PR #355, opened 2019-12-03
CAS-53 Fix/ChannelWatchRequest and Add ChannelWatchRequest Test Case

### LINKED ISSUE #352: Can't create a channel without ID.
I can't create channel using SDK.
After small investigation I've found possible source of issue:
1. `queryChannel` method in APIService takes ChannelQueryRequest argument as body for request;
2. Channel.watch method takes ChannelWatchRequest argument, fill it with `data` with members ids and pass to `queryChannel`;
3. https://github.com/GetStream/stream-chat-android/blob/master/library/src/main/java/com/getstream/sdk/chat/rest/request/ChannelWatchRequest.java#L14 - but `data` field in ChannelWatchRequest has no `@Expose` annotation, so it was ignored by Gson and wasn't included into request, this field is private, so it don't override ancestor's `data` field.

Simplified code used to send request
```
val data: HashMap<String, Any?> = hashMapOf("members" to members)
val client: Client = TODO()
val channel: Channel = client.channel("messaging", null, data)
channel.watch(ChannelWatchRequest(), object : QueryWatchCallback{
    override fun onSuccess(response: ChannelState?) {
        //success
    }

    override fun onError(errMsg: String?, errCode: Int) {
        //error
    }

})
```
Response:
```
D/OkHttp: --> POST https://XXX.stream-io-api.com/channels/messaging/query?api_key=XXX&user_id=XXX&client_id=XXX
D/OkHttp: Content-Type: application/json; charset=UTF-8
D/OkHttp: Content-Length: 44
D/OkHttp: {"presence":false,"state":true,"watch":true} ---> `data` wasn't included into request
D/OkHttp: --> END POST (44-byte body)
D/OkHttp: <-- 400 Bad Request https://XXX.stream-io-api.com/channels/messaging/query?api_key=XXX&user_id=XXX&client_id=XXX (750ms)
D/OkHttp: Server: fasthttp
D/OkHttp: Content-Type: application/json;charset=utf-8
D/OkHttp: Content-Length: 156
D/OkHttp: Cache-Control: no-cache
D/OkHttp: Access-Control-Allow-Origin: *
D/OkHttp: Access-Control-Allow-Headers: x-requested-with, content-type, accept, origin, authorization, x-csrftoken, x-stream-client, stream-auth-type
D/OkHttp: Access-Control-Allow-Methods: GET, POST, PUT, PATCH, DELETE, OPTIONS
D/OkHttp: Access-Control-Max-Age: 86400
D/OkHttp: {"code":4,"message":"GetOrCreateChannel failed with error: \"When using member based IDs specify at least 2 members\"","StatusCode":400,"duration":"0.00ms"}
```

adrian09h:
# Submit a pull request

- close #352 
- fix `ChannelWatchRequest`/ `ChannelQueryRequest`
- add Test Case for ChannelWatchRequest
- fix typo
- update CHANGELOGO.md


adrian09h:
> * I think you are testing the wrong client method; `channel.query` is the method that is not working here but you are testing `queryChannel` instead.

I have added `Channel`
I didn't call `channel.query`

library/src/test/java/com/getstream/sdk/chat/rest/ChannelWatchRequestTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ChannelWatchRequestTest {
+    @Test
+    void ChannelWatchReuestTest() {
```
elevenetc: Method name shouldn't begin with capital.

library/src/test/java/com/getstream/sdk/chat/rest/core/ClientTest.java
```diff
@@ -951,6 +953,95 @@ void queryChannelWithoutIdSuccessTest() {
         verify(storage).insertMessagesForChannel(channel, response.getMessages());
     }
 
+    @Test
+    void watchChannelWithIdSuccessTest() {
```
tbarbugli: test is called `watchChannelWithIdSuccessTest` but calls and tests `queryChannel`, it does not make much sense :) Please rename + add a direct test for `channel.watch` and `channel.create` to make sure this is not creating regressions

adrian09h: @tbarbugli , I have removed the problem code.
We have already `channel.watch` test case in `ChannelTest`

https://github.com/GetStream/stream-chat-android/blob/master/library/src/test/java/com/getstream/sdk/chat/models/ChannelTest.java#L112

adrian09h: @tbarbugli, added `createTest`, either.

PR #356, opened 2019-12-03
[WIP] Fix/Contain Markdown test case

adrian09h:
# Submit a pull request

Currently `containsMarkdown` is wrong. 
It misses many cases markdown text as well as returns true about non Markdown text.
For example `TestMarkDown`. it won't work.


library/src/test/java/com/getstream/sdk/chat/GetDeletedOrMentionedTextTest.java
```diff
@@ -81,4 +86,75 @@ void convertMentionTextTest() {
         text = "@@@Adrian a @@This is @A";
         assertEquals("@@@Adrian a @@This is @Adrian", StringUtility.convertMentionedText(text, userName));
     }
+
+    @Test
+    void containMarkdownTextText() {
```
tbarbugli: if you break this down into different tests life will be better:

```java
    @Test
    void containMarkdownTextPlain() {}

    @Test
    void containMarkdownTextBold() {}
```
```

adrian09h:
@tbarbugli ,  `StringUtility.containsMarkdown` completely messed up:

image

I will add detail test case soon.

adrian09h:
@andriizhumela , I have added `MarkDownTextTest` for Markdown Text in `update/markdown-test-case` BR.

https://github.com/GetStream/stream-chat-android/blob/update/markdown-test-case/library/src/test/java/com/getstream/sdk/chat/MarkDownTextTest.java

`containMarkdown` that you created must pass these test cases.
Can you please fix the function on this BR?

PR #361, opened 2019-12-04
CAS-55 Fix/detecting markdown

andriizhumela:
- Fixed detecting markdown attributes
- Added detecting deleted messages and mentions
Снимок экрана 2019-12-04 в 11 41 54

Result
Screenshot_20191204-115722

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
+++ b/library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
@@ -16,7 +16,7 @@ import java.util.regex.Pattern;
 public class StringUtility {
 
     private static final String TAG = StringUtility.class.getSimpleName();
-    private static final String MARKDOWN_REGEX = "(?m)[.{2}!{3}?{2},{2}>*$_}/:`=^~#+\\-@\\]]";
+    private static final String MARKDOWN_REGEX = "(?m)[.{2}!{3}?{2},{2}>*_}/:`=^~#+\\-@\\]]";
 
     public static String stringFromNumbers(int... numbers) {
         StringBuilder sNumbers = new StringBuilder();
--- a/library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
+++ b/library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
@@ -15,7 +15,6 @@ import java.text.DecimalFormat;
 public class StringUtility {
 
     private static final String TAG = StringUtility.class.getSimpleName();
-    private static final String MARKDOWN_PATTERN = "[*@_~#{\\[\\]\"'`|>]";
 
     public static String stringFromNumbers(int... numbers) {
         StringBuilder sNumbers = new StringBuilder();
--- a/library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
+++ b/library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
@@ -16,7 +16,10 @@ import java.util.regex.Pattern;
 public class StringUtility {
 
     private static final String TAG = StringUtility.class.getSimpleName();
-    private static final String MARKDOWN_REGEX = "(?m)[.{2}!{3}?{2},{2}>*_}/:`=^~#+\\-@\\]]";
+    private static final String MARKDOWN_REGEX = "<([a-zA-Z0-9. ]+)>|\\*{1,3}([a-zA-Z0-9. ]+)\\*{1,3}|_(\\w+)_|=(\\w+)=" +
+            "|~(\\w+)~|~(\\w+)|~ (\\w+)|@(\\w+)|#{1,6} (\\w+)|---( +)|\\*\\*\\*( +)|\\+\\+\\+( +)" +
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
```diff
@@ -47,7 +43,39 @@ public static boolean isEmoji(String message) {
...
+        }
+        return false;
+    }
+
+    public static boolean containsMention(String message) {
+        return message.contains("@");
+    }
+
+    public static boolean isMessageDeleted(Message message) {
+        return message.getType().equals("deleted");
```
adrian09h: Sometimes it works wrong. maybe backend issue.
Please use `return message.getDeletedAt() != null;` instead.

andriizhumela: [Fixed](https://github.com/GetStream/stream-chat-android/pull/361/commits/1b71bd45a9b4fb981cdcbf9a0cc038068cd78b71
)

adrian09h:
LGTM,  @tbarbugli , what do you think about new update?

library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
```diff
@@ -47,7 +43,39 @@ public static boolean isEmoji(String message) {
...
+                || message.startsWith("# ") || message.startsWith("## ") || message.startsWith("### ")
+                || message.startsWith("#### ") || message.startsWith("##### ")
+        ) {
+            return true;
+        }
+        return false;
+    }
+
+    public static boolean containsMention(String message) {
+        return message.contains("@");
```
adrian09h: @andriizhumela , it looks wrong. in case message is "Test@", it is not mention text.
The function name is weird.

adrian09h: I don't think the function needs in here since your containMarkdown method contains it already.

andriizhumela: @adrian09h no, we need that function (it was renamed to `isContainsMention`) because we do not check for the "@" symbol, we check mentions with `message.getMentionedUsers()`. 
Example in `getDeletedOrMentionedText()` function

Also i removed testcase @MarkDown, because @ it is not correct creteria.

andriizhumela: Screenshot_20191204-150851

adrian09h: You updated `containsMention` method correctly. (it was wrong before.)

library/src/test/java/com/getstream/sdk/chat/MarkDownTextTest.java
```diff
@@ -35,13 +35,22 @@ void containStrikeThroughTextTest(){
...
+        text = "### h3";
+        assertTrue(StringUtility.containsMarkdown(text));
+
+        text = "#### h4";
+        assertTrue(StringUtility.containsMarkdown(text));
+
+        text = "##### h5";
+        assertTrue(StringUtility.containsMarkdown(text));
+
+        text = "###### h6";
```
adrian09h: @andriizhumela , I have updated `containHeaderTextTest()`
in case `text = "###### h6"`, doesn't work.

image

Please refer this test link: https://markdown-it.github.io/
And then SDK doesn't provide all Markwon features.
https://github.com/GetStream/stream-chat-android#markdown-support

When you are testing, please consider above.

andriizhumela: @adrian09h I expanded the test suite and added a missing character check.

adrian09h: @andriizhumela , currently we don't provide all markwon features.

image

However you have added `containTypographicTextTest` and updated `containsMarkdown`.

SDK doesn't provide obviously `Typographic` feature currently.

Can you please remove the code for non-provided markwon feature from `Test` and `containsMarkdown`?

andriizhumela:
@tbarbugli  Currently we have one error with have false positives on stand alone symbols like "::"

adrian09h:
@tbarbugli , Got it. I will remove `TablePlugin` of Markwon.
What do you mean about `horizontal rules`?

Now we are supporting `CorePlugin`, `LinkifyPlugin`, `ImagesPlugin`, `StrikethroughPlugin`, `TablePlugin`.

What plugins do you want to remove in addition to `TablePlugin`?

library/src/test/java/com/getstream/sdk/chat/MarkDownTextTest.java
```diff
@@ -94,28 +148,4 @@ void containCodeTextTest(){
                 "```";
         assertTrue(StringUtility.containsMarkdown(text));
     }
-
-    @Test
-    void notContainMarkdownTextTest() {
```
adrian09h: @andriizhumela , `containsMarkdown` must pass this function.

PR #363, opened 2019-12-04
CAS-57 Fix/Compare-Model

### LINKED ISSUE #358: NullPointerException in Channel.equals
Obvious typo in `Channel` class - null pointer exception in https://github.com/GetStream/stream-chat-android/blob/master/library/src/main/java/com/getstream/sdk/chat/model/Channel.java#L395.

adrian09h:
# Submit a pull request
close #358


library/src/main/java/com/getstream/sdk/chat/model/Channel.java
```diff
@@ -392,9 +392,11 @@ public void setReactionTypes(Map<String, String> reactionTypes) {
 
     @Override
     public boolean equals(@Nullable Object obj) {
-        if (getClass() != obj.getClass()) {
+        if (obj == null) {
             return false;
         }
+        if (!(obj instanceof Channel))
```
elevenetc: That is still invalid actually: hashCode must be overridden also, because Sets ( and HashSets and other collections and util classes) use hashCode to have reference to the object. There is no need to write manually equals/hashCode methods. IDE can generate it (with cmd + N). Also comparison with the current object is missed, so this one is correct and generated version:
```java
@Override
public boolean equals(Object o) {
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;
    Channel channel = (Channel) o;
    return Objects.equals(cid, channel.cid);
}

@Override
public int hashCode() {
    return Objects.hash(cid);
}
```

library/src/main/java/com/getstream/sdk/chat/rest/Message.java
```diff
@@ -201,9 +201,12 @@ public void setUserID(String userID) {
 
     @Override
     public boolean equals(@Nullable Object obj) {
-        if (getClass() != obj.getClass()) {
+        if (obj == null)
```
elevenetc: Everything the same as about `Channel`.

library/src/test/java/com/getstream/sdk/chat/models/ChannelComparisonTest.java
```diff
@@ -0,0 +1,89 @@
...
+        // TODO: I'm wandering that after changing some parameters x and x_clone should be the same.
+        assertTrue(x.equals(x_clone));
+        assertTrue(x_clone.equals(x));
+
+        assertFalse(x.equals(y));
+        assertTrue(x.equals(y) == y.equals(x));
+
+        y.setId(x.getId());
+        z.setId(y.getId());
+        if ((x.equals(y) && y.equals(z)))
```
elevenetc: Having conditions in a test is confusing.
Moreover without else case it might not work as expected.

library/src/test/java/com/getstream/sdk/chat/models/MessageComparisonTest.java
```diff
@@ -0,0 +1,86 @@
...
+        assertTrue(x.equals(x_clone) == x_clone.equals(x));
+
+
+        assertFalse(x.equals(y));
+        assertTrue(x.equals(y) == y.equals(x));
+
+        y.setId(x.getId());
+        z.setId(y.getId());
+
+        if ((x.equals(y) && y.equals(z)))
```
elevenetc: Having conditions in a test is confusing.
Moreover without else case it might not work as expected.

library/src/test/java/com/getstream/sdk/chat/models/UserComparisonTest.java
```diff
@@ -0,0 +1,78 @@
...
+        y.setId(x.getId());
+        z.setId(y.getId());
+
+        if ((x.equals(y) && y.equals(z)))
+            assertTrue(x.equals(z));
+    }
+
+    @Test
+    void hashCodeTest() {
+        assertEquals(x.hashCode(), x_clone.hashCode());
```
elevenetc: All tests are more or less about similar, so we might miss some line while supporting all of them. And we will have more tests. I would think about some generic test.
We can wrap objects into interface during tests:
```kotlin
interface EqualsHashObject {
  fun setId(String id)
  fun getId():String
  fun equals(Object o)
  fun hashCode(): Int
}
```
And define one test:
```kotlin
class TestEqualsHashUtil {
  fun test(a: EqualsHashObject, aClone: EqualsHashObject, b: EqualsHashObject){
    assertEquals(a.hashCode(), aClone.hashCode());
    //...
    a.setId(b.getId());
    assertEquals(a.hashCode(), b.hashCode());
  }
}
```
So all tests will look like this
```kotlin
class TestEqualsHashObjects {
  fun testChannels() {
    val channelA = wrapChannel(Channel());
    val channelAClone = wrapChannel(Channel());
    val channelB = wrapChannel(Channel());
    TestEqualsHashUtil.test(channelA, channelAClone, channelB);
  }

  fun wrapChannel(channel: Channel): EqualsHashObject {
    return object: EqualsHashObject {
      override fun setId (id: String) {
        channel.setId(id);
      }
    }
  }
}
```

library/src/test/java/com/getstream/sdk/chat/models/ChannelComparisonTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+        x = new Channel(client, TEST_CHANNEL_TYPE, TEST_CHANNEL_ID_1);
+        x_clone = new Channel(client, TEST_CHANNEL_TYPE, TEST_CHANNEL_ID_1);
+        y = new Channel(client, TEST_CHANNEL_TYPE, TEST_CHANNEL_ID_2);
+        z = new Channel(client, TEST_CHANNEL_TYPE, TEST_CHANNEL_ID_3);
+    }
+
+
+    @Test
+    void equalsTest() {
```
tbarbugli: - tests should be easy to read
- test 1 thing with 1 test
- extract common test code if needed
- test names should make sense (easier when you only test 1 thing)

```
@Test
void equalsIsTransitive() {
    assertTrue(a.equals(b) && b.equals(c) && a.equals(a));
}

@Test
void equalsIsReflexive() {
    assertTrue(a.equals(a));
}

@Test
void equalsIsSymmetric() {
    // with a and be equal
    assertEqual(a.equals(b), b.equals(a));

    // with a and be not equal
    assertEqual(a.equals(b), b.equals(a));
}
```

library/src/test/java/com/getstream/sdk/chat/models/UserComparisonTest.java
tbarbugli: this is how you make unreadable tests; my suggestion is to write 1 complete set of tests for one type and then use ctrl-c and ctrl-v

elevenetc: There is nothing to read about testing equals/hashCode tests. Java developers already for 25 years write such kind of utils, libraries and frameworks:
- https://jqno.nl/equalsverifier/
- https://gist.github.com/rherrmann/2970842
- and many more other private

Then Autovalue was developed by Google and now Kotlin is solving about the same problem internally.

Copy/paste will sooner or later trigger a bug.

adrian09h:
@tbarbugli , @elevenetc  I have changed `ChannelComparisonTest` first.
After passing review this, will fix `Message` and `User` as well.

Issue #335, opened 2019-11-27
Attachment message not localized to the current set user's device


savked:
In our app, we have a system messages that notify users within the chat for some events.

This attachment looks like this:

```
{
  text: 'Seems you both travel to Australia and could meet each other between August 17, 2015 and August 21, 2015',
  system_event: 'system:match',
  attachments: [
    {
      type: 'event',
      title: 'system:match'
    },
    {
      type: 'i18n',
      text: "{\"en\":\"Seems you both travel to Australia and could meet each other between August 17, 2015 and August 21, 2015\",\"de\":\"Es scheint als ob ihr nach Australien reisen werdet und euch zwischen 17. August 2015 und 21. August 2015 treffen könntet\",\"es\":\"Parece que ambas viajan a Australia y que podrían coincidir entre 17 de agosto de 2015 y 21 de agosto de 2015\"}"
    }
  ]
}
```

I am displaying the message with the type "i18n" however the messages list does not read this text localized. I have my phone on German for example, but the message is in English.

savked:
I have just came to a realisation that the chat is actually showin the message under field `text` and it's obviously in English. However you can see on the screenshot provided that there is an "empty" attachment above the message that is actually attachment with the message library should display in language. What are my options here, what can we do about this?

![testupload](https://user-images.githubusercontent.com/22063573/69805520-1cfedd00-11e1-11ea-9065-8da4aabd67b0.png)

Chikrix:
@savked how're you able to send a system message in a conversation? I've not been able to successfully do this

tbarbugli:
@Chikrix at the moment the APIs only allow you to send system messages when updating a channel (ie. updating the channel data, add/remove members, ban/unban members)

Beside this; you can also add a custom field to the message (ie. system:true) and let the Android SDK render them differently

tbarbugli:
@savked the SDK does not have built-in i18n for message text; I think the easiest approach is to use a custom message class which picks the right version of the message text. @adrian09h perhaps you can provide some advice here?

savked:
How would I implement that in a ChannelActivity, any documentation that you can refer me to?

adrian09h:
@savked ,SDK doesn't provide  `i18n` , provides attachment types as default such as `image`, `video`, `file`, `giphy`, `link` and `product`.
So if you want to use custom type of attachment, it needs some consideration about custom fields of attachment additionally.

[...]

PR #364, opened 2019-12-04
Cleanup ViewHolders

adrian09h:
# Submit a pull request

## Cleanup ViewHolders

- setStyle
In case using `CustomMessageListViewHolder` we have to consider setStyle in many classes such as `CustomViewHolderFactory`, `CustomAttachListViewHolder`,...

It makes harder to integrate CustomMessageListViewHolder, and developer might miss it.
I have realized it while answer to Sava from issue #335 

- setViewHolderFactory
When we integrate CustomMessageListView, what we must consider is setCustomViewHolderFacotry()

So I have created it abstract function on BaseMessageListViewHolder.


elevenetc:
@adrian09h could you fix failed builds?

adrian09h:
@elevenetc , this PR has not failed builds but the travis issue is from `master`

adrian09h:
@elevenetc , I'm going to update ReadMe about explanation and example: https://github.com/GetStream/stream-chat-android/issues/335#issuecomment-561339775 in new PR.

PR #375, opened 2019-12-09
CAS-64 Fix/Duplicated attachment issue when edit message.

adrian09h:
# Submit a pull request


elevenetc:
@adrian09h It seems fixes consequences, but not the reason: `messageInputController.getSelectedAttachments()` should not return duplicates. Could you also add check inside `MessageInputController` that doesn't allow to add duplicates to `selectedAttachments`.

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.java
```diff
@@ -475,14 +475,14 @@ private void editMessage(Message message) {
...
+                    fileType.equals(ModelType.attach_mime_mp4))
                 messageInputController.onClickOpenSelectMediaView(null, message.getAttachments());
-            } else {
+            else
                 messageInputController.onClickOpenSelectFileView(null, message.getAttachments());
-            }
         } else {
             messageInputController.onClickOpenSelectMediaView(null, message.getAttachments());
         }
+        message.setAttachments(null);
```
tbarbugli: weird approach (hack?); looks like you are not cleaning up afterwards. What if you edit the message (attachments is reset to null) and then cancel the edit?

adrian09h:
@tbarbugli , @elevenetc , this regression has been recovered from custom-message-input PR. so I close this PR.

PR #377, opened 2019-12-10
CAS-65 Fix/Wrong scroll position

andriizhumela:
Fixed problem with chat scrolling after edit.


library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -206,10 +207,13 @@ public void setViewModel(ChannelViewModel viewModel, LifecycleOwner lifecycleOwn
...
                 Message lastMessage = entities.get(entities.size() - 1).getMessage();
                 if (lastMessage != null
-                        && scrolledBottom()
                         && justUpdated(lastMessage)) {
                     int newPosition = adapter.getItemCount() - 1;
-                    layoutManager.scrollToPosition(newPosition);
+
+                    Handler handler = new Handler();
+                    handler.postDelayed(() -> layoutManager.scrollToPosition(newPosition), 200);
+                    handler.sendEmptyMessage(0);
```
elevenetc: There is no need to create new `Handler` and send empty message. Each view has already handler internally and can deliver delayed message to UI thread, so all 3 lines might be just replaced with:
```java
postDelayed(() -> layoutManager.scrollToPosition(newPosition), 200);
```

PR #378, opened 2019-12-10
CAS-51 Feature/init to builder

andriizhumela:
Removed static init function.
Added Builder pattern.

For initialization chat:
```
ApiClientOptions apiClientOptions = new ApiClientOptions.Builder()
      .BaseURL(BuildConfig.API_ENDPOINT)
      .Timeout(BuildConfig.API_TIMEOUT)
      .CDNTimeout(BuildConfig.CDN_TIMEOUT)
      .build();

StreamChatStyle style = new StreamChatStyle.Builder()
      .setDefaultFont(R.font.lilyofthe_valley)
      .setDefaultFont("fonts/odibeesans_regular.ttf")
      .build();

StreamChat.Config configuration = new StreamChat.Config(context, BuildConfig.API_KEY);
        configuration.setApiClientOptions(apiClientOptions);
        configuration.setStyle(style);
        
StreamChat.init(configuration);
```


tbarbugli:
Replacing `init` with builder patter looks wrong; I think we should keep separate. ie.

1. you build all options
2. you init chat with the options


```java
new StreamChat.Builder(this)
      .setApiKey(BuildConfig.API_KEY)
      .setApiClientOptions(apiClientOptions)
      .setStyle(style)
      .build();
```

@elevenetc what do you think?

andriizhumela:
@elevenetc As for me static init it’s bad solution. But i can left init method as one of the variant.

elevenetc:
The problem that static init is anyway needed, although it might be hidden inside builder. And the reason is that components of SDK need to have static reference to the instance of `Client`. So there must be:
```kotlin
StreamChat.getClient();
```
We can hide static client initialisation inside builder, but that will look for developer that they can create multiple clients:
```
new StreamChat.Builder().build();
new StreamChat.Builder().build();
```
Which is not valid.

Lets rename `StreamChat.Builder` into `StreamChat.Config` and pass it into `init`:
```kotlin
val config = new StreamChat.Config();
config.setNotification(NotificationsManager);
config.setStyle(Style);)
StreamChat.init(new StreamChat.Config());
```
So it fits all requirements:
- no builder pattern which may tell developers that it's possible to build something several times
- no builder, so no internal static magic
- static init is easy
- StreamChat.getClient stays there, so no changes.

PR #381, opened 2019-12-11
CAS-62 Fix/Send button behavior

adrian09h:
# Submit a pull request

- fix send button behavior(prevent sending empty message)
- cleanup `MessageInputController`

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListAdapter.java
+++ b/library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListAdapter.java
@@ -9,6 +9,8 @@ import android.widget.ImageView;
 import android.widget.ProgressBar;
 import android.widget.TextView;
 
+import androidx.annotation.Nullable;
+
 import com.getstream.sdk.chat.R;
 import com.getstream.sdk.chat.model.Attachment;
 import com.getstream.sdk.chat.utils.Constant;
@@ -25,13 +27,17 @@ public class AttachmentListAdapter extends BaseAdapter {
     private List<Attachment> attachments;
     private boolean localAttach;
     private boolean isTotalFileAdapter;
+    private OnAttachmentCancelListener cancelListener;
 
-
-    public AttachmentListAdapter(Context context, List<Attachment> attachments, boolean localAttach, boolean isTotalFileAdapter) {
+    public AttachmentListAdapter(Context context,
+                                 List<Attachment> attachments,
+                                 boolean localAttach,
+                                 boolean isTotalFileAdapter) {
         this.attachments = attachments;
         this.layoutInflater = LayoutInflater.from(context);
         this.localAttach = localAttach;
         this.isTotalFileAdapter = isTotalFileAdapter;
+
     }
 
     @Override
@@ -112,6 +118,8 @@ public class AttachmentListAdapter extends BaseAdapter {
             holder.iv_large_file_mark.setVisibility(file.length()> Constant.MAX_UPLOAD_FILE_SIZE ? View.VISIBLE : View.INVISIBLE);
         } else {
             holder.tv_close.setVisibility(View.VISIBLE);
+            if (cancelListener != null)
+                cancelListener.onAttachmentCancel(attachment);
             if (!attachment.config.isUploaded()) {
                 holder.progressBar.setVisibility(View.VISIBLE);
                 holder.progressBar.setProgress(attachment.config.getProgress());
@@ -124,5 +132,9 @@ public class AttachmentListAdapter extends BaseAdapter {
         ImageView iv_file_thumb, iv_select_mark, iv_large_file_mark;
         TextView tv_file_title, tv_file_size, tv_close;
         ProgressBar progressBar;
+    }
+
+    public interface OnAttachmentCancelListener {
+        void onAttachmentCancel(Attachment attachment);
     }
 }
--- a/library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListAdapter.java
+++ b/library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListAdapter.java
@@ -9,8 +9,6 @@ import android.widget.ImageView;
 import android.widget.ProgressBar;
 import android.widget.TextView;
 
-import androidx.annotation.Nullable;
-
 import com.getstream.sdk.chat.R;
 import com.getstream.sdk.chat.model.Attachment;
 import com.getstream.sdk.chat.utils.Constant;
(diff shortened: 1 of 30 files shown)


library/src/main/java/com/getstream/sdk/chat/utils/StringUtility.java
```diff
@@ -87,6 +87,10 @@ public static String convertMentionedText(String text, String userName) {
         String[] names = text.split("@");
         String last = names[names.length - 1];
         return text.substring(0, text.length() - last.length()) + userName;
+    }
 
+    public static boolean isValidTextMessage(String text){
```
tbarbugli: I suggest we rename this into `isEmptyTestMessage` and add some basic javadoc

library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListAdapter.java
```diff
@@ -125,4 +137,8 @@ private void configureFileAttach(ViewHolder holder, Attachment attachment) {
         TextView tv_file_title, tv_file_size, tv_close;
         ProgressBar progressBar;
     }
+
+    public interface OnAttachmentCancelListener {
+        void onAttachmentCancel(Attachment attachment);
```
tbarbugli: rename it into `onCancel`

library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListAdapter.java
```diff
@@ -25,13 +27,18 @@
     private List<Attachment> attachments;
     private boolean localAttach;
     private boolean isTotalFileAdapter;
+    private OnAttachmentCancelListener cancelListener;
 
-
-    public AttachmentListAdapter(Context context, List<Attachment> attachments, boolean localAttach, boolean isTotalFileAdapter) {
+    public AttachmentListAdapter(Context context,
```
tbarbugli: what about we overload the constructor instead of breaking it?

library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentViewHolder.java
```diff
@@ -149,7 +149,7 @@ private void configFileAttach() {
                 attachments.add(attachment);
             }
         }
-        AttachmentListAdapter attachAdapter = new AttachmentListAdapter(context, attachments, false, false);
+        AttachmentListAdapter attachAdapter = new AttachmentListAdapter(context, attachments, false, false, null);
```
tbarbugli: see previous comment about constructor overloading (if OK this code is not needed)

library/src/main/java/com/getstream/sdk/chat/adapter/MediaAttachmentSelectedAdapter.java
```diff
@@ -48,16 +48,16 @@ public MyViewHolder onCreateViewHolder(ViewGroup parent,
...
 
     @Override
     public int getItemCount() {
         return attachments.size();
     }
 
-    public interface OnItemClickListener {
-        void onItemClick(int position);
+    public interface OnAttachmentCancelListener {
+        void onAttachmentCancel(Attachment attachment);
```
tbarbugli: rename into `onCancel`?

elevenetc:
@adrian09h It's possible to send empty message:
screen-record mp4

And the solution looks complicated. Especially with this:
```
new Handler().postDelayed(()->{
            if (!uploadingFile)
                setSendButtonState(true);
        },100);
```
It might trigger send button flickering on low speed connection.
Also `MessageInputController` is growing. It seems we need to extract upload part into some dedicated class which internally will be responsible for handling of uploading and move this methods out of `MessageInputController`:
````
uploadFile
getUploadFileCallBack
fileUploadSuccess
fileUploadFailed
fileUploading
cancelUploadingAttachment
```

Inside we can avoid delays and multiple checks. Uploading is just a queue:
```kotlin
class UploadManager {
  val queue = mutableList<Attachment>()
  fun upload(attachment) {
    queue.add(attachment)
    // call listeners that uploading is in progress, so button send might be hidden
  }
  private fun fileUploadSuccess(attachment) {
    queue.remove(attachment) 
    // if queue is empty call listeners that uploading is done, so button send might become visible
  }
}
```

library/src/main/java/com/getstream/sdk/chat/utils/UploadManager.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+public class UploadManager {
+
+    private static List<Attachment> queue;
+
+    public static void uploadFile(Channel channel, List<Attachment> attachments,
```
elevenetc: All these methods shouldn't be static. Static methods are almost only for set of pure util functions.
`UploadManager` should be just a regular class which keeps state inside and doesn't share it staticly.
```kotlin
val uploadManager = new UploadManager()
uploadManager. uploadFile(...)
```

library/src/main/java/com/getstream/sdk/chat/utils/MessageInputController.java
```diff
@@ -46,40 +44,54 @@
...
+            if (selectedAttachments == null)
+                selectedAttachments = new ArrayList<>();
+            selectedAttachments.add(attachment);
+        }else{
+            selectedAttachments.remove(attachment);
+        }
+    }
+
+    public boolean isUploadingFile(){
+        if (uploadManager == null) return false;
```
elevenetc: This null check is not needed. `uploadManager` is initialised in constructor. And if you mark `uploadManager` field as final then IDE will not even shoe NPE warning.

library/src/main/java/com/getstream/sdk/chat/utils/UploadManager.java
```diff
@@ -0,0 +1,125 @@
...
+        }else {
+            queue.remove(attachment);
+        }
+    }
+
+    public boolean isUploadingFile() {
+        return queue != null && !queue.isEmpty();
+    }
+
+    public void setQueue(List<Attachment> queue) {
```
elevenetc: If it's used only for clearing queue I would not allow to pass anything externally, but replace it actual method:
```java
public void reset() {
  queue.clear();
  //maybe callbacks that notifies that there is no uploading processes
}
```

library/src/main/java/com/getstream/sdk/chat/utils/UploadManager.java
```diff
@@ -0,0 +1,125 @@
...
+    private void fileUploading(Attachment attachment,
+                               Integer percentage,
+                               UploadFileListener fileListener) {
+        attachment.config.setProgress(percentage);
+        fileListener.onProgress(attachment);
+    }
+
+    public void updateQueue(Attachment attachment, boolean add){
+        if (queue == null)
+            queue = new ArrayList<>();
```
elevenetc: Might be initialised at class field:
```java
private List<Attachment> queue = new ArrayList<>();
```

library/src/test/java/com/getstream/sdk/chat/EmptyTextMessageTest.java
```diff
@@ -0,0 +1,35 @@
+package com.getstream.sdk.chat;
+
+import com.getstream.sdk.chat.utils.StringUtility;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class EmptyTextMessageTest {
```
elevenetc: null case is still not covered. It is crucial thing because right now it might trigger NPE and crash of an app.

PR #384, opened 2019-12-11
CAS-7 Feature/logger

andriizhumela:
Logging of information from the SDK is possible using the built-in StreamChatLogger logger.
StreamChatLogger allows you to adjust the log display level.
To create and configure a logger, you must call the following:
```
StreamLogger logger = new StreamChatLogger.Builder()
      .build ();
```
**Setting the logging level.**
StreamChatLogger provides the following logging levels:
- All
- Debug
- Warning
- Error
- Nothing

Setting the logging level is as follows:
```
StreamLogger logger = new StreamChatLogger.Builder()
      .loggingLevel(StreamLoggerLevel.INFO)
      .build ();
```
Setting the logging level will allow you to display the logs of the set level and all levels above. For example, if you set the Debug display level, you will also see warnings and errors.

If you want turn off logging errors and crashes - use `StreamLoggerLevel.NOTHING`

**Third party logging**
The StreamLoggerHandler interface is provided for third party logging.
```
StreamLoggerHandler loggerHandler = new StreamLoggerHandler () {
   @Override
   public void logI (@NonNull String className, @NonNull String message) {
      // display info logs here
   }

   @Override
   public void logD (@NonNull String className, @NonNull String message) {
      // display debug logs here
   }

   @Override
   public void logW (@NonNull String className, @NonNull String message) {
      // display warning logs here
   }

   @Override
   public void logE (@NonNull String className, @NonNull String message) {
      // display error logs here
   }

@Override
   public void logT (@NonNull Throwable throwable) {
       // display throwable logs here
   }

   @Override
   public void logT (@NonNull String className, @NonNull Throwable throwable) {
       // display throwable logs here
   }
};

StreamLogger logger = new StreamChatLogger.Builder ()
                .setLoggingHandler (loggerHandler)
                .build ();
```
Errors and crashes are logged by default.


elevenetc:
@andriizhumela using class as a tag by default may look convenient at first, but that just inconvenient in a set of cases:
- in a class which has too many responsibilities (and should be refactored ideally) different tags may be required
- in a generic listener interface which may be used in different places 
- during debugging or development of new features it's often convenient to set different tags within a single class.
- if we switch to Kotlin there will be cases without class environment at all

So I would replace it with `Object` and inside `getTag` check if it's a string instance or another class:
```kotlin
fun getTag(tag: Object): String {
  if(tag == null) return "null"
  if(tag is String) {
    return tag as String
  } else {
    return tag.class.simpleName()
  }
}
```

elevenetc:
Another thing is that we need to log exceptions. Currently they are either ignored or by default printed with `printStackTrace`. So there should be something like:
```kotlin
try {
  //...
} catch(t: Throwable) {
  logger.log(t)
}
```
And maybe with optional tag, because often it doesn't really matter which tag is used:
```kotlin
fun log(t: Throwable)
fun log(tag: Object, t:Throwable)
```

andriizhumela:
@elevenetc fixed PR

PR #389, opened 2019-12-13
CAS-60, CAS-67/Remove markdown from last message

adrian09h:
# Submit a pull request

 markdown on rendering last message in channels list makes some problems 
- doesn't display RTL langues for limit width(3dots to end). (CAS-67)
- disable click event of RecyclerView(ChannelListView)
- wrong displaying for some kind of mark down.(CAS-60)

So it is not suitable to use markdown on last message in channel list.


tbarbugli:
we are better off doing this from scratch

PR #390, opened 2019-12-16
CAS-61 Fix/Edit attachment behavior

adrian09h:
# Submit a pull request


library/src/main/java/com/getstream/sdk/chat/utils/MessageInputController.java
```diff
@@ -150,27 +151,29 @@ public void onClickCloseBackGroundView() {
...
+    private void configSelectAttachView(@Nullable List<Attachment> editAttachments, boolean isMedia) {
         binding.setIsAttachFile(!isMedia);
         
         selectedAttachments = (editAttachments != null) ? editAttachments : new ArrayList<>();        
         localAttachments = getAttachmentsFromLocal(isMedia);
         
         ((Activity) context).runOnUiThread(() -> {
-            if (!localAttachments.isEmpty()){
-                setAttachmentAdapters(isMedia);
+            if (editAttachments == null || editAttachments.isEmpty()){
```
tbarbugli: why don't you use `selectedAttachments` here?

library/src/main/java/com/getstream/sdk/chat/utils/MessageInputController.java
```diff
@@ -150,27 +151,29 @@ public void onClickCloseBackGroundView() {
...
+                if (!localAttachments.isEmpty()){
```
tbarbugli: please always avoid if-not when possible it makes it much easier to read code. ie

```java
if (selectedAttachments.isEmpty()) {
    Utils.showMessage(context, context.getResources().getString(R.string.stream_no_media_error));
    onClickCloseBackGroundView();
} else {
    setAttachmentAdapters(isMedia);
}
```

PR #392, opened 2019-12-16
CAS-68 add support for channel.hide with clear history + events

tbarbugli:
Requires a major release since it break channel, client and channel list view model `hide` methods


elevenetc:
@tbarbugli think we need to add second tab to Sample app to show hidden channels where we can make channels visible again. Otherwise we can't test `channel.visible`.

PR #393, opened 2019-12-18
CAS-56 Feature/channel property in search

andriizhumela:
Added `Channel` field to search message result.


library/src/main/java/com/getstream/sdk/chat/rest/controller/RetrofitClient.java
```diff
@@ -112,4 +114,10 @@ public static Retrofit getAuthorizedCDNClient(CachedTokenProvider tokenProvider,
                 .addConverterFactory(GsonConverterFactory.create(GsonConverter.Gson()))
                 .build();
     }
+
+    private static HttpLoggingInterceptor setupLoggingInterceptor() {
+        loggingInterceptor = new HttpLoggingInterceptor();
+        loggingInterceptor.level(HttpLoggingInterceptor.Level.BODY);
```
elevenetc: @tbarbugli do you remember if we disabled it deliberately some time ago or not?

andriizhumela: @elevenetc i can disable it if needed

elevenetc:
@andriizhumela lets add also search messages into Sample app. Currently `client.searchMessages` is not even used and we need to test search eventually as it's a basic feature of almost any messaging app. 

`ChannelActivity/ChannelViewModel` is too complicated right now, so it might be hard to add search there. So I would add search icon in Toolbar of `ChannelActivity` and open new activity `MessageSearchActivity`. Where we can have edit text and recycler view for messages.

andriizhumela:
@elevenetc Add to sample - Ok. But what i should to do with loggingInterceptor?

andriizhumela:
@elevenetc Added sample with search.
Would be nice add metadata to search response for correct pagination abbillity.
Also would be nice add somthing like scrollToMessageById feature to messageList. For auto scroll to message what we looked.

andriizhumela:
Added search to ChannelActivity.

PR #396, opened 2019-12-26
CAS-71 Fix/crash in attach document activity

andriizhumela:
Fix crash in case restore app from background. 
I can't recreate Client instance(too many static items and components), so we can just restart app from background without crash.


library/src/main/java/com/getstream/sdk/chat/view/activity/AttachmentDocumentActivity.java
```diff
@@ -62,7 +64,20 @@ private void configUIs() {
...
+        Client client = null;
+        BaseStorage storage = null;
+        String fileUrl = null;
+
+        try {
+            client = StreamChat.getInstance(this);
+            storage = client.getUploadStorage();
+            fileUrl = storage.signFileUrl(url);
+            webView.loadUrl("https://docs.google.com/gview?embedded=true&url=" + StreamChat.getInstance(this).getUploadStorage().signFileUrl(url));
+        } catch (NullPointerException ex) {
```
elevenetc: That's a workaround. There is clear reason for NPE: `getUploadStorage` returns null. So there could just basic check:
```kotlin
if(StreamChat.getInstance(this). getUploadStorage() == null){
  //load url
} else{
  finish();
}
```

sample/src/main/java/io/getstream/chat/example/ChannelActivity.java
```diff
@@ -54,6 +54,13 @@ protected void onCreate(Bundle savedInstanceState) {
         String channelID = intent.getStringExtra(ChannelListFragment.EXTRA_CHANNEL_ID);
         Client client = StreamChat.getInstance(getApplication());
 
+        // We have bad client instance
```
elevenetc: What `bad` means exactly?

library/src/main/java/com/getstream/sdk/chat/rest/core/Client.java
```diff
@@ -681,11 +682,15 @@ public void onWSEvent(Event event) {
     /**
      * the opposite of {@link #disconnectWebSocket()}
      */
-    public void reconnectWebSocket() {
-        if (getUser() == null) {
+    public synchronized void reconnectWebSocket() {
+        User user = getUser() != null ? getUser() : lastUser;
```
elevenetc: Why it's added here?
Because it's breaks existing logic:
1. developer calls `disconnect`
2. `getUser` returns null
3. `reconnectWebSocket` is not working anymore

With `lastUser` it maybe reconnected again. `reconnectWebSocket` is public and it can be called outside also.

PR #408, opened 2020-01-10
CAS-65, CAS-79 Fixed problem with wrong scroll.

andriizhumela:
- Fixed layout
- Merge with master


sample/src/main/res/layout/activity_channel.xml
```diff
@@ -10,7 +10,7 @@
             type="com.getstream.sdk.chat.viewmodel.ChannelViewModel" />
     </data>
 
-    <androidx.constraintlayout.widget.ConstraintLayout
+    <RelativeLayout
```
tbarbugli: AFAIK RelativeLayout performs much worse than ConstraintLayout https://developer.android.com/topic/performance/rendering/optimizing-view-hierarchies

In this case such difference seems relevant because we are changing the top container. @elevenetc @andriizhumela can we solve this and keep `ConstraintLayout`?

andriizhumela: @tbarbugli If there is no nesting relativelayout show much better performance. In case simple screens and if we have no any nesting better to use relativelayout.
https://android.jlelse.eu/constraint-layout-performance-870e5f238100

elevenetc: That's indeed not so obvious what layout performs better in particular environment. We need to measure somehow performance and track regressions. Will create a ticket to make a research.

PR #410, opened 2020-01-13
CAS-83 fix/Change okHttp settings

andriizhumela:
- keepAlive 45 seconds
- timeout 45 seconds
- Add tests for checking timeout


thesyncim:
as far as i can see okHTTP supports multiple timeout settings, and in this PR we are changing the wrong ones.
we should set :
connectTimeout  to 6 seconds (max time allowed dialling to the client)
readTimeout 6 seconds (max time allowed to read a response from the server)
keepAliveDuration to 45s

andriizhumela:
@thesyncim Could you check it now?

PR #414, opened 2020-01-20
CAS-87 Fix/mediaplayer

andriizhumela:
- Fix crash mediaplayer
- Add additional condition to check mime_type


library/src/main/java/com/getstream/sdk/chat/navigation/destinations/AttachmentDestination.java
```diff
@@ -121,21 +122,25 @@ private void loadFile(Attachment attachment) {
...
-        } else if (mimeType.equals("application/msword") ||
-                mimeType.equals(ModelType.attach_mime_txt) ||
-                mimeType.equals(ModelType.attach_mime_pdf) ||
-                mimeType.contains("application/vnd")) {
-
-            Intent intent = new Intent(context, AttachmentDocumentActivity.class);
-            intent.putExtra("url", url);
-            start(intent);
+        if (mimeType == null) {
+            StreamChat.getLogger().logE(this, "MimeType is null");
```
elevenetc: We should react somehow, otherwise user will see nothing when he tries to open attachment. Let's add at least `Toast` with a message like "Something went wrong. Unable to open attachment":
```kotlin
Utils.showMessage(context, context.getString(R.string.stream_attachment_invalid_mime_type));
```

Issue #418, opened 2020-02-02
ANR from Gson


jeffdgr8:
I have logs indicating ANRs caused by the Stream SDK doing JSON deserialization on the main UI thread. This should be done on a background thread.

```
java.util.HashSet.<init> (HashSet.java:105)
com.google.gson.internal.$Gson$Types.resolve ($Gson$Types.java:337)
com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields (ReflectiveTypeAdapterFactory.java:160)
com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create (ReflectiveTypeAdapterFactory.java:102)
com.google.gson.Gson.getAdapter (Gson.java:458)
com.google.gson.Gson.toJson (Gson.java:696)
com.google.gson.Gson.toJson (Gson.java:683)
com.google.gson.Gson.toJson (Gson.java:638)
com.google.gson.Gson.toJson (Gson.java:618)
com.getstream.sdk.chat.rest.StreamWebSocketService$1.run (StreamWebSocketService.java:88)
android.os.Handler.handleCallback (Handler.java:873)
android.os.Handler.dispatchMessage (Handler.java:99)
android.os.Looper.loop (Looper.java:193)
android.app.ActivityThread.main (ActivityThread.java:6971)
java.lang.reflect.Method.invoke (Method.java)
com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:493)
com.android.internal.os.ZygoteInit.main (ZygoteInit.java:865)
```

Looking at the code in `StreamWebSocketService`, it seems this routine gets [run every 30 seconds](https://github.com/GetStream/stream-chat-android/blob/master/library/src/main/java/com/getstream/sdk/chat/rest/StreamWebSocketService.java#L90).

It's actually getting posted to run again [here](https://github.com/GetStream/stream-chat-android/blob/master/library/src/main/java/com/getstream/sdk/chat/rest/StreamWebSocketService.java#L120) after a 1 second delay, with what seems to be a bug. Seems `mMonitor` should be posting `mMonitor` to run again after `monitorInterval`, right? Definitely don't want that running every 1 second on the main UI thread though. As is, `mMonitor` only gets run once after `startMonitor()` is called. But then `mHealthCheck` gets called twice every 30 seconds, each 1 second apart.

[...]

PR #419, opened 2020-02-02
Event thread looper

jeffdgr8:
# Submit a pull request

## CLA

- [X] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [X] The code changes follow best practices
- [X] Code changes are tested (add some information if not applicable)

Run health check and monitor routines on background event thread, instead of UI thread.

Fixes https://github.com/GetStream/stream-chat-android/issues/418

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
         espressoVersion = '3.2.0'
         glideVersion = '4.9.0'
         googleServiceVersion = '4.3.2'
-        gradleVersion = '3.5.2'
+        gradleVersion = '3.5.3'
         gradlePluginVersion = '2.1'
         crashlyticsVersion = '2.10.1'
         fabricVersion = '1.31.1'
--- a/library/src/main/java/com/getstream/sdk/chat/rest/EventHandlerThread.java
+++ b/library/src/main/java/com/getstream/sdk/chat/rest/EventHandlerThread.java
@@ -1,11 +1,20 @@
 package com.getstream.sdk.chat.rest;
 
+import android.os.Looper;
+
 class EventHandlerThread extends Thread {
     EventHandler mHandler;
+    private WebSocketService webSocketService;
 
     public EventHandlerThread(WebSocketService webSocketService) {
-        mHandler = new EventHandler(webSocketService);
+        this.webSocketService = webSocketService;
     }
 
-
+    @Override
+    public void run() {
+        super.run();
+        Looper.prepare();
+        mHandler = new EventHandler(webSocketService);
+        Looper.loop();
+    }
 }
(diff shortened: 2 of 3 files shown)


tschellenbach:
Sorry for the slow reply. We're currently working on splitting out our low-level client. WIP is here: https://github.com/getstream/stream-chat-android-client

Related: https://github.com/GetStream/stream-chat-android-client/blob/master/library/src/main/java/io/getstream/chat/android/client/socket/HealthMonitor.kt#L12

@elevenetc this pull request looks good in my opinion, what do you think?

Few questions while you're here:

* Are you using the custom views or working with the lowlevel client or livedata directly?
* Are you using RXJava or Livedata/databinding for your app?

jeffdgr8:
@tschellenbach thanks for following up. We are using LiveData and data binding in our app. We have some RxJava, but new code utilizes Kotlin Coroutines wherever possible. @Chikrix can add additional context. I know one view customization that we've needed is the ability to show a conversation as selected for displaying the conversation list and conversation detail on the screen at the same time on larger displays.

elevenetc:
@jeffdgr8 Thread `EventHandlerThread` is indeed useless currently. It just finishes and `EventHandler` is used as it is. 
Wrapping handler into thread fixes the issue, the only problem `Looper.loop()` never stops.
```
Looper.prepare();
mHandler = new EventHandler(webSocketService);
Looper.loop();
```
And we're not reusing `StreamWebSocketService`. So `EventHandlerThread` runs forever and keeps reference to instance of `StreamWebSocketService`. And SDK currently creates new instance of `StreamWebSocketService` when app goes into foreground and when socket is reconnected. Looper should be stopped.

jeffdgr8:
@elevenetc with your understanding of how this should be working, can you suggest edits such that it works as expected on a background thread?

`mMonitor` and `mHealthCheck` both reschedule themselves to be run again after an interval of 1 or 30 seconds. As I described in https://github.com/GetStream/stream-chat-android/issues/418, the current behavior is actually that `mHealthCheck` is getting called twice every 30 seconds, 1 second apart and on the main UI thread. So is this then happening even more often as `StreamWebSocketService` is recreated and schedules additional `mHealthCheck` routines?

elevenetc:
Health monitor should send ping message every N seconds while socket is connected, and try to reconnect when ping is not returned in a certain amount of time.
Current implementation is quite messy and eventually requires refactoring and it is done partly in the next version (`4.x.x`) of SDK which Thierry mentioned. We're still planning to support `3.x.x`, so I don't think it would be easy to avoid recreation of `StreamWebSocketService`, but monitor could be extracted from `StreamWebSocketService` as another service/thread which is:
1. created when `StreamWebSocketService.connect` is called
2. started when socket is connected
2.5 calls reconnection if ping is not returned
3. paused when socket is disconnected
4. destroyed when `StreamWebSocketService.disconnect` is called

tschellenbach:
closing this since the client has been moved to:
https://github.com/GetStream/stream-chat-android-client

PR #470, opened 2020-05-12
Migrate avatar view

JcMinarro:
**Create a new `AvatarView`**

This new view instead of add an extra `ImageView` for every avatar that is rendered, draws all of them on the same `ImageView`.
The new `AvatarView` is an `ImageView` instead of a `RelativeLayout`, flatting the view hierarchy inside of this UI componet.

On the other hand, I have added Coil-kt lib to load remote pictures. It has been added with an own wrapper that will help our SDK to not depend directly on the third-party library and allows us to switch to another one easier in the future if it is needed. The following PR will replace Glide calls to use the new wrapper (Adding needed methods to follow the same flow)

photo_2020-05-12_18-09-21
photo_2020-05-12_18-09-24
photo_2020-05-12_18-09-26


tschellenbach:
👍 

looks great. I'd like to add docs on getstream.io/chat/docs/ and some dokka/javadoc documentation. But we can do that later.

elevenetc:
Any particular reason why `Coil` was used? Some comparison with Glide, Picasso or anything else? Maybe some advantages like performance or binaries size or something else.
My concerns:
- the library is young, less than a year of development `=>` unclear future and support
- the library is less popular than Picasso for example: 3k stars vs 17k start `=>` less bugs fixed, smaller community. It's unclear how many production app used it actually.
- according to commits there is only one guy who is supporting it `=>` more chances that we will wait or fix ourself if there're bugs. Or developer might just stop supporting it.

The only better thing I see is that it doesn't use annotation processing.

Image caching and loading is an important part of scrollable views. We might not notice it first, but customers and users might notice slow scrolling for example. And such kind of things are address and fixed within old libraries like Picasso just because they exist for many years and huge amount of apps used it already.

JcMinarro:
@elevenetc 
It is true that it is younger than Picasso/Glide, and it is the higher reason why it is "less popular".
I have been using it on some apps and works really well. On the other hand, I have been doing some benchmark over our sample project (https://github.com/GetStream/stream-chat-android/tree/benchmark-coil-glide) and the performance is almost the same.
Comparing libs size:
Glide: ~573Kb
Coil: ~38Kb

I have not checked the number of methods of both libs, but on the Coil website, it shows it is smaller than Glide.
It is created on the top of Kotlin coroutines, and it is known better performance of coroutines than threads
It really easy to use because of the use of Extension Function and function as parameters to update the behavior to load/transform pictures.

BTW, to not pollute our code with external third-party libs and to be able to switch to another lib if it is needed (Or maybe allowing other devs to provide the lib they prefer to use, it will need some extra job) I have created this wrapper. Maybe we can extract third party libs to an external module to be sure we only use our code on all the SDK

tschellenbach:
Coil is very popular in the compose dev channel. Both Glide & Coil seem like viable options. Think it's best to use Coil and if we do find some issues in the future reevaluate.

PR #482, opened 2020-05-15
Fix default avatar with an empty name

### LINKED ISSUE #478: Crash when loading channel list
I updated to the latest version of the library `4.2.5` and changed from `AvatarViewGroup` to `AvatarView` and since then the app crashes when I try to load the channel list, I get the error
```
java.lang.IllegalArgumentException: width and height must be > 0
        at android.graphics.Bitmap.createBitmap(Bitmap.java:1033)
        at android.graphics.Bitmap.createBitmap(Bitmap.java:1000)
        at android.graphics.Bitmap.createBitmap(Bitmap.java:950)
        at android.graphics.Bitmap.createBitmap(Bitmap.java:911)
        at com.getstream.sdk.chat.view.AvatarView.createImageRounded(AvatarView.kt:87)
        at com.getstream.sdk.chat.view.AvatarView.createBitmap(AvatarView.kt:68)
        at com.getstream.sdk.chat.view.AvatarView$createBitmap$2.invokeSuspend(Unknown Source:12)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6669)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

JcMinarro:
When a channel or user has an empty name, the text to be rendered on the bitmap that contains the initials of the name has a size of 0. This size was used as the radio of the bitmap be center the name on it, but a bitmap can't have a size of 0, so it needs to have a default value on the case that user/channel doesn't have an associated name.

Fix #478

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/view/AvatarView.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/view/AvatarView.kt
@@ -82,7 +82,7 @@ class AvatarView @JvmOverloads constructor(
 		}
 		val textBounds = Rect()
 		paintText.getTextBounds(initials, 0, initials.length, textBounds);
-		val radius = max(textBounds.width(), textBounds.height()).takeUnless { it <= 0 } ?: MIN_RADIO_SIZE
+		val radius = max(textBounds.width(), textBounds.height()).takeUnless { it <= 0 } ?: MIN_RADIUS_SIZE
 		val bitmapSize = (radius * 2)
 		val output = Bitmap.createBitmap(bitmapSize, bitmapSize, Bitmap.Config.ARGB_8888)
 		val canvas = Canvas(output)
@@ -95,7 +95,7 @@ class AvatarView @JvmOverloads constructor(
 	}
 }
 
-private const val MIN_RADIO_SIZE = 100
+private const val MIN_RADIUS_SIZE = 100
 private const val FACTOR = 1.7
 private class AvatarDrawable(bitmaps: List<Bitmap>) : Drawable() {
 	private val avatarBitmaps = bitmaps.take(3)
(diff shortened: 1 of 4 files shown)


library/src/main/java/com/getstream/sdk/chat/view/AvatarView.kt
```diff
@@ -95,6 +95,7 @@ class AvatarView @JvmOverloads constructor(
 	}
 }
 
+private const val MIN_RADIO_SIZE = 100
```
samiuelson, line 98: So, this is a default radius size, right? 

1.  Its value should be defined in density-independent size, otherwise, it will have different actual size on different screen's densities. 

2. I'd call it MIN_RADIUS_SIZE or DEFAULT_RADIUS_SIZE

JcMinarro: 1-. It is not needed, in the end it is to create an image that will be rendered and resized on the drawing process according to with the Avatar size (can be changed on runtime)
2-. Yes, I have updated it ;)

samiuelson: Alright, thanks for explanation :)

PR #487, opened 2020-05-15
Fix for wrong lifecycle owner used to observer livedata

VGJohn:
viewLifecycleOwner is preferred over the fragment lifecycleOwner otherwise you can end up with multiple observers for each livedata


tschellenbach:
@samiuelson is revamping the sample app on our end. it needed a good refresh :)

PR #491, opened 2020-05-18
Migrate channelHeaderView

JcMinarro:
-. Migrate `ChannelHeaderView` to Kotlin
-. Decouple `ChannelHeaderView` from `ChannelViewModel`
-. Create a new Binding method to facilitate the use of the viewModel


library/src/main/java/com/getstream/sdk/chat/viewmodel/ChannelViewModelBinding.kt
```diff
@@ -0,0 +1,40 @@
...
+		LlcMigrationUtils.getChannelNameOrMembers(channel)
+				.takeUnless { it.isBlank() }
+				?.let { view.setHeaderTitle(it) }
+	})
+	anyOtherUsersOnline.observe(lifecycleOwner, Observer { view.setActiveBadge(it) })
+	view.onBackClick = {
+		if (isThread) {
+			resetThread()
+		} else {
+			view.closeActivity()
```
samiuelson, line 29: Imo, the logic of navigation should be extracted outside the view. For example, in the new sample app there is `androidx.navigation` used to navigate between Fragments. So, I'd like the view model to emit state `State.NavigateBack` and handle it in the fragment.

Eventually, ChannelViewModel will be rewritten in kotlin, so we can refactor this later on.

tschellenbach: Agreed. Also, new UI from Josh focuses more on replies than threads. (It's a more popular pattern on mobile)

tschellenbach:
It's a good step forward

library/src/main/java/com/getstream/sdk/chat/viewmodel/ChannelViewModelBinding.kt
JcMinarro: I agree. It was migrated on the same way it was previously implemented, to keep functionalities, but I have been thinking about to remove it and try to extract the other "icon actions" out of the view/ViewModel as well

JcMinarro: On the other hand, I think `ChannelViewModel` should be split, it has lot of logic that could be simplified and only to use what we really need for the View it renders

PR #495, opened 2020-05-19
Feature/new chat builder

samiuelson:
# Submit a pull request

## CLA

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [x] The code changes follow best practices
- [x] Code changes are tested (add some information if not applicable)

This PR redesigns `Chat.Builder` API and refactors `Chat` and `ChatImpl` accordingly. 
1. `Chat.Builder ` has a new API. 
a. It demands passing `ChatDomain` instance.
b. `Builder` throws informative exceptions in `build()` function

2. `Chat` has a new interface and is rewritten in Kotlin. 
a. `Chat` singleton is refactored to be stored in `Chat` companion object (`ChatImpl` previously) and it throws informative exception if it's being obtained before initializing.
b. `Chat` exposes now `setUser(user: User)` function. It is responsible for initializing ready/immutable instances of`ChatDomain` and `ChatClient` with the current user data.

3. `ChatImpl` is refactored to implement a new `Chat` interface. It is also cleaned from redundant code:
a. Removed references to global singletons
b. Implemented `setUser()` function

4. In sample app the global-level `Chat` setup is moved into `Application` class. The user-related logic remains in `ChannelsFragment`.


library/src/main/java/com/getstream/sdk/chat/Chat.kt
```diff
@@ -0,0 +1,76 @@
...
+        fun style(style: ChatStyle): Builder = apply {
+            this.style = style
+        }
+
+        fun navigationHandler(navigationHandler: ChatNavigationHandler) = apply {
+            this.navigationHandler = navigationHandler
+        }
+
+        fun build(): Chat {
+            if (chatDomain == null) {
```
JcMinarro, line 53: If `ChatDomain` is mandatory, it should be asked on the Builder constructor

library/src/main/java/com/getstream/sdk/chat/Chat.kt
```diff
@@ -0,0 +1,76 @@
...
+    fun setUser(user: User, token: String, callbacks: InitConnectionListener)
+
+    class Builder(private val context: Context) {
+        var navigationHandler: ChatNavigationHandler? = null
+        var style: ChatStyle = ChatStyle.Builder().build()
+        var urlSigner: UrlSigner = DefaultUrlSigner()
+        var markdown: ChatMarkdown = ChatMarkdownImpl(context)
+        var chatDomain: ChatDomain? = null
+
+        fun chatDomain(chatDomain: ChatDomain): Builder = apply {
```
JcMinarro, line 40: Do we need those methods?
The `chatDomain` and all other attributes are "public", I'm not sure we need it.
On the other hand, maybe we could do the builder "More Kotlin way" with default parameters in the constructor

library/src/main/java/com/getstream/sdk/chat/Chat.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.getstream.sdk.chat.style.ChatFontsImpl
+import com.getstream.sdk.chat.style.ChatStyle
+import com.getstream.sdk.chat.utils.strings.ChatStrings
+import com.getstream.sdk.chat.utils.strings.ChatStringsImpl
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.client.socket.InitConnectionListener
+import io.getstream.chat.android.livedata.ChatDomain
+import java.lang.IllegalStateException
+
+interface Chat {
```
JcMinarro, line 19: Do we really need this interface?
`ChatImpl` is a class that only implements this interface, doesn't hide anything else. This interface doesn't have any "boundary" reason neither it is implemented in different classes. Maybe we can simplify it

library/src/main/java/com/getstream/sdk/chat/Chat.kt
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    companion object {
+        private var instance: Chat? = null
+
+        @JvmStatic
+        fun getInstance(): Chat = instance?.let {
+            return it
+        } ?: throw IllegalStateException("Chat.Builder::build() must be called before obtaining Chat instance")
```
JcMinarro, line 74: Because we are returning the instance value, `let` expresion can be simplify, something like:
```suggestion
        fun getInstance(): Chat = instance ?: throw IllegalStateException("Chat.Builder::build() must be called before obtaining Chat instance")
```

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -109,8 +100,23 @@ public String getVersion() {
         return BuildConfig.BUILD_TYPE + ":" + BuildConfig.VERSION_NAME;
     }
 
-    void init() {
+    @Override
+    public void setUser(@NotNull User user, @NotNull String token, @NotNull InitConnectionListener callbacks) {
+        ChatClient.instance().setUser(user, token, new InitConnectionListener() {
```
JcMinarro, line 105: If we remove `ChatClient` as an attribute and delegate it on the singleton, we are assuming the user know that it needs to be initialized by himself. But it is a bit weird that I need to init another different class to be able to use `ChatDomain`'s methods. 
Why don't we ask for it on the `ChatDomain.Builder` or have any way to init it internally?

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -109,8 +100,23 @@ public String getVersion() {
...
+            @Override
+            public void onSuccess(@NotNull ConnectionData data) {
+                ChatDomain.instance().setCurrentUser(user);
```
JcMinarro, line 108: Why `ChatDomain` is mandatory on the Builder, but it is not injected into `Chat` instance?

library/src/main/java/com/getstream/sdk/chat/Chat.kt
samiuelson: Yes, I agree this would be a good idea to refactor this further and get rid of the interface because of the reasons you mentioned

library/src/main/java/com/getstream/sdk/chat/Chat.kt
samiuelson: Yes, we don't need them. Fixed. Java code will use setters now.

library/src/main/java/com/getstream/sdk/chat/Chat.kt
samiuelson: Simplified 👍

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
samiuelson: Imo, there's no reason for saving references to `ChatClient` and `ChatDomain` instances into any class fields (like before), as they are public, global singleton instances. For me, they are designed in a way that allows users to get them anywhere.

I think the one root problem here with the `Chat.setUser()` function is related to `ChatDomain`'s design. `ChatDomain` stores a duplicate `User` property. This should be redundant since it also uses `ChatClient` which has the user set.  So at the very beginning, `User` should be set only in one place and, imo, it should be the `ChatClient`.

library/src/main/java/com/getstream/sdk/chat/Chat.kt
samiuelson: Done

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
samiuelson: Since `ChatDomain` like `ChatClient` is designed to be public, immutable singletons I don't think there's any reason to store some more local references to the global one.

Anyway, I find it a bit weird because having those singletons `Chat.Builder` class acts rather as a validator and checks if `ChatClient` and `ChatDomain` are initialized properly. 

Since `Chat.setUser()` was supposed to be main action/entry point, maybe it would be better to validate initialization of `ChatClient` and `ChatDomain` inside it? 

Imo, this would be cleaner not to force passing `ChatClient` and `ChatDomain` to builder. But could be a bit misleading for the user. 

1. Any ideas for where to do the validation? 
2. Do you think we should force users to pass `ChatClient` and `ChatDomain` to builder just for validation purposes? 

This would require the user to perform the following operations in the correct order.
```
1. ChatClient.Builder().build()
2. ChatDomain.Builder().build()
3. Chat.Builder().build()
```
So this "correct order" requirement could be errorgenous. 

Another way would be to hide the `ChatClient` and `ChatDomain` initialization completely from the user and wrap it by `Chat.Builder().build()` or `Chat.setUser()`.

Calling for ideas here, @tschellenbach @elevenetc @tbarbugli

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
samiuelson: @tschellenbach need your opinion here whether it would be possible to instantiate `ChatDomain` without passing a `User` to it ☝️

library/src/main/java/com/getstream/sdk/chat/Chat.kt
elevenetc: Interface might be useful for customers who want to mock/fake chat for testing or other purposes. It might help also to migrate them to another provider if needed. 
And as a user of a library I find it much easier to read interface rather than stuffed with code implementation.

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
elevenetc: Yep, `Chat` shouldn't touch client at all. Layers must not be mixed.
`Chat` is a dumb ui. It should just consume data from domain and pass user events.
```kotlin
Chat {
  fun setUser(user:User) {
    domain.setUser(user)
  }
}

ChatDomain {
  fun setUser(user: User){
    // sync or replace
   client.setUser(user)
  }
}
```

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
elevenetc: Build everything at once without required `client` and `domain`, because `setUser` might be called later or never.
```kotlin
val chat = Chat.Builder(apiKey).build()
//much later or never
chat.setUser(user)
```

JcMinarro: Some points:
-. Our dependencies are "kind of singleton" but they are not immutable, whenever you call `build()` method, the instance is mutated
-. Because we have they  defined as `interface`, an user could create his own implementation, past it on the builder and later get an error that `ChatDomain` or `ChatClient` has not been initialized
-. On the "test side", if we are using a global instance instead of an injected one, we can't run our tests in parallel, or at least be sure our tests are working as we expect

I am not sure, but `ChatDomain` is a kind of wrapper of `ChatClient` and `Chat` a kind of wrapper of `ChatDomain`?

Maybe we can have a "common interface" (The same one that ChatClient uses), so the final user can use "whatever" implementation of `ChatClient` the want to use.
`ChatDomain` could implement it by delegating it to LLC, and `Chat` does the same, but delegating it on `ChatDomain`

elevenetc: @JcMinarro common interface would be huge and not really needed. Top layers should simplify bottom ones by reducing amount of methods and narrowing down responsibilities:
```
chat: top layer which provided UI related api
domain: data layer (repository) which provides everything about data
client: network layer which provides raw network api
```

So UI components like messages list don't need `client` methods or know anything about it. Messages list should retrieve it from data layer.

library/src/main/java/com/getstream/sdk/chat/Chat.kt
JcMinarro: A class can be mock/fake without an interface, it is not extremely needed.
If we have our public method documented with some Javadoc, and it is shown by the IDE, the user doesn't need to go to the code implementation.
On the other hand, if a user wants to go to the source code, it is because he wants to know what it is doing under the wood. Having an interface that hides it is an extra step the user needs to do

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
JcMinarro: Ok, I agree that "Top Layers" should simplify "Bottom ones".
So we need that a user that is using "UI Chat", doesn't need to know anything about `ChatDomain` neither `ChatClient`.

I mean, a user should be able to do something like:
```
val chatUI = ChatUI.Builder(....).build() 
chatUI.sendMessage(...)
```

Without "Knowledge" of the bottom layer

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
samiuelson: Looks good to me. We'd need to provide `ChatDomain.setUser()` then. 

cc @tschellenbach

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
samiuelson: I opt for:

1. Hiding `ChatDomain` initialization into `Chat.Builder.build()`:
```
Chat.Builder.build() {
    // initialization of ChatDomain 
    ChatDomain.Builder().build()
    // building actual Chat instance
    // ....
}
```
2. Hiding `ChatClient` initialization into `ChatDomain.Builder.build()`:
```
ChatDomain.Builder.build() {
    // initialization of ChatClient 
    ChatClient.Builder().build()
    // building actual ChatDomain instance
    // ....
}
```
3. Adding `setUser()` function into `ChatDomain`, so that:
```
fun Chat.setUser(user: User) {
    ChatDomain.instance.setUser(user)
}
```
wdut?

elevenetc: Yep

library/src/main/java/com/getstream/sdk/chat/Chat.kt
samiuelson: I'd park this topic atm.
This can be brainstormed and refactored later on, without modifying the sdk's API. 
Let's focus on the API in this PR.

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
elevenetc: Reviewed livedata library. Instance of `ChatDomain` deeply depends on current user, basically all internal components require current user. It is completely stateful and can't exists without current user. And that can't be refactored easily and fast.
Talked also with Thierry and it's done because database contains particular user related data. Which makes sense.

So for now it's probably better to create it when setUser is called. Later can be optimised. because deleting/creating database is async process. Async + stateful initialisation is tricky.
```kotlin
val chat = Chat.Builder(apiKey, context).build()
chat.setUser(user) // init client and domain
```

tschellenbach:
Three things I'd like to point out:

1. Customers will typically use a combination of low level client, chatDomain & chat UX. These are not full abstractions, you typically need all 3. (Or all 2 if you are not using our UX)

2. Both hiding the construction of chatDomain/client vs passing it are valid options

3. To switch user on the chatDomain you'll want to call chatDomain.disconnect() and create a new chatDomain instance using the builder.

@samiuelson feel free to pick the approach you like best at the moment, and perhaps reevaluate in a few months?

samiuelson:
Thanks for the details @tschellenbach.

I've redesigned `Chat.Builder` as discussed and reimplemented `Chat.setUser()`  accordingly. 
Ready for review.

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -109,8 +118,46 @@ public String getVersion() {
         return BuildConfig.BUILD_TYPE + ":" + BuildConfig.VERSION_NAME;
     }
 
-    void init() {
+    @Override
+    public void setUser(@NotNull User user,
+                        @NotNull String userToken,
+                        @NotNull InitConnectionListener callbacks) {
```
elevenetc, line 124: `InitConnectionListener` might be nullable. Developers can get connection event or error via `ChatObservable`.

library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -109,8 +118,46 @@ public String getVersion() {
...
+
+    private void disconnectChatDomainIfAlreadyInitialized() {
+        try {
+            final ChatDomain chatDomain = ChatDomain.instance();
+            BuildersKt.launch(GlobalScope.INSTANCE,
+                    Dispatchers.getIO(),
+                    CoroutineStart.DEFAULT,
+                    (scope, continuation) -> chatDomain.disconnect(continuation));
+        } catch(UninitializedPropertyAccessException e) {
+            Log.d("ChatImpl", "ChatDomain was not initialized yet. No need to disconnect.");
```
elevenetc, line 156: Client has logger. Better to use it for development logs, so customers could disable it, intercept it or filter by common tag prefix: `adb logcat | grep "Chat:"`
```kotlin
ChatLogger.get("tag").logD("log message")
```

PR #497, opened 2020-05-20
Feature/add channel header view to demo

JcMinarro:
-. Create a new ViewModel that only has the needed logic for `ChannelHeaderView`
-. Add `ChannelHeaderView` to the Sample App
-. Update navigation on the Sample App

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/viewmodel/ChannelHeaderViewModel.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/viewmodel/ChannelHeaderViewModel.kt
@@ -1,7 +1,6 @@
 package com.getstream.sdk.chat.viewmodel
 
 import androidx.lifecycle.LiveData
-import androidx.lifecycle.MutableLiveData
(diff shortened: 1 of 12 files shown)


library/src/main/java/com/getstream/sdk/chat/viewmodel/ChannelHeaderViewModel.kt
```diff
@@ -0,0 +1,32 @@
...
+	val anyOtherUsersOnline: LiveData<Boolean>
+
+	init {
+		val channelController: ChannelController = chatDomain.useCases.watchChannel.invoke(cid, messageLimit).execute().data()
+		members = channelController.members
+		channelState = map(channelController.channelData) { channelController.toChannel() }
+		anyOtherUsersOnline = map(members) { members ->
+			members.asSequence()
+					.filter { it.user != chatDomain.currentUser }
+					.any { it.user.online }
```
samiuelson, line 29: This little piece of custom logic could be unit-tested

PR #503, opened 2020-05-25
First iteration over MessageInputView

JcMinarro:
# First iteration over `MessageInputView`

## How was it implemented?
The previous `MessageInputView` was implemented using a mix of a lot of collaborators (ViewModel, Controller, Binding, UploadManager, PermissionChecker) without any architecture, View was being updated from different collaborators in different ways, observing ViewModel, changing the state of the "binding" instance, by external methods and so on. With this PR I want to standardize a little bit the flow to update the handling logic on this view.

## Final goal on the new MessageInputView architecture
The final idea of this refactor is to improve the view performance and make a clear boundary between different collaborators that will help devs to use it with and without ViewModel.

-. `MessageInputView` -> Stupid views that only propagates events and renders stuff on the screen
-. `MessageInputController` -> Internal logic of the view (feature logic like select files, show users to do a mention, show commands that the users can do....)
-. `MessageInputViewModel` -> Use the Livedata library to communicate with Stream platform, for instance, get "channel data", send a message and so on
-. `MessageInputViewModelBinding` -> Extension fun that communicate ViewModel with the "public API" of the View

## Changes on this PR
-. First of all, I have created a new ViewModel instead of using the `ChannelViewModel` that doesn't make sense for this view
-. Classes have been converted to Kotlin
-. ViewModel dependency has been removed on Controller (It is a dependency into the View yet, but will be removed from there as well)
-. Some methods have been created to configured the channel members and supported commands, instead to depend on a ViewModel call
-. UploadManager has been refactored and unit tested (It is not perfect yet and would be great to use any other android background process like a service or a Worker of WorkManager to do it, and maybe it could be on the LiveData lib)
-. A new method to bind the ViewModel to the view has been created, It is not completed yet, and to keep the functionality of this view the viewModel is been set to the view yet.


tschellenbach:
It would be good to also support replies in addition to editing a message and starting a thread. The new designs have a more WhatsApp style reply interface.

tschellenbach:
I like the idea of shipping the basic improvements and figuring out background image upload and background file uploading in the next release. That's probably one of the most annoying UX issues with the current lib though, so definitely something to prioritize in the near future.

library/build.gradle
```diff
@@ -134,7 +134,7 @@ dependencies {
     testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
     testImplementation "org.json:json:20190722"
     testImplementation "androidx.arch.core:core-testing:2.1.0"
-    testImplementation 'org.amshove.kluent:kluent-android:1.61'
+    testImplementation 'org.amshove.kluent:kluent:1.61'
```
elevenetc, line 137: Is it better than Truth from google or AssertJ?
Among 3 libraries we already have 3 assertion libraries and that's just not very convenient to switch between them.

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
```diff
@@ -0,0 +1,501 @@
...
+	 */
+	private val style: MessageInputStyle = MessageInputStyle(context, attrs)
+
+	/**
+	 * Fired when a message is sent
+	 */
+	private var messageSendListener: MessageSendListener? = null
+
+	/**
+	 * Permission Request listener
```
elevenetc, line 73: Quite useless comments. When field name is just splitted by spaces it's a bad sign.

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
```diff
@@ -0,0 +1,501 @@
...
+	 */
+	private var permissionRequestListener: PermissionRequestListener? = null
+
+	/**
+	 * The viewModel for handling typing etc.
+	 */
+	private lateinit var viewModel: MessageInputViewModel
```
elevenetc, line 80: Views shouldn't have reference to viewmodels

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
```diff
@@ -0,0 +1,501 @@
...
+	// region listeners
+	protected fun setMessageSendListener(manager: MessageSendListener?) {
+		messageSendListener = manager
+	}
+
+	fun setPermissionRequestListener(l: PermissionRequestListener?) {
+		permissionRequestListener = l
+	}
+
+	private fun generateMessageID(): String {
```
elevenetc, line 477: Message is generated at llc before it's sent.

library/src/main/java/com/getstream/sdk/chat/viewmodel/MessageInputViewModel.kt
```diff
@@ -0,0 +1,155 @@
...
+
+	@get:Deprecated("")
+	val channel: Channel
+		get() = channelController.toChannel()
+
+	fun getEditMessage(): LiveData<Message?> {
+		return editMessage
+	}
+
+	fun setEditMessage(editMessage: Message?) {
```
elevenetc, line 61: We need to stick to single way of interactive with viewmodels. Either with events and states (as its done by Samuel in `ChannelsViewModel`) or via methods. I prefere first one.

library/src/main/kotlinX/Primitives.kt
```diff
@@ -0,0 +1,5 @@
+inline fun Boolean.whenTrue(crossinline f: () -> Unit): Boolean = also { if (this) f() }
```
elevenetc, line 1: SDK extension functions should be internal.

library/build.gradle
samiuelson: Kluent is convenient to use in Kotlin tests because of its concise dsl. It is an extension. Under the hood, it uses `kotlin.test` library for assertions. 

Imo, Truth/AJ is a good choice for Java, and it is worth to try some more convenient syntax in Kotlin.

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
JcMinarro: Good to know, BTW, it was implemented on that way on the previous code, and it has not been refactored yet. The idea is that `MessageInputView` will have a listener to notify a new message needs to be sent, and the ViewModel (Or the implementation that the developer provides) will handle it and use the LLC to send the message, no message-id will be generated internally on the view

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
JcMinarro: It is the goal, but it is been used on lot of methods, so it will be achieved on following iterations

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
JcMinarro: It was done by "Java to Kotlin auto-convert" from the previous `MessageInputView.java`

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
elevenetc: Ok, clear.

JcMinarro:
> Not sure whats exactly will be resolved in next iteration, so let me know.

View and ViewModel should be completely decoupled, but I will need to change lot of code to achieve it. That is because I have created this initial PR, where I am decoupling `MessageInputController` of `ViewModel`. 
Next Iterations:
-. `MessageInputController` will only handle the internal logic of the view
-. `MessageInputController` shouldn't communicate with "Stream API" to send messages/files, it should be notified from the view to an external collaborator (Our ViewModel or whatever implementation the user want to use)
-. `MessageInputView` shouldn't know anything about ViewModel
-. `MessageInputView` needs to provide methods to inject the desired behavior to send messages/files, update a message and so on
-. Implement binding logic to setup our ViewModel with the previous behavior
-. Commands/Mentions adapter is not Type-Safe, it will be refactored as well

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
elevenetc: That's new. Ok.
But still if you see excessive comment just delete it. 1 second improvement makes easier future reviews.

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
elevenetc: Cool.

library/build.gradle
elevenetc: Have concerns because it seems a pet project of guy and not very popular. And in the worst case some assertion might miss requirement. So it's in the end `safety vs nice looking api`.
But ok, let's try it. I'll try it on llc as well. We can evaluate later whats fits better our projects.

samiuelson:
> It would be good to also support replies in addition to editing a message and starting a thread. The new designs have a more WhatsApp style reply interface.

We will need soon to work out communication between `MessageListViewModel` and `MessageInputViewModel`. I'm working on the `MessageListViewModel` atm.

`MessageListViewModel` is going to receive "Message edit" and "Message reply" events, it will inform its associated view to render UI and will need to inform the TextInput part to update its state as well.

cc @JcMinarro @tschellenbach

library/build.gradle
samiuelson: Alright 👍

PR #504, opened 2020-05-26
Refactor send message behavior

JcMinarro:
Refactor "send text message" behavior 
Typing state has been handled as well


library/src/main/java/com/getstream/sdk/chat/viewmodel/MessageInputViewModel.kt
```diff
@@ -102,6 +102,10 @@ class MessageInputViewModel(private val cid: String, private val chatDomain: Cha
 		chatDomain.useCases.sendMessage.invoke(message).execute()
 	}
 
+	fun sendMessage(message: String) {
+		chatDomain.useCases.sendMessage(Message(cid = cid, text = message)).execute()
```
elevenetc, line 106: `chatDomain` could be just called `domain`. It's the only entity with such name in the project and chat prefix is excessive.

library/src/main/java/com/getstream/sdk/chat/view/MessageInputView.kt
```diff
@@ -64,16 +66,22 @@ class MessageInputView(context: Context, attrs: AttributeSet?) : RelativeLayout(
...
 	 */
 	private var permissionRequestListener: PermissionRequestListener? = null
 
+	var onSendMessageListener: OnSendMessageListener = object : OnSendMessageListener {
+		override fun onSendTextMessage(message: String) {
+			throw IllegalStateException("MessageInputView#onSendMessageListener needs to be configured to send messages")
+		}
+	}
+
+	var typeListener: TypeListener = object : TypeListener {
```
elevenetc, line 80: The field should be internal. Otherwise if developer overrides this listener then keystroke/typing of domain is not called. And it would hard to notice why typing indicator stopped working in the app, but it might be that developer just wanted to add analytics in the app by overriding this listener.

In general it's better to avoid adding public methods and do it deliberately only with clear reason. Concise api is easer to use and support and breaking changes happen less often.

And also about difference between `handler` and `listener`. It should be safe to override listener, but not handler. Handler does something, listener just called without side effects. So listeners might be public, handler are internal if they do something (like calling domain). Handler overriding might be public in case we allow developer to override default behaviour.

```kotlin
class PublicEntity {
  var listener:Listener? = null
  internal var handler: Handler = {
    listener?.call()
    doSomeJob()
  }
}
```

The same applied to `onSendMessageListener`.
@samiuelson

samiuelson: Imo, `onSendMessageListener` is a listener here, not a handler. It emits "send button click" events publicly. 

We subscribe to those events in binding and propagate them to the viewmodel we provide. That is our ready to use building block. When a user wants to use MessageInputView without the viewmodel they are free to set the listener to their custom one. 

If we want to make it convenient to attach another listener (e.g. for analytics) one option would be to allow setting multiple listeners or another one would be to expose those events as a stream (allowing multiple observers to subscribe to it).

elevenetc: If developer overrides `onSendMessageListener` and using our viewmodel sending messages stops working without clear reason. And the only reason to understand it is either debugging or sending support message to us - that is bad. Also most people don't read documentation unfortunately. So in the current implementation it is still handler.

Hence we need to indicate clearly what's safe to override and whats not. Observable is an option indeed, although it requires a bit more code and support comparing to dumb and old listener and handler. And in case if we need to provide way to override default behaviour it's not enough to have events Observable, so there will be another handler logic.

So, there should be:
- common and clear way to indicate if it's safe to override listener (whether viewmodel is used or not)
- in some cases common and clear way to allow developers override default behaviour

JcMinarro: `typeListener` is a listener and it needs to be possible to be overrided by the dev. We are doing our view as simple as possible, they notify about the event and the dev is the one that needs to implement the behavior. The View are not connected with domain.
The `onSendMessageListener` is the same, the view notify that the user wants to send this message, and the dev needs to implement the action of send the message or delegate it in our `ViewModel`

samiuelson: I'd prefer to make a view which is responsible for emitting events purely. That's because of the overriding safety which you've mentioned. 

I'd accept having a handler calling listeners internally if you feel this is a simpler/robust approach. However, I can see a few drawbacks here as well: 
1. Introducing terms for and asking devs to get familiar with nomenclature (handler vs listener). 
2. It's hard to inject a handler/function implementation into the view and make it immutable since we have limited access to the view constructor. Previously, there was ViewModel reference passed to and controlled by the view internally. And if we want to separate view and data layers with this approach we'd need to inject the handler/function.

PR #506, opened 2020-05-28
Refactor commands and mentions behavior

JcMinarro:
# Refactor commands and mentions behavior

Improve suggestions when user wants to use a commands or create a mention to anyone else

BAACAgQAAxkBAAEnxY5ez6QNwlMq50bwFQuaf0LlSerbkwAC1wYAAofQgVL_RiDovvjDaBkE_1590666440 86

BAACAgQAAxkBAAEnxbRez6cdvNHxb29XYKhCKQ8p0iJmFgAC2AYAAofQgVL9plnJcEBBFBkE_1590667059 52

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/adapter/CommandsAdapter.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/adapter/CommandsAdapter.kt
@@ -14,7 +14,10 @@ import com.getstream.sdk.chat.view.BaseStyle
 import com.getstream.sdk.chat.view.MessageInputStyle
 import io.getstream.chat.android.client.models.Command
 
-class CommandsAdapter(private val style: BaseStyle) : ListAdapter<Command, CommandViewHolder>(
+class CommandsAdapter(
+		private val style: BaseStyle,
+		private val onCommandSelected: (Command) -> Unit
+) : ListAdapter<Command, CommandViewHolder>(
 		object : DiffUtil.ItemCallback<Command>() {
 			override fun areItemsTheSame(oldItem: Command, newItem: Command): Boolean = oldItem.name == newItem.name
 			override fun areContentsTheSame(oldItem: Command, newItem: Command): Boolean = oldItem == newItem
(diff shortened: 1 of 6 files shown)


tschellenbach:
Nice, definitely an improvement.

library/src/main/java/com/getstream/sdk/chat/adapter/CommandsAdapter.kt
```diff
@@ -0,0 +1,50 @@
...
+					StreamItemCommandBinding.inflate(LayoutInflater.from(parent.context), parent, false),
+					style,
+					onCommandSelected
+			)
+	override fun onBindViewHolder(holder: CommandViewHolder, position: Int) = holder.bind(getItem(position))
+}
+
+class CommandViewHolder(
+		private val binding: StreamItemCommandBinding,
+		private val style: BaseStyle,
```
elevenetc, line 37: `BaseStyle` by the way also should be eventually refactored. Currently it has lots of particular (not base at all) properties like `readStateAvatarWidth`. Not sure if base style is even needed.

library/src/main/java/com/getstream/sdk/chat/utils/MessageInputController.kt
```diff
@@ -338,90 +332,30 @@ class MessageInputController(private val context: Context,
...
+}
 
-	private fun setMentionUsers(namePattern: String) {
-		commands = members.map { it.user }
-				.filter { it.getExtraValue("name", "").contains(namePattern, true) }
-				.toMutableList()
-	}
-}
\ No newline at end of file
+private fun String.isCommandMessage() = COMMAND_PATTERN.matcher(this).find()
```
elevenetc, line 357: `private vals` on the top, `private funs` on the bottom: looks not aligned. Do you follow some style?

library/src/main/java/com/getstream/sdk/chat/adapter/CommandsAdapter.kt
JcMinarro: I tried to refactor it when I started to work on `AvatarView` but it was no easy, I achieve to remove the unneeded generic type it was using, but BaseStyle is being used in lot of places, so it will be refactored in future tasks. I think we can simplify it a lot, and try to do all configuration with themes, styles, and custom attributes.

Right now I am only following the previous implementation, keeping it as simple as possible

library/src/main/java/com/getstream/sdk/chat/utils/MessageInputController.kt
JcMinarro: In that case I am putting "Constant values" on top of the class and "Private fun extensions" on the bottom, It has been useful for me in previous jobs, but we can change it 
I'm not sure if you have any defined style, if not, maybe we can work on a common style and would be great to use some checkstyle tool like [ktlint](https://ktlint.github.io/)

PR #525, opened 2020-06-11
Feature / Message list: thread support

samiuelson:
# Description of the pull request

This PR refactored legacy "messages thread" feature:
* Decoupled MessageListView from the data layer
* Reimplemented thread data loading logic into Kotlin

Note: Message sending logic should be further refactored in MessageInputViewModel.

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
@@ -67,7 +67,7 @@ class MessageListViewModel(private val cid: String,
                 onThreadModeEntered(event.parentMessage)
             }
             is Event.BackButtonPressed -> {
-                if (currentMode is Thread) {
+                if (currentMode is Mode.Thread) {
                     onNormalModeEntered()
                 } else {
                     stateMerger.postValue(State.NavigateUp)
--- a/library/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import com.getstream.sdk.chat.view.messages.MessageListItemWrapper
+import exhaustive
 import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.client.models.Message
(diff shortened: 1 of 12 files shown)


tschellenbach:
Seems like a good step forward.

library/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -29,44 +28,50 @@ class MessageListViewModel(private val cid: String,
...
-            addSource(resultState) {
-                value = it
-            }
+            addSource(loading) { value = it }
+            addSource(messageListData) { value = State.Result(it) }
         }
     }
 
     fun onEvent(event: Event) {
         when (event) {
```
JcMinarro, line 50: Are all event types handled?
What about adding a new one? This method won't handle it.
If you don't want to return anything on the function and not use the return function body, I created an extension (`.exhaustive`) that you can add at the end of the `when` sentence and It will fail in compile-time in the future if there is any new type that is not handled here

library/src/test/java/com/getstream/sdk/chat/Mother.kt
```diff
@@ -100,11 +100,14 @@ fun createMessage(
...
 		text: String = randomString(),
-		createdAt: Date? = Date.from(Instant.now())
-): Message = Message(id, cid, text, createdAt = createdAt)
+		createdAt: Date? = Date.from(Instant.now()),
+		parentId: String? = null
+): Message = Message(id, cid, text, createdAt = createdAt, parentId = parentId)
 
 fun createMessageList(size: Int = 10) = (0..size).map { createMessage() }.toList()
 
+fun createThreadMessageList(size: Int = 10, parentMessageId: String = createMessage().id) = (0..size).map { createMessage() }.toList()
```
JcMinarro, line 109: `parentMessageId` is not used. Did you forget to use it?
On the other hand, because you are only to use the id of it, you don't need to create a complete message, you  can use `randomString()` and it will be faster while tests are run

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModelTest.kt
```diff
@@ -136,4 +159,40 @@ class MessageListViewModelTest {
...
+
+    @Test
+    fun `Should return from thread to normal mode on back click`() {
+        val viewModel = MessageListViewModel(CID, domain, client)
+        viewModel.onEvent(MessageListViewModel.Event.ThreadModeEntered(MESSAGE))
+        val states = viewModel.state.observeAll()
+
+        viewModel.onEvent(MessageListViewModel.Event.BackButtonPressed)
+
+        states.last() shouldBeInstanceOf  MessageListViewModel.State.Result::class
```
JcMinarro, line 171: `MessageListViewModel.State.Result` is a data class, and it helps us generating `.equals()` method and so on.
Would be great to check, not only that the last state is instance of this class, but also that the content of the result is the one we expect.
You can create a variable `expectedState` with the expected result and update the assertion with
```
states.last `should be equal to` expectedState
```

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModelTest.kt
```diff
@@ -136,4 +159,40 @@ class MessageListViewModelTest {
...
+    }
+
+    @Test
+    fun `Should display thread messages when thread mode entered`() {
+        val viewModel = MessageListViewModel(CID, domain, client)
+        val states = viewModel.state.observeAll()
+
+        viewModel.onEvent(MessageListViewModel.Event.ThreadModeEntered(MESSAGE))
+
+        states.last().run {
```
JcMinarro, line 191: Same here

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModelTest.kt
samiuelson: This is not possible (in reasonable time) to define complete, expected instances of `MessageListViewModel.State.Result`. However, I did add more specific assertions to validate all the data that are supposed to be tested here.

samiuelson: (Classes wrapped by `Result` are too complex)

JcMinarro: Ok, no problem, I didn't check the complexity of the wrapped class.
As we improve our `Mother` class, the easier it will be to create these values for us

PR #531, opened 2020-06-16
Feature/take picture

JcMinarro:
Implement behavior to take a picture or record a video by another installed app on the device.
To be able to share File/Uri between different apps we need to declare a `FileProvider`, it is using the final `packageName` as part of the `android:authorities` attribute to be able to use the one defined on the Stream library into multiple apps.
With the new method `registerForActivityResult()` we are able to register callbacks that handles the "Activity Result" without needed to add logic to the activity/fragment where the activity was started

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/CaptureMediaContract.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/CaptureMediaContract.kt
@@ -1,0 +1,56 @@
+package com.getstream.sdk.chat
+
+import android.app.Activity
+import android.content.ComponentName
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.provider.MediaStore
+import androidx.activity.result.contract.ActivityResultContract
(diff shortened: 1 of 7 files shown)


library/src/main/java/com/getstream/sdk/chat/CaptureMediaContract.kt
```diff
@@ -0,0 +1,56 @@
...
+				flags = Intent.FLAG_GRANT_WRITE_URI_PERMISSION
+				component = ComponentName(it.activityInfo.packageName, it.activityInfo.name)
+				`package` = it.activityInfo.packageName
+			}
+		}
+	}
+
+	override fun parseResult(resultCode: Int, intent: Intent?): File? =
+			(pictureFile?.takeIf { it.exists() && it.length() > 0 }
+					?: videoFile?.takeIf { it.exists() && it.length() > 0 })
```
samiuelson, line 53: `{ it.exists() && it.length() > 0 }` can be extracted and reused

Issue #521, opened 2020-06-10
custom design when the channle list is empty


abhirup-patra:
unable to check whether the channel list is empty or not to add a custom empty image view

samiuelson:
@abhirup-patra In the new version of SDK it's going to be easy to subscribe to `ChannelsViewModel.State.NoChannelsAvailable` event and display custom view in such a case. 

You can see sample usage here: https://github.com/GetStream/stream-chat-android/pull/533

The change will be available soon in the beta.

abhirup-patra:
is it implemented in this beta 5 release? because i am not able to find that in this beta 5 release as well

samiuelson:
Not yet. It will be released in one of the next releases along with a new sample app, revamped view models, and bunch of technical improvements.

JcMinarro:
Hello @abhirup-patra 
As Samuel said, we have added an option to show an EmptyState View when no channels on #533 
It is already part of the version [4.2.11-beta-6](https://github.com/GetStream/stream-chat-android/releases/tag/4.2.11-beta-6), not a final release yet, we will update #554 when the final version is ready

abhirup-patra:
may i know how to implement that? @JcMinarro

JcMinarro:
Hello @abhirup-patra we are working to improve our documentation, it will be ready when we release the final version.
In the meantime you can check the sample apps in the repo, where we integrate the different SDK Views.
The new `ChannelsView` has two methods that allow you to define a customView for LoadingState and a customView for EmptyState.

You should do something like:
```
val myEmptyStateView = generateMyEmptyStateView()
channelsView.setEmptyStateView(myEmptyStateView)
```

With that, it should show your emptyState when no channels are found

abhirup-patra:
how to check whether the channel is empty or not?

JcMinarro:
If you are binding the UI to the ViewModel with the following method:
```
ChannelListViewModelBindingKT.bindView(channelsViewModel, channelsView, lifecycleOwner)
```

It will do it automatically. If you are not using the ViewModel, you can handle it by yourself calling:
```
channelsView.hideEmptyStateView() // Hide the empty state
channelsView.showEmptyStateView() // Show the empty state
channelsView.hideLoadingView() // Hide the loading state
channelsView.showLoadingView() // Show the loading state
```

PR #533, opened 2020-06-17
Display empty state label when there are no channels available

samiuelson:
This PR:
* Displays empty state view when there are no channels available

Issue: https://github.com/GetStream/stream-chat-android/issues/521

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelListViewModel.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelListViewModel.kt
@@ -52,10 +52,10 @@ class ChannelsViewModelImpl(
         val queryChannelsController = chatDomain.useCases.queryChannels(filter, sort).execute().data()
         queryChannelsController.run {
             channelsData = map(channels) {
-                if (it.isNotEmpty()) {
-                    ChannelsViewModel.State.Result(it)
-                } else {
+                if (it.isEmpty()) {
                     ChannelsViewModel.State.NoChannelsAvailable
+                } else {
+                    ChannelsViewModel.State.Result(it)
                 }
             }
             loadingMoreData = map(loadingMore) { ChannelsViewModel.State.LoadingNextPage(it) }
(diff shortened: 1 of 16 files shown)


sample/src/main/kotlin/io/getstream/chat/sample/feature/channels/ChannelsFragment.kt
```diff
@@ -35,21 +34,30 @@ class ChannelsFragment : Fragment(R.layout.fragment_channels) {
...
             }
         })
+
         ChannelsViewModelImpl().apply {
             bindView(channelsList, this@ChannelsFragment)
             state.observe(viewLifecycleOwner, Observer {
                 when (it) {
                     is ChannelsViewModel.State.Loading -> {
                         loadingProgressbar.visible(it.isLoading)
+                        emptyStateLabel.visible(false)
```
JcMinarro, line 50: Would be great if we have a customizable emptyView into the `ChannelListView`, then all this logic could be handled into the view, and the devs only need to design that view

library/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelListViewModel.kt
```diff
@@ -50,7 +51,13 @@ class ChannelsViewModelImpl(
     init {
         val queryChannelsController = chatDomain.useCases.queryChannels(filter, sort).execute().data()
         queryChannelsController.run {
-            channelsData = map(channels) { ChannelsViewModel.State.Result(it) }
+            channelsData = map(channels) {
+                if (it.isNotEmpty()) {
```
JcMinarro, line 55: "If Not" could be a bit confusing, what do you think about switching it?

sample/src/main/kotlin/io/getstream/chat/sample/feature/channels/ChannelsFragment.kt
samiuelson: Sounds good

PR #544, opened 2020-06-23
Feature Edit Message

JcMinarro:
Implement EditMessage behavior from `MessageInputView`
The coupled ViewModel has been removed from the View as well


samiuelson:
Looks good. It would be also good to unit test MessageInputViewModel.

Issue #433, opened 2020-03-22
Should `MessageListItem#copy` deep copy?


psylinse:
When I use `MessageListItem#copy` I expected a deep copy. I was confused when I started manipulating the nested `message` instance and caused a mutation bug, since this instance is shared with the object I copied from. I'm forced to do an awkward clone myself if I need to manipulate the message, however I can't set some of the attributes that `copy` does since they're private. Is it possible to have the `.copy` do a `.copy` on `message` (and the other attributes, such as `positions`)?

Here is an example of some awkward work around code:
```kotlin
val decryptedMessage = messageListItem.message.copy()
if (messageListItem.isMine) {
    decryptedMessage.text = eThree.decrypt(decryptedMessage.text)
} else {
    val publicKey = eThree.lookupPublicKeys(decryptedMessage.user.id)
        .get()[decryptedMessage.user.id]
    decryptedMessage.text = eThree.decrypt(decryptedMessage.text, publicKey)
}

val decryptedItem =
    MessageListItem(decryptedMessage, messageListItem.positions, messageListItem.isMine)
```

Here is my source: https://github.com/psylinse/stream-android-encrypted-chat/blob/master/android/app/src/main/java/io/getstream/encryptedchat/EncryptedMessageViewHolder.kt

tschellenbach:
yeah, easily fixed using a data class on messageListItem

samiuelson:
Thank you for reporting this, we are currently pushing on a new major release which will not be affected by this problem. You can already try the beta release and subscribe to this issue to stay up-to-date to the release https://github.com/GetStream/stream-chat-android/issues/554

PR #547, opened 2020-06-24
Refactor MessageListItem to sealed class

samiuelson:
## Description:

1. Refactors MessageListItem model class to sealed a class and set of data classes representing different message list items:
* DateSeparatorItem
* MessageItem
* TypingItem
* ThreadSeparatorItem

2. Adjusts all the dependent classes to the refactored data model

3. Fixes: https://github.com/GetStream/stream-chat-android/issues/433


tschellenbach:
Very clean

PR #559, opened 2020-07-06
[Fix] Loading indicator

samiuelson:
Loading indicator was not being hidden on a `Loading.isLoading = false` event

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelListViewModelBinding.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelListViewModelBinding.kt
@@ -13,8 +13,12 @@ fun ChannelsViewModel.bindView(view: ChannelsView, lifecycle: LifecycleOwner) {
                 view.hideLoadingView()
             }
             is ChannelsViewModel.State.Loading -> {
-                view.hideEmptyStateView()
-                view.showLoadingView()
+                if (it.isLoading) {
+                    view.hideEmptyStateView()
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelListViewModelBinding.kt
```diff
@@ -13,8 +13,12 @@ fun ChannelsViewModel.bindView(view: ChannelsView, lifecycle: LifecycleOwner) {
                 view.hideLoadingView()
             }
             is ChannelsViewModel.State.Loading -> {
-                view.hideEmptyStateView()
-                view.showLoadingView()
+                if (it.isLoading) {
+                    view.hideEmptyStateView()
+                    view.showLoadingView()
+                } else {
```
JcMinarro, line 19: Why do we have a `loading` state that is "Not loading"? :thinking: 
It seems a bit confusing
Maybe we could have an extra state if it is needed and split this logic

samiuelson: Loading event comes from livedata. `QueryChannelsControllerImpl.loading: LiveData<Boolean>` emmits true/false.

samiuelson: `QueryChannelsControllerImpl.loading: LiveData<Boolean>` is redundant for the need of app here. Simplified.

PR #562, opened 2020-07-08
App Distribute Process

JcMinarro:
Implement a workflow to distribute the sample App with Firebase.
I have added a plugin that increase the version of the app to be able to identify which version testers are using. This plugin use the git history to create a version code/name and be sure which commit they are testing.


tschellenbach:
👏

PR #568, opened 2020-07-15
Fix channel name validating

samiuelson:
* Fix validation of channel name
* Move network call to background


library/src/main/java/com/getstream/sdk/chat/viewmodel/CreateChannelViewModel.kt
```diff
@@ -30,31 +35,31 @@ class CreateChannelViewModel(
...
+                result.isError -> {
                     stateMerger.postValue(State.BackendError)
                 }
             }
         }
     }
 
     private fun validateChannelName(channelNameCandidate: String): Boolean {
-        return channelNameCandidate.isNotEmpty()
+        return channelNameCandidate.isNotEmpty() && channelNameCandidate.matches(Regex("^!?[\\w-]*\$"))
```
JcMinarro, line 62: Using a RegEx should be enought, `isNotEmpty()` call is redundant.
On the other hand, could we extract RegEx into a constant that describes what this RegEx is used for?

library/src/test/java/com/getstream/sdk/chat/viewmodel/CreateChannelViewModelTest.kt
```diff
@@ -42,43 +54,52 @@ internal class CreateChannelViewModelTest {
...
     }
 
     @Test
     fun `Should inform about validation error`() {
         whenever(createChannelResult.isError) doReturn false
         whenever(createChannelResult.isSuccess) doReturn true
-        val viewModel = CreateChannelViewModel(domain = chatDomain, client = chatClient)
-        val states = viewModel.state.observeAll()
-        val channelNameCandidate = ""
+        val channelNameCandidates =
```
JcMinarro, line 71: Can't we use `ParameterizedTest`?
If we use it instead of a list and anyone fails, all cases are run and show which ones are failing, using a list and iterating over them into the test method, it will fail on the first one but not check the rest of them

library/src/main/java/com/getstream/sdk/chat/viewmodel/CreateChannelViewModel.kt
samiuelson: This regexp accepts an empty string in Kotlin, so I'll leave isNotEmpty() check. 
I'll extract regex to give it a name.

PR #573, opened 2020-07-17
Feature/cas 162 night mode on test app

emitenma:
# Submit a pull request

If you want to skip checking the kotlinization of the 3 classes i converted, just review by commit, the changes i added for nightmode (simple ones).
Took some time to figure out how the binding + style classes worked all together.

## CLA

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [ ] The code changes follow best practices
- [ ] Code changes are tested (add some information if not applicable)


library/src/main/java/com/getstream/sdk/chat/view/ChannelHeaderViewStyle.kt
```diff
@@ -0,0 +1,104 @@
...
+    val optionsButtonHeight: Int
+    val isActiveBadgeShow: Boolean
+
+    internal val backButtonBackground: Drawable?
+    internal val optionsButtonBackground: Drawable?
+    private val offlineText: String?
+    private var channelWithoutNameText = ""
+
+    fun getChannelWithoutNameText(): String {
+        return if (!TextUtils.isEmpty(channelWithoutNameText)) channelWithoutNameText else context.getString(R.string.stream_channel_unknown_title)
```
JcMinarro, line 30: Because Kotlin provide some extension function over String, we could do something like
```suggestion
        return channelWithoutNameText.takeIf(String::isNotBlank) ?: context.getString(R.string.stream_channel_unknown_title)
```

library/src/main/java/com/getstream/sdk/chat/view/ChannelHeaderViewStyle.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fun getBackButtonBackground(): Drawable {
+        return backButtonBackground ?: getDrawable(R.drawable.stream_arrow_left)
```
JcMinarro, line 34: All this method that only have one line can be converted to "Expression Body"

library/src/main/java/com/getstream/sdk/chat/view/ChannelHeaderViewStyle.kt
emitenma: Agree, i still need to get used to this one, but i like it!

PR #584, opened 2020-08-04
Init Fresco library

JcMinarro:
Fresco library needs to be initialized before starting to use the views they provide.
I'm using a ContentProvider to start it because it is instantiated before any other part of the code


tschellenbach:
great

PR #585, opened 2020-08-04
Use proper CID from message when add/remove reactions

JcMinarro:
The cid property of Message entity is not initialized sometimes, we need to standardize the structure of our entities to be sure we are working always with the right data. In the meantime we need to add this kind of check to our code


tschellenbach:
I noticed that some events send the message without the CID. (the event has the cid, but the message doesn't).
There are probably more edge cases. This is a good fix for now, but we probably want to patch this at a higher level as well (either the LLC or livedata)

PR #602, opened 2020-08-29
Fix uploading files and capturing images on Android >= 10:

adasiewiczr:
Android 10 introduced a new storage paradigm - [scoped storage](https://developer.android.com/preview/privacy/storage). Starting from Android 10 we are unable to operate on external Files directly and that requires making copies inside internal app storage when sending messages. What exactly changed:
- Media files (Images, Videos, Audios) - can be accessed and resolved using ContentResolver
- Other files and documents - can be accessed using SAF instead of iterating through the whole ExternalStorage
Ref: https://developer.android.com/training/data-storage
Moreover, I've changed the folder in which captured images and videos are saved and I've added a fallback to internal app cache because external storage might not exist on some devices (e.g. emulators)


library/src/main/java/com/getstream/sdk/chat/CaptureMediaContract.kt
```diff
@@ -22,12 +23,12 @@ class CaptureMediaContract : ActivityResultContract<Unit, File>() {
...
     override fun createIntent(context: Context, input: Unit?): Intent {
         val takePictureIntents =
-            File(context.externalCacheDir, createFileName("STREAM_IMG", "jpg")).let {
+            File(context.getExternalFilesDir(Environment.DIRECTORY_PICTURES) ?: context.cacheDir, createFileName("STREAM_IMG", "jpg")).let {
                 pictureFile = it
                 createIntentList(context, MediaStore.ACTION_IMAGE_CAPTURE, it)
             }
         val recordVideoIntents =
-            File(context.externalCacheDir, createFileName("STREAM_VID", "mp4")).let {
+            File(context.getExternalFilesDir(Environment.DIRECTORY_PICTURES) ?: context.cacheDir, createFileName("STREAM_VID", "mp4")).let {
```
JcMinarro, line 31: Shouldn't we use `Environment.DIRECTORY_MOVIES` instead?

library/src/main/java/com/getstream/sdk/chat/adapter/MediaAttachmentSelectedAdapter.java
```diff
@@ -81,19 +81,13 @@ public void bind(AttachmentMetaData attachment, final OnAttachmentCancelListener
...
-                if (file.exists()) {
-                    Uri imageUri = Uri.fromFile(file);
-                    Glide.with(context)
-                            .load(imageUri)
-                            .into(binding.ivMedia);
-                }
-            } else if (attachment.isUploaded()) {
+            if (attachment.file != null) {
+                Glide.with(context)
+                        .load(Uri.fromFile(attachment.file))
```
JcMinarro, line 86: Why are we using `attachment.uri` into `MediaAttachmentAdapter` but we are building the uri by ourselve here?
Could we use the same method in both part of the code?

library/src/main/java/com/getstream/sdk/chat/model/AttachmentMetaData.java
```diff
@@ -1,41 +1,46 @@
...
-    public boolean isUploaded() {
-        return attachment != null;
+    public AttachmentMetaData(@NonNull Uri uri, String mimeType) {
+        this.uri = uri;
+        this.mimeType = mimeType;
     }
 
     public boolean isImage() {
-        return mimeType != null && mimeType.contains("image");
+        return type.equals(ModelType.attach_image);
```
JcMinarro, line 43: To avoid NPE here, you can flip your check, something like
```suggestion
        return ModelType.attach_image.equals(type);
```

library/src/main/java/com/getstream/sdk/chat/utils/StorageUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+    internal fun getCachedFileFromUri(
+        context: Context,
+        attachmentMetaData: AttachmentMetaData
+    ): File {
+        if (attachmentMetaData.file != null) {
+            return attachmentMetaData.file!!
+        }
+        val cachedFile = File(context.cacheDir, "STREAM_${attachmentMetaData.title}")
+        if (!cachedFile.exists()) {
+            context.contentResolver.openInputStream(attachmentMetaData.uri!!)?.use { inputStream ->
```
JcMinarro, line 41: `attachmentMetaData.uri` seems that could be null, we should avoid NPE

library/src/main/java/com/getstream/sdk/chat/utils/StorageUtils.kt
```diff
@@ -0,0 +1,156 @@
...
+        )
+
+    internal fun getCachedFileFromUri(
+        context: Context,
+        attachmentMetaData: AttachmentMetaData
+    ): File {
+        if (attachmentMetaData.file != null) {
+            return attachmentMetaData.file!!
+        }
+        val cachedFile = File(context.cacheDir, "STREAM_${attachmentMetaData.title}")
```
JcMinarro, line 39: Is it safe? What about two files with the same title but in different folders?

library/src/main/java/com/getstream/sdk/chat/adapter/MediaAttachmentSelectedAdapter.java
adasiewiczr: My bad, I've used `attachment.file` only for captured images/videos, but I think it can be simplified

library/src/main/java/com/getstream/sdk/chat/model/AttachmentMetaData.java
adasiewiczr: I think this method can be removed

library/src/main/java/com/getstream/sdk/chat/utils/MessageInputController.kt
```diff
@@ -200,11 +199,14 @@ class MessageInputController(
...
+    private suspend fun getAttachmentsFromLocal(
+        isMedia: Boolean,
+        treeUri: Uri? = null
+    ): List<AttachmentMetaData> =
         withContext(Dispatchers.IO) {
             when (isMedia) {
-                true -> Utils.getMediaAttachments(view.context)
-                false -> Utils.getFileAttachments(Environment.getExternalStorageDirectory())
+                true -> StorageUtils.getMediaAttachments(view.context)
+                false -> StorageUtils.getFileAttachments(view.context, treeUri!!)
```
JcMinarro, line 209: We shouldn't use `!!` operator.
If we want that something won't be null, we should indicate it on the signature of the method.

JcMinarro (approved):

PR #613, opened 2020-09-05
Add JvmField annotation to default filter and sort type inside ChannelListViewModel

adasiewiczr:
Adding @JvmField annotation improves UX when using SDK in Java because user can get those values directly from `ChannelListViewModel` instead of `ChannelListViewModel.Companion`


zsmb13 (approved):
Looks good 👍

zsmb13 (changes requested):
Sorry, on second look, these should be marked with `@JvmField` instead.

We currently get `ChannelsViewModel.getDEFAULT_FILTER()` as the Java syntax, but the desired one is `ChannelsViewModel.DEFAULT_FILTER` (the same as from Kotlin), which is what `@JvmField` will do.

zsmb13 (approved):

JcMinarro (approved):

PR #614, opened 2020-09-07
[CAS-232] Refactor MessageViewHolderFactory

zsmb13:
Migration of the `MessageViewHolderFactory` class to Kotlin.


library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    open fun createMessageViewHolder(adapter: MessageListItemAdapter, parent: ViewGroup, viewType: Int): BaseMessageListItemViewHolder<*>? {
+        return when (viewType) {
+            MESSAGEITEM_DATE_SEPARATOR ->
+                DateSeparatorViewHolder(R.layout.stream_item_date_separator, parent)
+            MESSAGEITEM_MESSAGE ->
+                MessageListItemViewHolder(R.layout.stream_item_message, parent).apply {
+                    setMessageClickListener(adapter.messageClickListener)
```
JcMinarro, line 67: What do you think about to decouple our `ViewHolder` from our `Adapter`?
As I see, all logic on this case is coupled.
`Factory` doesn't need to know about the `Adapter`. `Adapter` has some "listeners" that are set only to be send to the `factory` to instanciate the different `ViewHolders`.

At the end, the only one that need the listener is the `factory` class to be able to generate our `ViewHolder`.
It can be added to the `factory` constructor (If we want them mandatory) and try to remove them from `Adapter` class

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    open fun createAttachmentViewHolder(
+        adapter: AttachmentListItemAdapter,
+        parent: ViewGroup,
+        viewType: Int
+    ): BaseAttachmentViewHolder {
+        return when (viewType) {
+            VIDEO_ATTACHMENT, IMAGE_ATTACHMENT ->
+                AttachmentViewHolderMedia(R.layout.stream_item_attach_media, parent).apply {
+                    setGiphySendListener(adapter.giphySendListener)
```
JcMinarro, line 91: Same here

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    open fun getAttachmentViewType(
+        message: Message,
+        mine: Boolean,
+        position: Position,
+        attachments: List<Attachment>,
+        attachment: Attachment
+    ): Int {
```
JcMinarro, line 46: It seems that we don't need all those arguments

JcMinarro (commented):
I have left some comments on it.
Apart of that it looks good ;)
What do you think about adding some tests to be sure we don't break our Factory behavior in the future?

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
zsmb13: As far as I can tell, the purpose of this class is to allow for customization (by overriding these methods as needed in a subclass), so there could be implementations of it by users where these arguments are being used for something. Not sure which ones - if any - would be safe to remove.

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
zsmb13: We need more than half a dozen listeners here in the factory, to be able to create `MessageListItemViewHolder` instances. Passing all of those in separately seems impractical. If we want to decouple from the adapter, we could hide it behind an interface, something like a `ListenerProvider` (only with a better name).

One thing to note is that custom implementations of this factory, as it is now, could be using the adapter reference for more than just the click listeners it has. This, especially with the changes in https://github.com/GetStream/stream-chat-android/pull/617 includes the current channel, the bubble helper, the style, whether we're in a thread... Do we want to allow access to these things in the factory?

JcMinarro: I like the idea to have something like a `ListenerProvider` if we have a lot of listeners to be configured. It can be an interface that belongs to our Factory. If some listener is not mandatory, it could have a default implementation that does nothing, it will help dev to use it.

About the use of the adapter into our factory, in the future, it should be removed, in the end, the factory is who provides stuff to the adapter and not in the other direction. If it is hard to remove it now, we can iterate over it in another task.

Lastly, about the BubbleHelper, Style... I don't think it is needed into our factory, but if we would need it, it should be possible to add to the factory as another dependency on the constructor, keeping methods simple as possible

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
JcMinarro: It is true. But all those info could be only into the "item" that will be rendered, and we can be sure that all info needed for this type will be into its "itemClass". I mean, we could have different types of items to be rendered, and every of them will have only the needed data to be shown on the ViewHolder.
BTW. I can see this factory method is not used into `MessageListItemAdapter`, it is into `AttachmentListItemAdapter`, Maybe we can create another task to cover it instead of add lot of changes on this PR

zsmb13: How about only keeping the `attachment` parameter then? Feels like that should always be enough to figure out the type, and then everything else can be taken into account later, when binding date into the ViewHolder.

JcMinarro: Sound good, but maybe we can wrap it into some class like `AttachmentItem` that will live into our `AttachmentListItemAdater`, it would allow us to add extra info to be rendered on the `ViewHolder` without break our public API

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
ogkuzmin: My thoughts about this. Here we want to create an instance of ViewHolder, but listeners are required to setup it. We could decouple this two operations and provide functional object that know how to setup particular type of ViewHolder. It could have a such interface:

```kotlin
interface ViewHolderSetupHandler /* whatever name :-) */ {
  fun setup(viewHolder: BaseMessageListItemViewHolder)
}
```

And we can provide it via constructor. It is more flexible than providing listeners

ogkuzmin (approved):

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
zsmb13: For now, I've removed the extra parameters, and kept just the `Attachment`. Will follow with a wrapper class around `Attachment` in a separate PR.

library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListItemAdapter.java
```diff
@@ -42,7 +42,7 @@ public AttachmentListItemAdapter(Context context, @NonNull MessageListItem.Messa
     public int getItemViewType(int position) {
         try {
             Attachment attachment = attachments.get(position);
-            return factory.getAttachmentViewType(message, true, MessageViewHolderFactory.Position.BOTTOM, attachments, attachment);
+            return factory.getAttachmentViewType(attachment);
```
ogkuzmin, line 45: Much clearer 👍

ogkuzmin (approved):

JcMinarro (approved):

PR #617, opened 2020-09-07
[CAS-234] Refactor MessageListItemAdapter

zsmb13:
This PR migrates `MessageListItemAdapter` from Java to Kotlin.

- Getter/setter pairs have been converted into Kotlin properties. The channel, the bubble helper, and the factory are now exposed from the adapter - previously, they only had setters, but we can't really have write-only properties in Kotlin. **If we don't want these exposed, we could throw exceptions from their getters and mark them deprecated.**
- `BaseMessageListItemViewHolder` was converted to Kotlin, and now has a new `bindListItem` method, as it needs to be able to accept any `MessageListItem` instead of just `T` specifically (the latter can't be passed in from Kotlin in the adapter, as its type isn't proven, and raw types are not allowed).
- `MessageListItemDiffCallback` has been converted to Kotlin, since it's a very simple class used only by the adapter.


library/src/main/java/com/getstream/sdk/chat/adapter/BaseMessageListItemViewHolder.kt
```diff
@@ -0,0 +1,43 @@
...
+    resId: Int,
+    parent: ViewGroup
+) : RecyclerView.ViewHolder(
+    LayoutInflater.from(parent.context).inflate(resId, parent, false)
+) {
+
+    /**
+     * Workaround to allow a downcast of the MessageListItem to T
+     */
+    fun bindListItem(
```
JcMinarro, line 21: This method seems to have lot of arguments and I think not all of them are needed to bind an item to its view.
`context` is not needed because we have an already inflated view where we can get the context.
A `ViewHolder` shouldn't need the factory that already creates itself
The rest of arguments, if they are needed to the render process, should be into our `MessageListItem`

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
```diff
@@ -0,0 +1,81 @@
...
+        parent: ViewGroup,
+        viewType: Int
+    ): BaseMessageListItemViewHolder<*> {
+        return viewHolderFactory.createMessageViewHolder(this, parent, viewType)
+    }
+
+    override fun onBindViewHolder(holder: BaseMessageListItemViewHolder<*>, position: Int) {
+        holder.bindListItem(
+            context,
+            requireNotNull(channel) { "Channel was not set" },
```
JcMinarro, line 73: If we require them, we shouldn't allow create an adapter with null values

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
```diff
@@ -0,0 +1,81 @@
...
+            messageListItemList[position],
+            requireNotNull(style) { "Style was not set" },
+            requireNotNull(bubbleHelper) { "BubbleHelper was not set" },
```
JcMinarro, line 76: Same for those two

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
```diff
@@ -0,0 +1,81 @@
...
+        )
+        result.dispatchUpdatesTo(this)
+        messageListItemList = newEntities
+    }
+
+    override fun getItemCount(): Int = messageListItemList.size
+
+    override fun getItemId(position: Int): Long = messageListItemList[position].getStableId()
+
+    override fun getItemViewType(position: Int): Int {
```
JcMinarro, line 54: Which case we are asked for a position out of our collection? :thinking:

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
```diff
@@ -0,0 +1,81 @@
...
+) : RecyclerView.Adapter<BaseMessageListItemViewHolder<*>>() {
+
+    var bubbleHelper: BubbleHelper? = null
+
+    var messageClickListener: MessageClickListener? = null
+        set(value) {
+            if (style?.isReactionEnabled == true)
+                field = value
+        }
+    var messageLongClickListener: MessageLongClickListener? = null
```
JcMinarro, line 32: I were talking about those listeners in other PR, I think they shouldn't be into our Adapter, it doesn't need them.
On the other hand, would be great if we remove `null` value here, At the end those listener should be easy to convert into an empty function instead of a null value.
If they are mandatory and couldn't be added to our constructor because some reason, we can implement them as a function that throw an exception if no implementation is assigned.
At the end, no null values does our code is more readable

library/src/main/java/com/getstream/sdk/chat/adapter/BaseMessageListItemViewHolder.kt
zsmb13: This is covered by a separate task, CAS-231 🙂

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
zsmb13: Will review in CAS-237

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
zsmb13: Will review in CAS-237

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
zsmb13: Will refactor in CAS-237, where these will be moved from the adapter to the factory altogether

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
zsmb13: Removed the check, as it should never fail 👍

JcMinarro (approved):

Issue #588, opened 2020-08-06
gif messages not rendering 


pixnbit:
Android (4.11-beta7) vs iOS
![Screenshot_1596739448](https://user-images.githubusercontent.com/729730/89569899-37f70780-d7da-11ea-8932-4fa065e76dc5.png)

![Simulator Screen Shot - iPhone 11 Pro Max - 2020-08-06 at 11 44 10](https://user-images.githubusercontent.com/729730/89569878-32012680-d7da-11ea-8d2b-487e99b1493d.png)

JcMinarro:
Hello @pixnbit 
It was fixed on #619 and added to the last released.
Could you check if it is working fine for you now?

pixnbit:
Yes this has been resolved with upgrading to the latest SDK

PR #619, opened 2020-09-08
[cas-159] Fix giphy attachment rendering

samiuelson:
Fix giphy attachment rendering:
* Update .gif rendering code to latest Glide API


adasiewiczr (approved):

JcMinarro:
Do you think it will fix #588 as well?
Do we have any way to identify that an attachment is a gif file?
What happen if we share a gif from our gallery instead of from "Giphy Command"?

samiuelson:
> Do you think it will fix #588 as well?
> Do we have any way to identify that an attachment is a gif file?
> What happen if we share a gif from our gallery instead of from "Giphy Command"?

Probably we will need a file MIME type parser. It will rather not fix rendering .gif files send as file attachment (#588) without this. @JcMinarro

JcMinarro (approved):

PR #620, opened 2020-09-09
[CAS 216] Fix close thread issue

ogkuzmin:
This fix ensures that thread messages are shown if only we're in state of thread showing.


ogkuzmin:
Working with class MessageListItemLiveData I thought that design of this class isn't clear (especially when I was writing tests). The live data android logic is coupled with business logic of merging messages (thread messages/not thread messages), handling observers and thread state, posting updates. It could be decoupled and bring more clearness to this class. We could put refactoring of this live data class to backlog and do later

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
```diff
@@ -0,0 +1,122 @@
...
+        private set
+
+    override fun onChanged(value: T?) {
+        lastObservedValue = value
+    }
+
+    fun cleanState() {
+        lastObservedValue = null
+    }
+}
```
zsmb13, line 122: This looks like something we should be reusing in tests, so it could probably live in its own file. My own approach to an Observer like this is [here](https://github.com/rainbowcake/rainbowcake/blob/d1911c0e75d74b2f4f1fc6c9054dca0a37ab77b4/rainbow-cake-test/src/main/java/co/zsmb/rainbowcake/test/MockObserver.kt), and it stores a list of values instead of just the last one, and then things can be asserted on it neatly with extensions.

The `cleanState` method is a good idea though, I don't have that myself. Maybe it could be called `reset`, like a Mockito mock can be reset?

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
```diff
@@ -0,0 +1,122 @@
...
+
+    private lateinit var sut: MessageListItemLiveData
+
+    @BeforeEach
+    fun setUp() {
+        observer = TestObserver()
+        messages = MutableLiveData(listOf())
+        threadMessages = MutableLiveData(listOf())
+        typing = MutableLiveData(listOf())
+        reads = MutableLiveData(listOf())
```
zsmb13, line 37: Nit: could use [`emptyList`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/empty-list.html) for these

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Test
+    fun `When messages are changed if live data is not in the thread state should observe last assigned value to messages`() {
+        sut.observeForever(observer)
+        val sendingMessage = createMessage()
+
+        messages.value = listOf(sendingMessage)
+
+        observer.lastObservedValue.also { result ->
```
zsmb13, line 60: In a context like this, when you want to introduce a name for a value, `let` is usually the function used instead of `also`, as a convention in Kotlin.

In this specific case, just going with `val result = observer.lastObservedValue` seems to be good enough too, and it would simplify the code.

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
```diff
@@ -0,0 +1,122 @@
...
+import io.getstream.chat.android.client.models.Message
+import io.getstream.chat.android.client.models.User
+import org.amshove.kluent.shouldBeEqualTo
+import org.amshove.kluent.shouldNotBeEqualTo
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+
+@ExtendWith(InstantExecutorExtension::class)
+class MessageListItemLiveDataTests {
```
zsmb13, line 19: Happy to see these tests, nicely done 👍

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
```diff
@@ -0,0 +1,122 @@
...
+    @Test
+    fun `When messages are changed if live data is not in the thread state should observe last assigned value to messages`() {
+        sut.observeForever(observer)
+        val sendingMessage = createMessage()
+
+        messages.value = listOf(sendingMessage)
+
+        observer.lastObservedValue.also { result ->
+            result shouldNotBeEqualTo null
+            result!!.listEntities.any { item ->
```
zsmb13, line 62: Kluent has a `shouldNotBeNull` method we could use on the first line here, and then we wouldn't need to use `!!` later on, because it has [contract support](https://github.com/MarkusAmshove/Kluent/blob/e94cbd8dd1a4a7a5cc3271f937840bbf1b380974/common/src/main/kotlin/org/amshove/kluent/Basic.kt#L26-L33).

zsmb13 (approved):
Left some comments on the tests with improvement ideas, but the changes look good to me in general 👍

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
JcMinarro: As I see `also` is to apply some side-effect to our computation.
On that case, it is not a side-effect, it is part of our check-task.
I will go with 
```
val result = observer.lastObservedValue
result.shouldNotBeNull()
....
```

JcMinarro (approved):
Good Job.
Some comments on the tests side.
BTW, I have been checking `MessageListItemLiveData` class and it is a good candidate to be refactored in the future, the logic it does is not clear. Good thing, now we have tests over it, so it will be easier to update

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
ogkuzmin: Thanks for the feedback! I renamed that method according to Mockito style. Also moved this observer class to a standalone file. In future If we need list of all observed value like in your version we can easily change it

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
ogkuzmin: Done

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
ogkuzmin: Thanks guys, fixed it

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveDataTests.kt
ogkuzmin: Thanks for the comment, I hasn't worked with fluent before

zsmb13 (approved):

PR #621, opened 2020-09-10
Introduce AttachmentListItem wrapper class

zsmb13:
Wraps the `Attachment` network model into the `AttachmentListItem` class, so that it's not used directly in `AttachmentListItemAdapter`. This allows us to add extra properties to the list item later, if needed.

See previous discussion in https://github.com/GetStream/stream-chat-android/pull/614#discussion_r484452890

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListItemAdapter.java
+++ b/library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListItemAdapter.java
@@ -11,8 +11,8 @@ import com.getstream.sdk.chat.view.MessageListViewStyle;
 
 import java.util.List;
 
-import io.getstream.chat.android.client.models.Attachment;
 import io.getstream.chat.android.client.models.Message;
+import kotlin.collections.CollectionsKt;
 
 
 public class AttachmentListItemAdapter extends RecyclerView.Adapter<RecyclerView.ViewHolder> {
@@ -22,7 +22,7 @@ public class AttachmentListItemAdapter extends RecyclerView.Adapter<RecyclerView
     private MessageViewHolderFactory factory;
     private Context context;
     private MessageListItem.MessageItem messageListItem;
-    private List<Attachment> attachments;
+    private List<AttachmentListItem> attachments;
     private MessageListViewStyle style;
     private MessageListView.AttachmentClickListener attachmentClickListener;
     private MessageListView.MessageLongClickListener longClickListener;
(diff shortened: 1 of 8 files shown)


library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentViewHolder.java
```diff
@@ -69,7 +69,7 @@ public void bind(@NonNull Context context,
         this.context = context;
         this.messageListItem = messageListItem;
         this.message = message;
-        this.attachment = attachment;
+        this.attachment = attachmentItem.getAttachment();
```
zsmb13, line 72: Decided to not change the type of the private fields inside the ViewHolder classes, for now. This can be changed later so that the entire item is stored if needed, but that's all private implementation details, and it doesn't break the public API.

library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListItemAdapter.java
```diff
@@ -22,27 +22,29 @@
...
-    public AttachmentListItemAdapter(Context context, @NonNull MessageListItem.MessageItem messageListItem, @NonNull MessageViewHolderFactory factory) {
+    public AttachmentListItemAdapter(Context context,
+                                     @NonNull MessageListItem.MessageItem messageListItem,
+                                     @NonNull MessageViewHolderFactory factory) {
         this.context = context;
         this.messageListItem = messageListItem;
         this.message = messageListItem.getMessage();
         this.factory = factory;
-        this.attachments = message.getAttachments();
+        this.attachments = CollectionsKt.map(message.getAttachments(), AttachmentListItem::new);
```
zsmb13, line 40: Example of how handy Kotlin collection functions are, even when writing Java code

ogkuzmin (approved):

adasiewiczr (approved):

library/src/main/java/com/getstream/sdk/chat/adapter/AttachmentListItemAdapter.java
JcMinarro: I have never used it from Java, Taking note ;)

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
```diff
@@ -37,9 +36,9 @@ open class MessageViewHolderFactory {
     }
 
     open fun getAttachmentViewType(
-        attachment: Attachment
+        attachmentItem: AttachmentListItem
     ): Int {
-        return when (attachment.type) {
+        return when (attachmentItem.attachment.type) {
```
JcMinarro, line 41: It seems that `AttachmentListItem` could be a `sealed class`, so it will be easier to handle multiple types.
It could be covered on another PR when `ViewHolder` and `Adapter` will refactored

JcMinarro (dismissed):

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
zsmb13: I'd leave this for later, the wrapper feels like enough preparation for future expansion for now

JcMinarro:
BTW, @zsmb13 could you update our `CHANGELOG.md` with the new updates?

JcMinarro (approved):

PR #622, opened 2020-09-10
Fix/channel read state

adasiewiczr:
Channels list wasn't updated properly when calling `setUser()` without disconnecting the previous one because of socket disconnection problem. We did it in our sample app every time we return from `ChannelFragment` to `ChannelsFragment`.
We also need to update the LLC dependency because the current version doesn't include (https://github.com/GetStream/stream-chat-android-client/pull/66/files).

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
+++ b/library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
@@ -133,12 +133,6 @@ class ChatImpl implements Chat {
                         @NotNull String userToken,
                         @NotNull InitConnectionListener callbacks) {
         final ChatClient client = ChatClient.instance();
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -133,6 +133,12 @@ public void setUser(@NotNull User user,
                         @NotNull String userToken,
                         @NotNull InitConnectionListener callbacks) {
         final ChatClient client = ChatClient.instance();
+        if (client.getCurrentUser() != null) {
```
JcMinarro, line 136: Shouldn't we move this check to LLC? :thinking:

samiuelson (approved):

samiuelson (approved):

PR #624, opened 2020-09-11
WIP: draft implementations for ListenerContainer

zsmb13:
This PR refactors only the `messageClickListener` used by `MessageListItemViewHolder`, as an example of what the new listener setup could look like. Instead of the ViewHolder receiving the click listener through a setter after it's created, it now accesses it through the `ListenerContainer` class.

There are two implementations of `ListenerContainer` in the same file, one of them is a more manual implementation, but perhaps easier to understand. The second implementation abstracts away the common pattern that would have to be repeated for each listener into a property delegate.

Listeners set into the Container can not be accessed directly, they can only be called through a wrapper listener exposed from the Container. Holding a reference to this wrapper is safe, as when the actual listener underneath is swapped out,  calls will start going to the new listener.


zsmb13:
Closing WIP PR, will open separate one for actual changes

PR #625, opened 2020-09-11
Fix/cas 188 show new channels

ogkuzmin:
Since we made 
``` kotlin  
Chat.getInstance().setUser(...)
```
in onViewCreated of Fragment it led to overriding singleton instance of ChatDomain. That's why we lost updating of channels by new ones and last messages in ChannelList screen.
I moved that code to the Login screen logic.


adasiewiczr:
🙈 @ogkuzmin it's already implemented in https://github.com/GetStream/stream-chat-android/pull/622

ogkuzmin:
@adasiewiczr hah))) Not bad. I'll close my PR :)

PR #626, opened 2020-09-11
[CAS-237] Refactor ViewHolder listeners, introduce ListenerContainer

zsmb13:
This PR refactors the way that ViewHolders in `MessageListView` receive their listeners.

Instead of `MessageListView` and `MessageListItemAdapter` both holding onto listener instances individually, a `ListenerContainer` class now owns and controls the listeners. It also wraps them into references that are safe for the ViewHolders to hold onto, even if new listeners are set on the `MessageListView`.

The `MessageViewHolderFactory` that creates the ViewHolders no longer depends on `MessageListItemAdapter`. It receives a reference to the `ListenerContainer` as a constructor parameter instead.

As documented, setting `null` values for listeners will revert the listeners to their default implementations, now clearly stored in constants in `MessageListView`.

---

For additional context, take a look at CAS-237 and its comments.


library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
```diff
@@ -4,38 +4,18 @@ import android.content.Context
...
-import com.getstream.sdk.chat.view.MessageListView.MessageClickListener
-import com.getstream.sdk.chat.view.MessageListView.MessageLongClickListener
-import com.getstream.sdk.chat.view.MessageListView.ReactionViewClickListener
-import com.getstream.sdk.chat.view.MessageListView.ReadStateClickListener
 import com.getstream.sdk.chat.view.MessageListViewStyle
 import io.getstream.chat.android.client.models.Channel
 
 class MessageListItemAdapter @JvmOverloads constructor(
     private var context: Context,
+    var viewHolderFactory: MessageViewHolderFactory,
```
JcMinarro, line 13: Do we want it mutable?

zsmb13: One of our customization examples in the tutorial is subclassing the factory and replacing it, so I think we want to support swapping it out.

library/src/main/java/com/getstream/sdk/chat/adapter/ListenerContainer.kt
```diff
@@ -0,0 +1,109 @@
...
+import com.getstream.sdk.chat.view.MessageListView.GiphySendListener
+import com.getstream.sdk.chat.view.MessageListView.MessageClickListener
+import com.getstream.sdk.chat.view.MessageListView.MessageLongClickListener
+import com.getstream.sdk.chat.view.MessageListView.ReactionViewClickListener
+import com.getstream.sdk.chat.view.MessageListView.ReadStateClickListener
+import com.getstream.sdk.chat.view.MessageListView.UserClickListener
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
+
+class ListenerContainer(
```
adasiewiczr, line 13: I am just thinking if we shouldn't make this class `open` to allow customization and annotate its constructor with @JvmOverloads. WDYT?

zsmb13: Until we have a use case for this, I'd like to avoid customizing this class. It would be hard to manage the lifecycle of custom instances of it (since it's really important that we only have a single one, owned by `MessageListView`). In fact, now that I think about it, I'd make this `internal`, and only expose it to the factory through a public interface, because clients overriding the factory might need to reference the listeners.

---

As for `@JvmOverloads`, since there's no clear order or priority between the listeners here, I don't think it would help a whole lot. I don't see it likely that you'd only provide the first couple of them specifically.

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
zsmb13: Made this and other ctor properties of the adapter read-only, so now customization can only happen before the adapter instance is created.

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
```diff
@@ -25,6 +25,8 @@ open class MessageViewHolderFactory {
         const val FILE_ATTACHMENT = 4
     }
 
+    lateinit var listenerContainer: ListenerContainer
```
zsmb13, line 28: This is injected manually by `MessageListView`. It's public, as anyone extending `MessageViewHolderFactory` might need to access the listeners for their own custom ViewHolders.

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -129,24 +192,15 @@ private void parseAttr(Context context, @Nullable AttributeSet attrs) {
...
-        setMessageClickListener(messageClickListener);
-        setMessageLongClickListener(messageLongClickListener);
-        setAttachmentClickListener(attachmentClickListener);
-        setReactionViewClickListener(reactionViewClickListener);
-        setUserClickListener(userClickListener);
-        setReadStateClickListener(readStateClickListener);
-        setMessageLongClickListener(messageLongClickListener);
+    @Deprecated
+    public final void setAdapter(Adapter adapter) {
+        throw new IllegalArgumentException("Use setMessageListItemAdapter instead please");
```
zsmb13, line 197: The method being referenced here became `private` a while ago. Do we forbid setting a custom adapter on purpose, so that users only set custom things used by our adapter, such as the ViewHolder factory, the bubble helper, etc?

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -116,7 +180,6 @@ private void init(Context context) {
 
         this.setLayoutManager(layoutManager);
         hasScrolledUp = false;
-        setBubbleHelper(DefaultBubbleHelper.initDefaultBubbleHelper(style, context));
```
zsmb13, line 119: Moved into `initAdapter`

library/src/main/java/com/getstream/sdk/chat/adapter/MessageViewHolderFactory.kt
JcMinarro: Could we have an "empty implementation" of this field instead of a `lateinit var`?

zsmb13: It's guaranteed to be initialized by our own library code, we just can't do it at the time when the constructor is called. If it's somehow used before it's set, it _should_ crash, because that's an error in the init calls.

JcMinarro (approved):

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
zsmb13: @samiuelson, any idea why we changed this to `private`? It was your change a couple months ago. https://github.com/GetStream/stream-chat-android/commit/f7dfc64844e6350f2b17587a6a600c3563e8db7b#diff-778483d318fbb6393eb8add88b4fb9e1R127

samiuelson: it just wasn't used outside the class, but probably it's a mistake on my end. feel free to make it public

samiuelson: I think, previously the concept of working with `MessageListView` was to allow calling `setAdapterWithStyle` to set the adapter manually; or to call `init(Channel channel, User currentUser)` which calls `initAdapter()` internally. 

So, `setAdapterWithStyle` and `init(Channel channel, User currentUser)` were supposed to be public.

samiuelson: Also, feel free to refactor this. Function names are a bit confusing.

zsmb13: From out point of view, it's better if we can leave it `private`, as we then have control over what the adapter being used is... If we don't hear any complaints about this, I'd leave it as `private. It can be customized in a bunch of ways without swapping out our adapter implementation.

JcMinarro (approved):

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
```diff
@@ -4,40 +4,20 @@ import android.content.Context
...
-import com.getstream.sdk.chat.view.MessageListView.ReadStateClickListener
 import com.getstream.sdk.chat.view.MessageListViewStyle
 import io.getstream.chat.android.client.models.Channel
 
 class MessageListItemAdapter @JvmOverloads constructor(
     private var context: Context,
-    var channel: Channel? = null,
-    private var messageListItemList: List<MessageListItem> = emptyList(),
-    var viewHolderFactory: MessageViewHolderFactory = MessageViewHolderFactory()
+    val channel: Channel,
```
ogkuzmin, line 13: I like it! Very nice to provide required dependencies via constant values in the constructor

ogkuzmin (approved):

JcMinarro (approved):

PR #627, opened 2020-09-11
Remove dead code

JcMinarro:
Remove dead code


adasiewiczr (approved):
LGTM

library/src/main/java/com/getstream/sdk/chat/enums/QueryOptions.java
```diff
@@ -1,4 +0,0 @@
-package com.getstream.sdk.chat.enums;
-
-public class QueryOptions {
-}
```
zsmb13, line 4: 🤔

zsmb13 (approved):
The most satisfying kind of PR 😄  👍

library/src/main/java/com/getstream/sdk/chat/enums/QueryOptions.java
JcMinarro: :rofl: I had the same thought on it

zsmb13 (approved):

PR #630, opened 2020-09-14
[CAS-225] Fix message date is not shown

JcMinarro:
The date/time where the message was sent wasn't shown properly on the `MessageListView` because it wasn't used proper fields into `message` entity.
It has been updated and some old code that is not used anymore has been removed


adasiewiczr (approved):

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemViewHolder.java
```diff
@@ -263,12 +263,7 @@ protected void configUserNameAndMessageDateStyle() {
 
         if (style.isMessageDateShow()) {
             tv_messagedate.setVisibility(View.VISIBLE);
-            if (message.getDate() == null)
-                LlcMigrationUtils.setStartDay(Arrays.asList(message), null);
-            if (message.getDate().equals(TODAY.getLabel()) || message.getDate().equals(YESTERDAY.getLabel()))
-                tv_messagedate.setText(message.getTime());
-            else
-                tv_messagedate.setText(context.getString(R.string.stream_message_date, message.getDate(), message.getTime()));
```
zsmb13, line 271: So we're throwing out this behaviour, and simply always showing `HH:mm` time instead?

---

The `stream_message_date` resource is now unused 🙂

zsmb13 (approved):
Looks okay to me

PR #632, opened 2020-09-15
Fix reactions

tschellenbach:
- upgrade gradle
- use latest livedata lib that fixes support for offline reactions


tschellenbach:
closing since its hard to read with the base changed.

PR #642, opened 2020-09-22
Fix MessageListItemDiffCallback

JcMinarro:
Fix MessageListItemDiffCallback


zsmb13 (commented):
I can't comment on that line, but are we content (no pun intended) with how the `areContentsTheSame` method works? Is that a valid way to compare the instances, by their `equals` method? I know there were some concerns around this.

JcMinarro:
> I can't comment on that line, but are we content (no pun intended) with how the `areContentsTheSame` method works? Is that a valid way to compare the instances, by their `equals` method? I know there were some concerns around this.

For now we can continue with this implementation, it is not the best and we will need to find other solution.
The wrong case with this solution is that a message is rendered on every change, despite of the change is important for the UI or not

tschellenbach:
I don't think there are any message changes that change the message but don't result in a UI update. Can't think of any at least.

JcMinarro:
> I don't think there are any message changes that change the message but don't result in a UI update. Can't think of any at least.

Some example could be `MessageListItem.MessageItem.Message.User::online`
This flag notify us if a user is online or not, but it is not shown on the MessageListView. 
If it is updated, the message will be rendered again, without any change for the user that is reading the chat

zsmb13 (approved):
This is an obvious improvement for now, and we can get back to the content comparison at a later time.

PR #647, opened 2020-09-25
Use ViewBinding in ViewHolders

zsmb13:
Replaces the previous implementation of inflating layouts in the base ViewHolder with the base class accepting an already inflated layout, and each individual ViewHolder inflating their own, using ViewBinding.


library/src/main/java/com/getstream/sdk/chat/utils/frescoimageviewer/MultiTouchViewPager.java
```diff
@@ -7,7 +7,7 @@
 import androidx.annotation.NonNull;
 import androidx.viewpager.widget.ViewPager;
 
-class MultiTouchViewPager extends ViewPager {
+public class MultiTouchViewPager extends ViewPager {
```
zsmb13, line 10: This was needed for ViewBinding to be able to generate bindings that contain this class.

adasiewiczr (approved):

samiuelson (approved):

PR #649, opened 2020-09-25
Convert navigation classes to Kotlin

zsmb13:


library/src/main/java/com/getstream/sdk/chat/navigation/ChatNavigatorImpl.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class ChatNavigatorImpl : ChatNavigator {
+    private var handler: ChatNavigationHandler? = null
+
+    override fun setHandler(handler: ChatNavigationHandler) {
+        this.handler = handler
+    }
+
+    override fun navigate(destination: ChatDestination) {
+        val handler = handler
```
JcMinarro, line 13: Why do we need this local variable?

library/src/main/java/com/getstream/sdk/chat/navigation/ChatNavigatorImpl.kt
```diff
@@ -0,0 +1,22 @@
...
+        if (handler == null || !handler.navigate(destination)) {
```
JcMinarro, line 14: Could we do it in a more "Kotlin Style"?

library/src/main/java/com/getstream/sdk/chat/navigation/destinations/ChatDestination.kt
```diff
@@ -0,0 +1,17 @@
...
+
+abstract class ChatDestination(protected var context: Context) {
+    abstract fun navigate()
+
+    protected fun start(intent: Intent) {
+        context.startActivity(intent)
+    }
+
+    protected fun startForResult(intent: Intent, requestCode: Int) {
+        (context as Activity).startActivityForResult(intent, requestCode)
```
JcMinarro, line 15: This casting could fails, not all `Context` are `Activity`, we should handle it

library/src/main/java/com/getstream/sdk/chat/navigation/ChatNavigatorImpl.kt
zsmb13: `handler` is a mutable var, so smart casts don't work on it otherwise. The simplest the checks here get is what you see on the next line. The temporary variable could be introduced with a `let {}`, but it's really not any more readable.

library/src/main/java/com/getstream/sdk/chat/navigation/destinations/ChatDestination.kt
zsmb13: True, added a more explicit check and a nicely worded exception message for this

library/src/main/java/com/getstream/sdk/chat/navigation/ChatNavigatorImpl.kt
JcMinarro: What about having a "Non ChatNavitationHandler" that return false and avoid null values?

```
private var handler: ChatNavigationHandler = object  : ChatNavigationHandler {
    override fun navigate(destination: ChatDestination?): Boolean = false
}
```

zsmb13: Ah, I thought I allowed setting `null` handlers through the setter function, but apparently not. If we don't need to be able to `null` out the `handler` later on, this is a great idea, I'll add it.

JcMinarro (approved):

PR #650, opened 2020-09-25
[CAS-271] Show button to scroll bottom

leandroBorgesFerreira:
# Submit a pull request

## CLA

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [x] The code changes follow best practices
- [x] Code changes are tested (add some information if not applicable)

In this PR I add a small button so the user can go to the bottom of the screen after scrolling up. Also if new messages arrive at the chat the user can see how much new messages are available.

I still didn't put the correct icon for the button and I'm not sure if the position of the button is correct, but the logic seems to be OK. 

About unit tests, the logic is inside a view, which makes this code a bit harder to tests. Normally I would put the logic in a separated class, but I don't know how this would conflict with the ViewModel for this class. I'm open for ideas about where to move the code the unit tests are possible to be written. 

Here is a little demo:

Sep-29-2020 23-17-43

And that's my PR.

giphy
_Buttons are fun_


library/src/test/java/com/getstream/sdk/chat/view/DefaultScrollButtonBehaviourTest.kt
```diff
@@ -0,0 +1,95 @@
...
+import com.nhaarman.mockitokotlin2.whenever
+import org.junit.Test
+import org.mockito.Mockito
+
+private const val NEW_MESSAGE_SINGULAR: String = "%s New Message"
+private const val NEW_MESSAGES_PLURAL: String = "%s New Messages"
+
+internal class DefaultScrollButtonBehaviourTest {
+
+    private val unseenBottomBtn: ViewGroup = Mockito.mock(ViewGroup::class.java)
```
zsmb13, line 18: Can be done like this using `mockito-kotlin` (just needs `import com.nhaarman.mockitokotlin2.mock`)

```suggestion
    private val unseenBottomBtn: ViewGroup = mock()
```

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -557,4 +667,74 @@ public void setOnStartThreadListener(Function1<Message, Unit> onStartThreadListe
...
-}
\ No newline at end of file
+
+    public interface ScrollButtonBehaviour {
+
+        void userScrolledUp();
+
+        void userScrolledToTheBottom();
+
+        void unreadMessages(int count);
```
zsmb13, line 677: How about something with a verb for this method name? `onUnreadMessageCountChanged`, `updateUnreadMessageCount`, something like that?

library/src/main/res/layout/stream_message_list_view.xml
```diff
@@ -0,0 +1,65 @@
...
+        android:minWidth="50dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        android:layout_marginBottom="10dp"
+        android:orientation="horizontal"
+        android:background="@drawable/stream_shape_round"
+        android:visibility="gone"
+        tools:visibility="visible"
+        android:animateLayoutChanges="true">
```
zsmb13, line 31: From the order of these attributes here, I think this file is not formatted. Really nice use of `tools` attributes all around though 👍

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -269,6 +365,36 @@ private void initAdapter() {
         this.setMessageListItemAdapter(adapter);
     }
 
+    public void setScrollButtonBehaviour(ScrollButtonBehaviour scrollButtonBehaviour) {
+        this.scrollButtonBehaviour = scrollButtonBehaviour;
+    }
+
+    public void setNewMessagesBehaviour(int newMessagesBehaviour) {
```
zsmb13, line 372: Could we use an enum for this instead when we're dealing with it in code? I know we also need the int handling for the XML attributes, but it would be nicer to deal with ints only in that part.

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -269,6 +365,36 @@ private void initAdapter() {
...
+
+    public void setScrollButtonBackgroundResource(int backgroundRes) {
+        unseenBottomBtn.setBackgroundResource(backgroundRes);
+    }
+
+    public void setScrollButtonBackground(Drawable drawable) {
+        unseenBottomBtn.setBackground(drawable);
+    }
+
+    public void setScrollButtonIconResource(int backgroundRes) {
```
zsmb13, line 388: These parameters should have `@DrawableRes` annotations

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -212,10 +297,20 @@ public void onScrolled(RecyclerView recyclerView, int dx, int dy) {
                         endRegionReachedHandler.invoke();
                     }
 
-                    hasScrolledUp = currentLastVisible <= (adapter.getItemCount() - 3); //TODO: investigate magic number 3
```
zsmb13, line 215: Happy to see this gone, nicely done 😄

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -2,17 +2,25 @@
...
+import android.content.res.TypedArray;
 import android.graphics.drawable.Drawable;
 import android.util.AttributeSet;
+import android.view.LayoutInflater;
 import android.view.View;
+import android.view.ViewGroup;
+import android.widget.ImageView;
+import android.widget.TextView;
 
+import androidx.annotation.IntDef;
```
zsmb13, line 14: There are some unused imports in this file, such as this one

zsmb13 (changes requested):
Looks good as an overall solution, left a bunch of small questions / improvement suggestions

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
leandroBorgesFerreira: I thought about `enum`, but I'm not sure if we should use it or not in our libraries as it increases the dex much more than I simple constant: https://medium.com/better-programming/android-how-enums-can-impact-the-performance-f787ef5903dd.

Since this is a SDK that others will use, I think we should avoid enums to create the smallest library as possible and go easier on memory as possible. What do you think?

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
leandroBorgesFerreira: Thanks =]

sample/src/main/res/values/strings.xml
```diff
@@ -8,4 +8,6 @@
     <string name="backend_error_info">Ups, something went wrong</string>
     <string name="app_title">Stream Chat</string>
     <string name="log_out_title">Log out</string>
-</resources>
\ No newline at end of file
+    <string name="new_messages_single">%s New Message</string>
+    <string name="new_messages_plural">%s New Messages</string>
```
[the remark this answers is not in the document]
leandroBorgesFerreira: If we use plurals we won't be able to configure this using XML. We can use plurals to configure the View programatically, but then the XML configuration and code configuration would be different, which I don't consider to be ideal.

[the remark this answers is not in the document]
zsmb13: Hmm, so we can't get plural strings as the parameter for an XML attribute?

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
zsmb13: For a while now, the official recommendation on this was that enums are fine to use (e.g. [this bit from I/O 2018](https://youtu.be/IrMw7MEgADk?t=849)), especially in application level code. They're using it more and more in the framework as well, and we're much closer to writing application code here than a tight inner loop of the framework anyway.

Overall, the safety gained here from having a well-typed API seems to vastly outweigh the possible performance gains.

sample/src/main/res/values/strings.xml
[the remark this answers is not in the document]
leandroBorgesFerreira: No, only using code.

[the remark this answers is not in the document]
leandroBorgesFerreira: Well, it is possible to do it using databinding: https://stackoverflow.com/questions/17458965/getting-a-string-from-plurals-to-use-in-xml

But I prefer not to depend on it.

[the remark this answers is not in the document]
zsmb13: I found [this ugly way](https://stackoverflow.com/a/25648349) of parsing whether a value you got in the attribute is a plural, but it really doesn't seem like it would be worth it. Let's go with just single vs plural for now then 👍

zsmb13 (approved):

tschellenbach:
@leandroBorgesFerreira this PR has been open for a while, what's the status?

leandroBorgesFerreira:
It is done. I had to make a correction and lost the approve. I'll talk to Martón and request a second review

library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
```diff
@@ -666,6 +663,30 @@ public void setOnStartThreadListener(Function1<Message, Unit> onStartThreadListe
...
+        }
+
+        public static NewMessagesBehaviour parseValue(int value) {
+            if (value == SCROLL_TO_BOTTOM.getValue())
+                return SCROLL_TO_BOTTOM;
+            if (value == COUNT_UPDATE.getValue())
+                return COUNT_UPDATE;
+            throw new IllegalArgumentException(
+                    "Unknown behaviour type. It must be either SCROLL_TO_BOTTOM (int 0) or COUNT_UPDATE (int 1)"
+            );
```
zsmb13, line 690: Nicely done with the details 👍

zsmb13 (approved):

PR #651, opened 2020-09-28
[CAS-231] Remove position parameter from ViewHolder binding method

zsmb13:
Removes the last unnecessary parameter from the ViewHolder binding methods, leaving just the `MessageItem` itself.

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/adapter/BaseMessageListItemViewHolder.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/adapter/BaseMessageListItemViewHolder.kt
@@ -11,18 +11,10 @@ abstract class BaseMessageListItemViewHolder<T : MessageListItem>(
     /**
      * Workaround to allow a downcast of the MessageListItem to T
      */
-    fun bindListItem(
-        messageListItem: MessageListItem,
-        position: Int
-    ) {
-        @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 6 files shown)


library/src/main/java/com/getstream/sdk/chat/adapter/BaseMessageListItemViewHolder.kt
```diff
@@ -11,18 +11,10 @@ abstract class BaseMessageListItemViewHolder<T : MessageListItem>(
...
-        bind(messageListItem as T, position)
-    }
+    @Suppress("UNCHECKED_CAST")
+    fun bindListItem(messageListItem: MessageListItem) = bind(messageListItem as T)
 
-    protected abstract fun bind(
-        messageListItem: T,
-        position: Int
-    )
+    protected abstract fun bind(messageListItem: T)
```
JcMinarro, line 17: At the end it was possible!!!! :rocket: :partying_face:

JcMinarro (approved):

library/src/main/java/com/getstream/sdk/chat/adapter/BaseMessageListItemViewHolder.kt
zsmb13: It's very satisfying to finally have just the single parameter in this `bind` method, that's for sure 😄

PR #652, opened 2020-09-28
Break up Message ViewHolder into Configurators

zsmb13:
Breaks up the `configure...` methods of `MessageListItemViewHolder` into `Configurator` instances that are each invoked when an item is bound in the ViewHolder.

Also moves the ViewHolders and the new Configurators into separate packages instead of them residing in just the `adapter` package.


library/src/main/java/com/getstream/sdk/chat/adapter/viewholders/configurators/IndicatorConfigurator.kt
```diff
@@ -0,0 +1,128 @@
...
+        configReadIndicator(messageItem)
+        configParamsReadIndicator(messageItem)
+    }
+
+    private fun configDeliveredIndicator(messageItem: MessageItem) {
+        binding.ivDeliver.isVisible = false
+        binding.pbDeliver.isVisible = false
+
+        val message = messageItem.message
+        val lastMessage = LlcMigrationUtils.computeLastMessage(channel)
```
JcMinarro, line 44: I'm not sure if it would be work fine.
I know it was implemented on that way previously, but I am not sure if `channel::messages` is updated (In the future I would like to do it immutable) 
Maybe we could add this info into the `MessageItem` instead of have the channel injected by constructor

JcMinarro (commented):
Looks good the approach you are following ;)

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemViewHolder.kt
```diff
@@ -1,735 +0,0 @@
...
-import io.getstream.chat.android.client.models.Channel
-import io.getstream.chat.android.client.models.ChannelUserRead
-import io.getstream.chat.android.client.models.Message
-import io.getstream.chat.android.client.utils.SyncStatus
-import top.defaults.drawabletoolbox.DrawableBuilder
-import java.text.DateFormat
-import java.text.SimpleDateFormat
-import java.util.Arrays
-
-class MessageListItemViewHolder(
```
zsmb13, line 49: Not actually deleted, just changed a lot and moved to a new package 😄

library/src/main/java/com/getstream/sdk/chat/adapter/viewholder/message/configurators/MessageTextConfigurator.kt
```diff
@@ -0,0 +1,201 @@
...
+) : Configurator {
+
+    override fun configure(
+        messageItem: MessageItem
+    ) {
+        val message = messageItem.message
+
+        if (message.text.isEmpty() && !message.isDeleted()) {
+            binding.tvText.isVisible = false
+            return
```
samiuelson, line 49: How about using else branch instead of return in the middle?

samiuelson (approved):
I don't know what's the desired scope of this PR; I've marked a few repeating code smells that could be improved here and there, but maybe you'll want to do this in the next iterations. 

In general, I like the approach of `Configurator` interface; it's also a bit tempting for me to go for a more functional style here, but probably more efficient and safe to use interface and objects 👍🏼

library/src/main/java/com/getstream/sdk/chat/adapter/viewholder/message/configurators/MarginConfigurator.kt
```diff
@@ -0,0 +1,42 @@
...
+        configMarginStartEnd(binding.tvMessagedate)
+    }
+
+    private fun configMarginStartEnd(view: View) {
+        val avatarWidth = style.avatarWidth
+
+        view.updateLayoutParams<ConstraintLayout.LayoutParams> {
+            when (view) {
+                binding.tvUsername -> {
+                    leftMargin = Utils.dpToPx(10 + 5) + avatarWidth
```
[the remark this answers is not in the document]
zsmb13: Agreed, but I have no idea why they're calculated the way they are, and which ones are semantically the same and which ones aren't. Just moved this as-is for now.

library/src/main/java/com/getstream/sdk/chat/adapter/viewholder/message/configurators/MessageTextConfigurator.kt
zsmb13: I'm actually a fan of doing early returns like this, as it avoids the nesting that the `else` branch would introduce.

I think it's quite clear here that if the check fails then we don't have anything to display, so we're just hiding the View and skipping the rest of the operations.

JcMinarro (approved):

adasiewiczr (approved):
LGTM

PR #654, opened 2020-09-29
Fix/message input view model

tschellenbach:
- Don't block the main thread when running use cases


library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
```diff
@@ -139,6 +140,7 @@ internal class ChatSocketListener(
             is ConnectingEvent,
             is DisconnectedEvent,
             is ErrorEvent,
+            is ChannelUpdatedByUserEvent,
```
JcMinarro, line 143: It doesn't exist on the current version

JcMinarro (changes requested):

library/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemAdapter.kt
```diff
@@ -1,5 +1,6 @@
 package com.getstream.sdk.chat.adapter
 
+import android.os.Looper
```
samiuelson, line 3: Looks like the import is not used?

sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
```diff
@@ -35,13 +35,14 @@ class App : Application() {
...
+
+            // see crash/bug here: https://github.com/InsertKoinIO/koin/issues/871
+            koin.loadModules(listOf(
                     appModule,
                     dataModule,
                     loginModule
-                )
-            )
+            ))
+            koin.createRootScope()
```
samiuelson, line 45: I think we can move forward getting rid of koin in sample app. It turns out we use mostly singleton dependencies and don't need any advanced DI features. This will also make sample app's code more independent of external libs - easier to consume by users.

library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
samiuelson: Should be here already

JcMinarro:
This PR is breaking our tests. We shouldn't merge any PR until it passes our checkstyle and tests, because it will waste time for other devs on their development process

PR #658, opened 2020-09-30
Use ViewBinding in attachment ViewHolders

zsmb13:
Same update as performed on the Message ViewHolder classes in https://github.com/GetStream/stream-chat-android/pull/647


library/src/main/java/com/getstream/sdk/chat/adapter/viewholder/attachment/AttachmentViewHolder.kt
```diff
@@ -1,46 +1,38 @@
...
-        itemView.findViewById(R.id.tv_media_play)
-    private val tv_media_des: TextView =
-        itemView.findViewById(R.id.tv_media_des)
+class AttachmentViewHolder(
+    parent: ViewGroup,
+    private val binding: StreamItemAttachmentBinding =
+        StreamItemAttachmentBinding.inflate(parent.inflater, parent, false)
+) : BaseAttachmentViewHolder(binding.root) {
+
+    private val mediaBinding: StreamItemAttachMediaBinding = binding.clAttachmentMedia
```
zsmb13, line 35: Stores the binding for the included layout, so that it's easier to access, without having to go through `binding.clAttachmentMedia` for each View.

JcMinarro (approved):
It looks good with the "Binding" approach

adasiewiczr (approved):

PR #659, opened 2020-09-30
Clean up build warnings

zsmb13:
Fixes some deprecations / other build warnings


adasiewiczr (approved):

adasiewiczr (dismissed):

library/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -70,7 +70,7 @@ class MessageListViewModel @JvmOverloads constructor(
                 domain.useCases.deleteMessage(event.message).enqueue()
             }
             is Event.FlagMessage -> {
-                client.flag(event.message.user.id).enqueue()
+                client.flagUser(event.message.user.id).enqueue()
```
JcMinarro, line 73: It seems a mistake.
The event is `FlagMessage` but you are flagging an user

library/src/test/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModelTest.kt
```diff
@@ -175,7 +175,7 @@ class MessageListViewModelTest {
 
         viewModel.onEvent(MessageListViewModel.Event.FlagMessage(MESSAGE))
 
-        verify(client).flag(MESSAGE.user.id)
+        verify(client).flagUser(MESSAGE.user.id)
```
JcMinarro, line 178: This test should be updated as well

JcMinarro (changes requested):

library/src/main/java/com/getstream/sdk/chat/adapter/viewholder/message/configurators/UsernameAndDateConfigurator.kt
```diff
@@ -41,7 +41,7 @@ internal class UsernameAndDateConfigurator(
 
         if (style.isMessageDateShow) {
             binding.tvMessagedate.isVisible = true
-            binding.tvMessagedate.text = TIME_DATEFORMAT.format(messageItem.message.createdAt)
+            binding.tvMessagedate.text = TIME_DATEFORMAT.format(messageItem.message.createdAt!!)
```
leandroBorgesFerreira, line 44: You may have a NullPointerException here, as createdAt is optional. Maybe:
```
binding.tvMessagedate.text = messageItem.message.createdAt.let(TIME_DATEFORMAT::format)
```

To avoid the NPE.

zsmb13: We've always assumed that this exists, there was never a `null` check here previously. I'd rather have this very obviously break if it somehow ends up being a `null` value, as opposed to just silently not updating the date text.

The property itself is nullable because of how we have serialization set up, kind of an unfortunate side effect of rolling with Gson.

JcMinarro: I think it is nullable because local messages that has not been sent yet doesn't have `createAt` Date.
There is a related task about that already in our jira: https://stream-io.atlassian.net/browse/CAS-224

adasiewiczr (approved):

JcMinarro (approved):

PR #663, opened 2020-10-02
Update GitHub PR template

zsmb13:
### Description

Updates our default PR template with more relevant checklist items to us, while keeping the CLA signing requirement so that it can be used by external contributors as well. This is a meta-PR, see the format here for the proposed items of the template.

(Plus it moves the template file into the `.github` directory, so that it's not in the way in the root of the project)

### Checklist

- [ ] ~~Changelog updated with client-facing changes~~
- [ ] ~~New code is covered by unit tests~~
- [ ] ~~Comparison screenshots added for visual changes~~
- [x] Reviewers added


adasiewiczr (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,14 @@
...
+Describe what this PR changes, why we're making the change, how it should be tested...
+
+### Checklist
+
+- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
+- [ ] PR targets the `develop` branch
+- [ ] Changelog updated with client-facing changes
+- [ ] New code is covered by unit tests
+- [ ] Comparison screenshots added for visual changes
+- [ ] Reviewers added
```
ogkuzmin, line 14: Not sure, but what about optional checkpoint: "Files written in Java with new changes are converted to Kotlin@

ogkuzmin (approved):

samiuelson (approved):
Great!
One thing that maybe could be useful as well, would be to mention devices and OS version the PR was tested against by the author 🤔

JcMinarro (approved):

.github/pull_request_template.md
zsmb13: I think that's a nice to have, but I wouldn't enact it as a policy to convert to Kotlin if you're touching a file. It's generally harder to review a conversion + changes made at the same time than doing those separately.

PR #667, opened 2020-10-05
[CAS-262] Improve sample app - add checking current destination before navigating to ChannelFragment

adasiewiczr:
Sample app was crashing when two navigation events were triggered almost at the same time (e.g. multi-touch on ChannelsFragment). That's because the navigation controller changes the destination immediately after receiving the first event. I've added checking if the current destination is correct before navigating to Channel fragment


sample/src/main/kotlin/io/getstream/chat/sample/feature/channels/ChannelsFragment.kt
```diff
@@ -37,7 +37,11 @@ class ChannelsFragment : Fragment(R.layout.fragment_channels) {
 
     private fun setupOnClickListeners() {
         channelsListView.setOnChannelClickListener {
-            findNavController().navigate(ChannelsFragmentDirections.actionOpenChannel(it.cid))
+            with(findNavController()) {
+                if (currentDestination?.id == R.id.channelsFragment) {
+                    navigate(ChannelsFragmentDirections.actionOpenChannel(it.cid))
```
zsmb13, line 42: Don't we need this for every navigation we ever trigger? Maybe we should do something like [these extensions](https://github.com/woocommerce/woocommerce-android/pull/2401/files#diff-e012971d96c7f53a38c50392606490de)?

zsmb13 (approved):

PR #671, opened 2020-10-07
[CAS-283] Implement message retry handling

zsmb13:
https://stream-io.atlassian.net/browse/CAS-283

Implements the missing "click to retry" functionality for failed messages. (Plus some minor cleanup of adjacent project utilities.)

### DIFF
--- a/library/build.gradle
+++ b/library/build.gradle
@@ -62,7 +62,6 @@ android {
         all {
             it.java.srcDir "src/$it.name/kotlin"
         }
-        main.java.srcDirs += 'src/main/kotlinX'
     }
 
     kotlinOptions {
--- a/library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
@@ -142,7 +142,6 @@ internal class ChatSocketListener(
             is ConnectingEvent,
             is DisconnectedEvent,
             is ErrorEvent,
-            is ChannelUpdatedByUserEvent,
             is UnknownEvent -> { }
         }.exhaustive
     }
--- a/library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
@@ -1,6 +1,7 @@
 package com.getstream.sdk.chat
 
 import com.getstream.sdk.chat.enums.OnlineStatus
+import com.getstream.sdk.chat.utils.exhaustive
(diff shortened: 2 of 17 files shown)


library/build.gradle
```diff
@@ -62,7 +62,6 @@ android {
         all {
             it.java.srcDir "src/$it.name/kotlin"
         }
-        main.java.srcDirs += 'src/main/kotlinX'
```
JcMinarro, line 65: Why extension has been removed from the library?

library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
```diff
@@ -141,7 +142,6 @@ internal class ChatSocketListener(
             is ConnectingEvent,
             is DisconnectedEvent,
             is ErrorEvent,
-            is ChannelUpdatedByUserEvent,
```
JcMinarro, line 144: Why this event is not handled anymore?
Does it compile?

zsmb13: It was a duplicate in the list of events, listed twice in the same branch. Removing it removes a warning.

library/build.gradle
zsmb13: I moved them into the main source set from the separate `kotlinX` source set, simplifies the project. It was two files and ~10 lines in their own source set.

library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
JcMinarro: It seems that was added by two different PRs

library/build.gradle
JcMinarro: The idea with that is to have a common module where all "Third-party class extensions" live, so whenever anyone needs to check if an extension exists or create a new one, he/she knows where to go.
On the other hand, using the same package name of the class over which the extension acts, the imports are simpler and they start to make sense

zsmb13: I've never seen these forced into the package of the original types, always had them packaged under the project's own package name. After all, it's part of our codebase. Importing them from packages like `java.util` can only make things more confusing for developers. Discoverability is the same either way, and in general extensions should be found through IDE tooling, e.g. autocompletion.

It's a good idea to have them in a dedicated place so that we can look them up, how about either a `com.getstream.sdk.chat.extensions` package, or maybe under the utils, so in `com.getstream.sdk.chat.utils.extensions`? In there, we can have files named after the types that the extensions are for, like `Date.kt` or `When.kt`.

zsmb13:
Replaced by #673 and #674

PR #673, opened 2020-10-07
[CAS-283] Tap to retry failed message

zsmb13:
https://stream-io.atlassian.net/browse/CAS-283

Implements the missing "click to retry" functionality for failed messages.

Replaces #671


library/src/main/java/com/getstream/sdk/chat/ChatSocketListener.kt
```diff
@@ -141,7 +141,6 @@ internal class ChatSocketListener(
             is ConnectingEvent,
             is DisconnectedEvent,
             is ErrorEvent,
-            is ChannelUpdatedByUserEvent,
```
zsmb13, line 144: This was a duplicate type in this branch of the `when` statement, just cleaned up a warning by removing it

library/src/main/java/com/getstream/sdk/chat/adapter/viewholder/message/configurators/MessageTextConfigurator.kt
```diff
@@ -160,10 +161,13 @@ internal class MessageTextConfigurator(
...
     ) {
         binding.tvText.setOnClickListener {
-            if (message.isFailed() && !ChatClient.instance().isSocketConnected()) {
-                return@setOnClickListener
+            if (message.isFailed()) {
+                if (ChatClient.instance().isSocketConnected()) {
+                    messageRetryListener.onRetryMessage(message)
+                }
+            } else {
+                messageClickListener.onMessageClick(message)
```
JcMinarro, line 169: The connection logic should be out of the view
```suggestion
            if (message.isFailed()) {
                    messageRetryListener.onRetryMessage(message)
            } else {
                messageClickListener.onMessageClick(message)
```

On the other hand, why don't we attach the listener only on the "Message wasn't sent, try again" text?

zsmb13: I removed the network check. Tested it, and the UX is pretty okay even if we don't have a connection when we perform the retry.

Also improved the condition, it now matches the condition for displaying the tap-to-retry text exactly.

JcMinarro (approved):
LGTM

PR #674, opened 2020-10-07
Remove separate kotlinX source set for extensions

zsmb13:
Removes the separate source set we have for extensions, moves them into the main source set into the `utils` package.


ogkuzmin (approved):

JcMinarro (approved):

library/src/main/java/com/getstream/sdk/chat/utils/Primitives.kt
```diff
@@ -1,7 +1,7 @@
-package com.getstream.sdk.chat
+package com.getstream.sdk.chat.utils
 
 internal inline fun Boolean.whenTrue(crossinline f: () -> Unit): Boolean = also { if (this) f() }
 internal inline fun Boolean.whenFalse(crossinline f: () -> Unit): Boolean = also { if (!this) f() }
 
-internal val <T> T.exhaustive: T
+internal inline val <T> T.exhaustive: T
```
carterhudson, line 6: I've seen this used on a few `when` statements in the codebase - is this just a tagging convention for noting that a certain when statement is intentionally exhaustive vs using `else ->`?

JcMinarro: If you add `else ->`, whatever "new branch" is added to the `when`  by an external factor (new sealed class option, for example) will go to this statement, but maybe it should be handled in another different case. On the other hand, `else ->` is not required if all cases are handled, and it is only required if `when` return anything and it is assigned to a variable or as returned value, but sometimes you need to do a when that doesn't return anything, if we don't use this extension, we won't be notified that it needs to be handled on that cases

PR #676, opened 2020-10-07
[CAS-270] scroll problem in message list view

leandroBorgesFerreira:
[CAS-270](https://stream-io.atlassian.net/browse/CAS-270)
### Description
The diff utils calculate the diferences of messages in a new thread and then notifies the adapter about it in a later moment, asynchronously. This code was being treated as a synchronous code, we were requesting the recycler view to scroll to the bottom right after new itens were added. When the scroll-to-the-bottom request was made before the list could be updated, nothing happened and the list didn't scroll to the correct position. 

In this PR the logic to scroll to the bottom only happens after the recycler view is correctly updated and then no more items need to be removed/moved/added. This fixes the scrolling problem. 

On top of this correction this is a refactor in the class `MessageListItemLiveData` to improve its thread safety. Using local variables instead of final parameters in the methods can be dangerous as those values can change by other threads. In this PR the class variables are used only to keep the last state of a variable. 

Take a look at the result: 

Oct-08-2020 21-54-37

(Now it is scrolling)

### Checklist

- [x] Reviewers added

Bug caught!

giphy
(_Ok, I know that this guy is not catching the bug, but still... the gif makes sense_)


library/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.java
```diff
@@ -50,19 +57,25 @@ public MessageListItemLiveData(User currentUser,
         this.currentUser = currentUser;
         this.typing = typing;
         this.reads = reads;
-        this.messageEntities = new ArrayList<MessageListItem>();
+        this.messageEntities = new ArrayList<>();
         this.typingEntities = new ArrayList<>();
         this.isLoadingMore = false;
         // scroll behaviour is only triggered for new messages
         this.lastMessageID = "";
         this.hasNewMessages = false;
```
JcMinarro, line 65: After reviewing this class, I can see this field is not needed anymore. We are always to updating it before we call to `broadcastValue()` method. 
There is a public method that returns this value, but it is not used anymore, and I think it doesn't provide any useful value, so we can remove it

leandroBorgesFerreira: We need it to broadcast the current state of message entities, like:

```
private void onReadsChanged(List<ChannelUserRead> channelUserRead) {
        hasNewMessages = false;
        logger.logI("broadcast because reads changed");
        broadcastValue(hasNewMessages, isLoadingMore, typingEntities, messageEntities);
    }
```

Unless we can broadcast this as an empty list.

JcMinarro: Yep, but you only need to pass it on the method, it is not needed to be stored into the class

leandroBorgesFerreira: fixed

JcMinarro (dismissed):

JcMinarro (approved):

PR #677, opened 2020-10-08
Improve attachments related code

adasiewiczr:
### Description

- Add setting type and file size when creating AttachmentMetaData from file
- Remove unused FileAttachmentListAdapter
- Remove unused methods related to opening files chooser
- Replace isMedia flag with getting type from attachment if possible
- Adjust test to new implementation

### Checklist

- [x] Reviewers added


library/src/main/java/com/getstream/sdk/chat/view/messageinput/AttachmentsController.kt
```diff
@@ -41,31 +39,27 @@ internal class AttachmentsController(
...
+                setTotalMediaAttachmentAdapter(
                     totalAttachments.toList(),
                     selectedAttachments.toList(),
-                    messageInputType,
-                    isMedia
+                    messageInputType
                 )
             }
-            setSelectedAttachmentAdapter(messageInputType, isMedia)
+            setSelectedAttachmentAdapter(messageInputType, true)
```
ogkuzmin, line 62: Previously, there was one codebase with one logic for media and non-media files. That's why we had to pass isMedia flag, now there is not any reason to do that. I'd like to move out it completely. What if we divide method `setSelectedAttachmentAdapter` for two (media, non-media) and invoke it separately. I think it makes code cleaner and easier.
Also we can omit `showSelectedAttachments` and invoke view directly to show media non media attachments.

library/src/main/java/com/getstream/sdk/chat/view/messageinput/AttachmentsController.kt
```diff
@@ -249,23 +196,20 @@ internal class AttachmentsController(
             val attachments = withContext(dispatchersProvider.ioDispatcher) {
                 storageHelper.getAttachmentsFromUriList(view.context, uriList)
             }
-            attachments.forEach { selectAttachment(it, false) }
+            attachments.forEach { selectAttachment(it) }
```
ogkuzmin, line 199: For what reason do you do it iteratively instead of adding all files once? It's not so good from a performance point of view. I'd like to create a separate method for select one attachment (actually it works only for media mode) and add all attachments (adding all list to selected ones by one invocation)  from the platform picker.
It again divides code for two different modes (media/ non-media) that makes code cleaner.

library/src/main/java/com/getstream/sdk/chat/view/messageinput/MessageInputView.kt
```diff
@@ -356,12 +355,6 @@ class MessageInputView(context: Context, attrs: AttributeSet?) : RelativeLayout(
         gridLayoutManager.spanCount = MEDIA_ITEMS_PER_ROW
     }
 
-    internal fun showTotalFileAttachments(totalFileAttachmentAdapter: FileAttachmentListAdapter) {
-        gridSpacingItemDecoration.setSpanCount(FILE_ITEMS_PER_ROW)
```
ogkuzmin, line 360: Do we still need the constants? (MEDIA_ITEMS_PER_ROW, FILE_ITEMS_PER_ROW) Maybe just use them in init...

ogkuzmin (changes requested):
Good PR that makes code easier. I have some suggestions that would make code even more direct

library/src/main/java/com/getstream/sdk/chat/view/messageinput/MessageInputView.kt
adasiewiczr: Nice one! Thanks!

library/src/main/java/com/getstream/sdk/chat/view/messageinput/AttachmentsController.kt
adasiewiczr: agree

ogkuzmin (approved):
Very nice!

PR #688, opened 2020-10-13
[CAS-270] Fixes for custom scroll button

leandroBorgesFerreira:
[CAS-270](https://stream-io.atlassian.net/browse/CAS-270)
### Description

This PR includes code to disable the default button for scroll in the MessageListView. That's very important for the user to be able to add its own view for the button and don't show the default one. Perhaps a user wants to include its own implementation like this:

```
<com.getstream.sdk.chat.view.MessageListView
        android:id="@+id/messageListView"
        android:layout_width="0dp"
        android:layout_height="0dp"
        app:streamMessageDateShow="true"
        android:clipToPadding="false"
        android:paddingBottom="@dimen/padding_medium"
        app:streamNewMessagesBehaviour="count_new_messages"
        app:streamDefaultButtonEnabled="false"
        >

        <Button
            android:id="@+id/scrollBtn"
            android:layout_width="wrap_content"
            android:layout_height="wrap_content"
            android:layout_marginBottom="10dp"
            android:animateLayoutChanges="true"
            android:text="Scroll - 1 New message"
            app:layout_constraintBottom_toBottomOf="parent"
            app:layout_constraintEnd_toEndOf="parent"
            app:layout_constraintStart_toStartOf="parent"
            />

    </com.getstream.sdk.chat.view.MessageListView>
```

That would like this:

Screenshot 2020-10-13 at 23 47 44

Ok... maybe that button is a bit ugly, but you get the idea.

### Checklist

- [x] Reviewers added

### DIFF
--- a/library/src/main/res/values/attrs.xml
+++ b/library/src/main/res/values/attrs.xml
@@ -171,6 +171,7 @@
     <!--MessageListView-->
     <declare-styleable name="MessageListView">
         <!--Scroll button-->
+        <attr name="streamDefaultButtonEnabled" format="boolean" />
         <attr name="streamButtonBackground" format="reference" />
         <attr name="streamButtonIcon" format="reference" />
         <attr name="streamNewMessagesTextSingle" format="string" />
--- a/library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
+++ b/library/src/main/java/com/getstream/sdk/chat/view/MessageListView.java
@@ -559,6 +559,10 @@ public class MessageListView extends ConstraintLayout {
                 && passedTime < 3000;
     }
 
+    public void scrollToBottom() {
+        layoutManager.scrollToPosition(adapter.getItemCount() - 1);
+    }
+
     /**
      * Sets the message click listener to be used by MessageListView.
      *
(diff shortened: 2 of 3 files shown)


library/src/main/res/values/attrs.xml
```diff
@@ -171,6 +171,7 @@
     <!--MessageListView-->
     <declare-styleable name="MessageListView">
         <!--Scroll button-->
+        <attr name="streamDefaultButtonEnabled" format="boolean" />
         <attr name="streamButtonBackground" format="reference" />
         <attr name="streamButtonIcon" format="reference" />
```
zsmb13, line 176: Could we give these attributes a more specific name than "Button"? Without looking at the comment above them, it's hard to tell what these do, and that's not there when clients are setting attributes on `MessageListView` in their own layout files.

zsmb13 (commented):
Just one small note, looks good otherwise 👍

library/src/main/res/values/attrs.xml
leandroBorgesFerreira: ScrollButton perhaps?

zsmb13: Yup, that sounds descriptive enough 👍 . If we change it, let's make sure it's in the changelog as well, because these attributes might already be used by some clients (though they're pretty new 😄 ).

leandroBorgesFerreira:
I recreated those changes inside #703

PR #689, opened 2020-10-14
Fix keyboard utility methods

zsmb13:
# Description

Changed the keyboard utility methods taking an `Activity` as a parameter to take a `Context` instead, as it's often enough to perform the action.

Added two new utilities that take `View` instances as parameters, as we usually invoke these methods from our custom Views.

This fixes some crashes where our custom Views are used in Fragments that have non-Activity `Context`s.

### Checklist

- [] ~Changelog updated with client-facing changes~
- [] ~Comparison screenshots added for visual changes~
- [x] Reviewers added

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/utils/Utils.java
+++ b/library/src/main/java/com/getstream/sdk/chat/utils/Utils.java
@@ -122,8 +122,6 @@ public class Utils {
     }
 
     public static void hideSoftKeyboard(@NonNull Context context) {
-        InputMethodManager imm = (InputMethodManager) context.getSystemService(Activity.INPUT_METHOD_SERVICE);
-
         View view = null;
         if (context instanceof Activity) {
             // Find the currently focused view, so we can grab the correct window token from it.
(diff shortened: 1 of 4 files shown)


samiuelson (approved):

library/src/main/java/com/getstream/sdk/chat/utils/Utils.java
```diff
@@ -117,20 +115,37 @@ public static void showMessage(Context context, String message) {
...
+        if (!imm.isAcceptingText())
+            imm.toggleSoftInput(InputMethodManager.SHOW_FORCED, 0);
     }
 
-    public static void hideSoftKeyboard(Activity activity) {
-        InputMethodManager imm = (InputMethodManager) activity.getSystemService(Activity.INPUT_METHOD_SERVICE);
-        //Find the currently focused view, so we can grab the correct window token from it.
-        View view = activity.getCurrentFocus();
-        //If no view currently has focus, create a new one, just so we can grab a window token from it
+    public static void hideSoftKeyboard(@NonNull Context context) {
```
JcMinarro, line 124: This method could call to `public static void hideSoftKeyboard(@NonNull View view)` method, and reuse the same logic

zsmb13: True, I was doing this the other way around with the _show_ methods already, not sure why I didn't do it here. Updated the PR 😄

JcMinarro (approved):

PR #695, opened 2020-10-15
[CAS-292] Fix GIFs loading issues

adasiewiczr:
- Replace Glide with Coil in AttachmentViewHolderMedia
- Improve showing loading indicator on media

Glide sometimes has problems with GIFs loading (probably it cannot decode them) and it doesn't provide any helpful exception. On the other hand, Coil renders everything correctly and its performance is much better comparing to Glide

Describe what this PR changes, why we're making the change, how it should be tested...

### Checklist

- [x] Reviewers added

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
@@ -4,11 +4,15 @@ import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.drawable.BitmapDrawable
 import android.os.Build
+import android.widget.ImageView
+import androidx.annotation.DrawableRes
 import coil.Coil
+import coil.api.load
 import coil.decode.GifDecoder
 import coil.decode.ImageDecoderDecoder
 import coil.request.GetRequest
 import coil.request.GetRequestBuilder
+import coil.request.RequestDisposable
 import coil.size.Precision
 import coil.transform.BlurTransformation
 import coil.transform.CircleCropTransformation
@@ -37,8 +41,27 @@ internal object ImageLoader {
             }
     }
 
-    fun getImageLoaderWithGifSupport(
+    fun ImageView.loadWithGifSupport(
+        uri: String?,
+        @DrawableRes placeholderResId: Int?,
+        onStart: () -> Unit = {},
+        onCancel: () -> Unit = {},
+        onError: (throwable: Throwable) -> Unit = {},
+        onSuccess: () -> Unit = {},
+    ): RequestDisposable =
+        load(uri, getImageLoaderWithGifSupport(context)) {
+            placeholderResId?.let { placeholder(it) }
+            listener(
+                onStart = { onStart() },
+                onCancel = { onCancel() },
+                onError = { _, throwable -> onError(throwable) },
+                onSuccess = { _, _ -> onSuccess() },
+            )
+        }
+
+    private fun getImageLoaderWithGifSupport(
         context: Context,
+        // TODO: We should probably allowHardware for performance improvements but we do software rendering in PorterShapeImageView
         allowHardware: Boolean = false,
         precision: Precision = Precision.EXACT
     ): coil.ImageLoader {
--- a/library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
@@ -12,7 +12,6 @@ import coil.decode.GifDecoder
 import coil.decode.ImageDecoderDecoder
 import coil.request.GetRequest
 import coil.request.GetRequestBuilder
-import coil.request.RequestDisposable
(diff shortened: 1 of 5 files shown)


library/src/main/java/com/getstream/sdk/chat/adapter/viewholder/attachment/AttachmentViewHolderMedia.kt
```diff
@@ -5,7 +5,8 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
-import com.bumptech.glide.Glide
+import coil.api.load
```
JcMinarro, line 8: Because we have a wrapper (ImageLoader) over `Coil`, all "Coil Stuff" should be don't inside of it. Then, if we would need to switch to another alternative, it is easy to do, only one part of our code

JcMinarro (changes requested):

zsmb13:
Would it be way out of scope for this PR to remove Glide entirely? Having two image loaders feels like our library is getting a bit bloated, and if we don't remove it now, I'm afraid that we'll have both of them included for a long, long time.

adasiewiczr:
@zsmb13 I've been talking about that with Jc yesterday and we agreed that we will wait for some time to check if Coil works properly. I've already created a separate task for that - CAS-301

zsmb13:
Alright, so we'll merge this now, see if attachment images work fine for a while, and then check in in a couple weeks, and decide which way to go with it?

adasiewiczr:
yes, that was the idea

library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -33,6 +41,43 @@ internal object ImageLoader {
...
+                onStart = { onStart() },
+                onCancel = { onCancel() },
+                onError = { _, throwable -> onError(throwable) },
+                onSuccess = { _, _ -> onSuccess() },
+            )
+        }
+
+    private fun getImageLoaderWithGifSupport(
+        context: Context,
+        // TODO: We should probably allowHardware for performance improvements but we do software rendering in PorterShapeImageView
```
zsmb13, line 64: Does using software rendering in `PorterShapeImageView` conflict with this? The docs for Coil say to disable hardware only if we run into issues otherwise.

library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -33,6 +41,43 @@ internal object ImageLoader {
...
+    ): coil.ImageLoader {
+        return coil.ImageLoader.Builder(context)
+            .allowHardware(allowHardware)
+            .precision(precision)
+            .componentRegistry {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+                    add(ImageDecoderDecoder())
+                } else {
+                    add(GifDecoder())
+                }
```
zsmb13, line 76: Nice 👍

library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -33,6 +41,43 @@ internal object ImageLoader {
             }
     }
 
+    fun ImageView.loadWithGifSupport(
+        uri: String?,
+        @DrawableRes placeholderResId: Int?,
+        onStart: () -> Unit = {},
+        onCancel: () -> Unit = {},
+        onError: (throwable: Throwable) -> Unit = {},
+        onSuccess: () -> Unit = {},
```
zsmb13, line 50: Nit: We could wrap all of this up in an `onComplete`, if we don't care about the details at the call site for now.

zsmb13 (approved):
LGMT overall, few couple small comments / questions

library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
adasiewiczr: Unfortunately yes, we are getting ` java.lang.IllegalArgumentException: Software rendering doesn't support hardware bitmaps` when trying to render image inside `PorterShapeImageView`

library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
adasiewiczr: You're right. I've wanted to follow Coil API but it doesn't make sense for now

library/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
zsmb13: Ah, okay, ouch. I guess we can revisit this and try to work around it if / when we remove Glide and do more Coil work.

zsmb13 (approved):

JcMinarro (approved):

Issue #694, opened 2020-10-15
Crash after migrate to the latest androidx beta versions


letos:
Hi! Chat's Views crashed our app after we updated `androidx.fragment` and `androidx.activity` to the latest versions.

**Versions:**
```
implementation 'com.github.GetStream:stream-chat-android:4.3.0-beta-6'

implementation 'androidx.activity:activity-ktx:1.2.0-beta01'
implementation 'androidx.fragment:fragment-ktx:1.3.0-beta01'
```

**Error stacktrace from sample app:**
```
android.view.InflateException: Binary XML file line #43: Binary XML file line #43: Error inflating class com.getstream.sdk.chat.view.messageinput.MessageInputView
    Caused by: android.view.InflateException: Binary XML file line #43: Error inflating class com.getstream.sdk.chat.view.messageinput.MessageInputView
    Caused by: java.lang.reflect.InvocationTargetException
        at java.lang.reflect.Constructor.newInstance0(Native Method)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:430)
        at android.view.LayoutInflater.createView(LayoutInflater.java:645)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:787)
        at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:727)
        at android.view.LayoutInflater.rInflate(LayoutInflater.java:858)
        at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:821)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:518)
        at android.view.LayoutInflater.inflate(LayoutInflater.java:426)
        at androidx.fragment.app.Fragment.onCreateView(Fragment.java:1922)
        at androidx.fragment.app.Fragment.performCreateView(Fragment.java:2950)
        at androidx.fragment.app.FragmentStateManager.createView(FragmentStateManager.java:518)
        at androidx.fragment.app.FragmentStateManager.moveToExpectedState(FragmentStateManager.java:277)
        at androidx.fragment.app.FragmentManager.executeOpsTogether(FragmentManager.java:2177)
        at androidx.fragment.app.FragmentManager.removeRedundantOperationsAndExecute(FragmentManager.java:2094)
        at androidx.fragment.app.FragmentManager.execPendingActions(FragmentManager.java:1990)
        at androidx.fragment.app.FragmentManager$5.run(FragmentManager.java:524)
[...]

PR #697, opened 2020-10-15
Fix crash caused by registering for Activity results too late

zsmb13:
Bugfix for issue #694 

### Description

The `registerForActivityResult` method had a behaviour change in recent Activity / Fragment KTX versions (in ``, specifically) where callbacks can no longer be registered after the Activity got into a `STARTED` state. Instead of relying on the Activity's lifecycle, we can register and unregister our callbacks dynamically, by accessing the underlying ActivityResultRegistry directly.

### Checklist

- [x] Reviewers added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # To be released:
+- Fix a crash related to behaviour changes in 1.3.0-alpha08 of the AndroidX Fragment library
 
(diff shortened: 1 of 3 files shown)


JcMinarro (approved):
Good Job ;)

adasiewiczr (approved):
Nice one!

PR #699, opened 2020-10-16
Message input context wrapper

carterhudson:
### Description

Found more crashes happening when using Hilt & MessageInputView. Hilt makes the view's context a FragmentContextWrapper, so casting to an activity causes a crash.

### Checklist

- [x] Reviewers added

### DIFF
--- a/library/src/main/java/com/getstream/sdk/chat/view/common/Extensions.kt
+++ b/library/src/main/java/com/getstream/sdk/chat/view/common/Extensions.kt
@@ -1,5 +1,7 @@
 package com.getstream.sdk.chat.view.common
 
+import android.content.Context
+import android.content.ContextWrapper
 import android.view.View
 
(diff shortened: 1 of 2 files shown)


library/src/main/java/com/getstream/sdk/chat/view/common/Extensions.kt
```diff
@@ -9,3 +11,11 @@ fun View.visible(isVisible: Boolean) {
         View.GONE
     }
 }
+
+/**
+ * Ensures the context being accessed in a View can be cast to Activity
+ */
+fun Context.ensure() = when (this) {
```
JcMinarro, line 18: Could we do it internal?

carterhudson: Done

JcMinarro (approved):

PR #704, opened 2020-10-17
update channel diff to fix 698

tschellenbach:
Minor cleanup while investigating 698
The prior version of the diff doesn't work now that we have locallyUpdatedAt


library/src/main/java/com/getstream/sdk/chat/adapter/ChannelListDiffCallback.java
```diff
@@ -53,10 +53,6 @@ public boolean areContentsTheSame(int oldItemPosition, int newItemPosition) {
             contentTheSame = false;
         } else if (newChannel.getUpdatedAt() != null && oldChannel.getUpdatedAt().getTime() < newChannel.getUpdatedAt().getTime()) {
             contentTheSame = false;
-        } else if (oldChannel.getLastMessageAt() == null && newChannel.getLastMessageAt() != null) {
-            contentTheSame = false;
-        } else if (newChannel.getLastMessageAt() != null && oldChannel.getLastMessageAt().getTime() < newChannel.getLastMessageAt().getTime()) {
-            contentTheSame = false;
```
tschellenbach, line 59: the message comparison below already checks this

library/src/main/java/com/getstream/sdk/chat/utils/LlcMigrationUtils.java
```diff
@@ -213,15 +213,7 @@ public static ChannelUserRead getRead(Channel channel, String userId) {
...
-        if (oldLastMessage != null &&
-                newLastMessage != null &&
-                newLastMessage.getUpdatedAt() != null &&
-                oldLastMessage.getUpdatedAt() != null &&
-                oldLastMessage.getUpdatedAt().getTime() < newLastMessage.getUpdatedAt().getTime()) {
-            return false;
-        } else {
-            return true;
-        }
+        return oldLastMessage == newLastMessage;
```
tschellenbach, line 216: simplify this check now that the message is a data class with correct comparison methods

carterhudson: Do we need that method anymore if it's a straight equality check? Can we also just do a straight equality check in the diff callback's `areContentsTheSame` method?

tschellenbach: no, then again this entire file needs to be cleaned up/converted to kotlin. don't want to do it now.

adasiewiczr (dismissed):

library/src/main/java/com/getstream/sdk/chat/utils/LlcMigrationUtils.java
zsmb13: In Java, using `==` is a comparison of the references, and won't use the generated `equals` method of the data class. If we want to use `equals`, we also have to perform null checks on the messages first, because (at least according to the `computeLastMessage` method's declaration) either might be `null`, and calling `equals` on that would crash.

zsmb13 (changes requested):
Let's make sure we get the comparisons right, see other comment

library/src/main/java/com/getstream/sdk/chat/utils/LlcMigrationUtils.java
tschellenbach: how about we close this PR and move the whole thing to kotlin?

tschellenbach: well, the current version is broken, hmm.

zsmb13 (approved):

samiuelson (approved):

PR #722, opened 2020-10-21
[CAS-311] Wrap RetrofitCdnApi with FileUploader interface

adasiewiczr:
### Description

Wrap RetrofitCdnApi with FileUploader interface:
- Add StreamFileUploader
- Set StreamFileUploader as a default FileUploader
- Add tests

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -65,100 +63,49 @@ internal class ChatApi(
...
-                channelId,
-                part,
-                apiKey,
-                userId,
-                connectionId
-            )
-            .call
-            .enqueue(RetroProgressCallback(callback))
-    }
+    ) = fileUploader.sendFile(channelType, channelId, file, callback)
```
zsmb13, line 74: Explicit return types for these functions would be nice, hard to tell whether they implicitly return something or not

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -65,100 +63,49 @@ internal class ChatApi(
...
+            fileUploader.deleteFile(channelType, channelId, url)
+            Result(Unit, null)
+        }
     }
 
     fun deleteImage(channelType: String, channelId: String, url: String): Call<Unit> {
-        return retrofitCdnApi.deleteImage(channelType, channelId, apiKey, connectionId, url)
-            .map { Unit }
+        return FileUploaderCall {
+            fileUploader.deleteImage(channelType, channelId, url)
```
zsmb13, line 106: The file _uploader_ interface has methods that _send_ and _delete_ files, we might want to consider finding a better name for it, because this is kind of odd

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileUploaderCall.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import io.getstream.chat.android.client.call.ChatCallImpl
+import io.getstream.chat.android.client.utils.Result
+
+internal class FileUploaderCall<T : Any>(private val call: () -> Result<T>) : ChatCallImpl<T>() {
+
+    override fun execute(): Result<T> = call()
+
+    override fun enqueue(callback: (Result<T>) -> Unit) {
+        callback(call())
```
zsmb13, line 11: Isn't the `enqueue` implementation meant to be async, non-blocking?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -65,100 +63,49 @@ internal class ChatApi(
...
+        }
     }
 
     fun deleteFile(channelType: String, channelId: String, url: String): Call<Unit> {
-        return retrofitCdnApi.deleteFile(channelType, channelId, apiKey, connectionId, url)
-            .map { Unit }
+        return FileUploaderCall {
+            fileUploader.deleteFile(channelType, channelId, url)
+            Result(Unit, null)
+        }
```
zsmb13, line 101: Really like the design of this call ctor 👍

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/StreamFileUploader.kt
```diff
@@ -0,0 +1,122 @@
...
+import okhttp3.RequestBody.Companion.asRequestBody
+import java.io.File
+
+internal class StreamFileUploader(
+    private val apiKey: String,
+    private val retrofitCdnApi: RetrofitCdnApi
+) : FileUploader {
+
+    private var userId: String = ""
+    private var connectionId: String = ""
```
zsmb13, line 18: We could make these nullable, and throw errors if the methods of `FileUploader` are called before `setConnection` sets these values. Would be more explicit than something quietly failing because we pass in an empty string somewhere.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileUploader.kt
```diff
@@ -5,14 +5,16 @@ import java.io.File
...
+
     public fun sendFile(
         channelType: String,
         channelId: String,
         file: File,
         callback: ProgressCallback
     )
 
-    public fun sendFile(channelType: String, channelId: String, file: File): String
+    public fun sendFile(channelType: String, channelId: String, file: File): String?
```
zsmb13, line 17: This interface would be a bit hard to implement for a client just by looking at this file. For example, what are these `send...` methods supposed to return? Why are there two `send...` methods? Would be nice to add KDoc comments for these.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
adasiewiczr: I agree it can be confusing but on the other hand, image is a kind of file

zsmb13: Oh, sorry, it's not the image uploads I wanted to highlight.

What's weird is that we have an _uploader_, which never _uploads_ anything, but it _sends_ and _deletes_ things instead._Send_ is kind of a synonym (though why are we using two different words for the same thing), but _delete_ is the opposite of uploading something, so it's odd for something called an _uploader_ to do that.

That being said, let's _not_ call this `FileManager` or something... I don't know what a good name would be for the interface.

adasiewiczr: oh okay, you're right. I like `FileManager`

zsmb13: >let's _not_ call this `FileManager`

I'd like to avoid the very generic manager naming if possible 😄

adasiewiczr: 🙈 naming is indeed one of the hardest things in Computer Science

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -65,100 +66,49 @@ internal class ChatApi(
     fun setConnection(userId: String, connectionId: String) {
         this.userId = userId
         this.connectionId = connectionId
+        fileUploader.setConnection(userId, connectionId)
```
JcMinarro, line 69: `setConnection` is a bit weird on the `FileUploader` interface.
Maybe we could move those arguments to `send()`, `delete()` methods

adasiewiczr: I was thinking about it. We need it in `userId` and `connectionId` in our implementation but I'm not sure if user's implementations are going to need that. The question is if it's better to add an additional method to the interface or add those parameters to every method even if they won't be used by other users?

JcMinarro: I would prefer to add all argument to the methods, then they don't need to know the internal lifecycle that we use with this class.
If they don't need this argument, they can ignore it

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/uploader/StreamFileUploaderTest.kt
```diff
@@ -0,0 +1,212 @@
...
+            eq(connectionId)
+        )
+    }
+
+    @Test
+    fun `Should call api delete file when deleting file`() {
+        whenever(retrofitCdnApi.deleteFile(any(), any(), any(), any(), any())).thenReturn(
+            RetroSuccess(CompletableResponse()).toRetrofitCall()
+        )
+        streamFileUploader.deleteFile(channelType, channelId, userId, connectionId, url)
```
ogkuzmin, line 186: Nit: There is a classic way of how to structure tests. And usually arrange, act, and assert are divided by empty lines. So for reading purpose it's better to add an empty line before `streamFileUploader.deleteFile(channelType, channelId, userId, connectionId, url)`

ogkuzmin (approved):

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/uploader/StreamFileUploaderTest.kt
adasiewiczr: nice catch, thanks!

ogkuzmin (dismissed):

JcMinarro (dismissed):

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/uploader/StreamFileUploaderTest.kt
```diff
@@ -0,0 +1,222 @@
...
+
+        streamFileUploader.sendFile(channelType, channelId, userId, connectionId, File(""))
+
+        verify(retrofitCdnApi).sendFile(
+            eq(channelType),
+            eq(channelId),
+            file = any(),
+            eq(apiKey),
+            eq(userId),
+            eq(connectionId)
```
zsmb13, line 67: I have mixed feelings about these tests. They mostly just copy the implementation of the `StreamFileUploader` class, which is just to pass all parameters forward to the CDN API. Doesn't seem too meaningful. And for the parameter that involves business logic in our own class, creating the form data from the `File`, we aren't asserting anything.

The question I always ask: _What will it mean if this unit test fails?_ If it just means that we've updated the implementation, so now we have to come to this other file and update it the same way... That doesn't sound very productive.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -72,93 +73,41 @@ internal class ChatApi(
...
-                channelId,
-                part,
-                apiKey,
-                userId,
-                connectionId
-            )
-            .call
-            .enqueue(RetroProgressCallback(callback))
-    }
+    ): Unit = fileUploader.sendFile(channelType, channelId, userId, connectionId, file, callback)
```
zsmb13, line 76: We have _a lot_ of String parameters here, it would be a good idea to add names to the arguments being passed in, so that we clearly see that they're in the right position.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/uploader/StreamFileUploaderTest.kt
adasiewiczr: Good point. Most of them are checking if we call CDN API with proper parameters and I agree that if something change, we will need to change the code in both places. Let me improve that on Monday

JcMinarro (approved):

PR #726, opened 2020-10-21
CAS-241 Custom login screen

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-241

### Description

Resubmitted this PR: 
https://github.com/GetStream/stream-chat-android/pull/719

### Checklist

- [X] Reviewers added


stream-chat-android-sample/src/main/res/values/strings.xml
```diff
@@ -10,4 +10,17 @@
...
+    <string name="login_screen_title">Welcome to Stream Chat</string>
+    <string name="login_summary_text">Get started by filing out the following:</string>
+    <string name="login_button_text">Login</string>
+    <string name="login_development_users_text">Development users</string>
+    <string name="login_field_api_key_hint">Client API key</string>
+    <string name="login_field_user_id_hint">User ID</string>
+    <string name="login_field_user_token">User token</string>
+    <string name="login_field_user_name_hint">User name (optional)</string>
+    <string name="login_sdk_version_template">Stream SDK v %s</string>
+    <string name="login_sample_data">sample data</string>
```
zsmb13, line 24: This string seems unused

stream-chat-android-sample/src/main/res/layout/fragment_login.xml
```diff
@@ -1,32 +1,156 @@
...
+            android:layout_width="match_parent"
+            android:layout_height="48dp"
+            android:background="?selectableItemBackground"
+            android:gravity="center"
+            android:text="@string/login_development_users_text"
+            android:textColor="@color/colorPrimary"
+            android:textSize="18sp"
+            app:layout_constraintBottom_toTopOf="@id/sdkVersion"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent" />
```
zsmb13, line 130: This button on the UI feels a bit hidden to me. For devs taking a first look at the SDK, it might be intimidating to get the form with a ton of fields requiring keys and stuff, and they might not bother trying out our app because of that.

I don't know if we've already made these design considerations, but maybe we could have a more balanced screen, maybe a quick choice like "I just want to try the sample quickly" and "I already have an API key, let me try the app with that".

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/login/LoginFragment.kt
```diff
@@ -6,63 +6,98 @@ import androidx.core.view.isVisible
...
     }
 
-    private fun redirectToChannel(cid: String) {
-        findNavController().navigateSafely(LoginFragmentDirections.actionLoginFragmentToChannelFragment(cid))
+    private fun collectCredentials(): LoginCredentials {
+        return LoginCredentials(
+            apiKey = apiKeyEditText.text.toString().trim(),
+            userId = userIdEditText.text.toString().trim(),
+            userToken = userTokenEditText.text.toString().trim(),
+            userName = userNameEditText.text.toString().trim()
```
zsmb13, line 54: This could be cleaned up a bit with something like

```kotlin
private val EditText.trimmedText: String
        get() = text.trim().toString()
```

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/login/LoginFragment.kt
```diff
@@ -6,63 +6,98 @@ import androidx.core.view.isVisible
...
+    private fun clearValidationErrors() {
+        apiKeyEditText.error = null
+        userIdEditText.error = null
+        userTokenEditText.error = null
     }
 
     private fun changeLoadingIndicatorVisibility(isVisible: Boolean) {
         loadingProgressBar.isVisible = isVisible
-        usersList.isVisible = !isVisible
+    }
```
zsmb13, line 88: At this point, we might as well remove this method, it's body is a single line that's shorter than actually calling it

zsmb13 (commented):
Looks good overall, my most important comment about is about the discoverability of the dev users button below, the others are just small notes

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/ChatInitializer.kt
```diff
@@ -0,0 +1,23 @@
+package io.getstream.chat.sample.application
+
+import android.content.Context
+import com.getstream.sdk.chat.Chat
+import io.getstream.chat.android.client.notifications.handler.NotificationConfig
+import io.getstream.chat.sample.R
+
+class ChatInitializer(private val context: Context) {
```
samiuelson, line 8: 👍🏼

samiuelson (dismissed):

stream-chat-android-sample/src/main/res/layout/fragment_login.xml
bychkovdmitry: Lets leave it as is for now, the design is about to change anyway

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
```diff
@@ -1,35 +1,24 @@
...
-            val notificationConfig =
-                NotificationConfig(
-                    firebaseMessageIdKey = "message_id",
-                    firebaseChannelIdKey = "channel_id",
-                    firebaseChannelTypeKey = "channel_type",
-                    smallIcon = R.drawable.ic_chat_bubble
-                )
-            notificationHandler = SampleNotificationHandler(this@App, notificationConfig)
-        }.build()
+        chatInitializer.init(appConfig.apiKey)
```
adasiewiczr, line 21: I'm just thinking if it won't be misleading that we initialize the Chat twice (second time in LoginViewModel). WDYT?

samiuelson: Good point, maybe let's remove it from `App.kt`

bychkovdmitry: Yes, just a temp solution. OfflineSyncFirebaseMessagingService crashes the app on start because we don't have apiKey set. Will fix that in another PR.

zsmb13: One thing to keep in mind here is that having these multiple sign-in options complicates the code of our SDK integration. For someone new to the SDK, the extra `ChatInitializer` we create, and initializing it in multiple places might be hard to understand at first.

For most clients, we just want them to initialize everything in their `Application` subclass, once, don't we? With these changes, our sample will no longer show off that approach.

I don't have a solution, because we do need the multiple sign-in options here, I'm just pointing out the potential issue of the sample no longer being that good of a sample for clients.

bychkovdmitry: I don't see the point in using this custom login screen. For developer considering the sample app it is easier to modify AppConfig with their api key and user token.

bychkovdmitry: I updated the PR and removed chat initialization from App. To prevent the app from crashing had to add a null check into `OfflineSyncFirebaseMessagingService.onNewToken()`.

adasiewiczr: Agree, I don't have a better idea of how to handle that but I think we should leave Chat initialization in App class and give an explanation in `LoginViewModel` why do we need to initialize it for the second time.

adasiewiczr:
I'm not sure if we support dark mode at the moment, but:
Screenshot 2020-10-22 at 12 10 44

zsmb13:
> I'm not sure if we support dark mode at the moment

Not really, most things in the sample are broken with dark mode. But we have [a pending ticket](https://stream-io.atlassian.net/browse/CAS-229) for it, and should get to it soon.

bychkovdmitry:
> I'm not sure if we support dark mode at the moment, but:
> Screenshot 2020-10-22 at 12 10 44

Anyway, I've added minor changes to support dark mode on this screen

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
zsmb13: I'm leaning towards what @adasiewiczr proposed above ^, as it shows the regular init we expect people to do. Samples are important 😄.

bychkovdmitry: I've added again Chat init to the App class and added some explanation to the viewmodels

zsmb13 (approved):
LGTM

adasiewiczr (approved):

PR #727, opened 2020-10-21
Adding flipper

leandroBorgesFerreira:
### Description

Adding Flipper. This can be a very good tool to debug our databases. 

### Checklist

- [x] Reviewers added


zsmb13:
We already have Stetho in the project, so maybe we should remove that if we're adding Flipper. Let's have a discussion about these 😄

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
```diff
@@ -30,6 +37,19 @@ class App : Application() {
                 )
             notificationHandler = SampleNotificationHandler(this@App, notificationConfig)
         }.build()
+
+        configFlipper()
```
ogkuzmin, line 41: It's good to split App for debug and other build types and make flipper a debug dependency

ogkuzmin (changes requested):

ogkuzmin:
> We already have Stetho in the project, so maybe we should remove that if we're adding Flipper. Let's have a discussion about these 😄

I agree. It's better to remove stetho

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
```diff
@@ -30,6 +37,19 @@ class App : Application() {
...
+    }
+
+    private fun Application.configFlipper() {
+        SoLoader.init(this, false);
+
+        if (BuildConfig.DEBUG && FlipperUtils.shouldEnableFlipper(this)) {
+            AndroidFlipperClient.getInstance(this).apply {
+                addPlugin(InspectorFlipperPlugin(this@App, DescriptorMapping.withDefaults()))
+                addPlugin(DatabasesFlipperPlugin(this@App))
```
ogkuzmin, line 50: So instead of Stetho let's add network plugin

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
leandroBorgesFerreira: I thought it wouldn't be a problem since this is just the sample app. Splitting the Application into Debug and Release version will work, but it is simpler to use the no-op version of Flipper for release. Like it is suggested: https://fbflipper.com/docs/getting-started/android-native. Just for the sample app, of course, we don't want to send this dependency together with our SDKs.

zsmb13: If I remember correctly, we'll run into the issue that the main Flipper library has a no-op version, but the network interceptor add-on - and others we might include - doesn't. So we'll end up having to go the same route as we have now with Stetho, with an initializer that has different implementations for release/debug variants.

leandroBorgesFerreira: No-op includes the add ons, so I don't think we would have this sort of problem. I added a commit that shows that: `adding network add-on for flipper`.

I'm not really advocating that we use the no-op, both solutions are fine.

stream-chat-android-sample/src/main/AndroidManifest.xml
```diff
@@ -19,6 +19,8 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <activity android:name="com.facebook.flipper.android.diagnostics.FlipperDiagnosticActivity"
```
JcMinarro, line 22: Add it only for debug buildtype

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
JcMinarro: We can have a `ExtraDependecies` class that set up this configuration and have different implementations for Debug and Release. Then we don't need to do any `BuildConfig.DEBUG` check

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/App.kt
```diff
@@ -30,6 +37,20 @@ class App : Application() {
...
             notificationHandler = SampleNotificationHandler(this@App, notificationConfig)
         }.build()
+
+        configFlipper()
+    }
+
+    private fun Application.configFlipper() {
+        SoLoader.init(this, false)
+
+        if (BuildConfig.DEBUG && FlipperUtils.shouldEnableFlipper(this)) {
```
JcMinarro, line 47: `FlipperUtils` should be accessible ONLY from debug buildtype

leandroBorgesFerreira:
I'll continue this PR once I'm done with CAS-303 so I focus in the ticket.

stream-chat-android-client-sample/src/main/java/io/getstream/chat/android/client/sample/utils/StethoWebSocketsFactory.kt
```diff
@@ -1,82 +0,0 @@
-package io.getstream.chat.android.client.sample.utils
```
leandroBorgesFerreira, line 1: I'm not very sure about this class... this is not being used and I'm not trying to add new stuff in this PR, just use Flipper instead of Stheto. I decided to remove it.

samiuelson (dismissed):

stream-chat-android-client/src/debug/java/io/getstream/chat/android/client/di/NetworkFlipper.kt
```diff
@@ -0,0 +1,8 @@
+package io.getstream.chat.android.client.di
+
+import com.facebook.flipper.plugins.network.FlipperOkhttpInterceptor
+import com.facebook.flipper.plugins.network.NetworkFlipperPlugin
+
+public val networkFlipper: NetworkFlipperPlugin = NetworkFlipperPlugin()
+
+public fun flipperInterceptor(): FlipperOkhttpInterceptor = FlipperOkhttpInterceptor(networkFlipper)
```
zsmb13, line 8: Any reason for having this in a separate file?

zsmb13, line 8: Any reason for having this in a separate file?

stream-chat-android-sample/src/debug/kotlin/io/getstream/chat/sample/application/ExtrasDependenciesImpl.kt
```diff
@@ -0,0 +1,25 @@
...
+import android.app.Application
+import com.facebook.flipper.android.AndroidFlipperClient
+import com.facebook.flipper.android.utils.FlipperUtils
+import com.facebook.flipper.plugins.databases.DatabasesFlipperPlugin
+import com.facebook.flipper.plugins.inspector.DescriptorMapping
+import com.facebook.flipper.plugins.inspector.InspectorFlipperPlugin
+import com.facebook.soloader.SoLoader
+import io.getstream.chat.android.client.di.networkFlipper
+
+class ExtrasDependenciesImpl : ExtraDependencies {
```
zsmb13, line 12:
```suggestion
class ExtraDependenciesImpl : ExtraDependencies {
```

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/ExtraDependencies.kt
```diff
@@ -0,0 +1,8 @@
+package io.getstream.chat.sample.application
+
+import android.app.Application
+
+interface ExtraDependencies {
+
+    fun configFlipper(application: Application)
```
zsmb13, line 7: We could have just a single `config` function in this class, so that this interface doesn't reference `Flipper` directly

zsmb13 (commented):
Look good overall, just typo fixes / small questions

ogkuzmin (dismissed):

stream-chat-android-client/src/debug/java/io/getstream/chat/android/client/di/NetworkFlipper.kt
leandroBorgesFerreira: `networkFlipper` is being used both in ChatModule and `ExtraDependenciesImpl`. It would be weird to put everything inside ChatModule

zsmb13 (approved):

ogkuzmin (approved):

PR #732, opened 2020-10-22
Compilation improvement

leandroBorgesFerreira:
### Description

This speeds up the compilation a little. It enabled the cache of our modules do be configured in parallel (if this is possible), cache the modules that were already compilied and allows incremental kotlin

### Checklist

- [x] Reviewers added


JcMinarro (dismissed):

gradle.properties
```diff
@@ -9,6 +9,9 @@
 android.enableJetifier=true
 android.useAndroidX=true
 org.gradle.jvmargs=-Xmx1536m
+org.gradle.parallel=true
+org.gradle.caching=true
+kotlin.incremental=true
```
zsmb13, line 14: Did we check if all our annotation processors are incremental?

leandroBorgesFerreira: As far as I know if they are not, the compilation simply won't be incremental, no harm.

leandroBorgesFerreira: Although... the kotlin.incremental is already true by default, in not changing it there: https://blog.jetbrains.com/kotlin/2017/03/kotlin-1-1-1-is-out/

leandroBorgesFerreira: I'll remove the `kotlin.incremental` as it is already default behaviour, it is not adding anything.

gradle.properties
```diff
@@ -1,24 +1,5 @@
-# Project-wide Gradle stream_settings.
-# IDE (e.g. Android Studio) users:
-# Gradle stream_settings configured through the IDE *will override*
-# any stream_settings specified in this file.
-# For more details on how to configure your build environment visit
-# http://www.gradle.org/docs/current/userguide/build_environment.html
-# Specifies the JVM arguments used for the daemon process.
-# The setting is particularly useful for tweaking memory stream_settings.
```
zsmb13, line 8: Fun fact: we changed every mention of the word `settings` in the project to `stream_settings` a while back, while refactoring a string resource's name 😄

zsmb13 (approved):
I think this has everything it can have for now 👍

tschellenbach (approved):

PR #737, opened 2020-10-23
CAS-326 Fix anonymous user authentication

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-326

### Description

Anonymous authentication is not working. 

- Generated a user with random user id, as "user_details" parameter is mandatory when connecting to WS. Actually, doesn't matter what user id we send to the server, the user id will be always `!anon`
- Removed sending firebase token to the server (`addDevice` invocation)

### Checklist

- [X] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClientImpl.kt
```diff
@@ -112,13 +112,13 @@ internal class ChatClientImpl(
...
         config.isAnonymous = true
         warmUp()
-        notifications.onSetUser()
         getTokenAndConnect {
             socket.connectAnonymously()
         }
     }
 
     override fun getGuestToken(userId: String, userName: String): Call<GuestUser> {
+        config.isAnonymous = true
```
JcMinarro, line 121: Why do we need to do this side-effect here?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/SocketFactory.kt
```diff
@@ -49,13 +50,21 @@ internal class SocketFactory(
...
+            "user_id" to user.id,
+            "server_determines_connection_id" to true,
+            "X-STREAM-CLIENT" to ChatClient.instance().getVersion()
+        )
         return parser.toJson(data)
     }
+
+    /**
+     * It doesn't matter what user id we send to the server as the anonymous
+     * user id will always be "!anon"
```
JcMinarro, line 65: Have you talked with Backend team about that?
If the `userId` of an anonymous user is always `!anon`, why are we using a random UUID instead of using `!anon`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClientImpl.kt
bychkovdmitry: Maybe it's better to have `isAnonymous` true by default in ChatClientConfig. This request should be `anonymous` (`stream-auth-type` set to `anonymous` in `HeadersInterceptor`)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/SocketFactory.kt
bychkovdmitry: I've inspected JS sources, and they are sending UUID, probably to indicate that any string is fine. I will replace UUID with "!anon", will be more meaningful

bychkovdmitry: Server not accepting "!anon" as the user id due to the exclamation mark, will use "anon" then

bychkovdmitry: I also added a tiny explanation that no matter what user id we send to the server we get "!anon" in return

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClientImpl.kt
JcMinarro: What I mean is that we shouldn't mutate the status of `config.isAnonymous` inside of a method that says `getGuestToken()`, because this method is doing more stuff that it said to do.
It should be done outside

JcMinarro (approved):

PR #746, opened 2020-10-25
Tutorial

tschellenbach:
The one problematic thing in here is the logger, but that's already in the develop branch


stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -154,12 +154,11 @@ public void setUser(@NotNull User user,
         domainBuilder
                 .userPresenceEnabled()
                 .backgroundSyncEnabled(apiKey, userToken)
-                .notificationConfig(chatNotificationHandler.getConfig());
+                .notificationConfig(chatNotificationHandler.getConfig()).build();
```
tschellenbach, line 158: domainBuilder doesn't need to wait for setUser.

also the user object on the domain is automatically kept in sync, the setCurrentUser call isn't needed.

PR #747, opened 2020-10-25
CAS-355 Set guest user is not working

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-355

### Description

`getGuestToken()` server call was not working as `HeadersInterceptor` and `TokenAuthInterceptor` were trying to add authorization token to the request which was not present. Therefore, I moved the call to a separate Retrofit API class.

I also looked into the possibility to change `config.isAnonymous` variable when making `getGuestToken()` request:
```
    override fun getGuestToken(userId: String, userName: String): Call<GuestUser> {
        val temp = config.isAnonymous
        config.isAnonymous = true
        return api.getGuestUser(userId, userName).map {
            config.isAnonymous = temp
            it
        }
    }
```
This solution is not reliable as the user does not expect any side effects when calling a "getter" method. The method can be even called but not executed, that is the application will continue to function in "anonymous" mode.

### Checklist

- [X] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/HeadersInterceptor.kt
```diff
@@ -4,9 +4,9 @@ import io.getstream.chat.android.client.ChatClient
 import okhttp3.Interceptor
 import okhttp3.Response
 
-internal class HeadersInterceptor(val config: ChatClientConfig) : Interceptor {
+internal class HeadersInterceptor(private val isAnonymous: () -> Boolean) : Interceptor {
```
ogkuzmin, line 7: Why do you need lambda here instead of simple boolean flag?

bychkovdmitry: Yep, the value changes when we authenticate with "anonymous" or "normal"/"guest" users

ogkuzmin: But you provide a lambda as closure without any state (variable) I can't see that it changes when we authenticated. Or it will be done later in another PR?

bychkovdmitry: The idea in general: 
- For "getGuestToken" server call we have a provider which always returns true
- For every other method we pass `{ config.isAnonymous }` so that `isAnonymous: () -> Boolean` lambda returns different value depending on the auth state:
1) Not authenticated -> `isAnonymous()` should return false
2) Anonymous auth -> `isAnonymous()` should return true
3) Regular/guest auth -> `isAnonymous()` should return true

bychkovdmitry: The value of `config.isAnonymous` changes when we set a user:
```
    override fun setAnonymousUser(listener: InitConnectionListener?) {
        config.isAnonymous = true
         ...
    }

```

```
    override fun setUser(user: User, token: String, listener: InitConnectionListener?) {
        config.isAnonymous = false
         ...
    }

```

ogkuzmin: Now, I see. Thanks for clarification

ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -134,21 +150,23 @@ internal open class BaseChatModule(
     private fun buildApi(chatConfig: ChatClientConfig): ChatApi {
         return ChatApi(
             chatConfig.apiKey,
-            buildRetrofitApi(),
+            buildRetrofitApi(RetrofitApi::class.java, false),
+            buildRetrofitApi(RetrofitAnonymousApi::class.java, true),
```
zsmb13, line 154: We're passing in the same information - whether this is the anonymous API or not - twice here. Would be better to avoid this duplication.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -119,6 +124,17 @@ internal open class BaseChatModule(
...
 
+    private fun getAnonymousProvider(
+        config: ChatClientConfig,
+        isAnonymousApi: Boolean
+    ): () -> Boolean {
+        return if (isAnonymousApi) {
+            { true }
+        } else {
+            { config.isAnonymous }
+        }
```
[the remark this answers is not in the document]
zsmb13: Good idea, but it needs to be a lambda:

```suggestion
        return { isAnonymousApi || config.isAnonymous }
```

stream-chat-android-client/src/androidTest/java/io/getstream/chat/android/client/poc/ClientInstrumentationTests.kt
```diff
@@ -101,8 +100,8 @@ internal class ClientInstrumentationTests {
             }
             client.subscribe(connectedEventConsumer::onEvent)
         }.andThen {
-            await().atMost(5, SECONDS).until(setUserListener::onSuccessIsCalled)
-            await().atMost(5, SECONDS).until(connectedEventConsumer::isReceived)
+            await().atMost(50, SECONDS).until(setUserListener::onSuccessIsCalled)
+            await().atMost(50, SECONDS).until(connectedEventConsumer::isReceived)
```
zsmb13, line 104: This seems like a super long time to wait in this test, does it really take this long? We'd be better off leaving this test ignored if it does.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
JcMinarro: No, they are two different interfaces, one with the authenticated API endpoints and another with the anonymous API endpoints

stream-chat-android-client/src/androidTest/java/io/getstream/chat/android/client/poc/ClientInstrumentationTests.kt
JcMinarro: For now we are not running those tests in CI, they are instrumentation tests that aren't configured into our CI

JcMinarro (dismissed):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
zsmb13: I know, but we're passing in the same logical information - whether this is the authenticated API or the anonymous API being built - twice. Both parameters are telling the method the same thing.

And this allows for invalid sets of parameter values, for example calling the method like this makes no sense, but is technically possible now:

```kotlin
buildRetrofitApi(RetrofitAnonymousApi::class.java, false)
```

Having just a single parameter would avoid this.

zsmb13: Conversation resulted in https://github.com/GetStream/stream-chat-android/pull/767

JcMinarro (approved):

JcMinarro (approved):

PR #748, opened 2020-10-25
Simplify

tschellenbach:
Idea:

- remove chatclient interface
- use extension functions to access client.ux() and client.domain()
- use a handler system or similar to notify UX and offline support when setUser and disconnect are called
- move interfaces up to client level as needed


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -137,6 +141,24 @@ public fun ChatError.isPermanent(): Boolean {
...
 
+public fun ChatClient.domain(): ChatDomain {
+
+    // The builder/config logic happens at the ChatClient level
+    val domainBuilder = ChatDomain.Builder(appContext, this).setConfig(offlineConfig)
+
+    // We use listeners to tie into the ChatClient's setUser and disconnect flow
+    //
+    // // hookup listeners for pre set User and disconnect()
+    // setPreUserConnectListener {
```
ogkuzmin, line 152: We could write it more polymorphic. Provide users the possibility to enhance our SDKs by their own extensions.
We could create such an interface:
``` kotlin
interface ChatExtensionsModule /* whatever name */ {
  /* add here some lifecycle methods */
fun onSetUserPendind(user: User)
fun onSetUserCompleted(user: User)
fun onSetUserFailed(user: User, exception: Throwable)
fun onConnected()
fun onDisconnected()
...
}
```

We could even split such interface to several (i.e. `ChatConnectionListener`, `UserAuthorizationListener` etc) and make ChatExtensionsModule as a Facade.
Then add to ChatClient method `addExtension(moduleExtension: ChatModuleExtension)`. Then `ChatClient` will have a list of such extensions that it will notify about an event of the lifecycle.

ogkuzmin:
Anyway for user it seems as a goog change, easier to user ad configure. Also it's. important to developers cut down misleading while using all these ChatClient, ChatDomain, ChatUx

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/ClientExtensions.kt
```diff
@@ -137,6 +141,24 @@ public fun ChatError.isPermanent(): Boolean {
     return isPermanent
 }
 
+public fun ChatClient.domain(): ChatDomain {
```
zsmb13, line 144: Extensions are neat, but let's consider what using these will look like for Java clients

tschellenbach:
Great explanation of the limitations of kotlin extensions and Java interop: https://stackoverflow.com/a/28364983

PR #750, opened 2020-10-26
Clean up ViewModelFactory implementations

zsmb13:
### Description

- Makes factories implement `ViewModelProvider.Factory` instead of extending `ViewModelProvider.NewInstanceFactory()`. This prevents clients from misusing the factory, and creating instances of their own ViewModel types through it.
- Add checks for the class requested, fail with nice error message if incorrect class is passed into the factories.

### Checklist

- [x] Reviewers added


stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/factory/StreamViewModelFactory.kt
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A ViewModel factory for MessageListViewModel, ChannelHeaderViewModel and MessageInputViewModel
+ *
+ * @param cid: the channel id in the format messaging:123
+ *
+ * @see MessageListViewModel
+ * @see ChannelHeaderViewModel
+ * @see MessageInputViewModel
+ */
+public class StreamViewModelFactory(private val cid: String) : ViewModelProvider.Factory {
```
zsmb13, line 18: Not sure about the naming of this class yet, but we can't have `ChannelsViewModelFactory` and `ChannelViewModelFactory` in the API, way too confusing. Should review this with the things discussed [here](https://getstream.slack.com/archives/G016YAWD423/p1603467611056200).

bychkovdmitry: Maybe `ChannelListViewModelFactory` and `MessageListViewModelFactory`?

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelsViewModel.kt
```diff
@@ -40,7 +40,7 @@ public interface ChannelsViewModel {
     }
 }
 
-public class ChannelsViewModelImpl(
+internal class ChannelsViewModelImpl(
```
adasiewiczr, line 43: 🤔 Do we need `ChannelsViewModel` interface? Maybe we can easily remove that?

zsmb13: I think it's a nice, clean way of defining the contract of the ViewModel, even if we won't have multiple implementations of it. Kind of like an MVP contract interface.

I was actually thinking about introducing similiar interfaces for the other ViewModels 😅

adasiewiczr: I think it depends if want to "bind" our view with a specific contract or we rather give users our view model implementation together with binding and if they want to have some custom - they need to write the whole thing by themselves.
I'm not a big fan of creating interfaces that won't be reused and only defines contract because usually, you end up with exactly one implementation which is called ʻInterfaceIMPL` 🙈 
BTW currently we are using `ChannelsViewModelImpl` inside `ChannelsFragment` so we can't make it internal

zsmb13: Oops, I don't know how I haven't looked at the usages, but I just realized this whole change is unfeasible. We need the type passed into the factory to be a `ViewModel` of some sort, and an interface can not do that. Time to remove the interface 😄

adasiewiczr (approved):

JcMinarro (dismissed):

JcMinarro (approved):
LGTM

PR #755, opened 2020-10-27
Cas 356/fix channels sorting

ogkuzmin:
[Jira CAS-356](https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=49&projectKey=CAS&modal=detail&selectedIssue=CAS-356)

### Description
Main goal to speed up Channels sorting. The main consuming part was function `Collection<Channel>.applyPagination()` because of the `comparator` extension field that used reflection under the hood. To avoid reflection we can specify which field of model we want to use for sorting reasons, or provide backend field.
Also I refactored QuerySort data base workflow, because it used gson serialisation/deserialization with reflection to work with DB.
Summary:

- Changed QuerySort api. Now you have to specify model class for this
- Changed comparator calculation and inner implementation of QuerySort.
- Changed DB entities. mechanisms to store and extract it from DB

### Checklist

- [x] Reviewers added


stream-chat-android-client-sample/src/main/java/io/getstream/chat/android/client/sample/DocumentationSamples.kt
```diff
@@ -479,7 +480,7 @@ fun queryChannels() {
         .put("type", "messaging")
     val offset = 0
     val limit = 10
-    val sort = QuerySort().desc("last_message_at")
+    val sort = QuerySort.desc(Channel::lastMessageAt)
```
tschellenbach, line 483: ok, but how do we generate the API call to the backend using this setup?

ogkuzmin: As previously. Nothing changed. There is toDto method for that purpose

tschellenbach: toDto?

ogkuzmin: Yes, `QuerySort::toDto`

tschellenbach:
1. You would indeed expect the reflection approach to be slower than specifying the field. What's the performance improvement you saw with this change? 

2. Can you explain the need for the extra entity?
image

ogkuzmin:
@tschellenbach Thanks for questions.

1. Before I decided to make this refactoring, I had measured this part of a code flow and compared to other things. I saw that the Fragment with Channels spend the biggest amount of time locally for sorting channels list that had been stored locally. Then I saw that the reason was in the calculations of comparator using QuerySort. So I decided to rewrite it. 
 I did two iterations:
    1. Removed reflection in calculation of comparator by storing KPropery in QuerySort. Sorting was increased by that, but I noticed that selection of query entity from DB had grown unexpectedly. I looked at it, and inside it used GSON to serialize/deserialize list of that properties to DB.
   2. Secondly I looked at Room DB manual with best practises and refactored QueryEntity to store in DB using native mechanism of Room.
By the way, results of my experiments can be found [here](https://docs.google.com/spreadsheets/d/1wD_Ij1a6pD1nv6p-6ijGwJhqDh-igQgPIbz3HLr2GE0/edit?usp=sharing).
2. Since our QueryEntity contains a list of sorting params its' relation is one-to-many. In order to implement one-to-many using room you need to create extra entity (that represent a row entity in that list `ChannelsSortInnerEntity`) and non-entity class that describes a relation (`QueryChannelsWithSort`). And when you describes your queries in DAO Room will generate all required inner relations inside it using transactions. To get more details in this topic you can read [this article](https://developer.android.com/training/data-storage/room/relationships) about Room DB

tschellenbach:
1. Are you sorting using SQL? I didn't see how the code handles that
2. When you retrieve a query from the API you need to remove all old ChannelsSortInnerEntity and add the new ones, where is that implemented?

This seems promising, I'll ping you to setup a call for reviewing this after the tutorial goes live. (so probably Tuesday if that works for you.)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/QueryChannelsEntity.kt
```diff
@@ -1,26 +1,27 @@
...
-    var updatedAt: Date? = null
+@Entity(tableName = "channel_sort_inner_entity")
+internal data class ChannelSortInnerEntity(val name: String, val direction: Int, val queryId: String) {
+    @PrimaryKey
+    var id: String = name.hashCode().toString() + direction.hashCode().toString() + queryId
 }
+
+internal data class QueryChannelsWithSorts(
+    @Embedded val query: QueryChannelsEntity,
+    @Relation(parentColumn = "id", entityColumn = "queryId", entity = ChannelSortInnerEntity::class)
```
tschellenbach, line 25: Does room use a second query or a join for the @relation syntax?

tschellenbach: Why not define this relation on the QueryChannelsEntity class?

tschellenbach: You're saying this part is a lot faster than the old approach right? Any idea why?

ogkuzmin: Using relation you mark to room that it should generate two queries when fetching entity. You must create an additional class with relation and it must be not an entity but a simple class with @Relation field. It's Room requirement. You're able free not using this mechanism by create two simple entities and fetch them via two DAO and two queries and build them manually. This Room mechanism does the same but automatically generates all inner routine stuff.

ogkuzmin: I didn't measure it before. But previously QueryChannelsEntity stored QuerySort as json string using serialization via GSON. It think it's not so fast task, SQL queries to DB work should work faster.

ogkuzmin:
About sorting using SQL and cleaning DB entities. I didn't change anything in the lifecycle of entities or how we sort channels. I just refactored how we get a comparator for sorting and how we stored query entities in DB.
So we sort locally using a comparator, and we send sorting order in request to BE API. Also, it seems for cleaning we have to add new logic which absent right now.
Let speak and discuss it via a call

stream-chat-android-client-sample/src/main/java/io/getstream/chat/android/client/sample/DocumentationSamples.kt
```diff
@@ -214,7 +215,7 @@ fun queryUsers2() {
     val filter = Filters.`in`("id", listOf("jessica"))
     val offset = 0
     val limit = 10
-    val sort = QuerySort().desc("last_active")
+    val sort = QuerySort.desc(User::lastActive)
```
tschellenbach, line 218: So the API allows you to sort on custom fields and the field names are imploded. So **extraData.something** is specified as **something**.

Something to fix in a future release/when people ask for it.

tschellenbach (approved):
Solid improvement: 

- Not using reflection is cleaner
- Using the standard room approach here is cleaner
- Performance improved by ~200ms due to not using reflection (its weird how db got slower and than got faster again, it would be good to understand that better as well)

Further improvements for future iterations:
- IO can be improved on how we delete and insert the query channel results for additional performance improvement (has a big impact on how many deletes/inserts are done if only 1 or 2 channels are added/removed in the list)
- Ordering could be handled in the DB in some cases (which means you only need to retrieve the limit of items, instead of all. if you have a lot of channels this will impact performance)

PR #756, opened 2020-10-27
Remove Kotlin synthetics

zsmb13:
### Description

Starting in Kotlin 1.4.20, the synthetic imports provided by the Kotlin Android Extensions plugin are deprecated in favour of ViewBinding ([see here](https://youtrack.jetbrains.com/issue/KT-42121)). This PR removes usages of synthetics from all modules, and replaces them with ViewBinding.

Note: the ViewBinding code added is a bit verbose, as it simply follows the [official samples for implementing ViewBinding in Fragments](https://developer.android.com/topic/libraries/view-binding#fragments). It could be made neater at the call site by [extracting the common code into a delegate](https://medium.com/@Zhuinden/simple-one-liner-viewbinding-in-fragments-and-activities-with-kotlin-961430c6c07c), but I opted not to do this, because it would introduce complexity into sample code.

### Checklist

- [x] Reviewers added


stream-chat-android-offline/build.gradle
```diff
@@ -27,10 +26,7 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
-    
-    buildFeatures {
-        dataBinding true
-    }
```
zsmb13, line 33: This was still enabled, but not used anywhere

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/dialog/MessageMoreActionDialog.kt
```diff
@@ -49,47 +50,48 @@ internal class MessageMoreActionDialog(
     }
 
     private fun setupMessageActions() {
-        setContentView(R.layout.stream_dialog_message_moreaction)
+        val binding = StreamDialogMessageMoreactionBinding.inflate(context.inflater)
+        setContentView(binding.root)
```
zsmb13, line 54: Using `setContentView(View)` instead of `setContentView(@LayoutRes Int)` somehow changed how the layout was positioned. Previously, it was positioned at the bottom the screen correctly, but this change moved some of its contents to the top of the screen. The changes in the layout XML fix this.

adasiewiczr (approved):

PR #758, opened 2020-10-27
Fix multiple socket connections

bychkovdmitry:
### Description

There were multiple problems with socket initialization. We ended up with 2 or 3 socket connections at the same time, multile health monitors running at the same time (even when user is logged out), leaking sockets that were never closed.

1) Added `warmup` request to anonymous API as the call was executed in "authenticated" mode. To make matters worse, it was executed before token was loaded so it tried to load token and make yet another connection to the WS

2) Sometime after the socket was closed bu the client (e.g. during logout), `EventsParser.onFialure` was triggered and health monitor was restarted (and eventually the health monitor reconnected to the already closed socket). For now, I'm just ignoring these phantom errors when socket is closed intentionally by the client.

3) When we authenticate we register StreamLifecycleObserver, whose `onResume` callback was invoked immediately and forced the creation of yet another socket conneciton.

4) Also, I'm removing now StreamLifecycleObserver from lifecycle observers on disconnect, so that the observer is not added again after we set another user. 

### Checklist

- [X] Reviewers added


stream-chat-android/src/main/java/com/getstream/sdk/chat/StreamLifecycleObserver.java
```diff
@@ -8,14 +8,32 @@
 public class StreamLifecycleObserver implements LifecycleObserver {
```
JcMinarro, line 8: It is an internall class. We can convert it to Kotlin and mark it as internal

JcMinarro (changes requested):
The `warmup` shouldn't create a new WS connection if the token is not loaded. Reload the token doesn't interfiere on the WS Connection. On the other hand, it should be done on the same OkHttpClient to "warm up" the connection before the rest of calls are done. So WarmUp refactor need to be reverted.
The rest of the changes look good ;)

stream-chat-android/src/main/java/com/getstream/sdk/chat/StreamLifecycleObserver.java
bychkovdmitry: There's a small probability (1 out of 5 times on my emulator) that `TokenManager.loadSync()` from `TokenAthInterceptor` is executed when there is still a listener from `TokenManagerImpl.loadSync()` from `setUser` call. In that case connect() will be called twice

bychkovdmitry: Going to add a small delay before helth monitor is reset as :

```
    override fun connect(
        endpoint: String,
        apiKey: String,
        user: User?
    ) {
        state = State.Disconnected // triggers reconnect runnable in HealthMonitor immediately
       ...
        setupSocket()
    }
```

```
    private val reconnect = Runnable {
        if (needToReconnect()) {
            healthCallback.reconnect()
        }
    }
```

bychkovdmitry: - Refactored StreamLifecycleObserver to kotlin and made the class internal
- Added delay before health monitor start to prevent spontaneous reconnects
- Synchronized onTokenLoaded(). The class is still not totally thread-safe.
- Moved warmup() back to `RetrofitApi`. Though the call does not require auth token, we are not confident that different Retrofit APIs reuse the same TCP connection. 

I tested the code by doing login and logut 20 times and didn't notice any leaking sockets / health monitors.

PR #759, opened 2020-10-27
Small network config and parsing fixes

zsmb13:
### Description

A collection of fixes and improvements in the network configuration / parsing code. **Probably easier to review commit by commit.**

- Clean up BaseChatModule class, make some things final, add explicit typing
- Remove separate timeout parameters, as we always pass in the same value for all of them
- Move the configuration of the call adapter of our Retrofit instance to `BaseChatModule`, instead of having it performed by the concrete `ChatParser` implementation internally
- Reuse a single base `OkHttpClient` instance, so that threadpools and connection pools can be shared across instances (see the [docs of the class](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/) for more details)
- Rework the synchronization around date parsing.
    - The core issue with the `DateAdapter` was that `SimpleDateFormat` is not thread safe. Previously, [we added synchronization in a bunch of places](https://github.com/GetStream/stream-chat-android/commit/672db623e5ab39d530df36234dea81f699f5b468) to account for this, but this was done more than it was necessary, and not always in the correct place. Using thread-local `SimpleDateFormat` instances solves the core issue instead. The new `ThreadLocalDelegate ` class provides a convenient way to introduce thread-local instances.

### Checklist

- [x] Reviewers added


adasiewiczr (dismissed):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ThreadLocalDelegate.kt
```diff
@@ -0,0 +1,19 @@
+package io.getstream.chat.android.client.utils
+
+import kotlin.properties.ReadOnlyProperty
+import kotlin.reflect.KProperty
+
+internal class ThreadLocalDelegate<T : Any>(
```
bychkovdmitry, line 6: Cool. DefaultDateTypeAdapter from Gson uses synchronization for thread safety, but this approach definitely better.

zsmb13: I never thought to check it out, but that's actually a great idea. Looking at it now, they synchronize on the `DateFormat` instance(s), which is better than the method level synchronization that we had.

Still, thread local makes this even cleaner, so I think it's worth it.

bychkovdmitry (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ThreadLocalDelegate.kt
bychkovdmitry: Yep, I like this solution

bychkovdmitry (approved):

PR #761, opened 2020-10-27
Allow custom loading and empty state views

samiuelson:
https://stream-io.atlassian.net/browse/CAS-345

### Description

* Add empty state views to channel list view and message list view components
* Allow setting custom empty state views
* Add loading view to message list view
* Allow setting custom loading view

### Checklist

- [x] Reviewers added


stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/channels/ChannelsFragment.kt
```diff
@@ -15,6 +15,7 @@ import com.getstream.sdk.chat.viewmodel.channels.bindView
 import com.getstream.sdk.chat.viewmodel.factory.ChannelsViewModelFactory
 import io.getstream.chat.sample.R
 import io.getstream.chat.sample.common.navigateSafely
+import kotlinx.android.synthetic.main.fragment_channel.*
```
zsmb13, line 18: This looks really suspicious, with no other changes in the file.

But also: [pls no](https://github.com/GetStream/stream-chat-android/pull/756) 😅  (if it's actually needed, maybe rebase after that PR is merged)

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -52,6 +55,9 @@ import io.getstream.chat.android.client.models.User
  * - Customizing the click and longCLick (via the adapter)
  * - The list_item_message template to use (perhaps, multiple ones...?)
  */
+
+private val LOADING_VIEW_ID = R.id.message_list_loading_view_id
```
zsmb13, line 59: This clearly belongs to the class itself (as opposed to the file), I would prefer we place it within the class. Makes the class level comment's position weird as well.

stream-chat-android/src/main/res/layout/stream_message_list_view.xml
```diff
@@ -58,4 +58,38 @@
...
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:visibility="gone"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintLeft_toLeftOf="parent"
+        app:layout_constraintRight_toRightOf="parent"
+        app:layout_constraintTop_toTopOf="parent">
+
+        <ProgressBar
+            android:id="@+id/defaultProgressbar"
```
zsmb13, line 72: It would be more consistent to name this `defaultLoadingView`, to match the pattern with the IDs of the empty views.

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModelBinding.kt
```diff
@@ -26,8 +26,19 @@ public fun MessageListViewModel.bindView(view: MessageListView, lifecycleOwner:
     view.setOnMessageRetryHandler { onEvent(RetryMessage(it)) }
 
     state.observe(lifecycleOwner) { state ->
-        if (state is MessageListViewModel.State.Result) {
-            view.displayNewMessage(state.messageListItem)
+        when (state) {
+            is MessageListViewModel.State.Loading -> {
+                view.showLoadingView()
+            }
```
zsmb13, line 32: We should hide the empty state view on this branch. Even if right now we never get into a `Loading` state after `Result`, it's a cleaner implementation of binding state to UI if we get the binding right from any previous state.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -389,6 +420,34 @@ public class MessageListView : ConstraintLayout {
         setMessageListItemAdapter(adapter)
     }
 
+    public fun setLoadingView(view: View, layoutParams: FrameLayout.LayoutParams = defaultChildLayoutParams()) {
+        loadingViewContainer.removeView(loadingView)
+        loadingView = view.apply { id = LOADING_VIEW_ID }
```
zsmb13, line 425: What's the use of setting these IDs on the Views that are set?

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -389,6 +420,34 @@ public class MessageListView : ConstraintLayout {
         setMessageListItemAdapter(adapter)
     }
 
+    public fun setLoadingView(view: View, layoutParams: FrameLayout.LayoutParams = defaultChildLayoutParams()) {
```
zsmb13, line 423: Could use documentation to describe that the View passed as a parameter will be added to the View hierarchy by us, and we'll manage it. Otherwise, someone might add `MessageListView` and their own desired loading View to a layout, and then call this function, which I assume would produce an exception, since the View is already added somewhere.

stream-chat-android/src/main/res/values/attrs.xml
```diff
@@ -107,9 +107,13 @@
             <flag name="bold" value="1"/>
             <flag name="italic" value="2"/>
         </attr>
-        <!--Custom ChannelListLayout-->
         <attr name="streamChannelPreviewLayout" format="reference"/>
     </declare-styleable>
+    <declare-styleable name="ChannelsView">
+    <!--Custom ChannelListLayout-->
+        <!-- Text of the default empty state label -->
```
zsmb13, line 114: I think we don't need the word `default` in these descriptions, as it's for the customizable attribute

```suggestion
        <!-- Text of the empty state label -->
```

zsmb13 (commented):
I occasionally got a "No messages" label when opening a channel before the messages in it loaded, though I couldn't really reproduce it. More of a state handling issue in the ViewModel than an issue of the UI implementation here.

The implementation here looks nice, I left a bunch of tiny comments.

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/channels/ChannelsFragment.kt
samiuelson: Ups, some conflict reloving left-over 🤦🏼

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -71,6 +77,10 @@ public class MessageListView : ConstraintLayout {
 
     private lateinit var adapter: MessageListItemAdapter
     private lateinit var layoutManager: LinearLayoutManager
+    private lateinit var loadingView: View
```
tschellenbach, line 80: typically there is also a loading state for loading more. IE scrolling up, which tends to have a different loading UI than the first page load...

tschellenbach:
Really nice how this standardizes things between channel list and message list

tschellenbach:
is this good to go, do we want to get this live?

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
samiuelson: That's right, I'd like to handle "load more" in another PR to merge this one sooner

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
samiuelson: It seems they are required only on `RecyclerView`. Removing them from empty state and loading views.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
samiuelson: Hm, maybe kind of View factory or producing function would be better instead of passing the `View` reference here to avoid memory leaks also.

samiuelson: Added docs to explain this to the user

leandroBorgesFerreira (dismissed):

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -389,6 +420,42 @@ public class MessageListView : ConstraintLayout {
         setMessageListItemAdapter(adapter)
     }
 
+    /**
+     * @param view will be added to the view hierarchy of [ChannelsView] and managed by it.
+     * The view should not be added to another [ViewGroup] instance elsewhere.
+     * @param layoutParams defines how the view will be situated inside it's container ViewGroup.
+     */
+    public fun setLoadingView(view: View, layoutParams: FrameLayout.LayoutParams = defaultChildLayoutParams()) {
```
leandroBorgesFerreira, line 428: Why not add a way for the user to pass their own view using XML instead of passing the view programmatically? I think it is easier to create the layout using XML than using code.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -448,6 +515,12 @@ public class MessageListView : ConstraintLayout {
         layoutManager.scrollToPosition(adapter.itemCount - 1)
     }
 
+    private fun defaultChildLayoutParams() = FrameLayout.LayoutParams(
```
carterhudson, line 518: Could this be a `val` or live in a companion?

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModelBinding.kt
```diff
@@ -26,8 +26,20 @@ public fun MessageListViewModel.bindView(view: MessageListView, lifecycleOwner:
...
-        if (state is MessageListViewModel.State.Result) {
-            view.displayNewMessage(state.messageListItem)
+        when (state) {
+            is MessageListViewModel.State.Loading -> {
+                view.hideEmptyStateView()
+                view.showLoadingView()
+            }
+            is MessageListViewModel.State.Result -> {
+                if (state.messageListItem.items.isEmpty()) {
+                    view.showEmptyStateView()
```
carterhudson, line 36: Do we need two methods for what amounts to boolean logic? If we did `fun showEmptyStateView(show: Boolean = true) { ... }`, this could become any of 
- `view.showEmptyStateView(state.messageListItem.items.isEmpty())`
- ` state.messageListItem.items.isEmpty().let { itemsEmpty -> view.showEmptyStateView(itemsEmpty) }`
- ` state.messageListItem.items.isEmpty().let(view::showEmptyStateView)` 
but could also still be called as `view.showEmptyStateView`.

Just a thought that crossed my mind /shrug.

carterhudson (changes requested):

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModelBinding.kt
samiuelson: IMO, `showEmptyState(true)` and `showEmtpyState(false)` can be confusing.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
samiuelson: It would be a nice feature. I'd add this in the next iteration.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/channels/ChannelsView.kt
```diff
@@ -36,17 +35,40 @@ public class ChannelsView @JvmOverloads constructor(
...
+                    }
+                }
+            }
+        }
     }
 
+    /**
+     * @param view will be added to the view hierarchy of [ChannelsView] and managed by it.
+     * The view should not be added to another [ViewGroup] instance elsewhere.
+     * @param layoutParams defines how the view will be situated inside it's container ViewGroup.
```
zsmb13, line 56:
```suggestion
     * @param layoutParams defines how the view will be situated inside its container ViewGroup.
```

sorry 😅 

Good docs though, thank you for adding them!

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/channels/ChannelsView.kt
```diff
@@ -36,17 +35,40 @@ public class ChannelsView @JvmOverloads constructor(
...
         removeView(this.emptyStateView)
-        this.emptyStateView = view.apply { id = EMPTY_STATE_VIEW_ID }
+        this.emptyStateView = view
         addView(emptyStateView, layoutParams)
     }
 
+    /**
+     * @param view will be added to the view hierarchy of [ChannelsView] and managed by it.
+     * The view should not be added to another [ViewGroup] instance elsewhere.
+     * @param layoutParams defines how the view will be situated inside it's container ViewGroup.
```
zsmb13, line 67:
```suggestion
     * @param layoutParams defines how the view will be situated inside its container ViewGroup.
```

stream-chat-android/src/main/res/values/attrs.xml
```diff
@@ -367,6 +372,12 @@
 
         <attr name="streamUserNameShow" format="boolean"/>
         <attr name="streamMessageDateShow" format="boolean"/>
+
+        <!-- Text of the default empty state label -->
```
zsmb13, line 376:
```suggestion
        <!-- Text of the empty state label -->
```

stream-chat-android/src/main/res/values/attrs.xml
```diff
@@ -367,6 +372,12 @@
...
+        <attr name="streamMessagesEmptyStateLabelText" format="string"/>
+
+        <!-- Reference to the layout consisting of loading view -->
+        <attr name="streamMessagesLoadingView" format="reference"/>
```
zsmb13, line 380: Where is this used?

stream-chat-android/src/main/res/values/attrs.xml
```diff
@@ -113,6 +113,7 @@
     <!--Custom ChannelListLayout-->
         <!-- Text of the empty state label -->
         <attr name="streamChannelsEmptyStateLabelText" format="string"/>
+        <attr name="streamChannelsLoadingView" format="reference"/>
```
zsmb13, line 116: Where is this used?

zsmb13 (commented):
Really, really minor comments now

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/channels/ChannelsView.kt
samiuelson: My bad 👍🏼

stream-chat-android/src/main/res/values/attrs.xml
samiuelson: Removed 👍🏼

zsmb13 (approved):

carterhudson (approved):
lgtm

adasiewiczr (approved):
LGTM! Could you also update MessageListView docs: https://getstream.io/chat/docs/message_list_view/?language=kotlin ? It  would be great to mention those methods

PR #765, opened 2020-10-27
Chat initialization - Option 3 - CAS-351

tschellenbach:
Implementation of option 3 listed here while maintaining backwards compatibility with the current system:
https://stream-io.atlassian.net/browse/CAS-351

Tutorial PR is here: 
https://github.com/GetStream/android-chat-tutorial/pull/24/files

Other fixes:

- Removes the ChatClient interface in favor of the ChatClient implementation. Using extra interfaces complicates the navigation of code and makes it impossible to use kotlin extension functions
- Fix ChatUX to allow you to actually customize things
- Move app lifecycle disconnect behaviour to the low level client

### Initialization

**Old syntax**

```
Chat.Builder("b67pax5b2wdq", applicationContext).apply { offlineEnabled=true }.build()
Chat.setUser()
// this seems nice, but then later on in the code we use
ChatDomain.instance()
// and
ChatClient.instance()
// without you having an idea where those came from
```

**New Syntax**

```
val client = ChatClient.Builder("b67pax5b2wdq", applicationContext).logLevel(ChatLogLevel.ALL).build()
val domain = ChatDomain.Builder(client).build()
val ux = ChatUX.Builder(client, domain).build()
client.setUser()
 ```

### Changing the level of integration

A very common scenario is starting out with UX enabled and then going for the client, or client + offline route.

**Old syntax**

all 3 libs:

```
Chat.Builder("b67pax5b2wdq", applicationContext).apply { offlineEnabled=true }.build()
Chat.setUser()
```

2 libs (fun fact, our own example app got it wrong in the past)
```
        ChatClient.Builder chatBuilder = new ChatClient.Builder(this.apiKey, context)
                .notifications(chatNotificationHandler)
                .logLevel(chatLogLevel);

final ChatDomain.Builder domainBuilder = new ChatDomain.Builder(context, client, user);
domainBuilder
                .userPresenceEnabled()

// switch it when the user changes

        client.setUser(user, userToken, new InitConnectionListener() {
            @Override
            public void onSuccess(@NotNull ConnectionData data) {
                callbacks.onSuccess(data);
            }

            @Override
            public void onError(@NotNull ChatError error) {
                callbacks.onError(error);
            }
        });
```

**New Syntax**

```
val client = ChatClient.Builder("b67pax5b2wdq", applicationContext).logLevel(ChatLogLevel.ALL).build()
val domain = ChatDomain.Builder(client).build()
// remove 1 or 2 lines, everything still works as intended val ux = ChatUX.Builder(client, domain).build()
client.setUser()
 ```

### Conclusion

- The old syntax was confusing and frustrating for people that are changing their integration levels. (if you only use the UI it's fine, but if not you'll get confused)
- The different ways to initialize things also made it harder to document things in a consistent manner
- The new approach, makes it as simple as commenting a line and removing a dependency
- Note that this PR is not removing the Chat class, it's marked deprecated, but it still works so people can upgrade once everything is fully polished on our end in terms of docs.


stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUX.kt
```diff
@@ -0,0 +1,104 @@
...
+ * @param strings allows you to customize strings
+ * @param navigationHandler navigation handler for customizing things such as the media browsing experience
+ * @param markdown interface to to customize the markdown parsing behaviour, useful if you want to use more markdown modules
+ * @param urlSigner url signing logic, enables you to add authorization tokens for images, video etc
+ *
+ * @see ChatMarkdown
+ * @see UrlSigner
+ * @see ChatStrings
+ * @see ChatFonts
+ */
```
tschellenbach, line 29: this is the most user facing part of the change (hard to change things that are user/developer facing, easy to change internal things.) so lets get it right

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUX.kt
```diff
@@ -0,0 +1,104 @@
...
+    public val navigationHandler: ChatNavigationHandler? = null,
+    public val markdown: ChatMarkdown,
+    public val urlSigner: UrlSigner
+) {
+    public val version: String
+        get() = BuildConfig.BUILD_TYPE + ":" + BuildConfig.VERSION_NAME
+
+    public val navigator: ChatNavigator = ChatNavigatorImpl()
+
+    // TODO: would be nice to support configuring the loading icon in 1 place
```
tschellenbach, line 44: Would be nice to be able to set a default loading view here that's picked up by the other views. What do you think @samiuelson ?

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUX.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.getstream.sdk.chat.style.ChatFonts
+import com.getstream.sdk.chat.style.ChatFontsImpl
+import com.getstream.sdk.chat.style.ChatStyle
+import com.getstream.sdk.chat.utils.strings.ChatStrings
+import com.getstream.sdk.chat.utils.strings.ChatStringsImpl
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.livedata.ChatDomain
+
+/**
+ * ChatUX handles any configuration for the Chat UI elements. It replaces the older Chat class.
```
zsmb13, line 15: >ChatUX handles any configuration for the Chat UI elements.

So why isn't it called `ChatUI`? 😄

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -357,7 +412,7 @@ internal class ChatDomainImpl private constructor(
     /** stores the mapping from cid to channelRepository */
     var activeChannelMapImpl: ConcurrentHashMap<String, ChannelControllerImpl> = ConcurrentHashMap()
 
-    private val activeQueryMapImpl: ConcurrentHashMap<String, QueryChannelsControllerImpl> = ConcurrentHashMap()
+    private var activeQueryMapImpl: ConcurrentHashMap<String, QueryChannelsControllerImpl> = ConcurrentHashMap()
```
JcMinarro, line 415: Why it has been converted to mutable property?
We shouldn't ned to create new instance of this, because it has a `clear()` method

JcMinarro (changes requested):

JcMinarro:
I have extracted the `ChatClient` refactor into this PR #771  to simplify this one

tschellenbach:
The 2 key things to review are:
- Do you like the naming etc on ChatUX
- Do you agree that option 3 is the best way to initialize (see Jira)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatClientConfig.kt
```diff
@@ -1,24 +1,17 @@
...
-    var cdnHttpUrl: String,
-    var wssUrl: String,
-    var baseTimeout: Long,
-    var cdnTimeout: Long,
-    val warmUp: Boolean,
-    val loggerConfig: ChatLogger.Config,
-    val notificationsHandler: ChatNotificationHandler,
-    val fileUploader: FileUploader? = null,
-    val tokenManager: TokenManager = TokenManagerImpl()
+public class ChatClientConfig(
```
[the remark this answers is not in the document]
tschellenbach: because the config of client is used by the other 2 libs

tschellenbach:
- JC: notify the listener, if it's set after pre user already happened
-

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUX.kt
samiuelson: We are discussing it in the team on how to refactor `BaseStyle` code. We should have the possibility to customize global/common appearance and behaviour of our UI components. 

However, I'm sceptical about the idea of providing a global loading view setting. That's for a couple of reasons.
1. It would need to be passed as a factory or producer function instead of `View` instance to avoid a memory leak. That would overcomplicate the API.
2. I'm not convinced I'd need something like this as a developer. Loading indicators seems to be a thing which even if kept consistent across different places can be of a different size, variant, etc.

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUX.kt
tschellenbach: Do we all prefer UI instead of UX?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -89,16 +91,19 @@ public interface ChatDomain {
     public data class Builder(
         private var appContext: Context,
         private var client: ChatClient,
-        private var user: User
+        private var user: User? = null
```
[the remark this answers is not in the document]
tschellenbach: it's there for backwards compatibility, we can clean this up when we fully drop the Chat class, sometime in the future

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -22,290 +33,661 @@ import io.getstream.chat.android.client.models.Flag
...
-
-    public fun getCurrentUser(): User?
-
-    public fun channel(cid: String): ChannelController
+public class ChatClient internal constructor(
+    public val config: ChatClientConfig,
+    private val api: ChatApi,
+    private val socket: ChatSocket,
+    private val notifications: ChatNotifications,
+    public var appContext: Context,
```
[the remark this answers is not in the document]
tschellenbach: Any other solutions to this?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -203,24 +244,38 @@ internal class ChatDomainImpl private constructor(
...
+    init {
+        logger.logI("Initializing ChatDomain with version " + getVersion())
+
         useCases = UseCaseHelper(this)
 
-        // verify that you're not connecting 2 different users
-        if (client.getCurrentUser() != null && client.getCurrentUser()?.id != currentUser.id) {
-            throw IllegalArgumentException("client.getCurrentUser() returns ${client.getCurrentUser()} which is not equal to the user passed to this repo ${currentUser.id} ")
+        // if the user is already defined, just call setUser ourselves
+        val current = u ?: client.getCurrentUser()
```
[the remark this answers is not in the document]
tschellenbach: backwards compatibility

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -79,22 +86,29 @@ internal val gson = Gson()
  * chatDomain.errorEvents events for errors that happen while interacting with the chat
  *
  */
-internal class ChatDomainImpl private constructor(
+internal class ChatDomainImpl internal constructor(
     internal var client: ChatClient,
-    override var currentUser: User,
+    internal var u: User? = null,
```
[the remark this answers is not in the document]
tschellenbach: agreed, i rushed a bit to avoid the current merge conflict situation.

PR #768, opened 2020-10-28
Fix crash caused by relying on UninitializedPropertyAccessException

zsmb13:
### Description

https://github.com/GetStream/stream-chat-android/pull/764 changed `lateinit` `instance` properties to nullable. Turns out we were relying on `lateinit` sometimes throwing an `UninitializedPropertyAccessException`.

This PR adds a new method for checking whether the `ChatClient` or `ChatDomain` is already set. (This API could be marked internal or opt-in later, for now, it's probably fine for it to be `public`.)

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -202,5 +202,8 @@ public interface ChatDomain {
         @JvmStatic
         public fun instance(): ChatDomain = instance
             ?: throw IllegalStateException("ChatDomain.Builder::build() must be called before obtaining ChatDomain instance")
+
+        public val isInitialized: Boolean
+            get() = instance != null
     }
 }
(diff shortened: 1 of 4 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -202,5 +202,8 @@ public interface ChatDomain {
         @JvmStatic
         public fun instance(): ChatDomain = instance
             ?: throw IllegalStateException("ChatDomain.Builder::build() must be called before obtaining ChatDomain instance")
+
+        public val isInitialized: Boolean
```
ogkuzmin, line 206: What's about to make it JvmStatic?

ogkuzmin (approved):

bychkovdmitry (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
zsmb13: We have a `isInitialized` instance method already in `ChatDomain` that it would clash with. Plus I don't particularly want anyone to use this API, so it's not worth making friendly for Java clients 😅

PR #769, opened 2020-10-28
[CAS-350] Add GetChannelController use case

adasiewiczr:
Adding a separate use case for getting `ChannelController`. We shouldn't use `watchChannel` use case for that because it triggers unneeded calls to API and offline storage.

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/GetChannelImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/GetChannelImpl.kt
@@ -1,0 +1,35 @@
+package io.getstream.chat.android.livedata.usecase
+
+import io.getstream.chat.android.client.utils.Result
+import io.getstream.chat.android.livedata.ChatDomainImpl
+import io.getstream.chat.android.livedata.controller.ChannelController
+import io.getstream.chat.android.livedata.utils.Call2
+import io.getstream.chat.android.livedata.utils.CallImpl2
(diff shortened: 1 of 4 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/GetChannelImpl.kt
```diff
@@ -0,0 +1,35 @@
+package io.getstream.chat.android.livedata.usecase
+
+import io.getstream.chat.android.client.utils.Result
+import io.getstream.chat.android.livedata.ChatDomainImpl
+import io.getstream.chat.android.livedata.controller.ChannelController
+import io.getstream.chat.android.livedata.utils.Call2
+import io.getstream.chat.android.livedata.utils.CallImpl2
+import io.getstream.chat.android.livedata.utils.validateCid
+
+public interface GetChannel {
```
bychkovdmitry, line 10: A bit strange that we have GetThread, CetChannel use cases returning controllers. Would be better to name them GetThreadController, CetChannelController, etc. Unfortunately, these are breaking changes.

bychkovdmitry (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/GetChannelImpl.kt
tschellenbach: Get Channel is new, I like the GetChannelController name

tschellenbach:
good fix, will rename it

PR #772, opened 2020-10-28
[CAS-335] Refactor Call classes

zsmb13:
https://stream-io.atlassian.net/browse/CAS-335

### Description

The main goal of this PR is tidying up our usages of `Call` interfaces. We had two of these (one in the client, one in offline), and we had implementations with various different kinds of behaviour, misused coroutine scopes, etc.

See the highlights of the PR listed below, with more details in the inline comments on the changes.

---

With the new setup, we have a single `Call` interface, in the client. All implementations have the following ways of running them:
- `execute` is blocking and non-cancellable(so this should be called on a background thread)
- `enqueue` is main-safe, asynchronous and has cancellation support (up to the specific `Call` implementation to do this nicely), and *guarantees that the callback will be called on the main thread*
- `await` is a suspending extension function on `Call`, for easy usage of `Call` based tasks from Kotlin

---

This PR also introduces an `@InternalStreamChatApi` annotation, which can be used to share implementation between our modules, while deterring clients from using it themselves. This uses the `OptIn`/`RequiresOptIn` mechanisms of Kotlin (see more in [the official docs](https://kotlinlang.org/docs/reference/opt-in-requirements.html)). For now, this annotation lives in the client library (and package), but we can consider pushing into some kind of core module at a later time.

---

The coroutine scopes in `ChannelControllerImpl` and `QueryChannelsControllerImpl` have been removed, as they were unnecessary (and broken), and the `ChatDomain` scopes are now being used instead.

---

Use case classes are now all neatly formatted 🙂 

### Checklist

- [x] Reviewers added


stream-chat-android-client/build.gradle
```diff
@@ -56,6 +56,8 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
         freeCompilerArgs += [
                 '-progressive',
                 '-Xexplicit-api=strict',
+                '-Xopt-in=kotlin.RequiresOptIn',
+                '-Xopt-in=io.getstream.chat.android.client.internal.InternalStreamChatApi',
```
zsmb13, line 60: Adding the annotation as a compiler argument enables it for the entire module, so there's no opt-in required for each statement/class/file that uses it

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -0,0 +1,36 @@
...
+import io.getstream.chat.android.client.utils.Result
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+
+@InternalStreamChatApi
+public class CoroutineCall<T : Any>(
```
zsmb13, line 13: Replacement for `CallImpl`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -0,0 +1,36 @@
...
+    private val scope: CoroutineScope,
+    private val runnable: suspend () -> Result<T>
```
zsmb13, line 15: This `runnable` has to support proper coroutine cancellation for cancellation to work nicely, this is up to whoever creates instances of `CoroutineCall`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/Call.kt
```diff
@@ -1,23 +1,61 @@
...
+
+internal fun <T : Any, K : Any> Call<T>.map(mapper: (T) -> K): Call<K> {
+    return MapCall(this, mapper)
+}
 
-    public fun <K : Any> map(mapper: (T) -> K): Call<K>
-    public fun onError(handler: (ChatError) -> Unit): Call<T>
-    public fun onSuccess(handler: (T) -> Unit): Call<T>
-    public fun <K : Any> zipWith(call: Call<K>): Call<Pair<T, K>>
-    public fun <C : Any, B : Any> zipWith(callK: Call<C>, callP: Call<B>): Call<Triple<T, C, B>>
```
zsmb13, line 22: Half of these were unused, the rest were made internal and moved into extension functions. Clients should never use these in their code, they should just run `Call` instances in one of three ways.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/ZipCall.kt
```diff
@@ -2,88 +2,70 @@ package io.getstream.chat.android.client.call
...
-    fun <A : Any, B : Any> zip(callA: Call<A>, callB: Call<B>): Call<Pair<A, B>> {
-        return object : ChatCallImpl<Pair<A, B>>() {
             override fun execute(): Result<Pair<A, B>> {
-                val resultA = callA.execute()
+                val job = Job()
+                this.job = job
+
+                return runBlocking(job) {
+                    val deferredA = async { callA.await() }
+                    val deferredB = async { callB.await() }
```
zsmb13, line 29: These new implementation for `ZipCall` run the two `Call`s being zipped in parallel. We only use this in a single place, and this behaviour is appropriate there.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -190,9 +186,6 @@ internal class ChannelControllerImpl(
     val channelController = client.channel(channelType, channelId)
     override val cid = "%s:%s".format(channelType, channelId)
 
-    private val job = SupervisorJob()
-    val scope = CoroutineScope(Dispatchers.IO + domainImpl.job + job)
```
zsmb13, line 194: - This `job` was never cancelled, making it useless - the point of `Job` instances as a `CoroutineContext` and of `CoroutineScope` is cancellation.
- The `scope` created here is not a sub-scope of `domainImpl.job` in any way. Since `job` is also a `Job` element, it overwrites `domainImpl.job` in this combined `CoroutineContext`.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/utils/TestCall.kt
```diff
@@ -0,0 +1,20 @@
+package io.getstream.chat.android.livedata.utils
+
+import io.getstream.chat.android.client.call.Call
+import io.getstream.chat.android.client.utils.Result
+
+internal class TestCall<T : Any>(val result: Result<T>) : Call<T> {
```
zsmb13, line 6: Replaces `ChatCallTestImpl` and `TestResultCall `

samiuelson (dismissed):
Good improvement 👍🏼

tschellenbach:
Nice cleanup, I think the extra scope at the query channel controller and channel controller indeed doesn't make sense. You typically still want to complete those jobs even if you go close the specific query or channel. Typing events for instance, you still want to finish the API call to do stopTyping even when you navigate away. Makes sense to remove these.

Does this PR also address the issue that certain use cases are safe to run the UI thread? IE get channel controller is safe to run on the UI thread.

tschellenbach:
can you explain when we should use await vs the other 2 options?

tschellenbach (dismissed):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ErrorCall.kt
```diff
@@ -1,15 +1,24 @@
...
+import io.getstream.chat.android.client.call.Call
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.utils.Result
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+
+internal class ErrorCall<T : Any>(private val e: ChatError) : Call<T> {
+    override fun cancel() {
+        // Not supported
```
ogkuzmin, line 12: Why not throwing exception?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    override fun execute(): Result<T> {
+        return runBlocking { runnable() }
+    }
+
+    override fun enqueue(callback: (Result<T>) -> Unit) {
+        job = scope.launch {
+            val result = runnable()
+            withContext(Dispatchers.Main) {
```
ogkuzmin, line 31: As an improvement, we could specify dispatcher for result callback like operator `Flow.flowOn`. But for the first iteration we can omit it, it's not so often case

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/ZipCall.kt
```diff
@@ -2,88 +2,70 @@ package io.getstream.chat.android.client.call
...
-
-                if (!resultA.isSuccess) return Result(null, resultA.error())
-                val resultB = callB.execute()
-                if (!resultB.isSuccess) return Result(null, resultB.error())
-                val resultC = callC.execute()
-                if (!resultC.isSuccess) return Result(null, resultC.error())
-
-                return Result(Triple(resultA.data(), resultB.data(), resultC.data()), null)
-            }
+    fun <A : Any, B : Any> zip(callA: Call<A>, callB: Call<B>): Call<Pair<A, B>> {
```
ogkuzmin, line 15: nit: why function? You could also create a class with the same logic and name it ZipCall

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -115,7 +111,7 @@ internal class QueryChannelsControllerImpl(
             // - suspend/wait for a few seconds (yuck, lets not do that)
             // - post the refresh on a livedata object with only channel ids, and transform that into channels (this ensures it will get called after postValue completes)
             // - run the refresh channel call below on the UI thread instead of IO thread
-            scope.launch(Dispatchers.Main) {
+            domainImpl.scope.launch(Dispatchers.Main) {
```
ogkuzmin, line 118: What if we provide scope as an argument in the constructor? It helps us write tests

zsmb13:
>certain use cases are safe to run the UI thread? IE get channel controller is safe to run on the UI thread.

Addressed this on Slack as well, but to repeat it here: this happens to be an implementation detail for that use case, it's basically just lucky that it's relatively safe to call on the main thread. Best to treat it like all other `Call`s though, and use callback or `await` it, if not on a background thread.

> can you explain when we should use await vs the other 2 options?

Java clients are stuck using `execute` if they're on a background thread, or `enqueue` if they're on the main thread. Kotlin clients can use either of the two options above, or if they use coroutine in their apps, they can opt to use `await` instead, to get the best of both worlds: asynchronous non-blocking call that's easy to call and not having to nest callbacks.

Ideally, everyone would be on Kotlin & coroutines and just `await()` every `Call`. But not everyone will be using Kotlin yet, and not everyone will be using coroutines yet.

ogkuzmin (dismissed):
Really good job! It's a much cleaner and straight API than the previous one

adasiewiczr (dismissed):
Good job!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
zsmb13: Since clients don't know when they have an instance of `ErrorCall` and when they have some other, cancellable instance, I'd rather just silently not cancel here. Otherwise, cancelling a call (such as `client.queryChannel()`) will sometimes cancel it, and sometimes throw an exception for clients, which would be painful to deal with.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/ZipCall.kt
zsmb13: This was because of how the old implementation was structured, but good catch, I've changed this to match the style of `MapCall`.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
zsmb13: We can refactor this later if we need easier mocking for the scope in tests. For now, tests work fine through mocking the domain's scope.

adasiewiczr (approved):

ogkuzmin (approved):

PR #773, opened 2020-10-28
Chat initialization - Option 3 - CAS-351 - V2

tschellenbach:
Implementation of option 3 listed here while maintaining backwards compatibility with the current system:
https://stream-io.atlassian.net/browse/CAS-351

Tutorial PR is here: 
https://github.com/GetStream/android-chat-tutorial/pull/24/files

### Initialization

**Old syntax**

```
Chat.Builder("b67pax5b2wdq", applicationContext).apply { offlineEnabled=true }.build()
Chat.setUser()
// this seems nice, but then later on in the code we use
ChatDomain.instance()
// and
ChatClient.instance()
// without you having an idea where those came from
```

**New Syntax**

```
val client = ChatClient.Builder("b67pax5b2wdq", applicationContext).logLevel(ChatLogLevel.ALL).build()
val domain = ChatDomain.Builder(client).build()
val ux = ChatUX.Builder(client, domain).build()
client.setUser()
 ```

### Changing the level of integration

A very common scenario is starting out with UX enabled and then going for the client, or client + offline route.

**Old syntax**

all 3 libs:

```
Chat.Builder("b67pax5b2wdq", applicationContext).apply { offlineEnabled=true }.build()
Chat.setUser()
```

2 libs (fun fact, our own example app got it wrong in the past)
```
        ChatClient.Builder chatBuilder = new ChatClient.Builder(this.apiKey, context)
                .notifications(chatNotificationHandler)
                .logLevel(chatLogLevel);

final ChatDomain.Builder domainBuilder = new ChatDomain.Builder(context, client, user);
domainBuilder
                .userPresenceEnabled()

// switch it when the user changes

        client.setUser(user, userToken, new InitConnectionListener() {
            @Override
            public void onSuccess(@NotNull ConnectionData data) {
                callbacks.onSuccess(data);
            }

            @Override
            public void onError(@NotNull ChatError error) {
                callbacks.onError(error);
            }
        });
```

**New Syntax**

```
val client = ChatClient.Builder("b67pax5b2wdq", applicationContext).logLevel(ChatLogLevel.ALL).build()
val domain = ChatDomain.Builder(client).build()
// remove 1 or 2 lines, everything still works as intended val ux = ChatUX.Builder(client, domain).build()
client.setUser()
 ```

### Conclusion

- The old syntax was confusing and frustrating for people that are changing their integration levels. (if you only use the UI it's fine, but if not you'll get confused)
- The different ways to initialize things also made it harder to document things in a consistent manner
- The new approach, makes it as simple as commenting a line and removing a dependency
- Note that this PR is not removing the Chat class, it's marked deprecated, but it still works so people can upgrade once everything is fully polished on our end in terms of docs.

### Future improvements

- ChatDomain internal naming and docs can be cleaned up
- ChatUX naming


stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUI.kt
```diff
@@ -0,0 +1,102 @@
...
+        private var fonts: ChatFonts = ChatFontsImpl(style, client.appContext)
+        private var strings: ChatStrings = ChatStringsImpl(client.appContext)
+
+        public fun withMarkdown(markdown: ChatMarkdown): Builder {
+            this.markdown = markdown
+            return this
+        }
+
+        public fun withUrlSigner(signer: UrlSigner): Builder {
+            this.urlSigner = signer
```
carterhudson, line 68: All builder methods could be
```kotlin
 = apply {
    ...
}
```
if you want

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -288,10 +322,14 @@ public class ChatClient internal constructor(
     }
 
     public fun disconnect() {
+
```
leandroBorgesFerreira, line 325: No need for this blank line

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatClientConfig.kt
```diff
@@ -1,24 +1,17 @@
...
-    var baseTimeout: Long,
-    var cdnTimeout: Long,
-    val warmUp: Boolean,
-    val loggerConfig: ChatLogger.Config,
-    val notificationsHandler: ChatNotificationHandler,
-    val fileUploader: FileUploader? = null,
-    val tokenManager: TokenManager = TokenManagerImpl()
+public class ChatClientConfig(
+    public val apiKey: String,
+    public var httpUrl: String,
```
leandroBorgesFerreira, line 7: I know that you're not the creator of this class, but perhaps this can be changed to val instead of var.

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -99,6 +89,8 @@ public void stopped() {
 
         chatBuilder.build();
 
+
```
leandroBorgesFerreira, line 92: Not necessary

stream-chat-android/src/main/java/com/getstream/sdk/chat/navigation/destinations/WebLinkDestination.kt
```diff
@@ -3,10 +3,9 @@ package com.getstream.sdk.chat.navigation.destinations
 import android.content.Context
 import android.content.Intent
 import android.net.Uri
-import com.getstream.sdk.chat.Chat
 
 internal class WebLinkDestination(url: String, context: Context) : ChatDestination(context) {
-    private val url: String = Chat.instance().urlSigner().signFileUrl(url) ?: ""
+    private val url: String = com.getstream.sdk.chat.ChatUI.instance().urlSigner.signFileUrl(url) ?: ""
```
leandroBorgesFerreira, line 8: No need for the entire package

leandroBorgesFerreira (approved):
Looks much easier to understand. Some comments.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -633,7 +675,10 @@ public class ChatClient internal constructor(
     }
 
     private fun getTokenAndConnect(connect: () -> Unit) {
-        config.tokenManager.loadAsync {
+        tokenManager.loadAsync {
+            if (it.isSuccess) {
+                state.token = it.data()
```
JcMinarro, line 680: Instead of store the token into the `state` object, Why don't we use hte token into the `TokenManager`?
At the end it could be not updated, because the token manager could be asked to reload the token from multiple part of our code

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -50,17 +56,24 @@ import java.io.File
...
 
 public class ChatClient internal constructor(
-    private val config: ChatClientConfig,
+    public val config: ChatClientConfig,
     private val api: ChatApi,
     private val socket: ChatSocket,
-    private val notifications: ChatNotifications
-) {
+    private val notifications: ChatNotifications,
+    public var appContext: Context,
```
JcMinarro, line 63: Because we store an instance of the `ChatClient` into the companion object, it is similar than store it statically in Java. And here we store a context into an static reference, that could produce a memory leak. We should try to avoid to store `context` properies into our classes

tschellenbach: The easy fix here is to pass the context multiple times to the 3 builders. Is there another solution we should consider?

tschellenbach:
Will merge this and start a new PR for finetuning in order to avoid more merge conflicts given the large size of this change.

PR #774, opened 2020-10-28
implements a subscribefor method that is lifecycle sensitive

carterhudson:
https://github.com/GetStream/android-chat-tutorial/pull/22#discussion_r512397171

### Description
Supports a lifecycle sensitive `subscribeFor` method used in the tutorial project. Open to discussion about the best way to go about it. Is Destroy the right place to call `dispose()`? Would this need to be included in the changelog?

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -288,6 +289,32 @@ public class ChatClient internal constructor(
...
+        )
+
+        lifecycleOwner.lifecycle.addObserver(
+            object : LifecycleObserver {
+                @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
+                fun onDestroy() {
+                    disposable.dispose()
+                }
+            }
+        )
```
zsmb13, line 313: This performs _most_ of what `LiveData` does for safe observations:

- It only delivers values if your lifecycle is in a started state, so that you don't touch UI when you shouldn't
- It cleans up the observer when the lifecycle ends

It's missing one thing: if the lifecycle is not _started_, it won't queue values and wait for it to become _started_ again - it will simply drop them.

zsmb13:
I pushed these methods into the client and controller classes directly, as that's the only way they'll be available from Java. The extensions only provide Kotlin-friendly APIs on top of those core implementations, enabling the usage of `KClass` instead of `Class`, and generic type parameters instead of passing in `Class` or `KClass` references.

adasiewiczr:
Could you also add KDoc comments for those public methods?

tschellenbach:
@JcMinarro needs a check from you as well. I'm not sure we should even solve the lifecycle problem for the user. For the tutorial its a bit of a problem, for real code you would just ViewModelScope

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: I think it is fine, it should be started when it is subscribed

JcMinarro (approved):

adasiewiczr (approved):

tschellenbach:
nice work

PR #775, opened 2020-10-28
CAS-334/Fixing new messages for load more

leandroBorgesFerreira:
[CAS-334](https://stream-io.atlassian.net/browse/CAS-334)

### Description

Fixing the message "New Messages" when users load new messages. We can't show the this messages, because there's no new messages. Now it doesn't show anymore.

**Important:** I changed the way we spread of messages so we can handle the loadingMore in the `MessageListItemWrapper` which was always false until then. Please take a look at this change as it affects many parts of the code. 

### Checklist

- [ ] Changelog updated with client-facing changes - No need. This is just a bug fix.
- [ ] New code is covered by unit tests - No need. This is just a bug fix.
- [ ] Comparison screenshots added for visual changes - No need. This is just a bug fix.
- [ ] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -87,7 +88,7 @@ internal class ChannelControllerImpl(
 ) :
     ChannelController {
     private val editJobs = mutableMapOf<String, Job>()
-    private val _messages = MutableLiveData<Map<String, Message>>()
+    private val _messages = MutableLiveData<Pair<Boolean, Map<String, Message>>>()
```
ogkuzmin, line 91: Honestly, I think it's not right to create for the source of live data for messages of a channel such data structure as Pair<Boolean, Map<String, Message>>. Seems too complicated, for such a case I'd like to have some data class:
``` kotlin
 data class MessagesStates(val messages: List<Message>, val loadingState: Booleand = false)
```
or even clearer:
``` kotlin
sealed class MessagesState {
 class IdleMessages(val messages: List<Message>) : MessagesState()
 class LoadingMoreMessages(val messages: List<Message>) : MessagesState()
}
```

ogkuzmin (commented):
As I see the main reason for such a boolean flag is using it in `MessageListItemLiveData`. It seems for such reason you even could place just LiveData<Boolean> as a source of loading more state in `ChannelControllerImpl` and make other logic there (in `MessageListItemLiveData`). So much change would make fewer changes in the current codebase and let you achieve the same

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
```diff
@@ -26,7 +27,7 @@ public interface ChannelController {
     public val channelType: String
     public val channelId: String
     /** a list of messages sorted by message.createdAt */
-    public val messages: LiveData<List<Message>>
```
JcMinarro, line 29: It is a breaking change in our public API.
Apart of that, it is breaking the previous comment and the KDocs over the interface.
Why don't we provide it into another property?

JcMinarro (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
samiuelson: If a property must be deprecated, a proper `deprecated` annotation should be added, explaining what and how to use instead.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: Ok. I'll do that.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
leandroBorgesFerreira: I don't mind providing it in another property, but I don't see another way. I need to differentiate new messages from old ones so the property `loadingMore`  from `MessageListItemWrapper` gets the right information... until now it always get false.

JcMinarro: This "public livedata" could be a transformation of the new livedata you are creating.
We keep compatibility and the new feature is done

leandroBorgesFerreira: Without changing `public val messages: LiveData<List<Message>>` I don't see a way to differentiate new messages from old messages that were loaded from the history of the conversation.

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
```diff
@@ -81,19 +82,19 @@ internal class MessageListItemLiveData(
...
 
     @UiThread
-    internal fun messagesChanged(messages: List<Message>) {
-        messageItemsBase = groupMessages(messages)
+    internal fun messagesChanged(messagesUpdate: MessagesUpdate) {
+        messageItemsBase = groupMessages(messagesUpdate.messages)
         messageItemsWithReads = addReads(messageItemsBase, readsLd.value)
         val out = messageItemsWithReads + typingItems
-        val wrapped = wrapMessages(out, hasNewMessages)
+        val wrapped = wrapMessages(out, hasNewMessages, messagesUpdate.isLoadingMore)
```
bychkovdmitry, line 89: Not related to this PR, but why do we reconstruct MessageListItemWrapper every time instead of reducing its state? E.g. `value = value.copy(items = newItems)`

leandroBorgesFerreira:
#796  Substitute this one.

PR #776, opened 2020-10-29
Move library build jobs to a single GitHub Actions workflow

zsmb13:
### Description

We are using separate GitHub Actions _workflows_ that are triggered the same way, for each project. Each of them has a single job in it, which builds, ktlint's, and tests the given project.

This PR merges these into a single workflow with three jobs (which run in paralllel). It also formats the config and unifies it across all libraries.

---

I also tried splitting out the major parts of the jobs in two - lint and tests. These two parts each take roughly 3 minutes to execute. Surprisingly, when split out, tests take about twice as long to execute, eliminating any possible gains (see a regular run [here](https://github.com/GetStream/stream-chat-android/runs/1325399456?check_suite_focus=true), and a split one [here](https://github.com/GetStream/stream-chat-android/runs/1325401504)). Not sure why running ktlint before the tests makes the tests faster, as I don't believe ktlint should be building the project when it runs.

### Checklist

- [x] Reviewers added


.github/workflows/stream-chat-client-workflow.yml
```diff
@@ -1,12 +1,11 @@
-name: Stream Chat Android Client CI
+name: Stream Chat Android CI
```
samiuelson, line 1: Let's rename the file too 🕵🏼

samiuelson (changes requested):
Looks good 👍🏼  
Just 1 tiny request

.github/workflows/stream-chat-client-workflow.yml
zsmb13: Nice catch!

samiuelson (approved):

adasiewiczr (approved):
LGTM! I'm just thinking if we need triggers for both push and PR if we require branches to be up-to-date with develop. WDYT?

zsmb13:
The PR up-to-dateness topic is an open one, and yes, we can probably get rid of the `pull` trigger as we figure that out.

PR #779, opened 2020-10-29
Fix empty screen when returning back to channels

bychkovdmitry:
### Description

Steps to reproduce the issue:
1. Scroll channel list to the end
2. Navigate to any channel
3. Go back to the channel list and observe that the list is empty

This happens because the last state we receive on the channel list page is `EndPageReached`. When going back to that page, the UI is recreated with `EndPageReached` state and we observe an empty screen.

The root of the problem is that we try to implement our ViewModels in MVI/Redux style with a single `State` which is enough to recreate the whole UI at any point of time. Unfortunately, some of the `State` classes are more like events (e.g. `NavigateToLoginScreen`, `EndPageReached`) and they lack important bits of information to recreate the UI (e.g. when going back to the previous screen).

We have 2 options:

- Implement `State` classes in a way they contain all the information to render the screen from scratch at any point of time. The downside is that all the UI will be rendered again every time the state is reduced. I think we can't afford running diffutils every time something irrelevant changes (e.g. progress bar is shown over the channels list when hiding a channel).
- Split the state into several partial states to improve performance.

In any case, we need a separate event bus like `SingleLiveEvent` to trigger one shot events (navigation, toasts, etc.) 

### Checklist

- [X] Reviewers added

### DIFF
--- a/stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelsViewModel.kt
+++ b/stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelsViewModel.kt
@@ -30,12 +30,12 @@ public class ChannelsViewModel(
     private val limit: Int = 30
 ) : ViewModel() {
     private val channelsData: LiveData<State>
-    private val loadingMoreData: LiveData<State.LoadingNextPage>
     private val loadingData = MutableLiveData<State.Loading>()
-    private val endPageData: LiveData<State.EndPageReached>
     private val stateMerger = MediatorLiveData<State>()
-
     public val state: LiveData<State> = stateMerger
+
+    private val paginationStateMerger = MediatorLiveData<PaginationState>()
+    public val paginationState: LiveData<PaginationState> = paginationStateMerger
 
     init {
         val queryChannelsController = chatDomain.useCases.queryChannels(filter, sort, limit).execute().data()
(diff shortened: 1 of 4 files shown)


stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelsViewModel.kt
```diff
@@ -22,13 +22,13 @@ public class ChannelsViewModel(
...
     private val channelsData: LiveData<State>
-    private val loadingMoreData: LiveData<State.LoadingNextPage>
     private val loadingData = MutableLiveData<State.Loading>()
-    private val endPageData: LiveData<State.EndPageReached>
     private val stateMerger = MediatorLiveData<State>()
-
     public val state: LiveData<State> = stateMerger
 
+    private val paginationStateMerger = MediatorLiveData<PaginationState>()
+    public val paginationState: LiveData<PaginationState> = paginationStateMerger
```
ogkuzmin, line 30: I'd like to use it inside the main state of the screen as part of Result state with list of channels. It will let us to keep an MVI architecture

bychkovdmitry: Then, I guess, we will need to rewrite our neat sealed classes to something more generic like a single state data class

bychkovdmitry: We need to test DiffUtil performance and decide if we can afford to execute it frequently. For example, when we reach the end of the channel list we observe new state:

1. When loading more is set to true
2. When new chunk of channels arrive
3. When loading more is set to false
4. When `endOfChannels` emits true

Theoretically this can be reduced to just 2 emits: 
1. When loading more starts (probably displaying progress indicator)
2. When data arrives together with with `endOfChannels`flag

ogkuzmin: Your version works fine, it's just my sense of architecture approach and coding ideas. Let's discuss with others

zsmb13 (approved):
LGTM

JcMinarro (approved):

PR #783, opened 2020-10-30
Fix some issues in ChatDomainImpl

zsmb13:
### Description

See commit messages & inline comments

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -187,7 +187,6 @@ internal class ChatDomainImpl internal constructor(
         _online.value = false
         _totalUnreadCount.value = 0
         _channelUnreadCount.value = 0
-        _errorEvent.value = null
         _banned.value = false
         _mutedUsers.value = emptyList()
         if (activeChannelMapImpl != null) {
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -157,9 +157,17 @@ internal class ChatDomainImpl internal constructor(
      */
     override val errorEvents: LiveData<Event<ChatError>> = _errorEvent
 
+    /** the event subscription, cancel using repo.stopListening */
+    private var eventSubscription: Disposable? = null
+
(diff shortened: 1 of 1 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -187,16 +195,11 @@ internal class ChatDomainImpl internal constructor(
         _online.value = false
         _totalUnreadCount.value = 0
         _channelUnreadCount.value = 0
-        _errorEvent.value = null
```
zsmb13, line 190: Lint (as in Android Lint, not Ktlint) was failing on this line, we're just not running it very often. Can't set a `null` value in a `LiveData` that has a non-nullable type parameter. We don't really need to reset it though, because we're using it for events.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -187,16 +195,11 @@ internal class ChatDomainImpl internal constructor(
...
         _banned.value = false
         _mutedUsers.value = emptyList()
-        if (activeChannelMapImpl != null) {
-            activeChannelMapImpl.clear()
-        }
 
-        if (activeQueryMapImpl != null) {
-            activeQueryMapImpl.clear()
-        }
```
zsmb13, line 199: We had null checks on non-nullable, final values here, because this code was being called from an `init` block that was higher up in the class than the property declarations, and therefore those were not initialized yet. Moved the properties up top to avoid this.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -157,9 +157,17 @@ internal class ChatDomainImpl internal constructor(
      */
     override val errorEvents: LiveData<Event<ChatError>> = _errorEvent
 
+    /** the event subscription, cancel using repo.stopListening */
+    private var eventSubscription: Disposable? = null
+
+    /** stores the mapping from cid to channelRepository */
+    private val activeChannelMapImpl: ConcurrentHashMap<String, ChannelControllerImpl> = ConcurrentHashMap()
```
ogkuzmin, line 164: Oh, it doesn't mean good things for me... It seems that yesterday I was wrong saying "Flow doesn't give us any enhancement right now"

adasiewiczr (approved):

PR #785, opened 2020-10-30
CAS-308 create new module for the new design

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-308

### Description

Created UI components library and corresponding sample app

### Checklist

- [X] Reviewers added


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
```diff
@@ -0,0 +1,25 @@
+package io.getstream.chat.ui.sample.application
+
+import android.app.Application
+import io.getstream.chat.ui.sample.data.user.UserRepository
+
+class App : Application() {
+
+    lateinit var chatInitializer: ChatInitializer
```
JcMinarro, line 8: Instead of `lateinit var` we can use lazy

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
```diff
@@ -0,0 +1,25 @@
...
+    lateinit var userRepository: UserRepository
```
JcMinarro, line 9: It could be `val` as well

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/common/Extensions.kt
```diff
@@ -0,0 +1,76 @@
...
+    get() = extraData["image"] as String
+    set(value) {
+        extraData["image"] = value
+    }
+
+var User.name: String
+    get() = extraData["name"] as String
+    set(value) {
+        extraData["name"] = value
+    }
```
JcMinarro, line 27: It is already implemented in our `stream-chat-android-client` 
It shouldn't be needed here
```suggestion
```

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/data/user/User.kt
```diff
@@ -0,0 +1,13 @@
+package io.getstream.chat.ui.sample.data.user
+
+data class User(
```
JcMinarro, line 3: Could we use another name for this class?
It could be a bit confusing because we already have `User` in our SDK.
What about something like `DevelopmentUser` or any other name that reflects what it is?

stream-chat-android-ui-components/build.gradle
```diff
@@ -0,0 +1,127 @@
...
+
+    implementation Dependencies.androidxMedia
+    implementation Dependencies.coil
+    implementation Dependencies.coilGif
+
+    // ExoPlayer
+    api Dependencies.exoplayerCore
+    api Dependencies.exoplayerDash
+    api Dependencies.exoplayerHls
+    api Dependencies.exoplayerSmoothstreaming
```
JcMinarro, line 114: Do we really need to expose them?
Not only that, do we need to add them to our project?
Aren't they already added by our SDK?

stream-chat-android-ui-components-sample/build.gradle
```diff
@@ -0,0 +1,141 @@
...
+    implementation platform(Dependencies.firebaseBom)
+    implementation Dependencies.firebaseAnalytics
+    implementation Dependencies.firebaseCrashlytics
+
+    implementation Dependencies.timber
+
+    implementation Dependencies.glide
+    annotationProcessor Dependencies.glideCompiler
+
+    //debugImplementation Dependencies.leakCanary
```
zsmb13, line 140: Let's not add this if we're not using it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChannelsView.kt
```diff
@@ -0,0 +1,18 @@
...
+public class ChannelsView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : AppCompatTextView(context, attrs, defStyleAttr) {
+
+    public fun setChannels(channels: List<Channel>) {
+        text = channels.joinToString(separator = "\n") { it.name }
+    }
+}
```
zsmb13, line 18: Love this implementation 😄

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChannelsView.kt
```diff
@@ -0,0 +1,18 @@
+package io.getstream.chat.android.ui
+
+import android.content.Context
+import android.util.AttributeSet
+import androidx.appcompat.widget.AppCompatTextView
+import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.name
+
+public class ChannelsView @JvmOverloads constructor(
```
zsmb13, line 9: We use this in a lot of places, but it's [not entirely good practice](https://zsmb.co/jvmoverloads-for-android-views/)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChannelsViewModelBinding.kt
```diff
@@ -0,0 +1,18 @@
...
+package io.getstream.chat.android.ui
+
+import androidx.lifecycle.LifecycleOwner
+import com.getstream.sdk.chat.viewmodel.channels.ChannelsViewModel
+
+@JvmName("bind")
+public fun ChannelsViewModel.bindView(
+    view: ChannelsView,
+    lifecycle: LifecycleOwner
+): ChannelsViewModel = apply {
```
zsmb13, line 12: Doesn't seem like we really need a return value here

stream-chat-android-ui-components/build.gradle
```diff
@@ -0,0 +1,127 @@
...
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions {
+        freeCompilerArgs += [
+                '-progressive',
+                '-Xexplicit-api=strict',
+        ]
+    }
+}
+
+dependencies {
```
zsmb13, line 71: Until we're not using these dependencies, it would be best not to add them to this module. We'll never review them later, now is the best time to start with a minimal set, and only add what we need as we go.

stream-chat-android-ui-components-sample/src/debug/kotlin/io/getstream/chat/ui/sample/application/DebugMetricsHelper.kt
```diff
@@ -0,0 +1,19 @@
+package io.getstream.chat.ui.sample.application
+
+import android.os.StrictMode
+import timber.log.Timber
+
+class DebugMetricsHelper {
```
zsmb13, line 6: Should be an `object`, its instance isn't used in any way

stream-chat-android-ui-components-sample/src/release/kotlin/io/getstream/chat/ui/sample/application/AppConfig.kt
```diff
@@ -0,0 +1,36 @@
+package io.getstream.chat.ui.sample.application
+
+import io.getstream.chat.ui.sample.data.user.User
+
+object AppConfig {
```
zsmb13, line 5: This file is identical in debug and release, why do we have it duplicated?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
```diff
@@ -0,0 +1,25 @@
...
+
+        chatInitializer = ChatInitializer(this)
+        userRepository = UserRepository()
+
+        DebugMetricsHelper().init()
+    }
+
+    companion object {
+        lateinit var instance: App
+            private set
```
zsmb13, line 23: Might be worth a comment to recommend using a DI library (not a specific one, just any) in real applications instead of this approach 🙂

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/data/user/User.kt
zsmb13: `SampleUser`?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
zsmb13: This is the intended use case for `lateinit`, I think it's good as-is

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
zsmb13: I'd go for `private set`

stream-chat-android-ui-components-sample/src/release/kotlin/io/getstream/chat/ui/sample/application/DebugMetricsHelper.kt
```diff
@@ -0,0 +1,5 @@
+package io.getstream.chat.sample.application
```
zsmb13, line 1:
```suggestion
package io.getstream.chat.ui.sample.application
```

stream-chat-android-ui-components-sample/src/release/kotlin/io/getstream/chat/ui/sample/application/AppConfig.kt
bychkovdmitry: I have no idea TBH, do we have some kind of prod environment for sample app? Might be worth to have it with a cron job which occasionally purges all the data populates the db with some neat channels/messages.

stream-chat-android-ui-components-sample/build.gradle
```diff
@@ -0,0 +1,141 @@
...
+
+buildscript {
+    repositories {
+        jcenter()
+        google()
+    }
+    dependencies {
+        classpath Dependencies.navigationSafeArgsGradlePlugin
+    }
+}
```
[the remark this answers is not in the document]
bychkovdmitry: Nice catch

bychkovdmitry:
I addressed all the comments except for the one related to duplicate AppConfig with the same data.

bychkovdmitry:
Also, moved AppConfig to the main source set.

JcMinarro (approved):

PR #786, opened 2020-10-30
Change ConcurentHasMap to AtomicReference and usage of CAS alg

ogkuzmin:
### Description
In this PR #783, we brought ConcurrentHashMap to work with one collection in a multi-threaded environment. It's not the best way to work with such a thing, but such code lets you fix the problem fast. From my point of view, it's always better to avoid synchronized collections, they are slow. There are several non-blocking algorithms for the same purpose. One of them is [Compare-and-Swap (CAS)](https://en.wikipedia.org/wiki/Compare-and-swap). Its' performance is much better.
I've implemented it in a general way and added some extensions functions.

### Checklist

- [x] Reviewers added


zsmb13:
Let's look into whether the current implementation is actually significantly slowing us down / the new one is way faster. If that's not the case, the existing simple code is a better way to go, as both implementations seem logically correct.

PR #790, opened 2020-10-30
CAS-381/fixing background crash

leandroBorgesFerreira:
[CAS-381](https://stream-io.atlassian.net/browse/CAS-381)
### Description

The app is breaking when going in the background. 

Creating new clients, without disconnecting previous ones, create problems are many instances dispute the resources of the SDK. It necessary to disconnect the chat to prevent to stack many chat instances to be initialised. 

### Checklist

- [ ] Changelog updated with client-facing changes - No applicable.
- [ ] Comparison screenshots added for visual changes - Not applicable
- [ ] Reviewers added

### DIFF
--- a/stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/ChatInitializer.kt
+++ b/stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/ChatInitializer.kt
@@ -7,10 +7,18 @@ import io.getstream.chat.android.client.logger.ChatLogLevel
 import io.getstream.chat.android.client.notifications.handler.NotificationConfig
 import io.getstream.chat.android.livedata.ChatDomain
 import io.getstream.chat.sample.R
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
 
 class ChatInitializer(private val context: Context) {
 
+    var client: ChatClient? = null
+    var domain: ChatDomain? = null
+    var ux: ChatUI? = null
+
     fun init(apiKey: String) {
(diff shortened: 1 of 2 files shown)


stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/application/ChatInitializer.kt
```diff
@@ -21,8 +29,17 @@ class ChatInitializer(private val context: Context) {
...
+                .logLevel(ChatLogLevel.ALL).build()
+
+        domain =
+            ChatDomain.Builder(client!!, context).offlineEnabled().notificationConfig(notificationConfig).build()
+        ux = ChatUI.Builder(client!!, domain!!, context).build()
+    }
+
+    private fun disconnectChatsIfNecessary() {
+        client?.disconnect()
+        GlobalScope.launch(Dispatchers.Main) { domain?.disconnect() }
```
tschellenbach, line 43: i don't think this last disconnect is needed, the domain follows the .disconnect() call from the client. https://github.com/GetStream/stream-chat-android/blob/develop/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt#L280

tschellenbach: Code can be simplified using: https://github.com/GetStream/stream-chat-android/blob/develop/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt#L945

JcMinarro:
The real issue here was that we started our SDK twice.
I have fixed it here #795

PR #792, opened 2020-11-02
Store query entity in run query online flow.

ogkuzmin:
### Description

When manual checking I noticed that if we didn't query online at startup and ran query when connection was recovered, then we didn't store channel query. That's why after rerun of app we don't get channels from local repo.

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
@@ -143,6 +143,7 @@ internal class QueryChannelsControllerImpl(
             logger.logI("found ${channels.size} channels in offline storage")
         }
 
+        updateChannelsAndQueryResults(channels, pagination.isFirstPage)
         return channels
     }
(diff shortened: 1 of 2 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -202,7 +202,6 @@ internal class QueryChannelsControllerImpl(
             output = result
             if (result.isSuccess) {
```
adasiewiczr, line 203: Shouldn't we move the whole if into `runQueryOnline`? I think we won't call `updateChannelsAndQueryResults` after recovery which updates channels LiveData

ogkuzmin: Yes, you're right. I did it for the online/offline execution of a query. But there is some duplication in the recovery flow. It's not so easy to decouple right now. I think we can make some refactoring there, but not right now.

adasiewiczr (approved):

PR #796, opened 2020-11-02
[CAS-382] [CAS-388] Fix livedata updated in background thread

JcMinarro:
### Description

Some of our LiveData was been used from a background thread when calling `setValue()`, and it is not allowed.
There were other race-conditions with our eventListeners and the properties that should be initialized but weren't. All Listener Initialization has been moved after the user is set

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -258,7 +258,7 @@ internal class ChatDomainImpl internal constructor(
         }
 
         if (client.isSocketConnected()) {
-            setOnline()
+            postOnline()
         }
         startListening()
         initClean()
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -194,14 +194,13 @@ internal class ChatDomainImpl internal constructor(
     override var retryPolicy: RetryPolicy =
         DefaultRetryPolicy()
 
-    internal fun clearState() {
-        _initialized.value = false
-        _online.value = false
-        _totalUnreadCount.value = 0
-        _channelUnreadCount.value = 0
-        _banned.value = false
+    private fun clearState() {
+        _initialized.postValue(false)
+        _online.postValue(false)
+        _totalUnreadCount.postValue(0)
+        _channelUnreadCount.postValue(0)
+        _banned.postValue(false)
         _mutedUsers.value = emptyList()
-
         activeChannelMapImpl.clear()
         activeQueryMapImpl.clear()
     }
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -159,7 +159,10 @@ internal class ChatDomainImpl internal constructor(
     override val errorEvents: LiveData<Event<ChatError>> = _errorEvent
 
     /** the event subscription, cancel using repo.stopListening */
-    private var eventSubscription: Disposable? = null
+    private var eventSubscription: Disposable = object : Disposable {
+        override val isDisposed: Boolean = true
+        override fun dispose() { }
+    }
 
     /** stores the mapping from cid to channelRepository */
     private val activeChannelMapImpl: ConcurrentHashMap<String, ChannelControllerImpl> = ConcurrentHashMap()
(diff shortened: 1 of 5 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -515,7 +511,9 @@ internal class ChatDomainImpl internal constructor(
 
     internal fun setOnline() {
         isOnline = true
-        _online.value = true
+        scope.launch(Dispatchers.Main) {
+            _online.value = true
+        }
```
tschellenbach, line 516: this solves the error when calling this function outside of the UI thread (identical to doing _online.posValue(true) btw). however it introduces a new category of errors related to race conditions on the online value.

so far i've seen blogposts take different approaches:
1. specify which functions can run on the main thread
2. use withContext to switch to the right thread
3. use launch/postValue like you're doing here

I'm not sure what's right tbh

tschellenbach: here's a basic description of when/how to use the 3 types of async calls in kotlin: https://medium.com/@vcanato/withcontext-async-or-launch-inside-coroutine-kotlin-17c332854a08

so far my gut feeling is that the right approach is to:
- use withContext to ensure that code runs on the right thread
- be very careful around postValue and reading the value. it's easy to have race conditions if you call postValue and read the value outside of the UI thread (since it's possible you'll read a stale value)

tschellenbach: IE something like this: https://github.com/GetStream/stream-chat-android/pull/797

JcMinarro: The right way would be converting our `setUser()` to a suspend function, then we avoid all race-conditions and run every method on the thread it needs to.
The problem is that this class creates its own scope and is not ready to be used in tests. Doing this change our tests fail because some properties are not initialized before they are used

tschellenbach (dismissed):
longer term we need a cleaner setup for handling concurrency, not good to try to solve that today though, best to first fix the issue

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
zsmb13: The simplest thing to do is `withContext`, which will do a suspending thread switch (let go of the caller thread, especially if that's the main thread), but won't proceed to the next line until its contents are executed. Makes it easier to reason about the code. Using a single `launch` in a scope that's tied to a UI component (`LifecycleOwner`, `ViewModel`) and then only switching threads using `withContext` will result in very simple sequential code with coroutines. Rarely is nesting `launch` or `async` calls inside each other actually required.

In contrast, `launch` is akin to posting a `Runnable` to a thread, it will run when that thread is next available, i.e. it's asynchronous. Using `launch` to go to the main thread will not immediately execute that code, it will only run it when the main looper is next available.

>be very careful around postValue and reading the value

The core issue here is misusing `LiveData` for something it's simply not. If you need to read the `value` property of `LiveData`, you already lost. It will only behave predictably if you observe its values in a UI component, because *that's the only thing it's meant to do*. Frankly, the `value` getter should just throw an exception.

> The right way would be converting our `setUser()` to a suspend function

Can't do this as long as Java clients need to be able to call it. However, maybe it could somehow return a `Call` implementation, which would then allow synchronous, callback-based, and suspending use?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -159,7 +159,10 @@ internal class ChatDomainImpl internal constructor(
     override val errorEvents: LiveData<Event<ChatError>> = _errorEvent
 
     /** the event subscription, cancel using repo.stopListening */
-    private var eventSubscription: Disposable? = null
+    private var eventSubscription: Disposable = object : Disposable {
+        override val isDisposed: Boolean = true
+        override fun dispose() { }
+    }
```
zsmb13, line 165: It would be nice to move this `Disposable` implementation into a "static" field and give it a descriptive name

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
JcMinarro: `setUser()` can be a suspend function because it is not accesible out-side of the SDK, but because we are not injecting coroutine scope into our dependencies (They are created inside of the class) and we are mixing the use of coroutines and Livedata, our tests arrive to some race-conditions that is not easy to be fixed for now (Lot of tests will fails with the current implementation)

JcMinarro:
After spend lot of time refactoring our current using of coroutines (With not a really good result), I have reverted it and used the LiveData mechanism.

tschellenbach (approved):

PR #797, opened 2020-11-02
Fix/livedata updated in background thread2

tschellenbach:
Created this PR just for discussion purposes.
There are many different solutions that all ensure the code runs on the right thread.

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -511,7 +511,9 @@ internal class ChatDomainImpl internal constructor(
 
     internal fun setOnline() {
         isOnline = true
-        _online.value = true
+        scope.launch(Dispatchers.Main) {
+            _online.value = true
+        }
     }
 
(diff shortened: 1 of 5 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -233,6 +234,10 @@ internal class ChatDomainImpl internal constructor(
 
         // load channel configs from Room into memory
         initJob = scope.async(scope.coroutineContext) {
+            clearState()
+            if (client.isSocketConnected()) {
+                setOnline()
+            }
```
tschellenbach, line 240: withContext could also be used here, this prevents having to make changes to our test codebase

JcMinarro: `initJob` is not going to be completed in test-side, blocking some of our tests that will be waiting `initJob` to be completed

tschellenbach: indeed

PR #800, opened 2020-11-03
Add NORMAL synchronous mode for SQLite

ogkuzmin:
### Description
Speed up of messages fetching from DB. 
I noticed that sometimes message queries from DB for channel takes some long time. At my slow device _(1 CPU, 750 MB RAM_) for one channel it takes **246.570539 ms** in average. And also I see we have already added indices for table for its' selection params.
 I looked at SQLite manual and found very important setup parameter `PRAGMA synchronous` (docs available [here](https://www.sqlite.org/pragma.html#pragma_synchronous)). It defines how DB ensures its' sync via multithreaded transactional environment (it needs for critical situations as OS crashes). Since Lollipop SQLite uses `WAL mode` for journaling (faster then the previous `PERSIST mode`). And with `WAL` we don't need `FULL synchronous mode` it's too strict. We can apply `NORMAL synchronous mode` that makes our use-case of SQLite is faster but still reliable even when app crash occurs. 
I measured changed sync mode at testing emulator. 
Result: and querying of messages for channel had decreased from **246.570539 ms** in average to **121.3027312 ms** in average. Result of experiments are available [here](https://docs.google.com/spreadsheets/d/13ZCXOnNK8B4t_SdjLLWetyDCaK2piCW4gwXU7dxRYSc/edit?usp=sharing)
So in this PR I apply `NORMAL synchronous mode`

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDatabase.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDatabase.kt
@@ -41,7 +41,7 @@ import io.getstream.chat.android.livedata.entity.UserEntity
         ChannelConfigEntity::class,
         SyncStateEntity::class
     ],
-    version = 28,
+    version = 27,
     exportSchema = false
 )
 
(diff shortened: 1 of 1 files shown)


ogkuzmin:
P.S. Measurements are not 100 % precise, but gives you some approximation.

JcMinarro (commented):
Don't we need to increase the DB Version to have those chagnes applied?

ogkuzmin:
No, we don't need to increase DB version. It can be done even during DB work

samiuelson (approved):

tschellenbach:
Makes sense, nice find!

tschellenbach (dismissed):

tschellenbach (approved):

PR #801, opened 2020-11-03
CAS-334/fixing new messages button

leandroBorgesFerreira:
[CAS-334](https://stream-io.atlassian.net/browse/CAS-334)

### Description

This PR substitute #775. I decided to create another one since the approach was different.

Fixing the message button "New Messages". It was showing that new messages arrived even for old messages. Now it is fixed:

Nov-03-2020 21-10-32

### Checklist

- [x] Reviewers added

giphy (1)
_My reaction when the counting finally worked (that was a hard one o.O)_


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -379,7 +385,11 @@ internal class ChannelControllerImpl(
         val queryOnlineJob = if (domainImpl.isOnline()) { domainImpl.scope.async { runChannelQueryOnline(pagination) } } else { null }
         val localChannel = queryOfflineJob.await()
         if (localChannel != null) {
-            updateLiveDataFromLocalChannel(localChannel)
+            if (pagination.isLoadingMore) {
```
JcMinarro, line 388: Do we really need this extra property `isLoadingMore`?
Can't we do the same with the `pagination` property?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/request/QueryChannelPaginationRequest.kt
```diff
@@ -15,6 +15,8 @@ internal class QueryChannelPaginationRequest(var messageLimit: Int = 30) : Query
     var watcherLimit: Int = 30
     var watcherOffset: Int = 0
 
+    var isLoadingMore = false
```
JcMinarro, line 18: I am not sure we really need this property.
The name of this properthy is weird. What mean a pagination query that is not loading more? :thinking:

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/CollectionsUtils.kt
```diff
@@ -0,0 +1,34 @@
+package com.getstream.sdk.chat.utils
+
+// Use when it is known that the element is always close to the end
+internal fun <T> List<T>.reverseFind(predicate: (T) -> Boolean): T? {
```
JcMinarro, line 4: Isn't it the same like the Kotlin extension `lastOrNull()`?
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/last-or-null.html

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/CollectionsUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+        if (predicate(this[i])) {
+            return this[i]
+        }
+    }
+
+    return null
+}
+
+// Use when it is known that the element is always close to the end
+internal fun <T> List<T>.reverseIndexOf(element: T): Int {
```
JcMinarro, line 15: Same here:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/last-index-of.html

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/CollectionsUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+        if (this[i] == element) {
+            return i
+        }
+    }
+
+    return -1
+}
+
+// Use when it is known that the element is always close to the end
+internal fun <T> List<T>.reverseIndexOf(element: T, biPredicate: (T, T) -> Boolean): Int {
```
JcMinarro, line 26: Same here:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/index-of-last.html

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -593,13 +601,13 @@ public class MessageListView : ConstraintLayout {
...
             )
             lastMessageReadHandler.invoke()
-        } else if (listItem.loadingMore) {
-            // the load more behaviour is different, scroll positions starts out at 0
-            // to stay at the relative 0 we should go to 0 + size of new messages...
-
-            // TODO ???
-            val newPosition = layoutManager.findLastCompletelyVisibleItemPosition() + sizeGrewBy
-            layoutManager.scrollToPosition(newPosition)
+//        } else if (listItem.loadingMore) {
```
JcMinarro, line 604: Don't we need it anymore?
On that case we should remove this code

JcMinarro (changes requested):

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/CollectionsUtils.kt
zsmb13: Also the same as [`findLast`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/find-last.html), which has a better fitting name for here

leandroBorgesFerreira: True. I never checked the implementation of `findLast`... I though it did a search increasing the index instead of decreasing it.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
leandroBorgesFerreira: We never had the `listItem.loadingMore` as true before, so this never executed. When it started to execute, it was causing a scroll every time we load more data, which was not something to be expected. Let's remove it.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: I guess the this name is not good... I meant loading old messages, like "loading more" from the history. What about isLoadingFromHistory? We need this information because pagination is also used `loadOlderMessages` and `loadNewerMessages` and I need to differentiate then.

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/CollectionsUtils.kt
leandroBorgesFerreira: I this case I needed the `biPredicate`. Because I got some errors when simply looking for the element, the counter found messages that were much different than the true last one, the problem doesn't happens using getStableId... I can either use this function of override the equals from MessageListItem, what do you think?

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -404,8 +404,10 @@ public class MessageListView : ConstraintLayout {
         binding.chatMessagesRV.adapter = adapter
     }
 
-    private fun lastSeenMessagePosition(): Int = adapter.currentList.reverseIndexOf(lastSeenMessage) { a, b ->
-        a?.getStableId() == b?.getStableId()
+    private fun lastSeenMessagePosition(): Int {
+        return adapter.currentList.lastIndexOfBiPredicate(lastSeenMessage) { message1, message2 ->
+            message1?.getStableId() == message2?.getStableId()
+        }
```
zsmb13, line 410: Since `lastIndexOfBiPredicate` is always giving you its input parameter back as the second value in the predicate, isn't this the same as this?

```suggestion
        val lastMessage = lastSeenMessage
        return adapter.currentList.indexOfLast { message ->
            message?.getStableId() == lastMessage?.getStableId()
        }
```

(Local copy of `lastSeenMessage` so that it's the same value for all comparisons, and doesn't change while this function is running)

leandroBorgesFerreira: It is indeed. For some reason this didn't appear in my auto complete on android studio.

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/CollectionsUtils.kt
leandroBorgesFerreira: Nevermind, Martón presented a much better option.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: `Pagination` is an enum and it is different depend on `loadOlderMessagess` (`Pagination.LESS_THAN`) and `loadNewerMessages` (`Pagination.GREATER_THAN`)
That is because I don't think we need this extra property

leandroBorgesFerreira: Removed

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/request/QueryChannelPaginationRequest.kt
leandroBorgesFerreira: Removed

JcMinarro (approved):

PR #803, opened 2020-11-03
Callback threading

carterhudson:
### Description

Our custom call class was invoking callbacks without switching to the main thread. This meant if you used `.enqueue { ... }` and tried to touch the UI, your app would have a very bad time. This fix specifies a default main thread executor and uses it to invoke callbacks on the main thread. It also allows one to specify a callback executor when creating the call adapter factory.

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitCallAdapterFactory.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitCallAdapterFactory.kt
@@ -1,5 +1,7 @@
 package io.getstream.chat.android.client.api
 
+import android.os.Handler
+import android.os.Looper
 import io.getstream.chat.android.client.call.Call
 import io.getstream.chat.android.client.call.RetrofitCall
 import io.getstream.chat.android.client.parser.ChatParser
@@ -7,9 +9,11 @@ import retrofit2.CallAdapter
 import retrofit2.Retrofit
 import java.lang.reflect.ParameterizedType
 import java.lang.reflect.Type
+import java.util.concurrent.Executor
 
-internal class RetrofitCallAdapterFactory(
-    private val chatParser: ChatParser
+internal class RetrofitCallAdapterFactory private constructor(
+    private val chatParser: ChatParser,
+    private val callbackExecutor: Executor
 ) : CallAdapter.Factory() {
 
     override fun get(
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitCallAdapterFactory.kt
```diff
@@ -24,21 +28,31 @@ internal class RetrofitCallAdapterFactory(
...
-        return RetrofitCallAdapter<Any>(responseType, chatParser)
+        return RetrofitCallAdapter<Any>(responseType, chatParser, callbackExecutor)
     }
 
     companion object {
-        fun create(chatParser: ChatParser) = RetrofitCallAdapterFactory(chatParser)
+        val mainThreadExecutor by lazy {
+            Executor { r ->
+                Handler(Looper.getMainLooper()).post(r)
+            }
```
zsmb13, line 38: I'd bring this implementation closer to [what Retrofit has](https://github.com/square/retrofit/blob/7128bf34c649daca5977b1fbf93f0278fab4267c/retrofit/src/main/java/retrofit2/Platform.java#L138-L145), i.e. fetching the `Looper` and creating the `Handler` instance just once.

The `Handler` creation is relatively cheap (though it's still an object allocation), but grabbing the `Looper` is synchronized access of a static field.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitCallAdapterFactory.kt
```diff
@@ -24,21 +28,31 @@ internal class RetrofitCallAdapterFactory(
             throw IllegalArgumentException("Call return type must be parameterized as Call<Foo>")
         }
         val responseType: Type = getParameterUpperBound(0, returnType)
-        return RetrofitCallAdapter<Any>(responseType, chatParser)
+        return RetrofitCallAdapter<Any>(responseType, chatParser, callbackExecutor)
     }
 
     companion object {
-        fun create(chatParser: ChatParser) = RetrofitCallAdapterFactory(chatParser)
+        val mainThreadExecutor by lazy {
```
zsmb13, line 35: We probably don't need `lazy` here, we'll instantiate this immediately when our SDK is initialized anyway

zsmb13 (dismissed):
Solution looks correct, nicely done 👍

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitCallAdapterFactory.kt
JcMinarro: `Lopper.getMainLooper()` is Android SDK stuff that is not accesible in our tests. Because we are using this static field, some of our tests are using it and they crash if they try to get the Main Looper

JcMinarro (approved):

JcMinarro (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitCallAdapterFactory.kt
zsmb13: Ah, I see, we're mocking the Retrofit Api interface itself, and this class is being used with its real implementation in tests. Makes sense then 👍

zsmb13 (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitCallAdapterFactory.kt
JcMinarro: In the future maybe we should try to wrap this Looper or think into another "default argument implementation" into our `RetroSuccess` class

PR #804, opened 2020-11-04
CAS-180 Redesign avatar view

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-180

### Description

Redesigned `AvatarView` to support typical use case where avatar is used in the new design: user avatar, channel avatar, message avatar, etc. 

### Checklist

- [X] Reviewers added

photo_2020-11-04 02 58 01


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
```diff
@@ -0,0 +1,135 @@
...
+        return users.take(MAX_AVATAR_SECTIONS).map { createUserBitmap(it, style) }
+    }
+
+    private suspend fun createBitmapFromUrl(
+        url: String
+    ): Bitmap? = withContext(Dispatchers.IO) {
+        url.takeUnless { it.isBlank() }
+            ?.let {
+                (
+                    Coil.execute(
```
adasiewiczr, line 53: Our old ui-component module contains `ImageLoader` which is a wrapper for Coil. Can we reuse it here?
It would be great to not expose Coil API beyond `ImageLoader`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
```diff
@@ -0,0 +1,135 @@
...
+import io.getstream.chat.android.ui.utils.getIntArray
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+internal class AvatarBitmapFactory(private val context: Context) {
+    private val gradientBaseColors = context.getIntArray(R.array.avatar_gradient_colors)
+
+    internal suspend fun createUserBitmap(user: User, style: AvatarStyle): Bitmap {
+        return createBitmapFromUrl(user.image)
+            ?: createInitialsBitmap(style, LlcMigrationUtils.getInitials(user))
```
adasiewiczr, line 29: It would be great to get rid of LlcMigrationUtils at some point. What do you think about converting `getInitials` to extension function?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
```diff
@@ -0,0 +1,135 @@
...
+    private suspend fun createBitmapFromUrl(
+        url: String
+    ): Bitmap? = withContext(Dispatchers.IO) {
+        url.takeUnless { it.isBlank() }
+            ?.let {
+                (
+                    Coil.execute(
+                        ImageRequest.Builder(context)
+                            .data(it)
+                            .allowHardware(false)
```
adasiewiczr, line 56: Why are we disabling it?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
```diff
@@ -0,0 +1,67 @@
...
+            .style(
+                R.styleable.AvatarView_streamAvatarTextStyle,
+                Typeface.BOLD
+            )
+            .build()
+        onlineIndicatorEnabled = a.getBoolean(
+            R.styleable.AvatarView_streamAvatarOnlineIndicatorEnabled,
+            false
+        )
+    }
```
adasiewiczr, line 66: The TypedArray should be recycled when no longer needed. I think that we can wrap everything with `.use` to do that automatically

adasiewiczr (changes requested):
Good job! Looks great overall, I've added a few suggestions

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
```diff
@@ -0,0 +1,131 @@
...
+        borderPaint.strokeWidth = avatarStyle.avatarBorderWidth.toFloat()
+    }
+
+    private fun configureImageDrawable(generateAvatarDrawable: suspend () -> AvatarDrawable) {
+        GlobalScope.launch(Dispatchers.Main) {
+            layoutParams?.apply {
+                width = avatarStyle.avatarWidth
+                height = avatarStyle.avatarHeight
+            }?.let(::setLayoutParams)
+            setImageDrawable(generateAvatarDrawable())
```
zsmb13, line 73: - The first couple lines here are equivalent to an `updateLayoutParams` call from `Dependencies.androidxCoreKtx`
- Why do we need to update layout params before setting the new drawable?
- Is it okay that there might be some amount of time between `setLayoutParams` and `setImageDrawable` being called? `generateAvatarDrawable` is suspending here. If we want to call the two setters at the same time (for the same frame), we should create the Drawable first, and then call them.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
```diff
@@ -0,0 +1,131 @@
...
+            AvatarDrawable(listOfNotNull(bitmapFactory.createUserBitmap(user, avatarStyle)))
+        }
+    }
+
+    public fun toggleOnlineIndicatorVisibility(visible: Boolean) {
+        this.onlineIndicatorVisible = visible
+        invalidate()
+    }
+
+    public fun setStyle(avatarStyle: AvatarStyle) {
```
zsmb13, line 61: Does this have to be `public`, do we expect clients to create instances of `AvatarStyle` programatically? If not, that entire class could be `internal` too.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarDrawable.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    override fun getOpacity() = PixelFormat.TRANSLUCENT
+
+    override fun setColorFilter(colorFilter: ColorFilter?) {
+        paint.colorFilter = colorFilter
+    }
+
+    private fun reconfigureItems() {
+        avatarItems = when (avatarBitmaps.size) {
+            0 -> listOf()
```
zsmb13, line 44: 😉 

```suggestion
            0 -> emptyList()
```

stream-chat-android-ui-components/src/main/res/values/attrs.xml
```diff
@@ -0,0 +1,15 @@
...
+    <declare-styleable name="AvatarView">
+        <attr name="streamAvatarWidth" />
+        <attr name="streamAvatarHeight" />
+        <attr name="streamAvatarBorderWidth" />
+        <attr name="streamAvatarBorderColor" />
+        <attr name="streamAvatarTextSize" />
+        <attr name="streamAvatarTextColor" />
+        <attr name="streamAvatarTextFont" />
+        <attr name="streamAvatarTextFontAssets" />
+        <attr name="streamAvatarTextStyle" />
```
zsmb13, line 12: Shouldn't all of these specify the `format` required?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
JcMinarro: Good idea @adasiewiczr 
`LlcMigrationUtils` is a class that we should try to avoid, so if we are able to extract those method and refactor them, it would be the best solution

bychkovdmitry: Yep, also thought about that

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/Extensions.kt
```diff
@@ -0,0 +1,13 @@
+package io.getstream.chat.android.ui.utils
```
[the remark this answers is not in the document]
bychkovdmitry: Maybe ContextExt.kt?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
JcMinarro: Third-Party dependency should be wrapper to be able to replace them into one part of our code

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/Extensions.kt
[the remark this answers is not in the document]
zsmb13: `utils/extensions/Context.kt` would be my usual structuring for these kinds of extensions 😄 

Alternatively, `utils/ContextExtensions.kt`, no need to abbreviate in that case.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarDrawable.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    override fun getOpacity() = PixelFormat.TRANSLUCENT
+
+    override fun setColorFilter(colorFilter: ColorFilter?) {
+        paint.colorFilter = colorFilter
+    }
+
+    private fun reconfigureItems() {
+        avatarItems = when (avatarBitmaps.size) {
```
JcMinarro, line 43: Cant we use something like: `avatarBitmap.take(4)` and a generic method to configure them? :thinking: 
"Configure methods" seems really similar, I am not sure if we could extract some "common part" there.
Some ideas:
-. Bottom view is almost the same when 3 or 4 avatars
-. Top View is almost the same when 2 or 3 avatars
-. Top View Corner is always `0, 0`
-.View hieght is always `full-hieght` or `height/2`
-. View width is always `full-width` or `widht/2`

JcMinarro (commented):
Look great.
Some comments about something that we could improve there

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
bychkovdmitry: Using ImageLoader now

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
bychkovdmitry: I got some crashes as the bitmap was mutated which is not allowed in hardware mode. But now it works fine so I removed `allowHardware(false)`. Actually just utilized `ImageLoader.getBitmap()` where hardware mode is enabled.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/Utils.kt
```diff
@@ -0,0 +1,17 @@
+package io.getstream.chat.android.ui.utils
+
+import android.graphics.Color
+import kotlin.math.roundToInt
+
+internal fun adjustColorLBrightness(color: Int, factor: Float): Int {
```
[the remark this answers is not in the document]
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/Extensions.kt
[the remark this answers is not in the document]
bychkovdmitry: I've seen all the 3 naming options (`Context.kt`, `ContextExt.kt`, `ContextExtensions.kt`).  :) Lets use `utils/extensions/Context.kt`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
```diff
@@ -0,0 +1,67 @@
...
+import android.content.Context
+import android.content.res.TypedArray
+import android.graphics.Color
+import android.graphics.Typeface
+import android.util.AttributeSet
+import com.getstream.sdk.chat.style.TextStyle
+import io.getstream.chat.android.ui.R
+import io.getstream.chat.android.ui.utils.getDimension
+
+public class AvatarStyle(context: Context, attrs: AttributeSet?) {
```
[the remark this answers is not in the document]
bychkovdmitry: Are you sure?
```
public final TypedArray obtainStyledAttributes(@Nullable AttributeSet set,
            @NonNull @StyleableRes int[] attrs, @AttrRes int defStyleAttr,
            @StyleRes int defStyleRes)
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
```diff
@@ -0,0 +1,131 @@
...
+import android.graphics.RectF
+import android.util.AttributeSet
+import androidx.appcompat.widget.AppCompatImageView
+import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.User
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+
+public class AvatarView @JvmOverloads constructor(
```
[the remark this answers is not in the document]
bychkovdmitry: Used 3 separate constructors instead

stream-chat-android-ui-components/src/main/res/values/attrs.xml
bychkovdmitry: These attrs attrs are imported from `stream-chat-android` `attrs.xml`. Do you think it was better to duplicate them?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
bychkovdmitry: Used `ImageLoader`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
bychkovdmitry: I've just added add `a.recycle()` in the end. Don't want to wrap into`.use` as we will need to use `lateinit` fields, some of them like `Int` are considered primitives and need to be initialized with some default value.

zsmb13: Good call, better than going `lateinit` with everything. A real shame that AndroidX methods don't use Kotlin contracts yet, it would solve this issue really easily.

Although... We could add our own `use` method that supports contracts if we wanted - might be worth doing, since we'll do a bunch of attribute parsing like this. Would be pretty much like this:

```kotlin
internal inline fun TypedArray.use(block: (TypedArray) -> Unit) {
    contract { 
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }
    block(this)
    recycle()
}
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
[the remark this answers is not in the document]
zsmb13: My bad, I confused the parameters because of the mismatching names 🤦

stream-chat-android/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -19,8 +19,8 @@ import coil.transform.RoundedCornersTransformation
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 
-internal object ImageLoader {
-    suspend fun getBitmap(
+public object ImageLoader {
+    public suspend fun getBitmap(
```
zsmb13, line 23: Looks like this could be `@InternalStreamChatApi` so that we don't expose it completely (which we then have to opt into in our new module in the `build.gradle` config, like with some existing ones)

stream-chat-android-ui-components/src/main/res/values/attrs.xml
zsmb13: Ah, I didn't realize they were already defined elsewhere. Looks all good then 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
```diff
@@ -0,0 +1,131 @@
...
+                (height / 8f),
+                width / 8f,
+                onlineIndicatorOutlinePaint
+            )
+            canvas.drawCircle(
+                width - (width / 8f),
+                height / 8f,
+                width / 10f,
+                onlineIndicatorPaint
+            )
```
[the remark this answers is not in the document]
bychkovdmitry: Done

stream-chat-android/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarDrawable.kt
bychkovdmitry: Simplified the class a bit and removed duplication

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
zsmb13: Still looks to me like `AvatarStyle` should be `internal`.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarDrawable.kt
```diff
@@ -40,94 +39,93 @@ internal class AvatarDrawable(bitmaps: List<Bitmap>) : Drawable() {
...
+        }
     }
 
     private fun Bitmap.scaleCenterCrop(newWidth: Int, newHeight: Int): Bitmap {
         return ThumbnailUtils.extractThumbnail(this, newWidth, newHeight)
     }
 
     private data class AvatarItem(val bitmap: Bitmap, val position: Rect)
+
+    private enum class SectionType {
```
zsmb13, line 121: Really neat solution 👍

stream-chat-android-ui-components/build.gradle
```diff
@@ -72,7 +74,9 @@ dependencies {
     api project(":stream-chat-android")
 
     implementation Dependencies.kotlinStdLib
+    implementation Dependencies.coroutinesCore
     implementation Dependencies.androidxAppCompat
     implementation Dependencies.androidLegacySupport
     implementation Dependencies.androidxCoreKtx
+    implementation Dependencies.coil
```
adasiewiczr, line 81: Do we still need coil here?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
bychkovdmitry: Moved to extensions in LLC client

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
bychkovdmitry: Made AvatarStyle internal

stream-chat-android-ui-components/build.gradle
bychkovdmitry: Forgot to remove it, now it is removed

JcMinarro (approved):
Look Good!!!

adasiewiczr (approved):
Good job!

PR #810, opened 2020-11-04
[CAS-389] Replace LLC ChannelController by ChannelClient

adasiewiczr:
The ChannelController class already exists in offline module and it was easy to
use to wrong one by accident.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/controllers/ChannelControllerImpl.kt
```diff
@@ -72,6 +72,10 @@ import io.getstream.chat.android.client.utils.observable.ChatObservable
 import io.getstream.chat.android.client.utils.observable.Disposable
 import java.io.File
 
+@Deprecated(
+    message = "The ChannelControllerImpl has been replaced by ChannelClient",
+    replaceWith = ReplaceWith("ChannelClient")
+)
 internal class ChannelControllerImpl(
```
zsmb13, line 79: We can rename this class to `ChannelClient`, and have it implement the now deprecated `ChannelController` interface. Then we can remove that interface later on.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
```diff
@@ -0,0 +1,406 @@
...
+import io.getstream.chat.android.client.models.Member
+import io.getstream.chat.android.client.models.Message
+import io.getstream.chat.android.client.models.Mute
+import io.getstream.chat.android.client.models.Reaction
+import io.getstream.chat.android.client.utils.FilterObject
+import io.getstream.chat.android.client.utils.ProgressCallback
+import io.getstream.chat.android.client.utils.observable.Disposable
+import java.io.File
+
+public class ChannelClient(
```
JcMinarro, line 74: The constructor should be internal

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/controllers/ChannelControllerExtensions.kt
```diff
@@ -8,6 +8,10 @@ import kotlin.reflect.KClass
 /***
  * Subscribes to events of type [T] in the channel.
  */
+@Deprecated(
+    message = "The ChannelController has been replaced by ChannelClient",
+    replaceWith = ReplaceWith("ChannelClient::subscribeFor")
```
zsmb13, line 13: These are not valid `ReplaceWith` usages (the intention actions they generate break the code), we should remove them, as we can't provide good replacements for extensions being moved to a new package, this has to be done manually

zsmb13 (changes requested):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -720,6 +721,10 @@ public class ChatClient internal constructor(
      * @param channelType the channel type. ie messaging
      * @param channelId the channel id. ie 123
      */
+    @Deprecated(
+        message = "Use getChannelClient() instead",
+        replaceWith = ReplaceWith("this.getChannelClient")
+    )
```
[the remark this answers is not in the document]
adasiewiczr: I was thinking about it but it will be a breaking change for our customers.
Providing a new method will give them more time to migrate from `ChannelController` to `ChannelClient`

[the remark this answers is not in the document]
zsmb13: It's not breaking if `ChannelClient` is a `ChannelController`, is it?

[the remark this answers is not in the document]
adasiewiczr: ah I can see now. It makes sense when `ChannelClient` implements`ChannelController`. I'll do that

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClientExtensions.kt
```diff
@@ -1,4 +1,4 @@
-package io.getstream.chat.android.client.controllers
+package io.getstream.chat.android.client.channel
```
zsmb13, line 1: This is a breaking change for clients. It's very new and fancy API though, so it might be okay, if we document it in the changelog.

zsmb13 (approved):

PR #811, opened 2020-11-04
fix: CAS-358 fix channels being stuck in the unread state

### LINKED ISSUE #698: Latest message not updating
In the channel list, the latest message never updates in a same session. This is the latest 4.3.0-beta-6

out

carterhudson:
https://stream-io.atlassian.net/browse/CAS-358

### Description

Issues:
- Our logic for visually indicating as read in the channel list only renders as read if the last read date is newer than the last message date, but the API marks reads with the message's createdAt date. This makes the comparison EQUAL, especially from online query results.

- Additionally, if we got into the state where last read & last message dates are equal, we don't send markRead updates to the UI or the API because the logic indicates we only do that if the last read date is AFTER the last message date. So, we end up stuck in an unread state.

- We execute several concurrent data fetches for offline & online. The mark read usecase gets executed immediately when a message list view loads, even before the viewmodel has a chance to run a channel query for offline and online data. This means we optimistically update the UI, then send a message to the backend to mark a message as read, then execute an offline & online query. It's possible for these last read dates to differ / be old, especially since a MessageReadEvent will be received in the middle of these updates, triggering offline caching. The online job may also return outdated reads.

Changes:
- Since last lastRead == createdAt is expected in online responses, change the logic for rendering as read to >= instead of >.- Refactor updateReads to avoid pushing old read dates to the UI.
- Refactor markRead to check error states first and provide explicit logging.

### Checklist

- [x] Reviewers added

Fixes #698

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
@@ -238,24 +238,37 @@ internal class ChannelControllerImpl(
     }
 
     fun markRead(): Result<Boolean> {
-        if (!getConfig().isReadEvents) return Result(false, null)
+        if (!getConfig().isReadEvents) {
+            return Result(false, null)
+        }
+
         // throttle the mark read
         val messages = sortedMessages()
-        if (messages.isNotEmpty()) {
-            val last = messages.last()
-            val lastMessageDate = last.createdAt ?: last.createdLocallyAt
 
-            if (lastMarkReadEvent == null || lastMessageDate!!.after(lastMarkReadEvent)) {
-                lastMarkReadEvent = lastMessageDate
-                val userRead = ChannelUserRead(domainImpl.currentUser).apply {
-                    lastRead = last.createdAt ?: last.createdLocallyAt
-                }
-                _read.postValue(userRead)
-                client.markMessageRead(channelType, channelId, last.id).execute()
-                return Result(true, null)
-            }
+        if (messages.isEmpty()) {
+            logger.logI("No messages; nothing to mark read.")
+            return Result(false, null)
         }
-        return Result(false, null)
+
+        val last = messages.last()
+        val lastMessageDate = last.let { it.createdAt ?: it.createdLocallyAt }
+        val shouldNotUpdate =
+            lastMarkReadEvent != null && lastMessageDate?.after(lastMarkReadEvent) == false
+
+        if (shouldNotUpdate) {
+            logger.logI("Last message date [$lastMessageDate] is not after last read event [$lastMarkReadEvent]; no need to update.")
+            return Result(false, null)
+        }
+
+        lastMarkReadEvent = lastMessageDate
+
+        // optimistic UI update via LiveData
+        updateRead(ChannelUserRead(domainImpl.currentUser, lastMarkReadEvent))
+
+        // generates MessageReadEvent & triggers offline storage updates
+        client.markMessageRead(channelType, channelId, last.id).execute()
+
+        return Result(true, null)
     }
 
     private fun sortedMessages(): List<Message> {
@@ -381,6 +394,7 @@ internal class ChannelControllerImpl(
     suspend fun runChannelQuery(pagination: QueryChannelPaginationRequest): Result<Channel> {
         // first we load the data from room and update the messages and channel livedata
         val queryOfflineJob = domainImpl.scope.async { runChannelQueryOffline(pagination) }
+
         // start the online query before queryOfflineJob.await
         val queryOnlineJob = if (domainImpl.isOnline()) { domainImpl.scope.async { runChannelQueryOnline(pagination) } } else { null }
         val localChannel = queryOfflineJob.await()
@@ -391,8 +405,8 @@ internal class ChannelControllerImpl(
                 updateLiveDataFromLocalChannel(localChannel)
             }
         }
-        // if we are online we we run the actual API call
 
+        // if we are online we we run the actual API call
         return if (queryOnlineJob != null) {
             val response = queryOnlineJob.await()
             if (response.isSuccess) {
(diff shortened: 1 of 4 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -232,24 +232,37 @@ internal class ChannelControllerImpl(
...
-            }
+        if (messages.isEmpty()) {
+            logger.logI("No messages; nothing to mark read.")
+            return Result(false, null)
         }
-        return Result(false, null)
+
+        val last = messages.last()
+        val lastMessageDate = last.let { it.createdAt ?: it.createdLocallyAt }
+        val shouldNotUpdate =
```
tschellenbach, line 249: shouldMarkRead would be easier to read, the not X variable name is hard to read

carterhudson: Would you prefer `!shouldUpdate` vs `shouldNotUpdate`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -1017,20 +1044,39 @@ internal class ChannelControllerImpl(
...
-            _read.postValue(it)
-            println("updateReads for current user to ${it.lastRead} on channel $cid")
+        val previousUserIdToReadMap = _reads.value ?: mapOf()
+        val incomingUserIdToReadMap = reads.associateBy(ChannelUserRead::getUserId).toMutableMap()
+
+        /*
+        It's possible that the data coming back from the online channel query has a last read date that's before
+        what we've last pushed to the UI. We want to ignore this, as it will cause an unread state to show in the
+        channel list.
+         */
```
tschellenbach, line 1056: @carterhudson i don't get this, we don't show your own unread state...

tschellenbach: do we perhaps receive mark read events out of order? @tbarbugli is that possible?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -232,24 +232,37 @@ internal class ChannelControllerImpl(
...
-                return Result(true, null)
-            }
+        if (messages.isEmpty()) {
+            logger.logI("No messages; nothing to mark read.")
+            return Result(false, null)
         }
-        return Result(false, null)
+
+        val last = messages.last()
+        val lastMessageDate = last.let { it.createdAt ?: it.createdLocallyAt }
```
tschellenbach, line 248: @tbarbugli how does the backend API actually handle the mark read call? The usage of createdLocallyAt is a bit funky here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -515,7 +533,10 @@ internal class ChannelControllerImpl(
                 Result(processedMessage, null)
             } else {
 
-                logger.logE("Failed to send message with id ${newMessage.id} and text ${newMessage.text}: ${result.error()}", result.error())
+                logger.logE(
+                    "Failed to send message with id ${newMessage.id} and text ${newMessage.text}: ${result.error()}",
+                    result.error()
+                )
```
tschellenbach, line 539: @carterhudson which editor are you using and did you run `ktlint applyToIDEAProject`?

tschellenbach:
1. I think we need to understand how the backend handles mark read API calls for messages that aren't created yet. IE the mark read using message.createdLocallyAt

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
carterhudson: I'm using Android Studio with the Project code style selected and a pre-commit hook that requires ktlintFormat to be run.

tschellenbach (dismissed):
I think we still have edge cases with the createdLocallyAt, needs input from the backend team. 

It definitely seems better than before though, so lets go

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
carterhudson: Hmm It's possible I misunderstood this. I give this another look.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: `createdLocallyAt` would be used only for offline messages that has not been sent to the server (because no connection, for example)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: Which command do you use to format the code into your pre-commit hook?
Using `./gradlew ktlintFormat` doesn't format those lines in the current project.

JcMinarro (changes requested):
Look Good.
Could you update the Changelog?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
carterhudson: Formatting those lines is a result of the Project codestyle I'm using. It's possible mine differs if we're not tracking it in version control.

JcMinarro (approved):

PR #815, opened 2020-11-05
[CAS-371] Improve changing socket connection based on network state

adasiewiczr:
The old ChatSocketServiceImpl implementation didn't take NetworkConnectionState into account.
Socket connection status was updated only based on health monitor, so it wasn't immediately changed
after losing or recovering connection.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -121,7 +129,24 @@ internal class ChatSocketServiceImpl private constructor(
         this.endpoint = endpoint
         this.apiKey = apiKey
         this.user = user
-        setupSocket()
+
+        if (networkStateProvider.isConnected()) {
```
ogkuzmin, line 133: Shouldn't it be tested?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+            // Checks whether the network was switched or connection was lost
+            val isConnected = connectivityManager.activeNetworkInfo?.isConnected ?: false
+            if (!isConnected && this@NetworkStateProvider.isConnected) {
+                this@NetworkStateProvider.isConnected = false
+                listener?.onDisconnected()
+            }
+        }
+    }
+
+    private var isConnected: Boolean = false
```
ogkuzmin, line 27: I think it's better to use `@Volatile` or `AtomicBoolean` here instead of simple property.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -121,7 +129,24 @@ internal class ChatSocketServiceImpl private constructor(
         this.endpoint = endpoint
         this.apiKey = apiKey
         this.user = user
-        setupSocket()
+
```
bychkovdmitry, line 132: A small problem with this scenario:
1) Connection lost
2) Collapse app
3) Bring the app back to foreground
The state will be State.Disconnected instead of State.NetworkDisconnected, so the monitor will start running.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
adasiewiczr: Marking with @Volatile

adasiewiczr: Marking with Volatile

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
adasiewiczr: Nice catch!

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/socket/ChatSocketServiceImplTest.kt
```diff
@@ -0,0 +1,68 @@
...
+            tokenManager,
+            socketFactory,
+            eventsParser,
+            networkStateProvider
+        )
+        socketService.addListener(socketListener)
+    }
+
+    @Test
+    fun `Should start connecting to socket when connecting and network connectivity exists`() {
```
ogkuzmin, line 40: My personal recommendation. You can read about test namings [here](https://martinfowler.com/bliki/GivenWhenThen.html)  (Martin Fowler's blog).
Usually, if I want to use such naming and there is some condition(s) for the scenario I prefer the following scheme:
`Given arrange When act (Then) should assert`. For your case it is:
``` kotlin
@Test
    fun `Given network connectivity exists When connect Should move state to Connecting`()
```

ogkuzmin (approved):

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/socket/ChatSocketServiceImplTest.kt
adasiewiczr: I like to follow a shorter `Should` naming but I think we should discuss this topic and try to standardize our tests naming

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+    @Volatile
+    private var isConnected: Boolean = false
+    private var listener: NetworkStateListener? = null
+
+    fun isConnected(): Boolean {
+        isConnected = connectivityManager.activeNetworkInfo?.isConnected ?: false
+        return isConnected
+    }
+
+    fun subscribe(listener: NetworkStateListener) {
```
JcMinarro, line 36: `subscribe()` should add the listener to a list, instead of overriding the previous one

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -0,0 +1,51 @@
...
+        this.listener = listener
+        connectivityManager.registerNetworkCallback(NetworkRequest.Builder().build(), callback)
+    }
+
+    fun unsubscribe() {
+        listener = null
```
JcMinarro, line 42: `unsubscribe()` should receive the listener that wants stopping listen to, then remove from the list of listeners. If the list becomes empty, unregister the callback

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -0,0 +1,53 @@
...
+            if (!isConnected && this@NetworkStateProvider.isConnected) {
+                this@NetworkStateProvider.isConnected = false
+                listeners.forEach { it.onDisconnected() }
+            }
+        }
+    }
+
+    @Volatile
+    private var isConnected: Boolean = false
+    private var listeners = mutableListOf<NetworkStateListener>()
```
ogkuzmin, line 29: Seems it requires some synchronization (synchornizedCollections from JavaAPI for example)

ogkuzmin (changes requested):

ogkuzmin (approved):

JcMinarro (approved):

PR #817, opened 2020-11-05
[CAS-329] Basic message input view behaviour

leandroBorgesFerreira:
[CAS-329](https://stream-io.atlassian.net/browse/CAS-329)
### Description

Creating a new MessageInputView. 

It looks like this:

Nov-09-2020 10-37-03

### Checklist

- [ ] New code is covered by unit tests. No applicable for this view. 
- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
```diff
@@ -0,0 +1,165 @@
...
+                    R.styleable.StreamMessageInputView_inputTextSize,
+                    context.resources.getDimensionPixelSize(R.dimen.stream_text_size_input)
+                ).toFloat()
+
+            hint = typedArray.getText(R.styleable.StreamMessageInputView_inputHint)
+        }
+    }
+}
+
+private fun getColorList(normalColor: Int, pressedColor: Int, disabledColor: Int) = ColorStateList(
```
bychkovdmitry, line 158: Was it done intentionally out of the class? Maybe better to move to utils?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/TextInputHandler.kt
```diff
@@ -0,0 +1,32 @@
+package io.getstream.chat.android.ui.textinput
+
+import android.text.Editable
+import android.text.TextWatcher
+import android.widget.EditText
+
+public class TextInputHandler(
```
bychkovdmitry, line 7: Internal?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
```diff
@@ -0,0 +1,165 @@
...
+    private fun showSendMessageEnabled() {
+        binding.ivSendMessage.setImageResource(R.drawable.stream_ic_filled_up_arrow)
+    }
+
+    private fun hideSendMessageEnabled() {
+        binding.ivSendMessage.setImageResource(R.drawable.stream_ic_filled_right_arrow)
+    }
+
+    private fun configTextInput(typedArray: TypedArray) {
+        TextInputHandler(binding.etMessageTextInput, ::showSendMessageEnabled, ::hideSendMessageEnabled)
```
bychkovdmitry, line 130: Seems like this class is not doing much. What about replacing it with 
`binding.etMessageTextInput.addTextChangedListener(onTextChanged = {  })`

stream-chat-android-ui-components/src/main/res/layout/stream_message_input.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:background="@color/stream_white">
+
+    <LinearLayout
```
bychkovdmitry, line 9: `android:layout_width="0dp"` + `:app:layout_constraintXXX`

stream-chat-android-ui-components/src/main/res/layout/stream_message_input.xml
```diff
@@ -0,0 +1,62 @@
...
+            android:layout_marginEnd="8dp"
+            android:src="@drawable/stream_ic_lightning"
+            tools:src="@drawable/stream_ic_lightning" />
+
+        <androidx.appcompat.widget.AppCompatEditText
+            android:id="@+id/et_message_text_input"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:layout_weight="1"
```
bychkovdmitry, line 40: I think LinearLayout can be removed and weights can be implemented within `ConstraintLayout`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
```diff
@@ -0,0 +1,165 @@
...
+    ) {
+        init(context, attrs)
+    }
+
+    private lateinit var binding: StreamMessageInputBinding
+
+    private fun init(context: Context, attr: AttributeSet? = null) {
+        binding = StreamMessageInputBinding.inflate(LayoutInflater.from(context), this, true)
+
+        context.obtainStyledAttributes(attr, R.styleable.StreamMessageInputView).use { typedArray ->
```
bychkovdmitry, line 39: I like this class, looks really neat 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
```diff
@@ -0,0 +1,165 @@
...
+                    typedArray.getColor(
+                        R.styleable.StreamMessageInputView_lightningButtonIconPressedColor,
+                        ContextCompat.getColor(context, R.color.stream_gray_light)
+                    )
+                )
+            )
+
+            layoutParams.width = typedArray.getDimensionPixelSize(
+                R.styleable.StreamMessageInputView_lightningButtonWidth,
+                context.resources.getDimensionPixelSize(R.dimen.stream_attachment_button_width)
```
bychkovdmitry, line 111: Sometimes I start thinking that we need yet another module, something like `common-utils` because we seem to have extension function for stuff like that (`context.resources.getDimensionPixelSize`) everywhere :)

zsmb13: These things should go into `stream-chat-android` for now, and be marked with `@InternalStreamChatApi`.

Later on, we can pull out most of the non-UI parts of that module into a `stream-chat-android-ui-core` (name pending), and have both UI implementations depend on _that_ module.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/TextInputHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+    editText: EditText,
+    public var containsTextListener: () -> Unit = {},
+    public var emptyTextListener: () -> Unit = {}
+) {
```
bychkovdmitry, line 11: This class is not doing much ATM, but I suppose more logic will be added here like "commands" and "mentions", right?

stream-chat-android-ui-components/src/main/res/values/styles.xml
```diff
@@ -35,4 +35,50 @@
         <item name="streamAvatarTextSize">@dimen/stream_avatar_text_size_large</item>
         <item name="streamAvatarOnlineIndicatorEnabled">true</item>
     </style>
-</resources>
\ No newline at end of file
+
+    <!--MessageInputViewNew-->
+    <declare-styleable name="StreamMessageInputView">
```
zsmb13, line 40: All attributes here should be prefixed with `stream`, like the rest of them in the file.

stream-chat-android-ui-components/src/main/res/values/styles.xml
```diff
@@ -35,4 +35,50 @@
...
+        <attr name="attachmentButtonIconColor" format="color|reference" />
+        <attr name="attachmentButtonIconPressedColor" format="color|reference" />
+        <attr name="attachmentButtonIconDisabledColor" format="color|reference" />
+        <attr name="attachmentButtonSelectedIconColor" format="color|reference" />
+        <attr name="attachmentButtonWidth" format="dimension|reference" />
+        <attr name="attachmentButtonHeight" format="dimension|reference" />
+        <attr name="attachmentCloseButtonBackground" format="reference" />
+
+        <!--Lightning Button-->
+        <attr name="showLightningButton" format="boolean" />
```
zsmb13, line 53: We could name this something like `lightningButtonEnabled` to line up with the names of the related attributes, makes them easier to find

stream-chat-android-ui-components/src/main/res/values/styles.xml
bychkovdmitry, line 40: attrs.xml?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
```diff
@@ -0,0 +1,165 @@
...
+                    R.styleable.StreamMessageInputView_inputTextColor,
+                    ContextCompat.getColor(context, android.R.color.black)
+                )
+            )
+
+            setHintTextColor(
+                typedArray.getColor(
+                    R.styleable.StreamMessageInputView_inputHintTextColor,
+                    ContextCompat.getColor(context, android.R.color.darker_gray)
+                )
```
zsmb13, line 144: The input hint is a light blue colour in the gif in the PR description. Is that just out of date? Is it a bug due to using a system colour here somehow, instead of one of our own? (Same goes for all `android.R.color` resources in this file, are we sure we want to rely on those?)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
leandroBorgesFerreira: Makes sense.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/TextInputHandler.kt
leandroBorgesFerreira: Makes sensse

stream-chat-android-ui-components/src/main/res/layout/stream_message_input.xml
leandroBorgesFerreira: ConstraintLayout renders slower than LinearLayout, no? If we stack many of then, the view may time time to render itself.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/TextInputHandler.kt
leandroBorgesFerreira: That's what I believe will happen, yes. Maybe we want to show a preview of a reaction, like Slack does, or show available commands/users to be mention, like you said.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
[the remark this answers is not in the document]
leandroBorgesFerreira: 🤔  There's no side effects, just the construction of the `TextWatcher`. The idea is that this class handles different behaviours related to text, like commands, reactions... I can get big, so it I would prefer to keep it out of the `StreamMessageInputView` so the second doesn't get super big.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
leandroBorgesFerreira: You're right, I'll create the correct colour.

stream-chat-android-ui-components/src/main/res/layout/stream_message_input.xml
leandroBorgesFerreira: Oh, you mean use the root ContraintLayout, right?

leandroBorgesFerreira: You're right, makes more sense to use a group of views inside the ConstraintLayout than to use a LinearLayout.

stream-chat-android-ui-components/src/main/res/layout/stream_message_input.xml
leandroBorgesFerreira: We will probably add more views inside the ConstraintLayout, that's why I encapsulated this one inside the LinearLayout... anyways, it is better to just use the ConstraintLayout.

stream-chat-android/src/main/res/drawable/stream_ic_filled_up_arrow.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:pathData="M12,2C6.48,2 2,6.48 2,12C2,17.52 6.48,22 12,22C17.52,22 22,17.52 22,12C22,6.48 17.52,2 12,2ZM11,16V12H8L12,8L16,12H13V16H11Z"
+      android:fillColor="#006BFE"
```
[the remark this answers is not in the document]
leandroBorgesFerreira: 🤔  Not sure... this one I just picked from Frigma.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
leandroBorgesFerreira: Thanks!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
leandroBorgesFerreira: By the way, the blue hint in blue is just because I was testing the customisation of the hint when a created that gift. Good catch haha.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
leandroBorgesFerreira: Not sure if I get it. `getDimensionPixelSize` is from `android.content.res`. It is not our code

stream-chat-android/src/main/res/drawable/stream_ic_filled_up_arrow.xml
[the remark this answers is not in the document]
zsmb13: It's up to us to decide how we add them into the Android project, either already coloured or just plain black and then tinted in View attributes. On this other PR https://github.com/GetStream/stream-chat-android/pull/821#discussion_r519856150 we found that we can go black & tint when we use them for most cases (as long as they're a single colour icon).

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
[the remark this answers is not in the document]
zsmb13: What I meant was that we're constructing a class instance, which executes code in its constructor (which is a side effect of constructing the instance), and then we never use the instance itself. We're creating a class instance to just run a block of code.

To simplify, what I'm trying to point out is that doing something like this:

```kotlin
class A(val param: Int) {
    init {
        println(param)
    }
}

A(x)
```

Is equivalent to a simple function call:

```kotlin
fun a(param: Int) {
     println(param)
}

a(x)
```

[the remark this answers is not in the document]
zsmb13: If this `InputHandler` will eventually get larger, we can keep it around, but we could also move its `init` block into just a top level function, or maybe make it an `object` and call one of its functions, like so:

```kotlin
object A {
     fun init(param: Int) {
          println(param)
     }
}

A.init(param)
```

I just want to avoid using an `init` block and throwing away a class instance.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
zsmb13: I think he meant [this extension](https://github.com/GetStream/stream-chat-android/blob/1f16150c4025e5dc14ab7203f39d600998e92b55/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/common/Extensions.kt#L32-L34) specifically, and that we should have it available everywhere in the different modules, because we make these kinds of calls a lot. For now, it's fine to do it manually, we can revisit extracting this to simplify later.

stream-chat-android/src/main/res/drawable/stream_ic_filled_up_arrow.xml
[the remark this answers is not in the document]
leandroBorgesFerreira: I this case it does makes sense to be blue, as this is the arrow the indicates that the users can send messages.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
leandroBorgesFerreira: Oh I see.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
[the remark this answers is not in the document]
leandroBorgesFerreira: Ok, makes sense. Removing the logic from the constructor

bychkovdmitry (approved):
LGTM

zsmb13 (dismissed):

JcMinarro (approved):

PR #819, opened 2020-11-06
Fix/cas 396 fix channels from storage

ogkuzmin:
[CAS-396](https://stream-io.atlassian.net/browse/CAS-396?atlOrigin=eyJpIjoiYjBjMjQ4NDNiZWM2NDExNWI5MGFhYWE2OTZlNGJkNjciLCJwIjoiaiJ9)

### Description

Sometimes at start-up you could see a list of channels without names. It happened because of we received channels that were builded from some LiveData that had not been initialized yet.
So my fix is init live data's using suspend functions that update live data's values directly via Dispatchers.Main

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/LiveDataExtensions.kt
```diff
@@ -0,0 +1,10 @@
+package io.getstream.chat.android.livedata.extensions
+
+import androidx.lifecycle.MutableLiveData
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+internal suspend fun <T : Any?> MutableLiveData<T>.setOnUi(valueProvider: () -> T) =
+    withContext(Dispatchers.Main) {
+        value = valueProvider()
+    }
```
zsmb13, line 10: Is there a specific reason why we want the code inside `valueProvider` to be executed on the UI thread too? Seems like it'd be better to generate the value on a background thread if possible, and do just the setting on the UI thread.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
```diff
@@ -65,7 +65,7 @@ public interface ChannelController {
     public fun getMessage(messageId: String): Message?
     // This one needs to be public for flows such as running a message action
     // TODO: this is for handling actions, think we should expose this in a different way
-    public fun upsertMessage(message: Message)
+    public suspend fun upsertMessage(message: Message)
```
zsmb13, line 68: We should really address the comment above this method at this point. I don't think we have any suspending public API currently, and this change would introduce some. This would be uncallable from Java, for example.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
```diff
@@ -50,5 +50,5 @@ public interface QueryChannelsController {
      */
     public val loadingMore: LiveData<Boolean>
 
-    public fun addChannelIfFilterMatches(channel: Channel)
+    public suspend fun addChannelIfFilterMatches(channel: Channel)
```
zsmb13, line 53: Same as above, this is public suspending API now, not sure if this is a great idea.

We might be able to make this one internal though. Maybe it could be a method that's only on the `Impl` class but not on the interface?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -148,16 +148,12 @@ internal class QueryChannelsControllerImpl(
...
 
-        if (queryChannelsSpec != null) {
-            channels =
-                domainImpl.selectAndEnrichChannels(queryChannelsSpec.cids.toList(), pagination)
-            logger.logI("found ${channels.size} channels in offline storage")
+        return queryChannelsSpec?.let { query ->
+            domainImpl.selectAndEnrichChannels(query.cids.toList(), pagination).also {
+                logger.logI("found ${it.size} channels in offline storage")
+            }
         }
```
zsmb13, line 156:
```suggestion
        val query = domainImpl.repos.queryChannels.selectByFilterAndQuerySort(queryChannelsSpec)
            ?: return null

        return domainImpl.selectAndEnrichChannels(query.cids.toList(), pagination).also {
            logger.logI("found ${it.size} channels in offline storage")
        }
```

zsmb13 (commented):
Looks like a decent synchonization effort overall. When we do swap with `LiveData` for `Flow` everywhere, it'll be much, much easier to update values in them with a suspending call, but this is good prep in that direction for now.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
tschellenbach: The upsert was there to handle actions. I think Sam added support for that a while ago. As far as I know we can make it internal.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/LiveDataExtensions.kt
tschellenbach: Ah so you're waiting in the IO thread to write on the UI/Main thread. Well that solves the problem :)
Feels a bit hacky, but it works. 

A hard coded dispatcher is discouraged for testing purposes right?

tschellenbach (approved):
1. The sample app still works
2. Offline or online sorting is broken (you'll notice that the results load in a certain order, and after the API call completes they load in a different order). This can be fixed later though.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
ogkuzmin: Actually, we don't need to expose this interface. I'd like to hide it (but it's not a reason of this PR). This interface do nothing without `ChatDomain`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
ogkuzmin: Again same thing. Useless interface. I'd like to remove it and use it only internally

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/LiveDataExtensions.kt
ogkuzmin: Hardcoded dispatcher can be real problem in testing. It's not ideal solve for the problem. When we will have switched to flow it will be much easier to remove

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
ogkuzmin: But that comment makes me not so sure even though I don't see the usage of this in our codebase.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/LiveDataExtensions.kt
ogkuzmin: I think Marton is right that lambda can be heavy for UI. I changed it

zsmb13: We'll likely always swap the Main dispatcher in tests with `Dispatchers.setMain`, essentially in a static way, so it's _fine_ to use it directly. For every other dispatcher, they should really be injected so we can swap them out in tests.

We could even go as far as having a standardized `DispatcherProvider` interface that holds all the ones we use, and inject that everywhere. Then it would be easy to have a fake implementation of that for testing that we swap in.

---

Let's not forget the original comment up here though, that passing in the value `T` instead of the lambda that generates it would be a good idea here, so that the code in the lambda is not executed on the main thread 😄

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
zsmb13: cc @samiuelson, do you know if we can make this internal? 😄

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/LiveDataExtensions.kt
ogkuzmin: I like idea of DispatcherProvider. It gives a good level of flexibility

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
samiuelson: IMO, it should be internal. No need for actions/command use case to have it public

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/LiveDataExtensions.kt
zsmb13: Now that I think of it, proper injecting would be difficult in some places (e.g. individual functions that use Dispatchers, or small utility classes like `Call` implementations). I guess we can go the same way as the coroutine library works, have a static instance like `DispatcherProvider.instance().Main`, and then swap the Dispatchers inside inside there for tests.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
zsmb13: Sounds good, let's do that then 👍

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
zsmb13: Awesome, let's do that then 👍

zsmb13 (approved):

PR #821, opened 2020-11-06
CAS-397 Add design for mention and command lists

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-397

### Description

Designed suggestions list which supports `commands` and `mentions`.

### Checklist

- [X] I have signed the [Stream CLA]
- [X] Reviewers added


stream-chat-android-ui-components/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!--Command Suggestions-->
+    <string name="stream_commands_instant_commands">Instant Commands</string>
+    <string name="stream_command_command_template">/%s %s</string>
+
+    <!--Mention Suggestions-->
+    <string name="stream_mention_user_name_template">/%s</string>
```
JcMinarro, line 8: This string should start with `@` instead of `/`

JcMinarro (changes requested):

stream-chat-android-ui-components/src/main/res/values/strings.xml
bychkovdmitry: Done, changed the starting character to `@`

stream-chat-android-ui-components/src/main/res/drawable/stream_ic_command.xml
```diff
@@ -0,0 +1,9 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:fillColor="#006CFF"
```
zsmb13, line 7: Same question as here https://github.com/GetStream/stream-chat-android/pull/817#discussion_r519785377, should we stick with black by default, and then tint as needed?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListView.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun clearSuggestions() {
+        if (binding.suggestionsCardView.isVisible) {
+            commandsAdapter.submitList(emptyList())
+            mentionsAdapter.submitList(emptyList())
+            binding.suggestionsCardView.isVisible = false
+        }
+    }
+}
+
+public interface OnSuggestionClickListener {
```
zsmb13, line 69: I'd do more nesting on this PR. I'd move classes that belong directly to a single other class (e.g. this listener, the sealed class below, the ViewHolder class in the other file). More clearly indicates their intended scope. And if they're used in multiple places, they can live in their own file.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/MentionsAdapter.kt
```diff
@@ -0,0 +1,52 @@
...
+}
+
+internal class MentionViewHolder(
+    private val binding: StreamItemMentionBinding,
+    private val onUserClicked: (User) -> Unit
+) : RecyclerView.ViewHolder(binding.root) {
+
+    fun bind(user: User) {
+        binding.avatarView.setUserData(user)
+        binding.avatarView.toggleOnlineIndicatorVisibility(user.online)
```
zsmb13, line 44: Is this perhaps an indication that `AvatarView` should set the indicator visibility on its own, whenever `setUserData` is called?

stream-chat-android-ui-components/src/main/res/drawable/stream_ic_command.xml
bychkovdmitry: Not that simple. It is a bit verbose to set tint for `TextView` with drawable as you need to do it programmatically. And in the icon below with 2 colors it seems impossible to apply tint. Lets leave it as is.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListView.kt
bychkovdmitry: Might be a bit verbose at times but makes perfect sense

stream-chat-android-ui-components/src/main/res/drawable/stream_ic_command.xml
zsmb13: True, definitely can't do this for multi-colour icons. 

Doesn't [this `drawableTint` solution](https://stackoverflow.com/a/57899724/4465208) do what we need for single-colour icons though?

bychkovdmitry: Yep, something was wrong with my android studio. It works. Also added tint to `stream_ic_mention.xml`. Modified multi-color icon `stream_ic_command_circle.xml` to reference internally the blue color I extracted as color res.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListView.kt
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/MentionsAdapter.kt
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListView.kt
zsmb13: I usually add more specific imports for these, so that their usages are still as short as if they were top level classes.

zsmb13 (dismissed):
Good to go as far as I'm concerned 👍

zsmb13 (approved):

adasiewiczr (approved):

JcMinarro (approved):

PR #822, opened 2020-11-06
[CAS-398] Refactor BaseStyle class

samiuelson:
### Description

1. This PR separates and extracts data hold by `BaseStyle` into view-specific style classes: `AvatarStyle` and`ReadStateStyle`.
Thanks to that `BaseStyle` can be removed and other `...Style` classes are not forced to inherit from it. It makes view-styling code more readable and easier to maintain.

2. Convertes to kotlin: `ReadStateView`, `MessageListViewStyle`

### Checklist

- [x] Reviewers added


stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListViewStyle.kt
```diff
@@ -0,0 +1,618 @@
...
+import android.content.Context
+import android.graphics.Color
+import android.graphics.Typeface
+import android.util.AttributeSet
+import androidx.core.content.ContextCompat
+import com.getstream.sdk.chat.R
+import com.getstream.sdk.chat.style.TextStyle
+import com.getstream.sdk.chat.view.messages.AvatarStyle
+
+public class MessageListViewStyle(c: Context, attrs: AttributeSet?) {
```
samiuelson, line 12: Some of the `...Style` classes could not be made `internal` because either they are still used in Java views or are passed as args in public functions.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListViewStyle.kt
```diff
@@ -0,0 +1,574 @@
...
+            )
+        )
+
+        // Avatar
+        avatarStyle.avatarWidth = a.getDimensionPixelSize(
+            R.styleable.MessageListView_streamAvatarWidth,
+            res.getDimensionPixelSize(
+                R.dimen.stream_message_avatar_width
+            )
+        )
```
zsmb13, line 477: Could we move the logic for these nested styles into their own classes? E.g. something like `AvatarStyle.from(a)` here. I guess we'd also have to pass in the specific styleable IDs, so maybe `AvatarStyle.create(a, R.styleable.MLV_steramAvatarWidth, ...)`

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListViewStyle.kt
zsmb13: This also means that we definitely broke some of our public API with the changes here. If someone's using custom ViewHolders with either the attachment or message ViewHolder factory, this will mean syntax changes for them. While we probably don't want to list all the changes, we should add something about it into the changelog.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/ReadStateView.kt
```diff
@@ -0,0 +1,120 @@
...
+        init()
+    }
+
+    @SuppressLint("ResourceType")
+    private fun init() {
+        removeAllViews()
+
+        if (!readStateStyle.isReadStateEnabled || reads.isEmpty()) return
+
+        val chatFonts = instance().fonts
```
zsmb13, line 52: I know the Java-to-Kotlin converter does this, but let's not leave `instance()` calls in our code like this, super hard to discern

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/ReadStateView.kt
```diff
@@ -0,0 +1,120 @@
...
+        val user = reads[0].user
+        val image = user.getExtraValue("name", "")
```
zsmb13, line 54: This is available as `user.name` as an extension property. But also, we're assigning the name to an `image` variable, and loading that with Glide. Something seems fishy here.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/ReadStateView.kt
```diff
@@ -0,0 +1,120 @@
...
+            imageView.setPlaceholderTextSize(
+                TypedValue.COMPLEX_UNIT_PX,
+                readStateStyle.readStateText.size,
+                typeface
+            )
+        } else imageView.setPlaceholderTextSize(
+            TypedValue.COMPLEX_UNIT_PX,
+            readStateStyle.readStateText.size,
+            readStateStyle.readStateText.style
+        )
```
zsmb13, line 75: This `else` branch randomly doesn't have braces around it

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/ReadStateView.kt
```diff
@@ -0,0 +1,120 @@
...
+        if (!Utils.isSVGImage(image)) Glide.with(context)
+            .load(image) // TODO: llc check glide
+            // .load(StreamChat.instance().getUploadStorage().signGlideUrl(image))
+            .into(imageView)
```
zsmb13, line 79: This could also use surrounding braces for clarity

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/CommandMentionListItemAdapter.java
```diff
@@ -9,39 +9,34 @@
...
 import com.getstream.sdk.chat.view.MessageListViewStyle;
-import com.getstream.sdk.chat.view.messageinput.MessageInputStyle;
 
 import java.util.List;
 
-import io.getstream.chat.android.client.models.Command;
 import io.getstream.chat.android.client.models.User;
 
-public class CommandMentionListItemAdapter<STYLE extends BaseStyle> extends BaseAdapter {
+public class CommandMentionListItemAdapter<T> extends BaseAdapter {
```
zsmb13, line 19: This `T` is always `User`, isn't it?

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/CommandMentionListItemAdapter.java
```diff
@@ -53,57 +48,18 @@ public long getItemId(int position) {
     public View getView(int position, View convertView, ViewGroup parent) {
         ViewBinding binding;
         if (convertView == null) {
-            if (isCommand) {
-                convertView = layoutInflater.inflate(R.layout.stream_item_command, null);
-                binding = StreamItemCommandBinding.bind(convertView);
-            } else {
-                convertView = layoutInflater.inflate(R.layout.stream_item_mention, null);
-                binding = StreamItemMentionBinding.bind(convertView);
-            }
```
zsmb13, line 62: How come we removed one of these cases? Was it simply unused?

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/CommandMentionListItemAdapter.java
```diff
@@ -53,57 +48,18 @@ public long getItemId(int position) {
...
         } else {
             binding = (ViewBinding) convertView.getTag();
         }
-
-        if (isCommand) {
-            configCommands((StreamItemCommandBinding) binding, position);
-        } else {
-            configMentions((StreamItemMentionBinding) binding, position);
-        }
+        configMentions((StreamItemMentionBinding) binding, position);
```
zsmb13, line 57: If we always have just one kind of binding used here (as changed above), we can use its type for the local `binding` variable, and there's no need to cast here

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/ChannelHeaderViewStyle.kt
```diff
@@ -19,75 +21,138 @@ internal class ChannelHeaderViewStyle(context: Context?, attrs: AttributeSet?) :
...
             0
         )
 
         channelTitleText = TextStyle.Builder(attributes).apply {
-            size(R.styleable.ChannelHeaderView_streamChannelHeaderTitleTextSize, getDimension(R.dimen.stream_channel_header_initials))
-            color(R.styleable.ChannelHeaderView_streamChannelHeaderTitleTextColor, getColor(R.color.stream_channel_initials))
+            size(
+                R.styleable.ChannelHeaderView_streamChannelHeaderTitleTextSize,
+                res.getDimensionPixelSize(R.dimen.stream_channel_header_initials)
+            )
```
zsmb13, line 44: I know you didn't introduce the pattern, but we really shouldn't need to use `apply` here, the point of the `Builder` we have is to provide a nice API for chaining calls.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messageinput/MessageInputStyle.kt
```diff
@@ -10,24 +10,25 @@ import android.text.TextUtils
...
-
-    var inputText: TextStyle
-    val isShowAttachmentButton: Boolean
-    val inputButtonWidth: Int
-    val inputButtonHeight: Int
-    val attachmentButtonWidth: Int
-    val attachmentButtonHeight: Int
-    val attachmentCloseButtonBackground: Drawable
-    val inputSendAlsoToChannelTextColor: Int
+public class MessageInputStyle(private val context: Context, attrs: AttributeSet?) {
```
zsmb13, line 22: Why does this class have to be `public`?

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/CommandMentionListItemAdapter.java
samiuelson: That's right 👍🏼

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/CommandMentionListItemAdapter.java
samiuelson: It wasn't used anywhere. This adapter seems to be implemented to be used in a generic way in different places prematurely, and it wasn't.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messageinput/MessageInputStyle.kt
samiuelson: It doesn't have to, good catch!

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/ReadStateView.kt
samiuelson: Good catch! 🐡

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messages/AvatarStyle.kt
```diff
@@ -0,0 +1,110 @@
...
+            @DimenRes defaultValue: Int
+        ): Builder {
+            avatarBorderWidth = a.getDimensionPixelSize(
+                avatarBorderWidthStyleableId,
+                res.getDimensionPixelSize(defaultValue)
+            )
+            return this
+        }
+
+        public fun avatarBorderColoer(
```
zsmb13, line 62:
```suggestion
        public fun avatarBorderColor(
```

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messages/AvatarStyle.kt
```diff
@@ -0,0 +1,110 @@
...
+
+public data class AvatarStyle(
+    public val avatarWidth: Int = 0,
+    public val avatarHeight: Int = 0,
+    public val avatarBorderWidth: Int = 0,
+    public val avatarBorderColor: Int = 0,
+    public val avatarBackgroundColor: Int = 0,
+    public val avatarInitialText: TextStyle = TextStyle()
+) {
+    public class Builder(private val a: TypedArray, private val c: Context) {
```
zsmb13, line 20: The `AvatarStyle` class, or at least its nested `Builder` can likely be `internal`.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/channels/ChannelListViewStyle.kt
```diff
@@ -6,44 +6,38 @@ import android.graphics.Typeface
...
-    val lastMessageDateUnreadText: TextStyle
+    public val channelPreviewLayout: Int
+    public val channelTitleText: TextStyle
+    public val channelTitleUnreadText: TextStyle
+    public val lastMessage: TextStyle
+    public val lastMessageUnread: TextStyle
+    public val lastMessageDateText: TextStyle
+    public val lastMessageDateUnreadText: TextStyle
+    public var avatarStyle: AvatarStyle = AvatarStyle()
+    public val readStateStyle: ReadStateStyle = ReadStateStyle()
```
zsmb13, line 25: Looks to me like these should be `lateinit` and `private set`, or maybe just a `val` if we can refactor things around them a bit.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messages/AvatarStyle.kt
zsmb13, line 20: This immutable style class and `Builder` design is really neat, could we use this for `ReadStateStyle` too? Seems like the same kind of class. If all styles would be immutable, that'd be pretty perfect.

bychkovdmitry (approved):

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messages/AvatarStyle.kt
samiuelson: Yep, working on ReadStateStyle atm

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messages/AvatarStyle.kt
samiuelson: This is fixed

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Added new MessageInputView structure
 ## stream-chat-android
 - Remove `ChatClient` and `ChatDomain` as `ChatUI`'s dependencies
 - Replace Glide with Coil - SDK doesn't depend on Glide anymore.
+- Remove `BaseStyle` class and extract it's properties into `AvatarStyle` and `ReadStateStyle`.
```
zsmb13, line 10:
```suggestion
- Remove `BaseStyle` class and extract its properties into `AvatarStyle` and `ReadStateStyle`.
```

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/messages/AvatarStyle.kt
```diff
@@ -0,0 +1,105 @@
...
+import com.getstream.sdk.chat.R
+import com.getstream.sdk.chat.style.TextStyle
+
+public data class AvatarStyle(
+    public val avatarWidth: Int = 0,
+    public val avatarHeight: Int = 0,
+    public val avatarBorderWidth: Int = 0,
+    public val avatarBorderColor: Int = 0,
+    public val avatarBackgroundColor: Int = 0,
+    public val avatarInitialText: TextStyle = TextStyle()
```
zsmb13, line 18: Nit: I'd consider dropping the default values and making the ctor internal (same for `ReadStateStyle`)

zsmb13 (dismissed):
Looks great!

We have _so much_ style parsing code, it would be great to simplify it somehow, but it doesn't seem like we could get much more efficient with it, as we're almost down to just listing IDs of things to parse, and we'll always have to do that.

zsmb13 (approved):

zsmb13 (dismissed):

zsmb13 (approved):

PR #823, opened 2020-11-07
Testing

tschellenbach:
Old tests:

- leaking coroutines
- integration tests

New tests:

- clean coroutine test setup
- mock llc

The main improvements are in `BaseDomainTest2`, currently it replaced some of the slowest tests. The idea is to progressively replace more of the integration tests (llc is pretty reliable atm, we should have a few integration tests at the llc level to spot things like backwards incompatible API changes, serialization issues etc. no need to do that at the offline level though). Tests run in 25 seconds now


stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/BaseDomainTest2.kt
```diff
@@ -0,0 +1,258 @@
...
+            Truth.assertWithMessage(result.data().toString()).that(result.isError).isTrue()
+        }
+    }
+
+    @get:Rule
+    val rule = InstantTaskExecutorRule()
+
+    @Before
+    open fun setup() {
+        clientMock = createClientMock()
```
ogkuzmin, line 126: In test, it's also better to set the main Dispatcher. 
``` kotlin
...
  val testDispatcher = TestCoroutineDispatcher()
...
@Before
fun setup() {
  Dispatchers.setMain(testCoroutineDispatcher)
  ...
}

@After
fun tearDown() {
  Dispatchers.resetMain()
  ...
}

```

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/ChatDomainEventDomainImplTest.kt
```diff
@@ -3,194 +3,244 @@ package io.getstream.chat.android.livedata
...
 /**
  * Verify that all events correctly update state in room
  */
 @RunWith(AndroidJUnit4::class)
-internal class ChatDomainEventDomainImplTest : BaseConnectedIntegrationTest() {
+internal class ChatDomainEventDomainImplTest : BaseDomainTest2() {
 
     @Test
-    fun defaultConfig() = runBlocking(Dispatchers.IO) {
+    fun `verify that a missing channel config returns the default`() = testIOScope.runBlockingTest {
```
ogkuzmin, line 18: We can omit scope declaration in tests and use only `runBlockingTest`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -259,7 +257,13 @@ internal class ChatDomainImpl internal constructor(
         initClean()
     }
 
+    internal val job = SupervisorJob()
+    internal var scopeIO: CoroutineScope
+    internal var scopeMain: CoroutineScope
+
     init {
+        scopeIO = customIOScope ?: CoroutineScope(Dispatchers.IO + job)
```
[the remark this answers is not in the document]
tschellenbach: I cleaned this up to clearly have 2 dispatchers and 1 scope.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ThreadControllerImpl.kt
```diff
@@ -10,14 +10,15 @@ import io.getstream.chat.android.client.errors.ChatError
...
+import io.getstream.chat.android.livedata.ChatDomainImpl
 import kotlinx.coroutines.withContext
 import wasCreatedAfterOrAt
 
 internal class ThreadControllerImpl(
     override val threadId: String,
     val channelControllerImpl: ChannelControllerImpl,
-    val client: ChatClient
+    val client: ChatClient,
+    val domain: ChatDomainImpl,
```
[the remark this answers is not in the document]
tschellenbach: similar to above we pass domain to the controllers. we do this for all controllers. as the controllers become more complex this prevents you from having to add a larger list of params.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/BaseDomainTest2.kt
tschellenbach: yup, added it

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/ChatDomainEventDomainImplTest.kt
tschellenbach: If you have a look at `ExampleBaseDomainTest2` you'll notice that I use a testIOScope.launch, i'm not quite sure why that's needed, but if i change it into a simple launch or remove it the test fails with the error: "java.lang.IllegalStateException: This job has not completed yet"

image

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -147,6 +150,16 @@ public interface ChatDomain {
             return this
         }
 
+        internal fun withIODispatcher(scope: CoroutineDispatcher): Builder {
```
ogkuzmin, line 153: I saw in community very popular way to work with different dispatcher. The DispatcherProvider is provided via constructor and you take particular dispatcher from it. We already have it available via name `com.getstream.sdk.chat.infrastructure.DispatchersProvider`. And you can see how it is used in AttachmentsController.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -259,7 +260,16 @@ internal class ChatDomainImpl internal constructor(
...
 
+    internal val job = SupervisorJob()
+    internal var scope = CoroutineScope(job)
+    internal var scopeIO: CoroutineScope
+    internal var scopeMain: CoroutineScope
+
     init {
+        dispatcherIO = customDispatcherIO ?: Dispatchers.IO
+        dispatcherMain = customerDispatcherMain ?: Dispatchers.Main
+        scopeIO = CoroutineScope(job + dispatcherIO)
```
ogkuzmin, line 271: So do we still need these two scopes?

ogkuzmin (dismissed):
LGTM

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/ChatDomainEventDomainImplTest.kt
ogkuzmin: Indeed. I spent some time but couldn't reveal the cause. Seems some setup of BaseDomainTest2 runs async jobs that aren't completed when test gets executed. We could fix it later

zsmb13:
Updated this PR with a bunch of things, with a focus on simplifying things. As a rule of thumb, the simpler coroutine related code is, the more likely that it'll be correct. When you find yourself nesting scopes and contexts and dispatchers left and right, something is probably being done incorrectly.

So the changes:

- A centralized `DispatcherProvider` object in the client library that provides a `Main` and an `IO` dispatcher to use. All of our code should use this instead of using `Dispatchers` directly - we can enforce this with a lint rule at some point if we want to, but we should also be able to catch it on code reviews.
- A `TestCoroutineRule` and a `TestCoroutineExtension` (duplicated due to our module setups) for JUnit4 and JUnit5 based tests respectively, which both swap the `Main` and `IO` dispatchers in `DispatcherProvider` for the test implementation as needed. All tests now use either the rule or the extension for coroutine mocking.
- Removed lots of specified scopes and contexts that were unnecessary.

In general, our tests should use `coroutineTest.scope.runBlockingTest { ... }`, that's the best approach to take. However, due to Room's peculiar threading needs, its transactions will have to execute on a separate thread from the test thread, which would make the assertions for all jobs being done in `runBlockingTest` fail. Therefore we can fall back on `runBlocking` in those cases, which makes fewer coroutine correctness checks, but allows Room transactions to work.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/utils/TestCoroutineRule.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package io.getstream.chat.android.livedata.utils
+
+import io.getstream.chat.android.client.internal.DispatcherProvider
+import kotlinx.coroutines.test.TestCoroutineDispatcher
+import kotlinx.coroutines.test.TestCoroutineScope
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+internal class TestCoroutineRule : TestWatcher() {
```
tschellenbach, line 11: nice how it's a rule

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
tschellenbach: this was just a shortcut for running on main, it's not needed

ogkuzmin (approved):
The very good thing with TestCoroutineRule!

zsmb13 (approved):
I think this is good to go

PR #824, opened 2020-11-08
StateFlow for internal state - Merge after 4.4.2

tschellenbach:
We have a set of bugs on the offline support SDK that come from reading livedata.value. There are a few ways that we can solve this:

- run things on the IO thread with livedata
- add data structure on top of livedata
- use flow internally

Flow seems like the cleanest/most robust solution. The sample app works on this PR and tests pass, here are some things I like:

- Flow handles reading and writing from the flow much better than livedata, this prevents bugs
- There is no need for setValue vs postValue which simplifies the code
- Distinct until changed is enabled by default so we can remove checks that only write on livedata if it changed ( https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/distinct-until-changed.html)

To be sure we don't miss any issues i think we should merge this after 4.4.2 goes live.

The idea is to start with a minimal pr/update to adopt flow. In the future we can move more logic to flow.


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -83,53 +85,56 @@ internal class ChannelControllerImpl(
...
+        return Transformations.map(messages.asLiveData()) { messageMap ->
+            if (messageMap == null) { null } else {
+                messageMap.values
+                    .asSequence()
+                    .filter { it.parentId == null || it.showInChannel }
+                    .filter { hideMessagesBefore == null || it.wasCreatedAfter(hideMessagesBefore) }
+                    .sortedBy { it.createdAt ?: it.createdLocallyAt }
+                    .toList()
+                    .map { it.copy() }
+            }
```
ogkuzmin, line 124: The advantage of flow usage is flow operators. I'd suggest the next code for this snippet:

``` kotlin
    override val messages: LiveData<List<Message>> = _messages.messagesTransformation().asLiveData()

    override val oldMessages: LiveData<List<Message>> =
        _oldMessages.messagesTransformation().asLiveData()

    private fun Flow<Map<String, Message>?>.messagesTransformation(): Flow<List<Message>> =
        map { it?.values ?: emptyList() }
            .map { it.asSequence() }
            .map { messages -> messages.filter { it.parentId == null || it.showInChannel } }
            .map { messages ->
                messages.filter {
                    hideMessagesBefore == null || it.wasCreatedAfter(hideMessagesBefore)
                }
            }
            .map { messages -> messages.sortedBy { it.createdAt ?: it.createdLocallyAt } }
            .map { it.toList() }
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -83,53 +85,56 @@ internal class ChannelControllerImpl(
...
-    private fun messagesTransformation(messages: MutableLiveData<Map<String, Message>>): LiveData<List<Message>> {
-        return Transformations.map(messages) { messageMap ->
-            messageMap.values
-                .asSequence()
-                .filter { it.parentId == null || it.showInChannel }
-                .filter { hideMessagesBefore == null || it.wasCreatedAfter(hideMessagesBefore) }
-                .sortedBy { it.createdAt ?: it.createdLocallyAt }
-                .toList()
-                .map { it.copy() }
+    private fun messagesTransformation(messages: MutableStateFlow<Map<String, Message>?>): LiveData<List<Message>> {
```
ogkuzmin, line 114: Also is it good idea to use nullable type here? If so why messages and old messages live datas are non-nullable?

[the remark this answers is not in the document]
tschellenbach: From what I understand so far, LiveData always allows null as a value. So if you tell LiveData the type is String it's actually String?. See this blogpost for instance: https://proandroiddev.com/improving-livedata-nullability-in-kotlin-45751a2bafb7

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
tschellenbach: Trying to keep the PR small, but yes that approach is easier/cleaner than livedata's transformation step

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChannelData.kt
```diff
@@ -14,28 +14,30 @@ import java.util.Date
...
+    var createdAt: Date? = null,
     /** when the channel was updated */
-    var updatedAt: Date? = null
+    var updatedAt: Date? = null,
     /** when the channel was deleted */
-    var deletedAt: Date? = null
+    var deletedAt: Date? = null,
     /** all the custom data provided for this channel */
     var extraData: MutableMap<String, Any> = mutableMapOf()
+) {
```
tschellenbach, line 35: this data class had the important fields specified in the body. the comparison operators don't work on the body fields (only the arguments). because the objects compared as equal, and stateflow filters out data that's the same, updates on channel data didn't work.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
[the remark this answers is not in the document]
zsmb13: Solved this and removed a bunch of nullability in general in https://github.com/GetStream/stream-chat-android/pull/824/commits/e00f2c91a64e68decfcf8c5372cffc6063a1cf8c

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
zsmb13: I strongly agree, we should get rid of `LiveData` transformations as soon as we can, it's almost a shame that they were ever provided as a possible API to use, pretty misleading about what `LiveData` is supposed to be used for. I removed them in https://github.com/GetStream/stream-chat-android/pull/824/commits/8cd9b08fd4b6cd22b223f5a128802e6cfb85800b.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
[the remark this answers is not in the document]
tschellenbach: So how do you now distinguish between messages being:

- unknown
- empty
- a list of messages

You need some sort of state for that first one (previously represented as null)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
tschellenbach: We could also remove more livedata in the QueryChannel and thread controllers if we want to go for a full replace.

zsmb13: We should eventually, but if we don't need it for bugfixing elsewhere (like we did here), then it's just a nice to have thing, which can wait for now.

tschellenbach:
Making this a draft till I have the chance to refactor the ThreadController as well

buildSrc/src/main/kotlin/com/getstream/sdk/chat/Dependencies.kt
```diff
@@ -53,6 +53,7 @@ private const val MATERIAL_COMPONENTS_VERSION = "1.2.1"
 private const val MARKWON_VERSION = "4.6.0"
 private const val MOCKITO_KOTLIN_VERSION = "2.2.0"
 private const val MOCKITO_VERSION = "3.5.15"
+private const val MOCKK_VERSION = "1.10.2"
```
carterhudson, line 56: +1 for MockK

carterhudson (dismissed):
LGTM.

buildSrc/src/main/kotlin/com/getstream/sdk/chat/Dependencies.kt
zsmb13: I'm gonna be a killjoy here: in this form, -1 for MockK.

We have a tendency of adding dependencies on a whim, often ending up with two-three that perform the same task (look at our assertion libraries, or our image loaders). While the ones in tests at least don't degrade the user experience, they still introduce extra complexity, and require devs to jump back and forth between using different libraries as they touch different tests.

We should either:
- Stick to the mocking solution we already use everywhere (Mockito), if it works well enough for us.
- Make a concerted effort of migrating from the old solution to the new.

If we just add this for one test, I guarantee that in 6 months we'll be looking at a test suite that uses three different mocking libraries seemingly randomly.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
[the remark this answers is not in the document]
zsmb13: I tried looking up the original usages of these `LiveData`s, but I didn't find anywhere in the codebase where we handle a `null` value being emitted from them. Looks to me like they get to `MessageListItemLiveData` and we assume them to be non-null there.

buildSrc/src/main/kotlin/com/getstream/sdk/chat/Dependencies.kt
ogkuzmin: Agree with Marton. And even more MockK lets you to mock bad things like static dependencies or object dependencies. It tends to make us avoiding  of refactoring of such bad code coupling.

ogkuzmin (dismissed):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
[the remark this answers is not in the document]
tschellenbach: LiveData always emits null, so if we don't handle it that can be a problem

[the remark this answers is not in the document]
zsmb13: It can technically emit it, but only if we assign that value to it. Looks to me like we're okay with nulls in the current state of the PR, the changes to `Flow` should make things type safe.

buildSrc/src/main/kotlin/com/getstream/sdk/chat/Dependencies.kt
tschellenbach: That does't seem pragmatic, if you want to mock this with Mockito you have to add getter functions for those properties just for the sake of mocking. I don't feel convinced yet Mock is good enough to justify replacing everything, but it get some things like property mocking right.

zsmb13: I'm not sure what you mean about property mocking, this works with Mockito:

```kotlin
whenever(channelControllerMock.unfilteredMessages) doReturn channelMessages
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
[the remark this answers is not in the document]
tschellenbach: Ok it needs a bit of integration/testing with the UI side of things to make sure that we handle the "messages unknown" state well.

buildSrc/src/main/kotlin/com/getstream/sdk/chat/Dependencies.kt
tschellenbach: That breaks for me when an error talking about mockito limitations. I'm not quite sure why

ogkuzmin (approved):

buildSrc/src/main/kotlin/com/getstream/sdk/chat/Dependencies.kt
zsmb13: FYI, PR for removing it was here https://github.com/GetStream/stream-chat-android/pull/872

PR #827, opened 2020-11-09
Fix navigation in sample apps.

adasiewiczr:
Add startup fragment to prevent multiple ChatClient initializations.
Previous implementation allowed creating multiple ChatClient instances without deinitializing old ones.
This was happening when user was on channels fragment, clicked the back button and reopened the app which
was kept in memory.
That was causing multiple socket connections to be opened because socket connection is automatically restored
after resuming the app.
Moreover, back button wasn't working correctly after logging out (Activity wasn't destroyed).

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/startup/StartupFragment.kt
+++ b/stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/startup/StartupFragment.kt
@@ -3,16 +3,19 @@ package io.getstream.chat.sample.feature.startup
 import android.os.Bundle
 import androidx.fragment.app.Fragment
 import androidx.navigation.fragment.findNavController
-import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.sample.R
+import io.getstream.chat.sample.application.ChatInitializer
 import io.getstream.chat.sample.common.navigateSafely
+import org.koin.android.ext.android.inject
 
 class StartupFragment : Fragment() {
+
+    private val chatInitializer: ChatInitializer by inject()
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         findNavController().navigateSafely(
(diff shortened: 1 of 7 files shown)


stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/user_login/UserLoginFragment.kt
```diff
@@ -77,6 +78,17 @@ class UserLoginFragment : Fragment() {
                 viewModel.targetChannelDataReceived(cid)
             }
         }
+
+        activity?.apply {
```
zsmb13, line 82: Nit: we could use [`requireActivity`](https://developer.android.com/reference/kotlin/androidx/fragment/app/Fragment#requireActivity()) here in a few places

stream-chat-android-sample/src/main/kotlin/io/getstream/chat/sample/feature/startup/StartupFragment.kt
```diff
@@ -0,0 +1,22 @@
...
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.sample.R
+import io.getstream.chat.sample.common.navigateSafely
+
+class StartupFragment : Fragment() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        findNavController().navigateSafely(
+            if (ChatClient.isInitialized && ChatClient.instance().getCurrentUser() != null) {
```
zsmb13, line 15: How come we didn't add this logic to `ChatInitializer` in this sample app, but we did in the other?

adasiewiczr: 🙈 nice catch, forgot to refactor that

zsmb13 (approved):

PR #829, opened 2020-11-09
Refactor/chat ui

JcMinarro:
### Description
Our `ChatUI` is totally independent of our `ChatClient` and `ChatDomain` so doesn't make sense to keep an strong dependency on them anymore. The builder has been refactored to follow the KotlinStyle and `ChatNavigator` has been refactored as well

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
+++ b/stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
@@ -172,13 +172,14 @@ class ChatImpl implements Chat {
                 .enableBackgroundSync()
                 .notificationConfig(chatNotificationHandler.getConfig()).build();
 
-        // create a copy ChatUX implementation for backward compat
-        ChatUI.Builder uxBuilder = new ChatUI.Builder(context).withFonts(chatFonts).withMarkdown(markdown).withUrlSigner(urlSigner).withStrings(getStrings());
+        // create a copy ChatUI implementation for backward compat
+        ChatUI.Builder uiBuilder =
+                new ChatUI.Builder(context).withFonts(chatFonts).withMarkdown(markdown).withUrlSigner(urlSigner).withStrings(getStrings());
 
         if (navigationHandler != null) {
-            uxBuilder.withNavigationHandler(navigationHandler);
+            uiBuilder.withNavigationHandler(navigationHandler);
         }
-        uxBuilder.build();
+        uiBuilder.build();
 
         client.setUser(user, userToken, new InitConnectionListener() {
             @Override
--- a/stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
+++ b/stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
@@ -172,7 +172,7 @@ class ChatImpl implements Chat {
                 .notificationConfig(chatNotificationHandler.getConfig()).build();
 
         // create a copy ChatUX implementation for backward compat
-        ChatUI.Builder uxBuilder = new ChatUI.Builder(client(), domain, context).withFonts(chatFonts).withMarkdown(markdown).withUrlSigner(urlSigner).withStrings(getStrings());
+        ChatUI.Builder uxBuilder = new ChatUI.Builder(context).withFonts(chatFonts).withMarkdown(markdown).withUrlSigner(urlSigner).withStrings(getStrings());
 
         if (navigationHandler != null) {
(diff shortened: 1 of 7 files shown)


stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -172,7 +173,7 @@ public void setUser(@NotNull User user,
                 .notificationConfig(chatNotificationHandler.getConfig()).build();
 
         // create a copy ChatUX implementation for backward compat
```
adasiewiczr, line 175:
```suggestion
        // create a copy ChatUI implementation for backward compat
```

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatImpl.java
```diff
@@ -172,7 +173,7 @@ public void setUser(@NotNull User user,
...
-        ChatUI.Builder uxBuilder = new ChatUI.Builder(client(), domain, context).withFonts(chatFonts).withMarkdown(markdown).withUrlSigner(urlSigner).withStrings(getStrings());
+        ChatUI.Builder uxBuilder = new ChatUI.Builder(context).withFonts(chatFonts).withMarkdown(markdown).withUrlSigner(urlSigner).withStrings(getStrings());
```
adasiewiczr, line 176: It would be nice to rename it to `uiBuilder`

ogkuzmin (dismissed):

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUI.kt
```diff
@@ -29,66 +25,57 @@ import io.getstream.chat.android.livedata.ChatDomain
...
-        if (navigationHandler != null) {
-            navigator.setHandler(navigationHandler)
-        }
-    }
-
-    public data class Builder(
-        private var client: ChatClient,
-        private var chatDomain: ChatDomain,
-        private var appContext: Context
-    ) {
```
zsmb13, line 55: Changing this constructor signature is a significant breaking change for anyone who's initializing our library with any of the recent versions. I think it would be a very good idea to keep around a 3-param deprecated constructor and provide a migration path to this new API.

Or, at the very least, add more details to the release notes. But I'm really not a fan of breaking the library init for _all_ clients.

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUI.kt
```diff
@@ -29,66 +25,57 @@ import io.getstream.chat.android.livedata.ChatDomain
...
-    public val navigationHandler: ChatNavigationHandler? = null,
+    public val navigator: ChatNavigator,
     public val markdown: ChatMarkdown,
     public val urlSigner: UrlSigner
 ) {
     public val version: String
         get() = BuildConfig.BUILD_TYPE + ":" + BuildConfig.STREAM_CHAT_UI_VERSION
 
-    public val navigator: ChatNavigator = ChatNavigatorImpl()
+    public class Builder(private val appContext: Context) {
```
zsmb13, line 37: Happy to see that this is no longer a `data class` 👍

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUI.kt
JcMinarro: Because it is a builder, I can keep the old constructor, doing nothing with `ChatClient` and `ChatDomain` params ;)

zsmb13: Yup, that's what I was thinking. Plus we can add a nice `@Deprecated` annotation and `ReplaceWith("Builder(appContext)")`

zsmb13 (approved):

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUI.kt
JcMinarro: That was what I was thinking about ;)
DONE

adasiewiczr:
Would be nice to update tutorial 🙈

zsmb13:
We should wait for a release of this before we do that, but it's worth creating a ticket for already

JcMinarro:
Because we keep the previous constructor, it shouldn't be a breacking change on the tutorial.
BTW, I will update it when the new version is released.
Thanks @adasiewiczr

PR #830, opened 2020-11-09
[CAS-301] Replace Glide with Coil

adasiewiczr:
### Checklist

- [x] Reviewers added


stream-chat-android/build.gradle
```diff
@@ -102,12 +102,6 @@ dependencies {
 
     implementation Dependencies.gson
 
-    implementation Dependencies.glideRedirect
-    implementation(Dependencies.glide) {
-        exclude group: "com.android.support"
-    }
-    kapt Dependencies.glideCompiler
```
zsmb13, line 109: This means we can also disable `kapt` in this module

stream-chat-android/src/main/java/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -41,13 +43,28 @@ public object ImageLoader {
...
+
+    @JvmStatic
+    @JvmOverloads
+    public fun ImageView.load(
         uri: String?,
-        @DrawableRes placeholderResId: Int?,
+        @DrawableRes placeholderResId: Int? = null,
+        onStart: () -> Unit = {},
+        onComplete: () -> Unit = {},
+    ): Unit = loadAny(uri, placeholderResId, onStart, onComplete)
```
zsmb13, line 59: Looks like we should be able to merge these two methods into just one

adasiewiczr: I didn't want to expose `loadAny` outside this class. IMO it is better to use type-safe versions

zsmb13: Nevermind, didn't realize one of these took a `Uri` and the other a `String` as its parameter 🤦

zsmb13 (approved):

JcMinarro (commented):
Great!!!
We should add a comment into our Changelog that our SDK doesn't depend anymore on Glide :tada:

tschellenbach:
This is great 👍

bychkovdmitry (dismissed):

zsmb13 (approved):

PR #831, opened 2020-11-09
Chat client state refactoring

ogkuzmin:
[CAS-391](https://stream-io.atlassian.net/browse/CAS-391?atlOrigin=eyJpIjoiM2ZiYTk4ZWE1MmNjNDk1NGI3MGY5MWE2NTczYTk2ZjUiLCJwIjoiaiJ9)

### Description

Refactoring of possible states for ChatClient. 
Changes are the first part for feature "postponing query channels if set user is in progress"

Main changes:
1) New core java lib module
2) FiniteStateMachine in core module. Could be integrated in any required module.
3) ChatClientService as a logic for ClientState handling.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -274,8 +272,7 @@ public class ChatClient internal constructor(
     }
 
     public fun reconnectSocket() {
-        val user = state.user
-        if (user != null) socket.connect(user)
+        runCatching { clientStateService.state.userOrError().let(socket::connect) }
```
ogkuzmin, line 275: Maybe it's good to expose exception here

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientState.kt
```diff
@@ -0,0 +1,70 @@
...
+
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.client.utils.State
+
+internal sealed class ClientState : State {
+    object Idle : ClientState()
+    sealed class AnonymousUserState : ClientState() {
+        sealed class AnonymousUserPending : AnonymousUserState() {
+            object AnonymousPendingWithoutToken : AnonymousUserPending()
+            class AnonymousPendingWithToken(val token: String) : AnonymousUserPending()
```
zsmb13, line 11: It would simplify readability a lot if we used shorter names for all these states. For example, right now one of the inner ones here is called `ClientState.AnonymousUserState.AnonymousUserPending.AnonymousPendingWithToken`. This repeats a lot of information in the naming of the enclosing state. Could just be `ClientState.Anonymous.Pending.WithToken`, for example.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -791,7 +793,9 @@ public class ChatClient internal constructor(
     private fun getTokenAndConnect(connect: () -> Unit) {
         tokenManager.loadAsync {
             if (it.isSuccess) {
-                state.token = it.data()
+                clientStateService.onTokenReceived(it.data())
+            } else {
+                clientStateService.onTokenReceived("")
```
zsmb13, line 798: We have such strong typing everywhere else, can we avoid using an empty string to indicate the failure? Could we have a separate method, something like `clientStateService.onTokenLoadFailed()`?

Another question: do we handle the special case of the empty string somewhere later on in our code?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientStateService.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    fun onDisconnectRequested() {
+        stateMachine.sendEvent(ChatClientEvent.DisconnectRequestedEvent)
+    }
+
+    private val stateMachine: FiniteStateMachine<ClientState, ChatClientEvent> by lazy {
+        FiniteStateMachine {
+            initialState(ClientState.Idle)
+
+            defaultHandler { state, event -> state.inappropriateStateError("handling event $event") }
```
zsmb13, line 36: Looks like `inappropriateStateError` should just take `event` as its parameter.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientStateService.kt
```diff
@@ -0,0 +1,129 @@
...
+                        event.connectionId,
+                        event.user,
+                        state.token
+                    )
+                }
+            }
+        }
+    }
+
+    internal val state = stateMachine.state
```
zsmb13, line 119: This is a one-time assignment of this value, and not a getter that's executed every time - seems like a bug.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/FiniteStateMachine.kt
```diff
@@ -0,0 +1,96 @@
+package io.getstream.chat.android.client.utils
+
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlin.reflect.KClass
+
+internal typealias StateFunction<S, E> = FiniteStateMachine<S, E>.(S, E) -> S
+
+internal class FiniteStateMachine<S : State, E : Event>(
```
zsmb13, line 10: Is this completely our own implementation for this, or is it based on something?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: Marton, your comment is right. But this hasn't been implemented before and I don't know how to handle this case. I use empty token as it worked before.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/FiniteStateMachine.kt
ogkuzmin: It's completely mine implementation. Based on my previous experience and made a little less simple (usually in FSM and in graph there is some collection of side effects)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: @JcMinarro What do you think we suppose to do if we failed in token getting?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientState.kt
ogkuzmin: Good idea. Did it

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientStateService.kt
ogkuzmin: Good catch. Thanks

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -150,8 +147,8 @@ public class ChatClient internal constructor(
         if (!ensureUserNotSet(listener)) {
             return
         }
-        state.user = user
-        // fire a handler here that the chatDomain and chatUI can use
+        clientStateService.onSetUser(user)
+        // fire a handler here that the chatDomain and chatUX can use
```
JcMinarro, line 151: `ChatUX` doesn't exist

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: Would be good to have a error state on that case, and doesn't try to connect (It is not in a good state to be connected)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: I think it's some merge issue)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: I made the standalone [Jira task for that ](https://stream-io.atlassian.net/browse/CAS-437?atlOrigin=eyJpIjoiMWFjOTU2MTVlMTdiNGM2OTgyMmM4MTJjYmQxZTk0ZGUiLCJwIjoiaiJ9)

ogkuzmin:
For the case when token was received with error - I created [Jira task for that](https://stream-io.atlassian.net/browse/CAS-437?atlOrigin=eyJpIjoiOTdjZjg4ZGM3MTI2NGQ3YTg1NjFkZDhlYTUxMDVjMjkiLCJwIjoiaiJ9)

JcMinarro (approved):

PR #832, opened 2020-11-09
Feature/update project codestyle

JcMinarro:
### Description

-. Update Kotlin Codestyle used into the project.
-. Update XML Codestyle used into the project.
-. Add them to the repository to be updated a cross all members of the team (They are included on the `.idea` folder)

### Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] PR targets the `develop` branch
- ~[ ] Changelog updated with client-facing changes~
- ~[ ] New code is covered by unit tests~
- ~[ ] Comparison screenshots added for visual changes~
- [x] Reviewers added

PR #833, opened 2020-11-09
feat: add support for marking all channels as read

carterhudson:
https://stream-io.atlassian.net/browse/CAS-400

### Description

This implements functionality for marking all channels as read. It provides a usecase at the domain level for marking all as read. The usecase will optimistically update the UI for marking all as read, and subsequently invoke the client api to trigger backend events. The backend events trigger offline db updates and refreshing the UI if necessary.

~Note: This includes a dirty UI change in the sample app for marking all channels as read for ease of testing. This will need to be removed / changed before merging.~ Moved "Mark All Read" into the toolbar options menu.

### Checklist

- [x] Reviewers added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Added new MessageInputView structure
 - Update `totalUnreadCount` when user is connected
 - Update `channelUnreadCount` when user is connected
 - Fix bug when channels could be shown without names
+- Add support for marking all channels as read for the current user
 
 # Nov 4th, 2020 - 4.4.1
 ## Common changes for all artifacts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,14 @@ Added new MessageInputView structure
 
 ## stream-chat-android-client
 - Deprecate `User::unreadCount` property, replace with `User::totalUnreadCount`
-- Add MarkAllReadEvent
+- Added MarkAllReadEvent
 
 ## stream-chat-android-offline
 - Update `totalUnreadCount` when user is connected
 - Update `channelUnreadCount` when user is connected
 - Fix bug when channels could be shown without names
-- Add support for marking all channels as read for the current user
+- Added support for marking all channels as read for the current user.
+    - Can be accessed via `ChatDomain`'s use cases (`chatDomain.useCases.markAllRead()...`).
 
 # Nov 4th, 2020 - 4.4.1
 ## Common changes for all artifacts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Added new MessageInputView structure
 
 ## stream-chat-android-client
 - Deprecate `User::unreadCount` property, replace with `User::totalUnreadCount`
+- Add MarkAllReadEvent
 
 ## stream-chat-android-offline
 - Update `totalUnreadCount` when user is connected
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
@@ -29,6 +29,7 @@ import io.getstream.chat.android.client.events.ErrorEvent
 import io.getstream.chat.android.client.events.GlobalUserBannedEvent
 import io.getstream.chat.android.client.events.GlobalUserUnbannedEvent
 import io.getstream.chat.android.client.events.HealthEvent
+import io.getstream.chat.android.client.events.MarkAllReadEvent
 import io.getstream.chat.android.client.events.MemberAddedEvent
 import io.getstream.chat.android.client.events.MemberRemovedEvent
 import io.getstream.chat.android.client.events.MemberUpdatedEvent
@@ -212,7 +213,8 @@ public class ChannelClient internal constructor(
             is ConnectedEvent,
             is ConnectingEvent,
             is DisconnectedEvent,
-            is ErrorEvent -> false
+            is ErrorEvent,
+            is MarkAllReadEvent -> false
         }
     }
 
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
@@ -16,7 +16,7 @@ public sealed class ChatEvent {
 }
 
 public sealed class CidEvent : ChatEvent() {
-    public abstract val cid: String
+    public abstract val cid: String?
 }
 
 public interface UserEvent {
@@ -168,7 +168,7 @@ public data class MessageReadEvent(
     override val type: String,
     @SerializedName("created_at") override val createdAt: Date,
     override val user: User,
-    override val cid: String,
+    override val cid: String?,
     @SerializedName("channel_type") val channelType: String,
     @SerializedName("channel_id") val channelId: String,
     @SerializedName("watcher_count") val watcherCount: Int?
@@ -257,7 +257,7 @@ public data class NotificationMarkReadEvent(
     override val type: String,
     @SerializedName("created_at") override val createdAt: Date,
     override val user: User,
-    override val cid: String,
+    override val cid: String?,
     @SerializedName("channel_type") val channelType: String,
     @SerializedName("channel_id") val channelId: String,
     @SerializedName("watcher_count") val watcherCount: Int?,
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
@@ -16,7 +16,7 @@ public sealed class ChatEvent {
 }
 
 public sealed class CidEvent : ChatEvent() {
-    public abstract val cid: String?
+    public abstract val cid: String
 }
 
 public interface UserEvent {
@@ -168,7 +168,7 @@ public data class MessageReadEvent(
     override val type: String,
     @SerializedName("created_at") override val createdAt: Date,
     override val user: User,
-    override val cid: String?,
+    override val cid: String,
     @SerializedName("channel_type") val channelType: String,
     @SerializedName("channel_id") val channelId: String,
     @SerializedName("watcher_count") val watcherCount: Int?
@@ -257,13 +257,22 @@ public data class NotificationMarkReadEvent(
     override val type: String,
     @SerializedName("created_at") override val createdAt: Date,
     override val user: User,
-    override val cid: String?,
+    override val cid: String,
     @SerializedName("channel_type") val channelType: String,
     @SerializedName("channel_id") val channelId: String,
     @SerializedName("watcher_count") val watcherCount: Int?,
     @SerializedName("total_unread_count") val totalUnreadCount: Int?,
     @SerializedName("unread_channels") val unreadChannels: Int?
 ) : CidEvent(), UserEvent
+
+public data class MarkAllReadEvent(
+    override val type: String,
+    @SerializedName("created_at") override val createdAt: Date,
+    override val user: User,
+    @SerializedName("unread_count") val unreadCount: Int?,
+    @SerializedName("total_unread_count") val totalUnreadCount: Int?,
+    @SerializedName("unread_channels") val unreadChannels: Int?
+) : ChatEvent(), UserEvent
 
 public data class NotificationMessageNewEvent(
     override val type: String,
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
@@ -266,12 +266,12 @@ public data class NotificationMarkReadEvent(
 ) : CidEvent(), UserEvent
 
 public data class MarkAllReadEvent(
-    override val type: String,
+    override val type: String = "",
     @SerializedName("created_at") override val createdAt: Date,
     override val user: User,
-    @SerializedName("unread_count") val unreadCount: Int?,
-    @SerializedName("total_unread_count") val totalUnreadCount: Int?,
-    @SerializedName("unread_channels") val unreadChannels: Int?
+    @SerializedName("unread_count") val unreadCount: Int = 0,
+    @SerializedName("total_unread_count") val totalUnreadCount: Int = 0,
+    @SerializedName("unread_channels") val unreadChannels: Int = 0
 ) : ChatEvent(), UserEvent
 
 public data class NotificationMessageNewEvent(
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
@@ -656,3 +656,15 @@ private fun createReactionJsonString() =
             "created_at": "2020-06-29T06:14:28.000Z"
         }
     """.trimIndent()
+
+internal fun createMarkAllReadEventStringJson() =
+    """
+        {
+           "unread_count":0.0,
+           "unread_channels":0.0,
+           "total_unread_count":0.0,
+           "created_at":"2020-06-29T06:14:28.000Z",
+           "type":"notification.mark_read",
+           "user":${createUserJsonString()}
+        }
+    """.trimIndent()
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
@@ -660,9 +660,9 @@ private fun createReactionJsonString() =
 internal fun createMarkAllReadEventStringJson() =
     """
         {
-           "unread_count":0.0,
-           "unread_channels":0.0,
-           "total_unread_count":0.0,
+           "unread_count":0,
+           "unread_channels":0,
+           "total_unread_count":0,
            "created_at":"2020-06-29T06:14:28.000Z",
            "type":"notification.mark_read",
            "user":${createUserJsonString()}
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/EventArguments.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/EventArguments.kt
@@ -287,7 +287,7 @@ internal object EventArguments {
         Arguments.of(createNewMessageEventStringJson(), newMessageEvent),
         Arguments.of(createUnknownEventStringJson(), unknownEvent),
         Arguments.of(createUnknownEventStringJson("some.unknown.type"), otherUnknownEvent),
-        Arguments.of(createMarkAllReadEventStringJson(), markAllReadEvent)
+        Arguments.of(createMarkAllReadEventStringJson(), markAllReadEvent),
     )
 
     @JvmStatic
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/EventArguments.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/EventArguments.kt
@@ -17,6 +17,7 @@ import io.getstream.chat.android.client.createConnectedEventStringJson
 import io.getstream.chat.android.client.createGlobalUserBannedEventStringJson
 import io.getstream.chat.android.client.createGlobalUserUnbannedEventStringJson
 import io.getstream.chat.android.client.createHealthEventStringJson
+import io.getstream.chat.android.client.createMarkAllReadEventStringJson
 import io.getstream.chat.android.client.createMemberAddedEventStringJson
 import io.getstream.chat.android.client.createMemberRemovedEventStringJson
 import io.getstream.chat.android.client.createMemberUpdatedEventStringJson
@@ -66,6 +67,7 @@ import io.getstream.chat.android.client.events.ConnectedEvent
 import io.getstream.chat.android.client.events.GlobalUserBannedEvent
 import io.getstream.chat.android.client.events.GlobalUserUnbannedEvent
 import io.getstream.chat.android.client.events.HealthEvent
+import io.getstream.chat.android.client.events.MarkAllReadEvent
 import io.getstream.chat.android.client.events.MemberAddedEvent
 import io.getstream.chat.android.client.events.MemberRemovedEvent
 import io.getstream.chat.android.client.events.MemberUpdatedEvent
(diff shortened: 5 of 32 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -310,22 +310,43 @@ internal class EventHandlerImpl(
...
+                                )
+                            )
                         }
-                        batch.addChannel(updatedChannel)
-                    }
+                        ?.let(batch::addChannel)
+                        ?: batch
+                            .getAllChannels()
+                            .forEach { channel ->
+                                channel.updateReads(ChannelUserRead(event.user, event.createdAt))
```
JcMinarro, line 327: Why do we update all all channels?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -385,29 +406,42 @@ internal class EventHandlerImpl(
...
-        events.sortedBy { it.createdAt }
-        updateOfflineStorageFromEvents(events)
+        val sortedEvents = events.sortedBy { it.createdAt }
+        updateOfflineStorageFromEvents(sortedEvents)
 
         // step 3 - forward the events to the active channels
+        sortedEvents
+            .filterIsInstance<CidEvent>()
+            .let { cidEvents ->
+                // handle CidEvents with non-null CIDs
```
JcMinarro, line 416: `CidEvent` was created to not need this kind of check. We shouldn't allow null value there

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -385,29 +406,42 @@ internal class EventHandlerImpl(
...
-        events.filterIsInstance<CidEvent>()
-            .groupBy { it.cid }
-            .filterNot { it.key.isBlank() }
-            .forEach {
-                val (cid, eventList) = it
-                if (domainImpl.isActiveChannel(cid)) {
-                    domainImpl.channel(cid).handleEvents(eventList)
-                }
+                // some CidEvents, like those for marking reads,
+                // can have null channel ids to signal "all channels"
```
JcMinarro, line 427: On that case a new Event needs to be created

JcMinarro (changes requested):

tschellenbach:
Make sure to follow the guidelines from the Jira ticket: https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=49&projectKey=CAS&modal=detail&selectedIssue=CAS-400

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventBatchUpdate.kt
```diff
@@ -71,15 +72,15 @@ internal class EventBatchUpdate private constructor(
     }
 
     internal class Builder {
-        private val channelsToFetch = mutableSetOf<String>()
+        private val channelsToFetch = mutableSetOf<String?>()
```
tschellenbach: we can't fetch null channels, also we don't need to since we should store the markReadAll in 1 place.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
tschellenbach: this is not good, we should implement this with 1 write, see Jira

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
tschellenbach: That seems like the cleanest approach, to have a MarkAllRead event object, also ensures we handle it

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -43,8 +43,10 @@ internal class QueryChannelsControllerImpl(
     override val endOfChannels: LiveData<Boolean> = _endOfChannels
 
     private val _channels = MutableLiveData<Map<String, Channel>>()
+
     // Keep the channel list locally sorted
-    override var channels: LiveData<List<Channel>> = Transformations.map(_channels) { cMap -> cMap.values.sortedWith(sort.comparator) }
+    override var channels: LiveData<List<Channel>> =
+        Transformations.map(_channels) { cMap -> cMap.values.sortedWith(sort.comparator) }
```
tschellenbach, line 51: editor config issue?

zsmb13: These should be fixed after https://github.com/GetStream/stream-chat-android/pull/832

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
```diff
@@ -656,3 +656,15 @@ private fun createReactionJsonString() =
             "created_at": "2020-06-29T06:14:28.000Z"
         }
     """.trimIndent()
+
+internal fun createMarkAllReadEventStringJson() =
+    """
+        {
+           "unread_count":0.0,
+           "unread_channels":0.0,
+           "total_unread_count":0.0,
```
JcMinarro, line 665: Why are they double values instead of integer?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -308,24 +310,32 @@ internal class EventHandlerImpl(
...
+                        ?.let(batch::addChannel)
+
                 is NotificationMarkReadEvent -> {
-                    event.totalUnreadCount?.let { domainImpl.setTotalUnreadCount(it) }
+                    event.totalUnreadCount?.let(domainImpl::setTotalUnreadCount)
 
-                    batch.getCurrentChannel(event.cid)?.let {
-                        val updatedChannel = it.apply {
+                    event.cid
+                        .let(batch::getCurrentChannel)
```
JcMinarro, line 334: Using `let` here is redundant

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -308,24 +310,32 @@ internal class EventHandlerImpl(
                     batch.addChannel(event.channel)
                 }
 
-                is MessageReadEvent -> {
-                    // get the channel, update reads, write the channel
-                    batch.getCurrentChannel(event.cid)?.let {
-                        val updatedChannel = it.apply {
+                // we use syncState to store the last markAllRead date for a given
+                // user since it makes more sense to write to the database once instead of N times.
+                is MarkAllReadEvent -> domainImpl.repos.syncState.apply {
```
JcMinarro, line 315: `ChatDomain::totalUnreadCount` and `ChatDomain::channelUnreadCount` needs to be updated

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -385,29 +395,34 @@ internal class EventHandlerImpl(
...
+        val sortedEvents = events.sortedBy { it.createdAt }
+        updateOfflineStorageFromEvents(sortedEvents)
 
         // step 3 - forward the events to the active channels
-
-        events.filterIsInstance<CidEvent>()
-            .groupBy { it.cid }
-            .filterNot { it.key.isBlank() }
-            .forEach {
+        sortedEvents.filterIsInstance<CidEvent>().let { cidEvents ->
```
JcMinarro, line 402: Using `let` here is redundant

CHANGELOG.md
```diff
@@ -5,11 +5,12 @@
 Remove `ChatClient` and `ChatDomain` as `ChatUI`'s dependecies
 
 ## stream-chat-android-client
-- Depecrate `User::unreadCount` property, replace with `User::totalUnreadCount`
+- Deprecrate `User::unreadCount` property, replace with `User::totalUnreadCount`
```
JcMinarro, line 8: We need to notify that a new event has been created

JcMinarro (changes requested):

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
carterhudson: Copied directly from a real event's json payload

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
carterhudson: Left over from previous implementation.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
tschellenbach: if i remember correctly the json spec doesn't differentiate between float and ints

CHANGELOG.md
```diff
@@ -2,14 +2,16 @@
...
 
 ## stream-chat-android-client
-- Depecrate `User::unreadCount` property, replace with `User::totalUnreadCount`
+- Deprecate `User::unreadCount` property, replace with `User::totalUnreadCount`
+- Add MarkAllReadEvent
 
 ## stream-chat-android-offline
 - Update `totalUnreadCount` when user is connected
 - Update `channelUnreadCount` when user is connected
+- Add support for marking all channels as read for the current user
```
zsmb13, line 14: Would be nice to mention how to do this more precisely, i.e. class/method name

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
```diff
@@ -265,6 +265,15 @@ public data class NotificationMarkReadEvent(
     @SerializedName("unread_channels") val unreadChannels: Int?
 ) : CidEvent(), UserEvent
 
+public data class MarkAllReadEvent(
+    override val type: String = "",
+    @SerializedName("created_at") override val createdAt: Date,
+    override val user: User,
+    @SerializedName("unread_count") val unreadCount: Int = 0,
+    @SerializedName("total_unread_count") val totalUnreadCount: Int = 0,
+    @SerializedName("unread_channels") val unreadChannels: Int = 0
```
zsmb13, line 274: We don't use default parameter values for any other event, why are we using them here? Gson will ignore them anyway, which might be confusing.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/EventArguments.kt
```diff
@@ -283,7 +286,8 @@ internal object EventArguments {
         Arguments.of(createNotificationMutesUpdatedEventStringJson(), notificationMutesUpdatedEvent),
         Arguments.of(createNewMessageEventStringJson(), newMessageEvent),
         Arguments.of(createUnknownEventStringJson(), unknownEvent),
-        Arguments.of(createUnknownEventStringJson("some.unknown.type"), otherUnknownEvent)
+        Arguments.of(createUnknownEventStringJson("some.unknown.type"), otherUnknownEvent),
+        Arguments.of(createMarkAllReadEventStringJson(), markAllReadEvent)
```
zsmb13, line 290: Let's add a trailing comma here, to avoid these kinds of diffs in the future when adding new items here

image

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -109,14 +116,14 @@ internal class QueryChannelsControllerImpl(
...
-            // The reason is that we are on the IO thread and update ChannelControlelr using postValue()
+            // The reason is that we are on the IO thread and update ChannelController using postValue()
             //  ChannelController.toChannel() can read the old version of the data using livedata.value
             // Solutions:
             // - suspend/wait for a few seconds (yuck, lets not do that)
             // - post the refresh on a livedata object with only channel ids, and transform that into channels (this ensures it will get called after postValue completes)
             // - run the refresh channel call below on the UI thread instead of IO thread
             domainImpl.scope.launch(Dispatchers.Main) {
-                refreshChannel(event.cid)
+                event.cid.let(::refreshChannel)
```
zsmb13, line 126: Why is this better than what we had before?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -273,6 +284,14 @@ internal class QueryChannelsControllerImpl(
         refreshChannels(listOf(cId))
     }
 
+    /**
+     * Refreshes all channels returned in this query.
+     * Supports use cases like marking all channels as read.
+     */
+    fun refreshAllChannels() {
+        queryChannelsSpec.cids.let(::refreshChannels)
```
zsmb13, line 292: Why are we using `let` instead of making a function call directly?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/SyncStateEntity.kt
```diff
@@ -9,5 +9,6 @@ internal data class SyncStateEntity(
     @PrimaryKey var userId: String,
     var activeChannelIds: List<String> = mutableListOf(),
     var activeQueryIds: List<String> = mutableListOf(),
-    var lastSyncedAt: Date? = null
+    var lastSyncedAt: Date? = null,
+    var markedAllReadAt: Date? = null
```
zsmb13, line 13: Let's add a trailing comma

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/BaseConnectedIntegrationTest.kt
```diff
@@ -61,6 +64,8 @@ internal open class BaseConnectedIntegrationTest : BaseDomainTest() {
         return chatDomainImpl
     }
 
+    private val mainThreadSurrogate = newSingleThreadContext("UI")
```
zsmb13, line 67: Couldn't we use `TestCoroutineDispatcher` here?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/MarkReadImpl.kt
```diff
@@ -19,12 +20,19 @@ public interface MarkRead {
 }
 
 internal class MarkReadImpl(private val domainImpl: ChatDomainImpl) : MarkRead {
-    override operator fun invoke(cid: String): Call<Boolean> {
-        validateCid(cid)
+    override operator fun invoke(cid: String): Call<Boolean> =
+        domainImpl.channel(validateCid(cid)).let { channelController ->
+            CoroutineCall(domainImpl.scope) {
+                channelController.markRead().let { markedRead ->
```
zsmb13, line 26: Is there a reason to introduce all this nesting here with `let` instead of just assigning the results of the calls to regular variables?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/MarkAllReadImpl.kt
```diff
@@ -0,0 +1,32 @@
...
+                        channel.markRead()
+                    }
+                }.awaitAll() // wait for the UI updates to avoid races
+
+                // then update via remote API
+                domain.client.markAllRead().execute()
+                Result(true, null)
+            }
+        }.await()
+    }
```
zsmb13, line 31:
```suggestion
    override fun invoke(): Call<Boolean> {
        return CoroutineCall(domain.scope) {
            // update the UI first
            domain.allActiveChannels().map { channel ->
                domain.scope.async(Dispatchers.Main) {
                    channel.markRead()
                }
            }.awaitAll() // wait for the UI updates to avoid races

            // then update via remote API
            domain.client.markAllRead().await()
            Result(true, null)
        }
    }
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -385,29 +402,34 @@ internal class EventHandlerImpl(
     }
 
     private suspend fun handleEventsInternal(events: List<ChatEvent>) {
-        events.sortedBy { it.createdAt }
```
zsmb13, line 388: Ouch, nice catch.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -308,24 +310,39 @@ internal class EventHandlerImpl(
...
+                    event.unreadChannels.let(domainImpl::setChannelUnreadCount)
+
+                    // only update sync state if the incoming "mark all read" date is newer
+                    // this supports using event handler to restore mark all read state in setUser
+                    // without redundant db writes.
+                    domainImpl.repos.syncState.apply {
+                        select(event.user.id)?.let { state ->
+                            if (state.markedAllReadAt == null || state.markedAllReadAt?.before(event.createdAt) == true) {
+                                insert(state.copy(markedAllReadAt = event.createdAt))
+                            }
```
zsmb13, line 326: `apply` is generally used when we're doing something _to_ the object, e.g. setting properties during initialization. If we want to open a scope into the repo here, `run` or `with` would be more fitting.

But a regular variable would work too, and is easier to read (I needed some time to figure out what we're calling on the repo inside the `apply` block):

```suggestion
                    val syncStateRepo = domainImpl.repos.syncState
                    syncStateRepo.select(event.user.id)?.let { state ->
                        if (state.markedAllReadAt == null || state.markedAllReadAt?.before(event.createdAt) == true) {
                            syncStateRepo.insert(state.copy(markedAllReadAt = event.createdAt))
```

zsmb13 (commented):
The changes in general look okay to me, just a bit too much magic in the syntax here and there for my taste.

JcMinarro (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
carterhudson: It isn't "better". It's left over from when it was chaining an elvis operator on a null CID. I'm happy to change it back.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/MarkReadImpl.kt
carterhudson: Seems like it's just a style preference.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
carterhudson: When we create a MarkAllRead event after pulling the timestamp from syncState in setUser, it seemed cleaner than dropping a magic 0 in for three parameters.

CHANGELOG.md
carterhudson: Will do. Wasn't sure how long or explicit changelog comments were intended to be.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/EventArguments.kt
carterhudson: Sure.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/MarkAllReadImpl.kt
carterhudson: Is there a reason you prefer `return`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
carterhudson: You're right. Don't know why I chose apply in this case.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
```diff
@@ -212,7 +213,8 @@ public class ChannelClient internal constructor(
             is ConnectedEvent,
             is ConnectingEvent,
             is DisconnectedEvent,
-            is ErrorEvent -> false
+            is ErrorEvent,
+            is MarkAllReadEvent -> false
```
tschellenbach, line 217: i like how it's a separate event, makes it easy to check if we handle it

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -239,15 +240,19 @@ internal class ChatDomainImpl internal constructor(
...
+
+                // restore queries
+                repos.queryChannels.selectById(state.activeQueryIds).forEach { spec ->
+                    queryChannels(spec.filter, spec.sort)
                 }
+
+                // retrieve the last time the user marked all as read and handle it as an event
+                state.markedAllReadAt
+                    ?.let { MarkAllReadEvent(user = currentUser, createdAt = it) }
+                    ?.let { eventHandler.handleEvent(it) }
```
tschellenbach, line 255: I'm not sure this works...

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
tschellenbach, line 326: write path looks good to me

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -98,6 +101,10 @@ internal class QueryChannelsControllerImpl(
             addChannelIfFilterMatches(event.channel)
         }
 
+        if (event is MarkAllReadEvent) {
+            refreshAllChannels()
```
tschellenbach, line 105: good

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
tschellenbach: so here's my concern:

- you trigger this event
- the client receives a new list of channels (which overwrites the reads)
- you end up showing channels that you marked as read, as unread

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/MarkAllReadImpl.kt
carterhudson: Just a heads up - if I use `domain.client.markAllRead().await()`, `MarkAllReadImplTest` hangs indefinitely. I've used `suspendCancellableCoroutine` to wrap callbacks in the past, so I'm not sure why it's not working in this instance. I'll check it out later if I get some time.

tschellenbach (dismissed):

JcMinarro (approved):

PR #848, opened 2020-11-11
CAS-330 CAS-331 Integrate mentions and suggestions into StreamMessage…

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-330
https://stream-io.atlassian.net/browse/CAS-331

### Description

The ideal solution would be to wrap `SuggestionListView` into `PopupWindow`. But this solution turned out to be complex as technically `PopupWindow` is not bound the fragment's view hierarchy. As a result, it is hard to keep fixed position above the input field when `PopupWindow`'s height changes, keyboard is displayed/hidden, etc. I was not able to get rid of all of 'PopupWindow' movement lags completely, therefore for now I implemented suggestion list as a part of `StreamMessageInputView`'s layout.

- Changed activity's input mode to `adjustPan` so that the keyboard overlaps bottom navigation and is displayed just below message input
- Added margin to input field drawable so that there is some spacing between the opened keyboard and text input
- Moved as much logic as possible to `SuggestionListController` to simplify `StreamMessageInputView ` class 
- Handled clicks on suggestions

Improvement suggestions:
1) Implement and polish the solution with `PopupWindow`
2) Leave `SuggestionListView` as a part of `StreamMessageInputView` layout, but allow customers to provide external `SuggestionListView`:

```
<LinearLayout>
    <FrameLayout>
        <MessageListView>
        <SuggestionListView layout_alignParentBottom="bottom">
   </FrameLayout>
   <StreamMessageInputView>
</LinearLayout>

// any location customer wants, but typically aligned with the bottom of MessageListView
messageInputView.setCustomSuggestionView(suggestionsView) 
```

### Checklist

- [ ] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListController.kt
```diff
@@ -0,0 +1,71 @@
+package io.getstream.chat.android.ui.suggestions
+
+import android.widget.EditText
+import io.getstream.chat.android.client.models.Command
+import io.getstream.chat.android.client.models.Member
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.client.models.name
+import java.util.regex.Pattern
+
+internal class SuggestionListController(
```
JcMinarro, line 10: Good abstraction ;)

JcMinarro (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
```diff
@@ -170,6 +183,12 @@ public class StreamMessageInputView : ConstraintLayout {
     }
 
     private fun configTextInput(typedArray: TypedArray) {
+        suggestionListController = SuggestionListController(binding.suggestionListView, binding.etMessageTextInput)
+
+        binding.etMessageTextInput.doAfterTextChanged {
+            suggestionListController.onTextChanged(it.toString())
```
zsmb13, line 189: Careful, we're calling the [`Any?.toString()`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/to-string.html) extension here, because `it` is an `Editable?`. This will pass in `"null"` if `it == null`.

Using `it?.toString()` instead would return `null` if `it == null`, and we could then default to an empty string with an Elvis operator, or something like that instead.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputViewModelBinding.kt
```diff
@@ -0,0 +1,16 @@
+@file:JvmName("StreamMessageInputViewModelBinding")
```
zsmb13, line 1: We didn't prefix these with `Stream` before, do we want to adopt this new naming style?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
```diff
@@ -45,10 +49,19 @@ public class StreamMessageInputView : ConstraintLayout {
...
     private var iconDisabledSendButtonDrawable: Drawable? = null
     private var iconEnabledSendButtonDrawable: Drawable? = null
 
+    public fun configureMembers(members: List<Member>) {
+        suggestionListController.members = members
+    }
+
+    public fun configureCommands(commands: List<Command>) {
+        suggestionListController.commands = commands
+    }
```
zsmb13, line 63: Don't we have some kind of work to perform after these values change? Do we only want to update the next time that the text input changes as well?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListController.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    private fun String.isCommandMessage() = COMMAND_PATTERN.matcher(this).find()
+
+    private fun String.isMentionMessage() = MENTION_PATTERN.matcher(this).find()
+
+    private fun String.getCommandSuggestions(): SuggestionListView.Suggestions.CommandSuggestions {
+        val commandPattern = removePrefix("/")
+        return commands
+            .filter { it.name.startsWith(commandPattern) }
+            .toList()
```
zsmb13, line 55: These `toList` calls seem redundant in both functions, but maybe I'm missing something

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListController.kt
```diff
@@ -0,0 +1,71 @@
...
+            .let { SuggestionListView.Suggestions.CommandSuggestions(it) }
+    }
+
+    private fun String.getMentionSuggestions(): SuggestionListView.Suggestions.MentionSuggestions {
+        val namePattern = substringAfterLast("@")
+        return members.map { it.user }
```
zsmb13, line 61: We do this mapping every time that the input string changes, perhaps we should just store the mapped result to begin with (when `members` are set), so that we only compute it once

zsmb13 (commented):
Looks neat! Would be nice to have it added to the Component Browser so that it's easy to test this in the sample app, but that can also go into a separate PR.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputViewModelBinding.kt
JcMinarro: Added as a topic to our Weekly

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
bychkovdmitry: Really nice catch, thanks!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputViewModelBinding.kt
zsmb13: I think the decision here was to _not_ prefix class names, right?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputViewModelBinding.kt
bychkovdmitry: Removed `Stream` prefix for `MessageInputView` and `MessageInputViewModelBinding`. As discussed earlier, adding prefixes to resources in a library makes perfect sense as library users can accidentally override them. This problem does not apply to classes so to be consistent with what we have currently I'm omitting `Stream` prefix.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/StreamMessageInputView.kt
bychkovdmitry: Done, recalculate suggestions in controller when new values arrive

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListController.kt
bychkovdmitry: Done

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestions/SuggestionListController.kt
bychkovdmitry: Done

zsmb13 (approved):

PR #852, opened 2020-11-12
[CAS-378] Typing indicator

leandroBorgesFerreira:
[CAS-378](https://stream-io.atlassian.net/browse/CAS-378)

### Description

This PR creates the communication for the updates for typing event in the channels. With `typingUpdates` it is possible to listen to typing events in all active channels, this way we can update which channel has someone typing in. 

I'm separating the communication with the actual view as I wait for the new design of channels view to be ready and it is good to have a discussion of this code as I'm changing `ChatDomainImpl.kt ` which is a very base class. 

### Checklist

- [x] Changelog updated with client-facing changes. 
- [ ] New code is covered by unit tests -> **No unit tests**
- [ ] Comparison screenshots added for visual changes -> **No visual changes.**
- [x] Reviewers added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ Added new MessageInputView structure
 - Added support for marking all channels as read for the current user.
     - Can be accessed via `ChatDomain`'s use cases (`chatDomain.useCases.markAllRead()...`).
 - Fix bug when local channels could be sorted not properly
+- Typing events can be all tracked from `ChatDomain`
 
 # Nov 4th, 2020 - 4.4.1
 ## Common changes for all artifacts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ Added new MessageInputView structure
 - Added support for marking all channels as read for the current user.
     - Can be accessed via `ChatDomain`'s use cases (`chatDomain.useCases.markAllRead()...`).
 - Fix bug when local channels could be sorted not properly
-- Typing events can be all tracked from `ChatDomain`
+- Typing events can be all tracked with `ChatDomain.typingUpdates`
 
 # Nov 4th, 2020 - 4.4.1
 ## Common changes for all artifacts
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/TypeEvent.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/TypeEvent.kt
@@ -1,0 +1,3 @@
+package io.getstream.chat.android.client.models
+
+public data class TypeEvent(val channelId: String, val users: List<User>)
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/TypeEvent.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/TypeEvent.kt
@@ -1,3 +1,0 @@
-package io.getstream.chat.android.client.models
-
-public data class TypeEvent(val channelId: String, val users: List<User>)
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -8,6 +8,7 @@ import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.models.Config
 import io.getstream.chat.android.client.models.Mute
+import io.getstream.chat.android.client.models.TypeEvent
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.client.notifications.handler.NotificationConfig
 import io.getstream.chat.android.livedata.controller.QueryChannelsController
@@ -76,7 +77,7 @@ public interface ChatDomain {
     /** The retry policy for retrying failed requests */
     public var retryPolicy: RetryPolicy
 
-    public val typingUpdates: LiveData<Pair<String, List<User>>>
+    public val typingUpdates: LiveData<TypeEvent>
 
     /** a helper object which lists all the initialized use cases for the chat domain */
     public val useCases: UseCaseHelper
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -76,7 +76,6 @@ public interface ChatDomain {
     /** The retry policy for retrying failed requests */
     public var retryPolicy: RetryPolicy
 
-
     public val typingUpdates: LiveData<Pair<String, List<User>>>
 
     /** a helper object which lists all the initialized use cases for the chat domain */
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -76,6 +76,9 @@ public interface ChatDomain {
     /** The retry policy for retrying failed requests */
     public var retryPolicy: RetryPolicy
 
+
+    public val typingUpdates: LiveData<Pair<String, List<User>>>
+
     /** a helper object which lists all the initialized use cases for the chat domain */
     public val useCases: UseCaseHelper
 
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -8,7 +8,7 @@ import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.models.Config
 import io.getstream.chat.android.client.models.Mute
-import io.getstream.chat.android.client.models.TypeEvent
+import io.getstream.chat.android.client.models.TypingEvent
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.client.notifications.handler.NotificationConfig
 import io.getstream.chat.android.livedata.controller.QueryChannelsController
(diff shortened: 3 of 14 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -508,6 +515,9 @@ internal class ChatDomainImpl internal constructor(
                     this
                 )
             activeChannelMapImpl[cid] = channelRepo
+            GlobalScope.launch(Dispatchers.Main) {
```
zsmb13, line 518: - Why do we need to go to the main thread here?
- If we do need this, please use `DispatcherProvider` instead

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Added new MessageInputView structure
 - Added support for marking all channels as read for the current user.
     - Can be accessed via `ChatDomain`'s use cases (`chatDomain.useCases.markAllRead()...`).
 - Fix bug when local channels could be sorted not properly
+- Typing events can be all tracked from `ChatDomain`
```
zsmb13, line 27: Looking at this as a user, I don't know _how_ to track typing events. Would be useful to indicate the specific class/method to be used.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -76,6 +76,8 @@ public interface ChatDomain {
     /** The retry policy for retrying failed requests */
     public var retryPolicy: RetryPolicy
 
+    public val typingUpdates: LiveData<Pair<String, List<User>>>
```
zsmb13, line 79: `Pair` is a really inconvenient class to use in an API, we should almost always introduce our own class instead. Here, someone observing this `LiveData` would have to write down things like `value.second.size` to get the number of users typing, etc.

If we add our own class to wrap the channel ID and the list of users, it can have well-named properties instead of `first` and `second`.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -519,6 +529,22 @@ internal class ChatDomainImpl internal constructor(
...
+        _typingChannels.addSource(typingEvent, _typingChannels::postValue)
+    }
+
+    private fun removeTypingChannel(channelController: ChannelControllerImpl) {
+        val typingEvent: LiveData<Pair<String, List<User>>> = channelController.typing.map { userList ->
+            channelController.channelId to userList
+        }
+
+        _typingChannels.removeSource(typingEvent)
+    }
```
zsmb13, line 546: This is never called, and it removes a source that's created here, within the method. Seems like this wouldn't do practically anything.

zsmb13 (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
leandroBorgesFerreira: If addSource it called in another thread different than the main one, it crashes, so this is needed to move to the main thread.

ogkuzmin: I think GlobalScope usage is not the best practice to launch a coroutine. ChatDomain contains its' own scope, you can use it here.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
leandroBorgesFerreira: Where do we normally put our model classes? I see `io.getstream.chat.android.client.models`, but nothing similar in the   `stream-chat-android-offfile` project.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
leandroBorgesFerreira: It can remove a LiveData of a channel depending of our the `equals` and hash code method of LiveData works... But you're right, we never remove channels from activeChannels map, so this will never be used.

ogkuzmin: Since our offline support isn't used by customers without our view models then we could change that type (include there `channelId`) and omit mapping each time when add and remove source.

CHANGELOG.md
leandroBorgesFerreira: Done

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
leandroBorgesFerreira: Removed

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
leandroBorgesFerreira: Changed for `DispatcherProvider`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
zsmb13: Well, this touches on the sensitive issue where we should separate our domain models and our DTOs... We have an `entity` package in the `offline` module, which kind of contains models, but some of those are Room entities while some others aren't. Not sure what's the best place to put a new model.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -519,6 +529,14 @@ internal class ChatDomainImpl internal constructor(
         return currentUser.id + "-" + UUID.randomUUID().toString()
     }
 
+    private fun addTypingChannel(channelController: ChannelControllerImpl) {
+        val typingEvent: LiveData<Pair<String, List<User>>> = channelController.typing.map { userList ->
```
ogkuzmin, line 533: Our offline support feature (`ChatDomain`) isn't used by customers without our view models. Then we could change the type of `ChannelController::typing` and include there `channelId` . Result will be `LiveData<Pair<String, List<User>>>`. It lets us to omit mapping each time when add new channelController.

leandroBorgesFerreira: 🤔  But how would we create a way to listen for typing events in all channels at the same time? I use the mapping because new channels might be added and I need to listen for updates in all active channels and, if I understand correctly, a channelController listen to only of channel.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
leandroBorgesFerreira: Sorry @ogkuzmin, I didn't see you message. I'll change the scope.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
leandroBorgesFerreira: Ok. I'll change to a DTO and then we discuss where to place it, move classes is super simple

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
ogkuzmin: The same idea. When you create new `ChannelController` and add to a map of `ChatDomain` you also add its' `ChannelController::typing` as a source to all typingEvents into `ChatDomain::_typingChannels`. So implementation will be the same, we only can omit mapping of `ChannelController::typing`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
JcMinarro: Just create a class that model this info.
We will need to update it later with new "Typing Events" that will perform over threads (It is not implemented yet).
For now, the place to store our models is on the Client module (`io.getstream.chat.android.client.models`) In the future this package could live in a "common" module instead

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
leandroBorgesFerreira: Done. I also mapped the `Pair<String, List<User>>` to `TypeEvent`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
leandroBorgesFerreira: Done, mapped to TypeEvent.

JcMinarro (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/TypeEvent.kt
```diff
@@ -0,0 +1,3 @@
+package io.getstream.chat.android.client.models
+
+public data class TypeEvent(val channelId: String, val users: List<User>)
```
zsmb13, line 3: One last request: we call this _typing_ everywhere, including all the LiveDatas and also the events that eventually map into this (`TypingStartEvent`, `TypingStopEvent`). Let's call this `TypingEvent` as well.

(This current name also feels a bit like it's an event about the type of something.)

zsmb13 (changes requested):

ogkuzmin (approved):
for me LGTM

zsmb13 (approved):

PR #853, opened 2020-11-12
Adding unit tests only for debug build type

leandroBorgesFerreira:
### Description

This PR changes our github actions so we only run unit tests for the debug build type.

### Checklist

- [ ] Changelog updated with client-facing changes. **Not necessary**
- [ ] New code is covered by unit tests. **Not necessary**
- [ ] Comparison screenshots added for visual changes. **Not necessary**
- [x] Reviewers added

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -16,7 +16,7 @@ jobs:
       - name: ktlint
         run: ./gradlew :stream-chat-android-client:ktlintCheck
       - name: Unit tests
-        run: ./gradlew :stream-chat-android-client:test --stacktrace
+        run: ./gradlew :stream-chat-android-client:testDebug --stacktrace
       - name: Coverage
         run: ./gradlew :stream-chat-android-client:jacocoTestReport
       - name: Upload Coverage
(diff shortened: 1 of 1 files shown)


zsmb13:
What's the benefit of making this change?

leandroBorgesFerreira:
@zsmb13  This CI would run faster. There's no need to unit test both debug and release build types since there's difference between then in our code

zsmb13:
I mean, there _is_ a difference, and we've had failing release builds before due to never building that variant. But for tests, it's probably enough to run debug, agreed.

Though I don't see any speedup in the CI run on this PR compared to other ones, still takes 7 minutes overall. Might be worth looking into.

.github/workflows/pr-checks.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       - name: ktlint
         run: ./gradlew :stream-chat-android:ktlintCheck :stream-chat-android-sample:ktlintCheck
       - name: Unit tests
-        run: ./gradlew :stream-chat-android:test :stream-chat-android-sample:test --stacktrace
+        run: ./gradlew :stream-chat-android:testDebug :stream-chat-android-sample:testDebug --stacktrace
```
zsmb13, line 81: We should remove the uploading of the release test results if we make this change, it's a couple lines down from here in a step mistakenly named `Upload testDebugUnitTest results` due to a copy-paste error 😄

leandroBorgesFerreira: done

zsmb13: This is there for all different jobs, not just the one, we should remove it from all of them.

leandroBorgesFerreira: Oh, I see. Removed from all jobs now.

zsmb13 (dismissed):

JcMinarro (approved):

PR #864, opened 2020-11-13
feat: CAS-364 implement channel list items for redesign

carterhudson:
https://stream-io.atlassian.net/browse/CAS-364

### Description

This is a functioning re-implementation of channel list items for the redesign, ui-components module, and sample. The list items are not visually complete yet, but they provide enough functionality for hooking click events to complete the navigation flow to the chat/channel/message list screen.

Also contained in this implementation is a slight refactor of the channel list item view holder api. The previous implementation was unusable, as it required clients to provide an instance of a BaseChannelListItemViewHolder. When extending that class, you are required to override setStyle which takes a ChannelListViewStyle as a parameter. However, ChannelListViewStyle is internal to the SDK, and you can't proceed from there.

With the refactor, to specify your own views / viewholders, all you need to do is

1) Set the view holder layout as a ChannelsView attribute.
2) Create a ViewHolder that extends BaseChannelListItemViewHolder.
3) Create a ViewHolder factory that extends BaseChannelViewHolderFactory, and provide the ViewHolder from Step 2.

There is a method in `ChannelsFragment` that you can uncomment to use a test custom viewholder.

BREAKING CHANGE: Would require clients to extend BaseChannelViewHolderFactory instead of ChannelViewHolderFactory

### Checklist

- [ ] Reviewers added


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channels/ChannelsFragment.kt
```diff
@@ -35,22 +46,58 @@ class ChannelsFragment : Fragment() {
...
+
+            val loadingView = layoutInflater.inflate(
+                R.layout.channels_loading_view,
+                view as ViewGroup,
+                false
+            )
+
+            setLoadingView(loadingView, FrameLayout.LayoutParams(MATCH_PARENT, MATCH_PARENT))
+
+//            testViewHolderAPI()
```
carterhudson, line 60: Left this in for easy testing / verification that custom view holders can be used

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.java
```diff
@@ -0,0 +1,190 @@
+package io.getstream.chat.android.ui.channel.list;
```
carterhudson, line 1: Ports `ChannelListView` from old module to new so that VH factory could be modified. Also want to convert this to Kotlin.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
```diff
@@ -9,13 +9,19 @@ import io.getstream.chat.android.ui.R
...
 import io.getstream.chat.android.ui.utils.extensions.use
 
-internal class AvatarStyle(context: Context, attrs: AttributeSet?) {
-    val avatarWidth: Int
-    val avatarHeight: Int
-    val avatarBorderWidth: Int
-    val avatarBorderColor: Int
-    val avatarInitialText: TextStyle
-    val onlineIndicatorEnabled: Boolean
+public class AvatarStyle internal constructor(context: Context, attrs: AttributeSet?) {
```
carterhudson, line 12: Should we consider re-implementing the builder pattern for this style? It's used in `ChannelListViewStyle` where the old UI components expected it to be a `data class` and so provided a method to change the border color by using `avatarView.copy(...)`, which isn't possible with this one. The fields were originally `val`, but I made them `var` with `internal set` for now.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/diff/ChannelListDiffCallback.kt
```diff
@@ -0,0 +1,73 @@
+package io.getstream.chat.android.ui.channel.list.adapter.diff
+
+import androidx.recyclerview.widget.DiffUtil
+import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.name
+import io.getstream.chat.android.ui.utils.extensions.getLastMessage
+import io.getstream.chat.android.ui.utils.extensions.getLastMessageReadCount
+import io.getstream.chat.android.ui.utils.extensions.getUsers
```
ogkuzmin, line 8: Seems code style violation. It's better to check your branch via ktlint (seems we don't do it automatically)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolder.kt
```diff
@@ -0,0 +1,88 @@
...
+                }
+
+                if (lastMessageChanged) {
+                    lastMessageLabel.text = channel.getLastMessageDisplayText()
+                    channel.getLastMessageTime()
+                        ?.let(TIME_FORMAT::format)
+                        ?.let { lastMessageTimeLabel.text = it }
+
+                    channel.getCurrentUserRead()?.unreadMessages?.let { unreadCount ->
+//                        BadgeDrawable.create(context).apply {
```
ogkuzmin, line 54: What is it? Should it be deleted?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt
```diff
@@ -0,0 +1,69 @@
...
+                excludeCurrentUser -> users.withoutCurrentUser()
+                else -> users
+            }
+        }
+
+internal fun Channel.getDisplayName(): String = name.takeIf { it.isNotEmpty() }
+    ?: getUsers().joinToString { it.name }
+
+internal fun Channel.getLastMessage(): Message? =
+    messages
```
ogkuzmin, line 29: three times filtering can be expensive. We can use Kotlin sequences for this, it gives you some optimisation on big amount of channel's messages.

ogkuzmin (commented):
LGTM, but I left some comments

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.java
JcMinarro: Sure, all new code "should be written" in Kotlin

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/BaseChannelListItemAdapter.kt
```diff
@@ -0,0 +1,37 @@
...
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.RecyclerView
+import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.ui.channel.list.ChannelListView
+import io.getstream.chat.android.ui.channel.list.ChannelListViewStyle
+import io.getstream.chat.android.ui.channel.list.adapter.diff.ChannelListDiffCallback
+import io.getstream.chat.android.ui.channel.list.adapter.viewholder.BaseChannelListItemViewHolder
+
+public abstract class BaseChannelListItemAdapter :
+    RecyclerView.Adapter<BaseChannelListItemViewHolder>(),
```
JcMinarro, line 12: Why don't we use a [`RecyclerView.ListAdapter`](https://developer.android.com/reference/androidx/recyclerview/widget/ListAdapter) that already implement all logic about replace items and use the diffUtil Callback?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/BaseChannelListItemAdapter.kt
```diff
@@ -0,0 +1,37 @@
...
+    ChannelListViewClickDelegate {
+
+    protected open var channels: List<Channel> = emptyList()
+
+    override var userClickListener: ChannelListView.UserClickListener? = null
+
+    override var channelClickListener: ChannelListView.ChannelClickListener? = null
+
+    override var channelLongClickListener: ChannelListView.ChannelClickListener? = null
```
JcMinarro, line 21: Could we avoid nullable values here?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/BaseChannelListItemAdapter.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    public open var style: ChannelListViewStyle? = null
```
JcMinarro, line 23: Why the adapter needs the `style`?
It is not used within the adapter, it should be part of the ViewHolderFactory instead

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
```diff
@@ -0,0 +1,46 @@
...
+    override fun getItemViewType(position: Int): Int =
+        viewHolderFactory?.getItemViewType(channels[position])
+            ?: throw IllegalStateException("Please provide a view holder factory")
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): BaseChannelListItemViewHolder =
+        viewHolderFactory?.createChannelViewHolder(getChannelItemLayout(), parent, viewType)?.also { vh ->
+            vh.style = style
+            vh.userClickListener = userClickListener
+            vh.channelLongClickListener = channelLongClickListener
+            vh.channelClickListener = channelClickListener
```
JcMinarro, line 36: Why don't we have all those attributes in the Factory?

JcMinarro (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/diff/ChannelListDiffCallback.kt
carterhudson: Which part is a code style violation? I can't commit via git if I have ktlint errors, so this is all valid for ktlint.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolder.kt
carterhudson: This is still a WIP. Only made the PR so we could get the full flow in place.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/BaseChannelListItemAdapter.kt
carterhudson: I think so as well, but I wasn't sure how much I could change things.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
carterhudson: I agree. Wasn't sure how much I would be able to change things from the current API.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/diff/ChannelListDiffCallback.kt
ogkuzmin: It seems like a violation
image

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Any.kt
```diff
@@ -0,0 +1,8 @@
+package io.getstream.chat.android.ui.utils.extensions
+
+internal inline fun <reified AnyT> Any.cast() = this as AnyT
```
bychkovdmitry, line 3: Maybe just `T` instead of `AnyT` parameter?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt
```diff
@@ -0,0 +1,69 @@
...
+        .filter { it.createdAt != null || it.createdLocallyAt != null }
+        .filter { it.deletedAt == null }
+        .filter { it.type == "regular" } // migrate ModelType?
+        .maxByOrNull { it.createdAt ?: it.createdLocallyAt!! }
+
+internal fun Channel.getLastMessageDisplayText(): String =
+    getLastMessage()?.let { "${it.getDisplayName()}: ${it.text}" } ?: String.EMPTY
```
bychkovdmitry, line 36: Looks convenient, but I don't think we should do string concatenation for the UI in extension functions

bychkovdmitry (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/diff/ChannelListDiffCallback.kt
carterhudson: Ah, I see. I'll fix that.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
carterhudson: I agreed initially, but I've since found some issues when trying to do it that way (https://getstream.slack.com/archives/G016YAWD423/p1605638235150100)

stream-chat-android-ui-components/src/main/res/layout/stream_channel_list_item_view.xml
```diff
@@ -0,0 +1,80 @@
...
+        android:layout_marginEnd="8dp"
+        android:textSize="12sp"
+        tools:text="3:00PM"
+        app:layout_constraintBottom_toBottomOf="@+id/message_status_image_view"
+        app:layout_constraintTop_toTopOf="@+id/message_status_image_view" />
+
+    <ImageView
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        app:srcCompat="@drawable/ic_icon_check_all"
```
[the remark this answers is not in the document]
carterhudson: I think that gets set when you drag an ImageView in via GUI editor 🤔

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Any.kt
carterhudson: I generally prefer naming generic parameters with some sort of contextual name. We don't use single letter variable names, so why do it with generics? Makes the code easier to read and understand, in my opinion. Anyway, I can change it if we feel strongly about it.

stream-chat-android-ui-components/src/main/res/layout/stream_channel_list_item_view.xml
```diff
@@ -0,0 +1,80 @@
...
+        android:layout_marginStart="8dp"
+        android:layout_marginTop="12dp"
+        android:layout_marginBottom="12dp" />
+
+    <TextView
+        android:id="@+id/channel_name_label"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="8dp"
+        android:fontFamily="sans-serif-black"
```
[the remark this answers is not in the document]
carterhudson: This is just the default roboto bold font as far as I know. Seemed cleaner than using the Spannable API to make it bold.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt
carterhudson: Would you rather it be a string template with placeholders in `strings.xml`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
JcMinarro: For now we can leave it as it is, but we need to think about that. It doesn't work fine when different VHs live on the same adapter

JcMinarro (approved):

PR #866, opened 2020-11-16
Cas 391/postpone query channels

ogkuzmin:
[CAS-391](https://stream-io.atlassian.net/browse/CAS-391?atlOrigin=eyJpIjoiYjlmZDAyOWEyYjk2NDk0YmI3MmU4ODM5ZDYxNzFjNmUiLCJwIjoiaiJ9)
### Description

Logic that let us to postpone user's channel(s) queries while set user is being executed.
Must be merged after #831 

### Checklist

- [x] Reviewers added


JcMinarro (dismissed):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -479,22 +481,11 @@ public class ChatClient internal constructor(
         request: QueryChannelRequest
     ): Call<Channel> {
         // for convenience we add the message.cid field
-        return api.queryChannel(channelType, channelId, request)
-            .map { channel ->
-                channel.messages.forEach { message -> message.cid = channel.cid }
-                channel
-            }
+        return queryChannelsPostponeHelper.queryChannel(channelType, channelId, request)
```
zsmb13, line 484: The comment above this line no longer makes sense

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ## stream-chat-android
 
 ## stream-chat-android-client
+- There is no more the requirement to wait set user before querying channels
```
zsmb13, line 9:
```suggestion
- It's no longer required to wait for `setUser` to finish before querying channels
```

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/call/WrappedValueCall.kt
```diff
@@ -0,0 +1,11 @@
+package io.getstream.chat.android.client.call
+
+import io.getstream.chat.android.client.utils.Result
+
+internal class WrappedValueCall<T : Any>(private val value: T) : Call<T> {
```
zsmb13, line 5: Note: this is the same as the `TestCall` class in one of our other modules. Let's leave it as-is for this PR though, I'll move them into a common test utilities module after this is merged (along with some other test stuff).

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/call/CallTestUtils.kt
```diff
@@ -0,0 +1,5 @@
+package io.getstream.chat.android.client.call
+
+internal fun <T : Any> callFrom(valueProvider: () -> T): Call<T> = WrappedValueCall(valueProvider())
+
+internal fun <T : Any> T.asCall(): Call<T> = WrappedValueCall(this)
```
zsmb13, line 5: These feel a bit like overkill for their current usages, but they can stay if you like them 😄

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
```diff
@@ -0,0 +1,69 @@
...
+                    channel.messages.forEach { message ->
+                        message.cid = channel.cid
+                    }
+                }
+                channels
+            }
+        }
+    }
+
+    private tailrec suspend fun <T> doJob(attemptCount: Int = attemptsCount, job: () -> T): T {
```
zsmb13, line 50: The `tailrec` implementation is fancy, I would've just written a loop with `repeat(attemptsCount)`, but the end result is identical, so we can keep the fancy 👍

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
```diff
@@ -0,0 +1,69 @@
...
+        return when (clientStateService.state) {
+            is ClientState.Idle -> error("User must be set before querying channels")
+            is ClientState.User.Authorized,
+            is ClientState.Anonymous.Authorized -> job()
+            is ClientState.User.Pending,
+            is ClientState.Anonymous.Pending -> {
+                delay(delayDuration)
+                doJob(attemptCount - 1, job)
+            }
+        }
```
zsmb13, line 61: This is incredibly clean, the well-typed state definitions for the client were so worth it 🎉

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
```diff
@@ -0,0 +1,69 @@
...
+                        message.cid = channel.cid
+                    }
+                }
+                channels
+            }
+        }
+    }
+
+    private tailrec suspend fun <T> doJob(attemptCount: Int = attemptsCount, job: () -> T): T {
+        check(attemptCount > 0) { "Failed to perform job. Time out has reached" }
```
zsmb13, line 51: This error message is lying, we didn't reach a timeout, we ran out of retry attempts. Also, where is this exception handled, if anywhere?

zsmb13 (commented):
Looks great overall, nice work! Some small comments, mostly just noting things.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/call/CallTestUtils.kt
ogkuzmin: I think it will be convenient for tests. When I faced that we hadn't anything like this in tests I was wondered. Because Call is wide spread thing in LLC

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/call/WrappedValueCall.kt
ogkuzmin: Sounds good. Maybe we need not core module, but some test-core module. I think If we start to move everything in new core module it soon will became a garbage module

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
ogkuzmin: It's just my brain transformation of functional approach and haskell aspiration. I prefer recursive way instead of loops :-)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
ogkuzmin: Agree, that's why I like strongly typed approach, algebraic data structures and so on

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/call/WrappedValueCall.kt
zsmb13: Yeah, we need a separate test module, because we only want to include that as a `testImplementation` dependency. It was just wasn't possible before, because it requires us to move some stuff to `core` first, so that the test module can depend on it.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
ogkuzmin: It is not handled anywhere. I think it's completely ok since it is part of a network API layer, everything can go wrong and the code client should be ready for that.
Agree about message

zsmb13 (approved):

PR #869, opened 2020-11-16
Changing GlobalScope for CoroutineScope

leandroBorgesFerreira:
### Description

Small fix the StartStopBuffer. It is not necessary to use GlobalScope in this class, the CoroutinesScope is probably the best one here. 

### Checklist

- [ ] Changelog updated with client-facing changes. **Not necessary**
- [ ] New code is covered by unit tests.  **Not necessary**
- [ ] Comparison screenshots added for visual changes.  **Not necessary**
- [x] Reviewers added


samiuelson (approved):

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/StartStopBuffer.kt
```diff
@@ -27,7 +27,7 @@ internal class StartStopBuffer<T> {
     }
 
     private fun propagateData() {
-        GlobalScope.launch(DispatcherProvider.IO) {
+        CoroutineScope(DispatcherProvider.IO).launch {
```
zsmb13, line 30: This change makes no difference, by the way. An uncancelled new scope might as well be `GlobalScope`. If anything, it uses more resources by creating the new scope here.

PR #876, opened 2020-11-18
Create test module

zsmb13:
### Description

This PR creates a new `test` module for common utilities to be used in our tests.

- The `test` module depends on the `core` module
- The new `test` module contains the coroutine helpers for `JUnit4` and `JUnit5` so that we no longer have to duplicate them in each test source set
- The new `test` module contains the `TestCall` class so that we don't have to duplicate this either
    - This change required moving the `Call`, `Result`, and `ChatError` classes to the `core` module, so that they can be referenced by the test utilities. For the time being, they were moved to the new module, but kept the old package name. This should be fixed eventually (moved into the `core` package instead of the `client` package), but will be a breaking change and a deprecation cycle.

### Checklist

- [ ] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/logger/ChatLoggerImpl.kt
```diff
@@ -28,20 +28,23 @@ internal class ChatLoggerImpl constructor(
     }
 
     override fun logE(tag: Any, chatError: ChatError) {
+        val cause = chatError.cause
+        val message = chatError.message
```
zsmb13, line 32: Required because smart casts don't kick in across module boundaries

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/call/WrappedValueCall.kt
```diff
@@ -1,11 +0,0 @@
-package io.getstream.chat.android.client.call
-
-import io.getstream.chat.android.client.utils.Result
-
-internal class WrappedValueCall<T : Any>(private val value: T) : Call<T> {
```
zsmb13, line 5: Replaced by `TestCall`

stream-chat-android-core/src/main/java/io/getstream/chat/core/internal/fsm/builder/FSMBuilderMarker.kt
```diff
@@ -1,7 +0,0 @@
-package io.getstream.chat.core.internal.fsm.builder
-
-import io.getstream.chat.core.internal.InternalStreamChatApi
-
-@InternalStreamChatApi
-@DslMarker
-public annotation class FSMBuilderMarker
```
zsmb13, line 7: This class was moved and not deleted, weird that the diff didn't pick this up correctly

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
```diff
@@ -4,7 +4,7 @@ import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleObserver
 import androidx.lifecycle.OnLifecycleEvent
 import androidx.lifecycle.ProcessLifecycleOwner
-import io.getstream.chat.core.internal.coroutines.DispatcherProvider
+import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
```
JcMinarro, line 7: Do we need to move it into `android`? :thinking: 
What about if in the future we release a "LLC pure Kotlin"? It could use this `DispatcherProvider` but doesn't have any "android related stuff"

stream-chat-android-offline/build.gradle
```diff
@@ -70,13 +70,13 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
...
                 '-Xexplicit-api=strict',
-                '-Xopt-in=io.getstream.chat.core.internal.InternalStreamChatApi',
+                '-Xopt-in=io.getstream.chat.android.core.internal.InternalStreamChatApi',
         ]
     }
 }
 
 dependencies {
-    implementation project(':stream-chat-android-core')
+    api project(':stream-chat-android-core')
```
JcMinarro, line 79: `stream-chat-android-client` already provide `stream-chat-android-core` as api dependency, do we need to provide it from here? :thinking:

stream-chat-android-test/src/main/java/io/getstream/chat/android/test/TestCall.kt
```diff
@@ -1,10 +1,10 @@
-package io.getstream.chat.android.livedata.utils
+package io.getstream.chat.android.test
 
 import io.getstream.chat.android.client.call.Call
 import io.getstream.chat.android.client.utils.Result
 
-internal class TestCall<T : Any>(val result: Result<T>) : Call<T> {
-    var cancelled = false
+public class TestCall<T : Any>(public val result: Result<T>) : Call<T> {
```
JcMinarro, line 6: Sould it be annotated as `@InternalStreamChatApi` ?

stream-chat-android-ui-components/build.gradle
```diff
@@ -67,13 +67,13 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
...
                 '-Xopt-in=kotlin.contracts.ExperimentalContracts',
-                '-Xopt-in=io.getstream.chat.core.internal.InternalStreamChatApi'
+                '-Xopt-in=io.getstream.chat.android.core.internal.InternalStreamChatApi'
         ]
     }
 }
 
 dependencies {
-    implementation project(':stream-chat-android-core')
+    api project(':stream-chat-android-core')
```
JcMinarro, line 76: Same here

stream-chat-android/build.gradle
```diff
@@ -75,13 +75,13 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
...
                 '-Xexplicit-api=strict',
-                '-Xopt-in=io.getstream.chat.core.internal.InternalStreamChatApi'
+                '-Xopt-in=io.getstream.chat.android.core.internal.InternalStreamChatApi'
         ]
     }
 }
 
 dependencies {
-    implementation project(":stream-chat-android-core")
+    api project(":stream-chat-android-core")
```
JcMinarro, line 84: Same here

JcMinarro (commented):
Look good ;)
I have left some comments there

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
zsmb13: In that case, the `Main` dispatcher wouldn't make any sense, so this is definitely Android specific.

Even if we have things right now that are pure though, I think we can should keep them in the Android packages. Until we're explicitly working towards a pure Kotlin SDK, we're not paying attention to designing for it anyway, so we'd be bound to make incorrect assumptions along the way, putting things into the pure module that are actually platform specific somehow.

stream-chat-android-test/src/main/java/io/getstream/chat/android/test/TestCall.kt
zsmb13: I figured that since we're only including this module as `testImplementation`, it's easier to just have everything in it be public. But if you think people might start using this module themselves (?), then we could add annotations to scare them away from doing so.

zsmb13: Really, we just shouldn't be publishing this module, but Jitpack... Not sure if they allow us to configure this kind of thing, to exclude this module from publication.

stream-chat-android-offline/build.gradle
zsmb13: Looks like we don't need to be explicit about it, removed these.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
ogkuzmin: But right now it's not android specific, is it? Main Dispatcher has meaning not only in android world
image

zsmb13: Yes, technically, the Main dispatcher can map to different things, but I don't see us actually using it for any of those in practice, at least not the same way as we're using it now on Android.

Goes back to the previous point, if we haven't designed for it and figured out how it would work exactly, it's better not to have a pure module for now, because if anything, it'll be misleading to have it and think it's easy to reuse outside of Android.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/ClientStateServiceTests.kt
```diff
@@ -241,8 +242,8 @@ internal class ClientStateServiceTests {
 
         fun givenUserConnectedState(
             user: User = Mother.randomUser(),
-            token: String = Mother.randomString(),
-            connectionId: String = Mother.randomString()
+            token: String = randomString(),
```
ogkuzmin, line 245: I'd like have one mother object and move it to a common module with JFixture. It's more convenient. What do you think?

ogkuzmin: But I can do it later after merge by my own

zsmb13: Agreed, we can take a look at that later on, see if it works

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
JcMinarro: For now, `stream-chat-core` is pure kotlin, no Android stuff related.
The `Dispatcher.Main` is not Android related, you need to indicated which "main dispatcher" you want to use depending on which platform you are

stream-chat-android-test/src/main/java/io/getstream/chat/android/test/TestCall.kt
JcMinarro: I am not sure they provide any configuration to exclude this, I wouldn't take care about it for now, only limit the things we think are not needed for third-party devs

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
zsmb13: The module itself is called `stream-chat-android-core`, it's a core module for Android related things. It's pure Kotlin by coincidence. If we want a pure core module, that really should be named `stream-chat-core` (or `stream-chat-jvm-core`, perhaps). We can add that later if it's actually necessary and useful. For the time being, we have no real need for it.

JcMinarro: Ok, out-of-scope of this PR. We will rename this module in another PR

JcMinarro (approved):

PR #880, opened 2020-11-19
Feature/update chat client observation

JcMinarro:
### Description
We can't delegate observation to the initCallback, because the setUser could fails but have a reconnection process that complete this process. On that case our observer won't be registered. This behavior has been updated to be registered when the connection is done. 
Other corner-case happend here, because our observer could be registered multiple times, that is because our `StreamLifecycleObserver` has been updated as well to only allow observe once

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
@@ -5,7 +5,7 @@ import androidx.lifecycle.LifecycleObserver
 import androidx.lifecycle.OnLifecycleEvent
 import androidx.lifecycle.ProcessLifecycleOwner
 import io.getstream.chat.core.internal.coroutines.DispatcherProvider
-import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 
 internal class StreamLifecycleObserver(private val handler: LifecycleHandler) : LifecycleObserver {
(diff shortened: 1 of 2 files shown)


ogkuzmin (dismissed):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
```diff
@@ -10,19 +10,26 @@ import kotlinx.coroutines.launch
...
+    @Volatile private var isObserving = false
 
     fun observe() {
-        CoroutineScope(DispatcherProvider.Main).launch {
-            ProcessLifecycleOwner.get()
-                .lifecycle
-                .addObserver(this@StreamLifecycleObserver)
+        if (isObserving.not()) {
+            isObserving = true
+            CoroutineScope(DispatcherProvider.Main).launch {
```
ogkuzmin, line 18: Maybe it's good to store scope as a local private property?

JcMinarro: DONE ;)

ogkuzmin (dismissed):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
```diff
@@ -10,19 +10,27 @@ import kotlinx.coroutines.launch
 
 internal class StreamLifecycleObserver(private val handler: LifecycleHandler) : LifecycleObserver {
     private var recurringResumeEvent = false
+    @Volatile private var isObserving = false
+    private val coroutineScope = CoroutineScope(DispatcherProvider.Main)
```
zsmb13, line 14: If this scope is never cancelled, we might as well use `GlobalScope` instead (and specify a dispatcher in the coroutine builder, e.g. `launch`), because it makes no difference, and that's simpler.

JcMinarro: :+1:

ogkuzmin (approved):

zsmb13 (dismissed):

ogkuzmin (approved):

PR #882, opened 2020-11-19
Extract formatting of channel dates into DateFormatter

zsmb13:
### Description

Extracts the date formatting logic that we already had duplicated in two places for the list of channel items (and we'll use in at least two more places in the new design) into a helper class + adds tests.

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolder.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolder.kt
@@ -1,9 +1,9 @@
 package io.getstream.chat.android.ui.channel.list.adapter.viewholder
 
-import android.annotation.SuppressLint
 import android.util.TypedValue
 import android.view.View
 import androidx.core.view.isVisible
+import com.getstream.sdk.chat.utils.DateFormatter
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.ui.R
 import io.getstream.chat.android.ui.channel.list.ChannelListView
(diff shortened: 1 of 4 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolder.kt
```diff
@@ -19,18 +19,9 @@ import io.getstream.chat.android.ui.utils.extensions.getLastMessage
...
-import java.text.SimpleDateFormat
-import java.util.Locale
 
 public class ChannelListItemViewHolder(itemView: View) : BaseChannelListItemViewHolder(itemView) {
 
-    public companion object {
-        @SuppressLint("ConstantLocale")
-        private val DEFAULT_LOCALE: Locale = Locale.getDefault()
-
-        private val TIME_FORMAT = SimpleDateFormat("hh:mm", DEFAULT_LOCALE)
```
zsmb13, line 31: This was actually broken, as `hh` is the format for 12-hour time, so everything past 12:00 had 12 hours subtracted from it

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/DateFormatter.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Date
+import java.util.Locale
+
+@InternalStreamChatApi
+public object DateFormatter {
+
+    @SuppressLint("ConstantLocale")
+    private val DEFAULT_LOCALE = Locale.getDefault()
+    private val DATE_FORMAT: DateFormat = SimpleDateFormat("MMM d", DEFAULT_LOCALE)
+    private val TIME_FORMAT: DateFormat = SimpleDateFormat("HH:mm", DEFAULT_LOCALE)
```
samiuelson, line 17: In case the default `Locale` is `US`. Shouldn't we display time in AM/PM format?

zsmb13: This is a feature request, this is meant to be purely a refactor PR 😄 

Jokes aside, yes, we might want to consider different locales at some point. This way it'll be easier to change to a new implementation if we have to.

samiuelson (approved):
Good improvement 👍🏼

PR #885, opened 2020-11-20
use stateflow to fix chatDomainImpl.clearState

tschellenbach:
Use StateFlow to fix the error triggered by calling clearState from the IO thread:

https://console.firebase.google.com/project/stream-chat-internal/crashlytics/app/android:io.getstream.chat.sample.debug/issues/3d76c3f580802cc57d67fa27be437dcf?time=last-seven-days&sessionEventKey=5FB681280016000139B4D82C531C700B_1475287504195377751

Remove unused typingEvents implementation, design is dropping this feature request.

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -62,7 +62,7 @@ public interface ChatDomain {
      *   })
      *
      */
-    public val errorEvents: LiveData<Event<ChatError>?>
+    public val errorEvents: LiveData<Event<ChatError>>
 
     /**
      * list of users that you've muted
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -62,7 +62,7 @@ public interface ChatDomain {
      *   })
      *
      */
-    public val errorEvents: LiveData<Event<ChatError>>
+    public val errorEvents: LiveData<Event<ChatError>?>
 
     /**
      * list of users that you've muted
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -6,7 +6,6 @@ import android.os.Handler
 import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MediatorLiveData
-import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.asLiveData
 import androidx.room.Room
 import com.google.gson.Gson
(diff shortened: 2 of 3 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -435,36 +429,23 @@ internal class ChatDomainImpl internal constructor(
...
-            _banned.postValue(newBanned)
-        }
+        _banned.value = newBanned
     }
 
     fun setTotalUnreadCount(newCount: Int) {
-        val currentCount = _totalUnreadCount.value ?: 0
-        if (currentCount != newCount) {
-            _totalUnreadCount.postValue(newCount)
-        }
```
tschellenbach, line 467: really nice that this is built-in to stateflow

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -180,20 +178,16 @@ internal class ChatDomainImpl internal constructor(
      *   })
      *
      */
-    override val errorEvents: LiveData<Event<ChatError>> = _errorEvent
+    override val errorEvents: LiveData<Event<ChatError>> = _errorEvent.mapNotNull { it }.asLiveData()
```
zsmb13, line 181:
```suggestion
    override val errorEvents: LiveData<Event<ChatError>> = _errorEvent.filterNotNull().asLiveData()
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -77,8 +76,6 @@ public interface ChatDomain {
     /** The retry policy for retrying failed requests */
     public var retryPolicy: RetryPolicy
 
-    public val typingUpdates: LiveData<TypingEvent>
```
JcMinarro, line 80: Talking with designers, it is not totally dropped from our new design.
On the other hand, it is a public property that we expose in our SDK, we shouldn't break our public API because we don't use it into our new design

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -154,22 +152,22 @@ internal class ChatDomainImpl internal constructor(
...
      * list of users that you've muted
      */
-    override val muted: LiveData<List<Mute>> = _mutedUsers
+    override val muted: LiveData<List<Mute>> = _mutedUsers.asLiveData()
 
     /**
      * if the current user is banned or not
      */
-    override val banned: LiveData<Boolean> = _banned
+    override val banned: LiveData<Boolean> = _banned.asLiveData()
```
JcMinarro, line 170: In the future would be great to provide it as an extra artifact that provides extensions to work with LiveData, and remove LiveData as a dependency of our module

JcMinarro (changes requested):
This bug was already fixed on #796
As you can see all method you are updating on `clearState()` are calling `postValue()` that can be called from whatever thread.
However, migrate from LiveData to StateFlow is a good change

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
tschellenbach: @leandroBorgesFerreira why did you add this? was it for a customer or for the new design?

for our design the right way would be to expose a channel preview class that contains this information at the queryChannelController level.

JcMinarro (dismissed):

JcMinarro (approved):

PR #897, opened 2020-11-23
[CAS-445] Hotfix/channels view render offline

JcMinarro:
### Description
ChannelList weren't using the proper data to be rendered when it was loaded from offline storage

### Checklist

- [x] Reviewers added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@
 
 ## stream-chat-android-client
 - New artifact name: `io.getstream:stream-chat-android-client:STREAM_VERSION`
-
+- Show date of the last message into channels list when data comes from offline storage
+- Show text of the last message into channels list when data comes from offline storage
 
 ## stream-chat-android-offline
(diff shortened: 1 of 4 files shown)


ogkuzmin (dismissed):

adasiewiczr (approved):

tschellenbach:
I suspect there is a second bug as well. I mean we shouldn't be showing the offline version of the data when the API is online. Perhaps an issue with the detection of online/vs offline. (note how this code always shows the results from the online state if the API is online)

image

JcMinarro:
> I suspect there is a second bug as well. I mean we shouldn't be showing the offline version of the data when the API is online. Perhaps an issue with the detection of online/vs offline. (note how this code always shows the results from the online state if the API is online)
> 
> image

If we are online we are using online data, you can see it here: https://github.com/GetStream/stream-chat-android/blob/master/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt#L233

PR #904, opened 2020-11-25
[CAS-425] Add creating new DM channel logic 

adasiewiczr:
PR adds`AddChannelViewModel` which is responsible for providing data for AddUserView
and handling channel creation process.
Creation channel flow can be described in following steps:
1. Create draft channel when member is added
2. Update draft channel (remove draft flag) after sending the message
Draft channels are required to show past messages and they are not visible for user - they are filtered during queryChannels request.
Creating new DM channels flow works only while user is online.
In the next PR I'm going to move all `AddChannel*` related views to the sample app because they are custom and probably won't be used by our customer

Screenshot 2020-11-25 at 12 59 41

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelViewModel.kt
```diff
@@ -0,0 +1,139 @@
...
+import io.getstream.chat.android.client.api.models.QuerySort
+import io.getstream.chat.android.client.api.models.QueryUsersRequest
+import io.getstream.chat.android.client.channel.ChannelClient
+import io.getstream.chat.android.client.models.Filters
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.livedata.ChatDomain
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+
+class AddChannelViewModel : ViewModel() {
```
ogkuzmin, line 17: It should be in another module. shouldn't it?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelViewModel.kt
```diff
@@ -0,0 +1,139 @@
...
+            is Event.MembersChanged -> createDraftChannel(event.members)
+        }
+    }
+
+    private fun requestUsers(isRequestingMore: Boolean) {
+        if (!isRequestingMore) {
+            _state.value = State.Loading
+        }
+        viewModelScope.launch(Dispatchers.IO) {
+            val result = ChatClient.instance().queryUsers(
```
ogkuzmin, line 47: What's about offline support? Is it suppose to be added later or we don't think about it now?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelViewModel.kt
```diff
@@ -0,0 +1,139 @@
...
+            Event.MessageSent -> createChannel()
+            is Event.MembersChanged -> createDraftChannel(event.members)
+        }
+    }
+
+    private fun requestUsers(isRequestingMore: Boolean) {
+        if (!isRequestingMore) {
+            _state.value = State.Loading
+        }
+        viewModelScope.launch(Dispatchers.IO) {
```
ogkuzmin, line 46: We agreed to use `DispatchersProvider` instead of directly using `Dispatchers`

ogkuzmin (commented):
This PR brings a lot of new logic. I think it's better to add some unit tests for that. What do you think @adasiewiczr ?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelViewModel.kt
adasiewiczr: @ogkuzmin this one is in the right module. We are not gonna include AddChannelView in our SDK for now and I'm going to move all other related classes to the sample app in a separate PR

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelViewModel.kt
adasiewiczr: I think we should add support for this at some point

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelViewModel.kt
adasiewiczr: `DispatchersProvider` is internal StreamAPI and I don't think we should use it in sample app

stream-chat-android-ui-components/src/main/res/values/strings.xml
```diff
@@ -33,16 +33,19 @@
     <string name="stream_add_channel_back_icon_description">Back</string>
     <string name="stream_add_channel_add_member_icon_description">Add member</string>
     <string name="stream_add_channel_user_checked_description">User checked</string>
+    <string name="stream_add_channel_empty_users_description">No users</string>
     <string name="stream_add_channel_title">New Chat</string>
     <string name="stream_add_member_title">TO:</string>
     <string name="stream_add_member_hint">Type a name</string>
     <string name="stream_create_group">Create a Group</string>
     <string name="stream_add_channel_user_list_title">On the platform</string>
+    <string name="stream_add_channel_empty_users_title">No users matches these keywords…</string>
```
ogkuzmin, line 42: Mistypo here should be `No user_`
image

adasiewiczr:
@ogkuzmin I'm not sure if we want to write tests for features that are purely implemented in sample app and all `AddChannel*` related classes will be moved to sample in next PR. I just didn't want to include moving all of them in this

stream-chat-android-ui-components/src/main/res/values/strings.xml
adasiewiczr: nice catch!

ogkuzmin:
Discussed it. Since it's only for sample, there is no need for tests

ogkuzmin (approved):

ogkuzmin (approved):

PR #910, opened 2020-11-26
[CAS-424] Add search to add DM channel view

adasiewiczr:
Screenshot 2020-11-26 at 11 54 40

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/common/Debouncer.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+
+/**
+ * Utility class for debouncing high frequency events.
+ *
+ * [submit]ting a new piece of work to run within the debounce window
+ * will cancel the previously submitted pending work.
+ */
+class Debouncer(private val debounceMs: Long) {
```
adasiewiczr, line 16: Copy-paste from https://github.com/GetStream/stream-chat-android/blob/master/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/Debouncer.kt

ogkuzmin (dismissed):

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/common/Debouncer.kt
JcMinarro: What about to open the other one? 
Duplicate this class will be hard to maintain.

adasiewiczr: @JcMinarro we can do that, but I am not sure if we want to expose `Debouncer` to our SDK clients. What do you think?

JcMinarro: I think it is fine, it is into our own package name, so shouldn't be any issue with that

JcMinarro (approved):

samiuelson (approved):

PR #914, opened 2020-11-26
Message diff

tschellenbach:
We were previously triggering many rerenders of the message list item. Here's a list of the most common causes for rerendering the message:

```
- Message.user can have
-- different unread counts, we shouldn't rerender when that happens
-- different last online values, again shouldn't trigger a rerender
- The user objects on message.latestReactions and message.ownReactions have the same issue with last online and unread counts values on the user object
- messageReadBy returns a list of channel user read objects, the problem is that they contain unread message counts which can trigger a diff
- reaction.updated at wasn't stored in the reaction entity, so the offline & online version were always different
```

**Solutions:**
- MessageListItemDiffCallback implements areContentsTheSame focused on the UI that we use
- MessageListItemDiffCallback implements getChangePayload
- fix reaction to reaction entity conversion to use reaction.updated at

### DIFF
--- a/stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemDiffCallback.kt
+++ b/stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemDiffCallback.kt
@@ -1,6 +1,7 @@
 package com.getstream.sdk.chat.adapter
 
 import androidx.recyclerview.widget.DiffUtil
+import io.getstream.chat.android.client.models.User
 
 internal object MessageListItemDiffCallback : DiffUtil.ItemCallback<MessageListItem>() {
     override fun areItemsTheSame(oldItem: MessageListItem, newItem: MessageListItem): Boolean {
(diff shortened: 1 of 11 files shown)


stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemDiffCallback.kt
```diff
@@ -1,13 +1,59 @@
...
+                    false
+                } else if (oldItem.message.deletedAt != newItem.message.deletedAt) {
+                    false
+                } else if (oldItem.positions != newItem.positions) {
+                    false
+                } else oldItem.messageReadBy.map { it.getUserId() } == newItem.messageReadBy.map { it.getUserId() }
+            }
+            is MessageListItem.DateSeparatorItem, is MessageListItem.ThreadSeparatorItem, is MessageListItem.LoadingMoreIndicatorItem -> equal = true
+            is MessageListItem.TypingItem -> oldItem.users.map(User::id) == ((newItem) as MessageListItem.TypingItem).users.map(User::id)
+            is MessageListItem.ReadStateItem -> oldItem.reads.map { it.getUserId() } == ((newItem) as MessageListItem.ReadStateItem).reads.map { it.getUserId() }
```
adasiewiczr, line 36: Shouldn't we update `equal` variable here?  Maybe we can also get rid of `equal` variable and return the boolean directly from `when` expression?

JcMinarro: Good catch!!!

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/viewholder/message/MessageListItemViewHolder.kt
```diff
@@ -74,71 +72,32 @@ internal class MessageListItemViewHolder(
...
+            spaceConfigurator,
+            marginConfigurator,
+            messageTextConfigurator.takeIf { diff.text || diff.positions || diff.deleted || diff.reactions },
+            attachmentConfigurator.takeIf { diff.attachments },
+            reactionConfigurator.takeIf { diff.reactions },
+            replyConfigurator.takeIf { diff.replies },
+            indicatorConfigurator.takeIf { diff.syncStatus || diff.readBy },
+            userAvatarConfigurator.takeIf { diff.positions },
+            usernameAndDateConfigurator.takeIf { diff.positions }
+        ).forEach { it.configure(messageListItem) }
```
tschellenbach, line 101: cool trick

PR #918, opened 2020-11-27
clarify state to prevent bugs at the UI layer

tschellenbach:
1. We currently have UI bugs where we show "no result" instead of a loading icon. This leads to weird UI changes in both the message list and channel list

2. The root cause of this problem is confusion between the offline storage side and the UI components. Oleg had the excellent suggestion of using a sealed class for this. I think we should do that for both the queryChannelController.channels and channelController.messages

3. This is a draft, needs input from @JcMinarro on the right architecture. I just think it's important for efficient collaboration between teams and help solve one of the more noticable issues in our sample app (the weird UI flickring)

Video about the message list problem: https://www.loom.com/share/fc3877a260944c528f00d28ef385dc4c
Same issue applies to the channel list


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -112,19 +114,30 @@ internal class ChannelControllerImpl(
...
+    override val messages: LiveData<List<Message>> = sortedVisibleMessages.asLiveData()
+
+    private val _messagesState: StateFlow<ChannelController.MessagesState> =
+        _loading.combine(sortedVisibleMessages) { loading: Boolean, messages: List<Message> ->
+            when {
+                loading -> ChannelController.MessagesState.Loading
+                messages.isEmpty() -> ChannelController.MessagesState.OfflineNoResults
+                else -> ChannelController.MessagesState.Result(messages)
+            }
+        }.stateIn(domainImpl.scope, SharingStarted.Eagerly, ChannelController.MessagesState.NoQueryActive)
```
tschellenbach, line 127: Good to discuss this since the design/UI side of things can be different.

- some UIs show the loading icon overlayed on the chat messages. So showing it while there are results works well.
- other UIs only show the loading icon when there are no results available yet.. 

If i look at our designs it looks like we're trying to do the 2nd, this implement of state is different than those 2 options. (it triggers the loading interface once for offline and once for online)

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
```diff
@@ -51,8 +51,7 @@ import java.util.Date
  */
 internal class MessageListItemLiveData(
     private val currentUser: User,
-    newMessagesLd: LiveData<List<Message>>,
-    oldMessagesLd: LiveData<List<Message>>,
```
ogkuzmin, line 55: I like it, it seemed weird that we had two sources for actual and old messages

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/channels/ChannelsViewModel.kt
```diff
@@ -44,17 +43,19 @@ public class ChannelsViewModel(
...
                 val queryChannelsController = queryChannelsControllerResult.data()
                 stateMerger.addSource(
-                    map(queryChannelsController.channels) { channelList ->
-                        if (channelList.isEmpty()) {
-                            State.NoChannelsAvailable
-                        } else {
-                            State.Result(channelList.filter { it.hidden == false })
+                    map(queryChannelsController.channelsState) { channelState ->
+                        when (channelState) {
+                            is QueryChannelsController.ChannelsState.NoQueryActive,
```
ogkuzmin, line 53: So neat :heart_eyes:

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -112,19 +114,30 @@ internal class ChannelControllerImpl(
...
+    override val messagesState = _messagesState.asLiveData()
 
-    override val oldMessages: LiveData<List<Message>> = messagesTransformation(_oldMessages)
+    override val oldMessages: LiveData<List<Message>> = messagesTransformation(_oldMessages).asLiveData()
 
-    private fun messagesTransformation(messages: MutableStateFlow<Map<String, Message>>): LiveData<List<Message>> {
+    private fun messagesTransformation(messages: MutableStateFlow<Map<String, Message>>): Flow<List<Message>> {
```
ogkuzmin, line 132: Nit: with flows, it's convenient to use such things as extensions. It let you chain operators in a row

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
ogkuzmin: Good point, we can split loading for two cases just `Loding` and `LoadingWithCached(val cachedResult: Collection<Message>)`

ogkuzmin (commented):
Nice change, this logic flow is good typed and got clearer

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: This flow is following the 2nd option.
Loading is only returned when the first query is been done, doesn't mind if it is online or offline (both are done in parallel)
We have another livedata that represent that there is another query in process, used to show the "loading more" progressbar

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImplTest.kt
```diff
@@ -73,13 +81,14 @@ internal class QueryChannelsControllerImplTest {
         }
 }
 
-private class Fixture {
+private class Fixture(scope: CoroutineScope) {
     private val chatClient: ChatClient = mock()
     private val chatDomainImpl: ChatDomainImpl = mock()
```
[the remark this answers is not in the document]
JcMinarro: The coroutineScope depend on the test itself, it is created by jUnit with the extension, we can access directly to this scope from inside of this class, so it needs to be received by constructor

stream-chat-android/src/main/java/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
JcMinarro: It was used for "new messages counter", but it wasn't working properly, and at the end it wasn't needed at all

ogkuzmin (dismissed):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -113,19 +115,30 @@ internal class ChannelControllerImpl(
     val unfilteredMessages = _messages.map { it.values.toList() }
 
     /** a list of messages sorted by message.createdAt */
-    override val messages: LiveData<List<Message>> = messagesTransformation(_messages)
+    val sortedVisibleMessages = messagesTransformation(_messages).stateIn(domainImpl.scope, SharingStarted.Eagerly, emptyList())
```
zsmb13, line 118: Could be private, could benefit from having an explicit type

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
```diff
@@ -67,4 +76,29 @@ public interface ChannelController {
...
+         *
+         * @see loading
+         * @see loadingNewerMessages
+         * @see loadingOlderMessages
+         * */
+        public object Loading : MessagesState()
+        /** If we are offline and don't have channels stored in offline storage, typically displayed as an error condition. */
+        public object OfflineNoResults : MessagesState()
+        /** If we have no offline data and get an error from the API */
+        public class Failed(public val error: ChatError) : MessagesState()
```
zsmb13, line 97: It looks like we never go into this `Failed` state?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
```diff
@@ -39,8 +41,20 @@ public interface QueryChannelsController {
     public val endOfChannels: LiveData<Boolean>
     /**
      * The list of channels
+     * Typically we recommend using channelsState instead, it's a bit more complex but ensures
```
zsmb13, line 44:
```suggestion
     * Typically we recommend using [channelsState] instead, it's a bit more complex but ensures
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
```diff
@@ -49,4 +63,28 @@ public interface QueryChannelsController {
...
+         * For seeing if we're loading more results have a look at QueryChannelsController.loadingMore
+         *
+         * @see QueryChannelsController.loadingMore
+         * @see QueryChannelsController.loading
+         * */
+        public object Loading : ChannelsState()
+        /** If we are offline and don't have channels stored in offline storage, typically displayed as an error condition. */
+        public object OfflineNoResults : ChannelsState()
+        /** If we have no offline data and get an error from the API */
+        public class Failed(public val error: ChatError) : ChannelsState()
```
zsmb13, line 83: This `Failed` state also seems unused at the moment

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/BaseDomainTest.kt
```diff
@@ -239,6 +253,7 @@ internal open class BaseDomainTest {
                 return 1000
             }
         }
+        chatDomainImpl.scope = testCoroutines.scope
```
zsmb13, line 256: This should not be necessary, we're already mocking the dispatcher that's being used inside this scope via `DispatcherProvider` and `TestCoroutineRule`. Let's let the `ChatDomainImpl` create a real scope here, as it usually does. `scope` should actually be a `val` in that class, not sure why it's a `var` at the moment.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
zsmb13: I feel like I'm obligated to plug [my own content](https://zsmb.co/designing-and-working-with-single-view-states-on-android/) about distinct vs non-distinct states, as that's exactly what this discussion is about.

TLDR: There's nothing wrong with having a regular class with both a `loading` and `results` property at the same time, if those things are present at the same time on the UI. Not everything is neatly modellable by distinct cases of sealed classes, however awesome those are.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
tschellenbach: We should trigger the failed state when you hit an API error and no local storage results are available. This is important to get right since it can be very confusing if we show things like an infinite loading icon when there is in fact an error.

zsmb13: @JcMinarro ^

JcMinarro: Failure state is not needed in our current implementation.
We have 2 ways to obtain data, remote datasource (Stream Server) and local datasource (DB) that are run in parallel. If both of them fail, we will obtain an `OfflineNoResults` state, and the query will be marked as "recovery needed", so no more infinite loading state could be there. If a controller is marked as "recovery needed" (because a query failed) it is because there was a network connection failure, when the connection is live again or any other events come to the app, the query will be processed again. Because we are using a react approach from our DB, any new message/channel is inserted in our DB will trigger an update in our controller

zsmb13 (approved):
Good to go as far as I'm concerned, let's have multiple approvals on this one though

tschellenbach:
Showing an offline no result state seems not ideal from when you get an
error from the API.

PR #922, opened 2020-11-30
[CAS-369] Add Search Result list

zsmb13:
https://stream-io.atlassian.net/browse/CAS-369

### Description

More work on search and search results.

- Renamed `SearchView` to `SearchInputView` to better represent what it does
- Added `SearchViewModel` and `SearchResultListView` to execute search queries and display their results

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+        return StreamItemMentionListBinding
+            .inflate(LayoutInflater.from(parent.context), parent, false)
+            .let { MessagePreviewViewHolder(it.root) }
+    }
+
+    override fun onBindViewHolder(holder: MessagePreviewViewHolder, position: Int) {
+        holder.bind(getItem(position))
+    }
+
+    public fun setSearchResultSelectedListener(searchResultSelectedListener: SearchResultSelectedListener?) {
```
adasiewiczr, line 28: Is `searchResultSelectedListener` nullable on purpose?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+import androidx.recyclerview.widget.RecyclerView
+import io.getstream.chat.android.client.models.Message
+import io.getstream.chat.android.ui.databinding.StreamItemMentionListBinding
+import io.getstream.chat.android.ui.messagepreview.MessagePreviewView
+import io.getstream.chat.android.ui.search.SearchResultListAdapter.MessagePreviewViewHolder
+import io.getstream.chat.android.ui.search.SearchResultListView.SearchResultSelectedListener
+
+public class SearchResultListAdapter : ListAdapter<Message, MessagePreviewViewHolder>(MessageDiffCallback) {
+
+    private var searchResultSelectedListener: SearchResultSelectedListener? = null
```
adasiewiczr, line 16: What do you think about making it non-nullable and set an empty listener by default?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+    override fun onBindViewHolder(holder: MessagePreviewViewHolder, position: Int) {
+        holder.bind(getItem(position))
+    }
+
+    public fun setSearchResultSelectedListener(searchResultSelectedListener: SearchResultSelectedListener?) {
+        this.searchResultSelectedListener = searchResultSelectedListener
+    }
+
+    public inner class MessagePreviewViewHolder(
+        private val view: MessagePreviewView
```
adasiewiczr, line 33: Didn't we agree once that we are going to pass binding instead of views?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListView.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    private fun init(attrs: AttributeSet?) {
+        parseAttrs(attrs)
+
+        orientation = VERTICAL
+
+        binding.searchListView.apply {
+            setHasFixedSize(true)
+            adapter = this@SearchResultListView.adapter
+            layoutManager = LinearLayoutManager(context).apply {
```
adasiewiczr, line 43: Just out of curiosity - why not setting layout manager and orientation in XML?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListAdapter.kt
zsmb13: It is! It's `null` by default anyway, and this way you can unset it by passing in `null`, if you want to.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListAdapter.kt
zsmb13: It is! It's null by default anyway, and this way you can unset it by passing in null, if you want to.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListAdapter.kt
zsmb13: It's easier to clear the listener this way, otherwise you'd have to pass in your own empty implementation to stop listening to events. Same behaviour as `View#setOnClickListener`, which also allows you to set to `null` to clear it.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListAdapter.kt
zsmb13: You're right, changed to passing in the binding, that's more consistent 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListView.kt
zsmb13: Oops, this was leftover code from when the entire View itself was a `RecyclerView`, and XML config wasn't available on the `merge` tag 😄 . Moved to XML.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListView.kt
```diff
@@ -0,0 +1,72 @@
...
+    ) {
+        init(attrs)
+    }
+
+    private val adapter = SearchResultListAdapter()
+
+    private fun init(attrs: AttributeSet?) {
+        parseAttrs(attrs)
+
+        orientation = VERTICAL
```
adasiewiczr, line 38: NIT: This can be probably removed as it is already set in XML

adasiewiczr (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/search/SearchResultListView.kt
zsmb13: This is for the root LinearLayout, this still has to be set from code.

`SearchResultListView` is a vertical `LinearLayout`, which has a `TextView`("X results") and a `RecyclerView` inside it.

PR #924, opened 2020-11-30
Clean up some settings in build.gradle files

zsmb13:
### Description

A quick cleanup of some of our build.gradle configuration. See inline comments for explanations.

### Checklist

- [x] Reviewers added


stream-chat-android-client/build.gradle
```diff
@@ -19,12 +19,11 @@ ext {
...
 android {
-    compileSdkVersion 28
+    compileSdkVersion Configuration.compileSdkVersion
     buildToolsVersion "29.0.3"
     defaultConfig {
-        minSdkVersion 21
-        targetSdkVersion 28
-        versionCode 1
+        minSdkVersion Configuration.minSdkVersion
+        targetSdkVersion Configuration.targetSdkVersion
```
zsmb13, line 26: These values were previously set directly both here and in the `offline` module. As a result of this, the target and compile SDKs were already lagging behind the other modules.

stream-chat-android-client/build.gradle
```diff
@@ -48,19 +47,17 @@ android {
...
         targetCompatibility JavaVersion.VERSION_1_8
     }
-
-    testOptions.unitTests {
-        returnDefaultValues = true
-    }
-
     kotlinOptions {
         jvmTarget = '1.8'
-        noStdlib = true
```
zsmb13, line 58: A setting that we don't need to care about, plus it's `true` by default

stream-chat-android/build.gradle
```diff
@@ -17,26 +17,17 @@ ext {
...
 
     defaultConfig {
+        minSdkVersion Configuration.minSdkVersion
+        targetSdkVersion Configuration.targetSdkVersion
         versionName Configuration.versionName
 
-        buildConfigField "String", "DEFAULT_API_ENDPOINT", "\"$DEFAULT_API_ENDPOINT\""
-        buildConfigField "int", "DEFAULT_API_TIMEOUT", "$DEFAULT_API_TIMEOUT"
-        buildConfigField "int", "DEFAULT_CDN_TIMEOUT", "$DEFAULT_CDN_TIMEOUT"
-        buildConfigField "int", "DEFAULT_API_KEEP_ALIVE_TIMEOUT", "$DEFAULT_API_KEEP_ALIVE_TIMEOUT"
```
zsmb13, line 34: As far as I can tell, none of these were used in the project anymore

stream-chat-android-offline/build.gradle
```diff
@@ -19,21 +19,24 @@ ext {
...
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion Configuration.compileSdkVersion
     buildToolsVersion "29.0.3"
     defaultConfig {
-        minSdkVersion 21
-        targetSdkVersion 28
-        versionCode 1
+        minSdkVersion Configuration.minSdkVersion
```
bychkovdmitry, line 25: What do you think, is it a good idea to apply `buildToolsVersion`, `compileSdkVersion`, `compileOptions`, `minSdkVersion`, `targetSdkVersion`, (optionally kotlin source sets) to all subprojects in the root level gradle build script?

stream-chat-android-client/build.gradle
```diff
@@ -19,12 +19,11 @@ ext {
 apply from: "${rootProject.projectDir}/scripts/publish-mavencentral.gradle"
 
 android {
-    compileSdkVersion 28
+    compileSdkVersion Configuration.compileSdkVersion
     buildToolsVersion "29.0.3"
```
bychkovdmitry, line 23: Can also be extracted to a variable

bychkovdmitry (approved):
LGTM

stream-chat-android-client/build.gradle
zsmb13: Oops, left these here accidentally. They can just be removed, the plugin sets these implicitly anyway.

stream-chat-android-offline/build.gradle
zsmb13: Should be possible - there's a whole bunch of things that we could extract into common configuration. Don't want to touch too many things for now, just wanted to apply the quick fix of using the configured values instead of individual hardcoded ones.

bychkovdmitry (approved):

PR #929, opened 2020-12-01
Fix showing search result list on add channel view

adasiewiczr:
Before:
before

After:
after

### Checklist

- [ ] Reviewers added


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelFragment.kt
```diff
@@ -67,51 +67,29 @@ class AddChannelFragment : Fragment() {
...
-                    }
                     is AddChannelViewModel.State.ShowChannel -> initializeChannel(state.cid)
                     AddChannelViewModel.State.HideChannel -> cleanChannel()
                     is AddChannelViewModel.State.NavigateToChannel -> findNavController().navigateSafely(
                         AddChannelFragmentDirections.actionOpenChat(state.cid, null)
                     )
+                    AddChannelViewModel.State.Loading,
+                    AddChannelViewModel.State.Empty,
+                    is AddChannelViewModel.State.Result,
+                    is AddChannelViewModel.State.ResultMoreUsers -> Unit
```
zsmb13, line 82: Did these become unused now, should we remove them?

adasiewiczr: The VM is reused between to *AddChannelFrahments. Some states are common for both of them (and they are handled in bind view function) and some of them are not (handled directly in AddChsnnelFragment). That's why we have Unit here. View model design could be done in a better way, but I would leave it for now as this is a part of sample app and we can come back to this later on

zsmb13 (approved):

Issue #925, opened 2020-11-30
Camera crashes the app


CrispBull:
**Describe the bug**
On clicking the camera attachment icon in the messageInputView, the app crashes

**SDK version**
- 4.4.5

**To Reproduce**
Just running the app and clicking the icon, it crashes the app with no log shown. Ironically the sample app doesn't do this. What could I be doing wrong?

```kotlin
class StreamChatDetailFragment: Fragment() {


        override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
              super.onViewCreated(view, savedInstanceState)
              channelMessagesViewModel.bindView(chatDetailBinding.messageList, viewLifecycleOwner)
              messageInputViewModel.bindView(chatDetailBinding.messageInput, viewLifecycleOwner)
    }
}

```

I don't know what else I need to do

**Expected behavior**
On selecting the camera icon in the message input view and granting permission, I should be able to se the camera launched

**Device:**
 - Google Pixel 3 XL
 - Android version: 10

zsmb13:
The part where the crash produces no logs seems like a very odd problem, perhaps an issue with the specific device or logcat somehow. Can't see any reason why this would happen.

It would be really useful if you could provide a crash log or a minimal example project that reproduces the issue. The code provided so far seems fine on its own.

---

Just to further confirm, you're tapping the "Upload from a camera" option here, and that's what causes the crash?

![image](https://user-images.githubusercontent.com/12054216/100745656-e6edcd80-33df-11eb-8a86-186b1e189df9.png)

It also sounds like you've already granted permission, so you were prompted for permissions when you first tapped that button, granted them, and then the crash occurred?

zsmb13:
For the "no logs" issue: please make sure that you're not filtering the logs in any way. You should have no search input for Logcat, and "No Filters" set, like so:

![image](https://user-images.githubusercontent.com/12054216/100854549-6f747880-3489-11eb-9087-c9995fe53210.png)

zsmb13:
Any news on this, @CrispBull?

CrispBull:
@zsmb13 I have tried doing this and still get the same state of nothing being logged.

zsmb13:
We'll ship some fixes for FileProviders in the next release, which will be out later today, hopefully that will fix your issues with the camera.

In the meantime, if you could provide a minimal sample project that reproduces the problem you're having, we could look into it further, to verify the cause.

JcMinarro:
Hello @CrispBull 

[...]

PR #934, opened 2020-12-02
FileProvider updates

zsmb13:
### Description

Updates some usages of FileProvider after reviewing it due to https://github.com/GetStream/stream-chat-android/issues/925

- Use separate `StreamFileProvider` class instead of the androidx `FileProvider` ([as described here](https://commonsware.com/blog/2017/06/27/fileprovider-libraries.html))
- Update the way we create the authority string, fetch it from `PackageManager` instead of assembling it ourselves

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android/src/main/java/com/getstream/sdk/chat/CaptureMediaContract.kt
+++ b/stream-chat-android/src/main/java/com/getstream/sdk/chat/CaptureMediaContract.kt
@@ -46,6 +46,12 @@ public class CaptureMediaContract : ActivityResultContract<Unit, File>() {
     private fun createFileName(prefix: String, extension: String) =
         "${prefix}_${dateFormat.format(Date().time)}.$extension"
 
+    private fun getFileProviderAuthority(context: Context): String {
+        val compName = ComponentName(context, StreamFileProvider::class.java.name)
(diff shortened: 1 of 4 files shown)


stream-chat-android/src/main/java/com/getstream/sdk/chat/CaptureMediaContract.kt
```diff
@@ -46,14 +46,20 @@ public class CaptureMediaContract : ActivityResultContract<Unit, File>() {
     private fun createFileName(prefix: String, extension: String) =
         "${prefix}_${dateFormat.format(Date().time)}.$extension"
 
+    private fun getFileProviderAuthority(context: Context): String {
+        val compName = ComponentName(context, StreamFileProvider::class.java.name)
```
samiuelson, line 50: `StreamFileProvider` class should have a different name after obfuscation. Won't it break anything?

zsmb13: The `StreamFileProvider::class` reference should still point to the class correctly, even if it does get obfuscated. Since the provider is an entry point for the app, I'd expect it not to be obfuscated in the first place though.

samiuelson (approved):

PR #936, opened 2020-12-02
CAS-429 User reactions view

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-429

### Description

This PR only covers the user reactions card on "reaction screen". The white outline around `ReactionView` can be done later.

The overlay for "reaction screen" is similar to the one used in "message actions". However, the work on "message actions" overlay is still in progress, therefore putting the work on "reaction screen" on hold till it's done.  

### Checklist

- [X] Reviewers added

| Mentions screen | Message Preview component |
| --- | --- |
|photo_2020-12-02 13 25 06 |photo_2020-12-02 13 25 09
 |


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionItemDiffCallback.kt
```diff
@@ -0,0 +1,16 @@
+package io.getstream.chat.android.ui.messages.reactions
+
+import androidx.recyclerview.widget.DiffUtil
+
+public class ReactionItemDiffCallback : DiffUtil.ItemCallback<ReactionItem>() {
```
zsmb13, line 5: `internal`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionClickListener.kt
```diff
@@ -0,0 +1,7 @@
+package io.getstream.chat.android.ui.messages.reactions
+
+import io.getstream.chat.android.client.models.Reaction
+
+public fun interface ReactionClickListener {
+    public fun onReactionClick(reaction: Reaction)
+}
```
zsmb13, line 7: Was the intention here to reuse this in multiple places? `ReactionsView` still has its own copy of this interface nested inside it.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionAdapter.kt
```diff
@@ -0,0 +1,82 @@
...
+        fun bind(reaction: ReactionItem) {
+            this.reactionItem = reaction
+            bindUserAvatar()
+            bindUserName()
+            bindUserReaction()
+        }
+
+        private fun bindUserAvatar() {
+            val user = reactionItem.reaction.user
+            binding.avatarView.setUserData(user ?: User())
```
zsmb13, line 53: Can this happen, is there a chance that a reaction belongs to no user? If it can, shouldn't we filter out these reactions before we set them into the adapter?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionAdapter.kt
```diff
@@ -0,0 +1,82 @@
...
+                reactionContainer.updateConstraints {
+                    clear(R.id.reactionsView, ConstraintSet.START)
+                    clear(R.id.reactionsView, ConstraintSet.END)
+                }
+
+                val params = reactionsView.layoutParams as ConstraintLayout.LayoutParams
+                if (reactionItem.isMine) {
+                    reactionsView.setOrientation(ReactionsView.Orientation.LEFT)
+                    params.endToEnd = ConstraintSet.PARENT_ID
+                    params.marginEnd = context.getDimension(R.dimen.stream_spacing_small)
```
zsmb13, line 71: The start and end margins could probably be fixed values, we don't need to update them dynamically, it's enough to change the constraints

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionsView.kt
```diff
@@ -0,0 +1,75 @@
...
+    ) {
+        init()
+    }
+
+    public fun setMessage(message: Message) {
+        bindTitle(message)
+        bindReactionList(message)
+    }
+
+    public fun setReactionClickListener(reactionClickListener: ReactionsView.ReactionClickListener) {
```
zsmb13, line 43: I'd allow nullable parameters here, just so you can clear the listener easily - the property already allows for it

stream-chat-android-ui-components/src/main/res/values/strings.xml
```diff
@@ -54,4 +54,9 @@
 
     <!-- Message list -->
     <string name="stream_message_deleted_label">Message deleted</string>
+
+    <plurals name="stream_user_reactions_title">
+        <item quantity="one">%d Message reaction</item>
+        <item quantity="other">%d Message reactions</item>
```
zsmb13, line 60: According to the design, these should be

```suggestion
        <item quantity="one">%d Message Reaction</item>
        <item quantity="other">%d Message Reactions</item>
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/View.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import androidx.constraintlayout.widget.ConstraintLayout
+import androidx.constraintlayout.widget.ConstraintSet
+
+internal inline fun ConstraintLayout.updateConstraints(actions: ConstraintSet.() -> Unit) {
+    val set = ConstraintSet()
+    set.clone(this)
+    set.actions()
+    set.applyTo(this)
+}
```
zsmb13, line 11: This is already there in `stream-chat-android`, we could make it `@InternalStreamChatApi` to reuse it here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Context.kt
```diff
@@ -28,3 +29,6 @@ internal fun Context?.getFragmentManager(): FragmentManager? {
         else -> null
     }
 }
+
+internal inline val Context.inflater: LayoutInflater
+    get() = LayoutInflater.from(this)
```
zsmb13, line 34: Same here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionsView.kt
```diff
@@ -0,0 +1,75 @@
...
+            reactionCount,
+            reactionCount
+        )
+    }
+
+    private fun bindReactionList(message: Message) {
+        val layoutManager = binding.recyclerView.layoutManager as GridLayoutManager
+        layoutManager.spanCount = message.latestReactions
+            .size
+            .coerceAtMost(MAX_COLUMNS_COUNT)
```
zsmb13, line 64: Nicely done 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionsView.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        val reactionItems = message.latestReactions
+            .map { ReactionItem(it, message.ownReactions.contains(it)) }
+            .toMutableList()
```
zsmb13, line 68: Why do we need `.toMutableList()` here?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionsView.kt
```diff
@@ -56,6 +56,15 @@ public abstract class ReactionsView : RecyclerView {
...
     }
 
+    public fun setReaction(reaction: Reaction, isMyReaction: Boolean = false) {
+        // according to the design, current user reactions have the same style
+        // as reactions on the current user messages in the message list
+        this.isMyMessage = isMyReaction
+
+        setPadding(0, 0, 0, 0)
+        reactionsAdapter.submitList(listOf(ReactionItem(reaction, true)))
+    }
```
zsmb13, line 66: Am I understanding it correctly that we're using a separate instance of this RecyclerView (`ReactionsView`) for every reaction shown in `UserReactionsView`?

Maybe we should extract `ReactionViewHolder`'s layout into a View, and use that inside `UserReactionsView`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionAdapter.kt
bychkovdmitry: Good idea, I think reaction without user is an edge case. Maybe this can happen when user was deleted?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionClickListener.kt
bychkovdmitry: Yep, making careless mistakes. Removed redundant `ReactionClickListener` in `ReactionsView `

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionAdapter.kt
bychkovdmitry: Did that as clearing constraints seemed to clear the margins as well. Will look into it once again.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionAdapter.kt
bychkovdmitry: According to the discussion I've found, this was fixed on the server side a year ago. Added double-bang operator.

zsmb13: Would it be overkill to make the property non-nullable in the `Reaction` class? Sounds like that should also be fine, if it's guaranteed to be non-null on the server side.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionAdapter.kt
zsmb13: Oh, that might be the case, I didn't realize that would happen.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/UserReactionAdapter.kt
bychkovdmitry: Added filtering of reactions without user, just in case

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionsView.kt
zsmb13: We discussed this, and it would cause a lot of duplication, and the performance isn't necessarily an issue with the current implementation.

zsmb13 (approved):

PR #940, opened 2020-12-02
Fix crash on logout while in pending state

bychkovdmitry:
### Description

Steps to reproduce:
1) Turn off internet connection
2) Login
3) Logout
4) The app crashes with the following stacktrace:

```
 java.lang.IllegalStateException: Cannot handle event io.getstream.chat.android.client.clientstate.ClientStateService$ClientStateEvent$DisconnectRequestedEvent@4d413ac while being in inappropriate state io.getstream.chat.android.client.clientstate.ClientState$User$Pending$WithToken@f52f875
```

### Checklist

- [X] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientStateService.kt
```diff
@@ -52,13 +52,7 @@ internal class ClientStateService {
             state<ClientState.User.Pending.WithToken> {
                 onEvent<ClientStateEvent.ConnectedEvent> { state, event -> ClientState.User.Authorized.Connected(event.connectionId, event.user, state.token) }
                 onEvent<ClientStateEvent.DisconnectedEvent> { _, _ -> stay() }
-                onEvent<ClientStateEvent.ConnectedEvent> { state, event ->
-                    ClientState.User.Authorized.Connected(
-                        event.connectionId,
-                        event.user,
-                        state.token
-                    )
-                }
```
zsmb13, line 61: I get the new line being added here, but why is the handling for `ConnectedEvent` being removed?

bychkovdmitry: Maybe I don't get something about state machines, but the same event is handled above:
```
onEvent<ClientStateEvent.ConnectedEvent> { state, event -> 
      ClientState.User.Authorized.Connected(event.connectionId, event.user, state.token) }
```

JcMinarro: By some reason it seems taht was duplicated

JcMinarro (commented):
Shouldn't we do the same for `state<ClientState.Anonymous.Pending.WithToken>`? :thinking:

bychkovdmitry:
Removed duplication and handled `ClientState.Anonymous.Authorized.Disconnected` event while in `ClientState.Anonymous.Pending.WithToken` state.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientStateService.kt
ogkuzmin: Yeah, it was dublicated. Anyway there is a test for such case. It helps us not to break it
(`Given user pending with token state When connected Should move to user connected state`)

ogkuzmin (approved):
Good job! 👍

ogkuzmin (commented):
I think JC is right. It is also good to cover the same case with anonymous user

JcMinarro (approved):

PR #942, opened 2020-12-02
Update GitHub Actions workflow for PR checks

zsmb13:
### Description

Reconfigures our previous PR check jobs into two jobs (lint and test), which are per task, instead of per module. This simplifies our config, as we no longer need to update it when new modules are added.

Workflow execution times should remain roughly the same as before, around 7 minutes.

Removes coverage jobs from CI, as we don't really use them, and they take longer than the others.

### Checklist

- [x] Reviewers added


.github/workflows/pr-checks.yml
```diff
@@ -38,59 +26,11 @@ jobs:
...
       - name: Unit tests
-        run: ./gradlew :stream-chat-android-ui-components:testDebug :stream-chat-android-ui-components-sample:testDebug --stacktrace
+        run: ./gradlew testDebug --stacktrace
       - name: Upload testDebugUnitTest results
         uses: actions/upload-artifact@v2
         if: failure()
         with:
           name: testDebugUnitTest
-          path: stream-chat-android-ui-components/build/reports/tests/testDebugUnitTest
+          path: /**/build/reports/tests/testDebugUnitTest
```
JcMinarro, line 36: It will break tests report, right?
I mean, some files will be overridden

JcMinarro: Or it will keep folder structure?

zsmb13: I think that should be okay, [the docs for the Action](https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions) say this (see also the example below this in the docs):

> If a wildcard pattern is used, the path hierarchy will be preserved after the first wildcard pattern.

But I'll break some tests and verify that it works correctly.

zsmb13: This is the structure in the zip produced, I think we're good: 

image

zsmb13: That artifact is here https://github.com/GetStream/stream-chat-android/pull/942/checks?check_run_id=1491606030

JcMinarro (approved):

PR #945, opened 2020-12-02
[CAS-452] Handle keystroke and typing stop actions

samiuelson:
https://stream-io.atlassian.net/browse/CAS-452

### Description

1. Replace text view placeholders with `TypingIndicatorView`
2. Notify MessageInput view model about keystroke and typing stop actions detected in `MessageInputView`

### Checklist

~~- [ ] Changelog updated with client-facing changes~~
~~- [ ] Comparison screenshots added for visual changes~~
- [x] Reviewers added


adasiewiczr (dismissed):
LGTM

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/MessageInputView.kt
```diff
@@ -307,6 +310,14 @@ public class MessageInputView : ConstraintLayout {
         }
     }
 
+    private fun handleKeyStroke() {
+        if (messageText.isNotBlank()) {
```
zsmb13, line 314: Minor question: are we sure we don't want to check for `isEmpty()` here instead? I'd count hitting "space" as having started typing too 😄

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/textinput/MessageInputView.kt
```diff
@@ -82,6 +82,8 @@ public class MessageInputView : ConstraintLayout {
         configSendAlsoToChannelCheckbox()
     }
 
+    public var typeListener: TypeListener = EMPTY_TYPE_LISTENER
```
zsmb13, line 85: I think making these listeners nullable makes for a more Android-y API, like the default things like `setOnClickListener` also accept `null` to clear the listener (instead of having to create an empty one, if you want to remove a listener).

Creating a separate setter method for these is also a good idea, again, it brings it closer to Android platform APIs (even if it's less Kotlin-y). As a bonus, for `fun interfaces` (which is not the case here, but is in a lot of other places), you can use a lambda when you're passing it into the method, but you have to use a more verbose SAM constructor if you have to set a property.

`SearchInputView` is an example of doing these things.

zsmb13 (dismissed):
Looks good overall

adasiewiczr (approved):

PR #946, opened 2020-12-02
[CAS-364] last message attachments

carterhudson:
### Description

- Noticed there was no direction on handling attachments for channel list items & their previews, so asked design to provide some direction and implemented what they decided on.

- Only bold mentions of the current user

- Fixed user's last message status not updating when read

- Also updates the tests that were using `yield()` to use `advanceUntilIdle()`

Screen Shot 2020-12-02 at 4 41 31 PM
Screen Shot 2020-12-02 at 4 41 50 PM

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/KeystrokeImplTest.kt
+++ b/stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/KeystrokeImplTest.kt
@@ -12,11 +12,12 @@ internal class KeystrokeImplTest : BaseConnectedIntegrationTest() {
 
     @Test
     fun keystroke() = runBlockingTest {
-        advanceUntilIdle()
         var channelState = chatDomain.useCases.watchChannel(data.channel1.cid, 10).execute().data()
         val result = chatDomain.useCases.keystroke(data.channel1.cid).execute()
+        advanceUntilIdle()
         Truth.assertThat(result.data()).isTrue()
         val result2 = chatDomain.useCases.keystroke(data.channel1.cid).execute()
+        advanceUntilIdle()
         Truth.assertThat(result2.data()).isFalse()
     }
 }
--- a/stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/KeystrokeImplTest.kt
+++ b/stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/KeystrokeImplTest.kt
@@ -3,8 +3,7 @@ package io.getstream.chat.android.livedata.usecase
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.google.common.truth.Truth
 import io.getstream.chat.android.livedata.BaseConnectedIntegrationTest
-import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.yield
+import kotlinx.coroutines.test.runBlockingTest
 import org.junit.Test
 import org.junit.runner.RunWith
 
@@ -12,8 +11,8 @@ import org.junit.runner.RunWith
 internal class KeystrokeImplTest : BaseConnectedIntegrationTest() {
 
     @Test
-    fun keystroke() = runBlocking {
-        yield()
+    fun keystroke() = runBlockingTest {
+        advanceUntilIdle()
         var channelState = chatDomain.useCases.watchChannel(data.channel1.cid, 10).execute().data()
         val result = chatDomain.useCases.keystroke(data.channel1.cid).execute()
         Truth.assertThat(result.data()).isTrue()
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
@@ -10,7 +10,7 @@ import io.getstream.chat.android.ui.channel.list.adapter.diff.ChannelDiff
 public abstract class BaseChannelListItemViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
     public abstract fun bind(
         channel: Channel,
-        diff: ChannelDiff?,
+        diff: ChannelDiff,
         channelClickListener: ChannelListView.ChannelClickListener,
         channelLongClickListener: ChannelListView.ChannelClickListener,
         channelDeleteListener: ChannelListView.ChannelClickListener,
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
@@ -10,7 +10,7 @@ import io.getstream.chat.android.ui.channel.list.adapter.diff.ChannelDiff
 public abstract class BaseChannelListItemViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
     public abstract fun bind(
         channel: Channel,
-        diff: ChannelDiff,
+        diff: ChannelDiff?,
         channelClickListener: ChannelListView.ChannelClickListener,
         channelLongClickListener: ChannelListView.ChannelClickListener,
         channelDeleteListener: ChannelListView.ChannelClickListener,
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
@@ -10,7 +10,7 @@ import io.getstream.chat.android.ui.channel.list.adapter.diff.ChannelDiff
 public abstract class BaseChannelListItemViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
     public abstract fun bind(
         channel: Channel,
-        diff: ChannelDiff? = null,
+        diff: ChannelDiff,
         channelClickListener: ChannelListView.ChannelClickListener,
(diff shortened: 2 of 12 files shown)


ogkuzmin (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt
```diff
@@ -24,54 +27,60 @@ internal fun Channel.getDisplayName(): String = name.takeIf { it.isNotEmpty() }
...
-            channelRead.lastRead?.before(currentUserLastMessage.createdAt)?.not() ?: false
-        }
-    } ?: false
+internal fun Channel.isMessageRead(message: Message): Boolean =
+    read.filterNot { it.user.isCurrentUser() }
+        .mapNotNull { it.lastRead }
+        .any { it.time >= message.getCreationDate().time }
+
+internal fun Channel.isDirectMessaging() =
+    memberCount == 2 && members.map { it.user.id }.any { it == getCurrentUser().id }
```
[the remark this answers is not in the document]
zsmb13: If we're doing perf improvements and changing this line, I'd also fetch `getCurrentUser()` just once, outside the loop

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/KeystrokeImplTest.kt
```diff
@@ -3,17 +3,16 @@ package io.getstream.chat.android.livedata.usecase
...
 import org.junit.runner.RunWith
 
 @RunWith(AndroidJUnit4::class)
 internal class KeystrokeImplTest : BaseConnectedIntegrationTest() {
 
     @Test
-    fun keystroke() = runBlocking {
-        yield()
+    fun keystroke() = runBlockingTest {
+        advanceUntilIdle()
```
zsmb13, line 15: We should place usages of `advanceUntilIdle` to where we expect data propagation to occur (in this test, between keystrokes and assertions). It's a coincidence that placing it at the start of the function works, and we shouldn't rely on that.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
```diff
@@ -10,7 +10,7 @@ import io.getstream.chat.android.ui.channel.list.adapter.diff.ChannelDiff
 public abstract class BaseChannelListItemViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
     public abstract fun bind(
         channel: Channel,
-        diff: ChannelDiff? = null,
```
zsmb13, line 13: Nice, always happy to have fewer defaults if they're not necessary 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/User.kt
```diff
@@ -1,8 +1,14 @@
 package io.getstream.chat.android.ui.utils.extensions
 
+import android.content.Context
 import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.client.models.name
 import io.getstream.chat.android.livedata.ChatDomain
+import io.getstream.chat.android.ui.R
 
-internal fun List<User>.withoutCurrentUser() = this.filter { !it.isCurrentUser() }
+internal fun List<User>.withoutCurrentUser(): List<User> = this.filter { !it.isCurrentUser() }
```
zsmb13, line 9: Love these extensions, one nitpick: it would be good to not go through the `ChatDomain` inside the loop here on every iteration, but fetch the current user just a single time outside the loop.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolder.kt
```diff
@@ -94,8 +96,11 @@ public class ChannelListItemViewHolder(itemView: View) : BaseChannelListItemView
                 }
 
                 if (lastMessageChanged) {
-                    configureLastMessageLabel(channel)
-                    configureLastMessageTimestamp(channel)
+                    channel.getLastMessage().let { lastMessage ->
+                        configureLastMessageLabel(channel, lastMessage)
+                        configureLastMessageTimestamp(lastMessage)
```
zsmb13, line 101: This logic will also be needed for `MessagePreviewView` which is used in Search results and Mentions, but we can figure out how to extract it later, if you don't want to do that on this PR.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/String.kt
```diff
@@ -1,24 +1,34 @@
...
+        ranges.forEach {
+            setSpan(StyleSpan(typeface), it.first, it.last + 1, Spannable.SPAN_EXCLUSIVE_INCLUSIVE)
         }
     }
 
-internal fun String.getMentionRanges(mentionTags: List<String>): List<IntRange> =
-    mentionTags.flatMap { tag ->
-        Regex(tag).findAll(this).map { it.range }
-    }
+internal fun String.bold(items: List<String>? = null): SpannableString = applyTypeface(BOLD, getOccurrenceRanges(items))
```
zsmb13, line 31: Everything added here in this file is awesome 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Message.kt
```diff
@@ -51,4 +29,7 @@ internal fun Message.hasNoAttachments(): Boolean = attachments.isEmpty()
 
 internal fun Message.isEphemeral(): Boolean = type == ModelType.message_ephemeral
 
-internal fun Message.getCreatedDate(): Date? = createdAt ?: createdLocallyAt
+internal fun Message.getCreationDate(): Date =
+    createdAt ?: createdLocallyAt ?: throw IllegalStateException("Message $id : \"${text}\" missing creation metadata")
```
zsmb13, line 33: We could use this in a lot of places within our codebase, though in some places it would have to be nullable. Maybe we should add a `getCreationDateOrNull` variant.

zsmb13 (changes requested):
Great stuff, my main request is to be more precise with using `advanceUntilIdle` - great that we added usages of it though!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt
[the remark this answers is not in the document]
carterhudson: something like `internal val currentUser by lazy { ChatDomain.instance().currentUser }`? Would this cause issues if the user switches accounts? 🤔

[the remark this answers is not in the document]
carterhudson: I guess it wouldn't since it's per channel instance.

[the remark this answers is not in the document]
zsmb13: Wow, I never considered delegates being used in extension properties before. This should be illegal. Like literally, because it's super broken and nobody will have the right expectations for it, I'm shocked that the compiler allows it 😄 

It simply ignores that it's an extension, it will have a single delegate backing it. But it looks extremely confusing.

---

I'd just do this for this function:

```kotlin
internal fun Channel.isDirectMessaging(): Boolean {
    val currentUser = getCurrentUser()
    return memberCount == 2 && members.map { it.user.id }.any { it == currentUser.id }
}
```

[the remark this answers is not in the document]
zsmb13: [Example of how bad this is](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS40LjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwianNDb2RlIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJcbmltcG9ydCBrb3RsaW4ucmFuZG9tLlJhbmRvbVxuXG5jbGFzcyBVc2VyXG5cbnZhbCBVc2VyLmlkOiBMb25nIGJ5IGxhenkgeyBSYW5kb20ubmV4dExvbmcoKSB9XG5cbmZ1biBtYWluKCkge1xuICAgIHZhbCB1c2VyMSA9IFVzZXIoKVxuICAgIHByaW50bG4odXNlcjEuaWQpXG4gICAgXG4gICAgdmFsIHVzZXIyID0gVXNlcigpXG4gICAgcHJpbnRsbih1c2VyMi5pZClcbn0ifQ==)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolder.kt
carterhudson: Let's extract it later :)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/User.kt
carterhudson: Agreed!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt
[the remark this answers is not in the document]
carterhudson: Interesting - just saw this

zsmb13 (approved):
One last tiny thing I noticed, but everything else looks good to go AFAIC

ogkuzmin (approved):

PR #950, opened 2020-12-03
Add Some Code Snippet Docs

zetbaitsu:
### Description

* [Events](https://getstream.io/chat/docs/event_object/?language=kotlin) code-snippet from web documentation. 
* Adding remnants from [Client & Users](https://getstream.io/chat/docs/query_users/?language=kotlin)
* Handling `result.error` API call

### Checklist

- [x] Reviewers added


stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Events.kt
```diff
@@ -0,0 +1,90 @@
...
+import io.getstream.chat.android.client.utils.observable.Disposable
+
+class Events(val client: ChatClient, val channelController: ChannelClient) {
+
+    fun listenSpecificChannelEvents() {
+        // Subscribe for new message events
+        val disposable: Disposable = channelController
+            .subscribeFor<NewMessageEvent> { newMessageEvent ->
+                // to get the message
+                val message = newMessageEvent.message
```
zsmb13, line 24: I think it would be useful if the code here was exactly the same as what's in the docs, copy-pasted. These lines differ a bit from the current version of the docs, I'd update one of these to make them match.

This applies to a lot of these snippets, not just this specific one.

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Events.kt
```diff
@@ -0,0 +1,90 @@
...
+        // Dispose when you want to stop receiving events
+        disposable.dispose()
+    }
+
+    fun listenAllChannelEvents() {
+        val disposable: Disposable = channelController
+            .subscribe { event: ChatEvent ->
+                when (event) {
+                    is NewMessageEvent -> {
+                        // to get the message
```
zsmb13, line 36: I'd decide on either lower- or uppercase comments, and use the same style everywhere here, since it's going into the documentation. My vote is starting comments with uppercase.

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Events.kt
```diff
@@ -0,0 +1,90 @@
...
+        // sends an event typing.stop to all channel participants
+        channelController.stopTyping().enqueue()
+    }
+
+    fun notificationEvents() {
+        // an example of how listen event when a user is added to a channel
+        channelController.subscribeFor<NotificationAddedToChannelEvent> { notificationEvent ->
+            // Handle event
+        }
+    }
```
zsmb13, line 89: Took me a moment to figure out that these were on a different page than the other snippets above. What if we matched the structure of the docs with our code here, either by nesting classes or functions?

So in the case of Events
image


Something like

```kotlin
class Events {
    class EventObject {
         // snippet functions
    }
    class ListeningForEvents{
         // snippet functions
    }
}
```

or

```kotlin
class Events {
    fun eventObject() {
         // nested snippet functions
    }
    fun listeningForEvents() {
         // nested snippet functions
    }
}
```

stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/ClientAndUsers.java
```diff
@@ -151,4 +158,73 @@ public void onError(@NotNull ChatError error) {
             }
         });
     }
+
+    public void queryingUsersById() {
```
zsmb13, line 162: I was a bit confused that these querying related snippets ended up on this PR too, as it just says "Events" everywhere - but they look correct 😄

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Events.kt
adasiewiczr: I like nested classes solutions. Also, maybe it is worth it to add KDoc with a link to the documentation?

zetbaitsu: Ah good idea, I'd update the `Events` class. Will update for `ClientAndUsers` and `Messages` class in next PR too.

stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/ClientAndUsers.java
zetbaitsu: Sorry, I have updated PR's title and description 🙈

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Events.kt
zetbaitsu: Need some adjustment, can not exactly the same as what's in the docs cause some snippet code in the web is not updated/working like [this one ](https://getstream.io/chat/docs/event_typing/?language=kotlin) `channel.keystroke()`, I think the right code is `channel.keystroke().enqueue()`.

And @JcMinarro also suggests adding an example code that can be done inside a callback, instead of just an empty callback like in the web now.

zsmb13: Yup, that's a good example of what we should update on the web. I'm mostly asking _when_ we'll update those. Perhaps it would be easier to do it when we add it to the repo, instead of all at once later.

stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/ClientAndUsers.java
zsmb13: No worries, thanks 😄

zsmb13 (approved):

PR #972, opened 2020-12-07
Channels Code Snippet

zetbaitsu:
### Description

* [Channels](https://getstream.io/chat/docs/initialize_channel/?language=kotlin) `Kotlin` code snippet from web documentation. `Java` code snippet will be in the next PR https://github.com/GetStream/stream-chat-android/pull/982.

### Checklist

- [x] Reviewers added


stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Channels.kt
```diff
@@ -0,0 +1,570 @@
...
+                    Log.e(TAG, String.format("There was an error %s", it.error(), it.error().cause))
+                }
+            }
+        }
+    }
+
+    /**
+     * @see <a href="https://getstream.io/chat/docs/query_members/?language=kotlin">Query Members</a>
+     */
+    inner class QueryMembers {
```
zsmb13, line 478: If everything other than the filters are the same in the snippets here, I think it would be good to skip the copy-pasted `queryMembers` and `enqueue` parts, and just create each filter, e.g.

```kotlin
// Query members by user.name
val filterByName = Filters.eq("name", "tommaso")

// Query member by id
val filterById = Filters.eq("id", "tommaso")
...
```

Maybe we can add the full call just once at the beginning or end, to show off what that looks like, but other than that it looks like only the filters matter.

zsmb13 (approved):
Merged `develop` and pushed some tiny changes, didn't want to go through another round of PR reviews for these - nice work!

PR #976, opened 2020-12-07
Fix crash on AddChannel screen

adasiewiczr:
### Checklist

- [x] Reviewers added


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelViewController.kt
```diff
@@ -82,7 +82,7 @@ class AddChannelViewController(
 
     private fun showSectionedUsers(userInfoList: List<UserInfo>, usersSubmittedCallback: () -> Unit) {
         val sectionedUsers = userInfoList
-            .groupBy { it.user.name.first().toUpperCase() }
+            .groupBy { (it.user.name.firstOrNull() ?: Char.MIN_VALUE).toUpperCase() }
```
zsmb13, line 85: Apparently `MIN_VALUE` is `'\u0000'`, the unicode NULL character. How will this appear on the UI as a section when we run into this case?

Should we perhaps go about this the other way around, and group the usernames by letters and numbers first, and put the rest that doesn't start with an alphanumeric character into some kind of "Others" section?

adasiewiczr: Agree, "others" section is better than displaying those users at the beginning of the list. Btw, `\u0000` is displayed as an empty character.

zsmb13 (approved):

PR #977, opened 2020-12-07
Fix channel view holder

JcMinarro:
### Description
Our previous implementation of ChannelViewHolder used an internal class as a parameter of the `bind()` method, and it doesn't allow an override of this class. It has been opened now.

On the other hand, ChannelItemPayloadDiff behavior has been updated to handle multiple updates properly

### Checklist

- [x] Reviewers added


stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelListItemAdapter.java
```diff
@@ -87,29 +87,27 @@ public int getItemViewType(int position) {
...
-        ChannelItemPayloadDiff diff;
+        ChannelItemPayloadDiff diff = EMPTY_CHANNEL_ITEM_PAYLOAD_DIFF;
         if (payloads.isEmpty()) {
-            diff = noDiff;
+            diff = FULL_CHANNEL_ITEM_PAYLOAD_DIFF;
         } else {
-            diff = (ChannelItemPayloadDiff) payloads.get(0);
+            for (int i = 0; i < payloads.size(); i++) {
+                diff = diff.plus((ChannelItemPayloadDiff) payloads.get(i));
+            }
```
zsmb13, line 108: This would be so much nicer in Kotlin... Anyhow, well done, seems like a good change to get more precise with the diff.

JcMinarro: Yep, it is a `fold` call in Kotlin, but refactoring all class to Kotlin was out of the scope of this PR

zsmb13 (approved):

PR #979, opened 2020-12-07
Refactor attrs of MessageListView

samiuelson:
This PR:
1. Renames a few attrs set names that didn't have "stream" prefixes. 
2. Introduces 2 new attrs for `MessageListView`: `streamEditMessageActionEnabled` and `streamDeleteMessageActionEnabled`

### Checklist`

- [x] Reviewers added


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Use custom `StreamFileProvider` instead of androidx `FileProvider` to avoid conflicts
 - Add `ChatClient::setGuestUser` method to login as a guest user
 - Make `MessageListItemViewHolder` public and open, to allow customization by overriding the `bind` method
+- Add new attrs to `MessageListView`: `streamDeleteMessageActionEnabled`, `streamEditMessageActionEnabled`
```
JcMinarro, line 7: Wrong file.
We are using `UNRELEASED_CHANGELOG.md` file now

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/dialog/MessageMoreActionDialog.kt
```diff
@@ -63,6 +63,8 @@ public class MessageMoreActionDialog(
                 dismiss()
             }
             binding.flagMessageButton.isVisible = style.flagMessageActionEnabled
+            binding.editMessageButton.isVisible = style.editMessageActionEnabled
+            binding.deleteMessageButton.isVisible = style.deleteMessageActionEnabled
```
zsmb13, line 67: Why are we adding this to the branch where the message is not by the current user?

stream-chat-android/src/main/res/values/attrs.xml
```diff
@@ -178,7 +178,7 @@
         <attr name="streamSendAlsoToChannel" format="color|reference" />
     </declare-styleable>
     <!--MessageListView-->
-    <declare-styleable name="MessageListView">
+    <declare-styleable name="StreamMessageListView">
```
zsmb13, line 181: Looks like the layout editor relies on the name of the styleable matching the name of the View to show suggestions. I don't get any autocompletion results when I try to add attributes (for example, the newly added ones) to the `MessageListView` in `fragment_channel.xml` after this change.

I think we should remove prefixes from all of our styleables, as our Views don't have them, and then suppress the warning about their names, if possible.

zsmb13 (changes requested):

stream-chat-android/src/main/res/values/attrs.xml
samiuelson: It's reverted now. Let's clean it up in another PR.

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/dialog/MessageMoreActionDialog.kt
samiuelson: Fixed

zsmb13 (approved):

PR #990, opened 2020-12-08
[CAS-462] channel list pagination

carterhudson:
https://stream-io.atlassian.net/browse/CAS-462

### Description
Adds a loading progress bar to indicate channel list items are being fetched.

Waiting on answer about test data & for https://github.com/GetStream/stream-chat-android/pull/946 to be merged before marking ready for review.

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
```diff
@@ -1,19 +1,28 @@
...
 import io.getstream.chat.android.ui.utils.extensions.cast
 import io.getstream.chat.android.ui.utils.extensions.firstOrDefault
 
 public class ChannelListItemAdapter : BaseChannelListItemAdapter() {
 
-    public var viewHolderFactory: BaseChannelViewHolderFactory<BaseChannelListItemViewHolder> =
-        ChannelViewHolderFactory()
+    public var viewHolderFactory: ChannelListItemViewHolderFactory = ChannelListItemViewHolderFactory()
+
+    public var listenerContainer: ChannelListItemListenerContainer = ChannelListItemListenerContainer()
```
zsmb13, line 16: The idea behind the original `ListenerContainer` is that the container itself will never change, only the listeners inside it, therefore it's safe for the ViewHolders to hold onto the container reference, and then fetch the current, up-to-date listener values from it when they need it.

This should probably be a `private val`, so that the container is never swapped for a new instance, because then already created ViewHolders would still reference the old instance.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
```diff
@@ -25,45 +34,59 @@ public class ChannelListItemAdapter : BaseChannelListItemAdapter() {
...
+    }
+
+    override fun getItemCount(): Int {
+        return super.getItemCount().let { realCount ->
+            when {
+                // if the list isn't empty, and we haven't reached the end, always offset +1 for the loading more view
+                !endReached && realCount > 0 -> realCount + 1
+                else -> realCount
+            }
+        }
```
zsmb13, line 59: In my experience, this kind of tricking of RecyclerView will eventually go wrong somewhere when a flag isn't updated or the offset isn't considered in some function. Everywhere else in our SDK we use a real item for the Loading item, that we add to the list that the adapter handles, so that it's rendered through the regular RecyclerView mechanisms.

In this case, we could have something like

```kotlin
sealed class ChannelListItem {
    data class ChannelItem(val channel: Channel) : ChannelListItem()
    object Loading : ChannelListItem()
}
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/BaseChannelListItemViewHolder.kt
```diff
@@ -8,13 +8,19 @@ import io.getstream.chat.android.ui.channel.list.ChannelListViewStyle
...
+
+    protected open val channelLongClickListener: ChannelListView.ChannelClickListener =
+        ChannelListView.ChannelClickListener.DEFAULT
+
+    protected open val channelDeleteListener: ChannelListView.ChannelClickListener =
+        ChannelListView.ChannelClickListener.DEFAULT
+
+    protected open val userClickListener: ChannelListView.UserClickListener = ChannelListView.UserClickListener.DEFAULT
+
+    protected open val style: ChannelListViewStyle? = null
```
zsmb13, line 23: Why do we need these in the base class, instead of in `ChannelViewHolder`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemListenerContainer.kt
```diff
@@ -0,0 +1,36 @@
+package io.getstream.chat.android.ui.channel.list.adapter.viewholder
+
+import io.getstream.chat.android.ui.channel.list.ChannelListView
+
+public class ChannelListItemListenerContainer(
```
zsmb13, line 5: Perhaps just `ChannelListListenerContainer`, for a slightly simpler name? We could also make this an interface with an internal implementation, just so that nobody gets the idea of creating their own instances of it.

Alternatively, if we're not exposing it anywhere, we could make it entirely internal. The original `ListenerContainer` is only public because it's provided as a property in both `MessageViewHolderFactory` and `AttachmentViewHolderFactory`, but I see that `ChannelListItemViewHolderFactory` is just getting each individual listener as separate parameters.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ListenerDelegate.kt
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Setting new listeners via the setter will update the underlying listener, and
+ * calls to the [wrapper] will then be forwarded to the latest [realListener] that
+ * was set.
+ *
+ * @param wrap A function that has to produce the wrapper listener. The listener being
+ *             wrapped can be referenced by calling the realListener() method. This
+ *             function always returns the current listener, even if it changes.
+ */
+internal class ListenerDelegate<L : Any>(
```
zsmb13, line 23: This won't be needed after https://github.com/GetStream/stream-chat-android/pull/1003

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/ViewGroup.kt
```diff
@@ -0,0 +1,7 @@
+package io.getstream.chat.android.ui.utils.extensions
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+
+internal val ViewGroup.inflater: LayoutInflater
```
zsmb13, line 6: This won't be needed after https://github.com/GetStream/stream-chat-android/pull/1003

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
carterhudson: Oversight on my part; I'll change it.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
carterhudson: We could, but is polluting the data set better? I've worked with people on both sides of the fence over the years. Could this even be a decorator? I don't have a strong preference, though. Will rework it if this is blocking.

zsmb13: If the implementation works, let's not block things with this now, but we should create a ticket to follow up on it. Fragmentation kills a codebase in the long term.

JcMinarro (approved):

PR #932, opened 2020-12-01
[CAS-428] Message options2

leandroBorgesFerreira:
[CAS-428](https://stream-io.atlassian.net/browse/CAS-428)

### Description

First iteration of messages options. This PR creates this view:

Dec-01-2020 23-33-00

The other view at yet to be created. They need to be added to container for the views. 

### Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] PR targets the `develop` branch
- [ ] Changelog updated with client-facing changes. **Not needed.** 
- [ ] New code is covered by unit tests. **Not needed.** 
- [x] Comparison screenshots added for visual changes
- [x] Reviewers added

PR #991, opened 2020-12-08
[CAS-465] Message options list

leandroBorgesFerreira:
[CAS-465](https://stream-io.atlassian.net/browse/CAS-465)

### Description

This PR creates the options view in the message options. It has the same approach for the Reactions screen. 

- It uses the Stylables from the stream-android-ui because we still don't have the new module to put the styles and share then. It can be done after the `stream-chat-android-ui-common` has the basic structure. 

- The ViewHolder for `MessageListItemViewType.ATTACHMENTS` must inflate a new layout, otherwise the app crashes when trying to add the viewholder for attachments with texts to dialog. 

- The MessageItem in message list is a mock, just like the reactions implementation. Both problems can be solved after the signature of the listeners are changed. 

### Checklist

- [ ] New code is covered by unit tests. **No need.** 
- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -152,14 +154,14 @@ public class MessageListView : ConstraintLayout, IMessageListView {
         MessageLongClickListenerView { _, view ->
             optionsView = view
 
-            binding.blurLayer.isVisible = true
+            binding.shadowLayer.isVisible = true
 
             if (view.parent != null && view.parent is ViewGroup) {
```
bychkovdmitry, line 159: Nice idea to take messages straight from the message list. But can you explain when this view is added again to the RecyclerView?

stream-chat-android-ui-components/src/main/res/drawable/stream_ic_mute.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+  <path
+      android:pathData="M5.889,16H2C1.7348,16 1.4804,15.8946 1.2929,15.7071C1.1054,15.5196 1,15.2652 1,15V9C1,8.7348 1.1054,8.4804 1.2929,8.2929C1.4804,8.1054 1.7348,8 2,8H5.889L11.183,3.668C11.2563,3.608 11.3451,3.57 11.4391,3.5584C11.5331,3.5469 11.6284,3.5623 11.714,3.6028C11.7996,3.6434 11.872,3.7074 11.9226,3.7875C11.9732,3.8675 12.0001,3.9603 12,4.055V19.945C12.0001,20.0397 11.9732,20.1325 11.9226,20.2126C11.872,20.2926 11.7996,20.3567 11.714,20.3972C11.6284,20.4377 11.5331,20.4531 11.4391,20.4416C11.3451,20.4301 11.2563,20.3921 11.183,20.332L5.89,16H5.889ZM10,7.22L6.603,10H3V14H6.603L10,16.78V7.22ZM23.95,15.536L20.414,12L23.95,8.464L22.536,7.05L19,10.586L15.464,7.05L14.05,8.464L17.586,12L14.05,15.536L15.464,16.95L19,13.414L22.536,16.95L23.95,15.536Z"
+      android:fillColor="#808080"
```
bychkovdmitry, line 8: As Marton suggested we set fill color to black for consistency and apply tint afterwards

stream-chat-android-ui-components/src/main/res/layout/stream_message_list_view.xml
```diff
@@ -101,32 +101,253 @@
...
+
+                <androidx.constraintlayout.widget.ConstraintLayout
+                    android:layout_width="match_parent"
+                    android:layout_height="match_parent"
+                    android:minWidth="200dp"
+                    android:paddingTop="12dp"
+                    android:paddingBottom="12dp"
+                    >
+
+                    <ImageView
```
bychkovdmitry, line 158: A bit of duplication here .Can be optimized later with LinearLayout with dividers and TextViews with `drawableLeft`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -152,14 +154,14 @@ public class MessageListView : ConstraintLayout, IMessageListView {
...
                 (view.parent as ViewGroup).removeView(view)
```
bychkovdmitry, line 160: Would be cleaner to refactor later and recreate the message view holder with viewholder factory. Would probably resolve the issue with message displayed without avatar (in the message list only last message in a row is displayed with avatar) and the issue with the list adjusting its scroll position.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -152,14 +154,14 @@ public class MessageListView : ConstraintLayout, IMessageListView {
...
             }
 
-            binding.messageOptionsContainer.addView(optionsView)
-            binding.messageOptionsScroll.isVisible = true
+            binding.optionsMessageContainer.addView(optionsView)
+            binding.optionRootContainer.isVisible = true
```
bychkovdmitry, line 164: The animations produced by `animateLayoutChanges` look really solid 👍

bychkovdmitry:
Looks good overall, but IMO message needs to be recreated from view holder factory, which can't be done at this point in time.

bychkovdmitry (commented):
Looks good overall, but IMO message needs to be recreated from view holder factory, which can't be done at this point in time.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
bychkovdmitry: Checked the design, looks like avatar should not be displayed at all

bychkovdmitry: My bad, checked the design, looks like a message should be displayed just like in the message list (either with avatar or without it)

stream-chat-android/src/main/res/values/attrs.xml
```diff
@@ -182,6 +182,20 @@
     </declare-styleable>
     <!--MessageListView-->
     <declare-styleable name="MessageListView">
+        <!--MessageOptions-->
```
zsmb13, line 185: Shouldn't these go in the new module?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
zsmb13: I'm not entirely sure about this solution. It takes the entire holder as-is, with the timestamp, avatar, replies, seen indicators included (the replies arrow is nearly invisible, grey on grey here)

image

The design only has the message bubble and the avatar:

image

I think we should use a separate View here instead of reusing the old one. It would let us solve things like truncating overly long messages too. I see that we are going with a `ScrollView` for that, but that feels like an odd solution, I'd rather do maxlines and ellipsize in that case (the design doesn't seem to address it directly).

bychkovdmitry:
BTW, all resources are prefixed with `stream_ui` now

bychkovdmitry (changes requested):
Please hide the `ScrollView`, as showing it all the time destroys the scrolling experience in the message list. The rest can be improved later.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
leandroBorgesFerreira: Thanks! 😃

stream-chat-android/src/main/res/values/attrs.xml
leandroBorgesFerreira: 🤔  Makes sense. I now see that we don't an stylable for the new MessageListView. I'll create a PR for that.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
leandroBorgesFerreira: List message is not added anymore, but you can see that the recyclerview doesn't lose it. Perhaps the recycler view recreate its views after this one is removed. 

About the recreation using the same view or recreating it using the ViewHolder, this review seams more related to the PR #932 then this one. This one just add the message options list in the bottom of the message in the new screen. 

To avoid using this PR to solve more than it is intended, what about fix this matter on a next PR?

leandroBorgesFerreira:
Let's move this back to draft as I refactor the message list options to use a dialog instead of the views.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionsOverlayDialogFragment.kt
```diff
@@ -142,14 +143,16 @@ internal class ReactionsOverlayDialogFragment : DialogFragment() {
         )
     }
 
+    private class MessageItemWrapper(val messageListItem: MessageListItem.MessageItem) : Serializable
```
bychkovdmitry, line 146: I made `MessageItem` serializable now to use the wrapper

leandroBorgesFerreira: Done

bychkovdmitry (approved):

stream-chat-android/src/main/res/values/attrs.xml
zsmb13: FYI, gonna move these on https://github.com/GetStream/stream-chat-android/pull/1022/

PR #994, opened 2020-12-09
Update the API of the Call class

zsmb13:
### Description

Using the `Call` class from Java is inconvenient, as evidenced by our documentation snippets. Using a Kotlin function type as the parameter of `enqueue` (the preferred way of running a `Call` from Java) means that we're forcing our clients to return `kotlin.Unit` explicitly in their implementations, and/or to implement an ugly `Function1<Result<T>, Unit>()` type.

This PR replaces the function type with a `fun interface`, which lets Java clients use a more natural API. The old `enqueue` signature is kept around to keep source compatibility for Java clients, but it's hidden from Kotlin clients. 

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
+++ b/stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
@@ -27,7 +27,31 @@ public interface Call<T : Any> {
      * Executes the call asynchronously, on a background thread. Safe to call from the main
      * thread. The [callback] will always be invoked on the main thread.
      */
-    public fun enqueue(callback: (Result<T>) -> Unit = {})
+    @Suppress("DeprecatedCallableAddReplaceWith", "NEWER_VERSION_IN_SINCE_KOTLIN")
+    @Deprecated(
+        level = DeprecationLevel.ERROR,
+        message = "Use the enqueue method with a Callback<T> parameter instead",
+    )
+    // Prevent usages from Kotlin, force call resolution to select the new enqueue function
+    @SinceKotlin("99999.9")
+    public fun enqueue(callback: (Result<T>) -> Unit = {}) {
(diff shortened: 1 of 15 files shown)


bychkovdmitry:
Great, when I reviewed Java code snippets few hours ago I had the same idea

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
```diff
@@ -27,14 +27,42 @@ public interface Call<T : Any> {
...
      * thread. The [callback] will always be invoked on the main thread.
      */
-    public fun enqueue(callback: (Result<T>) -> Unit = {})
+    @Suppress("DeprecatedCallableAddReplaceWith", "NEWER_VERSION_IN_SINCE_KOTLIN")
+    @Deprecated(
+        level = DeprecationLevel.ERROR,
+        message = "Use the enqueue method with a Callback<T> parameter instead",
+    )
+    // Prevent usages from Kotlin, force call resolution to select the new enqueue function
+    @SinceKotlin("99999.9")
```
bychkovdmitry, line 36: Really nice

bychkovdmitry (approved):

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
zsmb13: That's an overstatement, but it's a hack that works, and I don't know of another one 😄

bychkovdmitry: I mean, super call how it works. Completely hides the method from kotlin users.

adasiewiczr (approved):

adasiewiczr (approved):

PR #998, opened 2020-12-09
[CAS-472] message date custom formatting

carterhudson:
https://stream-io.atlassian.net/browse/CAS-472

### Description

- Allows clients to specify a custom `DateFormatter` implementation. 
- JC suggested we use the `DateFormatter` implemented in the UI components module for this, so I copied it over and tweaked it a bit.
- I kept the same formatting behavior we had previously, just used `DateFormatter` to wrap it. 
- The original request was to be able to toggle between 24 vs 12 hr formats. This delegates that power to the user via system settings (`DefaultDateFormatter`), or allows clients to override to specify their own behavior: 
```kotlin
        binding.messageListView.setMessageDateFormatter(object : DateFormatter {
            val dateFormatter = DateTimeFormatter.ofPattern("hh:mm")
            override fun format(localDateTime: LocalDateTime?): String = dateFormatter.format(localDateTime)
        })
```
- Originally considered a styleable attribute on `MessageListView` that allows clients to specify a format pattern (something like `app:streamMessageDateFormat="h:mm a"`). Open to other ideas.

### Checklist

- [x] Reviewers added


stream-chat-android/build.gradle
```diff
@@ -78,6 +78,8 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
 dependencies {
     api project(":stream-chat-android-offline")
 
+    implementation Dependencies.androidxStartup
+    api Dependencies.threeTenAbp
```
carterhudson, line 82: Since we use the threeten backport internally, I had to include it as an API dependency since `DateFormatter` accepts a `org.threeten.bp.LocalDateTime`. Do we foresee any issues doing this? Should we abstract this somehow? 🤔

JcMinarro:
> https://stream-io.atlassian.net/browse/CAS-472
> 
> ### Description
> * Allows clients to specify a custom `DateFormatter` implementation.
> * JC suggested we use the `DateFormatter` implemented in the UI components module for this, so I copied it over and tweaked it a bit.
> * I kept the same formatting behavior we had previously, just used `DateFormatter` to wrap it.
> * The original request was to be able to toggle between 24 vs 12 hr formats. This delegates that power to the user via system settings (`DefaultDateFormatter`), or allows clients to override to specify their own behavior:
> 
> ```kotlin
>         binding.messageListView.setMessageDateFormatter(object : DateFormatter {
>             val dateFormatter = DateTimeFormatter.ofPattern("hh:mm")
>             override fun format(localDateTime: LocalDateTime?): String = dateFormatter.format(localDateTime)
>         })
> ```
> 
> * Originally considered a styleable attribute on `MessageListView` that allows clients to specify a format pattern (something like `app:streamMessageDateFormat="h:mm a"`). Open to other ideas.
> 
> ### Checklist
> * [x]  I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
> * [x]  PR targets the `develop` branch
> * [x]  Changelog updated with client-facing changes
> * [ ]  New code is covered by unit tests
> * [ ]  Comparison screenshots added for visual changes
> * [x]  Reviewers added

The idea was to have only one `DateFormatter` across all SDKs, so it shouldn't be copied, but moved to `stream-chat-android` module.
Let me do it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/startup/ThreeTenInitializer.kt
```diff
@@ -0,0 +1,14 @@
+package io.getstream.chat.android.ui.startup
+
+import android.content.Context
+import androidx.startup.Initializer
+import com.jakewharton.threetenabp.AndroidThreeTen
+
+@Suppress("unused")
+internal class ThreeTenInitializer : Initializer<Unit> {
```
zsmb13, line 8: Wouldn't it be enough to have one of these in the old UI module?

UNRELEASED_CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 - Add new attrs to `MessageListView`: `streamDeleteMessageActionEnabled`, `streamEditMessageActionEnabled`
 - Improve Channel List Diff
 - Add new attrs to `MessageInputView`: `streamInputScrollbarEnabled`, `streamInputScrollbarFadingEnabled`
+- Add API for setting custom message date formatter in MessageListView via `setMessageDateFormatter(DateFormatter)`
+    - 24 vs 12 hr controlled by user's System settings.
```
zsmb13, line 10: For custom formatters, this system settings part doesn't apply, they choose how to format it from just a `LocalDateTime`

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/DateFormatter.kt
```diff
@@ -1,38 +1,21 @@
...
-        val dateCalendar = Calendar.getInstance().apply {
-            time = date
-        }
-        return today[Calendar.DAY_OF_YEAR] == dateCalendar[Calendar.DAY_OF_YEAR] &&
-            today[Calendar.YEAR] == dateCalendar[Calendar.YEAR]
-    }
+/**
+ * Extension to be able to format objects of the deprecated [Date] type.
+ */
+public fun DateFormatter.formatDate(date: Date?): String {
```
zsmb13, line 19: Does this have to be `public`, couldn't it be `@InternalStreamChatApi`?

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelListItemViewHolder.java
```diff
@@ -177,7 +179,8 @@ protected void configLastMessage(Channel channel) {
     }
 
     protected void configLastMessageDate(Channel channel) {
-        tv_date.setText(DateFormatter.formatAsTimeOrDate(channel.getLastMessageAt()));
+        String date = DateFormatterKt.formatDate(defaultDateFormatter, channel.getLastMessageAt());
```
zsmb13, line 182: If we make this `formatDate` public, we should rename the generated containing class for nicer Java API. Depends on the previous comment about it being public / internal though.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/startup/ThreeTenInitializer.kt
JcMinarro: Ups, I forgot to remove this one

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/DateFormatter.kt
JcMinarro: Why don't we allow to use it with `Date`? :thinking: 
`Date` is the one that we use in our models

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelListItemViewHolder.java
JcMinarro: Let me add it

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/DateFormatter.kt
zsmb13: We should at some point get rid of this one, because we shouldn't be using `Date` for our date handling. I created the core API of the new formatter with ThreeTenABP so that it's not immediately deprecated too - gotta start somewhere.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/startup/ThreeTenInitializer.kt
carterhudson: Are we eventually migrating everything over to `ui-components` at some point? Does this become something we would eventually migrate back?

carterhudson:
> > https://stream-io.atlassian.net/browse/CAS-472
> > ### Description
> > 
> > * Allows clients to specify a custom `DateFormatter` implementation.
> > * JC suggested we use the `DateFormatter` implemented in the UI components module for this, so I copied it over and tweaked it a bit.
> > * I kept the same formatting behavior we had previously, just used `DateFormatter` to wrap it.
> > * The original request was to be able to toggle between 24 vs 12 hr formats. This delegates that power to the user via system settings (`DefaultDateFormatter`), or allows clients to override to specify their own behavior:
> > 
> > ```kotlin
> >         binding.messageListView.setMessageDateFormatter(object : DateFormatter {
> >             val dateFormatter = DateTimeFormatter.ofPattern("hh:mm")
> >             override fun format(localDateTime: LocalDateTime?): String = dateFormatter.format(localDateTime)
> >         })
> > ```
> > 
> > 
> > 
> > * Originally considered a styleable attribute on `MessageListView` that allows clients to specify a format pattern (something like `app:streamMessageDateFormat="h:mm a"`). Open to other ideas.
> > 
> > ### Checklist
> > 
> > * [x]   I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
> > * [x]   PR targets the `develop` branch
> > * [x]   Changelog updated with client-facing changes
> > * [ ]   New code is covered by unit tests
> > * [ ]   Comparison screenshots added for visual changes
> > * [x]   Reviewers added
> 
> The idea was to have only one `DateFormatter` across all SDKs, so it shouldn't be copied, but moved to `stream-chat-android` module.
> Let me do it

Do we want the new date formatting criteria from the redesign present in the old sdk? Are we planning on getting rid of the old module once the redesign is complete? Is this something that we'd eventually have to migrate back to the `ui-components` module?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/startup/ThreeTenInitializer.kt
zsmb13: We can push this date formatting code down into the upcoming `stream-chat-android-ui-common`, it can live there. I don't think we'll move it up to `ui-components` at any point.

zsmb13 (approved):
Good to go as far as I'm concerned

JcMinarro (approved):

PR #1000, opened 2020-12-10
Reactions screen

bychkovdmitry:
### Description

An implementation of "Reactions" overlay. Implemented using full screen `DialogFragment` to avoid corner cases: no need to block the message input, back button behavior is not ambiguous and returns the user to the message list. Besides, in full screen mode we get more screen real estate to accommodate long messages. The message on the overlay is recreated with `MessageListItemViewHolderFactory`, therefore there is no need to detach existing view holder from the message list and reattach it to the overlay. Fancy animations can be done later.

What's blocking this PR and what needs to be done:
- Message `ViewHolder` always occupies the same width, regardless of the message length. Should be fixed in the message list.
- Reaction types that we use currently differ from reaction types used in the new UI components, so they are hardcoded for now
- There should be a consistent way to get message bounds from message view holders to position the `EditReactionsView` according to the designs. I think this can be postponed until all message view holders are implemented and we have a complete picture.
- Currently, `ReactionViewClickListener` accepts `Message` as an argument. Changing the argument type to `MessageListItem` and corresponding changes require significant efforts and will not be done in the context of this PR

### Checklist

- [X] Reviewers added

| Theirs message | Mine message |
| --- | --- |
| photo_2020-12-10 10 55 53 |photo_2020-12-10 10 55 49 |


samiuelson (dismissed):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionsOverlayDialogFragment.kt
```diff
@@ -0,0 +1,161 @@
...
+        binding.messageContainer.addView(
+            messageView,
+            FrameLayout.LayoutParams(
+                FrameLayout.LayoutParams.MATCH_PARENT,
+                FrameLayout.LayoutParams.WRAP_CONTENT
+            )
+        )
+    }
+
+    private class MessageItemWrapper(val messageListItem: MessageListItem.MessageItem) : Serializable
```
JcMinarro, line 148: Why do we need this wrapper class?
Can't we use `MessageListItem.MessageItem` instead?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -171,17 +173,21 @@ public class MessageListView : ConstraintLayout, IMessageListView {
...
-                onStartThreadHandler,
-                onMessageFlagHandler
-            ).show()
+            context.getFragmentManager()?.let {
+                // TODO: pass a real MessageItem instead of mock
+                val mockMessageItem = MessageItem(
+                    message.apply {
+                        latestReactions.forEach { it.type = ReactionType.LOVE.type }
+                        ownReactions.forEach { it.type = ReactionType.LOVE.type }
+                    },
```
JcMinarro, line 182: Would be good to update the rendered test bo be sure we don't release it to production.
We could use `copy()` method instead, to not update the one that comes from the listener.
something like:
```suggestion
                    message.copy(
                        latestReactions = message.latestReactions.forEach { it.type = ReactionType.LOVE.type },
                        ownReactions, = message.ownReactions.forEach { it.type = ReactionType.LOVE.type },
                        text = "MOCKED TEXT: ${message.text}"
                    },
```

JcMinarro (commented):
Look Good!!!! Some comments on the code

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionsOverlayDialogFragment.kt
bychkovdmitry: Yes, I will make `MessageItem` serializable then

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
bychkovdmitry: It is just is shallow copy, right?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionsOverlayDialogFragment.kt
zsmb13: Could we use `Parcelable` instead? Much more reliable in my experience. Also, `MessageItem` contains things like the network `Message` model, I'd be terrified of serializing it. Maybe we could introduce a new, simple model here that has just the necessary data, and pass that through to the dialog instead?

JcMinarro: Because the Dialog internally is using our ViewModelFacory, we need to use `MessageItem` with its real content.
About `Parcelable`? Why? It doesn't contribute anything. Serializable is easier for this case.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
JcMinarro: Yep, creating a copy to not update the previous message object

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/reactions/ReactionsOverlayDialogFragment.kt
zsmb13: Serializable is a reflection based Java API, which means it's a lot slower (doesn't matter in this case), and will not necessarily play nicely with Kotlin. I wouldn't be surprised if something unexpected happened while serializing and deserializing values here.

bychkovdmitry: Did it by making `MessageItem` serializable

zsmb13 (approved):

PR #1002, opened 2020-12-10
[CAS-468] Add possibility to add users to group channel

adasiewiczr:
This PR also adds missing channel id when creating group channel.
Creating channel without id results in a distinct channel being created (based on members) and that channel type doesn't allow to change members.

Screenshot 2020-12-10 at 17 38 50

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/group/users/GroupChatInfoAddUsersViewModel.kt
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/group/users/GroupChatInfoAddUsersViewModel.kt
@@ -28,7 +28,8 @@ class GroupChatInfoAddUsersViewModel(
     val userAddedState: LiveData<Boolean> = _userAddedState
 
     init {
-        chatDomain.useCases.getChannelController(cid).enqueue { result ->
+        viewModelScope.launch {
+            val result = chatDomain.useCases.getChannelController(cid).await()
             if (result.isSuccess) {
                 members = result.data().members
                 viewModelScope.launch {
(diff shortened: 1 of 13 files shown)


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/group/users/GroupChatInfoAddUsersViewModel.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    init {
+        chatDomain.useCases.getChannelController(cid).enqueue { result ->
+            if (result.isSuccess) {
+                members = result.data().members
+                viewModelScope.launch {
+                    fetchUsers()
+                }
+            }
+        }
```
zsmb13, line 38: Since we have coroutine here anyway, we could do this entire thing inside the coroutine, using `await` to get the controller instead of `enqueue`

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/group/users/GroupChatInfoAddUsersViewModel.kt
```diff
@@ -0,0 +1,137 @@
...
+                results = emptyList(),
+                isLoading = true,
+                canLoadMore = true,
+            )
+            fetchUsers()
+        }
+    }
+
+    private suspend fun fetchUsers() {
+        members.value?.let { currentMembers ->
```
zsmb13, line 86: Nit: `val currentMembers = members.value ?: return` would _let_ us avoid nesting the entire function body inside `let`

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/EndlessScrollListener.kt
```diff
@@ -2,9 +2,7 @@ package com.getstream.sdk.chat.view
 
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
-import io.getstream.chat.android.core.internal.InternalStreamChatApi
 
-@InternalStreamChatApi
```
zsmb13, line 7: I don't really want to provide a utility like this to our customers as public API, it's not super related to our library. They should use their own paging implementations.

Maybe we could opt into using this where we use it in the sample app? Not sure what other solution there is.

adasiewiczr: As discussed on Slack - we are going to keep it public

stream-chat-android/src/main/java/com/getstream/sdk/chat/view/MessageListView.kt
```diff
@@ -242,6 +241,9 @@ public class MessageListView : ConstraintLayout, IMessageListView {
         if (attr != null) {
             configureAttributes(attr)
         }
+        loadMoreListener = EndlessScrollListener(loadMoreThreshold) {
+            endRegionReachedHandler()
+        }
```
zsmb13, line 246: I think we could move this into `configureAttributes` to get rid of the `loadMoreThreshold` property (in both of these cases)

zsmb13 (approved):

PR #1005, opened 2020-12-11
Message preview formatting

zsmb13:
### Description

Updates the formatting of the sender and the message itself in `MessagePreviewView` to match the design.

<p float=left>

</p>

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/component_browser/messagepreview/ComponentBrowserMessagePreviewFragment.kt
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/component_browser/messagepreview/ComponentBrowserMessagePreviewFragment.kt
@@ -6,9 +6,10 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.fragment.app.Fragment
 import io.getstream.chat.android.client.models.Attachment
-import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.ChannelInfo
 import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.models.name
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import io.getstream.chat.ui.sample.databinding.FragmentComponentBrowserMessagePreviewBinding
 import io.getstream.chat.ui.sample.feature.component_browser.utils.randomUser
 import java.util.Date
@@ -32,6 +33,7 @@ class ComponentBrowserMessagePreviewFragment : Fragment() {
         _binding = null
     }
 
+    @InternalStreamChatApi
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
@@ -41,11 +43,10 @@ class ComponentBrowserMessagePreviewFragment : Fragment() {
                 user = randomUser(),
                 createdAt = Date(120, 7, 15, 14, 22),
                 text = "Hello world, how are you doing?",
-                channel = Channel(
+                channelInfo = ChannelInfo(
                     memberCount = 2,
-                ).apply {
-                    name = "Direct Message"
-                }
+                    name = "Direct Message",
+                ),
             )
         )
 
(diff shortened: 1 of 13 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt
```diff
@@ -105,9 +105,14 @@ internal fun Channel.getLastMessagePreviewText(
 
         val attachments: SpannableString? = message.attachments
             .takeIf { it.isNotEmpty() }
-            ?.map { attachment ->
+            ?.mapNotNull { attachment ->
                 attachment.title?.let { title ->
-                    "${getAttachmentPrefix(context, attachment)} $title"
```
zsmb13, line 110: `getAttachmentPrefix` has a nullable return type, inserting that into a string with a template could output the literal `"null"` string on the UI

stream-chat-android-ui-components/src/main/res/values/strings.xml
```diff
@@ -73,4 +73,8 @@
         <item quantity="one">%d Message Reaction</item>
         <item quantity="other">%d Message Reactions</item>
     </plurals>
+
+    <!-- Message Preview -->
+    <string name="stream_ui_message_file">File: %s</string>
+    <string name="stream_ui_message_sender_title_in_channel">&lt;b&gt;%1$s&lt;/b&gt; in &lt;b&gt;%2$s&lt;/b&gt;</string>
```
zsmb13, line 79: I'm so sorry 😄

bychkovdmitry: Maybe CDATA?

```
<![CDATA[
    Since this is a CDATA section
    I can use all sorts of reserved characters
    like > < " and &
    or write things like
    <foo></bar>
    but my document is still well formed!
]]>
```

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/component_browser/messagepreview/ComponentBrowserMessagePreviewFragment.kt
```diff
@@ -29,15 +33,89 @@ class ComponentBrowserMessagePreviewFragment : Fragment() {
         _binding = null
     }
 
+    @InternalStreamChatApi
```
bychkovdmitry, line 36: ???

zsmb13: Uses `ChannelInfo` in its body, this makes that compile (it propagates the opt-in requirement, but since it's a framework method, it doesn't matter). Didn't want to add any opt-in Gradle configuration to this module.

stream-chat-android-ui-components/src/main/res/values/strings.xml
zsmb13: Not sure if this would be much of an improvement, both feel and look kinda ugly and hacky 😄

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/mentions/MentionsListAdapter.kt
```diff
@@ -8,13 +8,17 @@ import androidx.recyclerview.widget.ListAdapter
...
 import io.getstream.chat.android.ui.mentions.MentionsListAdapter.MessagePreviewViewHolder
 import io.getstream.chat.android.ui.mentions.MentionsListView.MentionSelectedListener
 import io.getstream.chat.android.ui.messagepreview.MessagePreviewView
+import io.getstream.chat.android.ui.utils.extensions.asMention
+import io.getstream.chat.android.ui.utils.extensions.context
 
 public class MentionsListAdapter(
-    context: Context
+    context: Context,
+    private val chatDomain: ChatDomain,
```
bychkovdmitry, line 21: I would pass only the required data: currentUser instead of ChatDomain

bychkovdmitry (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/mentions/MentionsListAdapter.kt
zsmb13: Did this just in case the current user can somehow change while the adapter instance is alive (which is very unlikely, but still)

PR #1013, opened 2020-12-14
Fix/push notification

### LINKED ISSUE #578: Add a method to identify streamchat related notifications
As we already have a `FirebaseMessagingService` implementation, which handles a lot of different types of push notifications, we cannot directly add your `ChatFirebaseNotificationService` to the manifest. Instead, we need to be able to determine which type of push message it is, and then call the appropriate handler.

A common pattern for this on Android (e.g. intercom and others) is to provide a `isStreamChatMessage: Boolean` method, which we can use to then call `Client.onMessageReceived()`.

gz#4564

JcMinarro:
Fixes #578
### Description
Add method to know if a RemoteMessage is valid for Stream
Fix crash when a push notification arrives from another provider different than Stream

### Checklist

- [x] Reviewers added

### DIFF
--- a/UNRELEASED_CHANGELOG.md
+++ b/UNRELEASED_CHANGELOG.md
@@ -10,7 +10,7 @@
     - 24 vs 12 hr controlled by user's System settings.
 
 ## stream-chat-android-client
-- Add method to know if a RemoteMessage is valid for Stream
+- Add `ChatClient::isValidRemoteMessage` method to know if a RemoteMessage is valid for Stream
 
 ## stream-chat-android-offline
 - Add updating `channelData` after receiving `ChannelUpdatedByUserEvent`
--- a/UNRELEASED_CHANGELOG.md
+++ b/UNRELEASED_CHANGELOG.md
@@ -10,6 +10,8 @@
     - 24 vs 12 hr controlled by user's System settings.
 
 ## stream-chat-android-client
+- Add method to know if a RemoteMessage is valid for Stream
 
 ## stream-chat-android-offline
 - Add updating `channelData` after receiving `ChannelUpdatedByUserEvent`
+- Fix crash when a push notification arrives from other provider different than Stream
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -26,6 +26,7 @@ import io.getstream.chat.android.client.events.ChatEvent
 import io.getstream.chat.android.client.events.ConnectedEvent
 import io.getstream.chat.android.client.events.DisconnectedEvent
 import io.getstream.chat.android.client.events.ErrorEvent
+import io.getstream.chat.android.client.extensions.isValid
 import io.getstream.chat.android.client.helpers.QueryChannelsPostponeHelper
 import io.getstream.chat.android.client.logger.ChatLogLevel
 import io.getstream.chat.android.client.logger.ChatLogger
@@ -42,6 +43,7 @@ import io.getstream.chat.android.client.models.Reaction
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.client.notifications.ChatNotifications
 import io.getstream.chat.android.client.notifications.handler.ChatNotificationHandler
+import io.getstream.chat.android.client.notifications.handler.NotificationConfig
 import io.getstream.chat.android.client.socket.ChatSocket
 import io.getstream.chat.android.client.socket.InitConnectionListener
 import io.getstream.chat.android.client.socket.SocketListener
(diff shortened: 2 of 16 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -927,5 +935,8 @@ public class ChatClient internal constructor(
 
         public val isInitialized: Boolean
             get() = instance != null
+
+        public fun isValid(remoteMessage: RemoteMessage, defaultNotificationConfig: NotificationConfig = NotificationConfig()): Boolean =
```
adasiewiczr, line 939: NIT: I think it might be worth it to give all `isValid` methods  (except RemoteMessage extension function)  a more descriptive name, e.g. `isValidRemoteMessage`. What do you think?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/ChatFirebaseMessagingService.kt
```diff
@@ -1,19 +1,65 @@
...
+
+    private fun createSyncNotificationChannel() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            NotificationChannel(CHANNEL_ID, CHANNEL_NAME, NotificationManager.IMPORTANCE_HIGH).run {
+                getSystemService(NotificationManager::class.java).createNotificationChannel(this)
+            }
+        }
+    }
+
+    companion object {
```
adasiewiczr, line 60: This can probably be private

adasiewiczr (approved):

UNRELEASED_CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
     - 24 vs 12 hr controlled by user's System settings.
 
 ## stream-chat-android-client
+- Add method to know if a RemoteMessage is valid for Stream
```
zsmb13, line 13: Would be useful to name the specific class/method here

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/RemoteMessageExtensions.kt
```diff
@@ -0,0 +1,12 @@
+package io.getstream.chat.android.client.extensions
+
+import com.google.firebase.messaging.RemoteMessage
+import io.getstream.chat.android.client.notifications.handler.NotificationConfig
+
+public fun RemoteMessage.isValid(notificationConfig: NotificationConfig): Boolean =
```
zsmb13, line 6: Do we want to provide this as separate public API, shouldn't checks just go through `ChatClient`'s companion?

Having gone through all the code on this PR, it looks like we have... Five (?) public methods for checking if a message is valid now. Do we really need to expose all of these?

adasiewiczr (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -155,18 +155,15 @@ public interface ChatDomain {
             return this
         }
 
-        // TODO: do we even need this, or is it better to have it at the low level client?
+        @Deprecated("This method is deprecated, you should configure it into the ChatClient")
         public fun notificationConfig(notificationConfig: NotificationConfig): Builder {
-            this.notificationConfig = notificationConfig
+            ChatLogger.get("ChatDomain").logE("This method is deprecated, you should configure it into the ChatClient")
```
zsmb13, line 160: This method can still be called, but it's now a silent no-op, which will be unexpected for clients. Perhaps we want to mark this as `level = ERROR` so that it breaks the build? And even then, I'd keep it functional if possible, until we remove it.

zsmb13 (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
JcMinarro: Yep, `level = ERROR` is the best way for now. Keep old behavior could give us some problem with a bad configuration

zsmb13: We might still have people who already have compiled code that calls into this method, but I guess it's gonna be rare. Maybe throw an exception for that case, so that this never gets silently executed?

JcMinarro: :+1:

zsmb13 (approved):

samiuelson (approved):

PR #1038, opened 2020-12-16
Better channel list pagination

carterhudson:
### Description

The pagination for channel list we currently have differs from how we approach it in other places (having a list item that represents loading more). This PR puts things more in line with our other approaches and is easier to understand and maintain.

https://user-images.githubusercontent.com/2574098/102813902-e6b17400-4386-11eb-84e7-089bd83d7d57.mov

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -30,6 +31,7 @@ public class ChannelListView @JvmOverloads constructor(
     private val layoutManager: ScrollPauseLinearLayoutManager
     private val scrollListener: EndReachedScrollListener = EndReachedScrollListener()
     private val dividerDecoration: SimpleVerticalListDivider = SimpleVerticalListDivider()
+    private var loadingMore = false
```
zsmb13, line 34: This is duplicating state from the adapter, and might get out of sync if we ever forget to update it. We could check whether we are currently loading with something like `adapter.currentList.contains(LoadingMoreItem)`.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -116,12 +118,36 @@ public class ChannelListView @JvmOverloads constructor(
...
+
+            when {
+                showLoadingMore -> {
+                    loadingMore = true
+                    mutableCopy.add(ChannelListItem.LoadingMoreItem)
+                }
+
+                hideLoadingMore -> {
+                    loadingMore = false
+                    mutableCopy.remove(ChannelListItem.LoadingMoreItem)
```
zsmb13, line 141: When adding/removing this item, we could use [`filterIsInstance`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-is-instance.html) to make sure that we never end up with multiple loading items in the list.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
carterhudson: That was how I had initially implemented it, but it seemed like searching a potentially large list of items to see if we're loading might have some performance impacts.

zsmb13: I don't think performance will be an issue here, and if it is, we can always optimize it later.

- Channel lists probably won't be super long most of the time
- We don't need to do this often, only when the loading/pagination state changes
- Iterating an `ArrayList` should be very quick
- We only need to compare references, and not values, because we're looking for an `object`

carterhudson: Fair enough!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -117,12 +119,32 @@ public class ChannelListView @JvmOverloads constructor(
...
-        requireAdapter().submitList(channels)
+    public fun showLoadingMore(show: Boolean) {
+        requireAdapter().let { adapter ->
+            val currentList = adapter.currentList
+            val loadingMore = currentList.contains(ChannelListItem.LoadingMoreItem)
+            val showLoadingMore = show && !loadingMore
+            val hideLoadingMore = !show && loadingMore
+
+            val updatedList = when {
+                showLoadingMore -> currentList.toMutableList().fluentAdd(ChannelListItem.LoadingMoreItem)
```
zsmb13, line 134: This is the same as just

```suggestion
                showLoadingMore -> currentList + ChannelListItem.LoadingMoreItem
```

zsmb13 (approved):
LGTM

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
```diff
@@ -12,32 +11,55 @@ import io.getstream.chat.android.ui.channel.list.adapter.viewholder.ChannelListL
...
-
-            override fun areContentsTheSame(oldItem: Channel, newItem: Channel): Boolean =
-                !oldItem.diff(newItem).hasDifference()
-
-            override fun getChangePayload(oldItem: Channel, newItem: Channel): Any = oldItem.diff(newItem)
-        }
+        val DIFF_CALLBACK: DiffUtil.ItemCallback<ChannelListItem> =
+            object : DiffUtil.ItemCallback<ChannelListItem>() {
+                override fun areItemsTheSame(oldItem: ChannelListItem, newItem: ChannelListItem): Boolean {
+                    if (oldItem::class != newItem::class) {
```
leandroBorgesFerreira, line 27: You can use only boolean logic here:
```
return oldItem::class == newItem::class && (oldItem is ChannelListItem.ChannelItem && oldItem.channel.cid == newItem.safeCast<ChannelListItem.ChannelItem>()?.channel?.cid || oldItem !is ChannelListItem.ChannelItem)
```

leandroBorgesFerreira: It is a good idea to extract some parts of the boolean logic to a method to avoid a super big line.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
```diff
@@ -12,32 +11,55 @@ import io.getstream.chat.android.ui.channel.list.adapter.viewholder.ChannelListL
...
+                        }
+
+                        else -> true
+                    }
+                }
+
+                override fun areContentsTheSame(oldItem: ChannelListItem, newItem: ChannelListItem): Boolean {
+                    // this is only called if areItemsTheSame returns true, so they must be the same class
+                    return when (oldItem) {
+                        is ChannelListItem.ChannelItem -> {
```
leandroBorgesFerreira, line 43: The observation, you can use only boolean logic here. No need for the `when`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
carterhudson: Man, I write extensions and half the time I realize they're already available natively. This is one of the times I just... 🤦 .  Also GitHub is going to make you re-approve after I commit that suggestion.

zsmb13: Almost everything is available in the Standard Library, that's one of the best things in the language 😄 

By the way, you'll also have to remove the extension and its import separately, unfortunately these one-line suggestions are rarely applicable on their own, as convenient as they are 😞

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelsViewModelBinding.kt
```diff
@@ -24,7 +25,11 @@ public fun ChannelsViewModel.bindView(
                 }
             }
             is ChannelsViewModel.State.Result -> {
-                view.setChannels(channelState.channels)
+                channelState
+                    .channels
+                    .map { ChannelListItem.ChannelItem(it) }
```
leandroBorgesFerreira, line 30: I believe you can use function reference here. `.map(ChannelListItem::ChannelItem)`

leandroBorgesFerreira:
All my observations are just sugar syntax. Fell free to choose the implementation you prefer.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
zsmb13: I like the structure in this `ItemCallback` implementation as-is, it handles cases clearly, and it will be easy to update clearly if we ever need to add a new item type.

carterhudson: Just going for some readability & taking advantage of smart casting from `when`. I want to avoid the big line, but also avoid the manual cast of `oldItem` and evaluation of `oldItem.channel.cid == newItem.safeCast<ChannelListItem.ChannelItem>()?.channel?.cid`, which would happen if I extracted both into named variables (`return sameClass && sameCid`).

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
carterhudson: The `when` is just for implicit smart casting & readability

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelsViewModelBinding.kt
carterhudson: True 👍

leandroBorgesFerreira (approved):
🚢

PR #1039, opened 2020-12-16
[CAS-476] Feature/time decorator

JcMinarro:
### Description
Add the time where the message was sent

| Before | After |
| --- | --- |
| photo_2020-12-16_23-03-28 |photo_2020-12-16_23-03-30|

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/TimeDecorator.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun decorateOnlyMediaAttachmentsMessage(
+        viewHolder: OnlyMediaAttachmentsViewHolder,
+        data: MessageListItem.MessageItem
+    ) {
+        setupTime(viewHolder.binding.tvTime, data)
+    }
+
+    private fun setupTime(textView: TextView, data: MessageListItem.MessageItem) {
+        textView.text = dateFormatter.formatTime(data.message.getCreatedAtOrThrow())
```
ogkuzmin, line 33: Do we really want to throw an exception here? For example, component browser screen crashes because mocked messages don't contain such a field

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/BaseMessageItemViewHolder.kt
```diff
@@ -19,7 +21,8 @@ public abstract class BaseMessageItemViewHolder<T : MessageListItem>(
         GapDecorator(),
         MaxPossibleWidthDecorator(),
         AvatarDecorator(),
-        GravityDecorator()
+        GravityDecorator(),
+        TimeDecorator(DateFormatter.from(itemView.context))
```
zsmb13, line 25: Let's add a trailing comma, we gotta do it sometime for this list 😄

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/Extensions.kt
```diff
@@ -24,6 +24,16 @@ public fun ConstraintSet.constrainViewToParentBySide(view: View, side: Int) {
     connect(view.id, side, ConstraintSet.PARENT_ID, side)
 }
 
+@InternalStreamChatApi
+public fun ConstraintSet.constrainViewStartToEndOfView(startView: View, endView: View) {
+    connect(startView.id, ConstraintSet.START, endView.id, ConstraintSet.END)
+}
+
+@InternalStreamChatApi
+public fun ConstraintSet.constrainViewEndToEndOfView(startView: View, endView: View) {
```
zsmb13, line 33: The param names for this method don't seem to make complete sense

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/DateFormatter.kt
```diff
@@ -7,6 +7,7 @@ import java.util.Date
 
 public interface DateFormatter {
     public fun formatDate(localDateTime: LocalDateTime?): String
+    public fun formatTime(localDateTime: LocalDateTime?): String
```
zsmb13, line 10: This should take a `LocalTime?` as its parameter

zsmb13 (changes requested):
Looks good, let's change the param for the new formatter method and then it's good to go

stream-chat-android-ui-components/src/main/res/layout/stream_ui_item_message_media_attachments.xml
```diff
@@ -21,7 +22,7 @@
         android:padding="1dp"
         app:layout_constraintEnd_toEndOf="@+id/marginEnd"
         app:layout_constraintStart_toEndOf="@+id/avatarView"
-        app:layout_constraintTop_toBottomOf="@id/gapView"
+        app:layout_constraintTop_toTopOf="parent"
```
ogkuzmin, line 25: Why GapView is useless now? It supposes to be at the top of messages, giving additional space. Now it constrained to MediaAttachmentsGroupView and do not nothing

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/TimeDecorator.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    override fun decoratePlainTextMessage(viewHolder: MessagePlainTextViewHolder, data: MessageListItem.MessageItem) {
+        setupTime(viewHolder.binding.tvTime, data)
+    }
+
+    override fun decoratePlainTextWithMediaAttachmentsMessage(
+        viewHolder: PlainTextWithMediaAttachmentsViewHolder,
+        data: MessageListItem.MessageItem
+    ) {
+        setupTime(viewHolder.binding.tvTime, data)
```
ogkuzmin, line 22: I thought it should show time only for messages with the bottom position, not every message. The same about avatar. Or it does not suppose to be part of this PR?

ogkuzmin (commented):
I didn't get an idea of changes with gap view. TimeDecorator idea is clear but seems we need to limit showing only for bottom messages.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/TimeDecorator.kt
JcMinarro: CreatedAt sholdn't be never null (it is set by server or our local mechanism), if it is null, message has wrong data (Maybe because dev has set as null it).
Should we allow null there? :thinking:

stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/DateFormatter.kt
ogkuzmin: Judging by the implementation, both these methods format the same thing but with different output

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/LegacyDateFormatter.kt
```diff
@@ -1,10 +1,33 @@
...
+    final override fun formatTime(localTime: LocalTime?): String =
+        formatTime(localTime?.let(::toTime))
+
+    private fun toTime(localTime: LocalTime): Date {
+        return Date().apply {
+            hours = localTime.hour
+            minutes = localTime.minute
+            seconds = localTime.second
+        }
+    }
```
zsmb13, line 32: Added this only here and not to `DateConverter`, because it's not a great conversion, but it's what we need to do if we want to support `Date`

stream-chat-android-ui-common/build.gradle
```diff
@@ -106,6 +106,8 @@ dependencies {
     testRuntimeOnly Dependencies.junitJupiterEngine
     testRuntimeOnly Dependencies.junitVintageEngine
 
+    testImplementation Dependencies.threeTenBp
```
zsmb13, line 109: Strictly for tests only, for prod we need the Android backport instead

zsmb13 (approved):
Fixed up the DateFormatter API, good to go on my part

ogkuzmin (approved):
Fixed a visual part. Goot for me as well)

PR #1046, opened 2020-12-17
Cas 478/file attachments

ogkuzmin:
< Jira issue link, if applicable >

### Description

Brings support of FileAttachmentsVH to MessageListView
files_vh

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components-sample/build.gradle
```diff
@@ -117,6 +117,14 @@ android {
     }
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions {
+        freeCompilerArgs += [
+                '-Xopt-in=io.getstream.chat.android.core.internal.InternalStreamChatApi',
```
adasiewiczr, line 123: A few days ago I had a conversation about that with @zsmb13 and I'm not sure if we should do that in the sample app because someone could copy our `build.gradle` and they won't see any error while using our internal API. 
On the other hand, I don't have a better idea of how to solve that problem and it seems that this is still a better solution than copy-pasting or marking sample app methods with @InternalStreamChatApi annotation

adasiewiczr (dismissed):
LGTM

stream-chat-android-ui-components-sample/build.gradle
zsmb13: This is correct, we shouldn't need this in the sample app. The only place where using these internal APIs should come up is in the component browser, and there we should opt-in locally, for the given method or class.

zsmb13: Should we perhaps split out the component browser to a new module to avoid this issue? 🤔

ogkuzmin: Good point about copying this by customers. I think I can locally opt-in it

zsmb13: If it's something that the framework calls, you can add `@InternalStreamChatApi` on that method, which technically just propagates the opt-in requirement further, but happens to work out in that case.

Otherwise you can `@OptIn(InternalStreamChatApi::class)` locally, but that needs some Gradle config too, to enable using `@OptIn` itself 😄

ogkuzmin: I did locally `opt-in` and added a more generic compiler argument

zsmb13 (approved):
Good work 👍

adasiewiczr (approved):

PR #1049, opened 2020-12-18
Add extensions for setting soft input mode in Fragments

zsmb13:
### Description

We need different window handling behaviour depending on which Fragment we're in. Sometimes we want to resize the window so that it all fits above the keyboard, and sometimes we want to cover its bottom half with the keyboard.

This PR removes adds extensions for `Fragment` to dynamically set this for each screen.

https://user-images.githubusercontent.com/12054216/102590365-beb8db00-4110-11eb-9626-51437c4ffbed.mov

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelFragment.kt
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelFragment.kt
@@ -16,6 +16,7 @@ import io.getstream.chat.ui.sample.R
 import io.getstream.chat.ui.sample.common.initToolbar
 import io.getstream.chat.ui.sample.common.navigateSafely
 import io.getstream.chat.ui.sample.databinding.FragmentAddChannelBinding
+import io.getstream.chat.ui.sample.util.extensions.useAdjustResize
 
 class AddChannelFragment : Fragment() {
 
@@ -36,6 +37,11 @@ class AddChannelFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
         initToolbar(binding.toolbar)
         bindAddChannelView()
+    }
+
+    override fun onResume() {
+        super.onResume()
+        useAdjustResize()
     }
 
     override fun onDestroyView() {
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
@@ -19,6 +19,7 @@ import io.getstream.chat.android.ui.messages.header.bindView
 import io.getstream.chat.android.ui.textinput.bindView
 import io.getstream.chat.ui.sample.common.navigateSafely
 import io.getstream.chat.ui.sample.databinding.FragmentChatBinding
+import io.getstream.chat.ui.sample.util.extensions.useAdjustResize
 
 class ChatFragment : Fragment() {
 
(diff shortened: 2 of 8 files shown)


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelFragment.kt
```diff
@@ -38,6 +39,11 @@ class AddChannelFragment : Fragment() {
         bindAddChannelView()
     }
 
+    override fun onResume() {
+        super.onResume()
+        useAdjustResize()
```
bychkovdmitry, line 44: Can we remove android:windowSoftInputMode="adjustResize" from the manifest now?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/util/extensions/Fragment.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.fragment.app.Fragment
+
+fun Fragment.useAdjustNothing() = setSoftInputMode(SOFT_INPUT_ADJUST_NOTHING)
+
+fun Fragment.useAdjustPan() = setSoftInputMode(SOFT_INPUT_ADJUST_PAN)
+
+fun Fragment.useAdjustResize() = setSoftInputMode(SOFT_INPUT_ADJUST_RESIZE)
+
+private fun Fragment.setSoftInputMode(flags: Int) {
+    (activity ?: return).window.setSoftInputMode(flags)
```
bychkovdmitry, line 15: Maybe activity?.window?.setSoftInputMode(flags)?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
```diff
@@ -56,6 +57,11 @@ class ChatFragment : Fragment() {
         configureBackButtonHandling()
     }
 
+    override fun onResume() {
+        super.onResume()
```
bychkovdmitry, line 61: Good, I was worried that client would use incorrect resize mode which I defined in the manifest 👍

bychkovdmitry (approved):

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelFragment.kt
zsmb13: The really sad part is that this kind of dynamic resizing doesn't work if you don't have that in the manifest. I tried for a while, but it seems to be required, super weird. ([Same conclusion in this SO answer](https://stackoverflow.com/a/47629013/4465208))

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/util/extensions/Fragment.kt
zsmb13: It's the same thing, this just has fewer `?` in it 😄

adasiewiczr (approved):

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelFragment.kt
bychkovdmitry: Facepalm

PR #1050, opened 2020-12-18
Fix/watch channel

### LINKED ISSUE #1048: Failed message looks like it's sent
Failed message looks like this:

image

Error is:
"Can't find channel with id livestream:sandbox2"

JcMinarro:
### Description
Fix race-condition when a channel is watched but the socket connection is not online yet
Fix #1048

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -656,7 +656,7 @@ internal class ChatDomainImpl internal constructor(
         }
 
         // 4 recover events
-        if (isOnline()) {
+        if (recoveryNeeded && isOnline()) {
             replayEventsForActiveChannels()
         }
     }
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -662,7 +662,6 @@ internal class ChatDomainImpl internal constructor(
     }
 
     private suspend fun retryFailedEntities() {
-        delay(1000)
         // retry channels, messages and reactions in that order..
         val channelEntities = repos.channels.retryChannels()
         val messages = retryMessages()
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -598,12 +598,13 @@ internal class ChatDomainImpl internal constructor(
         // 0 ensure load is complete
         initJob.join()
 
+        // 1 Watch all channels that failed, it is needed for channels that are created locally, in other case it won't be created on the server
         activeChannelMapImpl
             .values
             .filter { it.errorWatching }
             .forEach { it.watch() }
 
-        // 1 update the results for queries that are actively being shown right now
+        // 2 update the results for queries that are actively being shown right now
         val updatedChannelIds = mutableSetOf<String>()
         val queriesToRetry = activeQueryMapImpl.values
             .toList()
@@ -623,7 +624,7 @@ internal class ChatDomainImpl internal constructor(
                 updatedChannelIds.addAll(response.data().map { it.cid })
             }
         }
-        // 2 update the data for all channels that are being show right now...
+        // 3 update the data for all channels that are being show right now...
         // exclude ones we just updated
         val cids: List<String> = activeChannelMapImpl
             .entries
@@ -650,12 +651,12 @@ internal class ChatDomainImpl internal constructor(
             }
         }
 
-        // 3 retry any failed requests
+        // 4 retry any failed requests
         if (isOnline()) {
             retryFailedEntities()
         }
 
-        // 4 recover events
+        // 5 recover events
         if (recoveryNeeded && isOnline()) {
             replayEventsForActiveChannels()
         }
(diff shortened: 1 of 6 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -598,6 +598,11 @@ internal class ChatDomainImpl internal constructor(
         // 0 ensure load is complete
         initJob.join()
 
+        activeChannelMapImpl
+            .values
+            .filter { it.errorWatching }
+            .forEach { it.watch() }
```
zsmb13, line 604: This might be worth a comment similar to the other ones in this method

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -651,13 +656,12 @@ internal class ChatDomainImpl internal constructor(
...
 
         // 4 recover events
-        if (isOnline()) {
+        if (recoveryNeeded && isOnline()) {
             replayEventsForActiveChannels()
         }
     }
 
     private suspend fun retryFailedEntities() {
-        delay(1000)
```
zsmb13, line 660: Do we know why this was here in the first place?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -431,15 +431,12 @@ internal class EventHandlerImpl(
...
-                    val recovered = domainImpl.isInitialized()
-
                     domainImpl.setOnline()
                     domainImpl.setInitialized()
-                    if (recovered && domainImpl.recoveryEnabled) {
-                        domainImpl.connectionRecovered(true)
-                    } else {
-                        domainImpl.connectionRecovered(false)
-                    }
+                    domainImpl.connectionRecovered(domainImpl.recoveryEnabled)
```
zsmb13, line 436: While we're here, looks like `recoveryEnabled` could be a `val`, because we never update it, we just set it at the creation of `ChatDomainImpl`.  (Also, `ChatDomainImpl` has an unused internal ctor?)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
JcMinarro: :+1:

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
JcMinarro: :+1:

zsmb13 (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -431,15 +431,12 @@ internal class EventHandlerImpl(
...
+                }
+                is HealthEvent -> {
+                    domainImpl.connectionRecovered(false)
```
tschellenbach, line 439: Doesn't this cause you to trigger connection recovered every 30 seconds?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
tschellenbach, line 436: the new code here doesn't differentiate between first connect and recovery flow

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -365,6 +365,7 @@ internal class ChannelControllerImpl(
             return
         }
         runChannelQuery(QueryChannelPaginationRequest(limit))
+        errorWatching = recoveryNeeded
```
tschellenbach, line 368: what's the difference between these 2 variables?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -598,7 +598,13 @@ internal class ChatDomainImpl internal constructor(
         // 0 ensure load is complete
         initJob.join()
 
-        // 1 update the results for queries that are actively being shown right now
+        // 1 Watch all channels that failed, it is needed for channels that are created locally, in other case it won't be created on the server
+        activeChannelMapImpl
+            .values
+            .filter { it.errorWatching }
+            .forEach { it.watch() }
```
tschellenbach, line 605: can be optimized by only calling watch for channels that we didn't receive after doing the query channels

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
JcMinarro: If it is first connect, there wasn't any pending channel, neither anyy calls to be done on the server

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
JcMinarro: Only for calls that were failing

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: Watch is only the first time. When you load more messages (older or newer), if they fails, we don't need to watch it again

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
JcMinarro: This list is going to be ~0 almost always

PR #1054, opened 2020-12-18
Bit fix fork

tschellenbach:
See the google doc about the 3 BIT issues

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/ChatEventsObservable.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/ChatEventsObservable.kt
@@ -1,25 +1,41 @@
 package io.getstream.chat.android.client.utils.observable
 
+import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.events.ChatEvent
 import io.getstream.chat.android.client.events.ConnectedEvent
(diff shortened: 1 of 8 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/ChatEventsObservable.kt
```diff
@@ -1,27 +1,41 @@
...
+import io.getstream.chat.android.client.logger.ChatLogger
 import io.getstream.chat.android.client.models.EventType
 import io.getstream.chat.android.client.socket.ChatSocket
 import io.getstream.chat.android.client.socket.SocketListener
 import java.util.Date
 
-internal class ChatEventsObservable(private val socket: ChatSocket) {
+internal class ChatEventsObservable(
+    private val socket: ChatSocket,
+    private var client: ChatClient
```
JcMinarro, line 18: It is wrong, we are coupling this class to our client, and it is not doing anything extra, because all subscriptions are called before the event is handled, the same that was done previously

tschellenbach: did you see the google doc i shared? it explains the root cause for some of these issues and what they are solving.

problem that we used to have here is this:

- client.observe<ConnectEvent> -> is called after the library knows its online
- client.setUser { is called earlier and before the library knows it's online }

With this change both the event and the callback are called after the library knows its online. 

There are of course many ways to solve the above problem, and I'm sure you can come up with a cleaner solution.

PR #1066, opened 2020-12-21
[CAS-483] Delivery state indicator

zsmb13:
https://stream-io.atlassian.net/browse/CAS-483

### Description

Adds a decorator to show pending / sent / seen state on messages.

#### Normal behaviour: 

https://user-images.githubusercontent.com/12054216/102864817-14ed8d00-4435-11eb-9cb8-b6ab07d7d754.mov 

#### Network outage and recovery:

https://user-images.githubusercontent.com/12054216/102864822-161eba00-4435-11eb-87d6-d0478e0a9349.mov

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/BaseDecorator.kt
```diff
@@ -0,0 +1,58 @@
...
+            is MessagePlainTextViewHolder -> decoratePlainTextMessage(viewHolder, data)
+            is OnlyMediaAttachmentsViewHolder -> decorateOnlyMediaAttachmentsMessage(viewHolder, data)
+            is PlainTextWithMediaAttachmentsViewHolder -> decoratePlainTextWithMediaAttachmentsMessage(viewHolder, data)
+            is OnlyFileAttachmentsViewHolder -> decorateOnlyFileAttachmentsMessage(viewHolder, data)
+            is PlainTextWithFileAttachmentsViewHolder -> decoratePlainTextWithFileAttachmentsMessage(viewHolder, data)
+            else -> Unit
+        }.exhaustive
+    }
+
+    protected abstract fun decoratePlainTextWithFileAttachmentsMessage(
```
ogkuzmin, line 29: Why did you make all these func abstract?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/DeliveryStatusDecorator.kt
```diff
@@ -0,0 +1,92 @@
...
+                    showIndicator(R.drawable.stream_ui_ic_check_all)
+                } else {
+                    showIndicator(R.drawable.stream_ui_ic_check_gray)
+                }
+            }
+            SyncStatus.FAILED_PERMANENTLY -> {
+                // This case is covered by a separate Decorator
+                hideIndicator()
+            }
+        }.exhaustive
```
adasiewiczr, line 90: Why do we need `exhaustive` here?

JcMinarro (approved):
LGTM

ogkuzmin (commented):
LGTM
Just question, why did you decide to make all funcs of BaseDecorators abstract?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/DeliveryStatusDecorator.kt
```diff
@@ -0,0 +1,92 @@
...
+        fun hideIndicator() {
+            imageView.isVisible = false
+        }
+
+        fun showIndicator(@DrawableRes drawableRes: Int) {
+            imageView.isVisible = true
+            imageView.setImageResource(drawableRes)
+        }
+
+        // TODO review this logic for the BOTTOM check
```
adasiewiczr, line 64: Do we still need that comment?

zsmb13: We do, sometimes things should be marked as being in a `BOTTOM` position, but they're not.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/DeliveryStatusDecorator.kt
zsmb13: It just makes sure that all values of the enum are covered.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/BaseDecorator.kt
zsmb13: Realized that having default implementations makes it easy to forget to add one of the overrides in a decorator (because I've done it the first time around). Feels better to be forced to make the decision that you want to override with an empty method.

ogkuzmin: I think it's kind of trade-off. None of these is the best (abstract and with empty impl). Why did then you add abstract for  decorate message deleted? Seems deleted message never requires to get decorated

zsmb13: I think being more explicit is better in this case, makes it harder to forget to decorate something. The kinds of ViewHolders we have are not completely obvious, e.g. someone might not remember that we have a separate one for `PlainTextWithFileAttachments` or something. Same goes for adding a new ViewHolder type, you might forget to update one of the decorators otherwise.

For the deleted ViewHolder: that might be the case, but making one method non-abstract would make it even trickier to work with the class. Based on all the others, you would assume that you're forced to implement the method for all ViewHolders, and then this one would be a surprising exception to that pattern. Consistency makes for safer code and fewer bugs 😄 .

ogkuzmin: I may agree about others, but such view holders as date divider, message deleted I'm not sure it worths to make such methods abstract

zsmb13: Alright, when we have more than one of those kinds of ViewHolders, we can add default implementations for them, then it's more obvious that some are required and some are not. (Though personally, I'd prefer to keep things one way, since it's not that verbose to do.)

PR #1070, opened 2020-12-22
Improve new chat screen

adasiewiczr:
Things covered by this PR:
- Showing `MessageListView` after clicking on `MessageInputView` if at least one member is added instead of showing it just after adding member
- Blocking send button when no members are added
- Hiding keyboard only if user name edit text has focus 
- Simplify displaying empty and result states - logic was moved to `AddChannelController` 

https://user-images.githubusercontent.com/17440581/102883261-0496db80-4450-11eb-9bbb-e146acaa7179.mp4

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelView.kt
```diff
@@ -52,6 +52,18 @@ class AddChannelView : FrameLayout {
...
+    override fun onInterceptTouchEvent(ev: MotionEvent): Boolean {
+        if (binding.messageInputView.isVisible && ev.action == MotionEvent.ACTION_DOWN) {
+            val messageInputViewRect = Rect().apply {
+                binding.messageInputView.getHitRect(this)
+            }
+            if (messageInputViewRect.contains(ev.x.toInt(), ev.y.toInt())) {
+                controller.messageInputViewClicked()
+            }
+        }
+        return super.onInterceptTouchEvent(ev)
```
zsmb13, line 64: Couldn't we somehow get this from `MessageInputView` in a cleaner way, with a listener or something?

zsmb13 (approved):
Looks good, one small question

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/add/AddChannelView.kt
adasiewiczr: It's not easy. `MessageInputView` has few children which consume on click event and I don't want to override `onInterceptTouchEvent ` inside `MessageInputView` cause I don't think discovering clicks on the whole view will be a common scenario.
I neither like this solution but I didn't find anything better 😞

zsmb13: And it's not enough to detect when its input field gets focused, we want all interactions with the view?

adasiewiczr: I think it's not enough because there are edge cases when input field won't gain focus (for example adding attachment).
It doesn't look good when you do that and you won't see channel preview

zsmb13: Alright, that makes sense, let's go with this then.

PR #1076, opened 2020-12-22
[CAS-428] Behaviour for Message Options

leandroBorgesFerreira:
[CAS-428](https://stream-io.atlassian.net/browse/CAS-428)

### Description

Adding the implementation for the message option. 

Is this PR I'm just binding the implementation that already exists to the list. We still need to create the behaviour for **reply**. Which is hidden.

This version: 
Screenshot_20201222-153823

New version: 
Screenshot_20201223-115640

### Checklist

- [ ] Changelog updated with client-facing changes. **No need**
- [ ] New code is covered by unit tests. **No need**
- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
@@ -822,14 +822,12 @@ public class MessageListView : ConstraintLayout {
     }
 
     public fun setOnBlockUserHandler(onUserBlockHandler: (User, Channel) -> Unit) {
-        val blockUserForThisChannel : (User) -> Unit = { user ->
+        val blockUserForThisChannel: (User) -> Unit = { user ->
             onUserBlockHandler(user, channel)
         }
 
         this.onBlockUserHandler = blockUserForThisChannel
-
     }
-
 
     public fun interface MessageClickListener {
         public fun onMessageClick(message: Message)
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
@@ -128,6 +128,9 @@ public class MessageListView : ConstraintLayout {
     private var onMessageReactionHandler: (Message, String) -> Unit = { _, _ ->
         throw IllegalStateException("onMessageReactionHandler must be set.")
     }
+    private var onUserMutedHandler: (User) -> Unit = {
+        throw IllegalStateException("onUserMutedHandler must be set.")
+    }
 
     private lateinit var messageOptionsConfiguration: MessageOptionsView.Configuration
 
@@ -157,6 +160,7 @@ public class MessageListView : ConstraintLayout {
                     threadReplyHandler = onStartThreadHandler,
                     editClickHandler = onMessageEditHandler,
                     flagClickHandler = onMessageFlagHandler,
+                    muteClickHandler = onUserMutedHandler,
                     deleteClickHandler = onMessageDeleteHandler
                 )
 
@@ -807,6 +811,10 @@ public class MessageListView : ConstraintLayout {
 
     public fun setOnMessageReactionHandler(onMessageReactionHandler: (Message, String) -> Unit) {
         this.onMessageReactionHandler = onMessageReactionHandler
+    }
+
+    public fun setOnUserMutedHandler(onUserMutedHandler: (User) -> Unit) {
+        this.onUserMutedHandler = onUserMutedHandler
     }
 
     public fun interface MessageClickListener {
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
@@ -169,11 +169,11 @@ public class MessageListView : ConstraintLayout {
                 )
 
                 MessageOptionsOverlayDialogFragment
-                    .newInstance(message.toMessageItemForOverlay(), messageOptionsConfiguration, handlers)
+                    .newInstance(message.toMessageItemForOverlay(), messageOptionsConfiguration)
                     .apply {
-                        setReactionClickListener { message, reactionType ->
-                            onMessageReactionHandler(message, reactionType)
-                        }
+                        setReactionClickListener(onMessageReactionHandler)
+
+                        setMessageOptionsHandlers(handlers)
                     }
                     .show(framentManager, MessageOptionsOverlayDialogFragment.TAG)
             }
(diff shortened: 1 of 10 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsOverlayDialogFragment.kt
```diff
@@ -140,22 +231,12 @@ internal class MessageOptionsOverlayDialogFragment : DialogFragment() {
...
-        }
-    }
+    internal class Handlers(
+        val threadReplyHandler: (Message) -> Unit,
+        val editClickHandler: (Message) -> Unit,
+        val flagClickHandler: (Message) -> Unit,
+        val muteClickHandler: (User) -> Unit,
+        val blockClickHandler: (User) -> Unit,
+        val deleteClickHandler: (Message) -> Unit
+    ) : Serializable
```
zsmb13, line 241: I never knew that Kotlin function types are serlializable, this is kind of shocking.

This seems extremely unsafe, given that these are instances of classes that implement the function interface. The code in there can have all sorts of references to other external objects (Views, ViewModels, Fragments, Activities, etc.), and those can not possibly be retained through serialization (or think about process death!).

I'd prefer to use the method we have in `ReactionsOverlayDialogFragment` for setting listeners on a nested Fragment, by calling a setter on it. That seems like it'll break after a process restoration too, but still seems a lot safer.

leandroBorgesFerreira: Let me take a look in the `ReactionsOverlayDialogFragment`

leandroBorgesFerreira: Moved the handlers as a setter instead of serialising it.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -757,6 +817,18 @@ public class MessageListView : ConstraintLayout {
...
+    public fun setOnMuteUserHandler(onUserMuteHandler: (User) -> Unit) {
+        this.onMuteUserHandler = onUserMuteHandler
+    }
+
+    public fun setOnBlockUserHandler(onUserBlockHandler: (User, Channel) -> Unit) {
+        val blockUserForThisChannel: (User) -> Unit = { user ->
+            onUserBlockHandler(user, channel)
+        }
+
+        this.onBlockUserHandler = blockUserForThisChannel
```
zsmb13, line 829: Really nice composition of functions 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsOverlayDialogFragment.kt
```diff
@@ -22,6 +27,26 @@ import java.io.Serializable
 
 internal class MessageOptionsOverlayDialogFragment : DialogFragment() {
 
+    companion object {
+        const val TAG = "messageOptions"
+
+        private const val ARG_MESSAGE = "message"
+        private const val ARG_OPTIONS_CONFIG = "optionsConfig"
+        private const val ARG_HANDLERS = "handlers"
```
zsmb13, line 35: This is now unused

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsView.kt
```diff
@@ -38,32 +72,63 @@ public class MessageOptionsView : FrameLayout {
...
         val blockIcon: Int,
         val deleteText: String,
-        val deleteIcon: Int
+        val deleteIcon: Int,
+        val copyTextEnabled: Boolean,
+        val deleteConfirmationEnabled: Boolean,
+        val deleteConfirmationTitle: String,
+        val deleteConfirmationMessage: String,
+        val deleteConfirmationPositiveButton: String,
+        val deleteConfirmationNegativeButton: String
```
zsmb13, line 90: Nit: there are a bunch of these long parameter / property lists on this PR where we could add a trailing commas to the last entry

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsView.kt
```diff
@@ -26,8 +27,41 @@ public class MessageOptionsView : FrameLayout {
...
+
+        binding.deleteTV.run {
+            text = configuration.deleteText
+            setTextColor(ContextCompat.getColor(context, R.color.stream_ui_light_red))
+            setCompoundDrawablesWithIntrinsicBounds(
+                ResourcesCompat.getDrawable(resources, configuration.deleteIcon, null),
+                null,
+                null,
+                null
+            )
```
zsmb13, line 63: How come this doesn't use `configureListItem` like all the other options above? Just because of the text color change?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsView.kt
```diff
@@ -26,8 +27,41 @@ public class MessageOptionsView : FrameLayout {
...
+            binding.flagTV.configureListItem(configuration.flagText, configuration.flagIcon, iconsTint)
+            binding.muteTV.configureListItem(configuration.muteText, configuration.muteIcon, iconsTint)
+            binding.blockTV.configureListItem(configuration.blockText, configuration.blockIcon, iconsTint)
+        } else {
+            binding.flagTV.isVisible = false
+            binding.muteTV.isVisible = false
+            binding.blockTV.isVisible = false
+        }
+
+        binding.deleteTV.run {
```
zsmb13, line 55: Should the _Delete_ option show up for both our own messages and messages of others? I assume we can only delete our own messages by default.

zsmb13 (commented):
Nice work, got a few questions

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsView.kt
leandroBorgesFerreira: Actually it makes sense to use it. Changed.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsView.kt
leandroBorgesFerreira: Added

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsOverlayDialogFragment.kt
leandroBorgesFerreira: Removed

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsView.kt
leandroBorgesFerreira: Yep, makes sense. Refactored so now it only appears for own messages

zsmb13 (approved):
Looks good to me 👍

PR #1082, opened 2020-12-22
Don't display loading indicator over channels from cache

bychkovdmitry:
### Description

This PR prevent channel list loading indicator from displaying over channel list received from cache. This is just a temp solution. We need to completely get rid of sealed classes in our view models and controllers:

```
data class State(
    val contentState: ContentState,
    val channels: List<Channel>,
    val loadingMore: Boolean,
    val hasMore: Boolean
)

enum class ContentState {
    CONTENT, LOADING, ERROR
}
```

### Checklist

- [X] Reviewers added

### Before
https://user-images.githubusercontent.com/9600921/102942457-6d209f80-44c6-11eb-9646-aad1e06f9334.mov

### After
https://user-images.githubusercontent.com/9600921/102942452-6a25af00-44c6-11eb-8ffc-8bef3644bec9.mov

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
@@ -26,7 +26,6 @@ public class ChannelListView @JvmOverloads constructor(
     defStyle: Int = 0
 ) : RecyclerView(context, attrs, defStyle) {
 
-    private val channelListItemAdapter: ChannelListItemAdapter
     private val layoutManager: ScrollPauseLinearLayoutManager
     private val scrollListener: EndReachedScrollListener = EndReachedScrollListener()
     private val dividerDecoration: SimpleVerticalListDivider = SimpleVerticalListDivider()
@@ -37,8 +36,7 @@ public class ChannelListView @JvmOverloads constructor(
         setHasFixedSize(true)
         layoutManager = ScrollPauseLinearLayoutManager(context)
         setLayoutManager(layoutManager)
-        channelListItemAdapter = ChannelListItemAdapter(parseStyleAttributes(context, attrs))
-        adapter = channelListItemAdapter
+        adapter = ChannelListItemAdapter(parseStyleAttributes(context, attrs))
         setSwipeListener(ChannelItemSwipeListener(this, layoutManager))
         setMoreOptionsClickListener { channel ->
             context.getFragmentManager()?.let { fragmentManager ->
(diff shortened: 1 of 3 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -26,16 +26,19 @@ public class ChannelListView @JvmOverloads constructor(
     defStyle: Int = 0
 ) : RecyclerView(context, attrs, defStyle) {
 
-    private var endReachedListener: EndReachedListener? = null
+    private val channelListItemAdapter: ChannelListItemAdapter
```
ogkuzmin, line 29: Why do we need to store one extra field? We can reuse just `adapter` for the same purpose, can't we?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -124,6 +127,10 @@ public class ChannelListView @JvmOverloads constructor(
         requireAdapter().submitList(channels)
     }
 
+    public fun hasChannels(): Boolean {
+        return channelListItemAdapter.itemCount > 0
```
ogkuzmin, line 131: `adapter?.itemCount ?: 0 > 0`
Isn't the same?

ogkuzmin (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
bychkovdmitry: There's a very minor difference, `BaseAdapter` not only provides items that were set to the adapter, but also pending items in async differ.

bychkovdmitry: So I felt, even though there is a very small chance chance that we invoke the method after we set channels but before async diff completes, it seems like the right thing to do.

ogkuzmin (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
bychkovdmitry: Ok, seems that I'm wrong here. Fixing it...

zsmb13 (approved):

PR #1099, opened 2020-12-27
CAS-429 Fix options overlay design

bychkovdmitry:
### Description

- Merged two overlays into one because they have so much in common
- Ignored some decorators on the overlay that are not appropriate (e.g. `ReactionsDecorator`) 
- `ownReactions` coming from `offline` is broken now, so temporarily relying only on `latestReactions`
- minor UI improvements according to the design

### Checklist

- [X] Reviewers added

| Message options overlay | Reactions overlay |
| --- | --- |
| photo_2020-12-28 01 20 03 | photo_2020-12-28 01 19 58 |


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Message.kt
```diff
@@ -93,4 +94,4 @@ public data class Message(
     @SerializedName("reply_to_message")
     var replyTo: Message? = null
 
-) : CustomObject
+) : CustomObject, Serializable
```
ogkuzmin, line 97: It's not right. We can't expose serializable mechanics min, such model, only for UI reason. Also, it could be even dangerous. Message class contains non-serializable fields (e.g. `List<Attachment>, List<User>` at least). It could be dangerous (I think we can get crash if the app is restored by Android on such a screen). How did it work before???
I think the proper way is to store message-id and extract it from the repo

ogkuzmin (changes requested):

ogkuzmin (approved):
Now looks good to me

PR #1108, opened 2020-12-29
Show user names in group chats

bychkovdmitry:
### Description

Added user names to message items according to the design. The conditions are as follows:
- Group chat
- Bottom message item
- Not a message of current user

As a side effect, had to refactor decorators.

### Checklist

- [X] Reviewers added

<https://user-images.githubusercontent.com/9600921/103288297-9de86380-49f5-11eb-92ef-7530b316f793.png src="" width="40%">


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/MessageListItemViewHolderFactory.kt
```diff
@@ -13,8 +12,11 @@ import io.getstream.chat.android.ui.messages.adapter.viewholder.OnlyMediaAttachm
 import io.getstream.chat.android.ui.messages.adapter.viewholder.PlainTextWithFileAttachmentsViewHolder
 import io.getstream.chat.android.ui.messages.adapter.viewholder.PlainTextWithMediaAttachmentsViewHolder
 import io.getstream.chat.android.ui.messages.adapter.viewholder.ThreadSeparatorViewHolder
+import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.DecoratorsProvider
 
-public open class MessageListItemViewHolderFactory(private val currentUser: User) {
+public open class MessageListItemViewHolderFactory(
+    private val decoratorsProvider: DecoratorsProvider
+) {
```
zsmb13, line 19: I think this is a good time to review our customization story for the Message List, I added that as a topic to our Trello board as well earlier today.

So, from here on out, it looks like we'll be allowing clients to:
 
- subclass this factory
- create their own implementation of `DecoratorsProvider` (and use that either in our factory, or their custom one)
- reuse our own existing `DecoratorsProvider` implementations
- create their own `Decorator` implementations

This prompts a bunch of questions:

- Do we need all of these, could we be more restrictive somewhere, expose less API?
- Do we see a chance of clients being able to write meaningful `Decorator`s, with how closely they're tied to the layouts of the ViewHolders?
- If we _do_ expose `Decorator`s, should we make the ones that are used by default `public` as well? They're currently internal.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/BaseMessageItemViewHolder.kt
```diff
@@ -3,45 +3,12 @@ package io.getstream.chat.android.ui.messages.adapter
...
-import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.MessageFooterDecorator
-import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.ReactionsDecorator
-import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.ReplyDecorator
-import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.ThreadRepliesDecorator
 
 public abstract class BaseMessageItemViewHolder<T : MessageListItem>(
-    currentUser: User,
-    itemView: View
+    itemView: View,
+    private val decorators: List<Decorator>
```
zsmb13, line 10: Like this a lot better with the decorators being passed in, good job 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/DecoratorsProvider.kt
```diff
@@ -0,0 +1,5 @@
+package io.getstream.chat.android.ui.messages.adapter.viewholder.decorator
+
+public interface DecoratorsProvider {
+    public fun getDecorators(): List<Decorator>
```
zsmb13, line 4: This could probably be just a `val`, looks like we're just returning a fixed, already existing list everywhere anyway

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/DecoratorsProvider.kt
```diff
@@ -0,0 +1,5 @@
+package io.getstream.chat.android.ui.messages.adapter.viewholder.decorator
+
+public interface DecoratorsProvider {
```
zsmb13, line 3: I'd name this `DecoratorProvider`, even if it provides a list of them - sounds nicer

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/Decorator.kt
```diff
@@ -3,6 +3,6 @@ package io.getstream.chat.android.ui.messages.adapter.viewholder.decorator
 import com.getstream.sdk.chat.adapter.MessageListItem
 import io.getstream.chat.android.ui.messages.adapter.BaseMessageItemViewHolder
 
-internal interface Decorator {
-    fun <T : MessageListItem> decorate(viewHolder: BaseMessageItemViewHolder<T>, data: T)
+public interface Decorator {
```
ogkuzmin, line 6: I think we'll need to figure out how not to expose it to customers. There is no sense in decorators being public

ogkuzmin (dismissed):
LGTM

zsmb13 (approved):
This is good to go, we'll review the public API for message ViewHolder customization later on

PR #1110, opened 2020-12-30
Add leaving channel

adasiewiczr:
### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
@@ -296,6 +296,20 @@ internal class QueryChannelsControllerImpl(
         }
     }
 
+    internal suspend fun removeChannel(cid: String) {
+        // Remove from queryChannelsSpec
+        if (queryChannelsSpec.cids.contains(cid)) {
+            queryChannelsSpec.cids = queryChannelsSpec.cids - cid
+            domainImpl.repos.queryChannels.insert(queryChannelsSpec)
+            // Remove from channel repository
+            domainImpl.repos.removeChannel(cid)
+
+            _channels.value = _channels.value.toMutableMap().apply {
+                remove(cid)
+            }
+        }
+    }
+
     /**
      * refreshes a single channel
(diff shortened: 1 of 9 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -296,6 +296,20 @@ internal class QueryChannelsControllerImpl(
...
+        // Remove from queryChannelsSpec
+        if (queryChannelsSpec.cids.contains(cid)) {
+            queryChannelsSpec.cids = queryChannelsSpec.cids - cid
+            domainImpl.repos.queryChannels.insert(queryChannelsSpec)
+            // Remove from channel repository
+            domainImpl.repos.removeChannel(cid)
+
+            _channels.value = _channels.value.toMutableMap().apply {
+                remove(cid)
+            }
```
zsmb13, line 309:
```suggestion
            _channels.value = _channels.value.minus(cid)
```

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/channels/ChannelsViewModel.kt
```diff
@@ -82,6 +82,10 @@ public class ChannelsViewModel(
         }.exhaustive
     }
 
+    public fun leaveChannel(channel: Channel) {
+        chatDomain.useCases.leaveChannel(channel.cid).enqueue()
+    }
```
zsmb13, line 87: This is lacking some error handling, but not sure if we have any specific behaviour we'd want here on an error

zsmb13 (approved):
Looks good to me in general 👍

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/channels/ChannelsViewModel.kt
adasiewiczr: For now, channel won't be removed from the list without any error. We also don't handle error in hiding channel method so I would leave it for now, but definitely, we need to think how to handle such cases

zsmb13 (approved):

PR #1111, opened 2020-12-30
[cas-484] Thread replies footnote for different message types

samiuelson:
https://stream-io.atlassian.net/browse/CAS-484

### Description

Adding thread replies count label for all of the available message types. 

Screenshot 2020-12-31 at 12 39 36
Screenshot 2020-12-31 at 12 39 25
Screenshot 2020-12-31 at 12 38 54
Screenshot 2020-12-31 at 12 38 29

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/ThreadRepliesDecorator.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/ThreadRepliesDecorator.kt
@@ -1,13 +1,13 @@
 package io.getstream.chat.android.ui.messages.adapter.viewholder.decorator
 
 import android.content.res.Resources
+import android.view.View
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.constraintlayout.widget.ConstraintSet
 import androidx.core.view.isVisible
 import com.getstream.sdk.chat.adapter.MessageListItem
 import com.getstream.sdk.chat.utils.extensions.updateConstraints
 import io.getstream.chat.android.ui.R
-import io.getstream.chat.android.ui.databinding.StreamUiItemMessageFootnoteBinding
 import io.getstream.chat.android.ui.databinding.StreamUiMessageThreadsFootnoteBinding
 import io.getstream.chat.android.ui.messages.adapter.viewholder.GiphyViewHolder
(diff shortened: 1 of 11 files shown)


zsmb13:
On the design, the timestamps and the thread reply info are at the same height and are the same font size. Looking at the screenshots here, I think we're off by a bit on one of them.

image

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/ThreadRepliesDecorator.kt
```diff
@@ -17,28 +23,187 @@ internal class ThreadRepliesDecorator : BaseDecorator() {
...
     }
 
-    override fun decorateGiphyMessage(viewHolder: GiphyViewHolder, data: MessageListItem.MessageItem) = Unit
+    override fun decorateGiphyMessage(viewHolder: GiphyViewHolder, data: MessageListItem.MessageItem) {
+        setupThreadRepliesView(viewHolder.binding, data)
+    }
+
+    private fun setupThreadRepliesView(
+        binding: StreamUiItemMessagePlainTextWithFileAttachmentsBinding,
+        data: MessageListItem.MessageItem
```
zsmb13, line 61: Looks like we're duplicating a lot of code here in this decorator. Instead of passing in the binding to these methods, you could have a single method and pass in the Views to manipulate (the root, and then the `threadRepliesFootnote`) as two parameters to it.

Take a look at some of the other decorators, like `ReplyDecorator` or `MessageFooterDecorator` for this pattern.

samiuelson: I've made the `setupThreadRepliesView()` reusable. Let me know if you can see a better way of doing this.

samiuelson:
> On the design, the timestamps and the thread reply info are at the same height and are the same font size. Looking at the screenshots here, I think we're off by a bit on one of them.
> 
> image

Cleaned up text sizes, dimens, and alignment of views.
Got rid of `android:fontWeight="600"`, replaced by `android:textStyle="bold"` to be compatible with all supporte OS versions.

leandroBorgesFerreira (approved):
looks good to me

leandroBorgesFerreira (approved):

PR #1118, opened 2021-01-04
[CAS-497] Feature/gallery share

JcMinarro:
### Description
Share pictures from GalleryView 

https://user-images.githubusercontent.com/4047514/103540466-0625c200-4e9a-11eb-920f-97fcbd55d695.mp4

### Checklist
- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -37,13 +57,22 @@ public class AttachmentGallery : ConstraintLayout {
     }
 
     public fun provideImageList(fragmentActivity: FragmentActivity, imageList: List<String>, currentIndex: Int = 0) {
-        binding.attachmentGallery.adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList)
+        adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList)
+        binding.attachmentGallery.adapter = adapter
         configPositionCount(imageList.size)
         binding.attachmentGallery.setCurrentItem(currentIndex, false)
+        binding.shareButton.setOnClickListener {
+            CoroutineScope(DispatcherProvider.IO).launch {
```
zsmb13, line 65: This can be `GlobalScope.launch(DispatcherProvider.Main) { ... }` and then we can drop the `withContext` call later on. `ImageLoader` has proper suspending functions, so it won't block the main thread.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -38,6 +42,29 @@ public object ImageLoader {
...
+
+    private fun Bitmap.getLocalBitmapUri(context: Context): Uri? {
+        var bmpUri: Uri? = null
+        try {
+            val file = File(
+                context.getExternalFilesDir(Environment.DIRECTORY_PICTURES),
+                "share_image_" + System.currentTimeMillis() + ".png"
+            )
+            val out = FileOutputStream(file)
+            compress(Bitmap.CompressFormat.PNG, 90, out)
```
zsmb13, line 59: I'm curious, why png instead of jpeg?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -9,14 +11,32 @@ import androidx.core.content.res.use
...
 import io.getstream.chat.android.ui.R
 import io.getstream.chat.android.ui.databinding.StreamUiAttachmentGalleryBinding
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
 public class AttachmentGallery : ConstraintLayout {
 
+    private var onSharePictureListener: (pictureUri: Uri) -> Unit = { pictureUri ->
+        ContextCompat.startActivity(
```
zsmb13, line 25: I'm getting a crash on this on a physical device, seems something is not right with where we place this image / what URI we use for it:

`android.os.FileUriExposedException: file:///storage/emulated/0/Android/data/io.getstream.chat.ui.sample.debug/files/Pictures/share_image_1609768254614.png exposed beyond app through ClipData.Item.getUri()`

zsmb13 (changes requested):

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/ImageLoader.kt
JcMinarro: JPEG doesn't support alpha channel and quality usually is lower

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
JcMinarro: It seems an issue with the new File API requirement.
Let me update it

zsmb13: Yeah, it's gonna be something like that. This was on Android 10, it's also worth testing on Android 11 though.

zsmb13: This crash is gone, also tested on Android 11, file permissions look all good now

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileProvider.kt
```diff
@@ -1,5 +1,22 @@
...
+import android.content.ComponentName
+import android.content.Context
+import android.net.Uri
 import androidx.core.content.FileProvider
+import java.io.File
 
-public class StreamFileProvider : FileProvider()
+public class StreamFileProvider : FileProvider() {
+
+    public companion object {
```
zsmb13, line 11: `@InternalStreamChatApi`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -9,14 +11,33 @@ import androidx.core.content.res.use
...
+            context,
+            Intent.createChooser(
+                Intent(Intent.ACTION_SEND).apply {
+                    type = "image/*"
+                    putExtra(Intent.EXTRA_STREAM, pictureUri)
+                    addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                    data = pictureUri
+                },
+                "Share Image"
```
zsmb13, line 33: This is a hardcoded UI string

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -9,14 +11,33 @@ import androidx.core.content.res.use
...
+    private var onSharePictureListener: (pictureUri: Uri) -> Unit = { pictureUri ->
+        ContextCompat.startActivity(
+            context,
+            Intent.createChooser(
+                Intent(Intent.ACTION_SEND).apply {
+                    type = "image/*"
+                    putExtra(Intent.EXTRA_STREAM, pictureUri)
+                    addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+                    data = pictureUri
+                },
```
zsmb13, line 32: I think there's _something_ wrong with this intent, but I couldn't debug the cause of the issues.

This is what I'm seeing with different target apps:
- Telegram: it works okay, but it looks like I've shared a file and not an image
- Telegram X: broken, toast appears with "this type of media is not yet supported"
- Messenger Lite: does nothing at all (doesn't even go the app, just closes the share sheet)
- Flamingo (Twitter client): opens a new tweet but the image is missing from it

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -38,6 +42,29 @@ public object ImageLoader {
...
+    private fun Bitmap.getLocalBitmapUri(context: Context): Uri? {
+        var bmpUri: Uri? = null
+        try {
+            val file = File(
+                context.getExternalFilesDir(Environment.DIRECTORY_PICTURES) ?: context.cacheDir,
+                "share_image_" + System.currentTimeMillis() + ".png"
+            )
+            val out = FileOutputStream(file)
+            compress(Bitmap.CompressFormat.PNG, 90, out)
+            out.close()
```
leandroBorgesFerreira, line 60: This will cause resource leaks. If the code crashes before `out.close` gets called the resource will stay open and memory won't be free. You can use `FileOutputStream(file).use` or use a `try/catch/finally` structure and call `close()` inside the finally block.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -37,13 +58,20 @@ public class AttachmentGallery : ConstraintLayout {
...
     public fun provideImageList(fragmentActivity: FragmentActivity, imageList: List<String>, currentIndex: Int = 0) {
-        binding.attachmentGallery.adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList)
+        adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList)
+        binding.attachmentGallery.adapter = adapter
         configPositionCount(imageList.size)
         binding.attachmentGallery.setCurrentItem(currentIndex, false)
+        binding.shareButton.setOnClickListener {
+            GlobalScope.launch(DispatcherProvider.Main) {
+                ImageLoader.getBitmapUri(context, adapter.getItem(binding.attachmentGallery.currentItem))
+                    ?.let { pictureUri -> onSharePictureListener(pictureUri) }
```
leandroBorgesFerreira, line 68: Function reference goes well here `.let(::onSharePictureListener)`

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -38,6 +42,29 @@ public object ImageLoader {
...
     }
 
+    public suspend fun getBitmapUri(
+        context: Context,
+        url: String,
+        transformation: ImageTransformation = ImageTransformation.None
+    ): Uri? = getBitmap(context, url, transformation)?.getLocalBitmapUri(context)
+
+    private fun Bitmap.getLocalBitmapUri(context: Context): Uri? {
+        var bmpUri: Uri? = null
```
leandroBorgesFerreira, line 52: No need to create this `bmpUri` variable. You can inline the return with the `try` block and return the last statement there.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/ImageLoader.kt
zsmb13: +1 for `use` here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
bychkovdmitry: `setType` and `setData` are mutually exclusive, so `setData` overrides the mime type: 

```
    public @NonNull Intent setType(@Nullable String type) {
        mData = null;
        mType = type;
        return this;
    }
    
        public @NonNull Intent setData(@Nullable Uri data) {
        mData = data;
        mType = null;
        return this;
    }
```
Try removing `data = pictureUri`

bychkovdmitry: At least without `data = pictureUri` I see that image is shared as an image with Telegram

bychkovdmitry (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
zsmb13: Nice catch, removing that fixes all of my sharing issues with the apps I'm testing with indeed. Also still works on both my Android 10 and Android 11 devices.

zsmb13:
Updated this PR to address all the feedback on it so far

bychkovdmitry (approved):

adasiewiczr (approved):

zsmb13 (approved):

PR #1121, opened 2021-01-04
[CAS-419] Attachment images menu

leandroBorgesFerreira:
[CAS-419](https://stream-io.atlassian.net/browse/CAS-419)

### Description

Making a image menu for the attachment gallery. 

giphy

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
@@ -97,7 +97,7 @@ public class AttachmentGallery : ConstraintLayout {
             ).let(binding.photoCount::setTextColor)
 
             imagesMenuTitle = tArray.getString(R.styleable.AttachmentGallery_streamUiImagesMenuTitle)
-            ?: context.getString(R.string.stream_ui_images_menu_default_title)
+                ?: context.getString(R.string.stream_ui_images_menu_default_title)
         }
     }
 
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
@@ -124,7 +124,7 @@ public class AttachmentGallery : ConstraintLayout {
             context.getFragmentManager()?.let { fragmentManager ->
                 ImagesMenuDialogFragment.newInstance(imagesMenuTitle, imageList)
                     .apply {
-                        setImageClickListener { _, position ->
+                        setImageClickListener { position ->
                             binding.attachmentGallery.setCurrentItem(position, true)
                         }
                     }
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
@@ -42,7 +42,7 @@ public class AttachmentGallery : ConstraintLayout {
 
     private var imagesMenuTitle: String = ""
 
-    private val imageList: MutableList<String> = mutableListOf()
+    private var imageList: List<String> = mutableListOf()
 
     public constructor(context: Context) : super(context)
 
@@ -69,8 +69,7 @@ public class AttachmentGallery : ConstraintLayout {
         configPositionCount(imageList.size)
         binding.attachmentGallery.setCurrentItem(currentIndex, false)
 
-        this.imageList.clear()
-        this.imageList.addAll(imageList)
+        this.imageList = imageList
 
         binding.shareButton.setOnClickListener {
             GlobalScope.launch(DispatcherProvider.Main) {
@@ -97,7 +96,8 @@ public class AttachmentGallery : ConstraintLayout {
                 ContextCompat.getColor(context, R.color.stream_ui_text_color_strong)
             ).let(binding.photoCount::setTextColor)
 
-            imagesMenuTitle = tArray.getString(R.styleable.AttachmentGallery_streamUiImagesMenuTitle) ?: "Photos"
+            imagesMenuTitle = tArray.getString(R.styleable.AttachmentGallery_streamUiImagesMenuTitle)
+            ?: context.getString(R.string.stream_ui_images_menu_default_title)
         }
     }
 
@@ -126,6 +126,7 @@ public class AttachmentGallery : ConstraintLayout {
                     .apply {
                         setImageClickListener { position ->
                             binding.attachmentGallery.setCurrentItem(position, true)
+                            dismiss()
                         }
                     }
                     .show(fragmentManager, null)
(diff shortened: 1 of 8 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -58,6 +68,8 @@ public class AttachmentGallery : ConstraintLayout {
                 R.styleable.AttachmentGallery_streamUiCountTextColor,
                 ContextCompat.getColor(context, R.color.stream_ui_text_color_strong)
             ).let(binding.photoCount::setTextColor)
+
+            imagesMenuTitle = tArray.getString(R.styleable.AttachmentGallery_streamUiImagesMenuTitle) ?: "Photos"
```
zsmb13, line 72: Hardcoded string here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/menu/ImagesMenuAdapter.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.recyclerview.widget.ListAdapter
+import androidx.recyclerview.widget.RecyclerView
+import coil.load
+import io.getstream.chat.android.ui.R
+import io.getstream.chat.android.ui.databinding.StreamUiItemAttachmentImageBinding
+
+internal class ImagesMenuAdapter(
+    private val imageList: List<String>,
+    private val clickListener: (Int) -> Unit
+) : ListAdapter<String, ImagesMenuAdapter.ImageViewHolder>(diffUtil()) {
```
zsmb13, line 15: We should use `submitList` to provide the items to `ListAdapter`, if we want its diffing functionality, because it's supposed to maintain its own list internally, and we have our own here. Or we can just use a regular `RecyclerView.Adapter` if we don't need the diffing at all in this case.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/menu/ImagesMenuAdapter.kt
```diff
@@ -0,0 +1,53 @@
+package io.getstream.chat.android.ui.images.menu
+
+import android.view.LayoutInflater
+import android.view.ViewGroup
+import androidx.recyclerview.widget.DiffUtil
+import androidx.recyclerview.widget.ListAdapter
+import androidx.recyclerview.widget.RecyclerView
+import coil.load
```
zsmb13, line 8: Let's use `ImageLoader` instead of Coil directly

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/menu/ImagesMenuDialogFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+import android.view.ViewGroup
+import androidx.recyclerview.widget.GridLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.getstream.sdk.chat.utils.GridSpacingItemDecoration
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import io.getstream.chat.android.ui.R
+import io.getstream.chat.android.ui.databinding.StreamUiImageMenuDialogBinding
+
+private const val SPAN_COUNT = 3
+private const val SPACING: Int = 2
```
zsmb13, line 15: These can go in the `companion`, they belong to the class directly

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/menu/ImagesMenuDialogFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+
+internal class ImagesMenuDialogFragment : BottomSheetDialogFragment() {
+
+    private var _binding: StreamUiImageMenuDialogBinding? = null
```
zsmb13, line 19: `onDestroyView` is missing, so we're never clearing this property

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -78,4 +90,18 @@ public class AttachmentGallery : ConstraintLayout {
...
     }
+
+    private fun configureImagesMenu() {
+        binding.menuButton.setOnClickListener {
+            context.getFragmentManager()?.let { fragmentManager ->
+                ImagesMenuDialogFragment.newInstance(imagesMenuTitle, imageList)
+                    .apply {
+                        setImageClickListener { position ->
+                            binding.attachmentGallery.setCurrentItem(position, true)
+                        }
```
zsmb13, line 101: Feels to me like we should also dismiss the dialog after we select an image in it, because otherwise it blocks the image we've scrolled to. What do you think?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -34,12 +40,16 @@ public class AttachmentGallery : ConstraintLayout {
...
+        configureImagesMenu()
     }
 
     public fun provideImageList(fragmentActivity: FragmentActivity, imageList: List<String>, currentIndex: Int = 0) {
         binding.attachmentGallery.adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList)
         configPositionCount(imageList.size)
         binding.attachmentGallery.setCurrentItem(currentIndex, false)
+
+        this.imageList.clear()
+        this.imageList.addAll(imageList)
```
zsmb13, line 52: Perhaps having a `var` `List` would be simpler here? It would also ensure we're only modifying the entire thing at the same time.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
leandroBorgesFerreira: Makes sense

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/menu/ImagesMenuAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    internal class ImageViewHolder(
+        private val binding: StreamUiItemAttachmentImageBinding
+    ) : RecyclerView.ViewHolder(binding.root) {
+
+        fun bind(imageUrl: String) {
+            binding.image.load(imageUrl, R.drawable.stream_placeholder) {
+                error(R.drawable.stream_placeholder)
```
zsmb13, line 35: ❗️  This now calls [`error`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/error.html) when the loading is completed

zsmb13 (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/menu/ImagesMenuAdapter.kt
leandroBorgesFerreira: That's why is was not working 🤦

zsmb13 (approved):

PR #1128, opened 2021-01-05
Feature/improve loading process

JcMinarro:
### Description
Improve loading process from our offline library

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -435,59 +435,38 @@ internal class ChannelControllerImpl(
         val queryOfflineJob = domainImpl.scope.async { runChannelQueryOffline(pagination) }
 
         // start the online query before queryOfflineJob.await
-        val queryOnlineJob = if (domainImpl.isOnline()) {
-            domainImpl.scope.async { runChannelQueryOnline(pagination) }
-        } else {
-            // if we are not offline we mark it as needing recovery
-            recoveryNeeded = true
```
zsmb13, line 442: We lose this line both here and in the other controller with these changes. Do we not need to set this flag when we're offline?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -435,59 +435,38 @@ internal class ChannelControllerImpl(
...
-        val localChannel = queryOfflineJob.await()
-        if (localChannel != null) {
+        val queryOnlineJob = domainImpl.scope.async { runChannelQueryOnline(pagination) }
+        val localChannel = queryOfflineJob.await()?.also {
             if (pagination.messageFilterDirection == Pagination.LESS_THAN) {
-                updateOldMessagesFromLocalChannel(localChannel)
+                updateOldMessagesFromLocalChannel(it)
             } else {
-                updateLiveDataFromLocalChannel(localChannel)
+                updateLiveDataFromLocalChannel(it)
```
zsmb13, line 443: `it` could be named better here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -435,59 +435,38 @@ internal class ChannelControllerImpl(
...
                     recoveryNeeded = true
                 }
-            }
-            response
-        } else {
-            if (localChannel != null) {
-                Result(localChannel)
-            } else {
-                Result(ChatError("Local channel was not found"))
+                localChannel?.let { Result(it) } ?: onlineResult
```
zsmb13, line 458: On this branch, `onlineResult` was an error, are we sure we want to return that if we don't have local results either? (Same in the other controller)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -435,59 +435,38 @@ internal class ChannelControllerImpl(
...
         }
         loader.value = false
         return result
     }
 
-    suspend fun runChannelQueryOffline(pagination: QueryChannelPaginationRequest): Channel? {
-        val selectedChannel = domainImpl.selectAndEnrichChannel(cid, pagination)
-
-        selectedChannel?.also { channel ->
-            channel.config = domainImpl.getChannelConfig(channel.type)
```
zsmb13, line 484: Do we not need this line that updates the channel config anymore?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: This flag is updated into `runChannelQueryOnline()` when the query fails, so it is not needed anymore here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: Yep, We will have a more clear error message on that case
Previously we had:
if online -> onlineResult (success/failure)
else -> if local -> localResult -> else -> Faiulre (Error without any helpful message)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
JcMinarro: `domainImpl.selectAndEnrichChannel()` already inject `config` into the returned channel, this call was redundant

zsmb13: Nice catch 👍

zsmb13 (dismissed):
LGTM

adasiewiczr (approved):

zsmb13 (approved):

PR #1145, opened 2021-01-06
[CAS-512] Original name for attachments

leandroBorgesFerreira:
[CAS-512](https://stream-io.atlassian.net/browse/CAS-512)

### Description

This PR removes the timestamp and the STREAM from the name of the attachment file. This way the user sees the original name, instead of just STREAM_202... 

Edit:

Removing `STREAM_` from the prefix and removing the maxline=1 from the textview. Now the item can show files with title with multiple lines. 

_The last one is the name with the change in the code._ 
Screenshot_20210106-163246

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
@@ -134,9 +134,7 @@ public class StorageHelper {
         return attachments
     }
 
-    private fun getFileName(attachmentMetaData: AttachmentMetaData): String =
-        "${attachmentMetaData.getTitleWithExtension()}_STREAM_${dateFormat.format(Date().time)}"
-}
+    private fun getFileName(attachmentMetaData: AttachmentMetaData): String = attachmentMetaData.getTitleWithExtension()
 
 private fun AttachmentMetaData.getTitleWithExtension(): String {
     val extension = MimeTypeMap.getFileExtensionFromUrl(title)
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
@@ -13,6 +13,7 @@ import com.getstream.sdk.chat.model.ModelType
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import java.io.File
 import java.text.SimpleDateFormat
+import java.util.Date
 import java.util.Locale
 
 @InternalStreamChatApi
@@ -133,7 +134,11 @@ public class StorageHelper {
         return attachments
     }
 
-    private fun getFileName(attachmentMetaData: AttachmentMetaData): String = attachmentMetaData.getTitleWithExtension()
+    private fun getFileName(attachmentMetaData: AttachmentMetaData): String {
+        return attachmentMetaData.run {
+            "${title}_STREAM_${dateFormat.format(Date().time)}.${getExtension()}"
+        }
+    }
 }
 
 private fun AttachmentMetaData.getTitleWithExtension(): String {
@@ -144,6 +149,10 @@ private fun AttachmentMetaData.getTitleWithExtension(): String {
         // TODO: Attachment's title should never be null. Review AttachmentMetaData class
         title ?: ""
     }
+}
+
+private fun AttachmentMetaData.getExtension(): String? {
+    return MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType)
 }
 
 internal sealed class MediaType(val contentUri: Uri, val modelType: String)
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
@@ -135,7 +135,7 @@ public class StorageHelper {
     }
 
     private fun getFileName(attachmentMetaData: AttachmentMetaData): String =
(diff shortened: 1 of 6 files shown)


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
```diff
@@ -134,8 +133,7 @@ public class StorageHelper {
         return attachments
     }
 
-    private fun getFileName(attachmentMetaData: AttachmentMetaData): String =
-        "STREAM_${dateFormat.format(Date().time)}_${attachmentMetaData.getTitleWithExtension()}"
```
zsmb13, line 138: This was here to avoid problems when multiple files with the same name are present, see the commit and its description here https://github.com/GetStream/stream-chat-android/commit/0c553c82a9c717b6ee27271bdc2f8a1ae672ae1a. Won't we have problems with that with the changes introduced here?

zsmb13: (That commit only introduced the milliseconds part specifically, but apparently even that was necessary?)

leandroBorgesFerreira: 🤔  I didn't dig much in it. We can add the mili seconds in the end, anyways

adasiewiczr: As far as I remember, the milliseconds part was added because every GIF sent from the keyboard has the same name. If you were sending them fast enough (or there was another case when you were sending multiple files with the same name) cached files were overridden and you were sending n * the same image instead of correct ones.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
```diff
@@ -134,8 +134,11 @@ public class StorageHelper {
...
+    private fun getFileName(attachmentMetaData: AttachmentMetaData): String {
+        return attachmentMetaData.run {
+            "${title}_STREAM_${dateFormat.format(Date().time)}.${getExtension()}"
```
adasiewiczr, line 139: Usually, the title contains a file extension and I think it would look a bit weird with our suffix (test.pdf_STREAM_date.pdf). Maybe we should strip extension from the title?

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
```diff
@@ -134,8 +134,11 @@ public class StorageHelper {
...
+        }
+    }
 }
 
 private fun AttachmentMetaData.getTitleWithExtension(): String {
```
adasiewiczr, line 144: It seems that this is no longer used and can be removed

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
leandroBorgesFerreira: I kept the prefix, but now it is smaller and we can show items that are more than one line long.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
leandroBorgesFerreira: I kept prefix instead of sufix, just removed the `STREAM_` part. And also the item can show multiple lines.

leandroBorgesFerreira: Edit: Now the prefix is added in the file, but taken away when the file is shown in the adapter.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
leandroBorgesFerreira: Edit: The prefix is added to the file, but the prefix is removed when showing the title.

zsmb13: At this point, couldn't we revert to the old prefix of `STREAM_yyyyMMdd_HHmmssSSS_`?

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StringUtils.kt
```diff
@@ -0,0 +1,14 @@
+package com.getstream.sdk.chat.utils
+
+public object StringUtils {
```
zsmb13, line 3: This probably doesn't have to be public API, at the very least it can be `@InternalStreamChatApi`

leandroBorgesFerreira: Makes sense, internal is much better.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
leandroBorgesFerreira: We need something constant in the stream in the beginning and end of the prefix. We could use `STREAM_yyyyMMdd_HHmmssSSS_prefix`. What do you think?

leandroBorgesFerreira: Nevermind... all we have to do it to replace it the first time, the original one can be kept.

adasiewiczr (approved):
LGTM

PR #1146, opened 2021-01-06
Update tint of thread replies ornaments

samiuelson:
https://stream-io.atlassian.net/browse/CAS-510
### Description

Update of "ornament" tint to match bubble background.
Screenshot 2021-01-07 at 10 34 13
Screenshot 2021-01-07 at 10 33 58

### Checklist

- [x] Reviewers added


leandroBorgesFerreira:
I think this part gets tinted... Let me check

leandroBorgesFerreira (dismissed):
I believe this is not changing the colour. This is where you should do it:

```
<androidx.appcompat.widget.AppCompatImageView
        android:id="@+id/threadsOrnamentLeft"
        android:layout_width="13dp"
        android:layout_height="33dp"
        android:src="@drawable/stream_ui_thread_replies_ornament_left"
        app:tint="@color/stream_ui_thread_reply_stroke"
        />
```

I think we should use the colour black in the drawables and then we set the tint the way we prefer. But if you prefer to do in the XML of the drawable, delete the tint line and then it will be fine.

zsmb13:
Agreed with the comment above, black + tinting should be preferred. Checking the screenshots in the PR description, I believe the ornaments are the same colour on both too.

bychkovdmitry:
We need to review all the icons as there are lots of icons with tint applied directly in corresponding drawables

zsmb13 (approved):
LGTM

PR #1150, opened 2021-01-06
Using pretty printing in our logs instead of huge lines

leandroBorgesFerreira:
### Description

Sorry for taking forever to do this 😅 

This PR adds pretty printing to our logger. Now we can see the json of the response body in a better way than a one giant line. 

The fact that were are removing Gson will eventually be a problem, but we can to substitute this logic with something using Moshi in a latter PR (or this one, no strong opinion about it)

### Checklist

- [x] Reviewers added


carterhudson (approved):
lgtm

zsmb13 (commented):
Looks good so far, two notes here:

- Lines are still getting cut by logcat because it has a max "line" limit, we could work around with [something like this](https://stackoverflow.com/a/7606118/4465208) if we wanted to display the entire response (would be huge in some cases though)
- We could also pretty print events if we wanted to, that would have to be added to `EventsParser` (we'd have to make sure we check the log level there before doing this double-parsing, because it shouldn't run in prod / if we don't need to log)

leandroBorgesFerreira:
I'll merge this and take a look at your comments in another PR @zsmb13

PR #1153, opened 2021-01-07
[CAS-513] Change unmuting user related methods return type to Call<Unit>

adasiewiczr:
Unmuting user was causing a crash because of the wrong return type.
We are getting `{"duration":"10.63ms"}` response instead of `Mute` class.
It was working "fine" before fixing `Result` class in #1069 just because we were able to create `Result` with both `data` and `error` as nulls

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitApi.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitApi.kt
@@ -256,7 +256,7 @@ internal interface RetrofitApi {
         @Query("user_id") userId: String,
         @Query("client_id") connectionId: String,
         @Body body: MuteUserRequest
-    ): RetrofitCall<MuteUserResponse>
+    ): RetrofitCall<CompletableResponse>
 
     @POST("/moderation/flag")
     fun flag(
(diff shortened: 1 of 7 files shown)


samiuelson (approved):
Nice catch 👍🏼

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/RetrofitApi.kt
```diff
@@ -256,7 +256,7 @@ internal interface RetrofitApi {
         @Query("user_id") userId: String,
         @Query("client_id") connectionId: String,
         @Body body: MuteUserRequest
-    ): RetrofitCall<MuteUserResponse>
+    ): RetrofitCall<CompletableResponse>
```
zsmb13, line 259: Did we check with the backend team that this was an intentional change in the API on their end?

adasiewiczr: I've double-checked that with other SDK and iOS, Flutter and js-client return `EmptyResponse` instead of `Mute` so I assumed that we've been returning `Mute` by mistake

zsmb13 (approved):

PR #1156, opened 2021-01-07
Hotfix/client state

JcMinarro:
### Description
Add unhandled states to our state machine

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/ClientStateServiceTests.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/ClientStateServiceTests.kt
@@ -26,6 +26,15 @@ internal class ClientStateServiceTests {
     }
 
     @Test
+    fun `Given Idle state When disconnected requested Should stay in Idle state`() {
+        val sut = Fixture().please()
+
+        sut.onDisconnectRequested()
+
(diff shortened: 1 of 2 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/ClientStateServiceTests.kt
```diff
@@ -25,6 +25,25 @@ internal class ClientStateServiceTests {
...
+    fun `Given Idle state When disconnected requested Should stay in Idle state`() {
+        val sut = Fixture().please()
+
+        sut.onDisconnectRequested()
+
+        sut.state shouldBeEqualTo ClientState.Idle
+    }
+
+    @Test
+    fun `Given Idle state When connected Should stay in Idle state`() {
```
zsmb13, line 38: Is the "Connected" event in this state invalid (we shouldn't have received it), so we're ignoring it?

zsmb13: Discussed this offline, we're not 100% sure why this event happens in this case. This will fix the crash that it produces, and we can keep investigating this later.

zsmb13 (approved):

PR #1158, opened 2021-01-07
Fixing error when connection doesn't work

leandroBorgesFerreira:
### Description

In this PR the app disconnects from the chat if the connection fails. The app was crashing because a wrong token was persisted in the app. 

### Checklist

- [x] Reviewers added


zsmb13 (commented):
What if we performed the saving of the user to the repository in the `onSuccess` callback instead? Might be simpler than adding/removing again.

zsmb13 (approved):
Works well, thanks for fixing this!

PR #1163, opened 2021-01-08
Channel List API updates

zsmb13:
### Description

Brings the API of the Channel List closer to the API of the Message List, injecting dependencies such as the Listener Container and the style information the same way.

Also fixes some random issues around Message List, visibility and accessibility of properties, the custom DateFormatter not being applied in the new implementation previously, etc.

See the commit messages for some details on the changes.

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelListItemViewHolderFactory.kt
```diff
@@ -1,58 +1,37 @@
...
 
 import android.view.ViewGroup
-import io.getstream.chat.android.ui.channel.list.ChannelListView
 import io.getstream.chat.android.ui.channel.list.ChannelListViewStyle
 import io.getstream.chat.android.ui.channel.list.adapter.ChannelItemType
 
 public open class ChannelListItemViewHolderFactory {
 
+    public lateinit var listenerContainer: ChannelListListenerContainer
+        internal set
```
zsmb13, line 10: Ideally, these should be `protected` with `internal` setters, but unfortunately the setter can't be more public than the property itself.

leandroBorgesFerreira (approved):

PR #1165, opened 2021-01-08
[CAS-517] Set transparent navigation bar

JcMinarro:
### Description
Set transparent navigation bar

| Before | After |
| --- | --- |
| device-2021-01-08-113008 | device-2021-01-08-111122 |

### Checklist
- [x] Reviewers added


stream-chat-android-ui-components-sample/src/main/res/values-v29/styles.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <style name="AppTheme" parent="Theme.MaterialComponents.DayNight.NoActionBar">
+        <item name="colorPrimary">@color/stream_color_primary</item>
+        <item name="colorPrimaryDark">@color/stream_color_primary_dark</item>
+        <item name="colorAccent">@color/stream_color_accent</item>
+        <item name="android:windowLightStatusBar">?attr/isLightTheme</item>
+        <item name="android:statusBarColor">@android:color/transparent</item>
+        <item name="android:navigationBarColor">@android:color/transparent</item>
```
zsmb13, line 10: Isn't this attribute available on all API levels we support? It's the only one being added for this configuration, but I'm not sure why we only want it from API 29.

leandroBorgesFerreira: navigationBar would change the old navigation bar (triangle, ball and square). I don't know how it would look like in the old vesion, that's something worth checking.

zsmb13: API 29 doesn't necessarily mean that you have the new gesture navigation bar that's on the PRs in the screenshot though.

leandroBorgesFerreira: Indeed. So we really need to see how the old navigation bar looks like with those changes.

JcMinarro: It is how it is shown on "three icons navigation". We are going to add it to API 23+
device-2021-01-08-131112

zsmb13 (approved):

PR #1168, opened 2021-01-08
Bold mentions ignoring case

JcMinarro:
### Description
Bold mentions ignoring case

| Before | After |
| --- | --- |
| photo_2021-01-08_13-07-24 | photo_2021-01-08_13-07-26 |

### Checklist
- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messagepreview/MessagePreviewView.kt
```diff
@@ -78,7 +78,7 @@ public class MessagePreviewView : FrameLayout {
                 )
             )
         } else {
-            message.user.name.bold()
+            message.user.name.bold(ignoreCase = true)
```
zsmb13, line 81: I don't think we need a change on this line, we're just making this entire string bold, there's no matching involved

zsmb13 (approved):

PR #1178, opened 2021-01-11
Fix/message cid

JcMinarro:
### Description
Enrich messages with cid when searching for them, that is needed to navigate to MessageList

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/4047514/104181506-3677de00-540f-11eb-8b36-ddb0852ae4cf.mp4" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/4047514/104181494-2eb83980-540f-11eb-82dd-6d26679c788f.mp4" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### Checklist
- ~[ ] Changelog updated with client-facing changes
- ~[ ] New code is covered by unit tests
- [X] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
@@ -182,7 +182,10 @@ internal class ChatApi(
         return retrofitApi.searchMessages(apiKey, connectionId, request)
             .map {
                 it.results.map { resp ->
-                    resp.message
+                    resp.message.apply {
+                        (cid.takeUnless { it.isBlank() } ?: channelInfo?.cid)
+                            ?.let { cidResult -> this.enrichWithCid(cidResult) }
+                    }
                 }
             }
     }
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
@@ -33,6 +33,7 @@ import io.getstream.chat.android.client.call.CoroutineCall
 import io.getstream.chat.android.client.call.map
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.events.ChatEvent
+import io.getstream.chat.android.client.extensions.enrichWithCid
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.client.models.Device
 import io.getstream.chat.android.client.models.Flag
@@ -374,24 +375,24 @@ internal class ChatApi(
             return noConnectionIdError()
         }
 
-        if (channelId.isEmpty()) {
-            return retrofitApi.queryChannel(
+        return if (channelId.isEmpty()) {
+            retrofitApi.queryChannel(
                 channelType,
                 apiKey,
                 userId,
                 connectionId,
                 query
-            ).map { flattenChannel(it) }
+            )
         } else {
-            return retrofitApi.queryChannel(
+            retrofitApi.queryChannel(
                 channelType,
                 channelId,
                 apiKey,
                 userId,
                 connectionId,
                 query
-            ).map { flattenChannel(it) }
-        }
+            )
+        }.map { flattenChannel(it) }
     }
 
     fun updateChannel(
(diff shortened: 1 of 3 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -181,7 +182,10 @@ internal class ChatApi(
         return retrofitApi.searchMessages(apiKey, connectionId, request)
             .map {
                 it.results.map { resp ->
-                    resp.message
+                    resp.message.apply {
+                        (cid.takeUnless { it.isBlank() } ?: channelInfo?.cid)
+                            ?.let { cidResult -> this.enrichWithCid(cidResult) }
```
ogkuzmin, line 187: Nit:
```suggestion
                        (cid.takeUnless(CharSequence::isBlank) ?: channelInfo?.cid)
                            ?.let(::enrichWithCid)
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -374,24 +378,24 @@ internal class ChatApi(
...
                 channelType,
                 channelId,
                 apiKey,
                 userId,
                 connectionId,
                 query
-            ).map { flattenChannel(it) }
-        }
+            )
+        }.map { flattenChannel(it) }
```
ogkuzmin, line 398: Nit:
```suggestion
        }.map(::flattenChannel)
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/MessageExtensions.kt
```diff
@@ -2,7 +2,7 @@ package io.getstream.chat.android.client.extensions
 
 import io.getstream.chat.android.client.models.Message
 
-public fun Message.enrichWithCid(cid: String) {
+public fun Message.enrichWithCid(cid: String): Message = apply {
```
ogkuzmin, line 5: Good change. Let's produce always new instead of mutating old one 👍

ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -181,7 +182,10 @@ internal class ChatApi(
         return retrofitApi.searchMessages(apiKey, connectionId, request)
             .map {
                 it.results.map { resp ->
-                    resp.message
+                    resp.message.apply {
+                        (cid.takeUnless { it.isBlank() } ?: channelInfo?.cid)
```
zsmb13, line 186: `cid.takeUnless` seems dangerous here. If there isn't a `cid` in the response, this will be `null` instead of being an empty string (thanks to Gson).

zsmb13: Oh wait, is this circumvented by `CustomObjectGsonAdapter`'s `null` restoration?

JcMinarro: If it is not present it will be empty string, because it is initialized by constructor by default parameters.
Gson only set null if literal null is in the json `{ cid: null }`, something that we are not going to handle

ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
zsmb13: The default parameters in the ctor don't take effect if the instance is created by Gson, because it's created by a `newInstance` call, and the ctor is never executed. But I believe our custom parser changes the `null` value to an empty string, according to [this impl](https://github.com/GetStream/stream-chat-android/blob/8870fb4d77a63e1c15592c068795af1cff93614d/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser/adapters/CustomObjectGsonAdapter.kt#L112-L138) and [this test](https://github.com/GetStream/stream-chat-android/blob/8870fb4d77a63e1c15592c068795af1cff93614d/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/CustomObjectParsingTests.kt#L33-L52).

PR #1181, opened 2021-01-11
MessageListView ViewHolder/Decorator cleanup

zsmb13:
### Description

1. **Use the real `MessageListView` class in the Component Browser for displaying messages**
    This avoids having to expose APIs just for the sake of the Browser, and gives us more of a real look at how those messages are rendered (instead of them being in a custom `RecyclerView`). It also removes a couple classes such as adapters from the Component Browser code.

2. **Make Decorators and ViewHolder implementations internal**
    Our decorators and built-in ViewHolders are very tightly coupled together, creating a new Decorator or DecoratorProvider would be extremely difficult and error prone for clients. With these restrictions, customization of `MessageListView` is restricted to providing a custom factory for its `ViewHolder`s, where the methods creating each types of ViewHolders can be overriden.

3. **Fix some generic type parameters around message ViewHolders**
    We were using a lot of star projections for no reason around our ViewHolder APIs, now most of these are replaced by more concrete types that are nicer to work with.

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/BaseMessageItemViewHolder.kt
```diff
@@ -3,27 +3,22 @@ package io.getstream.chat.android.ui.messages.adapter
 import android.view.View
 import androidx.recyclerview.widget.RecyclerView
 import com.getstream.sdk.chat.adapter.MessageListItem
-import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.Decorator
 
 public abstract class BaseMessageItemViewHolder<T : MessageListItem>(
     itemView: View,
-    private val decorators: List<Decorator>
```
zsmb13, line 10: Decorator related functionality was moved to another, `internal` base class, `DecoratedBaseMessageItemViewHolder`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/MessageListItemViewHolderFactory.kt
```diff
@@ -15,87 +25,89 @@ import io.getstream.chat.android.ui.messages.adapter.viewholder.PlainTextWithMed
 import io.getstream.chat.android.ui.messages.adapter.viewholder.ThreadSeparatorViewHolder
 import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.DecoratorProvider
 
-public open class MessageListItemViewHolderFactory(
-    private val decoratorProvider: DecoratorProvider
-) {
+public open class MessageListItemViewHolderFactory {
+
+    internal lateinit var decoratorProvider: DecoratorProvider
```
zsmb13, line 30: Injected the same way as listenerContainer now, so that it isn't a required constructor parameter (which clients couldn't provide if they extended this class)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -756,6 +759,16 @@ public class MessageListView : ConstraintLayout {
             messageClickListener ?: DEFAULT_MESSAGE_CLICK_LISTENER
     }
 
+    /**
+     * Sets the message long click listener to be used by MessageListView.
+     *
+     * @param messageLongClickListener The listener to use. If null, the default will be used instead.
+     */
+    public fun setMessageLongClickListener(messageLongClickListener: MessageLongClickListener?) {
```
zsmb13, line 767: This listener setter was simply missing from `MessageListView`

adasiewiczr (approved):
Good job!

PR #1182, opened 2021-01-11
[CAS-528] Attachment gallery full screen mode

leandroBorgesFerreira:
[CAS-528](https://stream-io.atlassian.net/browse/CAS-528)

### Description

This PR adds a full screen mode to the attachment gallery. 

giphy

### Checklist

- [x] I have signed the [Stream CLA](required)
- [x] Reviewers added

giphy (1)
_Now you can see your pics in the big screen_


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/ImageSlidePageFragment.kt
```diff
@@ -14,12 +14,16 @@ public class ImageSlidePageFragment : Fragment() {
 
     private lateinit var binding: StreamUiItemImageGalleryBinding
 
+    private var imageClickListener: (() -> Unit)? = {}
```
zsmb13, line 17: Since this is nullable, it could be `null` by default.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -63,8 +65,19 @@ public class AttachmentGallery : ConstraintLayout {
...
+        currentIndex: Int = 0,
+        imageClickListener: () -> Unit = {},
+    ) {
+        adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList) {
+            binding.bottomBarGroup.isVisible = isFullScreen
+
+            isFullScreen = !isFullScreen
+
+            imageClickListener()
+        }
```
zsmb13, line 80: The last lambda parameter is not very obvious here at the call site, it's hard to tell when it will be invoked. Being more verbose would be useful here, with something like:

```kotlin
AttachmentSlidePagerAdapter(
    fragmentActivity,
    imageList,
    imageClickListener = {
       binding.bottomBarGroup.isVisible = isFullScreen
       isFullScreen = !isFullScreen
       imageClickListener()
    }
)
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGalleryActivity.kt
```diff
@@ -28,7 +31,11 @@ public class AttachmentGalleryActivity : AppCompatActivity() {
         super.onStart()
         val urls = intent.getStringArrayExtra(EXTRA_KEY_URLS)?.toList().orEmpty()
         val currentIndex = intent.getIntExtra(EXTRA_KEY_CURRENT_INDEX, 0)
-        binding.attachmentGallery.provideImageList(this, urls, currentIndex)
+        binding.attachmentGallery.provideImageList(this, urls, currentIndex) {
+            binding.toolBarGroup.isVisible = isFullScreen
+
+            isFullScreen = !isFullScreen
+        }
```
zsmb13, line 38: Same for this lambda parameter, at first I thought this was something like `ListAdapter#submitList`, a callback after the new data has been set

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -63,8 +65,19 @@ public class AttachmentGallery : ConstraintLayout {
...
-    public fun provideImageList(fragmentActivity: FragmentActivity, imageList: List<String>, currentIndex: Int = 0) {
-        adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList)
+    public fun provideImageList(
+        fragmentActivity: FragmentActivity,
+        imageList: List<String>,
+        currentIndex: Int = 0,
+        imageClickListener: () -> Unit = {},
+    ) {
+        adapter = AttachmentSlidePagerAdapter(fragmentActivity, imageList) {
+            binding.bottomBarGroup.isVisible = isFullScreen
```
zsmb13, line 75: This line doesn't seem to make sense - the bottom group is visible when we're _not_ full screen, isn't it? I think we should flip the flag first, and _then_ update this visibility, e.g.

```kotlin
isFullScreen = !isFullScreen
binding.bottomBarGroup.isVisible = !isFullScreen
```

zsmb13 (commented):
Other than the code comments, I noticed two visual things while testing this code:

1. The background is white in light mode, not sure if design intended the fullscreen background to be always black or not



2. In full screen mode, there is a bar of "margin" on the bottom below the image, even if you zoom into it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/ImageSlidePageFragment.kt
JcMinarro: Instead of that, doesn't allow null here, then we don't need to handle nullability

zsmb13: We're clearing it in `onDestroyView` at the moment, that's why it's nullable - perhaps we don't need that though?

leandroBorgesFerreira: Removed the nullable part.

leandroBorgesFerreira:
Now it doesn't have the margin in the bottom and the background is now black in full screen

zsmb13 (approved):
LGTM

PR #1183, opened 2021-01-11
fixing locale compatibility problem

leandroBorgesFerreira:
### Description

`resources.configuration.locales` is only available prior to 24 API, we support 21+

### Checklist

- Changelog updated with client-facing changes
- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGalleryActivity.kt
```diff
@@ -44,7 +45,7 @@ public class AttachmentGalleryActivity : AppCompatActivity() {
             DateUtils.FORMAT_ABBREV_RELATIVE
         )
             .toString()
-            .decapitalize(resources.configuration.locales.get(0))
+            .decapitalize(Locale.getDefault())
```
zsmb13, line 48: This would also be equivalent, as it uses the default Locale internally:

```suggestion
            .decapitalize()
```

JcMinarro (approved):

PR #1184, opened 2021-01-11
CAS-520 Fix mention item position in the conversation

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-520

### Description

This PR ensures that we are taken to the right position in the conversation when we navigate to it via mention.

What's not covered:
- Two-way pagination from a message with mention (actually, one-way pagination doesn't work as well: if chat contains for example 50 new message, only latest 30 messages will be displayed plus the messages that were cached previously)
- Mentions in threads (not feasible with our current API)

### Checklist

- [X] Reviewers added

https://user-images.githubusercontent.com/9600921/104194604-96cc4700-5432-11eb-9043-f53320f20231.mov


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -492,11 +495,19 @@ public class MessageListView : ConstraintLayout {
...
-        val targetListItem = adapter.currentList.firstOrNull { it is MessageItem && it.message.id == message.id }
-        targetListItem?.let {
-            val position = adapter.currentList.indexOf(it)
-            binding.chatMessagesRV.layoutManager?.scrollToPosition(position)
-        }
+        Handler().postDelayed(
+            {
+                adapter.currentList
+                    .indexOfFirst { it is MessageItem && it.message.id == message.id }
+                    ?.let {
```
zsmb13, line 502: The safe call seems unnecessary, `indexOfFirst` can not be `null`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -492,11 +495,19 @@ public class MessageListView : ConstraintLayout {
     }
 
     public fun scrollToMessage(message: Message) {
-        val targetListItem = adapter.currentList.firstOrNull { it is MessageItem && it.message.id == message.id }
-        targetListItem?.let {
-            val position = adapter.currentList.indexOf(it)
-            binding.chatMessagesRV.layoutManager?.scrollToPosition(position)
-        }
+        Handler().postDelayed(
```
zsmb13, line 498: This could just be `postDelayed`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -403,34 +403,42 @@ internal class ChannelControllerImpl(
 
     private fun loadMoreMessagesRequest(
         limit: Int = 30,
-        direction: Pagination
+        direction: Pagination,
+        messageId: String? = null,
```
zsmb13, line 407: Took me a moment what this parameter meant, would be a good one to document briefly (can be used to specify the message we want to load relative to, otherwise the already loaded messages will be used in the query - something like that)

Same kind of doc could be added to the two methods calling this method, as they also have this slightly obscure parameter

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -403,34 +403,42 @@ internal class ChannelControllerImpl(
...
     ): QueryChannelPaginationRequest {
+        return QueryChannelPaginationRequest(limit).apply {
+            val messageId = messageId ?: getLoadMoreBaseMessageId(direction)
+            messageId?.let {
+                messageFilterDirection = direction
+                messageFilterValue = it
+            }
```
zsmb13, line 414: Is there a chance that `messageId` is `null` here? Does it make sense to make a request without these fields set?

zsmb13 (approved):

PR #1186, opened 2021-01-11
[CAS-529] video thumbnails fix

leandroBorgesFerreira:
[CAS-529](https://stream-io.atlassian.net/browse/CAS-529)

### Description

Using the correct fetcher so the video thumbnail appears. 

Screenshot_20210111-161542

### Checklist

- [x] Reviewers added


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/ImageLoader.kt
```diff
@@ -122,15 +122,15 @@ public object ImageLoader {
...
+            placeholderResId?.let(::placeholder)
             listener(
                 onStart = { onStart() },
                 onCancel = { onComplete() },
                 onError = { _, _ -> onComplete() },
                 onSuccess = { _, _ -> onComplete() },
             )
             if (videoContentUri) {
-                fetcher(VideoFrameUriFetcher(context))
+                fetcher(VideoFrameFileFetcher(context))
```
bychkovdmitry, line 133: I think we need both fetchers, otherwise thumbnails of videos from galley tab/files tab/file manager will not be displayed

bychkovdmitry (changes requested):

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/ImageLoader.kt
zsmb13: Looking at the implementation, I think the last `fetcher` call will be the only one that takes effect this way, because we're just setting the value of a single property inside Coil with this.

bychkovdmitry: That's bad, then we need two separate methods with different fetchers and more boilerplate

zsmb13: Let's add them, and afterwards I'll try to review all of the stuff we have in `ImageLoader` because it's getting out of hand 😄

JcMinarro: Our `CaptureMediaContract` can return an Uri instead of a File, then it could be rendered with the old `VideoFrameUriFetcher`.
We shouldn't need to add an extra method to render them.
Or we can create an Uri from file to be rendered when we capture video/picutre, if we still need the File itself to be sent to backend

JcMinarro: To create a proper Uri to our SDK you need to use `StreamFileProvider.getUriForFile(context, file)`

leandroBorgesFerreira: i'll take a look

leandroBorgesFerreira: Not this is only using `StreamFileProvider.getUriForFile(context, file)` the URI works as it should

zsmb13 (approved):
Looks good to me

PR #1190, opened 2021-01-12
Cas 526/fix giphy

ogkuzmin:
[CAS-526](https://stream-io.atlassian.net/browse/CAS-526)
### Description

Fix design feedback notes for giphy.
1) Fix shadows
2) Fix bottom footnote padding
3) Show for giphy sent time not created.
4) Add mathematical ceiling when converting dp to px

 https://user-images.githubusercontent.com/11807573/104294739-9edac400-54d0-11eb-8283-8893d4c93ce2.mov  

https://user-images.githubusercontent.com/11807573/104294781-ab5f1c80-54d0-11eb-9337-e7294193689b.mov

### Checklist

- [] <del> Changelog updated with client-facing changes </del>
- [ ] <del> New code is covered by unit tests </del>
- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Int.kt
```diff
@@ -1,6 +1,7 @@
 package io.getstream.chat.android.ui.utils.extensions
 
 import android.content.res.Resources
+import kotlin.math.ceil
 
-internal fun Int.dpToPx(): Int = dpToPxPrecise().toInt()
+internal fun Int.dpToPx(): Int = ceil(dpToPxPrecise()).toInt()
```
zsmb13, line 6: Do we _always_ want to round up for some reason, or do we just want to make sure that things like `0.99f` don't get converted to `0` int value?

For the latter, we could use [`roundToInt`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.math/round-to-int.html) which does proper nearest-integer rounding.

samiuelson (dismissed):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/MessageFooterDecorator.kt
```diff
@@ -92,15 +95,18 @@ internal class MessageFooterDecorator(
...
+
         when {
             data.isNotBottomPosition() || createdAt == null -> textView.isVisible = false
-            else -> {
-                textView.apply {
-                    isVisible = true
-                    text = dateFormatter.formatTime(createdAt)
-                }
-            }
+            data.message.isGiphyNotEphemeral() && updatedAt != null -> textView.showTime(updatedAt)
```
zsmb13, line 108: If it's not ephemeral, but `updatedAt == null`, I think we're just not doing anything with the view, so it'll remain in the previously bound state. I think we should handle that case, at least hide the view or something.

zsmb13 (commented):
Looks much nicer, well done! Got just two small thoughts about the details.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/MessageFooterDecorator.kt
ogkuzmin: For your described case old logic shows `createdAt` time. I think it's better than nothing. Also not clear why for regular giphy message updatedAt is null

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Int.kt
ogkuzmin: Yes, we always want, because the old approach will convert 0.99  to 0 and 2.65 to 2. Thanks for `roundToInt` is better than converting to int manually

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/MessageFooterDecorator.kt
zsmb13: You're right, that works 👍

zsmb13 (approved):

PR #1193, opened 2021-01-12
Task/improve links clicks

adasiewiczr:
https://user-images.githubusercontent.com/17440581/104323622-af963480-54e6-11eb-8fbf-52196b886f65.mp4

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/MediaAttachmentView.kt
```diff
@@ -44,7 +44,11 @@ internal class MediaAttachmentView : ConstraintLayout {
         defStyleRes
     )
 
-    fun showAttachment(attachment: Attachment, andMoreCount: Int = NO_MORE_COUNT) {
+    fun showAttachment(
+        attachment: Attachment,
+        andMoreCount: Int = NO_MORE_COUNT,
+        shouldSetClickListeners: Boolean = true,
```
adasiewiczr, line 50: Not the best solution - we shouldn't set those listeners if not needed but I didn't want to refactor this in the scope of this PR

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/LinkAttachmentView.kt
```diff
@@ -35,6 +38,22 @@ internal class LinkAttachmentView : FrameLayout {
...
+    fun setLinkPreviewClickListener(longClickTarget: View, linkPreviewClickListener: LinkPreviewClickListener) {
+        setOnClickListener {
+            previewUrl?.let { url ->
+                linkPreviewClickListener.onLinkPreviewClick(url)
+            }
+        }
+        setOnLongClickListener {
+            longClickTarget.performLongClick()
+            true
+        }
```
zsmb13, line 53: Leaving this unset or having it set but returning `false` doesn't work, right? Would be nice to get rid of the `longClickTarget` solution, at least here 😄

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/LongClickFriendlyLinkMovementMethod.kt
```diff
@@ -0,0 +1,53 @@
...
+ * forwarded to [longClickTarget].
+ *
+ * This class is reattached to the target [textView] every time its content changes,
+ * so you only need to call [set] one time.
+ */
+internal class LongClickFriendlyLinkMovementMethod private constructor(
+    private val textView: TextView,
+    private val longClickTarget: View,
+    private val onLinkClicked: (url: String) -> Unit,
+) : Utils.TextViewLinkHandler() {
```
ogkuzmin, line 22: First time I see such a framework feature. Looks interesting 👍

ogkuzmin (dismissed):
Interesting solution 👍 
It was new for me with that MovementHandler

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/MediaAttachmentView.kt
zsmb13: Is there a strong reason for using `MediaAttachmentView` inside `LinkAttachmentView`, by the way? Maybe we could replace it with a simpler `ImageView`.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/LinkAttachmentView.kt
adasiewiczr: Correct, the same problem as we had with LongClick inside movement method. I think I'll expose separate method for setting long click target

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/MediaAttachmentView.kt
adasiewiczr: I think it could be replaced easily. Let met check that

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/LongClickFriendlyLinkMovementMethod.kt
adasiewiczr: Yeah, the default `LinkMovementMethod` doesn't handle long clicks correctly so, together with Marton we had to workaround that somehow

zsmb13 (approved):
LGTM

ogkuzmin (approved):

PR #1195, opened 2021-01-12
[CAS-534] Feature/debounce share icon

JcMinarro:
### Description
Share actions needs to generated a file before the intent is generated, and it could spend some time. We need to disable the button during this process to prevent multiple share-process to be executed

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
@@ -20,6 +20,7 @@ import io.getstream.chat.android.ui.databinding.StreamUiAttachmentGalleryBinding
 import io.getstream.chat.android.ui.images.menu.ImagesMenuDialogFragment
 import io.getstream.chat.android.ui.utils.extensions.getFragmentManager
 import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 
 public class AttachmentGallery : ConstraintLayout {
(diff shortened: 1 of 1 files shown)


bychkovdmitry (dismissed):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -101,9 +101,11 @@ public class AttachmentGallery : ConstraintLayout {
         this.imageList = imageList
 
         binding.shareButton.setOnClickListener {
+            it.isEnabled = false
             GlobalScope.launch(DispatcherProvider.Main) {
                 ImageLoader.getBitmapUri(context, adapter.getItem(binding.attachmentGallery.currentItem))
                     ?.let(onSharePictureListener)
+                it.isEnabled = true
```
zsmb13, line 108: On my emulator, the coroutine here executes so quickly that it's done and has re-enabled the button by the time the second click event triggers the click listener, so it's still pretty easy to open the share sheet multiple times by spamming clicks. It's kind of weird that when `getBitmapUrl` suspends and lets go of the main thread, the additional click events don't get processed 🤔

One thing that fixes this is to yield the main thread in the coroutine before re-enabling the button, which lets the main thread process and throw away the additional events: 

```suggestion
                yield()
                it.isEnabled = true
```

... but even then, you can still click the share button again between the time when the intent chooser has been launched, the button re-enabled, and when the share sheet actually animates in... So a more sure fix is to delay this for a bit, giving time for the share sheet to appear:

```suggestion
                delay(500)
                it.isEnabled = true
```

zsmb13: (Note that these suggestions, of course, can not be committed as-is via GitHub, because they also require imports 🙈)

zsmb13 (approved):

PR #1198, opened 2021-01-12
[CAS-533] [CAS-537] Feature/update get attachments enpoints

JcMinarro:
### Description
Create a method to obtain only Messages with attachments.
Remove the unneeded model `AttachmentWithDate`
Refactor `getAttachements` methods

### Checklist
- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -450,15 +449,15 @@ public class ChatClient internal constructor(
...
         offset: Int,
         limit: Int
-    ): Call<List<AttachmentWithDate>> =
+    ): Call<List<Attachment>> =
         getAttachments(channelType, channelId, offset, limit, ATTACHMENT_TYPE_FILE)
 
     public fun getImageAttachments(
         channelType: String,
         channelId: String,
         offset: Int,
```
bychkovdmitry, line 458: Maybe worth mentioning in the java doc that this is a message offset and not attachment offset

bychkovdmitry (approved):
Overall good, only `offset` parameter of `getFileAttachments`/`getImageAttachments` is misleading

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
```diff
@@ -421,9 +421,12 @@ public class ChannelClient internal constructor(
         return client.queryMembers(channelType, channelId, offset, limit, filter, sort, members)
     }
 
-    public fun getFileAttachments(offset: Int, limit: Int): Call<List<AttachmentWithDate>> =
+    public fun getFileAttachments(offset: Int, limit: Int): Call<List<Attachment>> =
         client.getFileAttachments(channelType, channelId, offset, limit)
 
-    public fun getImageAttachments(offset: Int, limit: Int): Call<List<AttachmentWithDate>> =
+    public fun getImageAttachments(offset: Int, limit: Int): Call<List<Attachment>> =
```
zsmb13, line 427: These return type changes break public API (though it's very recently public API, so it might not have users yet)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -467,25 +466,21 @@ public class ChatClient internal constructor(
...
         limit: Int,
         type: String
-    ): Call<List<AttachmentWithDate>> {
+    ): Call<List<Attachment>> =
+        getMessagesWithAttachments(channelType, channelId, offset, limit, type).map { messages ->
```
adasiewiczr, line 470: As far as I remember this returns messages with all attachments when at least one of them has the type that we wanted, so if in some channel, we have a message with image and two files, and we want to get only `imageAttachments` this won't work correctly. We will get this message with all attachments and just map them, so our result we will get one image and two files. Shouldn't we filter only the desired attachments type?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -450,15 +449,15 @@ public class ChatClient internal constructor(
...
+    ): Call<List<Attachment>> =
         getAttachments(channelType, channelId, offset, limit, ATTACHMENT_TYPE_FILE)
 
     public fun getImageAttachments(
         channelType: String,
         channelId: String,
         offset: Int,
         limit: Int
-    ): Call<List<AttachmentWithDate>> =
+    ): Call<List<Attachment>> =
```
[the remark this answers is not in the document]
ogkuzmin: @bychkovdmitry it's not supported by our BE yet

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -467,25 +466,21 @@ public class ChatClient internal constructor(
...
+        limit: Int,
+        type: String
+    ): Call<List<Message>> {
         val channelFilter = Filters.`in`("cid", "$channelType:$channelId")
         val messageFilter = Filters.`in`("attachments.type", type)
-
-        return searchMessages(SearchMessagesRequest(offset, limit, channelFilter, messageFilter)).map { messages ->
-            messages.flatMap { message ->
-                if (type == ATTACHMENT_TYPE_IMAGE) {
-                    message.attachments.filter { attachment -> attachment.isImage }
```
zsmb13, line 477: Are we sure this filtering redundant in the previous version of the code? We added it quite recently, in https://github.com/GetStream/stream-chat-android/commit/9b7778d4ebd82a2bea78233163ff2e36dd5f347b. 

@adasiewiczr got any insight on this one?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/shared/ChatInfoSharedAttachmentsViewModel.kt
```diff
@@ -49,38 +51,33 @@ class ChatInfoSharedAttachmentsViewModel(
...
 
     private suspend fun fetchAttachments() {
-        val currentState = state.value!!
-        val result = if (attachmentsType == AttachmentsType.FILES) {
-            channelClient.getFileAttachments(offset = currentState.results.size, limit = QUERY_LIMIT).await()
-        } else {
-            channelClient.getImageAttachments(offset = currentState.results.size, limit = QUERY_LIMIT).await()
-        }
-
+        val result = channelClient.getMessagesWithAttachments(messages.size, limit = QUERY_LIMIT, attachmentsType.requestTypeKey).await()
```
zsmb13, line 54: It would be a good idea to name all parameters explicitly here, just like `limit`, so that we don't break this if we change the API we're calling into

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
zsmb13: I guess this https://github.com/GetStream/stream-chat-android/pull/1198#discussion_r556338320 is the answer 😃

ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
adasiewiczr: exactly :)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
JcMinarro: `AttachmentWithDate` was created recently and it hasn't been announced and this method is not documented on our webpage yet, I don't think it is an issue. BTW, `AttachmentWithDate` wasn't a good name neither a needed model. It was created because some UI requirements on the sample app

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -467,25 +466,21 @@ public class ChatClient internal constructor(
         offset: Int,
         limit: Int,
         type: String
-    ): Call<List<AttachmentWithDate>> {
+    ): Call<List<Attachment>> =
+        getMessagesWithAttachments(channelType, channelId, offset, limit, type).map { messages ->
+            messages.flatMap { message -> message.attachments.filter { it.type == type } }
+        }
+
+    public fun getMessagesWithAttachments(
```
adasiewiczr, line 474: It might be worth it to add KDOC explaining that this will return messages which contain at least one desired type attachment and not necessarily all of them will have a specified type.

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/shared/ChatInfoSharedAttachmentsViewModel.kt
```diff
@@ -133,3 +135,10 @@ class ChatInfoSharedAttachmentsViewModelFactory(
         return ChatInfoSharedAttachmentsViewModel(cid, type) as T
     }
 }
+
+private fun List<Message>.toAttachmentItems(): List<SharedAttachment.AttachmentItem> =
+    flatMap { message ->
+        message.attachments.map {
```
adasiewiczr, line 141: I think we should additionally filter attachments type here

adasiewiczr (approved):

PR #1199, opened 2021-01-13
[CAS-523] Thread reply behaviour

leandroBorgesFerreira:
[CAS-523](https://stream-io.atlassian.net/browse/CAS-523)

### Description

The title and subtitle now changes when the user is in thread mode. 

Screenshot_20210113-120500

### Checklist

- Changelog updated with client-facing changes. **no need.** 
- New code is covered by unit tests. **no need.** 
- [x] Reviewers added

Love sent for the reviewers of this PR:

giphy


stream-chat-android-ui-components/src/main/res/values/strings.xml
```diff
@@ -77,6 +77,9 @@
     <string name="stream_ui_send_label">Send</string>
     <string name="stream_ui_cancel_label">Cancel</string>
     <string name="stream_ui_attachment_downloading_started">Downloading started</string>
+    <string name="stream_ui_title_thread_reply">Thread Reply</string>
+    <string name="stream_ui_subtitle_thread_reply_single_user">with %s</string>
+    <string name="stream_ui_subtitle_thread_reply_many_users">with %d users</string>
```
zsmb13, line 82: Shouldn't the multiple user / channel case look like this? Or is this another case that I didn't find in the design?

image

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
```diff
@@ -108,27 +139,36 @@ class ChatFragment : Fragment() {
...
         // set external suggestion view which is displayed over message list
         binding.messageInputView.setSuggestionListView(binding.suggestionListView)
     }
 
     private fun initMessagesViewModel() {
         messageListViewModel
-            .apply { bindView(binding.messageListView, viewLifecycleOwner) } // TODO replace with new design message list
+            .apply {
+                bindView(binding.messageListView, viewLifecycleOwner)
+            } // TODO replace with new design message list
```
zsmb13, line 171: This TODO seems outdated, we can probably remove it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/header/MessagesHeaderViewBinding.kt
```diff
@@ -33,4 +34,12 @@ public fun ChannelHeaderViewModel.bindView(view: MessagesHeaderView, lifecycle:
         }
     }
     typingUsers.observe(lifecycle, view::showTypingStateLabel)
+
+    activeThread.observe(lifecycle) { message ->
+        if (message != null) {
+            view.setThreadTitle(view.context.getString(R.string.stream_ui_title_thread_reply))
```
zsmb13, line 40: Seems a bit odd that the binding contains the specific string to set. Could we push that into the View itself, and just call `setThreadMode()` or something like that here?

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/ChannelHeaderViewModel.kt
```diff
@@ -53,4 +57,10 @@ public class ChannelHeaderViewModel @JvmOverloads constructor(
             }
         }
     }
+
+    public fun isThreadMode(): Boolean = _activeThread.value != null
```
zsmb13, line 61: This seems unused, let's not provide it if it's not needed

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
```diff
@@ -55,6 +61,31 @@ class ChatFragment : Fragment() {
...
+        } else {
+            String.format(context.getString(R.string.stream_ui_subtitle_thread_reply_many_users), users.size)
+        }
+    }
+
+    private fun threadUsers(messageWrapper: MessageListItemWrapper): List<User> {
+        return messageWrapper.items
+            .filterIsInstance<MessageListItem.MessageItem>()
+            .filter { message -> message.isTheirs }
+            .map { messageItem -> messageItem.message.user }
```
zsmb13, line 88: Won't this result in 3 of the same user being returned if they've added 3 messages to the thread? Feels like something like a `.distinct()` or `toSet()` call is missing here.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/header/MessagesHeaderViewBinding.kt
leandroBorgesFerreira: 🤦  Yep. Translation would be totally wrong this way

leandroBorgesFerreira: But ChatFragment is actually in the sample app... so no problem with translations, ops.

zsmb13: Shouldn't the View still know how to display "Thread reply" on its own though?

leandroBorgesFerreira: If we push this tot the View, then it will have business logic. I believe it is better to simple set the title and the subtitle so the users of the library could decide the logic they should use. What do you think?

leandroBorgesFerreira: Now there's the normal mode and thread mode. The title is configured in the XML

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatViewModel.kt
```diff
@@ -38,6 +39,8 @@ class ChatViewModel(private val cid: String, private val chatDomain: ChatDomain
         }
     }
 
+    internal fun getChannelName(): String? = channelController?.toChannel()?.name
```
zsmb13, line 42: Date flows the wrong way around here - we're pulling it into the Fragment from the ViewModel, instead of the ViewModel exposing it to be observed.

`MessageListViewModel` already exposes the `Channel` that it's displaying messages for, maybe we can observe that?

zsmb13 (commented):
Thanks for the changes, this is in pretty good shape now! Got just one last note.

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatViewModel.kt
leandroBorgesFerreira: Let me add a mediator here then

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
```diff
@@ -55,6 +62,51 @@ class ChatFragment : Fragment() {
...
+            handleSubtitleChange(messageListViewModel.state.value, channel?.name).let(subtitleMediator::setValue)
+        }
+
+        subtitleMediator.observe(viewLifecycleOwner) { subtitle ->
+            subtitle?.let(binding.messagesHeaderView::setThreadSubtitle)
+        }
+    }
+
+    private fun handleSubtitleChange(state: MessageListViewModel.State?, channelName: String?): String? {
+        return if (state is MessageListViewModel.State.Result && state.messageListItem.isThread) {
```
leandroBorgesFerreira, line 85: This won't work... a new Message inside a thread can arrive even thought the user is not inside a thread... I need to change this.

PR #1204, opened 2021-01-13
Color resource fixes

zsmb13:
### Description

This PR removes nearly all color resources that don't correspond to named colours in the design, such as `concrete`, `alabaster`, various percentage alpha variants of black, etc. We now have 26 total colors (down from 83), and 15 dark mode overrides (down from 37).

⚠️  Warning: color resources like `stream_ui_white` and `stream_ui_black` now work like in the design, so they'll actually be the ~inverted colours when the app is in dark mode, and are not literal pure white and black values.

Remaining color resources are grouped:

- Greys, Accents, Specials: categories from the design palette
- Other: not on the palette, but named in the design
- Platform: contains just the `transparent` colour that we need to remove backgrounds sometimes
- Avatar gradients: used by `AvatarBitmapFactory`, not in design

🙈  No screenshots, as the list of visual changes is infinitely long here. Please build and run the app to test the colours. (And please note that this PR isn't meant to fix all colours in the app, though it fixes many dozens of previously incorrect colours usages.)

### Checklist

- [ ] ~Comparison screenshots added for visual changes~ 🚫
- [x] Reviewers added


stream-chat-android-ui-components/src/main/res/layout/stream_ui_item_message_giphy.xml
```diff
@@ -95,7 +95,7 @@
                 app:layout_constraintLeft_toRightOf="@id/giphyTextLabel"
                 app:layout_constraintRight_toRightOf="parent"
                 app:layout_constraintTop_toTopOf="@id/previousButton"
-                app:tint="@color/stream_ui_icon_strong_tint"
+                app:tint="#000000"
```
zsmb13, line 98: These button colors here are hardcoded and non-named in the design as well, and they don't change between light and dark modes

carterhudson (approved):
I made a commit to fix some elements that were using the wrong colors to save you some time.

zsmb13:
@carterhudson I was intentionally not using system color resources anywhere, because they aren't fully reliable on all devices (OEMs can change them 😢 ), see https://twitter.com/mandybess/status/910555678476951552

adasiewiczr (approved):
Good job!

JcMinarro (approved):
Great!!!

PR #1212, opened 2021-01-13
Fixing response error

leandroBorgesFerreira:
### Description

We are handling failed uploads of files the same way of successful one. There was an error with the Result.isSuccess. We also don't use uploadState. I need to investigate if there's a way to get a response from backend with the information if the upload if successful or not. 

Now a failed upload is displayed like this: 
upload error

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -617,6 +617,10 @@ internal class ChannelControllerImpl(
                     val result = uploadAttachment(it, attachmentTransformer)
                     if (result.isSuccess) {
                         attachment = result.data()
+                    } else {
+                        attachment.name = result.data().name
```
ogkuzmin, line 621: Here you can get an exception. Only `isSuccess` check guarantees that data isn't null.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FileAttachmentsView.kt
```diff
@@ -119,6 +119,18 @@ private class FileAttachmentViewHolder(
             fileTypeIcon.loadAttachmentThumb(attachment)
             fileTitle.text = attachment.getDisplayableName()
             fileSize.text = MediaStringUtil.convertFileSizeByteCount(attachment.fileSize.toLong())
+
+            // Todo: Use uploadState when possible
```
ogkuzmin, line 123: Why is commented code here? Do you really need it?

ogkuzmin (changes requested):

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/utils/Result.kt
```diff
@@ -17,7 +17,7 @@ public data class Result<T : Any> @Deprecated(
     public constructor(error: ChatError) : this(null, error)
 
     val isSuccess: Boolean
-        get() = data != null
+        get() = data != null && error == null
```
zsmb13, line 20: This should never be needed. Instead of making this change, let's find a way to eliminate the [last remaining use of the deprecated `Result` constructor](https://github.com/GetStream/stream-chat-android/blob/9e2b1ef20be75ad250ee502bbbe712c8c5f94947/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt#L707). Then we can remove the deprecation, make the constructor private, and we'll be sure that either `data` or `error` are set in all cases.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: Makes sense.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FileAttachmentsView.kt
leandroBorgesFerreira: Because we need to use `uploadState` to display the correct state of the upload. `uploadState` is not sent by the Backend, but I talk with backend engineers and they told me a way to get the state of the upload, so I can uncomment this and program the correct behaviour. I'll change the PR to draft and work on this,

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: Removed the name.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -24,12 +24,19 @@ public data class Attachment(
...
     var text: String? = null,
+    @SerializedName("type")
     var type: String? = null,
+    @SerializedName("image")
     var image: String? = null,
+    @SerializedName("url")
     var url: String? = null,
+    @SerializedName("name")
     var name: String? = null,
+    @SerializedName("fallback")
```
leandroBorgesFerreira, line 39: `@SerializedName ` so proguard doesn't kill us

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -42,7 +49,10 @@ public data class Attachment(
 
     @IgnoreSerialisation
     @IgnoreDeserialisation
-    override var extraData: MutableMap<String, Any> = mutableMapOf()
+    override var extraData: MutableMap<String, Any> = mutableMapOf(),
+
+    @SerializedName("upload_complete")
+    var uploadComplete: Boolean? = null
```
leandroBorgesFerreira, line 55: I'm using `uploadComplete` instead of serializing `uploadState` as a sealed class would require soe work in the Gson library... But I'm open to other options, we can make this a enum and serialise it... or make Gson serialiase a it as sealed class too.  What do you (whoever is reviewing) think?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
zsmb13: We have a [`consumer-proguard-rules.pro`](https://github.com/GetStream/stream-chat-android/blob/master/stream-chat-android-client/consumer-proguard-rules.pro) file, I think that should be enough to take care of this. (Not sure if it's up-to-date, Studio shows some errors in it on `develop` 😬)

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/utils/Result.kt
```diff
@@ -2,24 +2,26 @@ package io.getstream.chat.android.client.utils
 
 import io.getstream.chat.android.client.errors.ChatError
 
-public data class Result<T : Any> @Deprecated(
-    level = DeprecationLevel.WARNING,
-    message = "Use the constructors taking either the result or the error instead."
-) constructor(
```
zsmb13, line 8: It's nicer if we keep this primary ctor, and just make it private, like

```kotlin
public data class Result<T : Any> private constructor(
    private val ...
) { ... }
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -615,8 +615,12 @@ internal class ChannelControllerImpl(
                 var attachment: Attachment = it
                 if (it.upload != null) {
                     val result = uploadAttachment(it, attachmentTransformer)
+                    attachment.uploadComplete = result.isSuccess
```
ogkuzmin, line 618: As I see it's serialized field obtained from BE. Why do you set the value manually here?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
leandroBorgesFerreira: `@SerializedName` can be an extra prevention. If we move the class to another package (outside `io.getstream.chat.android.client.api.models`) we will trigger error related to proguard, if serialized name, we can move classes around. I've seen this exact error once, it took me quite a long time to find out that the problem was related to a model class being moved to a package that wasn't covered by the proguard rules.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: This information is not generated in the backend. I'm sending the information about the completeness of the upload so we can get it from the backend in a latter moment.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/SendMessageWithFilesTest.kt
```diff
@@ -145,40 +145,6 @@ internal class SendMessageWithFilesTest : BaseDomainTest2() {
...
-        val attachments = randomAttachmentsWithFile().toMutableList()
-        val files: List<File> = attachments.map { it.upload!! }
-
-        mockFileUploadsFailure(files)
-
-        for (attachment in attachments) {
-            val error = attachment.upload!!.toChatError()
-            val result = channelControllerImpl.uploadAttachment(attachment = attachment)
-            assertFailure(result)
-            Truth.assertThat(result.data().uploadState).isInstanceOf(Attachment.UploadState.Failed::class.java)
```
leandroBorgesFerreira, line 178: We should not be able to require data from a failed request, this is what caused this bug... I don't think that attachments with errors should have any format at all. The State is handled outside of the function `uploadAttachment`, in the `sendMessage` function.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/SendMessageWithFilesTest.kt
```diff
@@ -145,40 +145,6 @@ internal class SendMessageWithFilesTest : BaseDomainTest2() {
         }
     }
 
-    @Test
-    fun `Upload attachment with error should be configurable`() = testCoroutines.scope.runBlockingTest {
```
leandroBorgesFerreira, line 149: They should not be configurable, because `result.data()` should not be available. In the `sendMessage` the `Attachment` is still available and we can configure it there.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
zsmb13: Alright, I'm not bothered by it too much, it'll be nuked when we manage to remove Gson anyway 😄

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
ogkuzmin: If you want to add some custom fields to model there is `extraData` field. As I understood API convention it's the main reason for `extraData` existing. And your field could be backed field, something like:
```kotlin

class Attachment {
...
var uploadCompleted: Boolean?
 get() =
  if (extraData.contains("uploadCompleted")) 
       extraData.get("uploadCompleted") as Boolead
   else 
       null
  
 set(value: Boolean) = extraData.put("uploadCompleted", value)
}
```
@JcMinarro Could you prove my point?

bychkovdmitry (changes requested):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
leandroBorgesFerreira: That was my first option, but we ask to `extraData` don't be serialised in this class. 

```
@IgnoreSerialisation
    @IgnoreDeserialisation
    override var extraData: MutableMap<String, Any> = mutableMapOf(),
```

I can also change extraData and make it serialisable, I just wonder why was it make this way and maybe it shouldn't be changed. 🤔

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/utils/Result.kt
```diff
@@ -29,4 +26,26 @@ public data class Result<T : Any> @Deprecated(
...
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = data?.hashCode() ?: 0
+        result = 31 * result + (error?.hashCode() ?: 0)
+        result = 31 * result + isSuccess.hashCode()
+        result = 31 * result + isError.hashCode()
+        return result
+    }
```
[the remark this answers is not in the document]
leandroBorgesFerreira: The private constructor would be exposed by the generated copy method. 

Screenshot 2021-01-15 at 16 25 33

We would end up with the same problem as before.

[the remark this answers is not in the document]
zsmb13: Ah, okay, good call 👍  Let's just simplify these a bit, we can remove `isSuccess` and `isError` from the comparison code here as they're derived from the other properties that we're already comparing.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
ogkuzmin: Indeed, it's strange... I saw conversations that other teams definitely uses this to store custom attributes

leandroBorgesFerreira: Yep... overall I think that that using the extraData looks a better solution, instead of creating one more field. I don't see a problem in sending it do backend, I can be useful for future features

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -615,8 +616,12 @@ internal class ChannelControllerImpl(
                 var attachment: Attachment = it
                 if (it.upload != null) {
                     val result = uploadAttachment(it, attachmentTransformer)
+                    attachment.extraData[EXTRA_UPLOAD_COMPLETE] = result.isSuccess
```
ogkuzmin, line 619: What do you think if we expose for our domain this `uploadCompleted` as Boolean extensions field? And define set and get via this map with key `EXTRA_UPLOAD_COMPLETE `

leandroBorgesFerreira: Done

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -615,16 +618,23 @@ internal class ChannelControllerImpl(
...
```
ogkuzmin, line 621: Leo, idea of extension property was using it here as just property, doesn't expose how we use this mutable generic map

ogkuzmin (commented):
Now it's much better! Just the last small comment

ogkuzmin (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -91,6 +93,7 @@ internal class ChannelControllerImpl(
     override val channelId: String,
     val client: ChatClient,
     val domainImpl: ChatDomainImpl,
+    private val retryUtils: RetryUtils = RetryUtils
```
zsmb13, line 96:
```suggestion
    private val retryUtils: RetryUtils = RetryUtils,
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/utils/RetryUtils.kt
```diff
@@ -0,0 +1,44 @@
...
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.client.call.Call
+import io.getstream.chat.android.client.logger.TaggedLogger
+import io.getstream.chat.android.client.utils.Result
+import io.getstream.chat.android.livedata.extensions.isPermanent
+import kotlinx.coroutines.delay
+
+internal object RetryUtils {
+
+    internal suspend fun <T : Any> runAndRetry(
```
zsmb13, line 12: Why do we need to duplicate this from `ChatDomainImpl`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -40,9 +49,8 @@ public data class Attachment(
     @IgnoreDeserialisation
     var uploadState: UploadState? = null,
 
-    @IgnoreSerialisation
-    @IgnoreDeserialisation
```
zsmb13, line 44: Are we sure about this change? We were ignoring this so that Gson doesn't try to read a map called `extraData` (because it's never there), because we're populating and serializing this into the top level object manually.

zsmb13 (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/utils/RetryUtils.kt
leandroBorgesFerreira: I use a mock for the `ChatDomainImpl` in the unit test... when the  `runAndRetry` gets called, it always returns `null`, so it is not possible to unit test it.

ogkuzmin (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/utils/RetryUtils.kt
zsmb13: You can mock `runAndRetry` with something like this if you need to:

```kotlin
runBlocking {
    whenever(it.runAndRetry<Message>(any())) doAnswer {
        (it.arguments[0] as () -> Call<Message>).invoke().execute()
    }
}
```

Also, we have a lot of setup code in that test that creates a client, domain, etc, perhaps it should extend `BaseDomainTest2` which already does all of that?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
leandroBorgesFerreira: I talked with @zsmb13 and the this gets available anyways, we don't need to serialise it

leandroBorgesFerreira: Also tested, and the behaviour it still the same

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/utils/RetryUtils.kt
leandroBorgesFerreira: The mock makes sense. But I would prefer no to use `BaseDomainTest2` it add much more complexity than I would like... hieritance is normally more annoying than helpful.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -42,7 +51,7 @@ public data class Attachment(
 
     @IgnoreSerialisation
     @IgnoreDeserialisation
```
ogkuzmin, line 53: It's kind of black magic how it gets obtained from BE :-(

ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
zsmb13: It definitely is, but can we do anything better? There can be arbitrary dynamic data in the JSON we send/receive.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/utils/RetryUtils.kt
zsmb13: Our unit test setups are pretty complex, so it helps if they at least work the same way. But if you don't want the base test, using things like `TestCoroutineRule` as it's used in the `BaseDomainTest2` might still help simplify your test.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -6,6 +6,8 @@ import io.getstream.chat.android.client.parser.IgnoreDeserialisation
 import io.getstream.chat.android.client.parser.IgnoreSerialisation
 import java.io.File
 
+public const val EXTRA_UPLOAD_COMPLETE: String = "uploadComplete"
```
zsmb13, line 9: This key could probably go into `AttachmentExtensions`, and be private

bychkovdmitry (approved):

zsmb13 (approved):

PR #1215, opened 2021-01-14
ImageLoader refactor

zsmb13:
### Description

Cleans up our internal `ImageLoader` object which used weird member extensions and had lots of different methods to support different scenarios of image loading.

The new `StreamImageLoader` interface that replaces has just three methods to implement (and there's a corresponding `CoilStreamImageLoader` implementation), and will make it easy for us to make image loading swappable as a next step.

Loading images happens via extension functions on `ImageView`, such as `load` and `loadVideoThumbnail`.

⚠️  Some Coil APIs have been hidden from IDE autocompletion so that our own methods are prioritized instead. If you need to call these for some reason (you shouldn't have to), you can still add the imports manually, they just won't be suggested by the IDE.

### Checklist

- [x] Reviewers added


bychkovdmitry (dismissed):

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/images/StreamImageLoader.kt
```diff
@@ -0,0 +1,46 @@
...
+import android.content.Context
+import android.graphics.Bitmap
+import android.net.Uri
+import android.widget.ImageView
+import androidx.annotation.DrawableRes
+import androidx.annotation.Px
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+
+@InternalStreamChatApi
+public interface StreamImageLoader {
```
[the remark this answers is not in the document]
zsmb13: I'd like to keep `StreamImageLoader` an interface. We can modify it for proper pluggable instances later on (for example, if we move our Coil support to a separate module, which shouldn't be hard after this).

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/view/activity/AttachmentActivity.kt
```diff
@@ -0,0 +1,130 @@
...
+            Toast.makeText(this, "Error!", Toast.LENGTH_SHORT).show()
+            return
+        }
+        iv_image.isVisible = true
+        webView.isVisible = false
+
+        iv_image.load(
+            data = urlSigner.signImageUrl(url),
+            placeholderResId = R.drawable.stream_placeholder
+        )
```
zsmb13, line 108: Converted this Activity to Kotlin as well, because I couldn't be bothered to support Java callers for the new image loading APIs

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileProvider.kt
```diff
@@ -20,5 +23,22 @@ public class StreamFileProvider : FileProvider() {
 
         public fun getUriForFile(context: Context, file: File): Uri =
             getUriForFile(context, getFileProviderAuthority(context), file)
+
+        public fun writeImageToSharableFile(context: Context, bitmap: Bitmap): Uri? {
```
zsmb13, line 27: Moved this method here as it's related to file management and doesn't depend on our image loading library

adasiewiczr (approved):

PR #1216, opened 2021-01-15
[CAS-541] Fix scroll to bottom button with deleted messages

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-541

### Description

What's done:
- Scroll to bottom button shows correct number of unseen messages. The root of the problem: our zero height view holders (theirs deleted messages, typing indicator, etc.) can not be detected with `LinearLayoutManager.findLastVisibleItemPosition()` and `LinearLayoutManager.findFirstVisibleItemPosition()`. So when we  reach the bottom of the list , `LinearLayoutManager.findLastVisibleItemPosition()` can be less than `adapter.itemCount - 1`.
- Improved the performance a bit which is crucial as `RecyclerView.OnScrollListener.onScrolled()` method is invoked very frequently
- Scroll to bottom button appears in two cases: 1. when we scrolled a bit so that last message is not visible and new message is received 2. when we scrolled 30 items from the bottom
- Moved the logic related to scrolling to a separate helper class to simplify `MessageListView` class

Notes:
- Mark channel read logic can be significantly improved
- There is a bug: 1. open conversation 2. scroll up a bit 3. wait for 2 new messages to arrive 4. observe scroll to button with 2 unseen messages 5. scroll slightly down so that counter shows 1 unread message 6. navigate to channel list. Expected result: 1 unread message. Actual result: 2 unread messages. I wonder if there is an API endpoint to mark individual message as read or should we do it on our site in offline module?
- !!! If approx half of the last 30 messages are deleted, the list will not be scrollable and we won't be able to load more messages
- !!! Our `EndlessScrollListener` is broken as it relies on `findFirstVisibleItemPosition()`. If there are `loadMoreThreshold` deleted messages on top of our message list, load more listener will never be triggered

The last problem can not be solved by filtering out deleted messages of other users as customers might choose to use their own implementation of `MessageDeletedViewHolder` where deleted messages of other users are not hidden

### Checklist

- [X] Reviewers added

https://user-images.githubusercontent.com/9600921/104695193-6845b900-571d-11eb-8a45-396993fbbe9f.mov

https://user-images.githubusercontent.com/9600921/104695260-814e6a00-571d-11eb-9e0d-62666f2a6e05.mov


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListScrollHelper.kt
```diff
@@ -0,0 +1,155 @@
...
+                override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
+                    if (!scrollToBottomButtonEnabled || currentList.isEmpty()) return
+
+                    val lastVisibleItemPosition = layoutManager.findLastVisibleItemPosition()
+                    val lastPotentiallyVisibleItemPosition = currentList.indexOfLast { it.isValid() }
+                    val bottomOffset = lastPotentiallyVisibleItemPosition - lastVisibleItemPosition
+                    isAtBottom = bottomOffset == 0
+
+                    max(lastVisibleItemPosition, getLastSeenItemPosition())
+                        .coerceIn(0, currentList.lastIndex)
```
ogkuzmin, line 51: Do we need this coerce check? Seems it can't exceed limits

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListScrollHelper.kt
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    internal fun scrollToMessage(message: Message) {
+        recyclerView.postDelayed(
+            {
+                currentList.indexOfFirst { it is MessageListItem.MessageItem && it.message.id == message.id }
+                    .takeIf { it >= 0 }
+                    ?.let {
+                        with(recyclerView) {
+                            layoutManager
```
ogkuzmin, line 83: You already have cast layout manager, why do you not use it?

ogkuzmin (commented):
LGTM, some comments

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListScrollHelper.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    private var lastSeenMessageInChannel: MessageListItem? = null
+    private var lastSeenMessageInThread: MessageListItem? = null
+    private var isAtBottom = false
+    private var unreadCount = 0
+
+    private var currentList: List<MessageListItem> = emptyList()
+        get() {
+            return adapter.currentList
+        }
```
zsmb13, line 34: I agree with the lint warning here - I think this was meant to be a `val`?

image

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -315,9 +297,13 @@ public class MessageListView : ConstraintLayout {
...
     }
 
-    private fun initUnseenMessagesButton() {
-        binding.scrollToBottomButton.setOnClickListener {
-            binding.chatMessagesRV.scrollToPosition(lastPosition())
+    private fun initScrollHelper() {
+        scrollHelper = MessageListScrollHelper(
+            recyclerView = binding.chatMessagesRV,
+            scrollButtonView = binding.scrollToBottomButton,
+            scrollToBottomButtonEnabled = scrollToBottomButtonEnabled,
```
zsmb13, line 304: The `MessageListView` property we're setting here is set in `configureAttributes`, which is invoked later than `initScrollHelper`. I think this will result in us always passing in the default `true` here.

Then again, just changing the order of these is dangerous, because we can easily change it back again later without noticing. Perhaps we should make the `MessageListView` property a `Delegates.notNull` value (as it can't be `lateinit`, being a primitive).

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -638,6 +573,9 @@ public class MessageListView : ConstraintLayout {
 
     public fun setNewMessagesBehaviour(newMessagesBehaviour: NewMessagesBehaviour) {
         this.newMessagesBehaviour = newMessagesBehaviour
+        if (::scrollHelper.isInitialized) {
```
zsmb13, line 576: Is there a chance that it's not? As far I see, we'll always initialize it in the constructor.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListScrollHelper.kt
```diff
@@ -0,0 +1,155 @@
...
+        return currentList
+            .lastOrNull()
+            ?.safeCast<MessageListItem.MessageItem>()
+            ?.isMine
+            ?: false
+    }
+
+    private fun MessageListItem.isValid(): Boolean {
+        return this !is MessageListItem.TypingItem &&
+            !(this is MessageListItem.MessageItem && this.isTheirs && this.message.isDeleted())
```
zsmb13, line 144: Shouldn't we be handling the other item types as well? The logic could be something like

```kotlin
this is MessageListItem.MessageItem && !(this.isTheirs && this.message.isDeleted())
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -654,125 +592,23 @@ public class MessageListView : ConstraintLayout {
...
         adapter.isThread = listItem.isThread
-
-        val oldSize = adapter.itemCount
-
-        adapter.submitList(entities) {
-            continueMessageAdd(startThreadMode, listItem, entities, oldSize)
-        }
-    }
-
-    private fun continueMessageAdd(
```
zsmb13, line 676: So happy to see all of this gone 😄

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListScrollHelper.kt
```diff
@@ -0,0 +1,155 @@
...
+
+    private fun isLastMessageMine(): Boolean {
+        return currentList
+            .lastOrNull()
+            ?.safeCast<MessageListItem.MessageItem>()
+            ?.isMine
+            ?: false
+    }
+
+    private fun MessageListItem.isValid(): Boolean {
```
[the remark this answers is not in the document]
bychkovdmitry: We still need to ignore irrelevant items. I created a ticket https://stream-io.atlassian.net/browse/CAS-544 to filter out invisible items. After it is implemented we will replace this check with `this is MessageItem`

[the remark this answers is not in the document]
bychkovdmitry: @ogkuzmin started progress on CAS-544, so I'm resolving this conversation

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
bychkovdmitry: You mean `scrollToBottomButtonEnabled` property?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListScrollHelper.kt
bychkovdmitry: There are similar checks in the old `MessageListView`. I guess this was done for a reason to cover some edge case. So I'm inclined to leave it as is for now.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
bychkovdmitry: @zsmb13 Can you check the latest commit? I ensured that we crash if the ordering is broken

ogkuzmin (approved):
LGTM, let's wait for Marton

zsmb13 (approved):
LGTM

PR #1218, opened 2021-01-15
[CAS-538] Add support for pinned messages to LLC

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-538

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Message.kt
```diff
@@ -94,6 +94,14 @@ public data class Message(
...
-    var replyMessageId: String? = null
+    var replyMessageId: String? = null,
+
+    var pinned: Boolean = false,
+    @SerializedName("pinned_at")
+    var pinnedAt: Date? = null,
+    @SerializedName("pin_expires")
+    var pinExpires: Date? = null,
+    @SerializedName("pinned_by")
+    var pinnedBy: User? = null,
```
zsmb13, line 105: I assume we should also update `MessageEntity` and `ChannelEntity`?

zsmb13 (commented):
The JS implementation makes a bunch of updates to their channel state as message events (like `message.new`, `message.updated, `message.deleted` are coming in. Don't we have similar updates to perform?

On another note, is there anything here that we can add tests for?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Message.kt
adasiewiczr: As far as I know, for now, we are going to only support this in LLC so I'm not sure if there is a need to update those entities

zsmb13: Ah, okay, I didn't realize the scope of the task. The PR's title is actually clear, the Jira task perhaps not so much.

adasiewiczr:
@zsmb13 I'm not sure about the tests because basically, I've added a bunch of new fields and API calls which should return updated objects

zsmb13 (approved):
As far as covering the LLC, this LGTM 👍

PR #1222, opened 2021-01-15
[CAS-542] Thread participants

ogkuzmin:
[CAS-542](https://stream-io.atlassian.net/browse/CAS-542)

### Description

Show thread participants' avatar in thread reply footnote
Thread participants

### Checklist

- [ ] <del> Changelog updated with client-facing changes </del>
- [ ] <del> New code is covered by unit tests </del>
- [x] Reviewers added


stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_threads_footnote.xml
```diff
@@ -1,39 +1,108 @@
...
+        app:streamUiAvatarBorderWidth="0dp"
+        app:streamUiAvatarOnlineIndicatorEnabled="false"
+        tools:src="@tools:sample/avatars"
+        />
+
+    <io.getstream.chat.android.ui.avatar.AvatarView
+        android:id="@+id/secondMineUserImage"
+        android:layout_width="@dimen/stream_ui_message_thread_reply_avatar_size"
+        android:layout_height="@dimen/stream_ui_message_thread_reply_avatar_size"
+        android:layout_marginEnd="@dimen/stream_ui_message_thread_reply_avatar_half_size"
```
zsmb13, line 90: This margin seems unnecessary, it results in the avatars being disconnected from the ornament, like this:

image

vs the behaviour on the other side:

image

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Message.kt
```diff
@@ -104,4 +104,7 @@ public data class Message(
     @SerializedName("pinned_by")
     var pinnedBy: User? = null,
 
+    @SerializedName("thread_participants")
+    var threadParticipants: List<User> = emptyList()
```
zsmb13, line 108: A lot of these newly added lines in the `.kt` files could use trailing commas

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FootnoteView.kt
```diff
@@ -67,6 +69,52 @@ internal class FootnoteView : ConstraintLayout {
...
+
+            root.updateConstraints {
+                clear(threadRepliesButton.id, ConstraintSet.TOP)
+                connect(
+                    threadRepliesButton.id,
+                    ConstraintSet.TOP,
+                    if (isMine) firstMineUserImage.id else firstTheirUserImage.id,
+                    ConstraintSet.TOP
+                )
+            }
```
zsmb13, line 94: Why do we have to update this vertical constraint every time? Couldn't we constrain this View in a fixed way instead?

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_threads_footnote.xml
```diff
@@ -1,39 +1,108 @@
...
         android:layout_width="17dp"
         android:layout_height="40dp"
         android:src="@drawable/stream_ui_thread_replies_ornament_left"
         app:tint="@color/stream_ui_grey_whisper"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintBottom_toBottomOf="@id/bottomLine"
+        tools:visibility="gone"
+        />
+    
+    <io.getstream.chat.android.ui.avatar.AvatarView
```
zsmb13, line 33: In the design, the user avatars are stacked right-to-left here (the left one is on top), and the top one has a white border

image

Are planning to make those adjustments, or is our implementation different on purpose?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FootnoteView.kt
```diff
@@ -67,6 +69,52 @@ internal class FootnoteView : ConstraintLayout {
...
+            secondMineUserImage.isVisible = isMine
+
+            val (first, second) = getTwoLastUsers(threadParticipants)
+
+            applyUser(first, if (isMine) firstMineUserImage else firstTheirUserImage)
+            applyUser(second, if (isMine) secondMineUserImage else secondTheirUserImage)
+        }
+    }
+
+    private fun getTwoLastUsers(threadParticipants: List<User>): Pair<User?, User?> {
```
zsmb13, line 108: Not sure where things go wrong exactly with the data, but when there's a single reply to a thread, we end up showing two avatars, one of them being the sender of the original message:

image

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FootnoteView.kt
ogkuzmin: Thanks, I found out how to achieve it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FootnoteView.kt
ogkuzmin: I think it was done intentionally. So there is description by designers https://www.loom.com/share/9cadf788a5db4d0192b946dd9822d409
As I didn't do anything with data propagated from BE seems it is an implementation of requirements.

zsmb13: Ah, okay, I think this is confusing as a user (looks like more than 1 reply), but apparently it's intentional then

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_threads_footnote.xml
ogkuzmin: Thanks for such nice catch! I didn't notice it. Fixed

zsmb13: The last visual glitch here is that with the border added, the avatar with the border is smaller than the other one, as the border is drawn within the original size of the `AvatarView`:

image

We could try to increase its size here to match what the border clips off, but we should probably revisit the border stroke implementation inside `AvatarView` itself instead. Could go into another PR / Jira task though.

ogkuzmin: Made 18 dp size and 1 dp stroke by default for all avatars

zsmb13 (approved):
Looks great, nice work! 👍

Issue #1226, opened 2021-01-18
Inconsistent and wrong number of channel unread count


bibibilu:
I am using 4.5.0. I am trying to observe unread count for each channel but the channel unread count are not consistent and the sum up for that of all the channel is not equal to the total unread count shown on the bottom tab badge from time to time, as shown in the video. Do you have any ideal to make it work properly? Thanks!

https://user-images.githubusercontent.com/55115952/104872093-bd422480-5901-11eb-9545-e14bc50fab92.mp4

The code used to fetch total unread count 
```
val count = ChatClient.instance().getCurrentUser()?.totalUnreadCount ?: 0
badge.isVisible = count > 0
badge.number = count

ChatDomain.instance().totalUnreadCount.observe(viewLifecycleOwner, Observer {
  badge.isVisible = it > 0
  badge.number = it      
})
```
The code used to fetch channel unread count 
```
ChannelListItemViewHolder holder = new ChannelListItemViewHolder(view) {
      @Override
      public void bind(@NotNull Channel channel, int i, @NotNull ChannelItemPayloadDiff channelItemPayloadDiff) {
        super.bind(channel, i, channelItemPayloadDiff);
        LifecycleOwner lifecycleOwner = (LifecycleOwner) itemView.getContext();
        ChatDomain.instance().getTotalUnreadCount().observe(lifecycleOwner, num -> {
          TextView count = view.findViewById(R.id.channelUnreadCountTextView);
          unreadBadge.setVisibility(channel.getUnreadCount() == 0 ? View.GONE : View.VISIBLE);
          unreadBadge.setVisibility(View.GONE);
          count.setText(String.valueOf(channel.getUnreadCount()));
        });
      }
    };
```

bychkovdmitry:
@bibibilu, thank you for reporting this

As for total unread count, you are doing it correctly here:
```
ChatDomain.instance().totalUnreadCount.observe(viewLifecycleOwner, Observer {
  badge.isVisible = it > 0
  badge.number = it      
})
``` 

What about channel unread count, I see that you are registering lifecycle observer in `bind` method. Obviously you should not be doing that as it creates a memory leak (every time `bind` method is invoked a new observer is registered, but the old one is not disposed). My best guess is that the following code was not working for you:
```
ChannelListItemViewHolder holder = new ChannelListItemViewHolder(view) {
      @Override
[...]

PR #1227, opened 2021-01-18
[CAS-545] Incorrect unread counts in channel list

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-545
https://github.com/GetStream/stream-chat-android/issues/1226

### Description

`ChannelListDiffCallback` was not taking into account `unreadCount` field. As a result, unread counts could be stale in the channel list. 

### Checklist

- [X] Reviewers added

### DIFF
--- a/stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelItemPayloadDiff.kt
+++ b/stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelItemPayloadDiff.kt
@@ -6,6 +6,7 @@ public data class ChannelItemPayloadDiff(
     val lastMessage: Boolean,
     val lastMessageDate: Boolean,
     val readState: Boolean,
+    val unreadCount: Boolean,
 ) {
     public operator fun plus(other: ChannelItemPayloadDiff): ChannelItemPayloadDiff =
         ChannelItemPayloadDiff(
@@ -14,5 +15,6 @@ public data class ChannelItemPayloadDiff(
             lastMessage = lastMessage || other.lastMessage,
             lastMessageDate = lastMessageDate || other.lastMessageDate,
             readState = readState || other.readState,
+            unreadCount = unreadCount || other.unreadCount
         )
 }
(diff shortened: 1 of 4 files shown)


stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelListDiffCallback.kt
```diff
@@ -47,7 +48,8 @@ internal class ChannelListDiffCallback @JvmOverloads constructor(
             name = !channelNameIsTheSame(newChannel, oldChannel),
             avatarView = !channelUsersAreTheSame(newChannel, oldChannel),
             readState = channelUserReadIsDifferent(oldChannel, newChannel),
-            lastMessageDate = !channelLastMessageDatesAreTheSame(oldChannel, newChannel)
+            lastMessageDate = !channelLastMessageDatesAreTheSame(oldChannel, newChannel),
+            unreadCount = unreadCountIsDifferent(oldChannel, newChannel)
```
zsmb13, line 52: Finally, a suggestion you'll be able to apply via GitHub because it doesn't change imports 😄 

```suggestion
            unreadCount = unreadCountIsDifferent(oldChannel, newChannel),
```

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelItemPayloadDiff.kt
```diff
@@ -6,6 +6,7 @@ public data class ChannelItemPayloadDiff(
     val lastMessage: Boolean,
     val lastMessageDate: Boolean,
     val readState: Boolean,
+    val unreadCount: Boolean,
```
zsmb13, line 9: Should we also update [this code](https://github.com/GetStream/stream-chat-android/blob/2006964e2cd5149f2be92ab28d5de358d0c034df/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Channel.kt#L51-L57)?

bychkovdmitry: I think the difference is that in the old diff we only check if current user read changed while in the new diff this case is covered by the fact we calculate the diff for all `ChannelUserRead` objects.

zsmb13 (approved):
LGTM

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/ChannelItemPayloadDiff.kt
bychkovdmitry: However, will recheck it

zsmb13 (approved):

PR #1230, opened 2021-01-19
[CAS-544] Message filter for hidden messages

ogkuzmin:
### Description

For the clear, proper way of the message list item, we need to filter message items that we're not going to show. Previously we hide a view and made 0 height. 
This PR brings the message filter.

### Checklist

- [x] Reviewers added


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StartStopBuffer.kt
```diff
@@ -13,6 +13,7 @@ public class StartStopBuffer<T> {
     private val events: Queue<T> = ConcurrentLinkedQueue()
     private var active = AtomicBoolean(true)
     private var func: ((T) -> Unit)? = null
+    private var dataFilter: (T) -> T = { it }
```
leandroBorgesFerreira, line 16: I think that `dataFilter` fits better inside `MessageLIstView`. A buffer normally doesn't filter data, it only controls when the data is sent

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/HiddenMessageListItemFilter.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    private val theirDeletedMessagePredicate: (MessageListItem) -> Boolean = { item ->
+        item is MessageListItem.MessageItem && item.message.isDeleted() && item.isTheirs
+    }
+
+    private val theirGiphyEphemeralMessagePredicate: (MessageListItem) -> Boolean = { item ->
+        item is MessageListItem.MessageItem && item.message.isGiphyEphemeral() && item.isTheirs
+    }
+
+    override fun filterMessageListItem(messageListItem: List<MessageListItem>): List<MessageListItem> {
```
leandroBorgesFerreira, line 18: You can use `Iterable` instead of List for the parameter.

ogkuzmin: This use case is tightened to MessageListView, so it requires to return a list and receives a list as an argument. Since it's strictly defined use-cases I prefer to keep the List type here

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StartStopBuffer.kt
ogkuzmin: You're right. Reworked solution

leandroBorgesFerreira (dismissed):
Looks good

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -589,20 +596,31 @@ public class MessageListView : ConstraintLayout {
...
-        adapter.submitList(listItem.items) {
-            scrollHelper.onMessageListChanged(
-                isThreadStart = isThreadStart,
-                hasNewMessages = listItem.hasNewMessages,
-                isInitialList = isOldListEmpty && listItem.items.isNotEmpty()
-            )
-            buffer.active()
+    private fun handleNewWrapper(listItem: MessageListItemWrapper) {
+        CoroutineScope(DispatcherProvider.IO).launch {
+            messageListItemFilter.filterMessageListItem(listItem.items).let { filteredList ->
```
zsmb13, line 606: `let` is fancy, but is there any advantage of using it here instead of just `val filteredList = ...`? It seems to increase the nesting and complexity of the code for no gain.

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/HiddenMessageListItemFilter.kt
```diff
@@ -0,0 +1,24 @@
+package io.getstream.chat.ui.sample.feature.chat
+
+import com.getstream.sdk.chat.adapter.MessageListItem
+import io.getstream.chat.android.ui.messages.view.MessageListView
+import io.getstream.chat.android.ui.utils.extensions.isDeleted
+import io.getstream.chat.android.ui.utils.extensions.isGiphyEphemeral
+
+object HiddenMessageListItemFilter : MessageListView.MessageListItemFilter {
```
zsmb13, line 8: Shouldn't this be the default implementation that we ship in the library, instead of being part of the sample app? Feels like it's tied to the behaviour of the ViewHolders that we _do_ ship.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -589,20 +596,31 @@ public class MessageListView : ConstraintLayout {
...
+
+                    val isThreadStart = !adapter.isThread && listItem.isThread
+                    val isOldListEmpty = adapter.currentList.isEmpty()
+
+                    adapter.isThread = listItem.isThread
+                    adapter.submitList(filteredList) {
+                        scrollHelper.onMessageListChanged(
+                            isThreadStart = isThreadStart,
+                            hasNewMessages = listItem.hasNewMessages,
+                            isInitialList = isOldListEmpty && listItem.items.isNotEmpty()
```
zsmb13, line 618: We're still using `listItem.items` instead of the filtered list here - is that intentional?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Message.kt
```diff
@@ -17,36 +17,36 @@ internal fun Message.getSenderDisplayName(context: Context, isDirectMessaging: B
 private fun getSelfDisplayName(context: Context) =
     context.getString(R.string.stream_ui_channel_display_name_self)
 
-internal fun Message.isDeleted(): Boolean = deletedAt != null
+public fun Message.isDeleted(): Boolean = deletedAt != null
```
zsmb13, line 20: I'd leave these as internal API if possible, clients can always access the underlying data themselves if they really want to, and we'd have less API to worry about

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/HiddenMessageListItemFilter.kt
ogkuzmin: I'm not sure. For example in the old sample app, we showed their deleted messages intentionally.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Message.kt
ogkuzmin: From my point of view, these extensions look simpler and clearer than raw data inside models. There are no hidden implementation details, I think is entirely ok to make them public

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
ogkuzmin: Good point

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
ogkuzmin: Unintentional) Thanks for the catch!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/Message.kt
zsmb13: Alright, can we please rename the last three of them though? 😄 

They have really awkward names for public API, could be `hasSingleReaction`, `isReply`, and `hasText` instead.

ogkuzmin: No problem)

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/HiddenMessageListItemFilter.kt
zsmb13: This is a separate `MessageListView` class, so it wouldn't impact the old sample.

ogkuzmin: Hmm... Makes sense. Will fix it

zsmb13 (approved):
LGTM

PR #1231, opened 2021-01-19
Attachment options menu UI

samiuelson:
### Description

This is the 1st iteration of the attachment options menu feature. This PR adds:
1. UI for the options menu:
Screenshot 2021-01-19 at 12 48 10
2. Pass message data into `AttachmentGalleryActvity`

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/chat/navigation/GalleryImageAttachmentDestination.kt
```diff
@@ -25,10 +25,9 @@ public class GalleryImageAttachmentDestination(
             return
         }
 
-        val userName = message.user.name
         val createdAt = message.getCreatedAtOrThrow().time
         val attachmentIndex = message.attachments.indexOf(attachment)
 
-        start(AttachmentGalleryActivity.createIntent(context, userName, createdAt, attachmentIndex, imageUrls))
+        start(AttachmentGalleryActivity.createIntent(context, createdAt, attachmentIndex, message))
```
ogkuzmin, line 31: I see that previously we did filtering to ensure that we're going to show only images. I think we have to pass to `createIntent` also filtered list of attachments

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
```diff
@@ -50,6 +50,11 @@ public class AttachmentGallery : ConstraintLayout {
 
     private var isFullScreen = false
 
+    /**
+     * Informs about index of the currently visible attachment
+     */
+    public val currentItemIndex: Int by lazy { binding.attachmentGallery.currentItem }
```
ogkuzmin, line 56: Why is it lazy property instead of being just wrapped function like:
``` kotlin
public val currentItemIndex: Int
  get() = binding.attachmentGallery.currentItem
```
?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGalleryActivity.kt
```diff
@@ -66,24 +94,53 @@ public class AttachmentGalleryActivity : AppCompatActivity() {
...
+                cid = cid,
+                userName = userName,
+                imageUrl = this.imageUrl,
+                assetUrl = this.assetUrl,
+                name = this.name
+            )
     }
+
+    @Parcelize
+    public data class AttachmentData(
```
ogkuzmin, line 130: Good approach instead of serializing whole Message 👍

ogkuzmin (commented):
Few comments

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/images/AttachmentGallery.kt
samiuelson: Yep, getter is better. Fixed.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/chat/navigation/GalleryImageAttachmentDestination.kt
```diff
@@ -25,10 +25,9 @@ public class GalleryImageAttachmentDestination(
             return
         }
 
-        val userName = message.user.name
         val createdAt = message.getCreatedAtOrThrow().time
         val attachmentIndex = message.attachments.indexOf(attachment)
 
-        start(AttachmentGalleryActivity.createIntent(context, userName, createdAt, attachmentIndex, imageUrls))
+        start(AttachmentGalleryActivity.createIntent(context, createdAt, attachmentIndex, message, attachments))
```
ogkuzmin, line 31: @samiuelson "Filtering" is what we do at line 20 I mean. Seems we don't need `imageUrls` anymore, we have to filter attachments instead and check that the filtered collection isn't null. And then pass it to an activity

samiuelson: Right, fixed.

ogkuzmin (approved):
LGTM

PR #1234, opened 2021-01-19
CAS-545 Fix channel diff and unread count

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-545
https://github.com/GetStream/stream-chat-android/issues/1226

### Description

This PR addresses 2 problems:
1) Dropped channel list diff if channels are updated too frequently. As a result, we didn't always have up-to-date content in the channel list.
2) Unread count in `ChannelControllerImpl` was not updated synchronously when channel list was updated. As a consequence, if no other events were coming after `NewMessageEvent`, we were observing stale unread count.

### Checklist

- [X] Reviewers added


carterhudson (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/utils/ComputeUnreadCount.kt
```diff
@@ -8,15 +8,10 @@ import io.getstream.chat.android.client.models.User
...
-    messages: List<Message>? = null
-): Int? {
+    messages: List<Message> = emptyList(),
+): Int {
     val logger = ChatLogger.get("computeUnreadCount")
 
-    if (messages == null) {
-        logger.logD("Messages null; returning null")
-        return null
-    }
```
zsmb13, line 18: Wouldn't it be useful to do an early return here if we got an empty list?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/ChannelListItemAdapter.kt
```diff
@@ -74,19 +18,29 @@ internal class ChannelListItemAdapter(
...
-        bind(position, holder, payloads.firstOrDefault(EVERYTHING_CHANGED).cast())
+        val diff = (
+            payloads
+                .filterIsInstance<ChannelListPayloadDiff>()
+                .takeIf { it.isNotEmpty() }
+                ?: listOf(FULL_CHANNEL_LIST_ITEM_PAYLOAD_DIFF)
+            )
+            .fold(EMPTY_CHANNEL_LIST_ITEM_PAYLOAD_DIFF) { acc, channelListItemPayloadDiff ->
+                acc + channelListItemPayloadDiff
+            }
```
zsmb13, line 34: Awesome example of fancy, but not overdone Kotlin code 👍

zsmb13 (approved):

PR #1235, opened 2021-01-19
[CAS-523] thread title change

leandroBorgesFerreira:
[CAS-523](https://stream-io.atlassian.net/browse/CAS-523)

### Description

Substituting #1199... The actually behaviour is much simpler than what I thought it should be. 

The title and subtitle now changes when the user is in thread mode. 

Screenshot_20210113-120500

### Checklist

- Changelog updated with client-facing changes. **no need.** 
- New code is covered by unit tests. **no need.** 
- [x] Reviewers added

giphy
_It was much simpler that I initially thought._


stream-chat-android-ui-components/src/main/res/values/strings.xml
```diff
@@ -77,6 +77,8 @@
     <string name="stream_ui_send_label">Send</string>
     <string name="stream_ui_cancel_label">Cancel</string>
     <string name="stream_ui_attachment_downloading_started">Downloading started</string>
+    <string name="stream_ui_title_thread_reply">Thread Reply</string>
```
zsmb13, line 80: This is unused, and we use a hardcoded string in the sample app's layout instead

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/header/MessagesHeaderView.kt
```diff
@@ -142,6 +161,20 @@ public class MessagesHeaderView : ConstraintLayout {
         binding.avatar.isVisible = false
     }
 
+    public fun setNormalMode() {
+        binding.title.text = normalModeTitle
+        binding.onlineLabel.text = normalModeSubTitle
+    }
+
+    public fun setThreadMode(isGroupThread: Boolean) {
+        val title = threadModeTitle ?: normalModeTitle
```
zsmb13, line 170: If `threadModeTitle` ever ends up being `null` (i.e. the attribute was not set), do we really want to show `normalModeTitle` twice on the header? Perhaps we could use our own `"Thread Reply"` string by default for the main title, we have a string resource for it already anyway.

zsmb13 (commented):
Looks great, some tin questions

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/header/MessagesHeaderView.kt
leandroBorgesFerreira: "Thread Reply" only makes sense for someone that speaks english... I think that simply not changing the title (using normalModeTitle) would be better

zsmb13: All of our UI ships with English text on it by default, I don't see how this is a problem with this specific header text

stream-chat-android-ui-components/src/main/res/values/strings.xml
leandroBorgesFerreira: Now it is used for your second comment

zsmb13 (approved):
LGTM

PR #1237, opened 2021-01-19
API cleanup

zsmb13:
### Description

Enable the binary validator plugin (both in project and as an Actions workflow step), add generated API files, remove / rename some public API. See the [Workflow section](https://github.com/Kotlin/binary-compatibility-validator#workflow) of the plugin's README for details.

The first commit adds the generated `.api` files as they are on `develop`. See the second commit and the changes to the `.api` file there for what's changed in the public API.

Also adds a git pre-commit hook that can run API checks and ktlint so that you don't have to wait for it to fail on CI, configurable for the project by running the following locally:

```
git config core.hookspath hooks
```

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/mentions/MentionsListViewModel.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/mentions/MentionsListViewModel.kt
@@ -25,7 +25,7 @@ public class MentionsListViewModel : ViewModel() {
     )
 
     private companion object {
-        const val QUERY_LIMIT = 30
+        private const val QUERY_LIMIT = 30
 
         val INITIAL_STATE = State(
             results = emptyList(),
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -16,6 +16,19 @@ jobs:
       - name: ktlint
         run: ./gradlew ktlintCheck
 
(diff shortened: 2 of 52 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/mentions/MentionsListViewModel.kt
```diff
@@ -25,7 +25,7 @@ public class MentionsListViewModel : ViewModel() {
     )
 
     private companion object {
-        const val QUERY_LIMIT = 30
+        private const val QUERY_LIMIT = 30
```
zsmb13, line 28: Turns out `const` values inside companion objects are exposed as a public field in the bytecode, so they're worth making doubly `private`...

adasiewiczr (dismissed):

JcMinarro (dismissed):
LGTM

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/mentions/MentionsListViewModel.kt
zsmb13: To clarify, this is a bug that's being tracked here https://youtrack.jetbrains.com/issue/KT-25009

JcMinarro (approved):

PR #1244, opened 2021-01-21
Set ChatMode in MessageInputView

zsmb13:
### Description

Expose `isDirectMessage` about current channel from `MessageInputViewModel`, and set the `chatMode` in `MessageInputView` based on it.

This affects what copy we show next to the "Also send" checkbox:

| Group | Direct message |
| --- | --- |
| Screenshot_20210121_093909 | Screenshot_20210121_093933 |

The PR moves some utility methods to the `common` module as that's where the ViewModel lives.

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/extensions/Channel.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/extensions/Channel.kt
@@ -1,6 +1,32 @@
 package com.getstream.sdk.chat.utils.extensions
 
 import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+import io.getstream.chat.android.livedata.ChatDomain
 
 internal val Channel.isDraft: Boolean
     get() = getExtraValue("draft", false)
+
+@InternalStreamChatApi
+public fun Channel.isDirectMessaging(): Boolean = getUsers().size == 1
+
+@InternalStreamChatApi
+public fun Channel.getUsers(excludeCurrentUser: Boolean = true): List<User> =
+    members
+        .map { it.user }
+        .let { users ->
+            when {
+                excludeCurrentUser -> users.withoutCurrentUser()
+                else -> users
+            }
+        }
+
+private fun List<User>.withoutCurrentUser(): List<User> {
+    return if (ChatDomain.isInitialized) {
+        val currentUser = ChatDomain.instance().currentUser
+        filter { it.id != currentUser.id }
+    } else {
+        this
+    }
+}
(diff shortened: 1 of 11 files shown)


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/extensions/Channel.kt
```diff
@@ -1,6 +1,32 @@
...
 
 import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+import io.getstream.chat.android.livedata.ChatDomain
 
 internal val Channel.isDraft: Boolean
     get() = getExtraValue("draft", false)
+
+@InternalStreamChatApi
```
ogkuzmin, line 11: A nice method to be exposed to our customers

ogkuzmin (approved):

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/extensions/Channel.kt
zsmb13: I'll leave it like this for now, but we can expose it later on. (It would also be a good idea to move it to the new package name before doing so, don't want any more API that uses `com.getstream`)

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessageInputViewModel.kt
```diff
@@ -33,6 +34,9 @@ public class MessageInputViewModel @JvmOverloads constructor(
     public val editMessage: MutableLiveData<Message?> = MutableLiveData()
     public val repliedMessage: MediatorLiveData<Message?> = MediatorLiveData()
 
+    private val _isDirectMessage: MutableLiveData<Boolean> = MutableLiveData()
+    public val isDirectMessage: LiveData<Boolean> = _isDirectMessage
```
samiuelson, line 38: The `isDirectMessage` value will never change during the ViewModel's life. Maybe simple `Boolean` property would be enough instead of `LiveData<Boolean>`? 🤔

samiuelson (approved):
👍🏼

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessageInputViewModel.kt
zsmb13: It would be so ugly to have a regular plain value exposed from a ViewModel though - I agree this is overkill, but it's consistent API at least

PR #1250, opened 2021-01-21
Rework MessageListView handlers

zsmb13:
### Description

Removes usages of Kotlin function types from the interfaces of `MessageListView`, introduces functional interfaces for each handler to replace them. This is so that clients from Java don't have to deal with types like `kotlin.Unit`.

(Plus a bunch of other small fixes)

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsDialogFragment.kt
```diff
@@ -252,14 +253,14 @@ internal class MessageOptionsDialogFragment : FullScreenDialogFragment() {
...
-        val muteClickHandler: (User) -> Unit,
-        val blockClickHandler: (User) -> Unit,
-        val deleteClickHandler: (Message) -> Unit,
-        val replyClickHandler: (String, Message) -> Unit,
+        val threadReplyHandler: MessageListView.StartThreadHandler,
+        val retryHandler: MessageListView.MessageRetryHandler,
+        val editClickHandler: MessageListView.MessageEditHandler,
+        val flagClickHandler: MessageListView.MessageFlagHandler,
+        val muteClickHandler: MessageListView.MuteUserHandler,
+        val blockClickHandler: (user: User) -> Unit,
```
zsmb13, line 261: This is really awkward like this, but we can't take the `Handler` type defined in `MessageListView` as the `Channel` parameter for its method is being filled in by `MessageListView`, and not available here. However, this is all internal API, so we can refactor it later.

stream-chat-android-ui-components/build.gradle
```diff
@@ -83,7 +83,6 @@ dependencies {
     implementation Dependencies.androidLegacySupport
     implementation Dependencies.constraintLayout
     implementation Dependencies.materialComponents
-    implementation Dependencies.drawabletoolbox
```
zsmb13, line 86: This was simply unused

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsDialogFragment.kt
adasiewiczr: I think that we don't need the whole `Channel` object to block the user. It should be enough to pass `Channel::cid`

stream-chat-android-ui-components/build.gradle
adasiewiczr: Should we also remove it from `Dependencies`?

zsmb13: Used by the old UI, so we can't get rid of it quite yet. I'm not sure if I want to replace it with Material stuff there, or just leave the old implementation be and not break it.

adasiewiczr (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsDialogFragment.kt
zsmb13: Refactored so that we pass in the channel ID instead - take a look

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -194,12 +195,14 @@ public class MessageListViewModel @JvmOverloads constructor(
                 client.muteUser(event.user.id).enqueue()
             }
             is Event.BlockUser -> {
+                validateCid(cid)
+                val (channelType, channelId) = event.cid.split(':')
```
adasiewiczr, line 199: I think we can use `client.channel(cid)` to validate cid and obtain `ChannelClient` which methods don't require channel type or id and we won't have to make `validateCid` public

adasiewiczr (approved):
LGTM

PR #1251, opened 2021-01-21
[CAS-627] Isolate Sync state entity

ogkuzmin:
### Description

Remove usage of syncStateEntity from ChatDomain
Wrap syncState into MutableStateFlow

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -257,10 +258,9 @@ internal class ChatDomainImpl internal constructor(
             repos.configs.load()
 
             // load the current user from the db
-            val initialSyncState = SyncStateEntity(currentUser.id)
-            syncState = repos.syncState.select(currentUser.id) ?: initialSyncState
+            syncStateFlow.value = repos.syncState.select(currentUser.id) ?: SyncState(currentUser.id)
```
zsmb13, line 260: This should probably go into a local variable first, as this way we're writing it, and then reading it back twice later on in this function (which might have a new value by then)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -206,8 +206,9 @@ internal class ChatDomainImpl internal constructor(
     }
 
     internal lateinit var repos: RepositoryHelper
-    private var syncState: SyncStateEntity? = null
-    internal lateinit var initJob: Deferred<SyncStateEntity?>
+    @Volatile
+    private var syncStateFlow: MutableStateFlow<SyncState?> = MutableStateFlow(null)
```
zsmb13:
```suggestion
    private val syncStateFlow: MutableStateFlow<SyncState?> = MutableStateFlow(null)
```

zsmb13 (commented):
Just two comments, everything else look good 👍

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
zsmb13: Amazing, this got outdated between the time I drafted the comment and published the review, but GitHub opted not to remove it 😄

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/repository/SyncStateRepositoryTest.kt
```diff
@@ -1,23 +0,0 @@
...
-import com.google.common.truth.Truth
-import io.getstream.chat.android.livedata.BaseDomainTest
-import io.getstream.chat.android.livedata.entity.SyncStateEntity
-import io.getstream.chat.android.livedata.utils.calendar
-import kotlinx.coroutines.runBlocking
-import org.junit.Test
-import org.junit.runner.RunWith
-
-@RunWith(AndroidJUnit4::class)
-internal class SyncStateRepositoryTest : BaseDomainTest() {
```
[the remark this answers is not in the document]
ogkuzmin: In my opinion, every test should check that logic works correctly. This test checked that repo could write, read the same data. Actually, that repo is just a room wrapper. For me, there is no point to test the clean functions of mapping and Room driver as a storaging mechanism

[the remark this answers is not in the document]
zsmb13: It can still make sure that we're not losing something on data type conversions, for example, but I suppose that should be covered by tests of the Converter classes.

And we don't want to test the mapping here either?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
ogkuzmin: I'm not sure how it should work (we don't have any described logic for that or tests). I did some refactoring on that part, take a look

zsmb13: The refactor looks good to me, I think it makes sense this way 👍

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/repository/SyncStateRepositoryTest.kt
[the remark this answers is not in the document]
ogkuzmin: I think mapping testing is too much as they are clean functions itself. Such tests usually don't give any benefits.

zsmb13 (approved):
LGTM

PR #1253, opened 2021-01-22
Improve message list diff performance

bychkovdmitry:
### Description

This PR improves the performance of `MessageListItemDiffCallback`

What was measured:
The time span between `submitList` invocation and `commitCallback`. (Provided that all the messages in the "general" channel are loaded up to the 6th of December and we post a new message to the channel). 

Results vary greatly from time to time but on average 1 diff took:
Before the change - 300ms (150ms - 800ms)
After the change - 4ms (1ms - 8ms)

### Checklist

- [X] Reviewers added


ogkuzmin (approved):

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
```diff
@@ -24,16 +22,13 @@ import java.util.zip.Checksum
...
-        val plaintext = when (this) {
-            is TypingItem -> "id_typing"
-            is ThreadSeparatorItem -> "id_thread_separator"
-            is MessageItem -> message.id
-            is DateSeparatorItem -> date.toString()
-            is LoadingMoreIndicatorItem -> "id_loading_more_indicator"
+        return when (this) {
+            is TypingItem -> TYPING_ITEM_STABLE_ID
+            is ThreadSeparatorItem -> THREAD_SEPARATOR_ITEM_STABLE_ID
+            is MessageItem -> message.id.hashCode().toLong()
```
zsmb13, line 28: The only possible problem is a collision in these hashcodes, but I really hope that that won't be an issue in practice.

zsmb13 (approved):

adasiewiczr (dismissed):

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
bychkovdmitry: Yes, hope so, otherwise we'll have to add additional filed to Message class which will be calculated only once.

ogkuzmin (approved):

PR #1255, opened 2021-01-22
[CAS-641] Isolate ReactionEntity

ogkuzmin:
### Description

Remove ReactionEntity usage in ChatDomain

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -714,7 +714,8 @@ internal class ChatDomainImpl internal constructor(
         return messages
     }
 
-    private suspend fun retryReactions(): List<Reaction> {
+    @VisibleForTesting
+    internal suspend fun retryReactions(): List<Reaction> {
         return repos.selectReactionSyncNeeded().onEach { reaction ->
             reaction.user = null
             val result = if (reaction.deletedAt != null) {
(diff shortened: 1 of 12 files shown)


carterhudson (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -841,39 +839,35 @@ internal class ChannelControllerImpl(
...
+        domainImpl.repos.reactions.insert(reaction)
 
         // update livedata
         val currentMessage = getMessage(reaction.messageId)?.copy()
-        currentMessage?.let {
-            it.removeReaction(reaction, true)
-            upsertMessage(it)
-            domainImpl.repos.messages.insert(it)
-        }
+        currentMessage?.apply { removeReaction(reaction, true) }
```
adasiewiczr, line 852:
```suggestion
        currentMessage?.apply { removeReaction(reaction, updateCounts = true) }
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -713,6 +714,26 @@ internal class ChatDomainImpl internal constructor(
         return messages
     }
 
+    @VisibleForTesting
+    internal suspend fun retryReactions(): List<Reaction> {
+        return repos.selectReactionSyncNeeded().onEach { reaction ->
+            reaction.user = null
```
adasiewiczr, line 720: Why do we have to remove a user from the reaction before updating it on the backend?

ogkuzmin: I actually don't know it was copy-pasted from `ReactionRepository`. If you're sure that it's not right, we can change it

adasiewiczr: I'm not sure about that. Let's leave it for now but maybe it would be worth it to investigate that

adasiewiczr (approved):

PR #1264, opened 2021-01-25
[CAS-632] Thread footnote fix

ogkuzmin:
### Description

Threadfootnote didn't get shown when writing thread reply to the latest message. 
It happens because VH didn't get redrawn since `MessageListItemDiffCallback::areContentsTheSame` returned true. So we need to make messages different when in `ThreadMode` and when in `NormalMode`.  So:
1) Introduce the new field `isThreadMode` to `MessageListItem.MessageItem`
2) Remove from VHs' binding logic the `isThread`flag and use `isThreadMode` from `MessageItem`
3) Add this new flag into comparison logic of `MessageListItemDiffCallback` 

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
@@ -40,6 +40,7 @@ public sealed class MessageListItem {
         val positions: List<Position> = listOf(),
         val isMine: Boolean = false,
         val messageReadBy: List<ChannelUserRead> = listOf(),
+        val isThreadMode: Boolean = false
     ) : MessageListItem() {
         public val isTheirs: Boolean
             get() = !isMine
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/BaseMessageItemViewHolder.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/BaseMessageItemViewHolder.kt
@@ -25,14 +25,14 @@ public abstract class BaseMessageItemViewHolder<T : MessageListItem>(
      * Workaround to allow a downcast of the MessageListItem to T
      */
     @Suppress("UNCHECKED_CAST")
-    internal fun bindListItem(messageListItem: MessageListItem, isThread: Boolean, diff: MessageListItemPayloadDiff? = null) {
+    internal fun bindListItem(messageListItem: MessageListItem, diff: MessageListItemPayloadDiff? = null) {
         messageListItem as T
 
         this.data = messageListItem
-        bindData(messageListItem, isThread, diff)
+        bindData(messageListItem, diff)
     }
 
-    public abstract fun bindData(data: T, isThread: Boolean, diff: MessageListItemPayloadDiff?)
+    public abstract fun bindData(data: T, diff: MessageListItemPayloadDiff?)
 
     internal fun unbind() {
         cancelHighlightAnimation()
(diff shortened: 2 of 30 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/GapDecorator.kt
```diff
@@ -22,40 +22,34 @@ internal class GapDecorator : BaseDecorator() {
 
     override fun decorateDeletedMessage(
         viewHolder: MessageDeletedViewHolder,
-        data: MessageListItem.MessageItem,
-        isThread: Boolean
+        data: MessageListItem.MessageItem
```
zsmb13, line 25: The PR removes a bunch of commas in multiple files, making for extra line changes, but those could actually remain there as trailing commas

image

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
```diff
@@ -40,6 +40,7 @@ public sealed class MessageListItem {
         val positions: List<Position> = listOf(),
         val isMine: Boolean = false,
         val messageReadBy: List<ChannelUserRead> = listOf(),
+        val isThreadMode: Boolean = false
```
zsmb13, line 43:
```suggestion
        val isThreadMode: Boolean = false,
```

zsmb13, line 43: This new parameter in the `MessageItem` constructor is a potentially breaking change. I don't know if any of our clients are creating `MessageItem` instances manually, but it seems possible that they might be. I'd mention this in the changelog.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/BaseMessageItemViewHolder.kt
```diff
@@ -25,14 +25,14 @@ public abstract class BaseMessageItemViewHolder<T : MessageListItem>(
      * Workaround to allow a downcast of the MessageListItem to T
      */
     @Suppress("UNCHECKED_CAST")
-    internal fun bindListItem(messageListItem: MessageListItem, isThread: Boolean, diff: MessageListItemPayloadDiff? = null) {
```
zsmb13, line 28: It's really nice to see all of these parameters gone. Looks like we've ended up with both a more correct and a cleaner solution than what we originally had.

zsmb13 (commented):
This new solution works really nicely, great job 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/GapDecorator.kt
ogkuzmin: It how refactoring in AndroidStudio works)

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
ogkuzmin: Good point

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/GapDecorator.kt
zsmb13: Haha, I figured that that's what happened, that the "remove parameter" refactoring didn't format the files afterwards. Kind of weird that that doesn't happen automatically.

ogkuzmin: It requires tons of manual editing of sources. Since it isn't a strong violation or mistake I'd offer to keep it as is. What do you think? @zsmb13

zsmb13 (approved):
Ship it 🎉

PR #1265, opened 2021-01-25
[CAS-245] send attachment progress

leandroBorgesFerreira:
[CAS-245](https://stream-io.atlassian.net/browse/CAS-245)

### Description

This PR creates a ephemeral message to keep track of the updates of an attachment. 

The custom view is just a place holder, the goal of the PR is to create the communication and create a basic custom view that shows the progress of the upload.

`sendFileSync` Is just a placeholder name for a method that sends a file synchronously and expects a callback in the parameters, some ideas for the name would be good. This method allows the progress to be tracked without big changed in the `sendMessage` method.

**Note: Test this functionality in the demoRelease build type. The interceptors of OkHttp make the loading appear multiple times.**  [source](https://stackoverflow.com/questions/40691610/retrofit-2-requestbody-writeto-method-called-twice)

https://user-images.githubusercontent.com/10619102/106310634-9f01f000-6242-11eb-82e9-2fe3542259eb.mov

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/AttachmentsProgressViewHolder.kt
```diff
@@ -0,0 +1,61 @@
...
+                true
+            }
+        }
+    }
+
+    override fun bindData(data: MessageListItem.MessageItem, diff: MessageListItemPayloadDiff?) {
+        super.bindData(data, diff)
+
+        data.message.uploadId?.let(ProgressTrackerFactory.Companion::getOrCreate)?.let { tracker ->
+            GlobalScope.launch(Dispatchers.Main) {
```
leandroBorgesFerreira, line 42: Couldn't find a better way to launch the coroutine =|

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -608,26 +614,51 @@ internal class ChannelControllerImpl(
...
                     }
-                }
-                attachment
-            }.toMutableList()
 
-            logger.logI("Starting to send message with id ${newMessage.id} and text ${newMessage.text}")
+                    attachment
+                }.toMutableList()
+
+                if (hasAttachments) {
```
zsmb13, line 653: This is always `true`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -608,26 +614,51 @@ internal class ChannelControllerImpl(
...
+                    cancelProgressMessage()
```
zsmb13, line 654: Looking at the usage here, this is more `clear` than `cancel` (as in, it's not an exceptional case, because we completed the upload successfully). Also, it's a single-line method, maybe we can just inline it here, as we don't use it anywhere else.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/ProgressTracker.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import io.getstream.chat.android.client.errors.ChatError
+import io.getstream.chat.android.client.utils.ProgressCallback
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+public class ProgressTracker {
+
+    private var numberOfLaps = 1
+    private var currentProgress: MutableStateFlow<Int> = MutableStateFlow(0)
```
zsmb13, line 11: `currentLapProgress`? (or `currentItemProgress`, if the suggestion above is applied)

zsmb13 (commented):
Tested this and it looks like it works really well (thanks for the testing instructions!).

Don't be scared by the large amounts of comments, it's mostly just tiny questions / suggestions.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/AttachmentsProgressViewHolder.kt
zsmb13: I'll look into creating a proper scope here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/AttachmentsProgressViewHolder.kt
```diff
@@ -0,0 +1,61 @@
...
+            false
+        ),
+) : DecoratedBaseMessageItemViewHolder<MessageListItem.MessageItem>(binding.root, decorators) {
+
+    init {
+        binding.run {
+            root.setOnLongClickListener {
+                listeners.messageLongClickListener.onMessageLongClick(data.message)
+                true
+            }
```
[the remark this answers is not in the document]
leandroBorgesFerreira: Just a bad copy and paste from another class

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/ProgressTracker.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    public fun incrementLap() {
+        lapsCompleted.value += 1
+    }
+}
+
+public fun ProgressTracker.toProgressCallback(): ProgressCallback {
+    return object : ProgressCallback {
+        override fun onSuccess(file: String) {}
```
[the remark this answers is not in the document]
leandroBorgesFerreira: I would prefer to avoid chaining side effects. This way we would need to know which is the max value, which is something this class is not aware. I would prefer this class to have very little logic.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/ProgressTrackerFactory.kt
```diff
@@ -0,0 +1,18 @@
+package io.getstream.chat.android.client.uploader
+
+public class ProgressTrackerFactory {
```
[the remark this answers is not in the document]
leandroBorgesFerreira: This is `stream-chat-android-client` and is being used in `stream-chat-android-ui-components`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileUploader.kt
```diff
@@ -8,6 +8,15 @@ import java.io.File
...
 public interface FileUploader {
 
+    public fun sendFileSync(
+        channelType: String,
+        channelId: String,
+        userId: String,
+        connectionId: String,
+        file: File,
+        callback: ProgressCallback
+    ): String?
```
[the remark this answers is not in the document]
leandroBorgesFerreira: It makes sense, but it would break the api, I'm ok in this case, but not how the team feels about it.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -773,6 +832,10 @@ internal class ChannelControllerImpl(
         return client.sendFile(channelType, channelId, file).execute()
     }
 
+    private fun sendFile(file: File, callback: ProgressCallback): Result<String> {
+        return client.sendFileSync(channelType, channelId, file, callback).execute()
+    }
```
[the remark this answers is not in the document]
leandroBorgesFerreira: For this one, no problem, but I think it won't be nice to use the public methods from Java if we make then `suspend` funtions

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/ProgressTrackerFactory.kt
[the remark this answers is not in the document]
zsmb13: If we don't expect our users to call into it, we can make it `@InternalStreamChatApi`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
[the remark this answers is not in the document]
zsmb13: The similar methods above are `internal` (the entire class is)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/AttachmentsProgressViewHolder.kt
zsmb13: Added a commit that creates and cancels a scope as the ViewHolder is bound/unbound

leandroBorgesFerreira: thanks =]

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/ProgressTracker.kt
```diff
@@ -0,0 +1,41 @@
+package io.getstream.chat.android.client.uploader
```
zsmb13, line 1: `@InternalStreamChatApi` for everything in this file as well?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/AttachmentExtensions.kt
```diff
@@ -4,13 +4,13 @@ import io.getstream.chat.android.client.models.Attachment
 
 internal const val ATTACHMENT_TYPE_IMAGE = "image"
 internal const val ATTACHMENT_TYPE_FILE = "file"
-private const val EXTRA_UPLOAD_COMPLETE: String = "uploadComplete"
+private const val EXTRA_UPLOAD_ID: String = "uploadId"
```
zsmb13, line 7: We're duplicating this constant in two files now

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/SimpleListAdapter.kt
```diff
@@ -36,5 +36,7 @@ internal abstract class SimpleListAdapter<T : Any, VH : SimpleListAdapter.ViewHo
         val context: Context = itemView.context
 
         abstract fun bind(item: T)
+
+        open fun unbind() {}
```
zsmb13, line 40: If this is defined here in the base class, the base class should also do the invocation (currently, the inheriting adapters have to do it for themselves)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FileAttachmentsView.kt
```diff
@@ -121,20 +139,44 @@ private class FileAttachmentViewHolder(
...
+                fileSize.text = MediaStringUtil.convertFileSizeByteCount(attachment.upload?.length() ?: 0L)
+            } else {
+                actionButton.setImageResource(R.drawable.stream_ui_ic_icon_download)
+                val tintColor = ContextCompat.getColor(context, R.color.stream_ui_black)
+                ImageViewCompat.setImageTintList(actionButton, ColorStateList.valueOf(tintColor))
+                fileSize.text = MediaStringUtil.convertFileSizeByteCount(attachment.fileSize.toLong())
+            }
+
+            attachment.uploadId?.let(ProgressTrackerFactory::getOrCreate)?.let { tracker ->
+                scope = CoroutineScope(DispatcherProvider.Main)
```
zsmb13, line 156: Before creating the new scope, it's worth clearing the old one here, _just to be safe_

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileUploader.kt
```diff
@@ -19,7 +19,7 @@ public interface FileUploader {
         connectionId: String,
         file: File,
         callback: ProgressCallback
-    )
+    ): String?
```
zsmb13, line 22: The old implementation here was a fire and forget call for sending a file, likely not useful for anyone (and it gave no opportunity to handle errors).

Making this synchronous instead of async is a breaking API change, but we don't expect anyone to be calling the `FileUploader` APIs directly. We also don't expect any existing implementations other than our own, but those will break at compile time, and will have to be updated to be blocking.

(Still, we should probably add a line in the changelog about this.)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/GsonChatApi.kt
```diff
@@ -116,6 +111,20 @@ internal class GsonChatApi(
         }
     }
 
+    override fun sendImage(
+        channelType: String,
+        channelId: String,
+        file: File,
+        callback: ProgressCallback,
+    ): Unit = fileUploader.sendImage(
```
zsmb13, line 119: Perhaps we should make this (the `sendImage` method we're calling into) blocking as well? It's the only remaining async method in `FileUploader`. Some others return `Unit` too, but they are still synchronous.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -33,7 +33,7 @@ internal interface ChatApi {
         channelId: String,
         file: File,
         callback: ProgressCallback,
-    )
+    ): Call<String>
```
zsmb13, line 36: Just realized we should annotate all functions in this class that return a `Call` with `@CheckResult` - could we add that on this PR?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -201,13 +201,18 @@ public class ChatClient internal constructor(
...
         file: File,
-        callback: ProgressCallback,
-    ) {
-        api.sendFile(channelType, channelId, file, callback)
+        callback: ProgressCallback? = null,
+    ): Call<String> {
+        return if (callback != null) {
+            api.sendFile(channelType, channelId, file, callback)
+        } else {
+            api.sendFile(channelType, channelId, file)
```
zsmb13, line 214: We could probably merge those two methods in `ChatApi` so that we just call into a single one here

(But we shouldn't merge them in `FileUploader`, because it's public and Java clients couldn't make use of the default parameter value)

ogkuzmin (commented):
This solution looks good and solves the task of showing the progress of upload in the sample app.
But in terms of SDK enhancement and reuse in any other app by our customers it doesn’t solve the same task. What cons I see:

In terms of code cleanliness we mix here logic and model representation in the AttachmentsProgressViewHolder  class.
It can’t be reused by other apps based on our SDK
It makes more problems applying the “unidirectional data flow” pattern in our SDK.

As I’m concerned that our SDK lacks of the one architecture approach and the one data flow approach the last point from the list makes me worried.

What I’d suggest for the same purpose. We already have local field in the Attachment class uploadState . It is a sealed class with the following values:
public object InProgress : UploadState()
public object Success : UploadState()
public data class Failed(val error: ChatError) : UploadState()

I suggest to change InProgress  to public class InProgress(val completedPercentage: Float = 0f) . Now it shows a upload status progress from 0f to 1f
So when progress callback throws intermediate result, the result is observed by ChannelControllerImpl (from offline support package). It changes its’ flow value of `ChannelControllerImpl::_messagesin order to update this ephemeral message with such sending attachment file (image) . When the _messages  flow gets altered it triggers new value propagation:
 ChannelControllerImpl-> MessageListViewModel -> MessageListLiveData -> MessageListAdapter -> FileAttachmentsViewHolder . 
The last one receives a new value with the payload of upload status progress and shows the new result. So the same task will be solved and it could be reused by other customers and we keep our UI update flow.
If you had an intention of fast feature implementation then we can refactor it later. 
Also, I could miss something or don’t see why my suggested idea is worse. I’m open to discussing it here or via call.

zsmb13:
The suggestion above by @ogkuzmin makes sense to me, I'd also move towards that 👍  I think the question is whether we want to implement that now, or only later within a larger rewrite of the upload/download/file handling system in the SDK.

(Plus we've yet to find out how good our diffing is, if sending this many progress updates through the adapter is fast enough or not.)

leandroBorgesFerreira:
@ogkuzmin  I thought about updating the progress inside the `Message` and `Attachment` classes, but the amount of times that we would reload the views inside the adapter would be really high, it is very likely that we will run into performance issues 
this way. Also, I think that the view will be recreated many times and we will see it blinking every time it is redraw in the screen.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileUploader.kt
JcMinarro: `FileUploader` could be implemented by our customers to upload files to their own server

zsmb13: Those implementations will indeed break and have to be updated (should be fairly simple), but at least they won't get a silent behaviour change.

We're also betting on not many customers having done this yet, as it's pretty new API.

We pretty much can't fix this without a breaking change for existing implementations, as introducing a new method in the interface with a different name would also be a breaking change, and we can't adapt a fire-and-forget async API into a sync one, so a default implementation can't be provided either.

JcMinarro:
About the progress, we are discretizing it during the upload process, and values should go from 1 to 100. We shouldn't have tons of updated in our UI because buffer is about 2Kb and we doesn't allow very big files. BTW, on the worng case we will have 100 updates that should only update some part of our VH, the UI shouldn't blink on the rest of the VH neither on the other parts that are not updated into the Upload VH

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileUploader.kt
JcMinarro: Yep, I agree, I wanted to let you all know that it is a public interface because it can be (and it is)  implemented by our customers, and what is the reason they would like to implement it.
The breaking change is easy to fix the when they will need to update the integration with our SDK

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/ProgressTracker.kt
leandroBorgesFerreira: As we are discussing using DiffUtils instead if ProgressTracker, let's move it to `stream-chat-android-ui-components` and make then `internal`. Although they are not a UI component, I believe it is better keep then there so we can improve the communication with the views in a next PR and don't expose those classes that are going away. What do you think?

zsmb13: If we don't depend on this or the factory in the client module, then yes, moving them to the UI components module makes sense.

leandroBorgesFerreira: Yeah... not possible... 2 different modules depend of this class. Let's put the annotation then.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -608,26 +630,43 @@ internal class ChannelControllerImpl(
...
-                }
-                attachment
-            }.toMutableList()
 
-            logger.logI("Starting to send message with id ${newMessage.id} and text ${newMessage.text}")
+                    attachment
+                }.toMutableList()
 
+                uploadStatusMessage?.let { cancelMessage(it) }
+                uploadStatusMessage = null
```
zsmb13, line 662: This is inside the `if (online)` condition. If we're not online, this property is never cleared?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -202,13 +202,18 @@ public class ChatClient internal constructor(
...
+        }
```
zsmb13, line 216: This seems unnecessary now that `ChatApi` can handle nullable values for `callback` too

stream-chat-android-client/api/stream-chat-android-client.api
```diff
@@ -81,8 +81,8 @@ public final class io/getstream/chat/android/client/ChatClient {
...
 	public final fun sendEvent (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/Map;)Lio/getstream/chat/android/client/call/Call;
 	public static synthetic fun sendEvent$default (Lio/getstream/chat/android/client/ChatClient;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/Map;ILjava/lang/Object;)Lio/getstream/chat/android/client/call/Call;
-	public final fun sendFile (Ljava/lang/String;Ljava/lang/String;Ljava/io/File;)Lio/getstream/chat/android/client/call/Call;
-	public final fun sendFile (Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;)V
+	public final fun sendFile (Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;)Lio/getstream/chat/android/client/call/Call;
+	public static synthetic fun sendFile$default (Lio/getstream/chat/android/client/ChatClient;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;ILjava/lang/Object;)Lio/getstream/chat/android/client/call/Call;
```
zsmb13, line 85: This change of public method signatures on `ChatClient` should go in the changelog

stream-chat-android-client/api/stream-chat-android-client.api
```diff
@@ -2645,9 +2645,13 @@ public abstract interface class io/getstream/chat/android/client/uploader/FileUp
...
 	public abstract fun deleteImage (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;)V
 	public abstract fun sendFile (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;)Ljava/lang/String;
-	public abstract fun sendFile (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;)V
+	public abstract fun sendFile (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;)Ljava/lang/String;
 	public abstract fun sendImage (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;)Ljava/lang/String;
-	public abstract fun sendImage (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;)V
+	public abstract fun sendImage (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;)Ljava/lang/String;
+}
```
zsmb13, line 2651: These `FileUploader` interface changes should also be in the changelog

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/OnlyFileAttachmentsViewHolder.kt
```diff
@@ -56,5 +70,60 @@ internal class OnlyFileAttachmentsViewHolder(
...
+        }
+    }
+
+    private suspend fun trackFilesSent(uploadIdList: List<String>) {
+        var filesSent = 0
+        val totalFiles = uploadIdList.size
+
+        binding.sentFiles.isVisible = true
+        binding.sentFiles.text =
+            String.format(context.getString(R.string.stream_ui_upload_sending), filesSent, totalFiles)
```
zsmb13, line 100: Nit: `getString` can take the format arguments too, no need for `String.format`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/OnlyFileAttachmentsViewHolder.kt
```diff
@@ -56,5 +70,60 @@ internal class OnlyFileAttachmentsViewHolder(
...
+
+            scope.launch {
+                trackFilesSent(uploadIdList)
+            }
+        } else {
+            binding.sentFiles.isVisible = false
+        }
+    }
+
+    private suspend fun trackFilesSent(uploadIdList: List<String>) {
```
zsmb13, line 94: Could we share this method between the ViewHolders that use it?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/OnlyFileAttachmentsViewHolder.kt
```diff
@@ -56,5 +70,60 @@ internal class OnlyFileAttachmentsViewHolder(
...
+        }.collect { stateList ->
+            filesSent = stateList.size
+
+            if (filesSent == totalFiles) {
+                binding.sentFiles.text = context.getString(R.string.stream_ui_upload_complete)
+            } else {
+                binding.sentFiles.text =
+                    String.format(context.getString(R.string.stream_ui_upload_sending), filesSent, totalFiles)
+            }
+        }
```
zsmb13, line 117: I think this is the same logic - but please do check

```
val completionFlows: List<Flow<Boolean>> = uploadIdList.map { uploadId ->
    ProgressTrackerFactory.getOrCreate(uploadId).isComplete()
}
combine(completionFlows) { values -> values.count { complete -> complete } }
    .collect { completedCount ->
        filesSent = completedCount
        // ...
    }
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: Yes, but this is an ephemeral message. The user will see this message, but no update in the progress of the attachment. If the user leave the app, the message is not persistet

leandroBorgesFerreira: The current behaviour is that the message with the attachment stays in the conversation if the user is not online

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/OnlyFileAttachmentsViewHolder.kt
leandroBorgesFerreira: Much simpler o.O. I didn't know about the `combine` operator

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: Yep, checked.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -608,26 +630,43 @@ internal class ChannelControllerImpl(
...
+                    val attachmentProgress = attachmentProgressList[i]
+
+                    if (attachment.upload != null) {
+                        val result = uploadAttachment(
+                            attachment,
+                            attachmentTransformer,
+                            attachmentProgress.toProgressCallback()
+                        )
+
+                        if (result.isSuccess) {
```
ogkuzmin, line 648: It was previously a messy method, new lines make it more and more complicated. it requires refactoring... 😢 
What if extract result of upload attachments to some `handleUploadResult(result: Result<Attachment>, progressTracker: ProgressTracker)?`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: Made `uploadStatusMessage = null` in the offline mode as we discussed.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -608,26 +630,43 @@ internal class ChannelControllerImpl(
...
-                    if (result.isSuccess) {
-                        attachment = result.data()
-                    } else {
-                        attachment.uploadState = Attachment.UploadState.Failed(result.error())
-                        logger.logE("Failed to upload attachment for message")
+
+            if (hasAttachments) {
+                logger.logI("Uploading attachments for message with id ${newMessage.id} and text ${newMessage.text}")
+
+                newMessage.attachments = newMessage.attachments.mapIndexed { i, attach ->
```
ogkuzmin, line 637: Why do you need local `attachmentProgressList ` and mapping with index? ProgressTrackerFactory already can return existing progressTracker by some id.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -573,8 +580,9 @@ internal class ChannelControllerImpl(
         }
 
         newMessage.user = domainImpl.currentUser
+
         // TODO: type should be a sealed/class or enum at the client level
-        newMessage.type = if (newMessage.text.startsWith("/")) {
+        newMessage.type = if (newMessage.text.startsWith("/") || hasAttachments) {
```
ogkuzmin, line 585: Nit: This could be extracted to some `fun getMessageType(message: Message): String`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/FileAttachmentsView.kt
```diff
@@ -121,20 +136,60 @@ private class FileAttachmentViewHolder(
...
+                    )
+
+                    val fileProgress = progress.combine(completion, ::Pair)
+
+                    clearScope()
+                    scope = CoroutineScope(DispatcherProvider.Main)
+
+                    scope!!.launch {
+                        fileProgress.collect { (progress, isComplete) ->
+                            if (!isComplete) {
```
ogkuzmin, line 169: Please extract this to some smaller method. It's very hard to read (something like `handleUploadResult` )

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: Do you mean  the part inside `if (result.isSuccess) {`? This part of the code wasn't changed, but I can extract this to another method and extract some parts of `sendMessage` to another method.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
leandroBorgesFerreira: It is possible to create the ProgressTracker inside the `map` and don't use the index, but I create `attachmentProgressList` before `upsertMessage(newMessage)` is called, so when the `FileAttachmentsView::bind` is called all the `ProgressTracker` are already created. This makes the debug process easier... otherwise it would be the adapter creating the trackers.

leandroBorgesFerreira: I create all the trackers and store then in `attachmentProgressList` before the `upsertMessage(newMessage)` is called, so when the `FileAttachmentViewHolder::bind` is called all the trackers are already created. 

Although managing the instances with the factory would work, the adapter would create the trackers first if I call the `getOrCreate` inside `newMessage.attachments.mapIndexed`, making the logic harder to reason about. I could create the trackers and not assigning then to any list/variable, but that would look odd to me too.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -608,64 +626,116 @@ internal class ChannelControllerImpl(
...
+            } else {
+                attachmentProgress.setComplete(false)
+                newAttachment.uploadState = Attachment.UploadState.Failed(result.error())
+            }
+        }
+
+        return newAttachment
+    }
+
+    private suspend fun handleSendAttachmentSuccess(result: Result<Message>): Result<Message> {
```
ogkuzmin, line 688: Now it's much more readable 👍

ogkuzmin (approved):
LGTM
As we agreed to refactor it later

PR #1266, opened 2021-01-25
Fix adding date separators

adasiewiczr:
| Before | After |
| --- | --- |
|Screenshot 2021-01-25 at 14 21 48|Screenshot 2021-01-25 at 14 20 34|

1. Add possibility to configure when to add date separator through `DateSeparatorHandler`. We should reconsider the way we are passing it to `MessageListItemLiveData` because it won't work fine if the handler will be set after presenting results to the user. Maybe we should add an extra rerender?
2. Add missing TOP position to Message which has `DateSeparator` above

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
@@ -59,6 +59,24 @@ public class MessageListViewModel @JvmOverloads constructor(
     public val state: LiveData<State> = stateMerger
     public val currentUser: User = domain.currentUser
 
+    private var dateSeparatorHandler: DateSeparatorHandler? =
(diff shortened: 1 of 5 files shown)


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -210,6 +213,26 @@ public class MessageListViewModel @JvmOverloads constructor(
         }.exhaustive
     }
 
+    /**
+     * Sets the date separator handler which determines when to add date separators.
+     * By default, a date separator will be added if the difference between two messages' dates is greater than 4h.
+     *
+     * @param dateSeparatorHandler The handler to use. If null, [messageListData] won't contain date separators.
+     */
```
zsmb13, line 221: Nicely done with these docs 👍

zsmb13 (approved):
Looks great, well done!

PR #1268, opened 2021-01-25
[CAS-501] Enable reply feature

ogkuzmin:
### Description
1) Enable reply feature
2) Fix in dynamic view setup in ReplyView

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/options/MessageOptionsView.kt
```diff
@@ -64,8 +63,7 @@ public class MessageOptionsView : FrameLayout {
     private fun configureMineMessage(configuration: Configuration, syncStatus: SyncStatus) {
         val iconsTint = configuration.iconsTint
 
-        // Uncomment when the feature will be deployed to our backend
-        // binding.replyTV.isVisible = true
+        binding.replyTV.isVisible = true
```
zsmb13, line 66: If this is always visible, we could set it in the XML and not touch it from code

ogkuzmin: Make sense 👍

zsmb13 (approved):

PR #1269, opened 2021-01-25
Tutorial update chat event listener

carterhudson:
### Description

We recently updated several Kotlin functions used in our public API to functional interfaces so that Java lambdas don't have to `return Unit.INSTANCE`. This is the same thing (relevant in `ChannelActivity4.java`): 
`(ChatEvent) -> Unit` becomes 
```kotlin
fun interface ChatEventListener<EventT: ChatEvent> {
    fun onEvent(event: EventT)
}
```

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/Subscriptions.kt
```diff
@@ -32,7 +33,7 @@ internal open class SubscriptionImpl(
 
         if (filter(event)) {
             try {
-                listener!!.invoke(event)
+                listener!!.onEvent(event)
```
carterhudson, line 36: Why don't we safe unwrap here?

bychkovdmitry (approved):
LGTM

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/Subscriptions.kt
JcMinarro: Why don't we use a non-null property and assign a "no-op function" when it is disposed?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/ChatObservable.kt
```diff
@@ -14,7 +15,7 @@ public interface ChatObservable {
     @Deprecated("Use the subscribe methods on ChatClient or ChannelController directly instead of events()")
     public fun first(): ChatObservable
     @Deprecated("Use the subscribe methods on ChatClient or ChannelController directly instead of events()")
-    public fun subscribe(listener: (ChatEvent) -> Unit): Subscription
+    public fun subscribe(listener: ChatClient.ChatEventListener<ChatEvent>): Subscription
```
zsmb13, line 18: This method is deprecated already, no reason to change it - we'll remove it soon.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
```diff
@@ -99,21 +99,21 @@ public class ChannelClient internal constructor(
...
+        listener: ChatClient.ChatEventListener<ChatEvent>,
     ): Disposable {
         return client.subscribeFor(*eventTypes, listener = filterRelevantEvents(listener))
     }
 
     override fun subscribeFor(
         lifecycleOwner: LifecycleOwner,
         vararg eventTypes: String,
-        listener: (event: ChatEvent) -> Unit,
+        listener: ChatClient.ChatEventListener<ChatEvent>,
```
zsmb13, line 116: It's weird that in `ChannelClient` we use an interface nested in `ChatClient` - it should just be a top level interface instead.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/Subscription.kt
```diff
@@ -1,11 +1,12 @@
 package io.getstream.chat.android.client.utils.observable
 
+import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.events.ChatEvent
 
 @Deprecated("Use the subscribe methods on ChatClient or ChannelController directly instead of events()")
 public class Subscription(
     private val observable: ChatObservable,
-    private var listener: ((ChatEvent) -> Unit)?,
+    private var listener: ChatClient.ChatEventListener<ChatEvent>?,
```
zsmb13, line 9: This class is also deprecated already, no reason to update it

zsmb13 (changes requested):
This is a good idea, but it's a breaking change for nearly all Java clients that currently use this API. Here's some examples I checked that work with the old API but won't compile with the new one:

```java
client.subscribe(new Function1<ChatEvent, Unit>() {
    @Override
    public Unit invoke(ChatEvent chatEvent) {
        return null;
    }
});
client.subscribe(new Function1<ChatEvent, Unit>() {
    @Override
    public Unit invoke(ChatEvent chatEvent) {
        return Unit.INSTANCE;
    }
});
client.subscribe(chatEvent -> null);
client.subscribe(chatEvent -> Unit.INSTANCE);

client.subscribeForSingle(NewMessageEvent.class, new Function1<NewMessageEvent, Unit>() {
    @Override
    public Unit invoke(NewMessageEvent chatEvent) {
        return Unit.INSTANCE;
    }
});
client.subscribeForSingle(NewMessageEvent.class, chatEvent -> Unit.INSTANCE);
```

We should provide a nice migration path here if possible, or at least a very large and scary warning in the changelog.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/Subscriptions.kt
zsmb13: We know that `listener` is never `null` here, otherwise the instance would also be already marked as disposed (in which case, `onNext` should never be called on it in the first place).

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/Subscription.kt
carterhudson: > This is a good idea, but it's a breaking change for nearly all Java clients that currently use this API. Here's some examples I checked that work with the old API but won't compile with the new one:
> 
> ```java
> client.subscribe(new Function1<ChatEvent, Unit>() {
>     @Override
>     public Unit invoke(ChatEvent chatEvent) {
>         return null;
>     }
> });
> client.subscribe(new Function1<ChatEvent, Unit>() {
>     @Override
>     public Unit invoke(ChatEvent chatEvent) {
>         return Unit.INSTANCE;
>     }
> });
> client.subscribe(chatEvent -> null);
> client.subscribe(chatEvent -> Unit.INSTANCE);
> 
> client.subscribeForSingle(NewMessageEvent.class, new Function1<NewMessageEvent, Unit>() {
>     @Override
>     public Unit invoke(NewMessageEvent chatEvent) {
>         return Unit.INSTANCE;
>     }
> });
> client.subscribeForSingle(NewMessageEvent.class, chatEvent -> Unit.INSTANCE);
> ```
> 
> We should provide a nice migration path here if possible, or at least a very large and scary warning in the changelog.

What would a nice migration path look like? Supporting both the functional interface approach & the Kotlin function approach in subscribeFor methods, and deprecating one set?

carterhudson:
> We should provide a nice migration path here if possible, or at least a very large and scary warning in the changelog.

What would a nice migration path look like? Supporting both the functional interface approach & the Kotlin function approach in subscribeFor methods, and deprecating one set?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/controllers/ChannelController.kt
```diff
@@ -119,11 +120,26 @@ public interface ChannelController {
     )
     public fun events(): ChatObservable
 
-    public fun subscribe(listener: (event: ChatEvent) -> Unit): Disposable
+    public fun subscribe(listener: ChatEventListener<ChatEvent>): Disposable
+
+    @Deprecated(
+        message = "Use subscribeFor with ChatEventListener parameter",
+        level = DeprecationLevel.WARNING
+    )
```
zsmb13, line 128: This kind of Kotlin syntax will always resolve to the lambda parameter functions first (instead of the interface parameter functions), which means that all Kotlin users of these APIs will see warnings on their code, despite it being correct.

We can avoid this by adding this annotation to these deprecated functions, removing them from Kotlin call resolution entirely:

```kotlin
@SinceKotlin("99999.9")
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/controllers/ChannelController.kt
```diff
@@ -152,22 +187,32 @@ public interface ChannelController {
...
+        listener: (ChatEvent) -> Unit,
     ): Disposable
 
     /**
      * Subscribes for the next channel event with the given [eventType].
      *
      * @see [io.getstream.chat.android.client.models.EventType] for type constants
      */
-    public fun subscribeForSingle(eventType: String, listener: (event: ChatEvent) -> Unit): Disposable
+    public fun subscribeForSingle(eventType: String, listener: ChatEventListener<ChatEvent>): Disposable
```
zsmb13, line 208: There are some method here that don't have the old declarations anymore that take a function as a parameter, just the new one

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -320,7 +320,7 @@ public class ChatClient internal constructor(
     }
 
     public fun subscribe(
-        listener: (event: ChatEvent) -> Unit,
+        listener: ChatEventListener<ChatEvent>,
```
zsmb13, line 323: We didn't keep the old declarations in this class

UNRELEASED_CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Add `enforceUnique` parameter to `ChatClient::sendReaction` and `ChannelClient::sendReaction` methods .
 If reaction is sent with `enforceUnique` set to true, new reaction will replace all reactions the user has on this message.
 - Add suspending `setUserAndAwait` extension for `ChatClient`
+- Replace client event listener functions with ChatEventListener functional interface.
```
zsmb13, line 10: Might want to expand this explanation with something about the old APIs being deprecated

zsmb13 (commented):
This _is_ what I meant by a nice migration path! 👍   Some comments on the details.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/controllers/ChannelController.kt
carterhudson: Interesting, I'm not seeing that behavior in the updated docs, but I'll add the annotation anyway just to be safe.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
carterhudson: Added the old function back and deprecated it

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -311,16 +311,22 @@ public class ChatClient internal constructor(
...
+    ): Disposable {
+        return eventsObservable.subscribe(listener = listener)
+    }
+
+    @Suppress("NEWER_VERSION_IN_SINCE_KOTLIN")
+    @Deprecated(message = "Use subscribe with ChatEventListener parameter")
+    @SinceKotlin("99999.9")
```
carterhudson, line 327: Is there some way to compose / combine annotations in Kotlin without writing your own parser to unwrap / expand annotated annotations? Would be nice if we could have `@SinceKotlin("99999.9")` become something like `@ResolveLast` or something, or just be able to get rid of the string literal.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/observable/Subscriptions.kt
carterhudson: If we know it's not null, what's the `try` block for?

zsmb13: In the case when delivering the event throws (something errors out during the execution of the listener), we still want to run the after-event callback. We use it for the single-event subscribe mechanism, mostly. This way if that single delivery fails, disposal of the subscription still happens, and we don't end up delivering the next event as well.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
zsmb13: I believe we don't have a way to achieve this at the moment. Would be great though!

Meta-annotations are possible in Spring (e.g. here's a [custom `SpringTest` annotation](https://github.com/zsmb13/site/blob/6d55b09c587152d75926d9d8b4e69361b3093957/site-backend/src/test/kotlin/co/zsmb/site/backend/setup/SpringTest.kt) I created that adds a whole lot of config to a test in one annotation), but they have their own special processing there as far as I understand.

zsmb13 (approved):
Added a few commits, mostly to improve the GitHub diff view. Also fixed some param types and such.

I think this is good to go.

PR #1271, opened 2021-01-26
Add max height when only one attachment is rendered

JcMinarro:
### Description
Add max height when only one attachment is rendered

| Before | After |
| --- | --- |
|  photo_2021-01-26_09-42-39  | photo_2021-01-26_09-42-37  |

### Checklist

- [x] Reviewers added


ogkuzmin (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/view/MediaAttachmentsGroupView.kt
```diff
@@ -21,6 +22,9 @@ import io.getstream.chat.android.ui.utils.extensions.getOrDefault
 internal class MediaAttachmentsGroupView : ConstraintLayout {
     var attachmentClickListener: AttachmentClickListener? = null
     var attachmentLongClickListener: AttachmentLongClickListener? = null
+    val maxMediaAttachmentHeight: Int by lazy {
+        (Resources.getSystem().displayMetrics.heightPixels * MAX_HEIGHT_PERCENTAGE).toInt()
+    }
```
zsmb13, line 27: Does this really have to be display size percentage dependent? This would make it behave weirdly with setting the display size in System settings for example (it would be unaffected). I'd prefer a fixed `dp` amount instead.

JcMinarro: Fixed dp would cut tall pictures on big screens that are able to render them properly

JcMinarro: On the other hand, because I am getting the display metrics height (Obtained on runtime, so if anyone updates this property on the system, we use the proper one) and applying a percentage, we are sure that tall pictures will fit inside of the screen on all cases

zsmb13: I'd rather see the same relative sizing of images to other elements of the UI on different height devices, but we can go with this as well if you prefer it

bychkovdmitry (approved):

PR #1274, opened 2021-01-26
Implement basic message sending with Moshi

zsmb13:
https://stream-io.atlassian.net/browse/CAS-619

### Description

First, super basic implementation of a few endpoints with Moshi to verify that the basic setup works correctly. Testable by sending / deleting messages after setting the `config.enableMoshi` flag to `true` in `ChatClient`.

See the self-comments on the PR for more info.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MessageApi.kt
```diff
@@ -0,0 +1,52 @@
...
+import retrofit2.http.Body
+import retrofit2.http.DELETE
+import retrofit2.http.GET
+import retrofit2.http.POST
+import retrofit2.http.Path
+import retrofit2.http.Query
+
+@MoshiApi
+@AuthenticatedApi
+internal interface MessageApi {
```
zsmb13, line 16: Example of a new Api interface, not sure how many of these we'll want (but more than the single one in the old impl)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiApi.kt
```diff
@@ -0,0 +1,8 @@
+package io.getstream.chat.android.client.api2
+
+/**
+ * Temporary marker interface for Moshi based Retrofit API classes.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.CLASS)
+internal annotation class MoshiApi
```
zsmb13, line 8: Used to identify Retrofit interfaces that are used with Moshi - needed so that the Moshi parser is used with these instead of the Gson parser

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1,7 +1,62 @@
...
+
+    override fun sendMessage(channelType: String, channelId: String, message: Message): Call<Message> {
+        return messageApi.sendMessage(
+            channelType = channelType,
+            channelId = channelId,
+            apiKey = apiKey,
+            userId = userId,
+            connectionId = connectionId,
+            message = MessageRequest(mapper.toDto(message)),
+        ).map { response -> mapper.toDomain(response.message) }
```
zsmb13, line 32: Some duplication here, leaving it for now

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/DownstreamMessageDto.kt
```diff
@@ -0,0 +1,20 @@
+package io.getstream.chat.android.client.api2.model.dto
+
+import com.squareup.moshi.JsonClass
+import java.util.Date
+
+@JsonClass(generateAdapter = true)
+internal data class DownstreamMessageDto(
```
zsmb13, line 7: In the old `Message` class, we have many `@IgnoreSerialisation` annotations. Here, we'll use separate models for `Message` objects being sent and received instead, so that there are no always-unused fields to deal with.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDto.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+internal data class UserDto(
+    var id: String,
+    var role: String,
+    var invisible: Boolean,
+    var banned: Boolean,
+)
```
zsmb13, line 11: These model classes have a minimal set of data for now, for basic tests. Other properties will reference more models that have to be implemented (e.g. Attachments), plus the "custom object" behaviour and parsing is not implemented yet either.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/DtoMapper.kt
```diff
@@ -0,0 +1,59 @@
+package io.getstream.chat.android.client.api2
+
+import io.getstream.chat.android.client.api2.model.dto.DownstreamMessageDto
+import io.getstream.chat.android.client.api2.model.dto.UpstreamMessageDto
+import io.getstream.chat.android.client.api2.model.dto.UserDto
+import io.getstream.chat.android.client.models.Message
+import io.getstream.chat.android.client.models.User
+
+internal class DtoMapper {
```
zsmb13, line 9: Might want to split this out into multiple classes or rename it later, but probably good enough for now

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/DateAdapter.kt
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Date
+import java.util.Locale
+import java.util.TimeZone
+
+internal class DateAdapter : JsonAdapter<Date>() {
+
+    companion object {
+        private const val DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+        private const val DATE_FORMAT_WITHOUT_NANOSECONDS = "yyyy-MM-dd'T'HH:mm:ss'Z'"
+    }
```
zsmb13, line 17: Based on the Gson implementation

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/DownstreamMessageDto.kt
```diff
@@ -0,0 +1,20 @@
...
+    val id: String,
+    val cid: String,
+    val text: String,
+    val html: String,
+    val parent_id: String?,
```
adasiewiczr, line 12: We are doing that to avoid using `@Filed` right? IMO it would be better to keep it in camel case as we do everywhere else (even with views' ids). Does Moshi provide something similar to `GsonBuilder:: setFieldNamingPolicy `?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/DownstreamMessageDto.kt
```diff
@@ -0,0 +1,20 @@
...
+    val command: String?,
+    val user: UserDto,
+    val silent: Boolean,
+    val shadowed: Boolean,
+    val created_at: Date,
+    val updated_at: Date,
+    val deleted_at: Date? = null,
```
adasiewiczr, line 19: Just out of curiosity  - why are we adding default value only for `deleted_at` and not for every nullable field?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/DownstreamMessageDto.kt
zsmb13: Moshi does _not_ have a field naming policy, and we should be really happy about that! 😄 

Compared to Gson, from [their README](https://github.com/square/moshi#borrows-from-gson):

>Moshi is less configurable. There’s no field naming strategy, versioning, instance creators, or long serialization policy. Instead of naming a field `visibleCards` and using a policy class to convert that to `visible_cards`, Moshi wants you to just name the field `visible_cards` as it appears in the JSON.

Some reasons why this is great:

- Code is more straightforward, less magic involved
- No ugly `@SerializedName` annotations making the classes unreadable
- You can search for the JSON field's name in the project and find the corresponding property
- It makes you not want to use these fields in your code because they're ugly in Kotlin, encouraging you to not use your network models all over the place, but to map them to domain models instead

adasiewiczr: Fair enough! Thanks for great explanation

ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/DtoMapper.kt
```diff
@@ -0,0 +1,59 @@
...
+            user = toDomain(user),
+            silent = silent,
+            shadowed = shadowed,
+            createdAt = created_at,
+            updatedAt = updated_at,
+            deletedAt = deleted_at,
+        )
+    }
+
+    fun toDto(user: User): UserDto = user.run {
```
JcMinarro, line 42: What about fun extensions?? :thinking: 
```
internal fun User.toDto(): UserDto
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/DownstreamMessageDto.kt
zsmb13: Good catch, an oversight on my part. It makes no sense to have a default `null` value here, anything nullable is treated as optional and will be `null` if not present in the JSON.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/DtoMapper.kt
```diff
@@ -0,0 +1,59 @@
...
+            html = html,
+            parent_id = parentId,
+            command = command,
+            user = toDto(user),
+            silent = silent,
+            shadowed = shadowed,
+        )
+    }
+
+    fun toDomain(message: DownstreamMessageDto): Message = message.run {
```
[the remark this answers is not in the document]
zsmb13: The top level objects that will travel to/from the API already have `Request` and `Response` in their names, so I don't want to reuse those specific words. But I'm open for other suggestions to replace `Upstream`/`Downstream` 😄

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/DtoMapper.kt
zsmb13: I think I'll go with this approach, but I'll do it in one of the next PRs.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/DtoMapper.kt
zsmb13: Probably gonna go with splitting out into extensions https://github.com/GetStream/stream-chat-android/pull/1274#discussion_r565174986

adasiewiczr (approved):
LGTM

PR #1279, opened 2021-01-27
[CAS-551] Channel config entity isolation

ogkuzmin:
### Description

1) Isolation of Channel config entity from ChatDomain
2) Mapping to model/entity
3) Refactor table relation in order to implement the "one-to-many" pattern

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/model/ChannelConfig.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import io.getstream.chat.android.client.models.Config
+
+internal typealias ChannelConfig = Pair<String, Config>
+
+internal val ChannelConfig.type: String
+    get() = this.first
+
+internal val ChannelConfig.config: Config
+    get() = this.second
```
zsmb13, line 11: What's the benefit of this over a simple data class?

ogkuzmin: Why do you need to create any new class if it's just a simple pair? It covers all required use-cases for ChannelConfig.

zsmb13: For the exact reason why you've introduced the two extension properties here, to give the two values in it meaningful names. Creating new classes is super cheap in Kotlin - `Pair` really only exists to make creating Maps easier with `mapOf` (no idea why `Triple` exists). If you need a tuple, you can always create a meaningful data class in a single line for it.

With this solution, we have more code, we have `first` and `second` as available properties on `ChannelConfig`, as well as in its `toString` implementation, and other Pairs of these two values are cross-assignable with `ChannelConfig`.

ogkuzmin: I don't have any strong point against implementation via Pair. From my point of view, it's the same, and data classes contain their `componentN()` functions that can confuse (but really no one gets). Feels like a matter of taste. But if you really don't like it. I can change easily

P.S. `Pair`, `Triple` are famous Tuple concept. And it's one of the weakest parts of the Kotlin language design. For example [C# has a much better implementation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples)

zsmb13: `Pair` is also a data class, so it'll also have the component functions and all that functionality. The only difference between using `Pair` and your own data class is that you get to name your values nicely. Makes them easier to access, gives you a correct `toString` implementation, correct parameter names in the `copy` method, etc.

Kotlin used to have tuples, but since data classes are so concise and easy to create wherever you need them, they removed separate support for them. [This article](https://blog.jetbrains.com/kotlin/2012/09/how-do-you-traverse-a-map/) shows off some ancient history for this 😄

ogkuzmin: About topic - I'll change to data classes.

About tuples. I think it's also defined by the support requirement of java SDK background and its' work with `Map` and so on. data classes are too wordy when you want to just receive some values as `(Long, String, Double, SomeOtherType)`. It couldn't be related to logic/domain entoty model

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/ChannelConfigDao.kt
```diff
@@ -4,19 +4,37 @@ import androidx.room.Dao
...
+    suspend fun insertConfigs(channelConfigInnerEntities: List<ChannelConfigInnerEntity>)
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
-    suspend fun insert(channelConfigEntity: ChannelConfigEntity)
+    suspend fun insertCommands(commands: List<CommandInnerEntity>)
 
-    @Query(
-        "SELECT * FROM stream_chat_channel_config LIMIT 100"
-    )
+    @Transaction
```
zsmb13, line 37: Does `@Transaction` do anything on a method with no body?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -192,13 +192,8 @@ internal class QueryChannelsControllerImpl(
                 _endOfChannels.value = true
             }
             // first things first, store the configs
-            val configEntities = channelsResponse.associateBy { it.type }.values.map {
```
zsmb13, line 195: Was this `associateBy` not needed here, are there no duplicate types in this response?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -740,13 +741,13 @@ internal class ChatDomainImpl internal constructor(
 
     suspend fun storeStateForChannels(channelsResponse: List<Channel>) {
         val users = mutableMapOf<String, User>()
-        val configs: MutableMap<String, Config> = mutableMapOf()
```
zsmb13, line 743: Similar question here, was this a `Map` for no reason, do we never see duplicate channel types here that would override each other in the loop below?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/ChannelConfigDao.kt
```diff
@@ -4,19 +4,37 @@ import androidx.room.Dao
...
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertConfig(channelConfigInnerEntity: ChannelConfigInnerEntity)
+
+    @Insert(onConflict = OnConflictStrategy.REPLACE)
+    suspend fun insertConfigs(channelConfigInnerEntities: List<ChannelConfigInnerEntity>)
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
-    suspend fun insert(channelConfigEntity: ChannelConfigEntity)
+    suspend fun insertCommands(commands: List<CommandInnerEntity>)
```
zsmb13, line 35: These should never be called directly, outside of this Dao, right? We could make the Dao an abstract class and mark these methods as `protected` to avoid someone accidentally calling these in the future.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/ChannelConfigDao.kt
```diff
@@ -4,19 +4,37 @@ import androidx.room.Dao
...
 
 @Dao
 internal interface ChannelConfigDao {
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
-    suspend fun insertMany(channelConfigEntities: List<ChannelConfigEntity>)
+    @Transaction
+    suspend fun insert(channelConfigEntities: List<ChannelConfigEntity>) {
+        channelConfigEntities.map(ChannelConfigEntity::channelConfigInnerEntity).also { insertConfigs(it) }
+        channelConfigEntities.flatMap(ChannelConfigEntity::commands).also { insertCommands(it) }
```
zsmb13, line 19: Nit: I'd use `let` here to make use of the value, `also` is usually used when its return value is used in some way ("give me this object, but _also_ do stuff before that")

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/ChannelConfigEntity.kt
```diff
@@ -1,8 +1,56 @@
...
+)
+internal data class CommandInnerEntity(
+    val name: String,
+    val description: String,
+    val args: String,
+    val set: String,
+    val channelType: String,
+) {
+    @PrimaryKey
+    var id: Int = name.hashCode() + description.hashCode() + args.hashCode() + set.hashCode() + channelType.hashCode()
```
zsmb13, line 49: What's the purpose of using this combined hashcode value here? (As a separate thought, perhaps `hashCode()` itself could be used, which hashes the entire data class.)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/ChannelConfigEntity.kt
```diff
@@ -1,8 +1,56 @@
...
+    val isReadEvents: Boolean = false,
+    val isConnectEvents: Boolean = false,
+    val isSearch: Boolean = false,
+    val isReactionsEnabled: Boolean = false,
+    val isRepliesEnabled: Boolean = false,
+    val isMutes: Boolean = false,
+    val maxMessageLength: Int = Int.MAX_VALUE,
+    val automod: String = "",
+    val infinite: String = "",
+    val name: String = "",
```
zsmb13, line 26: Are all these default values ever used?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/ChannelConfigDao.kt
ogkuzmin: Yes, it does. It automatically fetches data from two tables and combine it to non-entity `ChannelConfigEntity` object

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/ChannelConfigEntity.kt
ogkuzmin: Nope. will remove it

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/ChannelConfigEntity.kt
ogkuzmin: Yes, you're right. Nice catch

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
ogkuzmin: For me, it is strange to have doubling here. Anyway if we had then it would get overridden in the Database. So I just changed the collection type to set.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/ChannelConfigDao.kt
ogkuzmin: Good idea

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
ogkuzmin: If there were any doubling we had to remove it before. When obtaining results. How do you can rely on an answer from BE and predict your app behaviour when you obtain data with doubling and store it in your DB?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/ChannelConfigDao.kt
zsmb13: Ah, that is awesome 👍

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/ChannelConfigDao.kt
ogkuzmin: Look at the latest commit. IDE suggested another :-)

zsmb13: Well, yes, that's the boring way I'd have gone with to begin with, but I wanted to give a suggestion given that you chose the functional style already 😄

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
ogkuzmin: Changed collection type to Set

zsmb13 (approved):
LGTM

PR #1280, opened 2021-01-28
Fix duplicate reactions

bychkovdmitry:
### Description

Motivation:
- `Message.addReaction` in `ClientExtensions` is flawed. For instance, user sends a reaction and `addReaction` method is invoked. In a moment, we receive `reaction.created` event and invoke this method for the second time. As a result, `ownReacitons` contains duplicate entries and counters don't show correct values. Basically this method should "upsert" reactions, which makes its implementation more complex.
- When reaction is sent to the server with `enforce_unique=true` flag, we receive `reaction.updated` event. However, the `enforce_unique=true` flag is not present in the event. Therefore, it is impossible to choose the right strategy to modify reactions of existing message (whether to remove other user reactions or not).

Prerequisites:
According to Mia and Ferhat, messages coming from `reaction.updated`, `reaction.created` and `reaction.deleted` events are fully enriched. It is more reliable to to update the whole message than reactions of a particular message with all the counters.

What's done:
- Renamed `addReaction` and `removeReaction` to `addMyReaction` and `removeMyReaction`. Did a small refactoring of these methods.
- `reaction.updated`, `reaction.created` and `reaction.deleted` now trigger full message update (just like `message.updated` event, excluding the logic of updating the last message)
- Reaction is sent to the server and saved locally with `score=1`, which is the default score the server will return with an event
- `ownReactions` from other user events (e.g. MessageUpdatedEvent) contain information about other user reactions. Had to use proper `ownReactions` of current user from corresponding message fetched from database
- Dropped some code in `ChannelControllerImplReactionsTest`. Sorry for that :)
- Didn't touch diff utils as the current implementation with `reactionCounts` and `reactionScores` seems to cover all possible cases

### Checklist

- [X] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
@@ -59,16 +59,15 @@ import io.getstream.chat.android.client.events.UsersUnmutedEvent
 import io.getstream.chat.android.client.extensions.enrichWithCid
 import io.getstream.chat.android.client.logger.ChatLogger
 import io.getstream.chat.android.client.models.ChannelUserRead
+import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.core.internal.exhaustive
-import io.getstream.chat.android.livedata.extensions.addReaction
-import io.getstream.chat.android.livedata.extensions.removeReaction
 import io.getstream.chat.android.livedata.extensions.setMember
 import io.getstream.chat.android.livedata.extensions.updateReads
 import kotlinx.coroutines.launch
 
 internal class EventHandlerImpl(
-    private val domainImpl: ChatDomainImpl
+    private val domainImpl: ChatDomainImpl,
 ) {
     private var logger = ChatLogger.get("EventHandler")
     private var firstConnect = true
@@ -180,7 +179,8 @@ internal class EventHandlerImpl(
                 is UserStartWatchingEvent,
                 is UserStopWatchingEvent,
                 is ChannelUserUnbannedEvent,
-                is MarkAllReadEvent -> Unit
+                is MarkAllReadEvent,
+                -> Unit
                 is ReactionNewEvent -> batchBuilder.addToFetchMessages(event.reaction.messageId)
                 is ReactionDeletedEvent -> batchBuilder.addToFetchMessages(event.reaction.messageId)
                 is ChannelMuteEvent -> batchBuilder.addToFetchChannels(event.channelMute.channel.cid)
@@ -209,19 +209,23 @@ internal class EventHandlerImpl(
                 // note that many of these events should also update user information
                 is NewMessageEvent -> {
                     event.message.enrichWithCid(event.cid)
+                    event.message.enrichWithOwnReactions(batch, event.user)
                     event.totalUnreadCount?.let { domainImpl.setTotalUnreadCount(it) }
                     batch.addMessageData(event.cid, event.message)
                 }
                 is MessageDeletedEvent -> {
                     event.message.enrichWithCid(event.cid)
+                    event.message.enrichWithOwnReactions(batch, event.user)
                     batch.addMessageData(event.cid, event.message)
                 }
                 is MessageUpdatedEvent -> {
                     event.message.enrichWithCid(event.cid)
+                    event.message.enrichWithOwnReactions(batch, event.user)
                     batch.addMessageData(event.cid, event.message)
                 }
                 is NotificationMessageNewEvent -> {
                     event.message.enrichWithCid(event.cid)
+                    event.message.enrichWithOwnReactions(batch, event.user)
                     event.totalUnreadCount?.let { domainImpl.setTotalUnreadCount(it) }
                     batch.addMessageData(event.cid, event.message)
                 }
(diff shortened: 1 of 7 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -209,19 +209,23 @@ internal class EventHandlerImpl(
...
                     batch.addMessageData(event.cid, event.message)
                 }
                 is MessageDeletedEvent -> {
                     event.message.enrichWithCid(event.cid)
+                    event.message.enrichWithOwnReactions(batch, event.user)
                     batch.addMessageData(event.cid, event.message)
                 }
                 is MessageUpdatedEvent -> {
                     event.message.enrichWithCid(event.cid)
+                    event.message.enrichWithOwnReactions(batch, event.user)
```
bychkovdmitry, line 223: When message is updated by other user, the event contains `ownReactions` of other user. Therefore, for such events we use previous `ownReactions` stored in the database

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -256,31 +260,19 @@ internal class EventHandlerImpl(
...
-                    // get the message, update the reaction data, update the message
-                    // note that we need to use event.reaction and not event.message
-                    // event.message only has a subset of reactions
-                    batch.getCurrentMessage(event.reaction.messageId)?.let {
-                        val updatedMessage = it.apply {
-                            addReaction(event.reaction, domainImpl.currentUser.id == event.user.id)
-                        }
-                        batch.addMessage(updatedMessage)
-                    }
+                    event.message.enrichWithCid(event.cid)
```
bychkovdmitry, line 263: We might have some issues with pagination if somebody updates very old message with "MessageUpdatedEvent", "ReactionNewEvent", etc and we persist the message from the event to our database.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -465,4 +458,12 @@ internal class EventHandlerImpl(
             queryRepo.handleEvents(sortedEvents)
         }
     }
+
+    private fun Message.enrichWithOwnReactions(batch: EventBatchUpdate, user: User) {
```
bychkovdmitry, line 462: The logic behind this:
If event is from current user then we are fine with `ownReactions` coming from event.
If event is from other user, then we replace other user's `ownReactions` with our own `ownReactions` from the database

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -1046,10 +1047,13 @@ internal class ChannelControllerImpl(
                 event.watcherCount?.let { setWatcherCount(it) }
             }
             is ReactionNewEvent -> {
-                upsertEventMessage(event.message)
+                upsertMessage(event.message)
```
bychkovdmitry, line 1050: We don't need to update the latest message so `upsertMessage ` instead of `upsertEventMessage `

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/extensions/MessageReactionExtensions.kt
```diff
@@ -0,0 +1,82 @@
...
+        if (newScore > 0) {
+            reactionScores[type] = newScore
+        } else {
+            reactionScores.remove(type)
+        }
+    }
+    ownReactions.clear()
+}
+
+private inline fun Message.updateReactions(actions: ReactionData.() -> Unit) {
```
bychkovdmitry, line 62: Some hacks for diffutil to see the different collection reference

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/controller/ChannelControllerImplReactionsTest.kt
```diff
@@ -47,28 +47,6 @@ internal class ChannelControllerImplReactionsTest {
             score = 234
         },
     )
```
bychkovdmitry, line 49: Removed otherReactions, because these methods are only used when current user creates/removes reaction

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/utils/TestDataHelper.kt
```diff
@@ -194,12 +194,24 @@ internal class TestDataHelper {
...
         cid = channel1.cid
         reactionScores = mutableMapOf("like" to 10)
-        reactionCounts = mutableMapOf("like" to 1); ownReactions =
-            mutableListOf(reaction1); latestReactions = mutableListOf(reaction1)
+        reactionCounts = mutableMapOf("like" to 1)
+        ownReactions = mutableListOf(reaction1)
+        latestReactions = mutableListOf(reaction1)
+    }
+
+    val reactionMessage2 = Message().apply {
```
bychkovdmitry, line 208: `ownReactions` contain own reactions of other user

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -296,6 +296,7 @@ public class MessageListViewModel @JvmOverloads constructor(
         val reaction = Reaction().apply {
             messageId = message.id
             type = reactionType
+            score = 1
```
bychkovdmitry, line 299: Set score to "1" so that locally saved reaction is consistent with the one coming from the server (on the server, default score is "1")

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/EventHandlerImpl.kt
```diff
@@ -465,4 +458,12 @@ internal class EventHandlerImpl(
...
+        if (domainImpl.currentUser.id != user.id) {
```
JcMinarro, line 463: Could we extract the `ownReactions` assignment out of the `if` statements?

JcMinarro (approved):
LGTM

adasiewiczr (approved):

PR #1283, opened 2021-01-28
CAS-535 Attachment reply option

samiuelson:
https://stream-io.atlassian.net/browse/CAS-535

### Description

Implement default attachment reply option handler

### Checklist

- [x] Reviewers added


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -367,7 +382,8 @@ public class MessageListViewModel @JvmOverloads constructor(
         public data class MuteUser(val user: User) : Event()
         public data class BlockUser(val user: User, val cid: String) : Event()
         public data class ReplyMessage(val cid: String, val repliedMessage: Message) : Event()
-        public data class AttachmentDownload(val attachment: Attachment) : Event()
```
zsmb13, line 370: I agree with renaming this, because it's named incorrectly, but it's been public API for two releases. We should probably leave it here as a deprecated class (with `replaceWith` to migrate to the new class), then we can clean it up later. Handling both events the same way should be easy.

zsmb13 (approved):

PR #1284, opened 2021-01-28
Creating a default gson instance

leandroBorgesFerreira:
### Description

This PR creates a default gson instance. As the instace of `gson` of `ChatDomainImpl` doesn't have a exclusion strategy, annotations like `@IgnoreDeserialisation` don't work with this instance. This create problems when setting the update state of an `Attachment` inside a `Message` so it can be upserted with this information. Without the correct exclusing strategy the app breaks with serialization/desarialization expcetion. 

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser/JsonUtils.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser/JsonUtils.kt
@@ -1,0 +1,39 @@
+package io.getstream.chat.android.client.parser
+
+import com.google.gson.ExclusionStrategy
+import com.google.gson.FieldAttributes
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+
+@InternalStreamChatApi
+public object JsonUtils {
+
+    public fun defaultStreamGson() : Gson {
+        return GsonBuilder()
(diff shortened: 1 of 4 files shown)


ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser/JsonUtils.kt
```diff
@@ -0,0 +1,39 @@
...
+import com.google.gson.ExclusionStrategy
+import com.google.gson.FieldAttributes
+import com.google.gson.Gson
+import com.google.gson.GsonBuilder
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+
+@InternalStreamChatApi
+public object JsonUtils {
+
+    public fun defaultStreamGson() : Gson {
```
zsmb13, line 12: This could be a lazy property within this class, so that there's literally just one instance of Gson in the app.

zsmb13 (approved):
Gson code that looks correct for now and should be removed soon if everything goes well - looks good to me 😄

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser/JsonUtils.kt
leandroBorgesFerreira: Makes sense, let's not have multiple instances of gson

ogkuzmin (approved):

PR #1293, opened 2021-01-29
Message entity storing refac #1

ogkuzmin:
As part of [CAS-552](https://stream-io.atlassian.net/browse/CAS-552?atlOrigin=eyJpIjoiNjNlMDYyMWUwYTcwNGE4NTg0NzczNmVlNWU4OTQxZGUiLCJwIjoiaiJ9)

### Description
The first iteration of message storing refactoring. 
1) Store Attachment as Entity with relation to MessageEntity
2) Remove from ChannelEntity embedded MessageEntity. Keep relation as a referenced id of the last message
3) Remove TypeConverter for Attachments

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/AttachmentEntity.kt
```diff
@@ -0,0 +1,53 @@
...
+    val image: String?,
+    val url: String?,
+    val name: String?,
+    val fallback: String?,
+    val uploadFilePath: String?,
+    @Embedded val uploadState: UploadStateEntity?,
+    val extraData: Map<String, Any>,
+) {
+    @PrimaryKey
+    var id: Int = hashCode()
```
bychkovdmitry, line 44: @PrimaryKey(autoGenerate = true) ?

ogkuzmin: Not sure. I consider Entity as a data class. So you can compare two instances by equal autogenerated code of kotlin data classes, so hashCode of a data class is a unique id, which is equal for two equal instances.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/dao/MessageDao.kt
```diff
@@ -4,55 +4,90 @@ import androidx.room.Dao
...
-            "WHERE stream_chat_message.id IN (:id)"
-    )
-    suspend fun select(id: String?): MessageEntity?
-
-    @Query(
-        "SELECT * FROM stream_chat_message " +
-            "WHERE stream_chat_message.syncStatus IN (:syncStatus) ORDER BY createdAt ASC"
-    )
-    suspend fun selectSyncNeeded(syncStatus: SyncStatus = SyncStatus.SYNC_NEEDED): List<MessageEntity>
+    @Transaction
```
bychkovdmitry, line 79: From the `Transaction` class docs:
"Putting this annotation on an {@link Insert}, {@link Update} or {@link Delete} method has no impact because those methods are always run inside a transaction. "

ogkuzmin: Yes, you're right. Nice catch 👍

bychkovdmitry (approved):

PR #1296, opened 2021-01-31
Show double tick mark when user message is read

bychkovdmitry:
### Description

In the old UI we used `MessageItem.messageReadBy` to show user avatar next to the last message read by this particular user. In the new UI, the use case is a bit different, we need an indication that message is read in every `MessageItem`. So, had to add `isMessageRead ` to `MessageItem`. Also, removed `messageReadBy` from the diff, as it is no longer used in our new UI.

### Checklist

- [X] Reviewers added

| Before | After |
| --- | --- |
| Screenshot 2021-02-01 at 00 41 30 | Screenshot 2021-02-01 at 00 38 06 |


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
```diff
@@ -41,6 +41,7 @@ public sealed class MessageListItem {
         val isMine: Boolean = false,
         val messageReadBy: List<ChannelUserRead> = listOf(),
         val isThreadMode: Boolean = false,
+        val isMessageRead: Boolean = true,
```
bychkovdmitry, line 44: Defaults to "true" as the most common case is when channel is read almost to the end

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
```diff
@@ -238,7 +238,41 @@ internal class MessageListItemLiveData(
...
+
+        return messages.map { messageListItem ->
+            if (messageListItem is MessageListItem.MessageItem) {
+                val isMessageRead = messageListItem.message
+                    .createdAt
+                    ?.let { it <= lastRead }
+                    ?: false
+
+                if (messageListItem.isMessageRead != isMessageRead) {
+                    messageListItem.copy(isMessageRead = isMessageRead)
```
bychkovdmitry, line 268: Typically only few copies here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/MessageListItemDiffCallback.kt
```diff
@@ -14,25 +13,16 @@ internal object MessageListItemDiffCallback : DiffUtil.ItemCallback<MessageListI
...
-                } else if (oldItem.message.syncStatus != newItem.message.syncStatus) {
-                    false
-                } else if (oldItem.message.deletedAt != newItem.message.deletedAt) {
-                    false
-                } else if (oldItem.positions != newItem.positions) {
-                    false
-                } else if (oldItem.messageReadBy.map(UserEntity::getUserId) != newItem.messageReadBy.map(UserEntity::getUserId)) {
-                    false
-                } else oldItem.isThreadMode == newItem.isThreadMode
+                oldItem.message.text == newItem.message.text &&
```
bychkovdmitry, line 16: Logical operations are short-circuit, previous optimization is not needed

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/MessageListItemDiffCallback.kt
```diff
@@ -14,25 +13,16 @@ internal object MessageListItemDiffCallback : DiffUtil.ItemCallback<MessageListI
...
+                    oldItem.message.reactionScores == newItem.message.reactionScores &&
+                    oldItem.message.reactionCounts == newItem.message.reactionCounts &&
+                    oldItem.message.attachments == newItem.message.attachments &&
+                    oldItem.message.replyCount == newItem.message.replyCount &&
+                    oldItem.message.syncStatus == newItem.message.syncStatus &&
+                    oldItem.message.deletedAt == newItem.message.deletedAt &&
+                    oldItem.positions == newItem.positions &&
+                    oldItem.isMessageRead == newItem.isMessageRead &&
```
bychkovdmitry, line 24: `messageReadBy` is ignored because it is not used in our UI

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/adapter/MessageListItem.kt
ogkuzmin: it's a change in public API. Should be stated in CHANGELOG

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
```diff
@@ -238,7 +238,41 @@ internal class MessageListItemLiveData(
             }
         }
 
-        return messagesCopy
+        return addMessageReadFlags(messagesCopy, reads)
```
ogkuzmin, line 241: I don't like that our logic becomes more and more complicated, loading UI thread. Seems right now we don't have any other choice. But in the future coroutine usage (with `Flow` for example) can help us with unloading UI thread.

ogkuzmin (commented):
LGTM, just update changelog

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
bychkovdmitry: Yep, but we'll have to change things a bit to avoid concurrency issues

ogkuzmin (approved):

PR #1299, opened 2021-02-01
Extend Moshi DTOs

zsmb13:
### Description

Moshi adventures, pt 3.

Adds all fields for Message DTOs, adding other necessary DTOs, new mapping code, and some bugfixes / code reorg along the way.

---

Planned next PR (and therefore not handled here yet):

- Expanding the User object
- Support for the custom object / extra data behaviour on User, Attachment, Reaction
- Sorting the fields within the DTOs by some logic (currently scrambled)

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/CustomObjectDtoAdapter.kt
```diff
@@ -68,10 +68,15 @@ internal abstract class CustomObjectDtoAdapter<Value : Any>(private val kClass:
...
         jsonWriter: JsonWriter,
-        value: Value,
+        value: Value?,
         mapAdapter: JsonAdapter<MutableMap<String, Any?>>,
         valueAdapter: JsonAdapter<Value>,
     ) {
+        if (value == null) {
+            jsonWriter.nullValue()
+            return
+        }
```
zsmb13, line 78: This was an unhandled case before, and caused a crash in tests.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/DtoMapper.kt
```diff
@@ -1,60 +0,0 @@
-package io.getstream.chat.android.client.api2
-
-import io.getstream.chat.android.client.api2.model.dto.DownstreamMessageDto
-import io.getstream.chat.android.client.api2.model.dto.UpstreamMessageDto
-import io.getstream.chat.android.client.api2.model.dto.UserDto
-import io.getstream.chat.android.client.models.Message
-import io.getstream.chat.android.client.models.User
-
-internal class DtoMapper {
```
zsmb13, line 9: Split out into several files as extension functions

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/MessageMapping.kt
```diff
@@ -0,0 +1,69 @@
...
+        text = text,
+        html = html,
+        parentId = parent_id,
+        command = command,
+        user = user.toDomain(),
+        replyCount = reply_count,
+        reactionCounts = reaction_counts.toMutableMap(),
+        reactionScores = reaction_scores.toMutableMap(),
+        latestReactions = latest_reactions.mapTo(mutableListOf(), ReactionDto::toDomain),
+        ownReactions = own_reactions.mapTo(mutableListOf(), ReactionDto::toDomain),
```
zsmb13, line 50: Some of the `MutableMap` and `MutableList` types in the domain models are really ugly, so we should eventually change them. For the time being though, we can deal with them just fine here with a bit of extra code.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UpstreamMessageDto.kt
```diff
@@ -1,17 +1,35 @@
 package io.getstream.chat.android.client.api2.model.dto
 
 import com.squareup.moshi.JsonClass
+import java.util.Date
 
 @JsonClass(generateAdapter = true)
 internal data class UpstreamMessageDto(
     val id: String,
     val cid: String,
+    val parent_id: String?,
```
leandroBorgesFerreira, line 10: I think that adding @Json(name = "[field name]") to all fields woudl be nice, so we prevent proguard problems. Just making sure that the `[...]/api2/model/dto` is in the proguard rules works too.

leandroBorgesFerreira (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UpstreamMessageDto.kt
leandroBorgesFerreira: I prefer doing both =P

zsmb13: We don't have these annotations on purpose, see https://github.com/GetStream/stream-chat-android/pull/1274#discussion_r564593801

As for Proguard: Moshi's generated adapters contain the string names of the fields as strings, so obfuscation shouldn't affect them at all.

Here's an example of the trie that Moshi uses for figuring out which field to parse being created in the generated code: 
image

leandroBorgesFerreira: Makes sense. Solving the conversation.

adasiewiczr (approved):

PR #1300, opened 2021-02-01
Message entity refac2

ogkuzmin:
Part2. Message Entity DB refac

### Description

Refac relation ReactionEntity and MessageEntity by native for SQLite mechanism of foreign key

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
```diff
@@ -12,6 +12,12 @@ internal data class MessageEntity(
     @Embedded val messageInnerEntity: MessageInnerEntity,
     @Relation(entity = AttachmentEntity::class, parentColumn = "id", entityColumn = "messageId")
     val attachments: List<AttachmentEntity>,
+    /** the reactions from the current user */
+    @Relation(entity = ReactionEntity::class, parentColumn = "id", entityColumn = "messageId")
+    val ownReactions: List<ReactionEntity> = emptyList(),
+    /** the last 5 reactions on this message */
```
adasiewiczr, line 18: It's just copy-paste but I think that `latestReactions` can contain more than 5 reactions

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ReactionRepository.kt
```diff
@@ -16,6 +18,7 @@ internal class ReactionRepository(private val reactionDao: ReactionDao) {
         insert(listOf(reaction.toEntity()))
     }
 
+    @VisibleForTesting
     internal suspend fun insert(reactions: List<Reaction>) {
```
adasiewiczr, line 22: Are we going to remove that at some point?

adasiewiczr (approved):
LGTM

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/entity/MessageEntity.kt
ogkuzmin: I'm not sure about logic of reaction domain. We can remove this if it's not true. Just as you mentioned I copied it

adasiewiczr: @bychkovdmitry can you confirm that?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ReactionRepository.kt
ogkuzmin: Yes, since there is no point to insert a list directly. Now standard mechanism of inserting reactions is inserting in a transaction with a message insertion. Only single direct reaction makes sense (when you make a new reaction and update offline storage)

PR #1303, opened 2021-02-02
CAS-546 AvatarView border improvement

samiuelson:
https://stream-io.atlassian.net/browse/CAS-546
### Description
In case the `AvatarView` displays the border the offset (padding) is set to the `ImageView` to avoid overlapping the image with the border stroke.

| Before | After |
| --- | --- |
| before | after |

### Checklist

- [x] Reviewers added


leandroBorgesFerreira (approved):
🚢

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
```diff
@@ -79,10 +79,12 @@ public class AvatarView : AppCompatImageView {
         setStyle(AvatarStyle(context, attrs))
     }
 
-    internal fun setStyle(avatarStyle: AvatarStyle) {
+    private fun setStyle(avatarStyle: AvatarStyle) {
         this.avatarStyle = avatarStyle
         borderPaint.color = avatarStyle.avatarBorderColor
         borderPaint.strokeWidth = avatarStyle.avatarBorderWidth.toFloat()
+        val borderWidth = this.avatarStyle.avatarBorderWidth
+        setPadding(borderWidth, borderWidth, borderWidth, borderWidth)
```
zsmb13, line 87: Perhaps this padding should be a tiny tiny tiny bit (1px?) less than the stroke width, so that we don't get any light/white pixels where the avatar and the border meet

PR #1307, opened 2021-02-02
[CAS-552] Channel entity isolation

ogkuzmin:
### Description

**Goal:** "Isolate ChannelEntity usage in ChannelDao/ChannelRepository only"

**What done:**
1) Remove usage ChannelEntity from business logic
2) Rewrite ChannelRepository methods in order to use only models
3) Move logic from ChannelEntity to RepositoryHelper
4) Add update methods to DAO 
5) Refactor ChannelEntity structure. Remove unused field `mutedTill`
6) Add several tests to cover new logic in RepositoryHelper

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -13,6 +13,7 @@ import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.api.models.QueryChannelsRequest
 import io.getstream.chat.android.client.api.models.QuerySort
 import io.getstream.chat.android.client.call.Call
+import io.getstream.chat.android.client.call.await
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.events.ChatEvent
 import io.getstream.chat.android.client.events.MarkAllReadEvent
@@ -688,7 +689,20 @@ internal class ChatDomainImpl internal constructor(
     }
 
     private suspend fun retryChannels(): List<ChannelEntity> {
-        return repos.channels.retryChannels()
+        val channelEntities = repos.channels.selectSyncNeeded()
+
+        for (channel in channelEntities) {
+            val members = channel.members.keys.toList()
+            val result = client.createChannel(channel.type, channel.channelId, members, channel.extraData).await()
+            if (result.isSuccess) {
+                channel.syncStatus = SyncStatus.COMPLETED
+                repos.channels.insert(channel)
+            } else if (result.isError && result.error().isPermanent()) {
+                channel.syncStatus = SyncStatus.FAILED_PERMANENTLY
+                repos.channels.insert(channel)
+            }
+        }
+        return channelEntities
     }
 
     @VisibleForTesting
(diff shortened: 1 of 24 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ChannelRepository.kt
```diff
@@ -1,85 +1,73 @@
...
-        for (cid in channelCIDs) {
-            val channelEntity = channelCache.get(cid)
-            channelEntity?.let { cachedChannels.add(it) }
-        }
+    suspend fun select(
+        channelCIDs: List<String>,
+        getUser: suspend (userId: String) -> User,
+        getMessage: suspend (messageId: String) -> Message?,
+    ): List<Channel> {
+        val cachedChannels: MutableList<Channel> = channelCIDs.mapNotNullTo(mutableListOf()) { channelCache.get(it) }
```
leandroBorgesFerreira, line 54: You can use just `channelCache::get` here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/RepositoryHelper.kt
```diff
@@ -94,6 +99,33 @@ internal class RepositoryHelper(
...
+
+    internal suspend fun updateLastMessageForChannel(cid: String, lastMessage: Message) {
+        selectChannelWithoutMessages(cid)?.also { channel ->
+            val messageCreatedAt = checkNotNull(
+                lastMessage.createdAt
+                    ?: lastMessage.createdLocallyAt
+            ) { "created at cant be null, be sure to set message.createdAt" }
+
+            val oldLastMessage = channel.lastMessage
+            val updateNeeded = if (oldLastMessage != null) {
```
leandroBorgesFerreira, line 114: you can use only boolean logic here, no need to the if/else.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -680,10 +683,33 @@ internal class ChatDomainImpl internal constructor(
...
+                channel.members.map(UserEntity::getUserId),
+                channel.extraData
+            ).await()
+
+            when {
+                result.isSuccess -> {
+                    channel.syncStatus = SyncStatus.COMPLETED
+                    repos.insertChannel(channel)
+                }
+                result.isError && result.error().isPermanent() -> {
```
leandroBorgesFerreira, line 707: What happens if the error is not permanent?

leandroBorgesFerreira (approved):
looks good

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
ogkuzmin: Feels like don't update right now these failed channels. I'm not an author of this logic. Just copied as it was [before](https://github.com/GetStream/stream-chat-android/blob/033455e584b93b6adadbce03e004df9256b68d9e/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ChannelRepository.kt#L78)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ChannelRepository.kt
ogkuzmin: Thanks! Nice, love fun references, seems more elegant for me

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/RepositoryHelper.kt
ogkuzmin: Yes, you're right but then the condition will look so messy, I'd like to keep it as is

PR #1309, opened 2021-02-02
[CAS-245] Adding InProgress as status before view type is requested

leandroBorgesFerreira:
[CAS-245](https://stream-io.atlassian.net/browse/CAS-245)

### Description

Fixing crash in the progress attachment. If the `uploadState` for the `attachments` if set after the `getMessageType`, it doesn't become ephemeral and then it can't be canceled. 

### Checklist

-  ~Changelog updated with client-facing changes~
-  ~New code is covered by unit tests~
- [x] Reviewers added


ogkuzmin (commented):
What about to add test for this specific case?

leandroBorgesFerreira:
Now this case is covered in 
`Attachments should be sent with success as uploadState when request success`
and in 
`Message with failed attachment upload should be upload send the right state`

ogkuzmin (approved):

PR #1311, opened 2021-02-02
[CAS-557] Visibility for config repository

leandroBorgesFerreira:
[CAS-557](https://stream-io.atlassian.net/browse/CAS-557)

### Description

Changing visibility of `RepositoryHelper.configs` and creating necessary methods. 

### Checklist

- ~New code is covered by unit tests~
- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -754,7 +754,7 @@ internal class ChatDomainImpl internal constructor(
         }
 
         // store the channel configs
-        repos.configs.insert(configs)
+        repos.configChannel(configs)
```
ogkuzmin, line 757: Why not `insertChannelConfig`?

leandroBorgesFerreira: My though is that this method is used to configure the channel, that fact that the configuration is stored can be extracted away. But no problems with `insertChannelConfig` either, what do you think?

ogkuzmin: I consider RepositoryHelper just as a database wrapper. ChatDomain is a key for logic. So I think straightforward insert is better than

leandroBorgesFerreira: Changed

ogkuzmin (approved):

PR #1314, opened 2021-02-02
[CAS-653] Feature/dev token

JcMinarro:
### Description
Move `devToken()` methiod into `ChatClient`

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1131,6 +1132,8 @@ public class ChatClient internal constructor(
     private fun isValidRemoteMessage(remoteMessage: RemoteMessage): Boolean =
         notifications.isValidRemoteMessage(remoteMessage)
 
+    public fun devToken(userId: String): String = ChatUtils.devToken(userId)
```
zsmb13, line 1135: Does it make any sense for this to be an instance method? We could also consider doing something like the new Swift SDK, where they have a `TokenProvider.development(userId)` call to create a dev token, [see here](https://github.com/GetStream/stream-chat-swift/blob/8f72ede75c64fd65e3513251d2767e968fbe63f9/Sources/StreamChat/Config/TokenProvider.swift#L23).

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ChatUtils.kt
```diff
@@ -3,9 +3,9 @@ package io.getstream.chat.android.client.utils
 import android.util.Base64
 import java.nio.charset.StandardCharsets
 
-public object ChatUtils {
+internal object ChatUtils {
```
zsmb13, line 6: We could just deprecate this class at first, instead of breaking public API immediately

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: It is a decision that has been taken by the iOS team, but it has not been accepted yet.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ChatUtils.kt
JcMinarro: This method is rarelly used (Only works for development apps, where fixed some weeks ago and we didn't receive any feedback about it), We don't need to deprecate it

zsmb13: In that case, can we just remove this file and move its contents into the client method?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
zsmb13: Alright, let's go with this for now then. Feels a bit weird to make a `client.setUser(client.devToken())` kind of call, but alright.

zsmb13 (approved):
Thanks for cleaning up its implementation too 👍

PR #1317, opened 2021-02-03
Add docs snippets for slow mode

zsmb13:
### Description

Adds documentation snippets for https://getstream.io/chat/docs/android/slow_mode/?language=kotlin

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/Channels.java
+++ b/stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/Channels.java
@@ -1,6 +1,7 @@
 package io.getstream.chat.docs.java;
 
 import android.os.Handler;
+import android.os.Looper;
 import android.util.Log;
 
 import org.jetbrains.annotations.NotNull;
@@ -594,7 +595,9 @@ public class Channels {
                         // After sending a message, block the UI temporarily
                         // The disable/enable UI methods have to be implemented by you
                         disableMessageSendingUi();
-                        new Handler().postDelayed(() -> enableMessageSendingUi(), cooldown);
+
+                        new Handler(Looper.getMainLooper())
+                                .postDelayed(() -> enableMessageSendingUi(), cooldown);
                     });
                 }
             });
--- a/stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/Channels.java
+++ b/stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/Channels.java
@@ -1,5 +1,6 @@
 package io.getstream.chat.docs.java;
 
+import android.os.Handler;
 import android.util.Log;
 
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/Channels.java
```diff
@@ -555,4 +555,49 @@ public void queryingMembers() {
...
+                    final Channel channel = channelResult.data();
+                    int cooldown = channel.getCooldown();
+
+                    Message message = new Message();
+                    message.setText("Hello");
+                    channelClient.sendMessage(message).enqueue((messageResult) -> {
+                        // After sending a message, block the UI temporarily
+                        // The disable/enable UI methods have to be implemented by you
+                        disableMessageSendingUi();
+                        new Handler().postDelayed(() -> enableMessageSendingUi(), cooldown);
```
zsmb13, line 597: `Handler` seemed like the most standard thing to use here

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Channels.kt
```diff
@@ -529,4 +530,45 @@ class Channels(val client: ChatClient, val channelClient: ChannelClient) {
...
+                    val channel = result.data()
+                    val cooldown = channel.cooldown
+
+                    val message = Message(text = "Hello")
+                    channelClient.sendMessage(message).enqueue {
+                        // After sending a message, block the UI temporarily
+                        // The disable/enable UI methods have to be implemented by you
+                        disableMessageSendingUi()
+                        Handler().postDelayed(::enableMessageSendingUi, cooldown.toLong())
+                    }
```
zsmb13, line 569: We could easily use `await()` and `delay()` in Kotlin snippets - we'll have to make a decision about this at some point, for now we have `enqueue` and callback lambdas everywhere

ogkuzmin: Maybe put handler as a class level property?
``` kotlin
private val uiHandler = Handler(Looper.getMainLooper())
```

zsmb13: Things outside the method won't be copied to the docs, so I'd rather create it here, otherwise it'll be a mystery where that variable comes from.

For the threading part, since `enqueue` runs the callback on the UI thread, the Handler will be on the UI thread too.

ogkuzmin: I understand why usage `Handler` in enqueue is not dangerous. It's just not obvious and can lead to inappropriate usage in `callback` methods for example

zsmb13: Made the loopers explicit

JcMinarro: I see it simple enough
Adding `await()` and `dealay()` will forze you to use suspend functions that could be a bit confuse when browsing our documentation

JcMinarro (approved):

PR #1323, opened 2021-02-03
Fix showing GIFs

### LINKED ISSUE #1315: Showing and playing gifs
**Describe the bug**
Showing and playing gifs is not working properly.

**SDK version**
- 4.5.0  
- 4.5.2

**To Reproduce**
1. Go to 'channel'
2. Send 'gif'

**Expected behavior**
Preview in the message list and playing gif in full screen or preview.

**Device:**
 - Pixel 2 API 29 Emulator
 - Xiaomi Redmi Note 5

https://user-images.githubusercontent.com/31736480/106709863-2ade9800-6606-11eb-9789-241c93542bc1.mp4

adasiewiczr:
GIFs weren't appropriately rendered because `PorterImageView` usage. It turned out that `PorterImageView` renders drawable only if the view needs to be [invalidated](https://github.com/GetStream/stream-chat-android/blob/develop/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/roundedImageView/PorterImageView.java#L129) (based on `invalidate` flag). The flag is changed to [false](https://github.com/GetStream/stream-chat-android/blob/develop/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/roundedImageView/PorterImageView.java#L109) when drawable to render is non-null, but the main problem is that the flag can be changed during `onDraw` method invocation (that happens every time when [invalidate](https://github.com/GetStream/stream-chat-android/blob/develop/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/roundedImageView/PorterImageView.java#L70) method is called) :(
The problem was resolved by introducing `invalidate` flag local copy, but this is rather a workaround and we should think about replacing`PorterImageView`.
PR also fixes displaying GIFs after clicking on attachment:
1. It turned out that Giphy attachment contains only `thumbURL`
2. We don't support animated images inside `ImageViewer`

| Before | After |
| --- | --- |
|Screenshot 2021-02-03 at 17 02 46|https://user-images.githubusercontent.com/17440581/106773905-de18b280-6641-11eb-9328-3eea1ab53d9d.mp4|

closes #1315 

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/navigation/destinations/AttachmentDestination.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/navigation/destinations/AttachmentDestination.kt
@@ -49,7 +49,7 @@ public open class AttachmentDestination(
                 }
             }
             ModelType.attach_video -> url = attachment.assetUrl
-            ModelType.attach_giphy -> url = attachment.assetUrl
+            ModelType.attach_giphy -> url = attachment.thumbUrl
             ModelType.attach_product -> url = attachment.url
         }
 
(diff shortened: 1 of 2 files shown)


ogkuzmin (approved):
Good catch! I think it was hard to investigate 👍

PR #1326, opened 2021-02-03
Message item attributes

carterhudson:
## Issue
https://stream-io.atlassian.net/browse/CAS-650

## Description

Supports tutorial content & goals by specifying a minimum viable subset of customizable attributes for message list items, specifically:
- background color
- text color
- link color

The current desired behavior for when a user specifies a background or text color for message items is to make _every_ item adopt that background color uniformly. However, our default design calls for "link" items to have a different background color than plain text items. If we used the default values supplied to our style class, without having separate attributes / properties for "link bg color," we wouldn't be able to keep them different colors as a default. I know we're still nailing down which attributes we want to support and expose, so, for now, if nothing is specified in the `MessageListView`'s attributes, a value of `VALUE_NOT_SET` (`Integer.MAX_VALUE`) is set as the a default value for the `defaultValue` parameter. It's coerced to `null` when `build()` is called. This lets the decorators specify their own default values for items when style values are `null`.

SO PRETTY 😍 :
Screen Shot 2021-02-04 at 1 02 39 PM

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListItemStyle.kt
```diff
@@ -0,0 +1,97 @@
...
+    @ColorInt val messageLinkTextColorTheirs: Int?,
+) : Serializable {
+
+    internal companion object {
+        internal const val VALUE_NOT_SET = Integer.MAX_VALUE
+    }
+
+    internal class Builder(private val attributes: TypedArray) {
+        @ColorInt
+        private var messageBackgroundColorMine: Int = 0
```
zsmb13, line 23: Shouldn't these have `VALUE_NOT_SET` as their default values? Looks like they'd remain `0` if we forget to call one of the builder methods.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListItemStyle.kt
```diff
@@ -0,0 +1,97 @@
...
+                messageTextColorMine = messageTextColorMine.nullIfNotSet(),
+                messageTextColorTheirs = messageTextColorTheirs.nullIfNotSet(),
+                messageLinkTextColorMine = messageLinkTextColorMine.nullIfNotSet(),
+                messageLinkTextColorTheirs = messageLinkTextColorTheirs.nullIfNotSet(),
+            )
+        }
+
+        private fun Int.nullIfNotSet(): Int? {
+            return if (this == VALUE_NOT_SET) null else this
+        }
```
zsmb13, line 95: Really neat solution!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -86,7 +87,8 @@ public class MessageListView : ConstraintLayout {
         private const val LOAD_MORE_THRESHOLD = 10
     }
 
-    private lateinit var style: MessageListViewStyle
+    private lateinit var messageListViewStyle:
+        MessageListViewStyle
```
zsmb13, line 91: Really weird line break here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -225,7 +228,8 @@ public class MessageListView : ConstraintLayout {
                         message,
                         messageOptionsConfiguration.copy(
                             threadEnabled = !adapter.isThread && !message.isInThread(),
-                        )
+                        ),
+                        messageListViewStyle.itemStyle
```
zsmb13, line 232:
```suggestion
                        messageListViewStyle.itemStyle,
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -637,8 +641,10 @@ public class MessageListView : ConstraintLayout {
         messageListItemViewHolderFactory.decoratorProvider = MessageListItemDecoratorProvider(
             currentUser = currentUser,
             dateFormatter = messageDateFormatter,
-            isDirectMessage = channel.isDirectMessaging()
+            isDirectMessage = channel.isDirectMessaging(),
+            messageListViewStyle.itemStyle
```
zsmb13, line 645:
```suggestion
            messageListViewStyle.itemStyle,
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/MessageListItemDecoratorProvider.kt
```diff
@@ -13,15 +13,19 @@ import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.LinkAt
...
 import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.ReactionsDecorator
 import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.ReplyDecorator
+import io.getstream.chat.android.ui.messages.adapter.viewholder.decorator.TextDecorator
+import io.getstream.chat.android.ui.messages.view.MessageListItemStyle
 
 internal class MessageListItemDecoratorProvider(
     currentUser: User,
     dateFormatter: DateFormatter,
     isDirectMessage: Boolean,
+    style: MessageListItemStyle,
```
ogkuzmin, line 23: Isn't better to pass and keep style attributes in VHs? And provider could take those values from VHs directly

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/TextDecorator.kt
```diff
@@ -0,0 +1,80 @@
...
+
+        val textColorTheirs = style.messageTextColorTheirs ?: defaultTextColor
+
+        val linkColorTextMine = style.messageLinkTextColorMine ?: defaultLinkTextColor
+
+        val linkColorTextTheirs = style.messageLinkTextColorTheirs ?: defaultLinkTextColor
+
+        val textColor = if (isMine) textColorMine else textColorTheirs
+
+        val linkColor = if (isMine) linkColorTextMine else linkColorTextTheirs
```
zsmb13, line 35: We compute more values here than we'll use, I'd go with

```suggestion
        val textColor = (if (isMine) style.messageTextColorMine else style.messageTextColorTheirs) 
            ?: ContextCompat.getColor(textView.context, R.color.stream_ui_black)
        val linkColor = (if (isMine) style.messageLinkTextColorMine else style.messageLinkTextColorTheirs)
            ?: ContextCompat.getColor(textView.context, R.color.stream_ui_accent_blue)
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/BackgroundDecorator.kt
```diff
@@ -126,22 +133,36 @@ internal class BackgroundDecorator : BaseDecorator() {
...
+                // however, if a user has specified a background color attribute, we use it for _all_ message backgrounds.
+                val myLinkBackgroundColor =
+                    style.messageBackgroundColorMine ?: ContextCompat.getColor(view.context, MESSAGE_LINK_BACKGROUND)
+
+                val myRegularBackgroundColor = style.messageBackgroundColorMine ?: ContextCompat.getColor(
+                    view.context,
+                    MESSAGE_CURRENT_USER_BACKGROUND
                 )
+
+                val backgroundTintColor = if (hasLink) myLinkBackgroundColor else myRegularBackgroundColor
```
zsmb13, line 146:
```suggestion
                val backgroundTintColor = if (hasLink) {
                    style.messageBackgroundColorMine
                        ?: ContextCompat.getColor(view.context, MESSAGE_LINK_BACKGROUND)
                } else {
                    style.messageBackgroundColorMine
                        ?: ContextCompat.getColor(view.context, MESSAGE_CURRENT_USER_BACKGROUND)
                }
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/TextDecorator.kt
```diff
@@ -0,0 +1,80 @@
...
+import io.getstream.chat.android.ui.messages.adapter.viewholder.PlainTextWithMediaAttachmentsViewHolder
+import io.getstream.chat.android.ui.messages.view.MessageListItemStyle
+
+internal class TextDecorator(val style: MessageListItemStyle) : BaseDecorator() {
+
+    private fun decorateMessageText(
+        textView: TextView,
+        isMine: Boolean,
+    ) {
+        val defaultTextColor = ContextCompat.getColor(textView.context, R.color.stream_ui_black)
```
ogkuzmin, line 21: If you keep those attributes in VHs, you could just provide the required color as an argument to this method

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/BackgroundDecorator.kt
```diff
@@ -126,22 +133,36 @@ internal class BackgroundDecorator : BaseDecorator() {
...
+
+                val theirRegularBackgroundColor =
+                    style.messageBackgroundColorTheirs ?: ContextCompat.getColor(
                         view.context,
-                        if (hasLink) MESSAGE_LINK_BACKGROUND else MESSAGE_OTHER_USER_BACKGROUND
+                        MESSAGE_OTHER_USER_BACKGROUND
                     )
-                )
+
+                val backgroundTintColor = if (hasLink) theirLinkBackgroundColor else theirRegularBackgroundColor
```
zsmb13, line 163:
```suggestion
                val backgroundTintColor = if (hasLink) {
                    style.messageBackgroundColorTheirs
                        ?: ContextCompat.getColor(view.context, MESSAGE_LINK_BACKGROUND)
                } else {
                    style.messageBackgroundColorTheirs
                        ?: ContextCompat.getColor(view.context, MESSAGE_OTHER_USER_BACKGROUND)
                }
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/MessageListItemDecoratorProvider.kt
carterhudson: I had it that way originally, but it seemed to me that decorators should be more concerned with keeping a style and how to apply it to each view holder, rather than having the view holder hold the style and give it to a decorator, so the decorator can then alter the view holder 🤔 What do you think?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/BackgroundDecorator.kt
carterhudson: Sure, works for me

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/TextDecorator.kt
carterhudson: I supposed that's true - just doesn't seem to be the domain of the VH to do that since we delegate all of the styling to decorators. The background decorator, for example, contains default colors in a companion.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/viewholder/decorator/TextDecorator.kt
carterhudson: I was attempting to make it easy to read, but I see your point. I'll see what I can do.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListItemStyle.kt
carterhudson: Oo, good catch. They definitely should.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListItemStyle.kt
carterhudson: Thanks! I was wondering if I made this more complicated than it needed to be, but it was the only way I could figure to do it until (if) we support all the other attributes.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
carterhudson: Haha good catch. You've got some eagle eyes!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
carterhudson: Should these be added automatically when I cmd+option+l? It does add some in the signatures, but not the calls 🤔

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/adapter/MessageListItemDecoratorProvider.kt
ogkuzmin: I think decorators only know how to do some decorations (setting background for views, applying max possible width and so on. They were introduced for avoiding code doublings for every type of VH), but ViewHolders hold/contain attributes and views for that. It's just my opinion and my way of thinking. If you see your solution as more concise and useful I don't mind anyway.

ogkuzmin (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
zsmb13: The formatter isn't quite as aggressive with them as it should be 😄

zsmb13 (approved):
Some things are not quite perfect, but I think this is good to go for now, and we can polish it later.

Screenshot 2021-02-05 at 11 47 11
Screenshot 2021-02-05 at 11 46 58

PR #1328, opened 2021-02-04
[CAS-558] Hide channel repository

ogkuzmin:
### Description

Make `channelRepository` the private field in `RepositoryHelper`

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ChannelRepository.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ChannelRepository.kt
@@ -71,7 +71,7 @@ internal class ChannelRepository(
         channelDao.setDeletedAt(cid, deletedAt)
     }
 
-    internal suspend fun setHiddenForChannel(cid: String, hidden: Boolean, hideMessagesBefore: Date?) {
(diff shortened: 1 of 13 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/ChannelRepository.kt
```diff
@@ -70,4 +70,14 @@ internal class ChannelRepository(
         channelCache.remove(cid)
         channelDao.setDeletedAt(cid, deletedAt)
     }
+
+    internal suspend fun setHiddenForChannel(cid: String, hidden: Boolean, hideMessagesBefore: Date?) {
```
adasiewiczr, line 74: Does `hideMessagesBefore` need to be nullable? If yes maybe we can join those two methods into one with default parameter?

ogkuzmin: Considering the data structure of the model it could be nullable, but in terms of an update, it doesn't make sense. You're right, I make argument non-nullable

adasiewiczr (approved):
LGTM

PR #1330, opened 2021-02-04
[CAS-640] Rename setUser to connectUser

zsmb13:
https://stream-io.atlassian.net/browse/CAS-640

### Description

Deprecates `setUser` (and similar methods), introduces new `connectUser` (and similar) methods instead. The new methods return `Call` instances instead of accepting a callback as their parameter, so clients can either `enqueue` or `await` them.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/androidTest/java/io/getstream/chat/android/client/utils/TestChatUtils.kt
```diff
@@ -1,20 +0,0 @@
-package io.getstream.chat.android.client.utils
-
-import org.assertj.core.api.Assertions.assertThat
-import org.assertj.core.api.Assertions.assertThatThrownBy
-import org.junit.Test
-
-internal class TestChatUtils {
```
zsmb13, line 7: We have tests for `devToken` elsewhere (plus it's not even in `ChatUtils` anymore, that class doesn't exist now)

stream-chat-android-client/src/androidTest/java/io/getstream/chat/android/client/utils/TestInitCallback.kt
```diff
@@ -0,0 +1,26 @@
+package io.getstream.chat.android.client.utils
+
+import io.getstream.chat.android.client.call.Call
+import io.getstream.chat.android.client.errors.ChatError
+import io.getstream.chat.android.client.socket.ConnectionData
+
+internal class TestInitCallback : Call.Callback<ConnectionData> {
```
zsmb13, line 7: Replaces `TestInitListener`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ClientExtensions.kt
```diff
@@ -74,36 +66,3 @@ public inline fun <reified T : ChatEvent> ChatClient.subscribeForSingle(
...
-/**
- * Runs [ChatClient.setUser] in a suspending way.
- * Throws exceptions if errors occur during the call.
- */
-public suspend fun ChatClient.setUserAndAwait(
-    user: User,
-    tokenProvider: TokenProvider,
-): InitConnectionListener.ConnectionData = suspendCoroutine { cont ->
-    setUser(user, tokenProvider, initConnectionListener(cont))
-}
```
zsmb13, line 98: This was brand new API, and is no longer needed with `connectUser` returning a `Call`

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
```diff
@@ -71,8 +71,11 @@ public interface Call<T : Any> {
  *
  * Does not throw exceptions. Any errors will be wrapped in the [Result] that's returned.
  */
-public suspend fun <T : Any> Call<T>.await(): Result<T> =
-    suspendCancellableCoroutine { continuation ->
+public suspend fun <T : Any> Call<T>.await(): Result<T> {
+    if (this is CoroutineCall<T>) {
+        return this.runnable.invoke()
+    }
```
zsmb13, line 77: To avoid calling `enqueue` below which then internally would just kick off yet another coroutine in the case of a `CoroutineCall`...

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/InitConnectionListener.kt
```diff
@@ -13,3 +17,5 @@ public abstract class InitConnectionListener {
 
     public data class ConnectionData(val user: User, val connectionId: String)
 }
+
+public data class ConnectionData(val user: User, val connectionId: String)
```
zsmb13, line 21: Duplicate of the class nested in `InitConnectionListener`, so that we can more easily remove the entire `InitConnectionListener` later.

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
```diff
@@ -71,8 +71,11 @@ public interface Call<T : Any> {
  *
  * Does not throw exceptions. Any errors will be wrapped in the [Result] that's returned.
  */
-public suspend fun <T : Any> Call<T>.await(): Result<T> =
-    suspendCancellableCoroutine { continuation ->
+public suspend fun <T : Any> Call<T>.await(): Result<T> {
+    if (this is CoroutineCall<T>) {
+        return this.runnable.invoke()
```
ogkuzmin, line 76: Not very clear name for `runnable`. Firstly, I thought it is java's `Runnable` and such invoke will blocking. But then looked at implementation details 🙂

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
```diff
@@ -126,31 +122,26 @@ internal class OfflineSyncFirebaseMessagingService : FirebaseMessagingService()
...
-                        continuation.resume(client)
-                    }
-
-                    override fun onError(error: ChatError) {
-                        val cause = error.cause ?: IllegalStateException(error.message)
-                        continuation.resumeWithException(cause)
-                    }
-                }
-            )
+        val notificationConfig = syncModule.notificationConfigStore.get()
```
ogkuzmin, line 127: Wow it looks much better now 👍

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/KeystrokeImplTest.kt
```diff
@@ -14,10 +14,8 @@ internal class KeystrokeImplTest : BaseConnectedIntegrationTest() {
     fun keystroke() = runBlockingTest {
         var channelState = chatDomain.useCases.watchChannel(data.channel1.cid, 10).execute().data()
         val result = chatDomain.useCases.keystroke(data.channel1.cid).execute()
-        advanceUntilIdle()
```
ogkuzmin, line 17: Glad to see less black test magic 😄

ogkuzmin (commented):
Nice job! API is getting more concise and clean.
Just small comments

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/InitConnectionListener.kt
JcMinarro: Shouldn't we move it out of this file?
Maybe insdide of `models` package?

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
zsmb13: `task`? `block`? `stuffToRunInACoroutine`?

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/utils/LiveDataTestUtil.kt
```diff
@@ -13,24 +11,12 @@ internal suspend fun waitForSetUser(
...
-            }
-
-            override fun onSuccess(data: ConnectionData) {
-                super.onSuccess(data)
-                lock.complete(Unit)
-            }
-        }
-    )
+    client.connectUser(user, token).enqueue {
+        lock.complete(Unit)
```
JcMinarro, line 18: Now that we have a `Call` here, we don't need the "lock" anymore.
The `delay()` and `withTimeout()` maybe is not needed anymore neither

JcMinarro: You will need to use `execute()` to block the call, instead of `enqueue()` if we want to remove the old implementation

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -166,10 +225,31 @@ public class ChatClient internal constructor(
...
+     *
+     * This method performs required operations before connecting with the Stream API.
+     * Moreover, it warms up the connection, sets up notifications, and connects to the socket.
+     * You can use [listener] to get updates about socket connection.
+     *
+     * @param user the user to set
+     * @param tokenProvider a [TokenProvider] implementation
+     */
+    @CheckResult
+    public fun connectUser(user: User, tokenProvider: TokenProvider): Call<ConnectionData> {
```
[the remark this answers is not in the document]
zsmb13: We want to keep `setUser` around for now without changing it. I agree it's a bit weird that the new API calls into now deprecated API, but that'll be fixed once we make `setUser` private (then we can also rename it, etc).

We could extract the contents of `setUser` to another private method already, but I'm not sure we'd gain anything by doing so. Deprecations are for the outside world anyway 😄

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
ogkuzmin: `task`, `block` are  better

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/utils/LiveDataTestUtil.kt
zsmb13: Try removing them and not breaking tests, I spent hours on it and couldn't get it simpler than this 😭 

Robolectric is a pain

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
JcMinarro: `blockingTask`? :thinking:

zsmb13: It's not blocking, it's suspending! That's the point! 😄

I think `suspendingTask` is gonna be the winner.

ogkuzmin (approved):

JcMinarro (approved):
New API is cleaner, So happy with this change

Issue #1313, opened 2021-02-02
NPE for io.getstream.chat.android.client.models.User.getId()


bibibilu:
I am using the latest 4.5.2.
I am using a custom ChannelViewHolderFactory, and my ChannelsView is bound to ChannelsViewModel: 
```
private val filter = and(
          eq("type", "messaging"),
          `in`("members", listOf(ChatDomain.instance().currentUser.id))
  )
private val sort = ChannelsViewModel.DEFAULT_SORT
private val limit = 100
private val viewModel: ChannelsViewModel by viewModels { ChannelsViewModelFactory(filter, sort, limit) }

viewModel.bindView(binding.chatChannelsView, viewLifecycleOwner)
binding.chatChannelsView.setViewHolderFactory(myChannelsViewHolderFactory)
```
Every time I receive new message in the foreground, it crashes for this exception:
```
E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-1
    Process: ai.zira.employee, PID: 29649
    java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String io.getstream.chat.android.client.models.User.getId()' on a null object reference
        at io.getstream.chat.android.livedata.EventHandlerImpl.enrichWithOwnReactions(EventHandlerImpl.kt:463)
        at io.getstream.chat.android.livedata.EventHandlerImpl.updateOfflineStorageFromEvents$stream_chat_android_offline_release(EventHandlerImpl.kt:228)
        at io.getstream.chat.android.livedata.EventHandlerImpl$updateOfflineStorageFromEvents$1.invokeSuspend(Unknown Source:12)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)

```
[...]

PR #1335, opened 2021-02-04
[CAS-667] Fix NPE when new message is received

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-667
https://github.com/GetStream/stream-chat-android/issues/1313

### Description

Only managed to reproduce the crash when `NotificationMessageNewEvent` is for some reason received when the app is the foreground.

What's changed:
- `user` field is removed from `NotificationMessageNewEvent` event as this field is absent
- made `user` parameter nullable in `enrichWithOwnReactions()` just in case

### Checklist

- [X] Reviewers added


ogkuzmin (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
```diff
@@ -229,29 +229,26 @@ public data class NotificationChannelTruncatedEvent(
     override val cid: String,
     @SerializedName("channel_type") val channelType: String,
     @SerializedName("channel_id") val channelId: String,
-    override val user: User,
```
JcMinarro, line 232: Shouldn't we keep it as nullable value(`User?`)? :thinking:

JcMinarro (approved):
LGTM

PR #1343, opened 2021-02-05
[CAS-659] messages header view documentation

leandroBorgesFerreira:
[CAS-659](https://stream-io.atlassian.net/browse/CAS-659)

### Description

Adding documentation for MessagesHeaderView. CMS: https://getstream.io/nessy/docs/chat_docs/android/messages_header_view

### Checklist

- ~New code is covered by unit tests~
- [x] Reviewers added


JcMinarro:
On this view the listener aren't so useful, they doesn't provide any argument, so maybe we doesn't need to add them to the CMS Doc, the initial line where they are explained I think is enough

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessagesHeaderViewModel.kt
```diff
@@ -14,11 +14,11 @@ import io.getstream.chat.android.livedata.ChatDomain
 /**
  * ViewModel class for [com.getstream.sdk.chat.view.ChannelHeaderView].
  * Responsible for updating channel information.
- * Can be bound to the view using [ChannelHeaderViewModel.bindView] function.
+ * Can be bound to the view using [MessagesHeaderViewModel.bindView] function.
  * @param cid the full channel id, i.e. "messaging:123"
  * @param chatDomain entry point for all livedata & offline operations
  */
-public class ChannelHeaderViewModel @JvmOverloads constructor(
+public class MessagesHeaderViewModel @JvmOverloads constructor(
```
adasiewiczr, line 21: This is a breaking change. Are we sure we want to do that?

leandroBorgesFerreira: But this is inside `stream-chat-android-ui-common`, so there are no no users for the module yet.

adasiewiczr: `stream-chat-android-ui-common` is shared between the old SDK `stream-chat-android` and the new one. So whoever uses it will be forced to change ViewModel's name

leandroBorgesFerreira: Indeed, I thought it was `stream-chat-android-ui-components`... You're right, it is a breaking change.

leandroBorgesFerreira: I'll revert the breaking change and we can discuss those renaming problems in the proper meeting.

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Android.kt
```diff
@@ -20,4 +23,17 @@ class Android {
             viewModel.bindView(channelListHeaderView, viewLifecycleOwner)
         }
     }
+
+    /**
+     * @see <a href="https://getstream.io/chat/docs/android/messages_header_view">Messages Header View</a>
+     */
+    inner class MessageListHeader(private val messagesHeaderView: MessagesHeaderView) : Fragment() {
```
bychkovdmitry, line 30: Why is this class `inner`? Also, fragment with constructor args looks weird.

stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/Android.java
```diff
@@ -22,4 +26,20 @@ public void bindingWithViewModel() {
             ChannelListHeaderViewModelBinding.bind(viewModel, channelListHeaderView, getViewLifecycleOwner());
         }
     }
+
+    /**
```
bychkovdmitry, line 30: This link is incorrect
@see <a href="https://getstream.io/chat/docs/android/channel_list_header_view">Channel List Header View</a>

bychkovdmitry (changes requested):

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/Android.kt
leandroBorgesFerreira: They can be not inner and private too... But please mind that this is just a code snipet that only has to compile correctly, so we know that no braking changes were made in those classes.

leandroBorgesFerreira: Better `public` as this is a documentation... changed to `public`.

bychkovdmitry (approved):

PR #1350, opened 2021-02-08
[CAS-672] Push sync service fixes

samiuelson:
### Description

~1. Don't connect to the socket when doing bg data sync API call~
_This is not possible at the moment without modifying LLC API client. Added task to allow setting user without opening WS connection https://stream-io.atlassian.net/browse/CAS-681_
2. Fix storing the bg sync data. Previously, data was persisted before the user was set into `ChatClient` resulting in missing userToken value. Now we are waiting for the first event which is emitted after user is connected to WS, we persist sync config and unsubscribe the listener to make sure it's not invoked on every event. 

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -16,6 +16,7 @@ import io.getstream.chat.android.client.call.Call
 import io.getstream.chat.android.client.call.await
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.events.ChatEvent
+import io.getstream.chat.android.client.events.ConnectedEvent
 import io.getstream.chat.android.client.events.MarkAllReadEvent
 import io.getstream.chat.android.client.extensions.enrichWithCid
 import io.getstream.chat.android.client.logger.ChatLogger
@@ -29,6 +30,7 @@ import io.getstream.chat.android.client.models.TypingEvent
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.client.models.UserEntity
 import io.getstream.chat.android.client.parser.StreamGson
+import io.getstream.chat.android.client.socket.SocketListener
 import io.getstream.chat.android.client.utils.FilterObject
 import io.getstream.chat.android.client.utils.Result
 import io.getstream.chat.android.client.utils.SyncStatus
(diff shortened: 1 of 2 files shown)


bychkovdmitry:
Could you close https://stream-io.atlassian.net/browse/CAS-672 once this one is merged?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
```diff
@@ -58,41 +51,37 @@ internal class OfflineSyncFirebaseMessagingService : FirebaseMessagingService()
...
+        client: ChatClient,
+        message: RemoteMessage,
+    ) {
+        domain.useCases.replayEventsForActiveChannels(cid).enqueue {
+            if (it.isSuccess) {
+                logger.logD("sync success.")
+            } else {
+                logger.logD("sync failed")
+            }
+            client.onMessageReceived(message)
```
bychkovdmitry, line 84: When app process was killed fails loading channel here
```
    private fun loadRequiredData(channelType: String, channelId: String, messageId: String) {
        val getMessage = client.getMessage(messageId)
        val getChannel = client.queryChannel(channelType, channelId, QueryChannelRequest())

        getChannel.zipWith(getMessage).enqueue { result ->
            if (result.isSuccess) {
                val (channel, message) = result.data()

                handler.getDataLoadListener()?.onLoadSuccess(channel, message)
                onRequiredDataLoaded(channel, message)
            } else {
                logger.logE("Error loading required data: ${result.error().message}", result.error())
                handler.getDataLoadListener()?.onLoadFail(messageId, result.error())
                showErrorCaseNotification()
            }
        }
    }
```
with `ChatNetworkError http status -1, stream error code 1007: No defined token. Check if client.setUser was called and finished`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
```diff
@@ -58,41 +51,37 @@ internal class OfflineSyncFirebaseMessagingService : FirebaseMessagingService()
...
-                        this@OfflineSyncFirebaseMessagingService,
-                        user,
-                        config.userToken,
-                        config.apiKey
-                    )
+                    val client = initClient(this@OfflineSyncFirebaseMessagingService, config.apiKey)
                     val domain = initDomain(user, client)
-                    performSync(domain, cid)
-                    client.onMessageReceived(message)
+                    logger.logD("Starting the sync, config: $syncConfig")
```
bychkovdmitry, line 63: I think we need to use the token from syncConfig after process is recreated. Currently, when message arrives, I see fallback notification.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -458,6 +452,28 @@ internal class ChatDomainImpl internal constructor(
         _totalUnreadCount.value = newCount
     }
 
+    private fun storeBgSyncDataWhenUserConnects() {
```
bychkovdmitry, line 455: Great, finally working notifications! 👍

bychkovdmitry (changes requested):

bychkovdmitry (approved):

PR #1357, opened 2021-02-09
[CAS-656] Add channel list view docs

adasiewiczr:
https://getstream.io/nessy/docs/chat_docs/android_chat_ux/channel_list_view

### Checklist

- [x] Reviewers added


stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/Android.java
```diff
@@ -35,6 +55,111 @@ public void bindingWithViewModel() {
...
+
+            // Add separator to the last item
+            channelsView.setShouldDrawItemSeparatorOnLastItem(true);
+        }
+
+        public void customViewHolderFactory() {
+            class CustomChannelListItemViewHolderFactory extends ChannelListItemViewHolderFactory {
+                @Override
+                public int getItemViewType(@NotNull ChannelListItem item) {
+                    // Override together with createViewHolder() to introduce different view holder types
```
ogkuzmin, line 129: Should we include such snippets, if there is not any customization? Maybe use an out-of-box factory and leave customization use case for another doc or cookbook?

adasiewiczr: IMO providing the code for the whole customization is too much for CMS docs, so I decided just to show methods available to override, but I agree that this might be a part of cookbook. @zsmb13 what do you think?

zsmb13: I like the snippet here like this for now, as it provides an idea of what you can do, and then a full customization sample can go somewhere else.

ogkuzmin (approved):

PR #1364, opened 2021-02-09
Display online status for direct chats

bychkovdmitry:
### Description

Tiny fix to display user online statuses in direct chats

### Checklist

- [X] Reviewers added

| Before | After |
| --- | --- |
| device-2021-02-09-164202 | device-2021-02-09-163918 |


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/ChannelViewHolder.kt
```diff
@@ -171,7 +172,12 @@ internal class ChannelViewHolder @JvmOverloads constructor(
     }
 
     private fun StreamUiChannelListItemForegroundViewBinding.configureAvatarView() {
-        avatarView.setChannelData(channel)
+        val otherUsers = channel.getUsers()
+        if (otherUsers.size == 1) {
```
adasiewiczr, line 176: What about group channels with only one member? Should we treat them as direct channels or group channels?

bychkovdmitry: Good catch, do we have a method to distinguish them? The only difference is non-empty name?

bychkovdmitry: So, the check should be 
```
if (otherUsers.size == 1 && channel.name.isEmpty) {
  // show user avatar with online status
} else {
  // show channel avatar with 1+ avatars without online indicator
}
```
?

adasiewiczr: not really, group channel can also have empty name. We can add an extension method similar to this: https://github.com/GetStream/stream-chat-android/blob/main/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/group/GroupChatInfoFragment.kt#L79
If channel is distinct and it has only one member (excluding current user) we can assume it's one to one conversation

bychkovdmitry: I checked flutter implementation and they are doing the same thing. Anyway, the "user.presence.changed" event is not working now.

bychkovdmitry: Created this task https://stream-io.atlassian.net/browse/CAS-700

adasiewiczr (approved):

JcMinarro (approved):
LGTM

PR #1365, opened 2021-02-09
Fixing method name

leandroBorgesFerreira:
### Description

Small correction for #1356. The name of the method was not supposed to be changed

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
@@ -662,7 +662,7 @@ public class MessageListView : ConstraintLayout {
         buffer.enqueueData(listItem)
     }
 
-    public fun setMessageListItemPredicate(messageListItemPredicate: MessageListItemPredicate) {
+    public fun setMessageListItemFilter(messageListItemPredicate: MessageListItemPredicate) {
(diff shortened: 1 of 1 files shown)


samiuelson (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/messages/view/MessageListView.kt
```diff
@@ -662,7 +662,7 @@ public class MessageListView : ConstraintLayout {
         buffer.enqueueData(listItem)
     }
 
-    public fun setMessageListItemPredicate(messageListItemPredicate: MessageListItemPredicate) {
```
ogkuzmin, line 665: Why do you call it filter? It's not filter anymore. Just object that holds predicated

ogkuzmin: We haven't published our docs yet. We can change namings

leandroBorgesFerreira: The user is setting the predicate that will be used to filter the messages in the `handleNewWrapper`. For me, filter seams an easier word than predicate. If you think this is misleading, we can maintain predicate.

leandroBorgesFerreira: Let's keep the predicate then, we can revisit this in a moment.

PR #1367, opened 2021-02-09
Constructor overloads

leandroBorgesFerreira:
### Description

This PR adds the overloads for the constructors of some important data classes for the Java users. Instead of passing 10+ parameters to `Message`, a user programming in Java can use just 1 parameter to create it. This can be good to simplify the code and to simplify code for testing purposes. 

It also makes our documentation simpler for the Java parts. 

### Checklist

- Changelog updated with client-facing changes
- [x] Reviewers added


zsmb13:
For Java users, you can already do this:

```java
Message message = new Message();
message.setId("some-id");
message.setText("Hello world");
```

stream-chat-android-client/api/stream-chat-android-client.api
```diff
@@ -2241,6 +2241,40 @@ public final class io/getstream/chat/android/client/models/Member : io/getstream
...
+	public fun <init> (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/util/List;Ljava/util/List;Ljava/util/List;ILjava/util/Map;Ljava/util/Map;Lio/getstream/chat/android/client/utils/SyncStatus;Ljava/lang/String;Ljava/util/List;Ljava/util/List;Ljava/util/Date;Ljava/util/Date;Ljava/util/Date;Ljava/util/Date;Ljava/util/Date;Lio/getstream/chat/android/client/models/User;Ljava/util/Map;ZZLjava/util/Map;ZLio/getstream/chat/android/client/models/ChannelInfo;Lio/getstream/chat/android/client/models/Message;Ljava/lang/String;ZLjava/util/Date;Ljava/util/Date;Lio/getstream/chat/android/client/models/User;)V
```
bychkovdmitry, line 2277: I think this is too much. Unfortunately, users will not be able to specify lets say the 1st and 10th argument. So, IMO doesn't make a lot of sense. I guess what java clients want is a builder here.

bychkovdmitry:
> For Java users, you can already do this:
> 
> ```java
> Message message = new Message();
> message.setId("some-id");
> message.setText("Hello world");
> ```

Are default args lost in that case?

leandroBorgesFerreira:
I didn't realise we could instantiate an empty constructor and then just use the setters. No need for this PR then.

PR #1374, opened 2021-02-10
[CAS-682] Add ViewModels for new ui module

adasiewiczr:
- Introduce abstract _BaseMessageListHeaderViewModel_ inside common module which contains logic for both _ChannelHeaderViewModel_ and _MessageListHeaderViewModel_
- Move _ChannelHeaderViewModel_ to old SDK package
- Introduce _ChannelListViewModel_ which has exactly the same logic as _ChannelsViewModel_ (I've changed _ChannelListViewModel.Event_ to _ChannelListViewModel.Action_). There was no easy way to share logic between those two without introducing breaking changes because of `State` and `Event` classes so I decided to copy it
- Move _ChanneslViewModel_ to the old SDK package
- Introduce new ViewModelFactories and move old ones into old SDK package

- [x] Reviewers added


stream-chat-android-ui-components/src/test/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModelTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import io.getstream.chat.android.ui.createUser
+import org.amshove.kluent.When
+import org.amshove.kluent.calling
+import org.amshove.kluent.shouldBeEqualTo
+import org.amshove.kluent.shouldBeInstanceOf
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.extension.ExtendWith
+
+@ExtendWith(InstantTaskExecutorExtension::class)
+internal class ChannelListViewModelTest {
```
adasiewiczr, line 31: Copy-pasted tests from common module

stream-chat-android-ui-components/src/test/kotlin/io/getstream/chat/android/ui/messages/header/viewmodel/MessageListHeaderViewModelTest.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.junit.jupiter.api.extension.ExtendWith
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
+
+private const val CID = "CID:messaging"
+private val CURRENT_USER = createUser(online = true)
+
+@ExtendWith(InstantTaskExecutorExtension::class)
+internal class MessageListHeaderViewModelTest {
```
adasiewiczr, line 34: Copy-pasted tests from common module

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -0,0 +1,130 @@
...
+                paginationStateMerger.addSource(queryChannelsController.endOfChannels) { endOfChannels ->
+                    setPaginationState { copy(endOfChannels = endOfChannels) }
+                }
+            } else {
+                stateMerger.postValue(State.NoChannelsAvailable)
+            }
+        }
+    }
+
+    public fun onAction(action: Action) {
```
samiuelson, line 75: It's a much better name  👍🏼

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    private fun requestMoreChannels() {
+        chatDomain.useCases.queryChannelsLoadMore(filter, sort).enqueue()
+    }
+
+    private fun setPaginationState(reducer: PaginationState.() -> PaginationState) {
+        paginationStateMerger.value = reducer(paginationStateMerger.value ?: PaginationState())
+    }
+
+    public sealed class State {
```
samiuelson, line 109: I'd suggest making `State` a data class, and keeping the view model design consistent with other view models in the `ui` module, eg. `SearchViewModel` or `MentionListViewModel`.

carterhudson: I think we should do it everywhere we have interactions with state & events so we don't lose values when we transition from one state to another. Is it within the scope of this PR, though?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
carterhudson: Smells like Redux :D

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
adasiewiczr: agree! let's do it in a separate PR. I would like to merge #1379 before that

carterhudson (approved):
LGTM

PR #1382, opened 2021-02-10
Cas 222 implement query members offline sdk

carterhudson:
https://stream-io.atlassian.net/browse/CAS-222

### Description

- Implements `QueryMembers` UseCase and adds it to `UseCaseHelper`
- Treats db as source of truth for offline operations, otherwise calls through to backend and updates database.

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/QueryMembersImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+import io.getstream.chat.android.client.utils.FilterObject
+import io.getstream.chat.android.client.utils.Result
+import io.getstream.chat.android.livedata.ChatDomainImpl
+
+public interface QueryMembers {
+
+    @CheckResult
+    public operator fun invoke(
+        channelType: String,
+        channelId: String,
```
adasiewiczr, line 17: I think we can replace `channelType` and `channelId` with `cid` to match other use cases

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/QueryMembersImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+        offset: Int = 0,
+        limit: Int = 0,
+        filter: FilterObject = FilterObject(),
+        sort: QuerySort<Member> = QuerySort<Member>().desc("created_at"),
```
adasiewiczr, line 21:
```suggestion
        sort: QuerySort<Member> = QuerySort.desc(Member::createdAt),
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/QueryMembersImpl.kt
```diff
@@ -0,0 +1,50 @@
...
+        sort: QuerySort<Member>,
+        members: List<Member>,
+    ): Call<List<Member>> {
+        return CoroutineCall(domainImpl.scope) {
+            if (domainImpl.isOffline() && domainImpl.offlineEnabled) {
+                val offlineMembers: List<Member> =
+                    domainImpl.channel(channelType, channelId).members.value ?: emptyList()
+                Result(offlineMembers)
+            } else {
+                domainImpl
```
adasiewiczr, line 42: Shouldn't we store members in the database and also add them to `ChannelController:members` and `Channel:members`?

carterhudson: Since this function provides the ability to obtain a subset of the actual members, should we be overwriting the channel members in the db?

ogkuzmin: I think for fresh data obtained from BE we need to update those values in local DB. For offline case we can fetch data from DB (like you said emulating BE behaviour)

adasiewiczr: To be honest I'm not sure what we should do in that case. Maybe this function should return 'ChannelController' which exposes Members LiveData and post updates there (like we do with messages) instead of returning subset of members? I think it might be worth it to check how other SDKs deal with it (I'm not sure if other teams already implemented that) or we should discuss that with other team members?

carterhudson: I'll check out the other SDKs and see how they handle it as well.

ogkuzmin: IMO, fetching data from DB directly is better than exposing the LiveData value of `ChannelController.` Because:
1) It might be not updated
2) We want to have one source of truth in the offline SDK. We agreed to have DB as such source

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/QueryMembers.kt
```diff
@@ -0,0 +1,76 @@
...
+                    .sortedWith(sort.comparator)
+                    .drop(offset)
+                    .let { afterDrop ->
+                        if (limit > 0) {
+                            afterDrop.take(limit)
+                        } else afterDrop
+                    }
+            } ?: emptyList()
+
+        return Result(membersFromDatabase)
```
ogkuzmin, line 74: Why do you do all these checks? Kotlin built-in collection extensions do all this stuff already. You don't need to check if the collection is empty for applying drop operator. With a negative limit value, the take operator will throw an exception.

```suggestion
return domainImpl.repos
            .selectMembersForChannel(cid)
            .sortedWith(sort.comparator)
            .drop(offset)
            .take(limit)
            .let(::Result)
```

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/QueryMembersTest.kt
```diff
@@ -0,0 +1,163 @@
...
+            on(domain.scope) doReturn TestCoroutineScope()
+            on(domain.offlineEnabled) doReturn true
+        }
+
+        queryMembers = QueryMembers(mockDomain)
+    }
+
+    @Test
+    fun `Given default arguments, and ONLINE, SHOULD make a remote call via client`() = runBlockingTest {
+        When calling mockDomain.isOffline() doReturn false
```
ogkuzmin, line 90: What about not using kluent anymore as we discussed in the last catch-up?

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/QueryMembersTest.kt
```diff
@@ -0,0 +1,163 @@
...
+import kotlinx.coroutines.test.runBlockingTest
+import org.amshove.kluent.When
+import org.amshove.kluent.`should be equal to`
+import org.amshove.kluent.any
+import org.amshove.kluent.calling
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+@OptIn(ExperimentalCoroutinesApi::class)
+internal class QueryMembersTest {
```
ogkuzmin, line 29: Neat and clean tests! My private respect 👍

ogkuzmin (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/QueryMembers.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    private suspend fun queryMembersOnline(
+        cid: String,
+        offset: Int,
+        limit: Int,
+        filter: FilterObject,
+        sort: QuerySort<Member>,
+        members: List<Member>,
+    ): Result<List<Member>> {
+        val cidComponents = cid.split(':')
```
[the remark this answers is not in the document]
adasiewiczr: We could use `ChannelClient` instead of `ChatClient` which already handles splitting cid:
`domainImpl.client.channel(cid).queryMembers(...)`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/QueryMembers.kt
carterhudson: Default take value of 0 will return an empty list, so some checks are needed.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/usecase/QueryMembersTest.kt
carterhudson: Sure, can update it.

ogkuzmin (approved):

PR #1386, opened 2021-02-11
CAS-681 Push data sync improvement

samiuelson:
https://stream-io.atlassian.net/browse/CAS-681
### Description

* Introduce `ChatClient::setUserWithoutConnecting`. This function is needed to initialize `ChatClient` without opening WS connection. This way `ChatClient` can be used to make REST calls in case the app's process is not alive when data sync operation is being triggered by a push message.
* Fix `::queryChannel` params validation. Allow to run `::queryChannel` without `clientId`
* Refactor `OfflineSyncFirebaseMessagingService` to use `ChatClient::setUserWithoutConnecting` instead of `ChatClient::connectUser` in case the app's process is dead.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -242,6 +242,27 @@ public class ChatClient internal constructor(
...
+     * Initializes [ChatClient] for a specific user and a given [userToken].
+     * Caution: This method doesn't establish connection to the web socket, you should use [connectUser] instead.
+     *
+     * This method initializes [ChatClient] to allow the use of Stream REST API client.
+     * Moreover, it warms up the connection, and sets up notifications.
+     *
+     * @param user the user to set
+     * @param userToken the user token
+     */
+    public fun setUserWithoutConnecting(user: User, userToken: String) {
```
zsmb13, line 255: > setUserWithoutConnecting

We should talk to other teams and see what they call methods like this, or if they have them available at all. We just had to do a bunch of renaming around `setUser`, would be good to be synchronized on new APIs around here.

samiuelson: Right, I'll start a cross-team discussion about this.

samiuelson: Since we don't have consistent APIs at the moment (especially compared to iOS) and other platforms don't need this feature. My suggestion is to add the function as it is and refactor it later along with iOS. I believe having it as a simple separate function would be easy to refactor in the future.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -242,6 +242,27 @@ public class ChatClient internal constructor(
...
+        if (isUserSet()) {
```
ogkuzmin, line 256: Why logic ensureUserNotSet doesn't work here?
I see a lot of doubling comparing with `setUser` can we somehow create some general method?

samiuelson: `ensureUserNotSet` uses `InitConnectionListener` which is not used in `setUserWithoutConnecting` because we are not starting WS connection.

I'll try to extract common parts of code.

UNRELEASED_CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 - Add `streamMessageActionButtonsBackground` attributes to `MessageListView`
 
 ## stream-chat-android-client
+- Fix `ConcurrentModificationException` on our `NetworkStateProvider`
```
adasiewiczr, line 12: I think it was already released
```suggestion
```

samiuelson: Ah, thanks!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: Can't we hide this method from our public API?
At the end it is not a method that our user will use, right?

samiuelson: But we need an access to this method in the `offline` module.

zsmb13: Sounds like `@InternalStreamChatApi`

samiuelson: Added!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/GsonChatApi.kt
```diff
@@ -374,29 +374,14 @@ internal class GsonChatApi(
...
-            retrofitApi.queryChannel(
-                channelType,
-                channelId,
-                apiKey,
-                userId,
-                connectionId,
-                query
-            )
-        }.map(::flattenChannel)
+        return retrofitApi.queryChannel(
```
ogkuzmin, line 377: There was `channelId` checking and you deleted it. Why it's not required anymore? I don't see that you move this checking elsewhere.

samiuelson: My bad, fixed.

ogkuzmin (approved):
LGTM

PR #1397, opened 2021-02-12
[CAS-394] Fetch users offline

ogkuzmin:
### Description

Step 1 for the story "Create channel offline"

1) Move logic for fetching users to UseCase
2) Augment logic for the offline case. Fetch all user from DB, make a search by name in DB
3) Cover logic by tests

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/PerformanceUtils.kt
```diff
@@ -50,17 +51,25 @@ public object PerformanceUtils {
 
     public fun <T> task(taskName: String, task: () -> T): T {
         synchronized(taskName) {
+            val count = tasks[taskName]?.count ?: 0
             val startTime = System.currentTimeMillis()
             return task().also {
-                doMeasureAndUpdateResult(taskName, TaskEntry(name = taskName, lastStart = startTime))
+                doMeasureAndUpdateResult(
```
ogkuzmin, line 57: Fix for a correct average duration calculation

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Filters.kt
```diff
@@ -17,52 +17,48 @@ import io.getstream.chat.android.client.utils.FilterObject
  */
 public object Filters {
 
+    public const val KEY_EXIST: String = "\$exists"
```
ogkuzmin, line 20: To be able to check filter values

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/SearchUsersByName.kt
```diff
@@ -0,0 +1,80 @@
...
+import io.getstream.chat.android.client.api.models.QueryUsersRequest
+import io.getstream.chat.android.client.call.Call
+import io.getstream.chat.android.client.call.CoroutineCall
+import io.getstream.chat.android.client.models.Filters
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.client.models.name
+import io.getstream.chat.android.client.utils.Result
+import io.getstream.chat.android.livedata.ChatDomainImpl
+
+public class SearchUsersByName internal constructor(private val chatDomainImpl: ChatDomainImpl) {
```
adasiewiczr, line 14: I really like dropping `interface` here. Would be nice to add some KDoc here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/SearchUsersByName.kt
```diff
@@ -0,0 +1,80 @@
...
+
+    @VisibleForTesting
+    internal val defaultUsersQueryFilter by lazy {
+        Filters.and(
+            Filters.ne("name", ""),
```
adasiewiczr, line 19:
```suggestion
            Filters.ne(FIELD_NAME", ""),
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/SearchUsersByName.kt
```diff
@@ -0,0 +1,80 @@
...
+            Filters.ne("id", chatDomainImpl.currentUser.id)
```
adasiewiczr, line 20:
```suggestion
            Filters.ne(FIELD_ID, chatDomainImpl.currentUser.id)
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/SearchUsersByName.kt
```diff
@@ -0,0 +1,80 @@
...
+            )
+        }
+
+        return chatDomainImpl.client.queryUsers(
+            QueryUsersRequest(
+                filter = filter,
+                offset = offset,
+                limit = userLimit,
+                querySort = USERS_QUERY_SORT,
+                presence = true
```
adasiewiczr, line 64: Should we allow to pass that in `invoke` method?

adasiewiczr (approved):
Good job,  neat solution! I just added some NITs and questions

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/usecase/SearchUsersByName.kt
ogkuzmin: There is no sense to pass QuerySort since searching by name, and by default ordering by name works (otherwise we can't support any other case in DB).
But for presence it makes sense

adasiewiczr: agree, would be nice to mention about default filter in KDOC.
`presence` also won't work with `offline` results right?

PR #1403, opened 2021-02-15
[CAS-708] Docs for moderation tools

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-708

### Description

Add doc snippets to the functionality which already exists. What's left to be done:
- Client side mutes
- Unflag message/user
- Global ban from all channels (+ `ip_ban ` feature)
- Global shadow ban
https://stream-io.atlassian.net/browse/CAS-710

### Checklist

- [X] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/controllers/ChannelController.kt
```diff
@@ -63,7 +63,7 @@ public interface ChannelController {
     public fun banUser(targetId: String, reason: String?, timeout: Int?): Call<Unit>
 
     @CheckResult
-    public fun unBanUser(targetId: String, reason: String?, timeout: Int?): Call<Unit>
+    public fun unBanUser(targetId: String): Call<Unit>
```
adasiewiczr, line 66: Nice catch! 
This might be a breaking change for our clients, so it's worth mentioning that in CHANGELOG. 
Maybe we should consider introducting a separate method inside `ChannelClient` and deprecate that one?

bychkovdmitry: Reflected the change in the changelog. Not sure that we need to keep the old method.

adasiewiczr (approved):

PR #1408, opened 2021-02-16
Extract OfflineSyncFirebaseMessagingService logic into delegate

samiuelson:
https://stream-io.atlassian.net/browse/CAS-671

### Description

Extracted all the logic form `OfflineSyncFirebaseMessagingService` into `StreamOfflineSyncCloudMessageDelegate` class. 
The delegate class is supposed to be used from a custom `FirebaseMessagingService` implementation in case the user doesn't want to use `OfflineSyncFirebaseMessagingService` directly (for example, they are working with multiple push providers). 

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
@@ -5,8 +5,8 @@ import com.google.firebase.messaging.RemoteMessage
 
 internal class OfflineSyncFirebaseMessagingService : FirebaseMessagingService() {
 
-    private val syncServiceDelegate: StreamOfflineSyncCloudMessageDelegate =
-        StreamOfflineSyncCloudMessageDelegate(this)
+    private val syncServiceDelegate: PushMessageSyncHandler =
+        PushMessageSyncHandler(this)
 
     override fun onNewToken(token: String) {
         syncServiceDelegate.onNewToken(token)
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
@@ -1,137 +1,19 @@
 package io.getstream.chat.android.livedata.service.sync
 
-import android.app.NotificationChannel
-import android.app.NotificationManager
-import android.content.Context
-import android.os.Build
-import androidx.core.app.NotificationCompat
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
-import io.getstream.chat.android.client.ChatClient
-import io.getstream.chat.android.client.logger.ChatLogger
-import io.getstream.chat.android.client.models.User
-import io.getstream.chat.android.client.notifications.FirebaseMessageParser
-import io.getstream.chat.android.client.notifications.FirebaseMessageParserImpl
-import io.getstream.chat.android.client.notifications.handler.ChatNotificationHandler
-import io.getstream.chat.android.client.notifications.handler.NotificationConfig
-import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
-import io.getstream.chat.android.livedata.ChatDomain
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.launch
 
 internal class OfflineSyncFirebaseMessagingService : FirebaseMessagingService() {
 
-    private val logger = ChatLogger.get("OfflineSyncFirebaseMessagingService")
-
-    private val syncModule by lazy { SyncProvider(this) }
-    private val notificationConfig: NotificationConfig by lazy { syncModule.notificationConfigStore.get() }
-    private val firebaseMessageParser: FirebaseMessageParser by lazy { FirebaseMessageParserImpl(notificationConfig) }
+    private val syncServiceDelegate: StreamOfflineSyncFirebaseServiceDelegate =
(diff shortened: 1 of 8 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/StreamOfflineSyncCloudMessageDelegate.kt
```diff
@@ -0,0 +1,160 @@
...
+import kotlinx.coroutines.launch
+
+/**
+ * A delegate class which can be used in [com.google.firebase.messaging.FirebaseMessagingService] in case the app uses
+ * multiple push notifications backend services.
+ *
+ * If the app uses Stream service as its only push notifications provider, the [OfflineSyncFirebaseMessagingService]
+ * class can be used as the  out-of-the box implementation instead of [StreamOfflineSyncCloudMessageDelegate] methods.
+ */
+public class StreamOfflineSyncCloudMessageDelegate(private val service: Service) {
```
samiuelson, line 30: All the code in the delegate class comes from the `OfflineSyncFirebaseService`.

ogkuzmin (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * A method to check if the cloud message is sent from Stream backend service.
+     *
+     * @param message instance of [RemoteMessage] delivered at [FirebaseMessagingService]
+     *
+     * @return true if the message is sent from Stream backend, false otherwise.
+     */
+    public fun isStreamMessage(message: RemoteMessage): Boolean {
```
ogkuzmin, line 106: We can make this method class level instead of instance level. Even make it static for Java users. If it makes sense. What do you think?

samiuelson: The problem is it requires `notificationsConfig`, so the static method would need to take a `Context` at least.

ogkuzmin (approved):
LGTM

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -0,0 +1,171 @@
...
+                    val user = User(id = config.userId)
+                    val token = config.userToken
+                    val client = initClient(service, user, token, config.apiKey)
+                    val domain = initDomain(user, client)
+                    logger.logD("Starting the sync, config: $syncConfig")
+                    performSync(domain, cid, client, message)
+                }
+            }
+        }
+        service.stopForeground(true)
```
zsmb13, line 96: This `stopForeground` call used to be in the body of the coroutine launched above, so it would run after sync has completed. Now we `launch` a coroutine, and don't wait for its completion, just immediately stop the service.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -0,0 +1,171 @@
...
+import kotlinx.coroutines.launch
+
+/**
+ * A delegate class which can be used in [com.google.firebase.messaging.FirebaseMessagingService] in case the app uses
+ * multiple push notifications backend services.
+ *
+ * If the app uses Stream service as its only push notifications provider, the [OfflineSyncFirebaseMessagingService]
+ * class can be used as the  out-of-the box implementation instead of [PushMessageSyncHandler] methods.
+ */
+public class PushMessageSyncHandler(private val service: Service) {
```
zsmb13, line 30: Passing in this `service` instance makes for easy integration (as in few lines) for the client, but I'm really not sure if they'll expect us to stop/start this service for them.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/OfflineSyncFirebaseMessagingService.kt
```diff
@@ -1,137 +1,19 @@
...
-import kotlinx.coroutines.launch
 
 internal class OfflineSyncFirebaseMessagingService : FirebaseMessagingService() {
 
-    private val logger = ChatLogger.get("OfflineSyncFirebaseMessagingService")
-
-    private val syncModule by lazy { SyncProvider(this) }
-    private val notificationConfig: NotificationConfig by lazy { syncModule.notificationConfigStore.get() }
-    private val firebaseMessageParser: FirebaseMessageParser by lazy { FirebaseMessageParserImpl(notificationConfig) }
+    private val syncServiceDelegate: PushMessageSyncHandler =
```
zsmb13, line 8: We still have the "delegate" naming in this property, also in some of the documentation comments added on this PR.

zsmb13 (commented):
Some late comments on this PR

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
samiuelson: Do you mean starting the service foreground execution and then stoping it after sync? 
```kotlin 
service.startForeground(NOTIFICATION_ID, this)
```
and
```kotlin
service.stopForeground(true)
```
Why is it bad?

zsmb13: I think they could have code that would run when the service is started or stopped, leading to potential side effects. Haven't worked with Services in a while though, just feels like we should be careful around these parts.

PR #1416, opened 2021-02-17
CAS-670 Fix for sync foreground execution end

samiuelson:
* Fix the place we stop foreground execution of sync. 
* Run the sync operation synchronously
* A bunch of improvements for docs

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
@@ -5,6 +5,7 @@ import android.app.NotificationManager
 import android.app.Service
 import android.content.Context
 import android.os.Build
+import androidx.annotation.WorkerThread
 import androidx.core.app.NotificationCompat
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
@@ -69,6 +70,7 @@ public class PushMessageSyncHandler(private val service: Service) {
      * @param message the [RemoteMessage] delivered to [com.google.firebase.messaging.FirebaseMessagingService]
      * registered at your app.
      */
+    @WorkerThread
     public fun onMessageReceived(message: RemoteMessage) {
         if (!isStreamMessage(message)) {
             return
@@ -114,14 +116,13 @@ public class PushMessageSyncHandler(private val service: Service) {
         client: ChatClient,
         message: RemoteMessage,
     ) {
-        domain.useCases.replayEventsForActiveChannels(cid).enqueue {
-            if (it.isSuccess) {
-                logger.logD("Sync success.")
-            } else {
-                logger.logD("Sync failed.")
-            }
-            client.onMessageReceived(message)
+        val result = domain.useCases.replayEventsForActiveChannels(cid).execute()
+        if (result.isSuccess) {
+            logger.logD("Sync success.")
+        } else {
+            logger.logD("Sync failed.")
         }
+        client.onMessageReceived(message)
     }
 
     private fun startForegroundExecution(smallIcon: Int) {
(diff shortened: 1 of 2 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -73,31 +79,31 @@ public class PushMessageSyncHandler(private val service: Service) {
             return
         }
         createSyncNotificationChannel()
-        showForegroundNotification(notificationConfig.smallIcon)
+        startForegroundExecution(notificationConfig.smallIcon)
 
-        GlobalScope.launch(DispatcherProvider.IO) {
```
samiuelson, line 78: Since it's a service there's no need to run sync on IO dispatcher. Running sync operation synchronously will prevent potential side effects (manual switching background/foreground mode by the user during sync operation).

zsmb13: Services run on the main thread by default ([see here](https://developer.android.com/guide/components/services)), but the `onMessageReceived` method in the Firebase SDK does promise to be on a worker thread.

zsmb13: Might be worth adding `@WorkerThread` annotations to these methods, if we expect them to always be called from a background thread

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -73,31 +79,31 @@ public class PushMessageSyncHandler(private val service: Service) {
...
+                val user = User(id = config.userId)
+                val token = config.userToken
+                val client = initClient(service, user, token, config.apiKey)
+                val domain = initDomain(user, client)
+
+                logger.logD("Starting the sync, config: $syncConfig")
+
+                performSync(domain, cid, client, message)
             }
+            service.stopForeground(true)
```
zsmb13, line 101: This only happens on the `else` branch now, seems suspicious

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -73,31 +79,31 @@ public class PushMessageSyncHandler(private val service: Service) {
             return
         }
         createSyncNotificationChannel()
-        showForegroundNotification(notificationConfig.smallIcon)
+        startForegroundExecution(notificationConfig.smallIcon)
```
zsmb13, line 82: Are these notification handling methods safe to call from any thread?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
samiuelson: I've added annotation to `onMessageReceived` and modified the sync call to run synchronously within the thread it's called.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
samiuelson: Yes, the `service.startForeground()` is safe to run from any thread.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
samiuelson: Yes, my bad. Fixed.

zsmb13 (approved):
LGTM

PR #1418, opened 2021-02-17
Introduce BaseMessageListHeaderViewModel::resetThread to make it consistent with thread handling in MessageInputViewModel

adasiewiczr:
Introduce `BaseMessageListHeaderViewModel::resetThread`method to match thread handling in MessageInputView. Make `BaseMessageListHeaderViewModel::setActiveThread` message parameter non-nullable

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/BaseMessageListHeaderViewModel.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/BaseMessageListHeaderViewModel.kt
@@ -53,7 +53,11 @@ public abstract class BaseMessageListHeaderViewModel constructor(
         }
     }
 
-    public fun setActiveThread(message: Message?) {
+    public fun setActiveThread(message: Message) {
         _activeThread.postValue(message)
+    }
+
+    public fun resetThread() {
+        _activeThread.postValue(null)
     }
 }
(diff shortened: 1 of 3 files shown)


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/BaseMessageListHeaderViewModel.kt
```diff
@@ -53,7 +53,11 @@ public abstract class BaseMessageListHeaderViewModel constructor(
         }
     }
 
-    public fun setActiveThread(message: Message?) {
```
zsmb13, line 56: Is this a breaking change, even for the old UI library, via `ChannelHeaderViewModel`?

adasiewiczr: That's a tricky one - `activeThread` was introduced few releases ago (at that time we only had one VM in the common module) but it is used only in the new UI component module. So technically this might be a breaking change. I'll revert that and just introduce a new method

zsmb13: Maybe it shouldn't be in the base class, if it's not used in the old module? All the more reason to split up & duplicate these ViewModels instead of sharing 😄

adasiewiczr: True, but on the other hand we did multiple releases before introducing different VMs for those modules and `activeThread ` was included. I'm not sure if removing it now is safe 😢

zsmb13: The base class itself should be unused by clients. We can duplicate its functionality into the inheriting classes, and then deprecate/remove the unnecessary parts.

adasiewiczr: Agree, I think we can do it after the release because it shouldn't be super important and that won't be a breaking change

zsmb13 (approved):

PR #1424, opened 2021-02-17
Fix ReadIndicator State

### LINKED ISSUE #1411: Read status not updating in MessageListView
**Describe the bug**
We currently have an issue in the `MessageListView` where the read status is just not updated live. So when we reload the view the read state is always accurate, but when we stay on the screen and the read status of my last messages changes, the avatar indicator does not move to the bottom. 

We can see the `message.read` events coming in the logs but the view never updates (it still updates when we scroll to the top which triggers some loading of previous messages, then scroll down).

Interesting is that as soon as we debug and set breakpoints in the `MessageListItemViewHolder` or in the `IndicatorConfigurator` and cycle through the `bind()` calls, the view updates correctly without any reloading. Not sure what side effect that might be but it could point to some sort of race condition or weird timing issue.

In general we use the View and ViewModel like in the docs and demos, so nothing special. We have our own `MessageListItemViewHolder`, but even without that we monitor the same behavior. 

**SDK version**
- 4.5.5

**To Reproduce**
1. Go to a `MessageListView`
2. Send a message
3. Wait for the message to be read by the other user
4. Read indicator (avatar of the other user) does not move to the message

**Expected behavior**
The read indicator should move to the new message automatically as soon as we receive the read event

**Device:**
 - Vendor and model: Google Pixel 3 XL
 - Android version: 11

JcMinarro:
### Description
Fix ReadIndicator State
Fixes #1411 

### Checklist

- ~[ ] Changelog updated with client-facing changes
- ~[ ] Comparison screenshots added for visual changes
- [x] Reviewers added

### DIFF
--- a/stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemDiffCallback.kt
+++ b/stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemDiffCallback.kt
@@ -47,7 +47,7 @@ internal object MessageListItemDiffCallback : DiffUtil.ItemCallback<MessageListI
                 syncStatus = oldItem.message.syncStatus != newItem.message.syncStatus,
                 deleted = oldItem.message.deletedAt != newItem.message.deletedAt,
                 positions = oldItem.positions != newItem.positions,
-                readBy = oldItem.messageReadBy.map { it.getUserId() } == newItem.messageReadBy.map { it.getUserId() }
+                readBy = oldItem.messageReadBy.map { it.getUserId() } != newItem.messageReadBy.map { it.getUserId() }
             )
         } else {
             null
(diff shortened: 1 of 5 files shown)


stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemDiffCallback.kt
```diff
@@ -47,7 +47,7 @@ internal object MessageListItemDiffCallback : DiffUtil.ItemCallback<MessageListI
                 syncStatus = oldItem.message.syncStatus != newItem.message.syncStatus,
                 deleted = oldItem.message.deletedAt != newItem.message.deletedAt,
                 positions = oldItem.positions != newItem.positions,
-                readBy = oldItem.messageReadBy.map { it.getUserId() } == newItem.messageReadBy.map { it.getUserId() }
+                readBy = oldItem.messageReadBy.map { it.getUserId() } != newItem.messageReadBy.map { it.getUserId() }
```
ogkuzmin, line 50: 🤨 Nice catch) We should pay more attention to tests

ogkuzmin (approved):

stream-chat-android/src/main/java/com/getstream/sdk/chat/adapter/MessageListItemDiffCallback.kt
bychkovdmitry: Haha, good, hard to notice as here the condition is inverted comparing to `areContentsTheSame`

bychkovdmitry (approved):

PR #1453, opened 2021-02-22
Raise deprecation levels

zsmb13:
### Description

Raises the deprecation level of already deprecated constructs in the SDK, according to the deprecations document.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION_ERROR")
```
zsmb13, line 1: These are needed in a handful of places because we have to reference the types that we are marking as error level deprecations in this PR. File level annotation targest allow us to import these types despite using them being marked as an error.

samiuelson (approved):

PR #1458, opened 2021-02-23
[CAS-698] Attachment view factory

ogkuzmin:
### Description

Introduce brand new `AttachmentViewFactory` in order to let create custom attachment views/link views
1) Create new  `AttachmentViewFactory`
2) Remove hardcoded `FileAttachmentsView`, `MediaAttachmentGroupView`, `LinkAttachmentView` from XML of VHs
3) Create `attachmentsView`/`linkAttachmentView` dynamically and add them to container when `ViewHolder::bing`
4) Remove redundant decorators

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentsGroupView.kt
```diff
@@ -191,6 +198,24 @@ internal class MediaAttachmentsGroupView : ConstraintLayout {
         }
     }
 
+    fun setupBackground(data: MessageListItem.MessageItem) {
```
ogkuzmin, line 201: Move from BackgroundDecorator

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/TextDecorator.kt
```diff
@@ -45,11 +45,6 @@ internal class TextDecorator(val style: MessageListItemStyle) : BaseDecorator()
         viewHolder.binding.apply {
             getStyleTextColor(data.isMine)?.let { textColor ->
                 messageText.setTextColor(textColor)
-                // set title & description colors as well
```
ogkuzmin, line 48: It's not right. We have link view in many others VHs, but somewhy we did setup only for `MessagePlainText`. Remove this right now. I will fix setup link attachments text in next PR

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/MessagePlainTextViewHolder.kt
```diff
@@ -66,5 +69,47 @@ internal class MessagePlainTextViewHolder(
...
+                setLinkPreviewClickListener { url ->
+                    listeners.linkClickListener.onLinkClick(url)
+                }
+                setLongClickTarget(binding.root)
+                showLinkAttachment(linkAttachment)
+            }
+        }
+    }
+
+    private fun applyMaxPossibleWidth(root: ConstraintLayout, messageContainer: View, message: Message) {
```
ogkuzmin, line 96: Moved from `LinkDecorator`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/PlainTextWithFileAttachmentsViewHolder.kt
```diff
@@ -108,6 +94,41 @@ internal class PlainTextWithFileAttachmentsViewHolder(
         }
     }
 
+    private fun setupAttachments(data: MessageListItem.MessageItem) {
+        binding.attachmentsContainer.removeAllViews()
```
ogkuzmin, line 98: A lot of doublings as an intermediate step for the solution. In next PR I will fuse these VHs into one

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
```diff
@@ -29,16 +30,21 @@ import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.Pla
 import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.ThreadSeparatorViewHolder
 
 public open class MessageListItemViewHolderFactory {
-
     internal lateinit var decoratorProvider: DecoratorProvider
 
     protected lateinit var listenerContainer: MessageListListenerContainer
         private set
 
+    private lateinit var attachmentViewFactory: AttachmentViewFactory
```
zsmb13, line 38: Shouldn't this be `protected` like `listenerContainer`, so that anyone extending `MessageListItemViewHolderFactory` can have access to it to generate Views for attachments?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/attachment/internal/StreamAttachmentViewFactory.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+import io.getstream.chat.android.ui.common.extensions.internal.hasLink
+import io.getstream.chat.android.ui.message.list.adapter.internal.MessageListItemViewTypeMapper.isMedia
+import io.getstream.chat.android.ui.message.list.adapter.view.internal.FileAttachmentsView
+import io.getstream.chat.android.ui.message.list.adapter.view.internal.LinkAttachmentView
+import io.getstream.chat.android.ui.message.list.adapter.view.internal.MediaAttachmentsGroupView
+import io.getstream.chat.android.ui.message.list.adapter.viewholder.attachment.AttachmentViewFactory
+
+@InternalStreamChatApi
+internal class StreamAttachmentViewFactory : AttachmentViewFactory {
```
zsmb13, line 16: Do we need a separate interface and internal implementation? Could we provide an open class to start from, like with `MessageListItemViewHolderFactory`? That way you can add a custom attachment type but otherwise delegate to the default implementation. That's the most basic use case, and it seems difficult with this interface / internal impl that we have so far. Or am I missing something?

Perhaps we could add a sample client implementation of `AttachmentViewFactory` to the `docs` module that implements something similar to the [tutorial's Imgur customization](https://getstream.io/tutorials/android-chat/#your-custom-viewholder), replacing the regular image attachment UI with something custom if the link is from imgur (doesn't have to be the overlay from the tutorial, can be just a red box in place of the image or something).

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/attachment/internal/StreamAttachmentViewFactory.kt
```diff
@@ -0,0 +1,40 @@
...
+    override fun createLinkAttachmentView(linkAttachment: Attachment, context: Context): View {
+        require(linkAttachment.hasLink()) { "Can create link view only for attachments with link" }
```
zsmb13, line 18: Why do links need their own method? Can't we check for whether something is a link in the `createAttachmentsView` method, with the same `hasLink` that we use here?

zsmb13 (changes requested):
Looks like the right direction, but there are a few things I'm not quite seeing how they'll work with this implementation

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/attachment/internal/StreamAttachmentViewFactory.kt
ogkuzmin: Right now we have simultaneously three modes for attachments:
1) Link + Attachment
2) Media attachments only
3) File attachments only

We define that if we have both attachments and links then we should show attachments on the top and link below. If we had links or other attachments then we could have a single method where we put all attachments that we have and it returns a single view for attachments.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/attachment/internal/StreamAttachmentViewFactory.kt
ogkuzmin: Yes, good point. I'll make it as open class

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
ogkuzmin: Will fix that

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/attachment/internal/StreamAttachmentViewFactory.kt
ogkuzmin: Maybe when we'll fuse VHs into one we will be able to make attachment view factory easy. Right now this PR is transit step for attachment customization

zsmb13: I'm a bit worried about already adding this as public API as-is, because we won't be able to release a new version until we've followed it up and changed it. But if we expect to move fast with the follow-up that cleans it up, I guess we can proceed like this. Would really like to see [all this logic](https://github.com/GetStream/stream-chat-android/blob/46cc09082c1290287a1b98629705696f00a5f722/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/PlainTextWithFileAttachmentsViewHolder.kt#L100-L129) moved to the factory if that's possible, so that the factory only exposes one method for clients to override.

Alternatively, we could ask design if it's okay to remove link previews if we already have other attachments on the message.

zsmb13 (approved):
Looks good so far, let's keep it moving 😄

PR #1470, opened 2021-02-24
[CAS-437] Refactor TokenManager and handle custom TokenProvider errors

bychkovdmitry:
### Description

This PR fixes the following issues:
1. `tokenManager.loadAsync` callback was executed on a worker thread
```
tokenManager.loadAsync {
     // This code is now executed on the Main thread
     if (it.isSuccess) {
         clientStateService.onTokenReceived(it.data())
     } else {
         clientStateService.onTokenReceived("")
     }
    connect()
}
```
2. `tokenManager.loadAsync` is now guaranteed to be executed only once which was not the case previously as  `TokenManager` class was not thread-safe
3. Retry `TokenProvider.loadToken()` several time in case of errors
4. Cancel lengthy `loadAsyncAndRetry` job before logout so that loaded token does not arrive when we are in the `Idle` state on the login screen (otherwise will crash)

What's not done:
Unclear what to do after several unsuccessful retries. Maybe we need a retry button in our headers.

### Checklist

- [X] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/token/TokenManagerImpl.kt
```diff
@@ -2,80 +2,89 @@ package io.getstream.chat.android.client.token
...
+    private suspend fun runAndRetry(): String {
+        var attempt = 0
+        var token: String = EMPTY_TOKEN
+
+        while (attempt < ATTEMPT_COUNT) {
+            token = loadSync()
+            if (token != EMPTY_TOKEN) {
+                break
+            } else {
+                delay(RETRY_DELAY_MILLIS)
```
JcMinarro, line 78: Could we use an Exponential backoff?

bychkovdmitry: Yes, will do. And later I will review our retry use cases, maybe we can come up with a neat solution.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -170,28 +174,36 @@ internal class ChatSocketServiceImpl private constructor(
...
+                }
+                is ConnectionConf.UserConnectionConf -> {
+                    state = State.Connecting
+                    socketConnectionJob = GlobalScope.launch(DispatcherProvider.IO) {
+                        tokenManager.ensureTokenLoaded()
+                        withContext(DispatcherProvider.Main) {
+                            socket = socketFactory.createNormalSocket(endpoint, apiKey, user)
+                        }
+                    }
+                }
```
bychkovdmitry, line 200: The job needs to cancelled as `tokenManager` can load token for quite a while and the following will have no effect
```
 private fun shutdownSocketConnection() {
        socket?.close(1000, "Connection close by client")	       
        socket = null	       
    }	
```

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/ClientStateServiceTests.kt
```diff
@@ -45,47 +45,37 @@ internal class ClientStateServiceTests {
     }
 
     @Test
-    fun `Given Idle state When set user Should move to state user pending without token`() {
+    fun `Given Idle state When set user Should move to state user pending`() {
```
bychkovdmitry, line 48: These tests are not that sophisticated now, can be simplified later

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/token/ConstantTokenProvider.kt
```diff
@@ -0,0 +1,5 @@
+package io.getstream.chat.android.client.token
+
```
bychkovdmitry, line 2: This name seems more descriptive to me

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -170,28 +174,36 @@ internal class ChatSocketServiceImpl private constructor(
...
+            }.exhaustive
+        }
     }
```
bychkovdmitry, line 204: Either token is loading, then cancel job. Or socket has already been created, then shutdown it

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/errors/ChatErrorCode.kt
```diff
@@ -16,9 +14,23 @@ public enum class ChatErrorCode(public val code: Int, public val description: St
     UNABLE_TO_PARSE_SOCKET_EVENT(1008, "Socket event payload either invalid or null"),
     NO_ERROR_BODY(1009, "No error body. See http status code"),
 
-    /**
-     * Backend
-     */
+    // server error codes
+    AUTHENTICATION_ERROR(5, "Unauthenticated, problem with authentication"),
     TOKEN_EXPIRED(40, "Token expired, new one must be requested."),
```
bychkovdmitry, line 19: Maybe just use 401 http code? 
https://getstream.io/chat/docs/android/api_errors_response/?language=kotlin&q=token

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -94,31 +100,29 @@ internal class ChatSocketServiceImpl private constructor(
...
-        ChatErrorCode.CANT_PARSE_CONNECTION_EVENT.code,
-        ChatErrorCode.CANT_PARSE_EVENT.code,
-        ChatErrorCode.UNABLE_TO_PARSE_SOCKET_EVENT.code,
-        ChatErrorCode.NO_ERROR_BODY.code -> {
-            // Nothing to do on that case
-        }
-        ChatErrorCode.TOKEN_EXPIRED.code -> {
+    private fun onChatNetworkError(error: ChatNetworkError) {
+        if (ChatErrorCode.isAuthenticationError(error.streamCode)) {
             tokenManager.expireToken()
```
JcMinarro, line 105: Don't we load a new one?

JcMinarro (approved):
LGTM

PR #1471, opened 2021-02-24
[CAS-699] Attachments and text VH

ogkuzmin:
### Description

This PR fuses all existing VH with attachments into one `TextAndAttachmentsVH`. So now we have:
1) `MessagePlainTextVH` is real plain text without any kind of attachments
2) `TextAndAttachmentsVH` is VH that contain text (optional) and some attachment(s) (mandatory)

Changes:
1) Introduced new `TextAndAttachmentsVH`
2) Removed container for attachments from `MessagePlainTextVH` layout and required logic for that
3) Deleted all old VHs for attachments (media, files, etc)
4) Changed `BaseDecorator` and its inheritors in order to decorate new VH
5) Apply text style for `LinkAttachmentView`

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/TextAndAttachmentsViewHolder.kt
```diff
@@ -71,36 +72,47 @@ internal class PlainTextWithFileAttachmentsViewHolder(
...
-            .attachments
-            .filter { attachment -> attachment.uploadState == Attachment.UploadState.InProgress }
-            .mapNotNull { attachment -> attachment.uploadId }
+        setupAttachment(data)
+        setupUploads(data)
+    }
 
-        val needUpload = uploadIdList.isNotEmpty()
+    private fun setupAttachment(data: MessageListItem.MessageItem) {
+        binding.attachmentsContainer.removeAllViews()
```
ogkuzmin, line 83: All related attachments code will go to the attachment view factory. I'm going to make `AttachmentViewFactory` easier only with one method for creating attachment content view (some `ViewGroup` that contain both link attachment and other attachment type view)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
```diff
@@ -38,6 +33,8 @@ public open class MessageListItemViewHolderFactory {
     protected lateinit var attachmentViewFactory: AttachmentViewFactory
         private set
 
+    private lateinit var style: MessageListItemStyle
```
zsmb13, line 36: This could be `internal` and then we wouldn't need a setter function for it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
```diff
@@ -82,6 +80,17 @@ public open class MessageListItemViewHolderFactory {
         }
     }
 
+    private fun createTextAndAttachmentViewHolder(parentView: ViewGroup): BaseMessageItemViewHolder<out MessageListItem> {
```
zsmb13, line 83: Should we make this `protected` so that clients can call it, like the other similar methods here?

stream-chat-android-ui-components/api/stream-chat-android-ui-components.api
```diff
@@ -813,14 +813,10 @@ public final class io/getstream/chat/android/ui/message/list/adapter/MessageList
 
 public class io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory {
 	public fun <init> ()V
-	protected final fun createAttachmentsViewHolder (Landroid/view/ViewGroup;)Lio/getstream/chat/android/ui/message/list/adapter/BaseMessageItemViewHolder;
```
zsmb13, line 816: Changes to public API here should be in the CHANGELOG in some form - though I expect very few people are already customizing things with the new UI, this is still breaking

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
ogkuzmin: Let it leave as is. I'm going to expose it as protected in next PR with AttachmentViewFactory improvement

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
ogkuzmin: Fixed

stream-chat-android-ui-components/api/stream-chat-android-ui-components.api
ogkuzmin: Fixed

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
zsmb13: Are we making the `MessageListItemStyle` class itself public then?

ogkuzmin: Yes, I'll have to do that. I see we had the same in old sample app

zsmb13 (approved):
LGTM

PR #1479, opened 2021-02-26
Uploads, more moderation, slow mode in Moshi API

zsmb13:
### Description

~Moshi goes brrrr~

Adds a couple new methods to our Moshi based API implementation.

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/GsonChatApi.kt
```diff
@@ -621,7 +620,7 @@ internal class GsonChatApi(
         channelType: String,
         channelId: String,
         shadow: Boolean,
-    ): Call<CompletableResponse> {
+    ): Call<Unit> {
```
leandroBorgesFerreira, line 623: Why not to change the signature of `RetrofitApi::banUser`?

zsmb13: The Retrofit interfaces return models that represent the raw API responses, that's fine as-is. They just shouldn't be leaked further out, outside of these `ChatApi` implementations, because they're internal details of each `ChatApi`.

In this case, I had to replace it with `Unit` on the public interface because `MoshiChatApi` uses a different `CompletableResponse` class internally.

leandroBorgesFerreira (approved):
🐑 🐑 🐑

PR #1480, opened 2021-02-26
[CAS-752] domain deprecate user

ogkuzmin:
### Description

We don't need to explicitly pass user to `ChatDomainImpl` anymore. Now user is obtained through listeners for `ChatClient`. 

1) Deprecate constructors of `ChatDomain.Builder` with user param
2) Remove user from `ChatDomainImpl` constructor
3) Update the deprecation table and doc example. Update CMS accordingly

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -283,24 +279,21 @@ internal class ChatDomainImpl internal constructor(
         logger.logI("Initializing ChatDomain with version " + getVersion())
 
         // if the user is already defined, just call setUser ourselves
-        val current = userOverwrite ?: client.getCurrentUser()
```
zsmb13, line 286: How do the changes in this class not break existing behaviour for users who have been using the three-param constructor, passing in a `User` object to the builder? We were previously performing logic based on that parameter, and now we're ignoring it.

ogkuzmin: Even if they passed user they could do nothing unless they invoke `ChatClient::connectUser`. IMO it was useless to have such logic from the moment when we had brought these listeners for set/unset users in ChatClient, wasn't it?

ogkuzmin: Also, it was used for such style, when you set/unset user you should create a new instance of ChatDomain. It can work with these changes as well

zsmb13: I'm honestly not quite sure what behaviour changes (if any) removing this will do, but it'd be easier and safer to just leave this code in place, and remove it later when we remove the public-facing API.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -283,24 +279,21 @@ internal class ChatDomainImpl internal constructor(
...
+        val current = client.getCurrentUser()
```
bychkovdmitry, line 282: Client user will only be available for anonymous users when socket is connectd

stream-chat-android-offline/api/stream-chat-android-offline.api
```diff
@@ -80,13 +80,13 @@ public abstract interface class io/getstream/chat/android/livedata/ChatDomain {
 }
 
 public final class io/getstream/chat/android/livedata/ChatDomain$Builder {
+	public fun <init> (Landroid/content/Context;Lio/getstream/chat/android/client/ChatClient;)V
 	public fun <init> (Landroid/content/Context;Lio/getstream/chat/android/client/ChatClient;Lio/getstream/chat/android/client/models/User;)V
-	public synthetic fun <init> (Landroid/content/Context;Lio/getstream/chat/android/client/ChatClient;Lio/getstream/chat/android/client/models/User;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
-	public fun <init> (Lio/getstream/chat/android/client/ChatClient;Landroid/content/Context;)V
```
zsmb13, line 85: It looks like we lost this constructor that had two parameters, `ChatClient` and then `Context`. This will break any calls in client code that look like `ChatDomain.Builder(client, applicationContext)`, just like they broke our docs samples. We should keep this constructor, and make it deprecated if we really want to make everyone move over to using `ChatDomain.Builder(applicationContext, client)` instead.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -97,14 +97,36 @@ public interface ChatDomain {
...
+            this.user = user
+        }
+
+        @Deprecated(
+            message = "Use constructor without user",
+            replaceWith = ReplaceWith("Use ChatDomain.Builder(appContext, chatClient) instead")
+        )
+        public constructor(client: ChatClient, appContext: Context, user: User?) : this(appContext, client) {
+            this.user = user
+        }
```
zsmb13, line 121: This looks like a brand new constructor that we didn't have before

ogkuzmin: Previously it was the main one

zsmb13: That was context/client/user, which we have above. This one is client/context/user.

It's new according to the API diff as well:

```diff
+ public fun <init> (Lio/getstream/chat/android/client/ChatClient;Landroid/content/Context;Lio/getstream/chat/android/client/models/User;)V
```

ogkuzmin: Oh, I see

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/cookbook/ui/Setup.kt
```diff
@@ -17,7 +17,7 @@ class Setup {
             .logLevel(ChatLogLevel.ALL)
             .build()
 
-        val domain = ChatDomain.Builder(client = client, appContext = applicationContext)
+        val domain = ChatDomain.Builder(client, applicationContext)
```
zsmb13, line 20:
```suggestion
        val domain = ChatDomain.Builder(applicationContext, client)
```

zsmb13 (approved):
One more suggestion to make all the samples use the same style for creating a builder, everything else looks good

PR #1483, opened 2021-02-26
Update pull_request_template.md

zsmb13:
Update the PR template based on our discussions.

New template preview: https://github.com/GetStream/stream-chat-android/blob/update-pr-template/.github/pull_request_template.md


.github/pull_request_template.md
```diff
@@ -1,14 +1,34 @@
-< Jira issue link, if applicable >
+_Jira issue link, if applicable_
```
adasiewiczr, line 1: Maybe it would be worth it to add here an additional line for the Github issue (e.g. _closes XYZ_). This will ensure that everyone will always remember to mention it and the PR will be bound to the issue?
On the other hand, we don't need that in most cases because we don't have so many issues

zsmb13: Will skip this for now

.github/pull_request_template.md
```diff
@@ -1,14 +1,35 @@
...
+
+### ☑️ Checklist
 
 - [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
 - [ ] PR targets the `develop` branch
-- [ ] Changelog updated with client-facing changes
+- [ ] Changelog is updated with client-facing changes
 - [ ] New code is covered by unit tests
 - [ ] Comparison screenshots added for visual changes
+- [ ] Affected documentation updated (CMS, cookbooks, tutorial)
```
bychkovdmitry, line 30: 👍

.github/pull_request_template.md
```diff
@@ -1,14 +1,35 @@
...
+_Jira issue link, if applicable_
 
-### Description
+### 🎯 Goal
 
-Describe what this PR changes, why we're making the change, how it should be tested...
+_Describe why we are making this change_
 
-### Checklist
+### 🛠 Implementation details
```
bychkovdmitry, line 7: Maybe just `Changes`. The `Goal` will cover the `why`. The `Changes` section will cover `what` and `how`. E.g.
`Goal`
Improve sign-in form
`Changes`
Added Csrf token to the template to prevent Csrf attack

bychkovdmitry (approved):
Overall very good, I like that we don't have too many sections

.github/pull_request_template.md
zsmb13: I think "Implementation details" is descriptive enough, I wouldn't add the "why" of the task in that section, only the "what" and "how", as you suggested.

adasiewiczr (approved):

PR #1496, opened 2021-03-01
[CAS-758] New filter object Structure

JcMinarro:
### Description
Create a new structure for our FilterObject using a sealed class

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/FilterObject.kt
```diff
@@ -0,0 +1,33 @@
+package io.getstream.chat.android.client.api.models
+
+public sealed class FilterObject
```
zsmb13, line 3: What will the usage of these new classes look like? Are clients expected to create these directly?

JcMinarro: It Will be used by `Filters` class, but it will be cover in another PR

adasiewiczr, line 3: Are we going to provide a way to create custom filters (for example searching for specific distinct channel: https://github.com/GetStream/stream-chat-android/blob/main/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/group/member/GroupChatInfoMemberOptionsViewModel.kt#L41)? If not, it would be great to double-check if we are not missing something

zsmb13: Should/could all these classes be `internal` then, at least for now?

JcMinarro: Ok, I can open on the next PR

JcMinarro: This custom filter should be the same than:
```
Filters.and(
            Filters.eq("distinct", true),
            Filters.eq("members", listOf(memberId, chatDomain.currentUser.id))
        )
```

With the new Structure it will be:
```
AndFilterObject(
            listOf(
                EqualsFilterObject("distinct", true),
                EqualsFilterObject("members", listOf(memberId, chatDomain.currentUser.id))
            )
        )
```

We will update `Filters` class to use the new structure, so they will be able to use both way

adasiewiczr (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/FilterObject.kt
adasiewiczr: I think I tried to use two `Filters.eq` instead of custom `FilterObject` and then the query was failing, but we can check that later on.

JcMinarro: I will check before my next PR where I will remove old FilterObject class

PR #1499, opened 2021-03-02
[CAS-751] Fix refreshing channel list after removing member

### LINKED ISSUE #1473: [Old UI] Channel List doesn't update if current-user is removed from a channel
When current-user is on channel-list page, and he gets removed from one of the channels, channel list doesn't update. But, if he is added to a channel, channel list updates to show the new channel. Channel list, also, updates if channel is deleted.

Keeping the channel history visible to removed user might be an intention. But, there is nothing like `readOnlyMode` in `MessageListView`, right? Also, if I disconnect the chat-client and set the user again, channel (from which user was removed) disappears. So, this doesn't seem like planned intention.

adasiewiczr:
Fixes #1473 

### Description
`QueryChannelsControllerImpl::refreshChannels` wasn't checking if the current user is still a member of the channel. 
I've added an additional check to ensure that we skip channels to which the current user doesn't belong to.

### Checklist

- [x] Reviewers added

### DIFF
--- a/UNRELEASED_CHANGELOG.md
+++ b/UNRELEASED_CHANGELOG.md
@@ -36,6 +36,7 @@
 
 ## stream-chat-android-offline
 ### 🐞 Fixed
+- Fixed refreshing channel list after removing member
 
 ### ⬆️ Improved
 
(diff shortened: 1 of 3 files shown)


ogkuzmin (approved):
Good job and tests! 👍

PR #1502, opened 2021-03-03
[CAS-754] Setup mock web server

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-754

### Description

Sample integration test using `MockWebServer`

### Checklist

- [X] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1304,6 +1308,13 @@ public class ChatClient internal constructor(
             warmUp = false
         }
 
+        @VisibleForTesting
```
bychkovdmitry, line 1311: Sorry for that. Better to add TLS support to `MockWebServer` (see `MockWebServer.useHttps()` method below). However, in that case, client needs to trust this self-signed certificate.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/mockwebserver/BaseMockWebServerIntegrationTest.kt
```diff
@@ -0,0 +1,219 @@
...
+    private fun connectUser() = runBlocking {
+        waitForSetUser(client, data.user, data.token)
+        Truth.assertThat(client.isSocketConnected()).isTrue()
+    }
+
+    /**
+     * Enables TLS support for mock web server. Doesn't work unless client is set up
+     * to trust self-signed certificates
+     */
+    private fun MockWebServer.useHttps() {
```
bychkovdmitry, line 209: Temporary unused

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/mockwebserver/BaseMockWebServerIntegrationTest.kt
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+    }
+
+    /**
+     * Sets the current user and awaits for network connection to be established.
+     * TODO: check why this doesn't work without a delay in `waitForSetUser`
+     */
+    private fun connectUser() = runBlocking {
+        waitForSetUser(client, data.user, data.token)
```
bychkovdmitry, line 201: `waitFirSetUser()` needs to be fixed as it slows down all integration tests

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
adasiewiczr: I think we should at least mark it with `@InternalStreamChatApi` annotation

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/mockwebserver/SendMessageUseCaseTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import kotlinx.coroutines.runBlocking
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.RecordedRequest
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+internal class SendMessageUseCaseTest : BaseMockWebServerIntegrationTest() {
+
+    override fun dispatchApiRequest(request: RecordedRequest): MockResponse {
```
adasiewiczr, line 18: How are we going to handle a case when inside one test class we want to check what happens when a message is sent successfully and what happens when a request fails?
Maybe it would be a good idea to replace this abstract method with the property that will allow configuring responses inside each test method?

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/mockwebserver/SendMessageUseCaseTest.kt
```diff
@@ -0,0 +1,50 @@
...
+            .execute()
+
+        assertSuccess(sendMessageResult)
+
+        val messagesFromDatabase = chatDomainImpl.repos.selectMessages(listOf(messageId))
+
+        Truth.assertThat(messagesFromDatabase.size).isEqualTo(1)
+        Truth.assertThat(messagesFromDatabase[0].syncStatus).isEqualTo(SyncStatus.COMPLETED)
+        Truth.assertThat(messagesFromDatabase[0].text).isEqualTo("message test")
+        Truth.assertThat(messagesFromDatabase[0].user.name).isEqualTo("Zetra")
```
adasiewiczr, line 48: I think we should allow configuring all those fields inside `createSendMessageRequestJsonString`. I was confused why does user name should be equal to `Zetra` because `Message ` doesn't have this property

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/mockwebserver/BaseMockWebServerIntegrationTest.kt
bychkovdmitry: This was fixed in another PR

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: IMO it's ok to expose it as `@InternalStreamChatApi`. If someone strongly decided to suppress such warning and use it then it is his fault

bychkovdmitry: I guess there will be no warnings in java

leandroBorgesFerreira (approved):
Let's use it, we can improve upon later =]

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/mockwebserver/SendMessageUseCaseTest.kt
```diff
@@ -0,0 +1,50 @@
...
+import okhttp3.mockwebserver.RecordedRequest
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+internal class SendMessageUseCaseTest : BaseMockWebServerIntegrationTest() {
+
+    override fun dispatchApiRequest(request: RecordedRequest): MockResponse {
+        return if (request.isRequestBodyEqualTo(createSendMessageRequestJsonString())) {
+            MockResponse()
```
leandroBorgesFerreira, line 20: The MockResponse().setResponseCode([some int]).setBody("Some string") and this else this `IllegalStateException` will be repeated in every test. You can make the user provide just a predicate and a Pair with response code and responde body then you create the `MockResponse` inside the`BaseMockWebServerIntegrationTest`.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/mockwebserver/BaseMockWebServerIntegrationTest.kt
```diff
@@ -106,25 +110,26 @@ internal abstract class BaseMockWebServerIntegrationTest {
...
+                        return MockResponse().withWebSocketUpgrade(object : WebSocketListener() {
+                            override fun onOpen(webSocket: WebSocket, response: Response) {
+                                super.onOpen(webSocket, response)
+                                // very likely that this instance of `WebSocket`
+                                // can be used to send further events
+                                webSocket.send(createConnectedEventStringJson())
+                            }
+                        })
+                    }
+                    serverCalls.forEach {
```
leandroBorgesFerreira, line 127: No need for `forEach`, you can use find.
Someting like:

```
return if (isWebSocketUrl(request)) {
                        return MockResponse().withWebSocketUpgrade(object : WebSocketListener() {
                            override fun onOpen(webSocket: WebSocket, response: Response) {
                                super.onOpen(webSocket, response)
                                // very likely that this instance of `WebSocket`
                                // can be used to send further events
                                webSocket.send(createConnectedEventStringJson())
                            }
                        })
                    } else {
                        serverCalls.find { it.isApplicable(request) }?.executeCall()
                            ?: throw IllegalStateException("Request not supported")
                    }
```

leandroBorgesFerreira (approved):
🐑

adasiewiczr (approved):

PR #1516, opened 2021-03-05
[CAS-685] Tests - Test reaction delete

leandroBorgesFerreira:
[CAS-685](https://stream-io.atlassian.net/browse/CAS-685)

### Description

New tests that cover the delete reaction logic.

### Checklist

- ~[ ] New code is covered by unit tests~ -> **The are unit tests =P**
- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelControllerImpl.kt
```diff
@@ -942,15 +942,15 @@ internal class ChannelControllerImpl(
             val result = domainImpl.runAndRetry { client.deleteReaction(reaction.messageId, reaction.type) }
             return if (result.isSuccess) {
                 reaction.syncStatus = SyncStatus.COMPLETED
-                domainImpl.repos.insertReaction(reaction)
```
JcMinarro, line 945: Why do you remove the insertion into our database?

leandroBorgesFerreira: This code gets called in the line 931 and there's no condition for it to run, so this line is just a duplication.

JcMinarro: But it update the `syncStatus` to be updated on the DB

leandroBorgesFerreira: Oh, I didn't realise that this new insertReaction is to propagate the new sync status.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/controller/ChannelControllerImplReactionsTest.kt
```diff
@@ -131,12 +144,77 @@ internal class ChannelControllerImplReactionsTest {
...
+                verify(mockReaction).syncStatus = SyncStatus.SYNC_NEEDED
+            }
+
+            result.isSuccess `should be equal to` true
+
+            verify(sut.domainImpl.repos).insertReaction(mockReaction)
+        }
+
+    @Test
+    fun `when delete reaction, the sync status is right and reaction is inserted`() =
```
adasiewiczr, line 170: I think it would be nice to rename it to make it more descriptive, e.g. `when deleting reaction failed, the sync status is right and reaction is inserted`. 
The current name looks almost the same as the next test (when deleting a reaction, the sync status is right and message is inserted) I had to review the implementation to spot the difference

stream-chat-android-offline/src/test/java/io/getstream/chat/android/livedata/controller/ChannelControllerImplReactionsTest.kt
```diff
@@ -131,12 +144,77 @@ internal class ChannelControllerImplReactionsTest {
...
+                .get()
+
+            whenever(sut.domainImpl.isOnline()) doReturn true
+            whenever(sut.domainImpl.runAndRetry<Message>(any())) doAnswer { Result(ChatError()) }
+
+            val result = sut.deleteReaction(mockReaction)
+
+            inOrder(mockReaction) {
+                verify(mockReaction).syncStatus = SyncStatus.IN_PROGRESS
+                verify(mockReaction).syncStatus = SyncStatus.SYNC_NEEDED
```
adasiewiczr, line 184: I think it would be nice to add an additional test for `SyncStatus.FAILED_PERMANENTLY` case

PR #1523, opened 2021-03-08
[CAS-764] Fix/user not stored

JcMinarro:
### Description
Ensure that all related users are stored before obtain messages/reactions from our DB

### Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/RepositoryFacade.kt
```diff
@@ -73,13 +75,35 @@ internal class RepositoryFacade constructor(
...
     }
 
     override suspend fun insertChannels(channels: Collection<Channel>) {
-        channelsRepository.insertChannels(channels)
         insertUsers(channels.flatMap(Channel::users))
+        channelsRepository.insertChannels(channels)
+    }
+
+    override suspend fun insertMessage(message: Message, cache: Boolean) {
+        insertUsers(message.users())
```
leandroBorgesFerreira, line 88: it it odd that insertUsers is called inside insertMessage, this is causing a side effect that I wouldn't normally expect.

samiuelson (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/repository/RepositoryFacade.kt
JcMinarro: It is because `Message` has an entity `user` inside of it, but our DB doesn't store nested object, so we need to be sure that our nested object are stored on the proper table

PR #1524, opened 2021-03-08
Implement search and query API calls with Moshi

zsmb13:
### Description

Implements all but two of the remaining `MoshiChatApi` methods: everything search and querying related. (The remaining two methods will be added once event parsing is implemented, in the next step.)

Other than the addition of the new endpoints, the following changes/fixes were necessary:
- Lots of new default values for DTOs that weren't needed / couldn't be tested before
- A custom Retrofit adapter for the Moshi implementation that's able to Moshi-serialize query parameters
- CustomObject behaviour via custom adapters for ChannelDto
- Null handling in CustomObject parsing

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1,6 +1,13 @@
 package io.getstream.chat.android.client.api2
 
 import io.getstream.chat.android.client.api.ChatApi
+import io.getstream.chat.android.client.api.ErrorCall
+import io.getstream.chat.android.client.api.models.FilterObject
+import io.getstream.chat.android.client.api.models.QueryChannelRequest
+import io.getstream.chat.android.client.api.models.QueryChannelsRequest
+import io.getstream.chat.android.client.api.models.QuerySort
+import io.getstream.chat.android.client.api.models.QueryUsersRequest
+import io.getstream.chat.android.client.api.models.SearchMessagesRequest
```
zsmb13, line 10: Really unfortunate, but these old network models are part of the API of the ChatAPI interface, so we have to reference them in the Moshi impl too, before converting them to new, proper DTOs

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser/FilterObjectTypeAdapter.kt
```diff
@@ -35,7 +35,7 @@ internal class FilterObjectTypeAdapter(val gson: Gson) : TypeAdapter<FilterObjec
     }
 }
 
-private fun FilterObject.toMap(): Map<*, *> = when (this) {
+internal fun FilterObject.toMap(): Map<*, *> = when (this) {
```
zsmb13, line 38: Decided to expose this logic within the module instead of duplicating it in the Moshi packages - we can duplicate it later if we want to decouple, but it's probably enough to keep it around for now & move it when the Gson impl is removed

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/requests/QueryChannelsRequest.kt
```diff
@@ -0,0 +1,16 @@
+package io.getstream.chat.android.client.api2.model.requests
+
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+internal class QueryChannelsRequest(
+    val filter_conditions: Map<*, *>,
```
zsmb13, line 7: Strongly typed API 😉

adasiewiczr (approved):
LGTM

PR #1552, opened 2021-03-15
Add possibility to query banned users in LLC

adasiewiczr:
### Description
Add support for querying banned users in LLC:
- Currently trying to query banned users fails because of the wrong type inside `Response::Channel::Config::Commands` (should be an array of objects instead of strings). This will be fixed on the backend side. If you want to test it - you can comment `BannedUser::channel` for now
- Docs will be updated in a separate PR

### Checklist

- [x] Reviewers added


bychkovdmitry (approved):
Looks good. Unfortunately not much is returned from the server ATM. As you've said commands have incorrect format, `expires` field is not returned, etc. So, not sure if we need to advertise the change in the changelog.

I would also add `ban_expires` field to `User` entity later.

PR #1555, opened 2021-03-15
Review channel config fields

bychkovdmitry:
### Description

Added new fields to `Channel.Config` entity:
- `uploadsEnabled`
-  `urlEnrichmentEnabled`
-  `customEventsEnabled`
-  `pushNotificationsEnabled`
-  `messageRetention`
-  `automodBehavior`
-  `blocklistBehavior`

Removed the `infinite` field from `Channel.Config` entity. Not mentioning it in `deprecations.md` as this field is useless as it is never returned from the server.

### Checklist

- [X] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Config.kt
```diff
@@ -4,39 +4,109 @@ import com.google.gson.annotations.SerializedName
...
+     * Determines if URL enrichment enabled to show they as attachments. Enabled by default.
+     */
+    @SerializedName("url_enrichment")
+    val isUrlEnrichment: Boolean = false,
+
+    @SerializedName("custom_events")
+    val isCustomEvents: Boolean = false,
 
+    @SerializedName("push_notifications")
+    val isPushNotifications: Boolean = false,
```
ogkuzmin, line 83: It's good to mention in `changelog`, that now we support such fields in our offline module

ogkuzmin (approved):

UNRELEASED_CHANGELOG.md
```diff
@@ -40,7 +41,7 @@ It is not possible to remove users from distinct channels anymore.
 - Deprecated `ChannelController::unBanUser`, the `ChannelController::unbanUser` method should be used instead
 
 ### ❌ Removed
-
+- Removed `infinite` field from channel config
```
zsmb13, line 44: So, we're mentioning it after all? PR description said we're not mentioning it on purpose 😄

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Config.kt
```diff
@@ -4,39 +4,109 @@ import com.google.gson.annotations.SerializedName
...
+     */
     @SerializedName("mutes")
-    var isMutes: Boolean = false,
+    val isMutes: Boolean = false,
+
+    /**
+     * Allows image and file uploads within messages. Enabled by default.
+     */
+    @SerializedName("uploads")
+    val isUploads: Boolean = false,
```
zsmb13, line 71: These property names like `isUploads` are really awkward to read, something like `areUploadsEnabled` or just `uploadsEnabled` would be much nicer, but now might not be the time to fix that

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Config.kt
```diff
@@ -4,39 +4,109 @@ import com.google.gson.annotations.SerializedName
...
     @SerializedName("search")
-    var isSearch: Boolean = false,
+    val isSearch: Boolean = false,
 
+    /**
+     * If users are allowed to add reactions to messages. Enabled by default.
+     */
     @SerializedName("reactions")
-    var isReactionsEnabled: Boolean = false,
+    val isReactionsEnabled: Boolean = false,
```
zsmb13, line 53: > Enabled by default.
> `= false`

This looks really confusing. I know these default values will probably never be used, but should we attempt to align them with what the default behaviour is?

bychkovdmitry: Need to check that our tests will make sense after that. I guess everything is `false` by default here and we are testing that updated config has all fields set to `true`

UNRELEASED_CHANGELOG.md
bychkovdmitry: Ultimately removed this line from the changelog. We don't want to advertise that we make such careless mistakes.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Config.kt
bychkovdmitry: Fixed all the fields which we added. Will be easier to deprecate the rest of the fields one day.

zsmb13 (approved):
LGTM

PR #1559, opened 2021-03-16
Merging the stream_ui_message_list_view so a container is removed from the hierarchy and the background is easier to customise for the clients.

dmarin:
### Description

Currently the stream_ui_message_list_view is hardcoding a background in the first child. This makes hard for the clients to override the color as it would leak implementation details to the clients apps.

Ie to change the color prior to this PR a client would need to do something similar to this:

`messageListView[0].background = null`

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
```diff
@@ -377,14 +378,15 @@ public class MessageListView : ConstraintLayout {
...
-        binding = StreamUiMessageListViewBinding.inflate(context.inflater, this, true)
+        binding = StreamUiMessageListViewBinding.inflate(context.inflater, this)
 
         initRecyclerView()
         initScrollHelper()
         initLoadingView()
         initEmptyStateView()
 
         configureAttributes(attr)
+        layoutTransition = LayoutTransition()
```
leandroBorgesFerreira, line 389: Why is this needed?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
```diff
@@ -454,6 +456,9 @@ public class MessageListView : ConstraintLayout {
 
         configureMessageOptions(tArray)
         tArray.recycle()
+        if (background == null) {
+            setBackgroundColor(ContextCompat.getColor(context, R.color.stream_ui_white_snow))
```
leandroBorgesFerreira, line 460: This way you're changing the background of the parent ConstraintLayout right? I don't thinkg that binding the background of both views should be done... If that's the case, we could simply never set the background for this view and it would simply propagate the background of the container.

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_list_view.xml
```diff
@@ -1,11 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
```
leandroBorgesFerreira, line 2: Some users would like to control the background of this view separately 🤔. We need to think about that.

leandroBorgesFerreira:
We could simply avoid de default color of the `MessageListView`... then we simply set the background color in the XML when declaring the view

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
dmarin: to keep the behaviour, previously in the XML there was a `android:animateLayoutChanges=true`. This is the programatic approach to do it. 
Personally I would be fine removing it.

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_list_view.xml
dmarin: mmmm, that is exactly my use case. I need the `MessageListView` to have a null background. I am a bit unsure about why this PR does not allow users to do it?

dmarin:
> Hey @dmarin, super thanks for helping us and participating in our development. Also, welcome here =].
> 
> So, I have a different approach for this problem. We could simply avoid the default color of the `MessageListView`... then we simply set the background color in the XML when declaring the view. Take a look: #1577
> 
> We lose the default color of the view, which is a not so cool, so in the end is just so much easier to change the background... What do you think? Makes sense?

Hi Leandro, thanks a lot for taking the time to review this. In terms of your PR that definitely solves the issue, but I still think adding the 2 `ConstraintLayouts` is not needed. 

Anyway, I would be happy to go with your proposal as that would allow me to customise the bg color. 

BTW same fix would be needed in the `MessageInputView`

leandroBorgesFerreira:
If we use `merge` instead of `ConstraintLayout` in `MessageListView` and `MessageInputView` I don't think it will be possible to have different colors for then in the same screen, because both will simply reproduce its parent `ConstraintLayout` color... Although there's a performance gain, this limitation is something that we can't have (in my point of view).

leandroBorgesFerreira:
I'm taking with other engineers about this problem (by the way, thanks for raising this). I'll wait for their input too.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
leandroBorgesFerreira: Makes sense. Let's keep it.

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_list_view.xml
leandroBorgesFerreira: I misunderstand your PR. I thought you were merging the views with the parent ConstraintLayout in `fragment_chat.xml`. Yep, merging with the ConstraintLayout of `MessageListView` makes total sense and it is way better

leandroBorgesFerreira (approved):
I misunderstand your PR. I thought you were merging the views with the parent ConstraintLayout in `fragment_chat.xml`. Your proposal corrects an error, there's no draw back. Let's 🚢

leandroBorgesFerreira:
Hi @dmarin. Before I merge this... did you sign the Stream CLA? Ir you don't, can you please do it and mark the check box?

dmarin:
> Hi @dmarin. Before I merge this... did you sign the Stream CLA? If you don't, can you please do it and mark the check box?

Signed and checkbox ticked. Thanks a lot!

leandroBorgesFerreira:
Thanks for you contribution!

PR #1561, opened 2021-03-16
[CAS-781] Add query banned users docs

adasiewiczr:
### Description
- [Docs](https://getstream.io/chat/docs/react/moderation/?language=kotlin)
- [LLC Cookbook](https://github.com/GetStream/stream-chat-android/wiki/Client-Cookbook#retrieving-banned-users-from-specific-channels)

### Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1141,6 +1141,7 @@ public class ChatClient internal constructor(
     }
 
     @CheckResult
+    @JvmOverloads
```
bychkovdmitry, line 1144: I think @JvmOverloads is not very helpful for java users as they still can't use the function with arbitrary parameters. E.g. somebody wants to use only last 2 parameters, still needs to pass something for the rest.

adasiewiczr: True but if you want to use only the first parameter, you don't have to pass others. I think that usually, users are going to use the first few parameters and JvmOverloads is helpful in such cases.

bychkovdmitry: Ok, I'm just worried that only first 1 or 2 overloads are useful, and others will mostly like not match the exact client's needs.

bychkovdmitry (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
adasiewiczr: yeah :( I'll leave it for now but, maybe we should somehow group pagination-related parameters into separate classes 🤔

PR #1572, opened 2021-03-17
Spike: Approach to decouple ui-components from ChatUI

samiuelson:
https://stream-io.atlassian.net/browse/CAS-731

### Description

Refactor which decouples `ui-components` from `ChatUI` and its dependencies.

Key changes:
1. `ChatUI` (also `ChatImpl` and `Chat`) moved to `chat-android` module
2. `ChatUI` dependencies: `navigation`, `style`, and `utils.strings` packages moved to `chat-android` module too
3.  Classes essential for new ui components: `ChatFonts`, `ChatFontsImpl`, and `TextStyle` duplicated and moved into `ui-components` module 

Other changes: 
1. `UrlSigner` and `ChatMarkdown` instances are set using `MessageListView`'s setter functions and are injected to other places via constructors rather than directly from `ChatUI.instance()` invoked all-around the codebase.

### Checklist

- [ ] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/gallery/AttachmentGalleryActivity.kt
```diff
@@ -250,13 +252,15 @@ public class AttachmentGalleryActivity : AppCompatActivity() {
...
+        private const val EXTRA_KEY_ATTACHMENT_URL_SIGNER: String = "extra_key_url_signer"
 
         internal const val EXTRA_ATTACHMENT_OPTION_RESULT = "extra_attachment_option_result"
 
         @JvmStatic
-        public fun createIntent(context: Context, initialIndex: Int): Intent {
+        public fun createIntent(context: Context, initialIndex: Int, attachmentUrlSigner: UrlSigner): Intent {
             return Intent(context, AttachmentGalleryActivity::class.java).apply {
                 putExtra(EXTRA_KEY_INITIAL_INDEX, initialIndex)
+                putExtra(EXTRA_KEY_ATTACHMENT_URL_SIGNER, attachmentUrlSigner)
```
ogkuzmin, line 263: Do not like this approach, because it's easy forget to implement `Serializable` interface for custom `UrlSigner`. What if pass signer to view model constructor and sign image previews there?

ogkuzmin (commented):
I don't know why we decided to do such changes. But my gut feelings that it becomes more complicated to customers to use our SDK. BTW, If we don't want to have `ChatUI` how it was before I'd like at least to have static singletones for fonts, markdown, urlSigner and etc

PR #1573, opened 2021-03-17
[CAS-801] Feature/open avatar bitmap factory

JcMinarro:
### Description
Allow custom implementation of our `AvatarBitmapFactory`

### Checklist

- [x] Reviewers added


ogkuzmin (commented):
As an idea I like it. The first step to give agility in terms of `AvatarView`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
```diff
@@ -0,0 +1,67 @@
...
+    public val avatarBorderWidth: Int,
+    @ColorInt public val avatarBorderColor: Int,
+    public val avatarInitialText: TextStyle,
+    public val onlineIndicatorEnabled: Boolean,
+    public val onlineIndicatorPosition: AvatarView.OnlineIndicatorPosition,
+) {
+
+    internal companion object {
+
+        operator fun invoke(context: Context, attrs: AttributeSet?): AvatarStyle =
```
zsmb13, line 23: Is this magic really worth it? Could just as well be a top level `AvatarView()` function if we wanted this syntax. Otherwise it could be something like `AvatarStyle.parse()`.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
```diff
@@ -0,0 +1,67 @@
...
+                    .build()
+                val onlineIndicatorEnabled = it.getBoolean(
+                    R.styleable.AvatarView_streamUiAvatarOnlineIndicatorEnabled,
+                    false
+                )
+                val onlineIndicatorPosition = it.getEnum(
+                    R.styleable.AvatarView_streamUiAvatarOnlineIndicatorPosition,
+                    AvatarView.OnlineIndicatorPosition.TOP
+                )
+                AvatarStyle(avatarBorderWidth, avatarBorderColor, avatarInitialText, onlineIndicatorEnabled, onlineIndicatorPosition)
```
zsmb13, line 64: Named parameters here, please

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
```diff
@@ -0,0 +1,283 @@
...
+        private const val GRADIENT_DARKER_COLOR_FACTOR = 1.3f
+        private const val GRADIENT_LIGHTER_COLOR_FACTOR = 0.7f
+
+        /**
+         * Marker object to detect whether methods have been implemented by custom subclasses.
+         */
+        private val NOT_IMPLEMENTED_MARKER = Bitmap.createBitmap(0, 0, Bitmap.Config.ALPHA_8)
+        internal lateinit var instance: AvatarBitmapFactory
+
+        public fun overrideDefaultInstance(avatarBitmapFactory: AvatarBitmapFactory) {
```
zsmb13, line 279: `setInstance` perhaps?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
```diff
@@ -0,0 +1,283 @@
...
+        val customDefaultBitmap = withContext(DispatcherProvider.IO) {
+            createDefaultChannelBitmapBlocking(channel, lastActiveUsers, style, avatarSize)
+        }
+        return when (customDefaultBitmap) {
+            NOT_IMPLEMENTED_MARKER -> createDefaultChannelBitmap(channel, lastActiveUsers, style, avatarSize)
+            else -> customDefaultBitmap
+        }
+    }
+
+    public open fun createChannelBitmapBlocking(
```
zsmb13, line 155: I'd like to have docs for all new public methods here, they're not completely trivial

zsmb13 (commented):
LGTM overall, some notes about style + docs

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
JcMinarro: We were using "AvatarView constructor" across our code with those arguments. `data class` doesn't allow you to have arguments on the constructor that aren't properties of the class, so, to not update all parts where this class is created, I chose this option. Do you see any problem it could have? It shouldn't be accessible outside

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
JcMinarro: It will have conflict with the property itself, should we remove this method and open the `instance` property on that case?

JcMinarro: BTW, we should move it to `ChatUI` when the refactor is completed

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarStyle.kt
zsmb13: Feels a bit like overkill when we could just have an `AvatarStyle()` top level function, but there's no specific problem with it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
zsmb13: `private set` for the property?

zsmb13 (approved):

PR #1580, opened 2021-03-18
[CAS-794] Add possibility to enable/disable commands

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-794

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
@@ -133,7 +133,7 @@ public class MessageInputView : ConstraintLayout {
 
     public fun setCommands(commands: List<Command>) {
         suggestionListController?.commands = commands
-        setCommandsEnabled(commands.isNotEmpty())
+        refreshControlsState()
     }
 
     public fun enableSendButton() {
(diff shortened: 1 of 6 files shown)


leandroBorgesFerreira (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -346,7 +360,8 @@ public class MessageInputView : ConstraintLayout {
             val hasValidContent = hasContent && !messageLengthExceeded
 
             attachmentsButton.isVisible = style.attachButtonEnabled && !commandMode
-            commandsButton.isVisible = style.lightningButtonEnabled && !commandMode && suggestionListViewAttached
+            commandsButton.isVisible =
```
JcMinarro, line 363: If the list of commands returned by backend is null, it should be hidden as well. Right?

adasiewiczr: Nice catch! Thanks

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -133,6 +133,7 @@ public class MessageInputView : ConstraintLayout {
 
     public fun setCommands(commands: List<Command>) {
         suggestionListController?.commands = commands
+        setCommandsEnabled(commands.isNotEmpty())
```
JcMinarro, line 136: This `setCommandsEnabled()` call could override a previous `setCommandsEnable(false)` that customer used, because this method is called from a livedata subscription (We don't know when data will arrive)
What about providing some property from `suggestionListControler` that tell us if "inside commands behavior" can be used and checking it from the `refreshControlsState()` method? :thinking:

adasiewiczr: 🙈 I totally forget about that. Thanks one more time!

JcMinarro (approved):

PR #1584, opened 2021-03-18
[CAS-768] Refactor - Moving code from QueryChannelsControllerImpl to QueryChanne…

leandroBorgesFerreira:
[CAS-786](https://stream-io.atlassian.net/browse/CAS-768)

### Description

Creating a QueryChannelsControllerStateFlow so we can have a query channels controller that only works with StateFlow. 

### Checklist

- ~[ ] Comparison screenshots added for visual changes
- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerStateFlow.kt
```diff
@@ -0,0 +1,389 @@
...
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+private const val MESSAGE_LIMIT = 10
+private const val MEMBER_LIMIT = 30
+private const val INITIAL_CHANNEL_OFFSET = 0
+private const val CHANNEL_LIMIT = 30
+
+internal class QueryChannelsControllerStateFlow(
```
leandroBorgesFerreira, line 39: If I implement the `QueryChannelsController` interface I'll need to add the live data methods =|

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerStateFlow.kt
```diff
@@ -0,0 +1,389 @@
+package io.getstream.chat.android.livedata.controller
```
ogkuzmin, line 1: According to task it should be in package `io.getstream.chat.android.offline`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerStateFlow.kt
ogkuzmin: I don't like to use `StateFlow`, we can keep the previous name and make a name alias for it with `StateFlow` suffix in the old `QueryChannelsController` internally. What do you think?

ogkuzmin (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerStateFlow.kt
leandroBorgesFerreira: I think I'm missing something, we can't have 2 `QueryChannelsControllerImpl`.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/controller/QueryChannelsController.kt
```diff
@@ -0,0 +1,391 @@
...
+    private val client: ChatClient,
+    private val domainImpl: ChatDomainImpl,
+) {
+    var newChannelEventFilter: (Channel, FilterObject) -> Boolean = { _, _ -> true }
+    var recoveryNeeded: Boolean = false
+
+    internal val queryChannelsSpec: QueryChannelsSpec = QueryChannelsSpec(filter, sort)
+
+    private val _channels = MutableStateFlow<Map<String, Channel>>(emptyMap())
+    internal val _loading = MutableStateFlow(false)
```
ogkuzmin, line 53: I think all properties with "_" such prefix should be only private, and usually they're mutable. So if you want to expose loading flow then it's better to have something like:

``` kotlin
internal class QueryChannelsController {
 private val _loading = MutableStateFlow(false)
 val loading: Flow<Boolean> = _loading
}

```

ogkuzmin: Because later we'll provide state flow versions to our customers and it's no good expose to them `_someValue`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -5,92 +5,64 @@ import androidx.lifecycle.asLiveData
...
+    val queryChannelsSpec: QueryChannelsSpec = controllerStateFlow.queryChannelsSpec
 
-    private val logger = ChatLogger.get("ChatDomain QueryChannelsController")
+    override val endOfChannels: LiveData<Boolean> = controllerStateFlow.endOfChannels.asLiveData()
 
-    private val _loading = MutableStateFlow(false)
-    override val loading: LiveData<Boolean> = _loading.asLiveData()
+    // Keep the channel list locally sorted
+    override val channels: LiveData<List<Channel>>
+        get() = controllerStateFlow._sortedChannels.asLiveData()
```
ogkuzmin, line 52: Why `controllerStateFlow` doesn't have `channels: Flow<List<Channel>>`? It should have the same interface but in terms of `Flow<*>`. If it was so you could do:
```kotlin
override val channels: LiveData<List<Channel>> = controllerStateFlow.channels.asLiveData()
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsControllerImpl.kt
```diff
@@ -5,92 +5,64 @@ import androidx.lifecycle.asLiveData
...
-            when {
-                loading -> QueryChannelsController.ChannelsState.Loading
-                channels.isEmpty() -> QueryChannelsController.ChannelsState.OfflineNoResults
-                else -> QueryChannelsController.ChannelsState.Result(channels)
-            }
-        }.stateIn(domainImpl.scope, SharingStarted.Eagerly, QueryChannelsController.ChannelsState.NoQueryActive)
+    override val loadingMore: LiveData<Boolean> = controllerStateFlow.loadingMore.asLiveData()
 
-    override val channelsState = _channelsState.asLiveData()
+    override val channelsState = controllerStateFlow._channelsState.asLiveData()
```
ogkuzmin, line 58: Still here weird name

ogkuzmin (approved):

ogkuzmin (approved):

PR #1597, opened 2021-03-22
[CAS-804] Alert for big files

leandroBorgesFerreira:
[CAS-804](https://stream-io.atlassian.net/browse/CAS-804)
### Description

Adding alert for big files.

### Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -206,6 +206,7 @@ public class MessageInputView : ConstraintLayout {
         binding.dismissInputMode.setOnClickListener { dismissInputMode(inputMode) }
         setMentionsEnabled(style.mentionsEnabled)
         setCommandsEnabled(style.commandsEnabled)
+        binding.messageInputFieldView.setAttachmentMaxFileMb(style.maxFileSize)
```
ogkuzmin, line 209: Why is it configurable? Are these limits extendable by our BE?

ogkuzmin (commented):
It's good changes. It makes UX already better. But I think about better solution. Can we show limitations with a grid view? It let us reduce user steps to zero instead of one (choose file -> show alert). User will be able to see a warning before they choose any file

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
leandroBorgesFerreira: The client can use its own backend, which can have different limits.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -50,6 +51,11 @@ public class MessageInputView : ConstraintLayout {
 
     private var onSendButtonClickListener: OnMessageSendButtonClickListener? = null
     private var typingListener: TypingListener? = null
+
+    public val attachmentBiggerThanMax: StateFlow<Boolean> by lazy {
```
ogkuzmin, line 55: Do not like to expose StateFlow in a view. We can pass just some simple handler like:
``` kotlin
fun interface BigFileSelectionHandler {
   fun handleBigFileSelected(fileSize: Int) 
}
```

It let us not to expose any state and don't use `StateFlow` that could be not so convenient for Java users.

ogkuzmin:
We also can enhance UX of media gallery selection. We can disable for selection media files with size bigger than max size

ogkuzmin (approved):

PR #1610, opened 2021-03-24
Run postponed queries as non-blocking

JcMinarro:
### Description

Run postponed queries as non-blocking
### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
@@ -5,17 +5,20 @@ import io.getstream.chat.android.client.api.ErrorCall
 import io.getstream.chat.android.client.api.models.QueryChannelRequest
 import io.getstream.chat.android.client.api.models.QueryChannelsRequest
 import io.getstream.chat.android.client.call.Call
+import io.getstream.chat.android.client.call.CoroutineCall
 import io.getstream.chat.android.client.clientstate.ClientState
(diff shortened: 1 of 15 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
```diff
@@ -5,17 +5,21 @@ import io.getstream.chat.android.client.api.ErrorCall
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
 import java.util.concurrent.TimeUnit
 
 internal class QueryChannelsPostponeHelper(
     private val api: ChatApi,
     private val clientStateService: ClientStateService,
+    private val coroutineScope: CoroutineScope = GlobalScope,
```
ogkuzmin, line 22: I'm not sure that it's a good way to use GlobalScope. Right now it will solve your task, but later other guys can copy such an approach for other places. What about to create scope with BG dispatcher in `ChatClient`? I think it can be reused for other purposes. We can later rethink how to handle connection lifecycle and clear scope when disconnected, for example

JcMinarro: Ok, for now, I am injecting it. 
We will have a `networkScope`, but, for now, I am not going to cancel any job yet, keeping implementation as it is.
We can decide late if this scope needs to be canceled at some time

ogkuzmin (approved):
LGTM

PR #1611, opened 2021-03-24
[CAS-808] Fix refreshing not messaging channels which don't contain current user as a member

### LINKED ISSUE #1585: [4.7.0] Not able to see custom channels
Our backend has created some custom channels that have ID format like this:
```
global:muc-random-hex-string
```

In v4.6.0, I was able to fetch these custom channels with this filter:
```
Filters.eq("type", "global")
```

But, in v4.7.0, I am not able to fetch these channels with same filter. Instead, empty-channels message appears in `ChannelsView`.

When I upgraded stream dependency to v4.7.0, I followed migration guide from release page. i.e. corrected package for `FilterObject`.

adasiewiczr:
fixes #1585 
We shouldn't assume that we don't want to display channels in which the current user is not a member - the assumption is correct only for messaging channels.
`refreshChannels` method should be refactored after the channel's filtering is done. 
### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/controller/QueryChannelsController.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/controller/QueryChannelsController.kt
@@ -221,7 +221,10 @@ internal class QueryChannelsController(
         val existingChannelMap = _channels.value.toMutableMap()
 
         newChannels.forEach { channel ->
-            if (channel.members.any { member -> member.getUserId() == domainImpl.currentUser.id }) {
+            // TODO: Update after channel's filtering refactoring is done.
+            // In order to interact with messaging channel user needs to be a member so for now,
+            // we are removing only channels of messaging type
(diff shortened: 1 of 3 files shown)


leandroBorgesFerreira (approved):
🚢 🚢 🚢

bychkovdmitry (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/controller/QueryChannelsController.kt
```diff
@@ -221,7 +221,10 @@ internal class QueryChannelsController(
         val existingChannelMap = _channels.value.toMutableMap()
 
         newChannels.forEach { channel ->
-            if (channel.members.any { member -> member.getUserId() == domainImpl.currentUser.id }) {
+            // TODO: Update after channel's filtering refactoring is done.
```
JcMinarro, line 224: Could you create a linked task for it on Jira?

PR #1613, opened 2021-03-25
Avoid memory leaks

bychkovdmitry:
### Description

- Lifecycle observer which is responsible for connecting/disconnecting to/from the WS was never disposed. Disposing it on `ChatClient.disconnect()`.
- We were using the deprecated `ChatDomain.Builder` constructor with user overwrite in our sample app. But we didn't call `ChatDomain.disonnect()` on log out. As a consequence, `ChatDomain` was never disposed. Using the new constructor without user overwrite.

Only one instance of `ChatCllient` and `ChatDomain` exists in memory now. Checked with weak references.

### Checklist

- [X] Reviewers added


ogkuzmin (approved):
Nice catch! 👍

PR #1621, opened 2021-03-26
Coroutine fixes

zsmb13:
### Description

Fixes some coroutine handling issues, see inline comments for details

Supercedes #1618 

### Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
@@ -45,6 +45,7 @@ import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
 import kotlinx.coroutines.CoroutineScope
 import okhttp3.OkHttpClient
 import retrofit2.Retrofit
+import java.util.concurrent.Executor
 import java.util.concurrent.TimeUnit
 
 internal open class BaseChatModule(
@@ -53,6 +54,7 @@ internal open class BaseChatModule(
     private val notificationsHandler: ChatNotificationHandler,
     private val fileUploader: FileUploader? = null,
     private val tokenManager: TokenManager = TokenManagerImpl(),
+    private val callbackExecutor: Executor?,
 ) {
 
     private val defaultLogger: ChatLogger = ChatLogger.Builder(config.loggerConfig).build()
@@ -121,7 +123,7 @@ internal open class BaseChatModule(
             .baseUrl(endpoint)
             .client(okHttpClient)
             .also(parser::configRetrofit)
-            .addCallAdapterFactory(RetrofitCallAdapterFactory.create(parser))
+            .addCallAdapterFactory(RetrofitCallAdapterFactory.create(parser, callbackExecutor))
             .build()
     }
 
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
@@ -41,8 +41,8 @@ import io.getstream.chat.android.client.token.TokenManagerImpl
 import io.getstream.chat.android.client.uploader.FileUploader
 import io.getstream.chat.android.client.uploader.StreamFileUploader
 import io.getstream.chat.android.client.utils.UuidGeneratorImpl
+import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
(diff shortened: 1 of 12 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -71,7 +71,7 @@ internal open class BaseChatModule(
         StreamFileUploader(buildRetrofitCdnApi())
     }
 
-    private val networkScope: CoroutineScope = CoroutineScope(Dispatchers.IO)
+    private val networkScope: CoroutineScope = CoroutineScope(DispatcherProvider.IO)
```
zsmb13, line 74: Use our own mockable dispatchers instead of the default ones

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
```diff
@@ -27,11 +28,11 @@ internal class QueryChannelsPostponeHelper(
         channelId: String,
         request: QueryChannelRequest
     ): Call<Channel> = CoroutineCall(coroutineScope) {
-        doSafeJob { api.queryChannel(channelType, channelId, request) }.execute()
```
zsmb13, line 30: This was performing a blocking operation inside a suspending function, which is never a good idea, see for example https://elizarov.medium.com/blocking-threads-suspending-coroutines-d33e11bf4761#4c8b

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -12,17 +12,23 @@ import kotlinx.coroutines.withContext
...
-    public val suspendingTask: suspend () -> Result<T>,
+    private val suspendingTask: suspend () -> Result<T>,
 ) : Call<T> {
 
     private var job: Job? = null
 
+    internal suspend fun awaitImpl(): Result<T> {
+        return withContext(scope.coroutineContext) {
+            suspendingTask()
+        }
```
zsmb13, line 23: Applies the context from the `scope` before running `suspendingTask`. This way, it'll run on the specified dispatcher (among other things), giving us a chance to put the `suspendingTask` on a background thread even if it's not a well-behaving suspending function, like the ones we used to have in `QueryChannelsPostponeHelper`, marked as `suspend` but actually blocking.

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -12,17 +12,23 @@ import kotlinx.coroutines.withContext
...
+    }
+
     override fun cancel() {
         job?.cancel()
     }
 
     override fun execute(): Result<T> {
-        return runBlocking { suspendingTask() }
+        return runBlocking(scope.coroutineContext) { suspendingTask() }
```
zsmb13, line 31: Same idea here, making sure that if we specified things in the `CoroutineContext` of `scope`, we're applying that to the suspending task when running it

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
```diff
@@ -73,7 +73,7 @@ public interface Call<T : Any> {
  */
 public suspend fun <T : Any> Call<T>.await(): Result<T> {
     if (this is CoroutineCall<T>) {
-        return this.suspendingTask.invoke()
+        return this.awaitImpl()
```
zsmb13, line 76: Pushed this responsibility to the impl class

adasiewiczr (approved):

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -12,17 +12,23 @@ import kotlinx.coroutines.withContext
...
+        }
+    }
+
     override fun cancel() {
         job?.cancel()
     }
 
     override fun execute(): Result<T> {
-        return runBlocking { suspendingTask() }
+        return runBlocking(block = suspendingTask)
```
zsmb13, line 31: Ended up not specifying the coroutine context here, because doing so makes tests fail when we're calling nested suspending Room Dao methods inside a transaction.

adasiewiczr (approved):

JcMinarro (approved):

PR #1630, opened 2021-03-28
[DO NOT MERGE] Push notifications when app is killed

bychkovdmitry:
### Description

Just food for thought on what is required to fix push notification when app process is killed. For the detailed description of what's happening please see the comments below.

### Checklist

- [X] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
@@ -83,8 +83,8 @@ internal class MoshiChatApi(
 
     val logger = ChatLogger.get("MoshiChatApi")
 
-    lateinit var userId: String
-    lateinit var connectionId: String
+    private var userId: String = ""
+    private var connectionId: String = ""
 
     override fun setConnection(userId: String, connectionId: String) {
         this.userId = userId
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -455,7 +455,6 @@ internal class ChatDomainImpl internal constructor(
             object : SocketListener() {
                 override fun onConnected(event: ConnectedEvent) {
                     storeBgSyncData()
-                    client.removeSocketListener(this)
                 }
             }
         )
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
@@ -3,7 +3,6 @@ package io.getstream.chat.android.livedata.service.sync
 import android.app.NotificationChannel
 import android.app.NotificationManager
 import android.app.Service
-import android.content.Context
 import android.os.Build
 import androidx.annotation.WorkerThread
 import androidx.core.app.NotificationCompat
@@ -14,7 +13,6 @@ import io.getstream.chat.android.client.logger.ChatLogger
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.client.notifications.FirebaseMessageParser
 import io.getstream.chat.android.client.notifications.FirebaseMessageParserImpl
-import io.getstream.chat.android.client.notifications.handler.ChatNotificationHandler
 import io.getstream.chat.android.client.notifications.handler.NotificationConfig
 import io.getstream.chat.android.livedata.ChatDomain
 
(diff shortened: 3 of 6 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -83,8 +83,8 @@ internal class MoshiChatApi(
 
     val logger = ChatLogger.get("MoshiChatApi")
 
-    lateinit var userId: String
-    lateinit var connectionId: String
+    private var userId: String = ""
+    private var connectionId: String = ""
```
bychkovdmitry, line 87: Not to crash from notification as we don't connect to the WS and therefore we don't have `connectionId`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -455,7 +455,6 @@ internal class ChatDomainImpl internal constructor(
             object : SocketListener() {
                 override fun onConnected(event: ConnectedEvent) {
                     storeBgSyncData()
```
bychkovdmitry, line 457: Is `ChatClient` is properly initialized in the Application class, then we might want this call to be executed multiple time when user changes account.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -43,18 +41,10 @@ public class PushMessageSyncHandler(private val service: Service) {
...
-        } else {
-            val syncConfig = syncModule.encryptedBackgroundSyncConfigStore.get()
-            syncConfig?.let {
-                val config = it
-                val user = User(id = config.userId)
-                val userToken = config.userToken
-                val client = initClient(service, user, userToken, config.apiKey)
-                client.onNewTokenReceived(token)
-            }
+        syncModule.encryptedBackgroundSyncConfigStore.get()?.let {
```
bychkovdmitry, line 44: `ChatClient` is always initilized if it is done in the Application class. There are two possible case:

-  App in the background, but the process is alive: `ChatClient` already initilizaed.
-  App process is killed. Then the following happens in this particular order:
1. Content providers are innitialized (irrelevant for us)
2. `Application` class is initialized, and therefore `ChatClient` is initilizaed
3. Services are initialized, and in this particular case `onNewToken` is invoked

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -79,22 +69,16 @@ public class PushMessageSyncHandler(private val service: Service) {
         startForegroundExecution(notificationConfig.smallIcon)
 
         val cid: String = firebaseMessageParser.parse(message).let { "${it.channelType}:${it.channelId}" }
```
bychkovdmitry, line 71: The same as the above. We assume that clients might have `ChatClient` not initialized, which is not correct. This branch should never have in a real app.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -79,22 +69,16 @@ public class PushMessageSyncHandler(private val service: Service) {
...
-                val user = User(id = config.userId)
-                val token = config.userToken
-                val client = initClient(service, user, token, config.apiKey)
-                val domain = initDomain(user, client)
+        val syncConfig = syncModule.encryptedBackgroundSyncConfigStore.get()
+        syncConfig?.let {
+            val config = it
+            val user = User(id = config.userId)
+            val token = config.userToken
+            ChatClient.instance().setUserWithoutConnecting(user, token)
```
bychkovdmitry, line 77: This method actually does something a bit different. Is sets the user, but only if  it was not set before. So we can safely call the method here.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
```diff
@@ -144,30 +128,6 @@ public class PushMessageSyncHandler(private val service: Service) {
         }
     }
```
bychkovdmitry, line 130: We can't initilizae `client` or `domain` here, because we are unaware what parameters to pass. For example, in our sample app we pass `SampleNotificationHandler`. And that is one of the reasons why push notifications don't work when app process is killed.

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
```diff
@@ -18,6 +18,7 @@ class App : Application() {
         super.onCreate()
         instance = this
         DebugMetricsHelper.init()
```
bychkovdmitry, line 20: Init `ChatClient` and `ChatDomain` here. All the static data needs to initialized in the `Application` class, otherwise after process death it will be lost. (E.g. bring app to the background, kill app process, locate the app in task manager).

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/home/HomeFragment.kt
```diff
@@ -80,6 +81,8 @@ class HomeFragment : Fragment() {
 
                 requireActivity().intent = null
```
bychkovdmitry, line 83: This is a hack. We did everything in push notification handler except connecting to the socket.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/PushMessageSyncHandler.kt
JcMinarro: Not all users are initializing `ChatClient` on `Application`

PR #1632, opened 2021-03-29
[CAS-829] Fix - Cursor crash for API bellow 29

leandroBorgesFerreira:
[CAS_829](https://stream-io.atlassian.net/browse/CAS-829)

### Description

Now it is only possible to change programatically the cursor for `MessageInputFieldView` for API 29+. In order to be able to change the cursor  `MessageInputFieldView`, users now need to override `stream_ui_message_input_cursor.xml`.

That's a little down size: The cursor won't respect the accent color by default anymore, which can be unexpected for some users. I would liek some options about this. Thanks!

### Checklist

- [x] Changelog updated with client-facing changes. -> Todo
- [x] Reviewers added


ogkuzmin (approved):
For me it's ok since other possible way is reflection usage that isn't good because some vendors can override the framework behaviour and reflection doesn't work.

PR #1670, opened 2021-04-02
[CAS-846] Create a fallback SharedPreferences in case the Encrypted one is not available

JcMinarro:
### 🎯 Goal
Some emulators and some Android devices are not able to created an Encrypted Shared Preferences throwing an error that is not handled. On case we are not able to generated an Encrypted Shared Preferences we fallback to a normal Shared Preferences instance

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/EncryptedBackgroundSyncConfigStore.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/EncryptedBackgroundSyncConfigStore.kt
@@ -4,27 +4,38 @@ import android.content.Context
 import android.content.SharedPreferences
 import androidx.security.crypto.EncryptedSharedPreferences
 import androidx.security.crypto.MasterKey
+import io.getstream.chat.android.client.logger.ChatLogger
 
 private const val MASTER_KEY_ALIAS = "_stream_sync_config_master_key_"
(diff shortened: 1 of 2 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/service/sync/EncryptedBackgroundSyncConfigStore.kt
```diff
@@ -4,27 +4,38 @@ import android.content.Context
...
+                masterKey,
+                EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+                EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
+            )
+        } catch (e: Exception) {
+            logger.logE("Error creating encrypted shared preferences", e)
+            context.applicationContext.getSharedPreferences(
+                SYNC_CONFIG_PREFS_NAME,
+                Context.MODE_PRIVATE
+            )
```
zsmb13, line 37: Is this safe enough for our needs? We must be using encrypted SharedPref normally for a reason 😄

JcMinarro: It stores the userId and the token.
It is into their own app sandbox that is not accessible by other apps (At least they have root permissions)

samiuelson (approved):

PR #1681, opened 2021-04-05
[CAS-842] Fix read indicator

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-842

### 🎯 Goal

This PR fixes the message read indicator in the message list.

### 🛠 Implementation details

Incoming reads from other users were ignored. Ensured that all the incoming reads (including the reads from other users) are propagated via corresponding `StateFlow` field.

### 🧪 Testing

GIVEN user_1 is reading a channel
AND user_2 is reading the same channel
WHEN user_1 sends a message
AND user_2 receives the message
THEN user_1 should observe that the message was read (double tick mark)

### ☑️ Checklist

- [X] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
@@ -1276,24 +1276,25 @@ internal class ChannelController(
          * before what we've last pushed to the UI. We want to ignore this, as it will cause an unread state
          * to show in the channel list.
          */
-        val incomingUserRead = incomingUserIdToReadMap[currentUserId] ?: return
+        incomingUserIdToReadMap[currentUserId]?.let {
 
-        // the previous last Read date that is most current
-        val previousLastRead = _read.value?.lastRead ?: previousUserIdToReadMap[currentUserId]?.lastRead
+            // the previous last Read date that is most current
+            val previousLastRead = _read.value?.lastRead ?: previousUserIdToReadMap[currentUserId]?.lastRead
 
-        // Use AFTER to determine if the incoming read is more current.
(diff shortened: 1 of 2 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -1276,24 +1276,25 @@ internal class ChannelController(
          * before what we've last pushed to the UI. We want to ignore this, as it will cause an unread state
          * to show in the channel list.
          */
```
bychkovdmitry, line 1278: Even if there is no incoming user read for current user we should still update the `_reads.value` below.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -1276,24 +1276,25 @@ internal class ChannelController(
...
 
-        // Use AFTER to determine if the incoming read is more current.
-        // This prevents updates if it's BEFORE or EQUAL TO the previous Read.
-        val shouldUpdateByIncoming = previousLastRead == null || incomingUserRead.lastRead?.inOffsetWith(
-            previousLastRead,
-            OFFSET_EVENT_TIME
-        ) == true
+            // Use AFTER to determine if the incoming read is more current.
+            // This prevents updates if it's BEFORE or EQUAL TO the previous Read.
+            val shouldUpdateByIncoming = previousLastRead == null || it.lastRead?.inOffsetWith(
```
adasiewiczr, line 1286: Using named parameter instead of `it` will increase code readability

adasiewiczr (approved):

Issue #1679, opened 2021-04-05
[MessageInputView] Attachment files list shows folders and crashes when selected & sent


rupinderjeet:
Here's the screenshot:
![Screenshot 2021-04-05 at 7 00 45 PM](https://user-images.githubusercontent.com/14011726/113578976-446d8180-9641-11eb-9333-6d178935458a.png)

Here's the crash [log](https://github.com/GetStream/stream-chat-android/files/6258346/crash.txt)

adasiewiczr:
Hey, thanks for reporting the issue, we'll get this fixed soon.

adasiewiczr:
Hey @rupinderjeet, the issue was fixed in the PR mentioned above and will be shipped in the next release.

JcMinarro:
Hi @rupinderjeet 
It has been fixed and we already released a [new versión](https://github.com/GetStream/stream-chat-android/releases/latest)
I'm closing this issue

PR #1693, opened 2021-04-06
Exclude folders and files with unknown mime type when getting files

adasiewiczr:
#1679 

### 🎯 Goal
Fix list of `Recent Files`  

### 🛠 Implementation details

In order to display `Recent Files` we are using a content resolver that queries for every file without specifying the type.
The content resolver doesn't provide an easy way to distinguish whether returned uri is a file or folder. 
The easiest solution is to check mime type (folder doesn't have it) and filter those files which has empty mime type. The downside of this solution is that we won't display other files in which mime type is not specified, but they can still be accessed using `+` button placed in the right corner.

### 🧪 Testing

1. Open `MessageInputView`
2. Open `File Attachment Menu`
3. Recent Files shouldn't contain folders and unknown files

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/StorageHelper.kt
```diff
@@ -47,7 +47,8 @@ public class StorageHelper {
     }
 
     public fun getFileAttachments(context: Context): List<AttachmentMetaData> {
-        return getFilteredAttachments(context, null)
+        // Excluding files with empty mime type just to be sure that we won't include folder and unknown files
+        return getFilteredAttachments(context, null).filterNot { it.mimeType.isNullOrEmpty() }
```
ogkuzmin, line 51: Can't we exclude such file using projection (where in SQL)? Something like `MediaStore.FileColumns.MIME_TYPE != ""`

adasiewiczr: Good idea, let me check that

ogkuzmin (approved):

PR #1698, opened 2021-04-06
CAS-809 Eliminate `useCases` prop from `ChatDomain`

samiuelson:
https://stream-io.atlassian.net/browse/CAS-809

### 🎯 Goal

Eliminate `ChatDomain.useCases` prop from public API. Expose usecase with `ChatDomain` functions.

### 🛠 Implementation details

* Expose use cases directly by `ChatDomain` functions
* Deprecate `ChatDomain.useCases`
* Replace `ChatDomain.useCases` calls by use case functions
* Replace deprecated `totalUnreadCount` occurrences by new implementation

### 🧪 Testing

Everything should work as before without any change 

### ☑️ Checklist

- [x] Affected documentation updated (CMS, cookbooks, tutorial)

- [x] Reviewers added

### 🎉 GIF

dog


UNRELEASED_CHANGELOG.md
```diff
@@ -44,7 +44,42 @@
 
 ### ✅ Added
 - Add `ChatDomain::removeMembers` method
+- Added the following use case functions to `ChatDomain` which are supposed to replace `ChatDomain.useCases` property:
+* `ChatDomain::replayEventsForActiveChannelsCall`
```
zsmb13, line 48: Is there a reason for postfixing all of these with `Call`? These methods should be similar to the ones on `ChatClient`.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -851,6 +888,115 @@ internal class ChatDomainImpl internal constructor(
         return repos.selectChannelConfig(channelType)?.config ?: defaultConfig
     }
 
+    // region use-case functions
+    override fun replayEventsForActiveChannelsCall(cid: String): Call<List<ChatEvent>> = ReplayEventsForActiveChannelsImpl(this).invoke(cid)
```
zsmb13, line 892: We could still use the `UseCaseHelper` internally so that we have just one instance of each use case, code might be simpler as well

UNRELEASED_CHANGELOG.md
samiuelson: The main reasons behind the naming:
1. We would have duplicate function names in `ChatDomainImpl` class. There are already functions like `ChatDomainImp::createChannel`, or `::replayEventsForActiveChannels`. They have different return types but the same names as some of the use cases so we can't keep both of them.
2. Those existing functions (`::createChannel`, `::replayEventsForActiveChannels`) names are imperative suggesting (for me) they will perform some action while use case functions are returning a `Call` instances that needs to be executed later.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
samiuelson: Sounds good, will do that.

UNRELEASED_CHANGELOG.md
zsmb13: All the existing functions are just internal API in the impl class, right? In that case we can find new names for them so that the public API is cleaner.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -380,7 +387,7 @@ internal class ChatDomainImpl internal constructor(
         return result
     }
 
-    suspend fun createChannel(c: Channel): Result<Channel> =
+    suspend fun createNewChannel(c: Channel): Result<Channel> =
```
samiuelson, line 390: This is an internal function. I needed to change its name because there would be a conflict with the use case function name.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -595,7 +602,7 @@ internal class ChatDomainImpl internal constructor(
      *
      * @param cid ensures that the channel with this id is active
      */
-    suspend fun replayEventsForActiveChannels(cid: String? = null): Result<List<ChatEvent>> {
+    suspend fun replayEvents(cid: String? = null): Result<List<ChatEvent>> {
```
samiuelson, line 605: The same as above.

UNRELEASED_CHANGELOG.md
samiuelson: Done.

UNRELEASED_CHANGELOG.md
```diff
@@ -39,8 +39,43 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Added the following use case functions to `ChatDomain` which are supposed to replace `ChatDomain.useCases` property:
+* `ChatDomain::replayEventsForActiveChannels`
```
JcMinarro, line 43: Could you add an small description of what every method do?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
JcMinarro: It is only used internally in offline SDK, right?
Can you modify the visibility of this method?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
JcMinarro: Same here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
samiuelson: Yep, added `internal` modifier

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
samiuelson: done

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -141,6 +147,7 @@ internal class ChatDomainImpl internal constructor(
     private val _banned = MutableStateFlow(false)
     private val _mutedUsers = MutableStateFlow<List<Mute>>(emptyList())
     private val _typingChannels = MediatorLiveData<TypingEvent>()
+    private val useCaseProvider: UseCaseHelper = UseCaseHelper(this)
```
ogkuzmin, line 150: Why did you decide to create new instance? Why can't we use just deprecated `useCaseHelper`?

ogkuzmin (commented):
Good job! API becomes more concise. Just one question

UNRELEASED_CHANGELOG.md
samiuelson: Added

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
samiuelson: Because it will be deprecated with `DeprecationLevel.Error` soon

ogkuzmin (approved):
LGTM

PR #1703, opened 2021-04-07
[CAS-867] Add missing team field to ChannelEntity and ChannelData

adasiewiczr:
### 🎯 Goal

Fix filtering channel based on `Channel::team` property.

### 🛠 Implementation details

Filtering wasn't working properly because `Channel::team` was always empty. In [`QueryChannelsController::refreshChannels` ](https://github.com/GetStream/stream-chat-android/blob/main/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/controller/QueryChannelsController.kt#L221) we are creating `Channel` from `ChannelController` which internally uses `ChannelData`. `Team` property was missing in both `ChannelData` and `ChannelEntity` and that's why filtering wasn't working properly (the property was always empty (default value) even if it was present in API response)

### 🧪 Testing

1. Add `Filters.eq("team", "someTeam")` filter to `ChannelListViewModelFactory` and ensure that API returns at least one channel
2. ChannelListView should contain this channel
### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

### DIFF
--- a/UNRELEASED_CHANGELOG.md
+++ b/UNRELEASED_CHANGELOG.md
@@ -39,6 +39,7 @@
 ### 🐞 Fixed
 - Fix Crash on some devices that are not able to create an Encrypted SharedPreferences
 - Fixed the message read indicator in the message list
+- Added missing `team` field to `ChannelEntity` and `ChannelData`
 ### ⬆️ Improved
 
 ### ✅ Added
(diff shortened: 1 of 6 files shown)


ogkuzmin (approved):
Nice catch!
I dream about the time when we have this only in DB and don't need this `ChannelData`

PR #1707, opened 2021-04-07
Migration to new Nexus plugin

zsmb13:
https://stream-io.atlassian.net/browse/CAS-859

### 🎯 Goal

We currently use this plugin which is going into maintenance mode https://github.com/Codearte/gradle-nexus-staging-plugin

Therefore we're migrating to this new plugin instead, as recommended on the old plugin's page https://github.com/gradle-nexus/publish-plugin/

### 🛠 Implementation details

See inline comments.

### 🧪 Testing

The new setup works for publishing to the local repository, still has to be tested for real publishing to MavenCentral.

### ☑️ Checklist

- [x] Reviewers added


scripts/publish-mavencentral.gradle
```diff
@@ -146,29 +146,18 @@ publishing {
             }
         }
     }
-    repositories {
-        maven {
-            name = "sonatype"
-
-            def releasesRepoUrl = "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
-            def snapshotsRepoUrl = "https://oss.sonatype.org/content/repositories/snapshots/"
-            url = version.endsWith('SNAPSHOT') ? snapshotsRepoUrl : releasesRepoUrl
```
JcMinarro, line 155: Is it already configured into the new plugin? Or because we are not supporting Snapshot for now has been removed?

zsmb13: I believe the plugin handles this based on the version number, yes. But we're not using snapshots anyway, so it's also fine if it just doesn't handle it 😄

scripts/publish-module.gradle
```diff
@@ -132,29 +108,12 @@ afterEvaluate {
...
 
-nexusStaging {
-    packageGroup = PUBLISH_GROUP_ID
-    stagingProfileId = sonatypeStagingProfileId
-    username = ossrhUsername
-    password = ossrhPassword
-}
+ext["signing.keyId"] = rootProject.ext["signing.keyId"]
+ext["signing.password"] = rootProject.ext["signing.password"]
+ext["signing.secretKeyRingFile"] = rootProject.ext["signing.secretKeyRingFile"]
```
zsmb13, line 116: All props are now fetched into the root projects, these ones need to be copied to the local project for the signing plugin to find them

scripts/publish-module.gradle
```diff
@@ -29,33 +29,9 @@ artifacts {
...
-        ext[name] = value
-    }
-} else {
-    ext["signing.keyId"] = System.getenv('SIGNING_KEY_ID')
-    ext["signing.password"] = System.getenv('SIGNING_PASSWORD')
-    ext["signing.secretKeyRingFile"] = System.getenv('SIGNING_SECRET_KEY_RING_FILE')
-    ext["ossrhUsername"] = System.getenv('OSSRH_USERNAME')
-    ext["ossrhPassword"] = System.getenv('OSSRH_PASSWORD')
-    ext["sonatypeStagingProfileId"] = System.getenv('SONATYPE_STAGING_PROFILE_ID')
-}
```
zsmb13, line 57: All of this moved into the root project, as the Nexus publishing plugin now needs to be initialized there and it relies on some of these values

JcMinarro (approved):

Issue #1688, opened 2021-04-06
[MessageListView] Change UI to show result of mute and block user options in message-actions


rupinderjeet:
When I mute a user, message-actions UI doesn't change its UI to show that user was muted i.e. either don't show mute-option again or show option to "unmute" user. Same for block user.

_source: https://github.com/GetStream/stream-chat-android/issues/1656_

zsmb13:
Thanks for asking about it, we'll be working on this feature in the coming weeks.

rupinderjeet:
Hello @zsmb13 

I understand that this isn't in the plan yet. But, do you think #1691 is possible?

I have tried to handle some issues with  `findViewById`. But, this is a dialog, I don't think I can access its view and hide the options with some stealth.

zsmb13:
Yes, we'll also be picking up that task in the next two weeks.

adasiewiczr:
Hey @rupinderjeet thanks for reporting that. It was fixed in the PR mentioned above and will be shipped in the next release

samiuelson:
Please check the new SDK release (version 4.9.0) which contains the fix. Let us know if you have any further issues.

PR #1733, opened 2021-04-14
[CAS-863] Make some message options stateful

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-863

### 🎯 Goal

#1688 
Improving message options:
- Add possibility to unmute muted user
- Add confirmation dialog to flag message option
- Add possibility to handle flag message result

### 🛠 Implementation details
PR adds multiple handlers that allow handling different actions inside `MessageListView`.
PR also includes some fixes for Moshi parsing.

### 🎨 UI Changes
Screenshot 2021-04-14 at 16 43 26
Screenshot 2021-04-14 at 16 43 39
Screenshot 2021-04-14 at 16 43 53
Screenshot 2021-04-14 at 16 44 12

### 🧪 Testing

- Open message options menu -> Flag message
- Open message options menu -> Mute User -> Unmute User

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
giphy


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/MuteMapping.kt
```diff
@@ -7,15 +7,15 @@ import io.getstream.chat.android.client.models.Mute
 internal fun Mute.toDto(): UpstreamMuteDto =
     UpstreamMuteDto(
         user = user.toDto(),
-        target = user.toDto(),
+        target = target.toDto(),
```
adasiewiczr, line 10: 🙈 it wasn't easy to find that

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Flag.kt
```diff
@@ -20,7 +20,7 @@ public data class Flag(
     @SerializedName("reviewed_at")
     val reviewedAt: Date,
     @SerializedName("reviewed_by")
-    val approvedAt: Date,
+    val approvedAt: Date?,
```
adasiewiczr, line 23: It turned out that `approvedAt` should be nullable. I have no idea how does it work with Gson because the property is always null although it's non-nullable

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
```diff
@@ -211,4 +229,7 @@ class ChatFragment : Fragment() {
         }
         return previousYear != year || previousDayOfYear != dayOfYear
     }
+
+    private fun Result<Flag>.isAlreadyExistsError(): Boolean =
+        isError && error().message?.contains("flag already exists") == true
```
adasiewiczr, line 234: There is no easy way to hide this option from menu so we allow to flag the message multiple times

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -182,7 +189,9 @@ public class MessageListViewModel @JvmOverloads constructor(
                 domain.deleteMessage(event.message).enqueue()
             }
             is Event.FlagMessage -> {
-                client.flagMessage(event.message.id).enqueue()
+                client.flagMessage(event.message.id).enqueue { result ->
```
ogkuzmin, line 192: IMO, creating live data aka event bus for async operation here is too much. Event bus can work fine for those events that we don't expect to receive. Here user is source of asyn requests, so if I did implementation I would add to `FlagMessage` new argument `val resultHandler: (Result<Flag>) -> Unit` and wrote this part as:
``` kotlin
 is Event.FlagMessage -> client.flagMessage(event.message.id).enqueue(event::resultHandler)
```

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
ogkuzmin: Maybe ask BE team? Because they can change message text easily in the future and our implementation will get broken

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
adasiewiczr: You're right, thanks! I was looking for simplest solution and I totally missed that

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatFragment.kt
adasiewiczr: I think we can rely on status codes instead. I'll double-check values with backend team tomorrow

ogkuzmin (approved):

PR #1737, opened 2021-04-14
[CAS-837] Fix - Adding Mute Icon

leandroBorgesFerreira:
[CAS-837](https://stream-io.atlassian.net/browse/CAS-837)

### 🎯 Goal

Add a icon to show that an channel is muted. 

### 🛠 Implementation details

Adding an icon if the channel is muted in that `User`. The muted channels are skep locally and can be requested using `ChatDomain`. 

The icon is placed next to the new messages badge, which is very close to what Whatsapp has. 

### 🎨 UI Changes

_ Add relevant screenshots_

| Before | After |
| --- | --- |
| Screenshot_20210414-153736 | Screenshot_20210414-154105 |

### 🧪 Testing

Mute a channel and check that the icon appears

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -59,10 +60,23 @@ public class ChannelListViewModel(
...
                                 channels = emptyList(),
                             )
-                            is QueryChannelsController.ChannelsState.Result -> currentState.copy(
-                                isLoading = false,
-                                channels = channelState.channels.filterNot { it.hidden == true || it.isDraft },
-                            )
+                            is QueryChannelsController.ChannelsState.Result -> {
+                                val channelMutesIds = chatDomain.currentUser.channelMutes.map { it.channel.id }
+
+                                val channels = channelState
```
ogkuzmin, line 66: What's about moving this logic in `QueryChannelsController`? It's more flexible to have such flag in domain logic instead only in one small UI part. The same code could be there when we update channels state flow

ogkuzmin (commented):
Also good to leave some description in our changelog

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListViewStyle.kt
```diff
@@ -50,6 +50,8 @@ public data class ChannelListViewStyle(
     @ColorInt public val foregroundLayoutColor: Int,
     public val unreadMessageCounterText: TextStyle,
     @ColorInt public val unreadMessageCounterBackgroundColor: Int,
+    public val mutedChannelIcon: Drawable,
+    @ColorInt public val mutedChannelIconTint: Int
```
adasiewiczr, line 54: Would be nice to update KDoc

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
```diff
@@ -67,10 +70,39 @@ internal class QueryChannelsController(
             when {
                 loading -> QueryChannelsController.ChannelsState.Loading
                 channels.isEmpty() -> QueryChannelsController.ChannelsState.OfflineNoResults
-                else -> QueryChannelsController.ChannelsState.Result(channels)
+                else -> {
+                    val channelMutes = domainImpl.currentUser.channelMutes
```
ogkuzmin, line 74: You shouldn't do any changes here. `channelState` is based on the `_sortedChannels` which one is based on the `_channels` property. We should make changes only in value of this `_channels`

ogkuzmin, line 74: You shouldn't do any changes here. `channelState` is based on the `_sortedChannels` which one is based on the `_channels` property. We should make changes only in value of this `_channels`

ogkuzmin:
Seems that we also have to calculate muted channels when first run query channels request. Info about muted channels can be obtained from `User::channelMutes`

ogkuzmin (approved):

JcMinarro (approved):

PR #1758, opened 2021-04-21
[CAS-912] Custom MimeType icons

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-912

### 🎯 Goal

Make file icons customizable

### 🛠 Implementation details

Unfortunately, making the default implementation (`MimeTypeIconProviderImpl`) extensible is problematic, as internally it not only checks MIME type to icon mapping but also has fallback icons for "audio" and "video" media types. We can add possibility to add custom icons for MIME type templates like "application/*" but not sure if it should override existing MIME types like "application/pdf".

### 🧪 Testing

```
ChatUI.mimeTypeIconProvider = MimeTypeIconProvider {
    R.drawable.stream_ui_ic_file
}
```

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (1)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/MimeTypeIconProvider.kt
```diff
@@ -1,9 +1,26 @@
...
+    /**
+     * Return icon drawable resource for the given MIME type.
+     *
+     * @param mimeType mimeType A MIME type (i.e. application/pdf)
+     * @return The drawable resource for the given MIME type
+     */
+    public fun getIconRes(mimeType: String?): Int
+}
+
+public class MimeTypeIconProviderImpl : MimeTypeIconProvider {
```
zsmb13, line 23: Is this impl class meant to be public?

JcMinarro: It should be to help our customers to only override some icons, but keeps the rest of them with our implementation

JcMinarro (approved):

Issue #1750, opened 2021-04-19
Rate Limiting for Create Device API call


rupinderjeet:
We are being hit with rate-limiting for Create Device API (_aka addDevice_). I have found two places this API could be called from:

- `chatClient.addDevice()` is one way to add devices. But, I am not using it since ChatClient is handling notifications for me.
- `ChatNotifications.onSetUser()` is another way for this API to be called internally. ChatClient calls this API whenever a `connected` event is received.

The [documentation for Create Device API](https://getstream.io/chat/docs/rest/#devices-createdevice) mentions following:
> Adds a new device to a user, if the same device already exists the call will have no effect

I think every call is being counted towards rate-limiting even if it had no effect, which is valid. But, at the same time, it is unfair to me because I have no option to skip this API call.

I suggest that whenever we call this API, we cache last *successfully* sent device(token?) in persistent storage. This way, next time we are about to hit this API, we can just compare current device(token?) with last cached device, and avoid the API call.

rupinderjeet:
I initialize chat-setup in `Application` class as instructed in [UI-Cookbook](https://github.com/GetStream/stream-chat-android/wiki/UI-Cookbook#initialize-sdk). The above mentioned API will hit every-time one of the following happens:

- app crashes & reopens
- chat socket reconnects to Stream
- user internet connection changes
- a user regularly closing & opening app

These scenarios are very common. Almost every app goes through them. For e.g. in the past, due to an app crash in chat-notifications, large number of apps/devices will reconnect to stream-chat at the same time, which may cause overflow for rate-limiting.

Stream helped us by elevating this limit to 900 calls per minute. But, we are still experiencing this issue. Please guide me if I can do something on my end.

samiuelson:
Thanks for reporting this. We are looking into the solutions and we will fix the initialization in one of the upcoming releases.

tschellenbach:
Things we can do to optimize this:

- we only need to register once, not every time you connect
- customers that don't use push don't need this API call at all, so there should be a way to disable it

JcMinarro:
Hi @rupinderjeet 
A new version has been released that fixes it

PR #1764, opened 2021-04-21
[CAS-914] Optimize number of Create Device API calls

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/1750
https://stream-io.atlassian.net/browse/CAS-914

### 🎯 Goal

Optimize the number of "Create Device" API calls

### 🛠 Implementation details

Persist user id and firebase token which have already been sent to the server

### 🧪 Testing

The following should not trigger "Create Device" API call
- reopening the app
- reconnecting to the socket
- bringing app to the foreground

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (2)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/PushTokenRepository.kt
```diff
@@ -0,0 +1,69 @@
+package io.getstream.chat.android.client.notifications
+
+import android.content.Context
+import android.content.SharedPreferences
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.client.extensions.getNonNullString
+import io.getstream.chat.android.client.logger.ChatLogger
+import io.getstream.chat.android.client.notifications.handler.ChatNotificationHandler
+
+internal class PushTokenRepository(
```
ogkuzmin, line 10: Not the best name since it's not only storing entity. Because it's shared preference storing mechanism is its' internal logic, API interface of this is only `updateTokenIfNecessary`. What about something like `PushTokenUpdateHandler`?

ogkuzmin (approved):
LGTM

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/PushTokenRepository.kt
samiuelson: Or maybe move `updateTokenIfNecessary()` into `ChatNotifications`?

bychkovdmitry: Renamed to `PushTokenUpdateHandler `

PR #1767, opened 2021-04-23
[CAS-949] Fix binding _MessageListViewModel_ before user is set

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-949

### 🎯 Goal

Fix crash related to accessing `ChatDomain::currentUser` in `MessageListViewModel` before user is connected

### 🛠 Implementation details
Workaround and first step to allow initializing VM before user is connected. I've added a separate task to investigate how to improve that.
NOTE: View model won't be refreshed when `watchChannel` fails because `ChatClient` was in an inappropriate state.
`MessageListViewModel` can be created before a user is connected and this scenario was causing a crash because we were accessing `ChatDomain::currentUser` and `ChatDomain::repos` during VM initialization.
Direct access to `ChatDomain::currentUser` was replaced with lambdas and I've added a few workaround if's to check if `repos` was already initialized.
I've also added refreshing `channel` live data when `channelController.channelData` changes.

### 🧪 Testing

1. Create `MessageListViewModel` and bind it with `MessageListView` before connecting the user

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
gif


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
```diff
@@ -530,7 +530,11 @@ public class MessageListView : ConstraintLayout {
     }
 
     public fun init(channel: Channel, currentUser: User) {
-        this.currentUser = currentUser
+        init(channel) { currentUser }
+    }
+
+    public fun init(channel: Channel, getCurrentUser: () -> User) {
```
JcMinarro, line 536: Do we need it as public method?

adasiewiczr: you're right, it can be internal

JcMinarro (approved):

PR #1768, opened 2021-04-23
[CAS-959] Feature/repository facade null object pattern

JcMinarro:
### 🎯 Goal
Create a NoOp RepositoryFacade that is used whenever an user is not set into our `ChatDomain`.
The same RepositoryFacade is used when an user is disconnected to avoid wrong data from another user

### 🛠 Implementation details
The NoOp RepositoryFacade use an InMemory implementation with empty data, with this implementation we don't need null-checking before to use it anymore.
This PR remove the Workaround added on the PR #1767 

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
source (1)

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
@@ -304,11 +304,11 @@ internal class ChatDomainImpl internal constructor(
             client.preSetUserListeners.add {
                 setUser(it)
             }
-            // disconnect if the low level client disconnects
-            client.disconnectListeners.add {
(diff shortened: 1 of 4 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -345,6 +346,7 @@ internal class ChatDomainImpl internal constructor(
         job.cancelChildren()
         stopListening()
         stopClean()
+        clearState()
```
adasiewiczr, line 349: Shouldn't we also replace current database with noOp?

adasiewiczr (approved):

PR #1817, opened 2021-04-29
Provide an unique socket listener whenever a new socket is created

JcMinarro:
### 🎯 Goal
On the SDK we keep only one WS connection opened with the server, but the previous socket keeps the old listener and reports any error that would happen inside it even if the connection was closed by the client.
The SDK client closes the WS connection whenever the main process is moved to the background and opens it again when the main process comeback to the foreground. This process was triggered multiple times because some exception was coming from old listeners that are reporting socket error that shouldn't be covered anymore.
Those errors shouldn't start a new reconnection 

### 🛠 Implementation details
Whenever a new WS Connection is created, a new listener is associated with it.
The listener reject any new event it receives after the client closes the connection

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

gif


samiuelson (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/EventsParser.kt
```diff
@@ -11,20 +11,18 @@ import okhttp3.Response
...
 ) : okhttp3.WebSocketListener() {
 
     private var connectionEventReceived = false
     private val logger = ChatLogger.get("Events")
-    private lateinit var service: ChatSocketService
-
-    fun setSocketService(service: ChatSocketService) {
-        this.service = service
-    }
+    private var closedByClient = true
```
ogkuzmin, line 20: Nice a quick solution, but we can iterate over this logic. Personally, I don't like having states using boolean flags, there are already two of such. We can use state patter or state machine inside it later

ogkuzmin (approved):
Nice fix. Let's iterate over this codebase in the future

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/EventsParser.kt
```diff
@@ -69,21 +71,37 @@ internal class EventsParser(
...
         }
     }
 
     private fun handleErrorEvent(error: ErrorResponse) {
-        service.onSocketError(ChatNetworkError.create(error.code, error.message, error.statusCode))
+        onSocketError(ChatNetworkError.create(error.code, error.message, error.statusCode))
+    }
+
+    private fun onSocketError(error: ChatError) {
+        if (!closedByClient) {
```
zsmb13, line 91: Can't we bring these checks earlier, to the `handleEvent` or even the `onMessage` functions? This way we're doing a bunch of parsing before we discard the result. If that's too early in some cases, maybe we can have checks in both places?

PR #1838, opened 2021-05-05
Fix/custom object filtering

JcMinarro:
### 🎯 Goal
Our local filtering process over CustomObject was not working fine when the filter was initialized with an Int value. 
The problem is that parsed value of extraData from our API Response always transforms "number" into "double" values. To work properly on that part we need to cast its value properly by using the functions that provide the `Number` class

### ☑️ Checklist

- [x] Reviewers added


adasiewiczr (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/utils/CustomObjectFiltering.kt
```diff
@@ -91,3 +91,18 @@ private fun <T : Comparable<T>> compare(a: T?, b: T?, compareFun: (Int) -> Boole
 
 private fun CustomObject.getMembersId(): List<String> =
     getMemberPropertyOrExtra(MEMBERS_FIELD_NAME, List::class)?.mapNotNull { (it as? Member)?.getUserId() } ?: emptyList()
+
+@Suppress("UNCHECKED_CAST")
+private fun <T : Any> CustomObject.getExtra(name: String, clazz: KClass<out T>): T? =
+    extraData[name]?.let {
+        when (clazz) {
```
bychkovdmitry, line 98: Was not sure about comparisons of numbers with floating points as such operations are not precise. But seems to work fine.

bychkovdmitry (approved):
LGTM

Issue #1773, opened 2021-04-23
Sometimes can't see titles of channel in 4.6.0 - 4.8.0. Missing messages in newer versions after 4.6.0


nin-echo:
**Describe the bug**
1. Titles of channels are randomly disappearing. 
2. Channels and messages are missing in the newer version after 4.6.0, with the same filters. 

**SDK version**
- 4.6.0

**Screenshots**
1. Titles are gone
    
![image](https://user-images.githubusercontent.com/12556474/115911640-55d5cc80-a43c-11eb-8234-bff1755b20be.png)

2. Messages are missing

![IMG_2192](https://user-images.githubusercontent.com/12556474/115912014-d1377e00-a43c-11eb-845a-4f591ad7d28d.JPG)

-   Use the same account and same filters:
    **<= 4.6.0**: shows correct numbers of channel
    ![image](https://user-images.githubusercontent.com/12556474/115942398-1036f500-a478-11eb-81c9-13c6b2514c1d.jpeg)

    **4.7.0 and later**: missing some channels
    ![image](https://user-images.githubusercontent.com/12556474/115942415-32c90e00-a478-11eb-85c6-405028e5905d.jpeg)

tschellenbach:
Note that permissions can also influence the channel list. IE it only returns channels that you have permissions to view. 

Can you share the filter that you're using for query channels?

The title issue is surprising, that almost seems like some sort of serialization issue with the extra data on the channel.

nin-echo:
@tschellenbach 
Hello, the title issue is happening randomly on different phone models. Usually, the titles will come back after a few hours or days.
I doubt it's a permission issue, we're using the same filter and server in <=4.6.0 and >4.6.0 versions, same as what the official doc shows. 
```
val filter = Filters.and(
            Filters.eq("type", "messaging"),
            Filters.`in`("members", listOf(user.id))
        )
```
after we upgrade to 4.7.0, the missing channel becomes a constant bug. But it will go away when we reverted back to 4.6.0 with the same code. We also tried 4.9.0 & 4.8.0, this issue still exists.

Thank you.

JcMinarro:
Hello @nin-echo 

I can see you are using the old UI-Components library, Did you have the choice to check our [new UI-Components SDK](https://github.com/GetStream/stream-chat-android/tree/main/stream-chat-android-ui-components)?

Re your issue, could you share with us the following info:
-. Are you using any custom view to render channels or are you using the default ones? 
[...]

PR #1855, opened 2021-05-12
Fixing filter for draft and hidden channels

leandroBorgesFerreira:
### 🎯 Goal

Fix the filter for hidden and draft channels. Until so far, it was not possible to make those channels appear. Now if the users don't use our default filter and don't ask for hidden and draft channels to be removed, they will appear. 

This will fix the filter problem of issue #1773.

**Important: This may cause some apps to include many channels that they may not want to, although this fixes a wrong behaviour.**

### 🛠 Implementation details

Removing the filter of those channels from `queryChannelsController.channelsState` and adding then in the default filter of `ChannelsViewModel` and `ChannelListViewModel`.

### 🧪 Testing

Ask for draft and hidden channels with and without this PR, you will see the difference =P.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
@@ -35,7 +35,7 @@ public class ChannelListViewModel(
         eq("type", "messaging"),
         `in`("members", listOf(chatDomain.currentUser.id)),
         or(Filters.notExists("draft"), ne("draft", true)),
-        eq("hidden", false)
+        or(Filters.notExists("hidden"), Filters.ne("hidden", true)),
     ),
     private val sort: QuerySort<Channel> = DEFAULT_SORT,
     private val limit: Int = 30,
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
@@ -5,13 +5,15 @@ import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.Transformations
 import androidx.lifecycle.Transformations.map
 import androidx.lifecycle.ViewModel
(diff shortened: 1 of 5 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -31,8 +33,9 @@ public class ChannelListViewModel(
     private val chatDomain: ChatDomain = ChatDomain.instance(),
     private val filter: FilterObject = Filters.and(
         eq("type", "messaging"),
-        Filters.`in`("members", listOf(chatDomain.currentUser.id)),
-        Filters.or(Filters.notExists("draft"), Filters.ne("draft", true)),
+        `in`("members", listOf(chatDomain.currentUser.id)),
+        or(Filters.notExists("draft"), ne("draft", true)),
+        ne("hidden", false)
```
JcMinarro, line 38: Shouldn't it be the opposite? :thinking: 
Why are we filtering for "channel with hidden not equal to false"?

leandroBorgesFerreira: This should be ne("hidden", true)).

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -31,8 +33,9 @@ public class ChannelListViewModel(
     private val chatDomain: ChatDomain = ChatDomain.instance(),
     private val filter: FilterObject = Filters.and(
         eq("type", "messaging"),
-        Filters.`in`("members", listOf(chatDomain.currentUser.id)),
-        Filters.or(Filters.notExists("draft"), Filters.ne("draft", true)),
+        `in`("members", listOf(chatDomain.currentUser.id)),
+        or(Filters.notExists("draft"), ne("draft", true)),
+        eq("hidden", false)
```
adasiewiczr, line 38: This filter won't work if extra data is null (backend issue, already reported). To workaround that, you can combine it with 'notExists' filter (just like for draft).
One more thing - shouldn't we also add this to VieModelFactories?

JcMinarro (approved):

PR #1857, opened 2021-05-12
[CAS-982] Client state service for pushes

ogkuzmin:
### 🎯 Goal

For push messages, we need to set user but without connection to the socket. It can be not established at all or just later.

### 🛠 Implementation details

1. First of all split `ClientState` to `SocketState` and `UserState` with according FSMs. Add tests for it.
2. Don't call connect socket when set user for pushes
3. Change `ChatClient::reconnectSocket` logic, cover it by unit tests.
4. Add new `DisconnectionCause` to `DisconnectedEvent`
5. Add new method `SocketListener::onDisconnected(cause: DisconnectCause)`
6. Add VALIDATION_ERROR as Server side error. Use it as unrecoverable error

### 🧪 Testing

1. Login with wrong credentials
2. Login with correct ones

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/ClientState.kt
```diff
@@ -1,45 +0,0 @@
-package io.getstream.chat.android.client.clientstate
-
-private typealias UserModel = io.getstream.chat.android.client.models.User
-
-internal sealed class ClientState {
```
ogkuzmin, line 5: Was split for SocketState, UserState

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
```diff
@@ -592,6 +593,7 @@ public data class ConnectingEvent(
 public data class DisconnectedEvent(
     override val type: String,
     override val createdAt: Date,
+    val disconnectCause: DisconnectCause = DisconnectCause.NETWORK_NOT_AVAILABLE
```
adasiewiczr, line 596:
```suggestion
    val disconnectCause: DisconnectCause = DisconnectCause.NETWORK_NOT_AVAILABLE,
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/SocketListener.kt
```diff
@@ -12,9 +13,16 @@ public open class SocketListener {
     public open fun onConnected(event: ConnectedEvent) {
     }
 
+    @Deprecated(
+        "Use method with DisconnectCause",
+        replaceWith = ReplaceWith("Use method with DisconnectCause instead of it")
```
adasiewiczr, line 18: Let's add level so it will be easier to update it in the future
```suggestion
        replaceWith = ReplaceWith("Use method with DisconnectCause instead of it"),
        level = DeprecationLevel.WARNING,
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -496,11 +510,13 @@ public class ChatClient internal constructor(
...
     public fun reconnectSocket() {
-        when (val state = clientStateService.state) {
-            is ClientState.Anonymous -> socket.connectAnonymously()
-            is ClientState.User -> socket.connect(state.user)
-            is ClientState.Idle -> {
+        when (socketStateService.state) {
+            is SocketState.Disconnected -> when (val userState = userStateService.state) {
+                is UserState.UserSet -> socket.connect(userState.user)
+                is UserState.Anonymous.AnonymousUserSet -> socket.connectAnonymously()
+                else -> error("Weird user state ${userState.javaClass.simpleName} without user being set!")
```
adasiewiczr, line 517: Is this intentional exception throwing? If yes, maybe we could add `@Throws` annotation, but maybe it would be better to log it instead

adasiewiczr (approved):
Great job! I just have one question and small suggestions

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: We did something similar with state management in ClientStateService previously. Such approach is tough but helps to prevent bugs and weird scenarios. I don't mind just logging. @JcMinarro what do you think?

ogkuzmin: Yes, it shouldn't be possible. But compilator let it happens

zsmb13:
```suggestion
                else -> error("Invalid user state $userState without user being set!")
```

I'd like to adjust the wording here a bit, "weird user state" is not something I want our clients to see in their logs 😄  

Also no need for the `javaClass.simpleName` solution if we make all the state objects used here either objects or data classes, so that they have a nice `toString` representation.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/DisconnectCause.kt
```diff
@@ -0,0 +1,5 @@
+package io.getstream.chat.android.client.clientstate
+
+public enum class DisconnectCause {
```
JcMinarro, line 3: Could it be a Sealed Class instead?
The "Error" case could have some "usefull info" about what happens, instead of only "error"

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/helpers/QueryChannelsPostponeHelper.kt
```diff
@@ -46,12 +46,10 @@ internal class QueryChannelsPostponeHelper(
...
             "Failed to perform job. Waiting for set user completion was too long. Limit of attempts was reached."
         }
-        return when (clientStateService.state) {
-            is ClientState.Idle -> error("User must be set before querying channels")
-            is ClientState.User.Authorized,
-            is ClientState.Anonymous.Authorized -> job()
-            is ClientState.User.Pending,
-            is ClientState.Anonymous.Pending -> {
+        return when (socketStateService.state) {
+            is SocketState.Idle -> error("Socket connection must be established before querying channels")
```
JcMinarro, line 50: What about if we allow user to use our ChatClient before the `SocketState` is initialized?
We can follow the same "delay" we have for Pending and Disconnected, and after X attempts, it will fail if they didn't connected

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/DisconnectCause.kt
ogkuzmin: What kind of info? ChatError? Where we can get it? We use it when `ChatSocketServiceImpl` we observe state transformation to `State.DisconnectedTemporarily`

JcMinarro (approved):
Looks great!!!!
So much clear the different states now.
I have left  some comments

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/DisconnectCause.kt
JcMinarro: I'm not sure if we could know "why" the error happend, but, keeping it as a Sealed Class, we are able to extend it easilly later. For now, if we don't have extra info for that, we can use only `object` instances

PR #1858, opened 2021-05-13
Add RepositoryFacade integration tests

bychkovdmitry:
### 🎯 Goal

Add a test which verifies that updating a message in the database works as expected after this fix:
https://github.com/GetStream/stream-chat-android/pull/1853/commits/f49ca7826214950c2f619907df20b73d07f92b47

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (4)


stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/repository/facade/RepositoryFacadeIntegrationTests.kt
```diff
@@ -0,0 +1,42 @@
...
+    fun `Given a message When persisting the message Should store html field`() = runBlocking {
+        val html = randomString()
+        val id = randomString()
+        val message = randomMessage(id = id, html = html)
+
+        repositoryFacade.insertMessages(listOf(message), cache = false)
+        val result = repositoryFacade.selectMessage(id)
+
+        Truth.assertThat(result).isNotNull()
+        Truth.assertThat(result!!.html).isEqualTo(html)
```
JcMinarro, line 40: What about to assert all fields of our message? :thinking: 
Because our `Message` class is a dataclass, check if the entity has the same values should be as simple as:
```
Truth.assertThat(result).isEqualTo(expectedMessage)
```

It will help us to verify any new field we add to our `Message` entity

bychkovdmitry: Not all the fields are stored in our database

bychkovdmitry: Overall, added 3 missing fields which were not persisted: `html`, `showInChannel` and `silent`. The `online` filed is ignored on purpose, there is no point to persist fields with quickly expiring values. 

`totalUnreadCount` and `unreadChannels` are also not persisted as they will be overwritten by null values coming from the server.

JcMinarro (approved):

PR #1873, opened 2021-05-19
Remove Awaitility dependency

zsmb13:
### 🎯 Goal

Remove the Awaitility dependency. It's used in a single test in very basic ways, we can replace it trivially with our own code. It's also not a great thing to use in tests anyway, as it employs polling.

### 🛠 Implementation details

It's a loop.

### ☑️ Checklist

- [x] New code is covered by unit tests ☑️
- [x] Reviewers added


stream-chat-android-client/src/androidTest/java/io/getstream/chat/android/client/poc/ClientInstrumentationTests.kt
```diff
@@ -114,9 +114,28 @@ internal class ClientInstrumentationTests {
...
+            if (predicate()) {
+                return@runBlocking
+            }
+
+            delay(100)
+            waited += 100
+        }
+
+        throw AssertionError("Predicate was not fulfilled within ${timeoutMs}ms")
+    }
```
ogkuzmin, line 140: Nit: I don't like mutability and while blocks. What do you think about tail recoursive version?
```suggestion
    tailrec suspend fun await(waited: Long, timeoutMs: Long) {
        return if (waited >= timeoutMs) {
            throw AssertionError("Predicate was not fulfilled within ${timeoutMs}ms")
        } else {
            if (predicate()) {
                return
            } else {
                delay(100)
                await(waited + 100)
            }
        }
    }

    runBlocking {
        await(0, timeoutSeconds * 1_000)
    }
```

ogkuzmin (approved):
LGTM

stream-chat-android-client/src/androidTest/java/io/getstream/chat/android/client/poc/ClientInstrumentationTests.kt
zsmb13: I'd prefer to use the more old school `while` loop solution in any serious production code, as I assume it'll be easier to understand - and if need be, change - for others on average than the recursive call.

However, for tests, especially for an obscure one like this, I'd be happy to drop in a local `tailrec suspend` function, because it's just a hilariously fun construct to have.

Sadly, there's a bug preventing us from using this. With the code here, the `waited` parameter would be forever stuck at `0`. See mostly [here](https://youtrack.jetbrains.com/issue/KT-38290), but also [here](https://youtrack.jetbrains.com/issue/KT-46218) and [here](https://youtrack.jetbrains.com/issue/KT-33982) for similar reported issues.

ogkuzmin: 😱  Wow, I didn't know about this bug. Thanks! 👍

PR #1874, opened 2021-05-19
Mark interfaces as sealed where possible

zsmb13:
https://stream-io.atlassian.net/browse/CAS-979

### 🎯 Goal

Kotlin 1.5 introduced sealed interfaces, which let us defined interfaces that can only be implemented in the same module and same package. We can use this to mark interfaces that only serve the purpose of hiding implementation classes, but should never be implemented by our clients.

### 🛠 Implementation details

Added the keyword to places.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListListenerContainerImpl.kt
```diff
@@ -1,4 +1,4 @@
-package io.getstream.chat.android.ui.message.list.adapter.internal
+package io.getstream.chat.android.ui.message.list.adapter
```
zsmb13, line 1: Had to move the impl class to the same package as the interface to comply with the requirements for sealed types.

adasiewiczr: Maybe we can remove this interface?

adasiewiczr (approved):
LGTM

bychkovdmitry (approved):

PR #1884, opened 2021-05-24
CAS-1019 Replace `FirebaseInstanceId` with `FirebaseInstallations`

samiuelson:
https://stream-io.atlassian.net/browse/CAS-1019

### 🎯 Goal

1. Update Firebase Cloud Messaging to the latest version.
2. Replace `FirebaseInstanceId` with `FirebaseInstallations` because `FirebaseInstanceId` is deprecated and not available anymore. 

### 🛠 Implementation details

Remove `ChatNotificationHandler::getFirebaseInstanceId()` and introduce `ChatNotificationHandler::getFirebaseInstallations()` instead.

### 🧪 Testing

Notifications should appear without issues.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/ChatNotificationHandler.kt
```diff
@@ -306,9 +306,9 @@ public open class ChatNotificationHandler @JvmOverloads constructor(
         )
     }
 
-    public open fun getFirebaseInstanceId(): FirebaseInstanceId? =
+    public open fun getFirebaseInstallations(): FirebaseInstallations? =
```
adasiewiczr, line 309: It's breaking change. I think we should deprecate it first

JcMinarro: Last version of FCM removed this class.
Because Firebase Dependencies can (and used to) be updated on the customer apps, we should try to use the most recent one that already removed this class.
For this concrete case I think we can bypass our "deprecation process"

samiuelson: To deprecate it first, we must not update the FCM lib to the latest version because the `FirebaseInstanceId` is not available anymore in this release. On the other hand, we need to update FCM to the latest version urgently because some clients are using the latest version of Firebase which overrides our transitive dependency in SDK (this is due to the fact Firebase uses BOM). This results in `FirebaseInstanceId` not being resolved. We decided to make an exception this time and that's a priority to make our SDK compatible with the latest version of Firebase ASAP.

adasiewiczr (approved):

PR #1890, opened 2021-05-25
[CAS-1024] Fix socket connection app background

ogkuzmin:
### 🎯 Goal

We need to release socket connection but not permanently and also not reconnect when in background

### 🛠 Implementation details

The main idea is to create a new state in ChatSocketServiceImpl in order to distinguish when we want disconnect and stop monitor but temporary in terms of not keeping resources

### 🧪 Testing
1. Go background
2. Check that connection is not established
3. Go in the foreground and see that new data is observed

### ☑️ Checklist

- [x] Reviewers added


adasiewiczr (approved):
LGTM

Issue #1889, opened 2021-05-25
Videos sent from React Sdk are not handled properly on Android Sdk.


harrunisk:
**Describe the bug**
We are using React and Android SDKs. Sending videos between the same SDKs is working properly. But when we try to send a video from React SDK to Android SDK, the video is not visible on the UI side(`MessageListView`). Meantime when we debugged the message list in the Android SDK, the message object which includes the video is available.

**SDK version**
- 4.10.0

**To Reproduce**
1. Send video by using React SDK  
2. Check the video on an android device

**Expected behavior**
Showing and playing video properly which sent by using React SDK.

**Device:**
 - Pixel 2 Api 30  
 - Xiaomi Redmi Note 5 api 28

**Screenshots**
An `Attachment` object sent by Android SDK  

![android](https://user-images.githubusercontent.com/31736480/119469598-2d4d3500-bd50-11eb-841b-e4665b145597.png)
An `Attachment` object sent by React SDK  

![web](https://user-images.githubusercontent.com/31736480/119469592-2cb49e80-bd50-11eb-9fc6-49009087e6c2.png)

ogkuzmin:
@harrunisk Thanks for submitting this issue! The patch has been merged, so it will be available in the next-release

JcMinarro:
Hello @harrunisk 
A new version has been released with the fix. You can check it [here](https://github.com/GetStream/stream-chat-android/releases/latest)

PR #1891, opened 2021-05-25
Fix/show attachment media activity for video mime type

ogkuzmin:
Related issue [here](https://github.com/GetStream/stream-chat-android/issues/1889)

### 🎯 Goal
Added new case for attachment when we should show AttachmentMediaActivity (activity with video player). It's for case when `type != "video"` but mime-type is video type

### ☑️ Checklist

- [x] Reviewers added


adasiewiczr (approved):
Nice catch! We should probably refactor those parts where we are using both `type` and `mimeType`.

PR #1900, opened 2021-05-26
[CAS-972] Refac - Removing current user usage

leandroBorgesFerreira:
[CAS-972](https://stream-io.atlassian.net/browse/CAS-972)

### 🎯 Goal

Remove currentUser from ChatDomain. 

### 🛠 Implementation details

Changing `currentuser` for `user.value`. In our sample app it is assumed that the user was set using ChatCllient, but in the SDK there are checks. 

### 🧪 Testing

Run the app and try to test everything you can =P.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


samiuelson (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -271,25 +273,25 @@ internal class ChatDomainImpl internal constructor(
...
 
-        currentUser = user
+        _user.value = user
+        _userLiveData.postValue(user)
 
         repos = RepositoryFacadeBuilder {
             context(appContext)
             database(db)
-            currentUser(currentUser)
+            _user.value?.let(::currentUser)
```
ogkuzmin, line 282: Maybe use `user` here?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -316,18 +318,15 @@ internal class ChatDomainImpl internal constructor(
...
 
         // if the user is already defined, just call setUser ourselves
-        val current = userOverwrite ?: client.getCurrentUser()
+        val current = client.getCurrentUser()
         if (current != null) {
             setUser(current)
         }
-        // past behaviour was to set the user on the chat domain
-        // the new syntax is to automatically pick up changes from the client
-        if (userOverwrite == null) {
```
ogkuzmin, line 325: Why did you just delete this? It was exposed to our customers, they could still use it. If we don't wanna let them pass `userOverwrite` in ChatDomain directly we should use our deprecation policy.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -371,7 +377,11 @@ public class ChannelController internal constructor(
     }
 
     internal suspend fun leave(): Result<Unit> {
-        val result = channelClient.removeMembers(domainImpl.currentUser.id).execute()
+        val currentUser = domainImpl.user.value
+
+        val result = currentUser?.let { user ->
+            channelClient.removeMembers(user.id).execute()
+        } ?: Result(ChatError("Current user is not set, then it is not possible to leave channel. Please set User for ChatDomain."))
```
ogkuzmin, line 384: Should we raise error here?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -790,10 +808,10 @@ public class ChannelController internal constructor(
      * If the request fails we retry according to the retry policy set on the repo
      */
     internal suspend fun sendReaction(reaction: Reaction, enforceUnique: Boolean): Result<Reaction> {
-        val currentUser = domainImpl.currentUser
+        val currentUser = domainImpl.user.value
```
ogkuzmin, line 811: Maybe we just need to skip all logic here if user is null

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/usecase/SearchUsersByName.kt
```diff
@@ -63,27 +65,35 @@ internal class SearchUsersByName(private val chatDomainImpl: ChatDomainImpl) {
...
     ): Result<List<User>> {
-        val filter = if (querySearch.isEmpty()) {
-            defaultUsersQueryFilter
-        } else {
-            Filters.and(
-                Filters.autocomplete(FIELD_NAME, querySearch),
-                Filters.ne(FIELD_ID, chatDomainImpl.currentUser.id)
-            )
-        }
+        val currentUser = chatDomainImpl.user.value
```
ogkuzmin, line 68: IMO it's better to have early return. It makes less nesting level

bychkovdmitry:
@leandroBorgesFerreira Could you explain what's the definition of done here? 
- Just using the new `LiveData` field and not crashing when the user is absent?
- Or subscribing to user state and initializing the `ViewModels` "lazily"? For example, when user is not set and `MessageListView` displays some empty placeholder (not sure about the design here). Once the user is set, `MessageListView` should display a list of messages without reinflating the view.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/internal/ChannelViewHolder.kt
```diff
@@ -204,7 +211,13 @@ internal class ChannelViewHolder @JvmOverloads constructor(
 
         lastMessage ?: return
 
-        lastMessageLabel.text = channel.getLastMessagePreviewText(context, channel.isDirectMessaging())
+        ChatDomain.instance().user.value?.let { user ->
+            lastMessageLabel.text =
+                channel.getLastMessagePreviewText(context, channel.isDirectMessaging(), user.asMention(context))
+        } ?: run {
```
ogkuzmin, line 217: Why we need `run { ... }` here?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/ReplyDecorator.kt
```diff
@@ -8,7 +8,7 @@ import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.Gip
 import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.MessagePlainTextViewHolder
 import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.TextAndAttachmentsViewHolder
 
-internal class ReplyDecorator(private val getCurrentUser: () -> User) : BaseDecorator() {
+internal class ReplyDecorator(private val getCurrentUser: () -> User?) : BaseDecorator() {
```
ogkuzmin, line 11: Seems like we should pass there Flow<User> or LiveData<User>. Maybe we can iterate over it and do later, but then we should put new task to our backlog

ogkuzmin (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -549,7 +550,7 @@ internal class ChatDomainImpl internal constructor(
         activeChannelMapImpl.values.toList()
 
     fun generateMessageId(): String {
-        return currentUser.id + "-" + UUID.randomUUID().toString()
+        return (_user.value?.id ?: "user_not_user") + "-" + UUID.randomUUID().toString()
```
zsmb13, line 553: Do we really want to silently swallow the error here? Not having a user for this function is an unexpected invalid state, isn't it?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -790,10 +808,10 @@ public class ChannelController internal constructor(
...
         reaction.apply {
             user = currentUser
-            userId = currentUser.id
+            userId = currentUser?.id ?: "user_not_set"
```
zsmb13, line 814: Why does it make sense to use this dummy ID here? Shouldn't we just fail the operation in some way if we don't have a user?

Additionally: what if a user's ID is `"user_not_set"`? I'd be very careful about using something like this anywhere in the SDK.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
```diff
@@ -69,7 +70,8 @@ internal class ChatDomainImpl internal constructor(internal val chatDomainStateF
             chatDomainStateFlow.currentUser = value
         }
 
-    override val user: LiveData<User?> = chatDomainStateFlow.user.asLiveData()
+    override val user: LiveData<User?>
+        get() = (chatDomainStateFlow as ChatDomainImpl)._userLiveData
```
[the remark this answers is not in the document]
leandroBorgesFerreira: asLiveData() starts with `value` as `null` even when the `value` is already set in the stateFlow. That makes our filtering to fail because it needs the CHatDomain.user to has a value right in it's initialization.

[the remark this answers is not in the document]
leandroBorgesFerreira: asLiveData doesn't pick the `.value` in the `StateFlow` so this value gets lost in the convertion and only the posted values get picked. 

That becomes a problem in the `ChannelListViewModel` as I don't see a way to wait to the currentUser.id as the filter needs it in the constructor. 

Only simple option that I see is to use the `offline.ChatDomain` (StateFlow) instead of `livedata.ChatDomain` inside `ChannelListViewModel`. That would change the API, but it seams a change that is going to happen anyways. What do you think about it?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
leandroBorgesFerreira: 🤔  Instead of returning the result with error? But wouldn't a crash be worse?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/internal/ChannelViewHolder.kt
leandroBorgesFerreira: removed

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/ReplyDecorator.kt
leandroBorgesFerreira: There was no need to make the return optional... that was probably a noise in the code. Removed.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
leandroBorgesFerreira: Removed now.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
leandroBorgesFerreira: Changed the signature for the method, now it is a pure function. No need to use `_user.value` anymore.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
leandroBorgesFerreira: done

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -182,14 +189,9 @@ internal class ChatDomainImpl internal constructor(
     private val _typingChannels = MutableStateFlow<TypingEvent>(TypingEvent("", emptyList()))
 
     private val _user = MutableStateFlow<User?>(null)
+    internal val _userLiveData = MutableLiveData<User?>()
```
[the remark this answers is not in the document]
leandroBorgesFerreira: Removed

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomainImpl.kt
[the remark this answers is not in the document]
leandroBorgesFerreira: We can use the new ChatDomain inside the ChannelListViewModel. Then we can remove this line.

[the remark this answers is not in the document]
leandroBorgesFerreira: I applied the new ChatDomain, then it is possible to remove this line.

leandroBorgesFerreira:
This PR is simply too big. Let me divide it into smaller tasks.

PR #1902, opened 2021-05-27
[CAS-1011] Add error feedback to MessageListView

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-1011

### 🎯 Goal

Add error toasts when something fails as a result of user actions

🛠 Implementation details

Handling errors for the following user actions:
- UI components
  - MessageListView
    - Mute/unmute user
    - Flag message
    - Block user
  - ChannelListView  
    - Hide channel
    - Delete channel
    - Leave channel
- Sample app
  - GroupChatInfoMemberOptionsDialogFragment
    - Remove member
  - AddGroupChannelFragment
    - Create channel
  - AddGroupChannelSelectNameFragment
    - Create channel
  - ChatInfoFragment
    - Mute/unmute channel
    - Block user
    - Delete channel
  - GroupChatInfoFragment
    - Change group name
    - Mute/unmute channel
    - Leave channel
  - GroupChatInfoAddUsersDialogFragment
    - Add member    

### ☑️ Checklist

- [x] Reviewers added

🎉 GIF

giphy (1)


stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
```diff
@@ -82,3 +83,20 @@ public fun <T : Any, K : Any> Call<T>.zipWith(call: Call<K>): Call<Pair<T, K>> {
 
 @InternalStreamChatApi
 public fun Call<*>.toUnitCall(): Call<Unit> = map {}
+
+private val onSuccessStub: (Any) -> Unit = {}
+private val onErrorStub: (ChatError) -> Unit = {}
+
+@InternalStreamChatApi
+public fun <T : Any> Call<T>.enqueue(
```
adasiewiczr, line 91: 🤔 Why it was added to `.api` file?

adasiewiczr (approved):

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
JcMinarro: Maybe because the annotation is not processed when the `.api` file is generated?
In the end, it is a public method that is accessible by the client

zsmb13: This is a bug in the binary validator plugin - it doesn't realize that the function that it generated a secondary synthetic `$default` function for (to support its default params in the bytecode) was originally annotated with our `@InternalStreamChatApi` annotation.

I've opened an issue for it here: https://github.com/Kotlin/binary-compatibility-validator/issues/58

PR #1904, opened 2021-05-27
CAS-900 Implement default Channel swipe actions behaviours

samiuelson:
https://stream-io.atlassian.net/browse/CAS-900

### 🎯 Goal
So far, channel options were shown to the user (when swiping channel item view) without default implementation. We shouldn't show clickable icons that do nothing by default.

### 🛠 Implementation details

1. Add default Channel "delete" option implementation coherent to default MessageListView's delete message implementation (showing confirmation dialog).
2. Make Channel "more" option not displayed by default since we're not able to provide any default implementation.

### 🧪 Testing
When you:
1. Remove `ChannelListView::configureDefaultMoreOptionsListener` invocation, 
2. Remove `app:streamUiChannelOptionsEnabled="true"` attribute from `fragment_channels.xml`, 
3. Remove `ChannelListView::setChannelDeleteClickListener` call,
and try out default implementation of Channel options provided by SDK.

Then:
1. "More options" should not be shown by default. 
2. "Delete channel" option should be shown only for the channel the current user is privileged to delete.
3. In case "More options" and "Delete channel" options are not available for a specific channel item swipe behavior should be disabled.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

swipe


stream-chat-android-ui-components/src/main/res/values/strings.xml
```diff
@@ -16,6 +16,10 @@
         <item quantity="one">%1d Member, %2d Online</item>
         <item quantity="other">%1d Members, %2d Online</item>
     </plurals>
+    <string name="stream_ui_channel_option_delete_confirmation_title">Delete Channel</string>
+    <string name="stream_ui_channel_option_delete_confirmation_message">Are you sure you want to permanently delete this message?</string>
```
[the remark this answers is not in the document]
samiuelson: I'd go for _conversation_ as it sounds generic enough. Anytime users can override it according to their preference.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModelBinding.kt
```diff
@@ -43,4 +45,18 @@ public fun ChannelListViewModel.bindView(
     view.setOnEndReachedListener {
         onAction(ChannelListViewModel.Action.ReachedEndOfList)
     }
+
+    view.setChannelDeleteClickListener {
+        AlertDialog.Builder(view.context)
```
zsmb13, line 50: The sample app now shows this new AlertDialog by default, because it calls `bindView` after setting up a bunch of listeners. We should fix that `bindView` call and make sure it happens before other customizations, and it'd be good to check the other `bindView` usages in the sample app as well.

zsmb13 (approved):
LGTM

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModelBinding.kt
samiuelson: Good catch 😓 Fixed.

PR #1908, opened 2021-05-28
CAS-880 Call `UploadAttachmentsWorker` in `ChatDomain::retryMessages`

samiuelson:
https://stream-io.atlassian.net/browse/CAS-880

### 🎯 Goal
Allow sending attachments while being offline.

### 🛠 Implementation details

Upload of pending attachments is done when the connection is recovered. When all pending attachments are uploaded the message is sent. 

### 🧪 Testing

1. Turn off data connection
2. Send message with attachments to a channel while being offline
3. Optional: kill the app and open it again
4. Turn on data connection
5. Attachments upload should resume and message should be send

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (1)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -760,25 +761,54 @@ internal class ChatDomainImpl internal constructor(
                 message.updatedAt != null || message.updatedLocallyAt != null -> {
                     client.updateMessage(message).execute()
                 }
-                else -> channelClient.sendMessage(message).execute()
+                else -> {
+                    val hasPendingAttachments = message.attachments.any {
```
bychkovdmitry, line 765: What do you think about adding `Message.attachmentSyncStatus` to Message? If the state is sync needed then we can start our worker, if sync is in progress then we do nothing, if attachments sync failed then we can mark the message as failed permanently.

samiuelson: Makes sense to me.

JcMinarro (approved):
It looks good and works fine

PR #1910, opened 2021-05-28
Fix publishing of jar artifacts

Miha-x64:
The original publishing script gave me good pom.xml for aar artifacts, but jar ones had no `<dependencies>` blocks. So here's a fix borrowed from Gradle docs.

### DIFF
--- a/scripts/publish-module.gradle
+++ b/scripts/publish-module.gradle
@@ -42,7 +42,7 @@ afterEvaluate {
                 if (project.plugins.findPlugin("com.android.library")) {
                     from components.release
                 } else {
-                    artifact("$buildDir/libs/${project.getName()}-${version}.jar")
+                    from components.java
                 }
 
                 artifact androidSourcesJar


zsmb13 (approved):
LGTM, but let's get a second review too, @JcMinarro

Miha-x64:
Oops, I'm sorry for misspelled 'of' in the commit message. Unfortunately, the change is done though GitHub interface and I can't reword the commit without checking out the whole project.

JcMinarro (approved):
Thanks for the PR

Miha-x64:
Thanks! Consider also updating [the article](https://getstream.io/blog/publishing-libraries-to-mavencentral-2021/) and [the gist](https://gist.github.com/zsmb13/56ed98c8fe916de441f2a9d8e060cd4a#file-publish-module-gradle-L38).

zsmb13:
@Miha-x64 Already have that on my TODO list, will update the article in all the different places where it's published as well 😄  Thanks for the catch again.

PR #1912, opened 2021-05-31
Applying dokkasarus only in the HTML output

leandroBorgesFerreira:
### 🎯 Goal

Fix dokka generation, it was crashing our javadocs generation. 

### 🛠 Implementation details

Instead of aplying dokkasaurus in all variants of dokka, now we only apply it in the HTML variant (yes, I know it doesn't make much sense, but the configuration of the multimodule tasks with dokkasarus would be harder than simply overriding the HTML output. We can improve this in the future if we need Dokka in the HTML format)

### 🧪 Testing

Run: `./gradlew androidSourcesJar javadocJar` and see that no conflict happens.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


stream-chat-android-client/build.gradle
```diff
@@ -107,5 +107,8 @@ dependencies {
     androidTestImplementation Dependencies.espressoCore
     androidTestImplementation Dependencies.androidxTestJunit
 
-    dokkaPlugin(Dependencies.dokkasaurus)
+    dokkaHtmlPlugin("org.jetbrains.dokka:dokka-base:1.4.32")
```
JcMinarro, line 110: Could we extract it into our `buildSrc` project and then use something like `Dependencies.dokkaBase`? :thinking:

leandroBorgesFerreira: No need to use dokka base anymore.

JcMinarro (approved):

PR #1921, opened 2021-06-02
[CAS-1046] Increasing message limit to 3

leandroBorgesFerreira:
[CAS-1046](https://stream-io.atlassian.net/browse/CAS-1046)

### 🎯 Goal

Recude the change to show an empty message.

### 🛠 Implementation details

The message limit our the Channels that we request in the ChannelListView is now 3, then we will show the last not deleted message even when the last and second last messages are deleted. 

### 🎨 UI Changes

Before: In the last channel the message show as empty, because the last message is deleted. 
After: The last non deleted message is not shown. 

| Before | After |
| --- | --- |
| Screenshot_20210602-175231 | Screenshot_20210602-175250 |

### 🧪 Testing

Delete the last message and notice that you won't see it without the PR in a fresh install of the app. 
### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/factory/ChannelListViewModelFactory.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/factory/ChannelListViewModelFactory.kt
@@ -28,7 +28,7 @@ public class ChannelListViewModelFactory @JvmOverloads constructor(
     ),
     private val sort: QuerySort<Channel> = ChannelListViewModel.DEFAULT_SORT,
     private val limit: Int = 30,
-    private val messageLimit: Int = 1,
+    private val messageLimit: Int = 3,
 ) : ViewModelProvider.Factory {
     override fun <T : ViewModel> create(modelClass: Class<T>): T {
         require(modelClass == ChannelListViewModel::class.java) {


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/factory/ChannelListViewModelFactory.kt
```diff
@@ -28,7 +28,7 @@ public class ChannelListViewModelFactory @JvmOverloads constructor(
     ),
     private val sort: QuerySort<Channel> = ChannelListViewModel.DEFAULT_SORT,
     private val limit: Int = 30,
-    private val messageLimit: Int = 1,
+    private val messageLimit: Int = 3,
```
bychkovdmitry, line 31: Just curious, what other SDKs are doing n that case?

leandroBorgesFerreira: RN: The show "message deleted", if the last message is deleted. So they don't have this problem.
Flutter: The same behaviour that we have.

bychkovdmitry: 🤔  I would prefer to show "message deleted", even if theirs deleted message is not visible in the chat itself

JcMinarro: I think we shouldn't update our MessageLimit only because this bad design thing.
I would prefer to follow the same RN approach, show that the last message was deleted.
This approach only cover the case where some of the last three messages wasn't deleted, but it is not a solution for all cases

tschellenbach: this will cause a regression for performance since the last message is available in offline storage. the last 3 require N extra queries

PR #1947, opened 2021-06-08
Fix QuerySort::hashcode

JcMinarro:
### 🎯 Goal
The `QuerySort::hashcode` method should return the same value allways that the content data is the same. This hashcode value is being used to obtain a `ChannelControler` from a map, if it is not calculated properly, new channelControlers are being used whenever a new QuerySort instance is used.

### 🛠 Implementation details
QuerySort used internally a property of a class that doesn't implemented the `hashcode()` method. Those classes have been changed to `data class` that create `hashcode()` method by us on compile time

### ☑️ Checklist

- [x] Reviewers added


ogkuzmin (approved):
Nice catch 👍

adasiewiczr (approved):

filbabic (approved):
Hey JC!

Thanks for this, I'm really happy we managed to find and resolve the issue. :]

PR #1967, opened 2021-06-10
[CAS-972] Removing current user from ChannelController

leandroBorgesFerreira:
[CAS-972](https://stream-io.atlassian.net/browse/CAS-972)

### 🎯 Goal

Remove ChatDomain.currentUser from ChannelController

### 🛠 Implementation details

Subscribing to user in typing, but using the value in the other fields as they expect an immediate value, not a stream of data.

### 🧪 Testing

Test that works as expected:
- Typing
- Reactions
- Unread count

### ☑️ Checklist

- [X] Reviewers added

### 🎉 GIF


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -574,7 +586,7 @@ public class ChannelController internal constructor(
             newMessage.enrichWithCid(cid)
         }
 
-        newMessage.user = domainImpl.currentUser
+        newMessage.user = domainImpl.user.value ?: return Result(ChatError("Current user null"))
```
ogkuzmin, line 589: What if we did it as first step of such method? There is no sense to execute any logic of this method if we don' have current user

ogkuzmin (approved):

PR #1990, opened 2021-06-18
[CAS-1089] Fix crash when adding reactions in threads

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-1089

### 🎯 Goal

Fix crash when adding reactions in threads

### 🛠 Implementation details

For some reason, we don't do caching when querying for messages in threads. Therefore, foreign key constraint for added reactions is not satisfied. The two possible approaches are:
1. Implement proper caching in threads
2. Don't store reactions via `ReactionsRepository` when adding a reaction in threads

This PR implements the second approach.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/9600921/122554364-8fdcdc80-d041-11eb-8471-20fc249ae106.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/9600921/122554263-6fad1d80-d041-11eb-8601-df2832377d45.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Steps to reproduce:
1. Sing in as user 1
2. Send a thread message
3. Sign in as user 2
4. Leave a reaction on the very same thread message

Expected result: 
The message contains 2 reactions: one reaction from user 1 and another one from user 2

Actual result:
Crash

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (2)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -804,7 +804,6 @@ public class ChannelController internal constructor(
                 deletedAt = Date()
             )
         }
-        domainImpl.repos.insertReaction(reaction)
```
bychkovdmitry, line 807: If I understand correctly, this line is redundant as reaction is saved as a part of the updated message several lines below.

tschellenbach: what about retries, network timeouts and the app getting killed by the OS? I mean its possible to never reach line 820 and 833 right?

bychkovdmitry: I've just rechecked and it seems to work fine as we persist the reaction on line 812:

```
val currentMessage = getMessage(reaction.messageId)?.copy() 
currentMessage?.let { // Don't understand how this can be null. If we are able to add a reaction then obviously the message should be present in `_messages` StateFlow
    it.addMyReaction(reaction, enforceUnique = enforceUnique)
    upsertMessage(it)
    domainImpl.repos.insertMessage(it) // Message is persisted here altogether with all the reactions 
}
```

Anyway, I'll ask for additional reviews on Monday. And we need to pay special attention to this functionality during regression testing before the next release.

tschellenbach: I don't think it retries the API call that way though. (it retries from the reactions table, not the messages table)

bychkovdmitry: Under the hood, `repos.insertMessage` persists reactions to a separate "reactions" table. If the sync status of a reaction is correct (sync_needed) then it should retry it.

```
internal abstract class MessageDao {

    @Transaction
    open suspend fun insert(messageEntities: List<MessageEntity>) {
        upsertMessageInnerEntities(messageEntities.map(MessageEntity::messageInnerEntity)) // messages table
        insertAttachments(messageEntities.flatMap(MessageEntity::attachments)) // attachments table
        insertReactions(messageEntities.flatMap { it.latestReactions + it.ownReactions }) // reactions table
    }
```

bychkovdmitry: Under the hood `insertMessage` persists reactions to the reactions table.

```
@Transaction
open suspend fun insert(messageEntities: List<MessageEntity>) {
        upsertMessageInnerEntities(messageEntities.map(MessageEntity::messageInnerEntity))
        insertAttachments(messageEntities.flatMap(MessageEntity::attachments))
        insertReactions(messageEntities.flatMap { it.latestReactions + it.ownReactions }) // persist to the reactions table with proper sync status (sync_needed)
}
```

tschellenbach: great. that seems good.

adasiewiczr (approved):

PR #1998, opened 2021-06-22
Update deprecations

adasiewiczr:
### 🎯 Goal

Update deprecations before release

### 🛠 Implementation details
Changing deprecation level or removing depends on DEPRECATIONS.md file.
NOTE:
- ChatDomain#currentUser deprecation level wasn't updated because it's still used in SDK and both sample apps. I'll add a task to remove all usages

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
giphy


samiuelson (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -178,14 +178,6 @@ public class MessageInputView : ConstraintLayout {
         refreshControlsState()
     }
 
-    @Deprecated(
```
bychkovdmitry, line 181: Added a note for myself to refactor the boilerplate around this block that was only necessary to support external SuggestionListView

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
```diff
@@ -84,7 +85,11 @@ public sealed interface ChannelController {
...
 
-    @Deprecated("Use offlineChannelData instead of it", replaceWith = ReplaceWith("ChannelController::offlineChannelData"))
+    @Deprecated(
+        level = DeprecationLevel.ERROR,
+        message = "Use offlineChannelData instead of it",
+        replaceWith = ReplaceWith("ChannelController::offlineChannelData"),
+    )
     public val channelData: LiveData<ChannelData>
 
     public val offlineChannelData: LiveData<OfflineChannelData>
```
bychkovdmitry, line 95: Are we going to rename this to `channelData` again after the method above is removed? The name is a bit confusing, looks like channels obtained only from the database.

bychkovdmitry (approved):
Great, a lot of code is now gone 👍

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
adasiewiczr: 👍 could you add a task so we won't forget about that?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/ChannelController.kt
adasiewiczr: @ogkuzmin what was the plan?

ogkuzmin: It's doubling. Since we expose live data ChannelData and have the same in flow version. We can't just delete it, that's why I had to deprecate old and expose the same from offline package with a little changed name

adasiewiczr: we can consider changing its name after we delete the old `channelData` but hopefully, we'll get rid of that at some point

PR #2002, opened 2021-06-23
[CAS-1074] Feature - Initializing repositories in the Worker

leandroBorgesFerreira:
[CAS-1074](https://stream-io.atlassian.net/browse/CAS-1074)

### 🎯 Goal

Sending attachments using the Worker event when the app is dead

### 🛠 Implementation details

Repository facade is initialized in ChatDomain inside Worker in case it is not initialized yet.

### 🧪 Testing

Select an attachment with no internet on your phone.
Check that is not send.
Kill the app.
Enable internet
Check in out logcat that the attachment is send.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
```diff
@@ -65,15 +65,20 @@ internal class MessageSendingService(
...
-                doSend(message)
+        return when {
+            domainImpl.online.value ->
+                if (message.hasPendingAttachments()) {
+                    waitForAttachmentsToBeSent(message)
+                } else {
+                    doSend(message)
+                }
+
+            message.hasPendingAttachments() -> waitForAttachmentsToBeSent(message)
```
ogkuzmin, line 76: Why do you want to wait for attachments if there is no online?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -315,6 +318,23 @@ internal class ChatDomainImpl internal constructor(
         initClean()
     }
 
+    internal fun initRepositoryFacade(user: User) {
```
ogkuzmin, line 321: This method isn't invoked anywhere

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -170,6 +170,8 @@ internal class ChatDomainImpl internal constructor(
     internal val job = SupervisorJob()
     internal var scope = CoroutineScope(job + DispatcherProvider.IO)
 
+    internal var isRepositoryInitialized = false
```
ogkuzmin, line 173: Do we really need this? Seems like `user.value != null` means the same

ogkuzmin (changes requested):
Leo, could you also add:
1) Clean job if there is no credentials/no attachments in DB to be uploaded
2) Unit tests for `UploadAttachmentsWorker`
?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
leandroBorgesFerreira: At this point, yes, it is the same. But we would be connecting something that don't need to be connected. At a later moment, we could decide that the initialization of RepositoryFacede is not inside `setUser`, by using `user.value != null` we could create a hidden bug in the future.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
leandroBorgesFerreira: My bad

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
leandroBorgesFerreira: Makes sense, there's no need to wait. It only enqueues the message

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
```diff
@@ -1,19 +1,40 @@
...
+        chatDomain: ChatDomainImpl,
+        chatClient: ChatClient
+    ): Result<Unit> {
         return try {
-            val domainImpl = (ChatDomain.instance() as ChatDomainImpl)
+            val domainImpl = chatDomain.apply {
+                if (!isRepositoryInitialized) {
+                    chatClient.setUserWithoutConnectingIfNeeded()
+
+                    chatClient.getCurrentUser()?.let(::initRepositoryFacade)
```
adasiewiczr, line 23: I'm not sure if this will work correctly. `chatClient.setUserWithoutConnectingIfNeeded()` sets the user,  then notifies `ChatDomain` which does the whole initialization. Why do we need to do it manually?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
ogkuzmin: 1. In my opinion, the boolean flag`isRepositoryInitialized` is a source of future bugs. It's always easy to forget to set it, to rewrite it wrongly, and so on. 
If you want to check that the repo has been initialized with data for some particular user you can do it. When we create stub-version in `createNoOpRepos` we don' pass to `RepositoryFacadeBuilder` any value for current user and we do it when we create real repository. You can use it, pass new field to `RepositoryFacade::currenUser: User?` and check if it's null or no for the same reason
2. IMO thinking about situations in the future when we want to distinguish the process of `setUser` and initializing of the repository is too early optimization. And it's a known common evil practice in CS. Let's do it when we really want. Right now the logic of our `ChatDomain` doesn't divide situations when the user is set and his data is cached in repo. IMO exposing `initRepositoryFacade` is not right, it's violation of encapsulation for ChatDomain

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
ogkuzmin, line 23: We already have an approach on how to init data with a user, and how to init all required stuff (repo with cache in this case), so we shouldn't do anything else except setting a user here.

ogkuzmin (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
adasiewiczr: We just need to somehow wait until ChatClient notifies ChatDomain about new user

ogkuzmin: I see it's done in synchronous way. Should work fine, but better to debug
```kotlin
private fun notifySetUser(user: User) {
        preSetUserListeners.forEach { it(user) }
    }
```

leandroBorgesFerreira: Ok. Now this flag is removed and no manual initialization of ChatDomain is happnening

leandroBorgesFerreira: You're right, no need to do it manually.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
leandroBorgesFerreira: Ok, removing the flag

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
leandroBorgesFerreira: Actually, no need for initRepositoryFacade as ChatClient already does this work.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -321,16 +321,20 @@ public class ChatClient internal constructor(
      * Moreover, it warms up the connection, and sets up notifications.
      *
      */
-    private fun setUserWithoutConnectingIfNeeded() {
+    @InternalStreamChatApi
+    public fun setUserWithoutConnectingIfNeeded(): Boolean {
```
ogkuzmin, line 325: The method with such a name shouldn't return anything. It's hard to treat output value. Do we really need it?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
```diff
@@ -65,15 +65,23 @@ internal class MessageSendingService(
...
-                doSend(message)
+        return when {
+            domainImpl.online.value ->
+                if (message.hasPendingAttachments()) {
+                    waitForAttachmentsToBeSent(message)
+                } else {
+                    doSend(message)
+                }
+
+            message.hasPendingAttachments() -> {
```
ogkuzmin, line 76: I think if we're offline we don't need to do anything at all. Just return what we have in else branch. Because when connection recovered we have handling logic already for this

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: Maybe it will be clearer if we can have something like:
```kotlin 
fun containsStoredCredentials(): Boolean {
  return  encryptedPushNotificationsConfigStore.get() != null
 }
 ```

File: stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt (hunk 6 above)
Comment (leandroBorgesFerreira) [reply stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt:76]: If we do this, then we won't enqueue the attachments and it won't be sent in the background. Basically the ticket won't be done

Comment (leandroBorgesFerreira) [reply stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt:76]: Our logic for recovered connection only works if the app is in the foreground as far as I know. 🤔  If that's what we want, then there's no need to merge this PR

Comment (adasiewiczr) [reply stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt:76]: `domainImpl.online` won't be `true` if WebSocket is disconnected and we're not connecting it while the app is in the background. Maybe we should distinguish between no internet connection and socket disconnected somehow 🤔

File: stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt [hunk 7]
```diff
@@ -333,6 +335,11 @@ public class ChatClient internal constructor(
         }
     }
 
+    @InternalStreamChatApi
+    public fun containsStoredCredentials(): Boolean {
+        return encryptedPushNotificationsConfigStore.get() != null
```
adasiewiczr, line 340: This is no longer `pushNotificationsConfigStore`. Maybe we should rename it to something more meaningful, for example `EncryptedUserConfigStorage`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
ogkuzmin: Yes, it can work! But for then to exclude double sending, before to start send attachments we need to update their status in DB for `InProgress`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
```diff
@@ -102,12 +110,16 @@ internal class MessageSendingService(
                     }
             }
             )
+        enqueueMessage(newMessage)
+        return Result.success(newMessage)
+    }
+
+    private fun enqueueMessage(message: Message) {
```
ogkuzmin, line 117: To be clear, maybe rename this method name as `enqueueAttachmentUploadForMessage` or something similar?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
```diff
@@ -1,24 +1,51 @@
...
             } else {
-                Result.error(ChatError())
+                val hasPendingAttachment = message.attachments.any { attachment ->
+                    attachment.uploadState == Attachment.UploadState.InProgress
+                }
+
+                if (!hasPendingAttachment) {
+                    return Result.success(Unit)
+                }
+
```
ogkuzmin, line 41: Seems like here we should insert message with updated attachments (their `uploadState` should be `InProgress`) for attachments that haven't been sent successfully yet

ogkuzmin (commented):
Nice job! Thank you, Leo, for addressing the comments. Last notes

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
leandroBorgesFerreira: I see that already happens in the `ChannelController.uploadAttachments`. It calls `domainImpl.repos.insertMessage(message)` after the upload happens. Then the message will be updated with the new status of its attachments

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
leandroBorgesFerreira: I believe this is already happening here:
```
.onSuccessSuspend { newMessage ->
            // we insert early to ensure we don't lose messages
            domainImpl.repos.insertMessage(newMessage)
            domainImpl.repos.updateLastMessageForChannel(newMessage.cid, newMessage)
        }
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
ogkuzmin: But just before sending we don't update their state. And consider we're opening app, `connectionRecovered` is triggered, we're checking for messages that wait for attachments and filter out those who's attachment have `InProgress` state, but we don't have such, then we send request to upload again. Theoretically we can have double sending job

ogkuzmin (approved):
Good job!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
ogkuzmin: Discussed it. It's already done in `MessageSendingService`

leandroBorgesFerreira: documentation: We talked about that in Slack

adasiewiczr (approved):

PR #2024, opened 2021-07-01
[CAS-1110] Allow adding custom HTTP headers to image requests

adasiewiczr:
### 🎯 Goal

Allow adding custom HTTP headers to image requests

### 🧪 Testing

Add custom image request header and check if it's included in the image request

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/images/StreamImageLoader.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/images/StreamImageLoader.kt
@@ -14,7 +14,7 @@ public sealed interface StreamImageLoader {
         public fun instance(): StreamImageLoader = CoilStreamImageLoader
     }
 
-    public var imageRequestHeaders: Map<String, String>
+    public var imageHeadersProvider: ImageHeadersProvider
 
     public fun load(
         target: ImageView,
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/images/StreamImageLoader.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/images/StreamImageLoader.kt
@@ -14,6 +14,8 @@ public sealed interface StreamImageLoader {
         public fun instance(): StreamImageLoader = CoilStreamImageLoader
     }
 
+    public var imageRequestHeaders: Map<String, String>
+
     public fun load(
         target: ImageView,
         data: Any?,
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
@@ -1,6 +1,7 @@
 package io.getstream.chat.android.ui
 
 import android.content.Context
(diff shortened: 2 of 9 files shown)


ogkuzmin (commented):
Do we need something like this for new UI?

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUI.kt
```diff
@@ -84,6 +85,10 @@ public class ChatUI internal constructor(
             this.strings = strings
         }
 
+        public fun withImageRequestHeaders(imageRequestHeaders: Map<String, String>): Builder = apply {
+            StreamImageLoader.instance().imageRequestHeaders = imageRequestHeaders
+        }
+
```
zsmb13, line 91: We probably don't need to add this to the old implementation

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
```diff
@@ -31,6 +33,9 @@ public object ChatUI {
     public var style: ChatStyle = ChatStyle()
     public var navigator: ChatNavigator = ChatNavigator()
     public var urlSigner: UrlSigner = UrlSigner.DefaultUrlSigner()
+    public var imageRequestHeaders: Map<String, String> by Delegates.observable(emptyMap()) { _, _, newState ->
+        StreamImageLoader.instance().imageRequestHeaders = newState
+    }
```
zsmb13, line 38: We can probably avoid storing this in ChatUI altogether like this:

```suggestion
    public var imageRequestHeaders: Map<String, String>
        get() = StreamImageLoader.instance().imageRequestHeaders
        set(value) {
            StreamImageLoader.instance().imageRequestHeaders = value
        }
```

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/images/StreamImageLoader.kt
```diff
@@ -14,6 +14,8 @@ public sealed interface StreamImageLoader {
         public fun instance(): StreamImageLoader = CoilStreamImageLoader
     }
 
+    public var imageRequestHeaders: Map<String, String>
```
zsmb13, line 17: This approach of having a `Map` makes the set of headers included static. Is that the requirement?

Otherwise we could have some `HeaderProvider` class that behaves similarly to `UrlSigner` that we keep invoking to get this data every time we load an image.

adasiewiczr: That might be a good idea. Let me change that

stream-chat-android/src/main/java/com/getstream/sdk/chat/ChatUI.kt
adasiewiczr: We do, it will be used in the old one

adasiewiczr:
@ogkuzmin I think we should keep it consistent and probably we'll need to add it at some point

zsmb13 (approved):
Looks good - would be nice to also add docs for the new public declarations, but then we're done

ogkuzmin (approved):

PR #2028, opened 2021-07-01
Fix - Stop filtering messages with sent attachments

leandroBorgesFerreira:
### 🎯 Goal

Fix the attachment sending in the background.

### 🛠 Implementation details

Removing the filter for only `InProgress` attachments. It is possible that the attachment is sent, but the message still wasn't sent. This is the case of this bug, the attachments were send by out Worker, but the message is not sent yet.

### 🧪 Testing

Case 1  - 
Send the attachment without internet.
Enabled the internet
Check that the message it sent.

Case 2 - 
Send the attachment without internet.
Close and kill the app
Enabled the internet
Come back to the app and the same channel.
Check that the message it sent.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (1)

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
@@ -2,7 +2,6 @@ package io.getstream.chat.android.offline
 
 import android.content.Context
 import android.os.Handler
-import android.util.Log
 import androidx.annotation.VisibleForTesting
 import io.getstream.chat.android.client.BuildConfig.STREAM_CHAT_VERSION
 import io.getstream.chat.android.client.ChatClient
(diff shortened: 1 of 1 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -730,7 +730,7 @@ internal class ChatDomainImpl internal constructor(
 
     private suspend fun retryMessagesWithAttachments(): List<Message> {
         val retriedMessages = repos.selectMessagesWaitForAttachments()
-            .filter { message -> message.attachments.any { it.uploadState == Attachment.UploadState.InProgress } }
```
ogkuzmin, line 733: I don't understand why removing solves this. You wrote in description that case is "attachments are already sent, but message isn't yet". But if attachment is sent it means that it has `uploadState == UploadState.Success`. And all attachments with `InProgress` state are already processing by worker

ogkuzmin (approved):
Agree to fix doubling in another PR

PR #2032, opened 2021-07-02
CAS-1057 Bug: Edit message when offline

samiuelson:
https://stream-io.atlassian.net/secure/RapidBoard.jspa?rapidView=81&projectKey=CAS&selectedIssue=CAS-1103

### 🎯 Goal
Allow editing message offline without issues.

There are multiple async operations executed in the `ChatDomainImp::connectionRecovered` method. Those operations affect a single DB table `stream_chat_message`. Resulting in unpredictable and wrong behavior related to race conditions. 

In ~80% of cases message offline message editing fails because before `::retryEntities` method is called from the `::connectionRecovered` method the edited message is overridden in DB already by the async logic intended to update Channel data.

This PR fixes those race conditions and synchronizes `ChatDomain::retryFailedEntities` suspending function.

### 🛠 Implementation details

1. Remove race conditions caused by async operations performed by `ChatDomainImpl::connectionRecovered` method, where the shared resource is `DB.stream_chat_message` table.
2. Synchronize `ChatDomainImpl::retryFailedEntities` method using `Mutex`. It's called from `::connectionRecovered` method and from `HealthEvent` and we're not guaranteed it's called on the same thread.

### 🎨 UI Changes

| Before | After |
| --- | --- |
|<video src="https://user-images.githubusercontent.com/4527432/124447813-0851d600-dd82-11eb-818b-427213872d8b.mp4"/>|<video src="https://user-images.githubusercontent.com/4527432/124447847-10aa1100-dd82-11eb-9f75-dd795e8c1418.mp4"/>|

### 🧪 Testing

When the message is edited while the device is offline, after switching to the online message should not be reverted to the old one.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -681,23 +690,21 @@ internal class ChatDomainImpl internal constructor(
...
-        // 4 retry any failed requests
-        if (online) {
-            retryFailedEntities()
-        }
     }
 
+    // An attempt to synchronize ::retryFailedEntities execution since it is called from multiple places. The shared resource is DB.stream_chat_message table.
+    private val entitiesRetryMutex = Mutex()
     internal suspend fun retryFailedEntities() {
-        delay(1000)
```
samiuelson, line 695: Good bye, old friend!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
```diff
@@ -101,6 +101,9 @@ internal class EventHandlerImpl(
                     }
                 }
                 is HealthEvent -> {
+                    // TODO: consider removing caling the code below on every HealthCheckEvent
+                    // We have retry logic for the most crucial network calls. Why would we need to rely on another mechanism for doing retries?
+                    // We already retry entities inside `domainImpl.connectionRecovered` which is called when the ConnectedEvent is received ☝️
                     domainImpl.scope.launch {
```
samiuelson, line 107: Can we remove this call and do nothing on HealthEvent as described ? 🤔

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -681,23 +689,21 @@ internal class ChatDomainImpl internal constructor(
...
         }
-
-        // 4 retry any failed requests
-        if (online) {
-            retryFailedEntities()
-        }
     }
 
+    // Synchronizing ::retryFailedEntities execution since it is called from multiple places. The shared resource is DB.stream_chat_message table.
+    private val entitiesRetryMutex = Mutex()
```
ogkuzmin, line 698: Maybe move to fields declaration place?

ogkuzmin (approved):
Good job, Sam!

adasiewiczr (approved):

PR #2034, opened 2021-07-05
[CAS-1106] Use filter in ChannelsVM as live data

ogkuzmin:
### 🎯 Goal

Our default filter object for channel list depends on current user. Since current user is reactive source, then filter object must be reactive as well. This should prevent race conditions when user isn't set yet but VM is already being used

### 🛠 Implementation details

1. Make filter object is optional in VM constructor and VM factory
2. Use by default reactive value for the filter object

### 🧪 Testing

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/extensions/Filters.kt
```diff
@@ -0,0 +1,16 @@
+package com.getstream.sdk.chat.utils.extensions
+
+import io.getstream.chat.android.client.api.models.FilterObject
+import io.getstream.chat.android.client.models.Filters
+import io.getstream.chat.android.client.models.User
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+
+@InternalStreamChatApi
+public fun Filters.defaultChannelListFilter(user: User?): FilterObject {
+    val userFiler = if (user == null) neutral() else `in`("members", listOf(user.id))
```
adasiewiczr, line 10: 🤔 Should we use a `neutral` filter if a user is not set? Consider the following scenario:
1. I'm connecting the user at the same time that I'm showing ChannelListView
2. Channels are loaded faster than the user is connected - I can see channels in which I'm not the member, but I can't view them because I don't have permissions
3. Channels are reloaded because the user was connected - wrong channels disappear

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -66,11 +68,17 @@ public class MessageListViewModel @JvmOverloads constructor(
...
     public val state: LiveData<State> = stateMerger
-    public val currentUser: User
-        get() = domain.user.value!!
 
     public val user: LiveData<User?> = domain.user
 
+    @Deprecated(
+        message = "Use user live data instead",
+        replaceWith = ReplaceWith("user.value!!"),
+        level = DeprecationLevel.WARNING
```
adasiewiczr, line 77:
```suggestion
        level = DeprecationLevel.WARNING,
```

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -66,11 +68,17 @@ public class MessageListViewModel @JvmOverloads constructor(
...
      */
     public val state: LiveData<State> = stateMerger
-    public val currentUser: User
-        get() = domain.user.value!!
 
     public val user: LiveData<User?> = domain.user
 
+    @Deprecated(
+        message = "Use user live data instead",
+        replaceWith = ReplaceWith("user.value!!"),
```
adasiewiczr, line 76:
```suggestion
        replaceWith = ReplaceWith("user.value"),
```

adasiewiczr (approved):

PR #2042, opened 2021-07-08
[CAS-1113] Add ChatClient::partialUserUpdate for partial user updates

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-1113

### 🎯 Goal

Add possibility for partial user updates

### 🛠 Implementation details

Added `ChatClient::partialUserUpdate`.
NOTE: You can only partially update `currentUser` using client-side API.

### 🧪 Testing
Try to partially update a user.

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -1084,6 +1084,11 @@ public class ChatClient internal constructor(
         set: Map<String, Any> = emptyMap(),
         unset: List<String> = emptyList(),
     ): Call<User> {
+        if (id != getCurrentUser()?.id) {
+            logger.logE("The client-side partial update allows you to update only the current user. Make sure the user is set before updating it.")
+            return ErrorCall(ChatError("The client-side partial update allows you to update only the current user. Make sure the user is set before updating it."))
+        }
+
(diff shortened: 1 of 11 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1069,6 +1069,28 @@ public class ChatClient internal constructor(
...
+     *
+     * @return executable async [Call]
+     */
+    @CheckResult
+    public fun partialUpdateUser(
+        id: String,
+        set: Map<String, Any> = emptyMap(),
+        unset: List<String> = emptyList(),
+    ): Call<User> {
+        return api.partialUpdateUser(
```
ogkuzmin, line 1087: What about add guard checks? We need to have user being set, we need to have set user with id that equals to argument. Can be platform side verification

adasiewiczr: I'm not sure if we should do them client-side. The server error is meaningful and we'll have to do additional changes in our SDK if this requirement changes on the server side.

adasiewiczr: I'll add a check

ogkuzmin (approved):

PR #2079, opened 2021-07-23
Make Channel::getLastMessage public

adasiewiczr:
### 🎯 Goal

Simplify getting the channel's last message. 

### 🛠 Implementation details

I've made `Channel::getLastMessage` extension function public so it can be reused in custom view holders.

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/Channel.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/Channel.kt
@@ -4,8 +4,10 @@ import android.content.Context
 import androidx.annotation.StringRes
 import com.getstream.sdk.chat.utils.extensions.getUsers
 import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.models.name
 import io.getstream.chat.android.ui.R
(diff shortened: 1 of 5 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/Channel.kt
```diff
@@ -26,3 +28,18 @@ public fun Channel.getDisplayName(
...
+ * @return Last message from the channel or null if it doesn't exist
+ */
+public fun Channel.getLastMessage(): Message? =
+    messages.asSequence()
+        .filter { it.createdAt != null || it.createdLocallyAt != null }
+        .filter { it.deletedAt == null }
+        .filter { !it.silent }
+        .filter { it.user.isCurrentUser() || !it.shadowed }
+        .filter { it.isRegular() || it.isSystem() }
+        .maxByOrNull { it.getCreatedAtOrThrow() }
```
ogkuzmin, line 45: Seems scary. What about to use here `getCreatedAtOrNull` instead of `getCreatedAtOrThrow`?

adasiewiczr: This should never happen because we are filtering all channels with `createdAt == null` before

ogkuzmin (approved):

PR #2094, opened 2021-07-28
Bump Compose version to 1.0.0

zsmb13:


filbabic (approved):
Props for taking this sweet sweet moment from me, of updating it to stable.

11/10. This makes more sense now.

buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
```diff
@@ -25,9 +25,9 @@ object Versions {
     internal const val ASSERTJ = "3.20.2"
     internal const val COIL = "1.3.0"
     internal const val COMPOSE_ACCOMPANIST = "0.14.0"
```
filbabic, line 27: This will probably happen today/tomorrow too, so we can be adamant of getting it up to date tomorrow I guess

PR #2124, opened 2021-08-02
2123 - Improved the way we send data in the composer

### LINKED ISSUE #2123: Expose a Message object in onSendMessage, in the MessageComposer
To allow our users to customize messages before they're sent, we should expose a `Message` instead of a `String` when the user taps on the **Send Message** button.

Action points:
- [x] Update the `onSendMessage` callbacks to accept a `Message`
- [x] Update the `MessageComposer` and `MessageInput` to create a message when tapping on the button
- [x] Saturate the message with any required data (e.g. attachments, parent message IDs), before passing it to the `onSendMessage` callback.
- [x] Propagate the message to our users/ViewModel

filbabic:
Fixes #2123  

### 🎯 Goal

We're exposing a `Message` in the `MessageComposer`, so that our users can customize/intercept the message before it's sent.

### 🛠 Implementation details

Built the message using our `ViewModel`, in the **bound** version of the `MessageComposer` and sent it through `onSendMessage()`.

### 🧪 Testing

The functionality should work the same, but we should now be able to intercept the `Message` and add any required data to it.

### ☑️ Checklist

    - [x] [Tutorial changes](https://github.com/GetStream/stream/pull/2835)
- [x] Reviewers added

### 🎉 GIF

giphy


zsmb13 (commented):
What about the missing items of the PR template checklist? Should we update the changelog & docs?

filbabic:
@zsmb13 Yeah we should. I'll do that tomorrow, make sure we get a good pass and everything is solid, before we release something.

adasiewiczr (approved):

PR #2125, opened 2021-08-03
[CAS-1109] Feature - Force light mode for message list view

leandroBorgesFerreira:
[CAS-1109](https://stream-io.atlassian.net/browse/CAS-1109)

### 🎯 Goal

Create the possibility to configure the app to force light theme. This PR creates the configuration for MessageListView. 

### 🛠 Implementation details

All the colours will have a different default if light theme if forced. This will respect the selected color, if this is was selected, but the default colours will be for the light theme if the configuration `streamUiForceLightMode` is set to `true`

### UI Changes

Forced light theme for MessageListView:
Screenshot_20210803-113544

### 🧪 Testing

Set `streamUiForceLightMode` to true, enable dark mode on your device and test everything you can in the `MessageListView`. Check that the colours are the same in the light mode.

### ☑️ Checklist

- ~[ ] Changelog is updated with client-facing changes.~ I'll add the documentation when all the views are done.
- ~[ ] Affected documentation updated (docusaurus, tutorial, CMS (task created)).~ I'll add the documentation when all the views are done.
- [x] Reviewers added. todo

### 🎉 GIF

giphy
_Light theme_


JcMinarro (changes requested):
Does this approach work when color/drawable/blablabla are configured by xml properties?
I am not sure it will work on that case and seems a complex way because we need to be sure we are using the "force" property on all case where we are calling "getColor/getDrawable". Because it is not an standard way, a new developer that comes to the team (or even us that "already know about it") could introduce an error on there because using the standard method (getColor/getDrawable without the forze attribute).

All our views use a the extension `Context.createStreamThemeWrapper()` to have a `ContextThemeWrapper` with the proper style it needs to use to inflate our views. We should be able to add there the needed logic to use the dark mode on case a property is configured.
I suggest to create a property into `ChatUI` object, and, inside of the previous method I shown, use this property to decide which theme needs to be used. For that, the only thing you will need to do is, if the property "forze drak mode" is true, create a new context and override the configuration inside of it with the forze dark mode enabled

leandroBorgesFerreira:
@JcMinarro Suggestion is a better way to do this.

PR #2137, opened 2021-08-04
PoC ChatDomain in LLC - alternative way of getting state

adasiewiczr:
### 🎯 Goal

Show alternative way of getting state

### 🛠 Implementation details

The idea is based on [DocumentReference](https://firebase.google.com/docs/reference/node/firebase.firestore.DocumentReference) class from Firestore that gives you two ways for getting document - either getting a single doc or subscribing to doc snapshots. 
PR adds `QueryReference` class that exposes two functions:
- `get()` - that returns `Call<T>` (similarly to `DocumentReference::get`)
- `asState()` - added as an extension function inside `offline` module (because state is not available in LLC) (similarly to `DocumentReference::onSnapshot`)

This way we can simplify getting state because users can always decide if they want to take a single `Call` or `State`. Moreover, it reduces the number of operations needed to subscribe to state updates and removes the additional call from [here](https://github.com/GetStream/stream-chat-android/compare/poc/mergeDomain2...poc/mergeDomain3#diff-62f3e6e8b5b7c39ba5b41a51e957365edb94741923db9d83d47aea5e97ef31ebR112). There is a downside - we'll need to change queries return types inside LLC.


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/adapter/ChatClientReferenceAdapter.kt
```diff
@@ -0,0 +1,9 @@
+package io.getstream.chat.android.offline.plugin.adapter
+
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.client.api.models.QueryChannelsRequest
+import io.getstream.chat.android.offline.querychannels.QueryChannelsReference
+
+public class ChatClientReferenceAdapter(private val chatClient: ChatClient) {
```
ogkuzmin, line 7: here we will add all requests that available currently in `ChatClient` and wrap them into `QueryReference<T,S>`

PR #2140, opened 2021-08-04
2138 - Exposed `isMine` on MessageItem.kt

### LINKED ISSUE #2138: Expose a property on `MessageItem` if the message is mine or not
We want users to be able to know if each `MessageItem` is owned by the currently logged in user, or another user.

In the core SDK, we exposed an `isMine` property. We can do the same for Compose.

Action points:
- [x] Expose `isMine` in `MessageItem`.
- [x] Compute the property value wherever `MessageItem`s are created
- [x] Use the `isMine` property instead of comparing user IDs

filbabic:
Fixes #2138.

Cleaned up some code, update docs as well. 

Essentially our users will be able to know what messages are theirs, when they consume/override our `itemContent` in `MessageList`.

We also cleaned up some properties, redundant TODOs and minor customization access for `MessageList`. 

### 🎯 Goal

We want users to be able to know which message is theirs and which is someone else's, when they override `MessageList`'s `itemContent` parameter.

### 🛠 Implementation details

Exposed a new property in `MessageItem`, removed usage of `currentUser` and ID comparison outside of our `ViewModel`.

### 🧪 Testing

No behavior/UI changes should happen, this is just a code accessibility improvement.

### ☑️ Checklist

    - [ ] We might want to update a tutorial part where we use the `MessageList`. I'll do that today. 
- [x] Reviewers added

### 🎉 GIF


docusaurus/docs/Android/04-compose/04-message-components/03-message-list.md
```diff
@@ -109,15 +109,14 @@ fun MessageList(
         DefaultMessageContainer(
             message = message,
             onThreadClick = onThreadClick,
-            onLongItemClick = onLongItemClick,
-            currentUser = viewModel.currentMessagesState.currentUser
+            onLongItemClick = onLongItemClick
```
adasiewiczr, line 112:
```suggestion
            onLongItemClick = onLongItemClick,
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/items/MessageItem.kt
```diff
@@ -6,4 +6,5 @@ public data class MessageItem(
     val message: Message,
     val position: MessageItemGroupPosition,
     val parentMessageId: String? = null,
+    val isMine: Boolean = false
```
adasiewiczr, line 9:
```suggestion
    val isMine: Boolean = false,
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
```diff
@@ -68,8 +67,7 @@ public fun MessageList(
         DefaultMessageContainer(
             messageItem = it,
             onThreadClick = onThreadClick,
-            onLongItemClick = onLongItemClick,
-            currentUser = viewModel.currentMessagesState.currentUser
+            onLongItemClick = onLongItemClick
```
adasiewiczr, line 70:
```suggestion
            onLongItemClick = onLongItemClick,
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
```diff
@@ -108,8 +106,7 @@ public fun MessageList(
         DefaultMessageContainer(
             messageItem = it,
             onThreadClick = onThreadClick,
-            onLongItemClick = onLongItemClick,
-            currentUser = currentState.currentUser
+            onLongItemClick = onLongItemClick
```
adasiewiczr, line 109:
```suggestion
            onLongItemClick = onLongItemClick,
```

adasiewiczr (approved):
LGTM! I just added missing trailing commas but feel free to ignore that

PR #2154, opened 2021-08-05
2122 - Auto-initializing ViewModels by default

### LINKED ISSUE #2122: Improve Component architecture/bound component documentation
A user reported they didn't realise our bound components (powered by ViewModels) require a `start()` call and they were stuck a bit on an empty screen.

We added `start()` functions to help users control over the lifecycle of our `ViewModels`. Based on their feedback, we should make sure we call the `start()` function in some of our bound components (so they don't have to) and/or that we add more guidance in the documentation, showing people they should call `start()` themselves.

Action points:
- [x] Improve our documentation for bound components
- [x] Add default `start()` calls where possible

filbabic:
Fixes #2122.

### 🎯 Goal

The idea behind this is to make the startup operations of `ViewModel`s automatic, when the object is initialized. That way, users don't have to think about calling any specific startup code, which makes it easier to set up components.

Additionally, it makes it easier for anyone who is switching from the XML SDK to Compose, as the startup behavior is the same. 

### 🛠 Implementation details

Removed or moved `start()` functions to the `init` block. Removed manual initialization in various components and examples.

### 🧪 Testing

Components should work the same, without any extra setup or manual work.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

initialize


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/list/ChannelList.kt
```diff
@@ -66,10 +65,6 @@ public fun ChannelList(
     },
 ) {
 
-    LaunchedEffect(Unit) {
```
adasiewiczr, line 69: This looks similar to React Hooks. Didn't know compose has something like this

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -56,7 +56,7 @@ import java.util.Date
 public class ImagePreviewActivity : AppCompatActivity() {
 
     private val factory by lazy {
-        ImagePreviewViewModelFactory(ChatClient.instance())
+        ImagePreviewViewModelFactory(ChatClient.instance(), intent?.getStringExtra(KEY_MESSAGE_ID) ?: "")
```
adasiewiczr, line 59: Does it make sense to run it with an empty message id? If not - maybe we should inform the user that something went wrong?

adasiewiczr (commented):
Should we mention this in the changelog?

filbabic:
@adasiewiczr Yeah I've had both PRs up w/o the changelog updates, so I'll just update the changelog in one of the PRs once I start merging everything.

It's easier than having conflicts between each PR for the same changelog location.

LMK if there's a better way to do this - maybe an intermediate branch that represents just this milestone? Then we'd have everything resolved there. :]

adasiewiczr:
We usually update the changelog in the same PR in which we do related changes. Conflicts are usually very easy to solve so we didn't think about another way but feel free to do what best for you

filbabic:
Oh it's not that they were hard to solve, it's just annoying if I create 4-5 PRs, I have to update the PRs and fix conflicts 4-5 times, which means i have to wait for all the CI checks that many times.

If we also do work on develop (e.g. merge core SDK PRs), and it happens while the checks are running, I can't merge because I have to update the PR, which again triggers all the checks.

So instead of waiting half an hour or an hour to resolve all this and merge, having one PR from an intermediate/milestone branch will just mean I have to wait once.

Does that make sense?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -69,8 +69,6 @@ public class ImagePreviewActivity : AppCompatActivity() {
             finish()
```
filbabic, line 69: Maybe we should throw or log an error that there is no `messageId` to view these images.

adasiewiczr (approved):

PR #2178, opened 2021-08-10
0000 - Updated compose accompanist

aldefy:
Fixes #0000 

### 🎯 Goal
By updating to latest Jetpack Compose accompanist library, we can also bring in more bug fixes for compose support libraries.

### 🧪 Testing
We should probably do a round of testing of the basics, just to make sure things still work with the Compose standard functions.

### ☑️ Checklist
- [x] Reviewers added
- [x] Changelog updated

### 🎉 GIF

upgrade


filbabic (changes requested):
We need to create a ticket for this and put it in the project, so we can track it.

Other than that, looks good.

PR #2180, opened 2021-08-10
[CAS-1195] Fix setting SuggestionListView's style when using it as a standalone component

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-1195

### 🎯 Goal

Fix setting SuggestionListView's style when using it as a standalone component

### 🛠 Implementation details
`SuggestionListViewStyle` resolves attributes based on `MessageInputView` attrs. That's why `SuggestionListViewStyle` was nullable and set only when we were creating it inside `MessageInputView`.
I've changed that and created a `default` style during view initialization that allows overriding different attributes either by using `suggestionListViewTheme` or `suggestionListStyleTransformer `

### 🧪 Testing

1. Create a sample that uses `SuggestionListView` as a standalone components
2. Override `suggestionListViewTheme` and check if the style is applied

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestion/list/SuggestionListViewStyle.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestion/list/SuggestionListViewStyle.kt
@@ -173,5 +173,7 @@ public data class SuggestionListViewStyle(
                 ).let(TransformStyle.suggestionListStyleTransformer::transform)
             }
         }
+
+        fun createDefault(context: Context) = invoke(context, null)
     }
 }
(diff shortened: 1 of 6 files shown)


ogkuzmin (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestion/list/SuggestionListViewStyle.kt
```diff
@@ -173,5 +173,7 @@ public data class SuggestionListViewStyle(
                 ).let(TransformStyle.suggestionListStyleTransformer::transform)
             }
         }
+
+        fun createDefault(context: Context) = invoke(context, null)
```
JcMinarro, line 177: Can't we add a default argument to our `operator fun invoke()` method?

PR #2186, opened 2021-08-10
Feature/2115 improve last message logic

### LINKED ISSUE #2115: Implement advanced last message preview logic for the channel list
The current logic shows _any_ messages's text content that happens to be the last message in the channel list.

Here's some of the desired logic in the UI Component implementation https://github.com/GetStream/stream-chat-android/blob/4734f119c29c1ab31e073f81d4a898f858166c4e/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/Channel.kt#L35

We should also make sure that deleted messages are not shown in the channel list as a preview of the channel's content.

### LINKED ISSUE #2119: Add fallbacks to Avatars (Channel list, Message list)
We need to show a fallback state when a user or a channel don't have images defined. The default behavior should be to show the initial of a user or the channel name, or a placeholder image in that case.

Action points:
* [x] Add fallback logic for `ChannelListItem`
* [x] Add fallback logic for `ChannelListHeader`
* [x] Add fallback logic for `DefaultMessageContainer`
* [ ] Add fallback logic for `MessageListHeader`

filbabic:
Fixes #2115 and fixes #2119.

### 🎯 Goal
We needed better fallback logic for our avatars and last messages, to show to our users. Now, they'll be able to preview if the last message was sent by them or someone else and if it has attachments or not and.

We also improved the way we load avatars in the SDK, for `Channel`s and `User`s. We have **initials** fallback for both `Channel`s and `User`s and we show different images, based on how many members there are in a `Channel`.

### 🛠 Implementation details
Build the `UserAvatar`, `ChannelAvatar` and `InitialsAvatar`. Added more logic for the `ChannelItem` using `Channel.getLastMessage()`.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| before | after |

### 🧪 Testing

### ☑️ Checklist

~~- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)~~
~~- [ ] Affected documentation updated (docusaurus, tutorial, CMS (task created))~~
- [x] Reviewers added

### 🎉 GIF


UNRELEASED_CHANGELOG.md
```diff
@@ -100,9 +101,14 @@
...
 
 ### ✅ Added
+- Added lots of improvements to Avatars - added a `UserAvatar`, `ChannelAvatar` and an `InitialsAvatar` to load different types of data.
+- We now show a matrix of user images in case we're in a group DM.
+- We also show initials in case the user doesn't have an image.
 
 ### ⚠️ Changed
 - `ViewModel`s now initialize automatically, so you no longer have to call `start()` on them. This is aimed to improve the consistency between our SDKs.
+- Added a `Shape` parameter to `Avatar` to customize the shape.
+- The `User` parameter in the `ChannelListHeader` is now non-nullable.
```
zsmb13, line 111: The UI Components library allows hiding the avatar from the header as a customization option (with attributes). Perhaps the Compose SDK should have a slot for the content that shows up at the start of the header?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MessageUtils.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @return - If the message type is regular.
+ * */
+public fun Message.isRegular(): Boolean = type == ModelType.message_regular
+
+/**
+ * @return - If the message type is system.
+ * */
+public fun Message.isSystem(): Boolean = type == ModelType.message_system
```
zsmb13, line 31: Could these be internal?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/avatar/UserAvatar.kt
```diff
@@ -0,0 +1,44 @@
...
+ * @param shape - The shape of the avatar.
+ * @param contentDescription - The content description of the avatar.
+ * @param onClick - The handler when the user clicks on the avatar.
+ * */
+@Composable
+public fun UserAvatar(
+    user: User,
+    modifier: Modifier = Modifier,
+    shape: Shape = ChatTheme.shapes.avatar,
+    contentDescription: String? = null,
```
zsmb13, line 28: Would it make sense to default this to `user.name`, or is that nonsense in some cases?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ImageUtils.kt
```diff
@@ -1,21 +1,40 @@
...
+        a,
+        r.coerceAtMost(255),
+        g.coerceAtMost(255),
+        b.coerceAtMost(255)
+    )
+}
 
 @Composable
-internal fun rememberChannelImagePainter(channel: Channel, currentUser: User?): Painter {
+internal fun initialsGradient(initials: String): Brush {
```
zsmb13, line 28:
```suggestion
@ReadOnlyComposable
internal fun initialsGradient(initials: String): Brush {
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/avatar/ChannelAvatar.kt
```diff
@@ -0,0 +1,118 @@
...
+            UserAvatar(
+                modifier = modifier,
+                user = user,
+                contentDescription = user.name
+            )
+        }
+        /**
+         * If the channel has more than two members - group - we load a matrix of their images or initials.
+         * */
+        else -> {
```
zsmb13, line 75: Does the logic here work nicely for 3 avatars? 🤔

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/avatar/InitialsAvatar.kt
```diff
@@ -0,0 +1,42 @@
...
+    Box(
+        modifier = modifier
+            .clip(shape)
+            .background(brush = initialsGradient)
+    ) {
+        Text(
+            modifier = Modifier.align(Alignment.Center),
+            text = initials,
+            style = ChatTheme.typography.title1,
+            color = ChatTheme.colors.textHighEmphasis
```
zsmb13, line 39: Something minor I noticed here: the UI Components use hardcoded white text for this - we might want to make these consistent one way or the other

UNRELEASED_CHANGELOG.md
filbabic: We could expose a property for the start of the header. I think we can then move back to the nullable user and just show a spacer in case there's nothing (to make sure things are aligned)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MessageUtils.kt
filbabic: Yeah I think so, good catch

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ImageUtils.kt
filbabic: Good catch!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/avatar/UserAvatar.kt
filbabic: I think it's better to be null, as we shouldn't fill the accessibility data with names of users - screen readers should be able to catch the text elements anyway.

When it's used in channel lists, when we have 1:1 DMs, people will likely use SRs to read the "channel name" label anyway. This would be a description on an image that doesn't give much info.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/avatar/ChannelAvatar.kt
filbabic: As per the before/after images, it should work the same as for attachments - we show 2 avatars on the left side, split screen, and the third avatar on the right (full height).

zsmb13 (approved):

PR #2199, opened 2021-08-11
[CAS-1196] Add custom linkifying implementation to properly format links in message [Old Client]

amitkma:
https://stream-io.atlassian.net/browse/CAS-1196

### 🎯 Goal

It seems that formatting doesn’t work well - links are not highlighted nor clickable. It is happening because we have autolink property set to TextView.

### 🛠 Implementation details

TextView autolink mask removes all the URLSpan when it tries to link the text. So in order to support both (markdown links and normal text links), a custom Linkify implementation is added. This implementation doesn't remove already added links (via markdown plugins). 

### 🎨 UI Changes

| Before | After |
| --- | --- |
|  |  |

### 🧪 Testing

This can be tested very easily. You can send a message having text link, markdown link or both. For example ->
```
Hey
[google](https://www.google.com) 
https://www.google.com
```
Both link should be clickable. 

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android/src/main/java/com/getstream/sdk/chat/utils/Linkify.kt
```diff
@@ -0,0 +1,212 @@
...
+import java.util.regex.Matcher
+import java.util.regex.Pattern
+
+/**
+ * Linkify links the part of the text based on matching pattern.
+ *
+ * This class is a simplified version of [Linkify] and differs only in one following way
+ * It doesn't remove any existing URLSpan from the Spannable.
+ */
+internal object Linkify {
```
adasiewiczr, line 23: What do you think about moving it to `ui-common` module? We could mark it with @InternalStreamApi annotation and reuse it in both `ui` modules

amitkma: You mentioned that "we are going to remove ui-common sometime". Is it still a good idea to do that?

adasiewiczr: We don't know when we are going to do that and it should be fine because this object shouldn't be used outside our SDK

amitkma:
This PR is being closed because it is implemented by #2203

PR #2200, opened 2021-08-11
[CAS-1185] Refac - Retry connection

leandroBorgesFerreira:
[CAS-1185](https://stream-io.atlassian.net/browse/CAS-1185)

### 🎯 Goal

Create a retry logic when the connection fails in a nonpermanent way.

### 🛠 Implementation details

When there's no internet and the socket executes an error handling, it retries 3 times to reconnect with exponential backoff. 

### 🧪 Testing

Enable and disable internet while using the app

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -122,7 +127,13 @@ internal class ChatSocketServiceImpl constructor(
             ChatErrorCode.UNABLE_TO_PARSE_SOCKET_EVENT.code,
             ChatErrorCode.NO_ERROR_BODY.code,
             -> {
-                // Nothing to do on that case
+                if (!networkStateProvider.isConnected() && reconnectionAttempts < 3) {
```
ogkuzmin, line 130: It's better not to expand the complexity of this class by this logic. We already have `HealthMonitor` with similar logic, we can modify it to cover our case.

ogkuzmin: Move this magic number to constants

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -72,6 +76,7 @@ internal class ChatSocketServiceImpl constructor(
                     }
                     is State.Connected -> {
                         healthMonitor.start()
+                        reconnectionAttempts = 0
```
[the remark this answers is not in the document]
leandroBorgesFerreira: 🤔  Do you mean logout from the user? This prevents retries stop happening if a user:
1 - Retries 3 times.
2 - Successfully connects
3 - Loses connection again. 

If we only reset in a logout, then this case is not covered.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
leandroBorgesFerreira: `by this logic` do you mean `!networkStateProvider.isConnected()`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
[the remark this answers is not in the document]
ogkuzmin: We don't need to restore retries count when connected. We need this only for first connection
I agree we can leave the case with logout

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
ogkuzmin: Also retries condition

leandroBorgesFerreira: This way `HealthMonitor` will start to be too much specific and handle logic looks outside of its scope.

ogkuzmin (approved):
LGTM for me now!

PR #2203, opened 2021-08-11
[CAS-1196] Add custom linkifying implementation to properly format links in messages

amitkma:
https://stream-io.atlassian.net/browse/CAS-1196

### 🎯 Goal

It seems that formatting doesn’t work well - links are not highlighted nor clickable. It is happening because we have autolink property set to TextView.

### 🛠 Implementation details

TextView autolink mask removes all the URLSpan when it tries to link the text. So in order to support both (markdown links and normal text links), a custom Linkify implementation is added. This implementation doesn't remove already added links (via markdown plugins). 

### 🎨 UI Changes

| Before | After |
| --- | --- |
|  |  |

### 🧪 Testing

This can be tested very easily. You can send a message having text link, markdown link or both. For example ->
```
Hey
[google](https://www.google.com) 
https://www.google.com
```
Both link should be clickable. 

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/internal/Linkify.kt
```diff
@@ -0,0 +1,212 @@
...
+ * Linkify links the part of the text based on matching pattern.
+ *
+ * This class is a simplified version of [Linkify] and differs only in one following way
+ * It doesn't remove any existing URLSpan from the Spannable.
+ */
+internal object Linkify {
+
+    private val COMPARATOR: Comparator<LinkSpec> = object : Comparator<LinkSpec> {
+        override fun compare(a: LinkSpec, b: LinkSpec): Int {
+            if (a.start < b.start) {
```
leandroBorgesFerreira, line 27: You can use `when` instead of `if/else`: 
```
return when {
                a.start < b.start -> -1
                a.start > b.start, a.end < b.end -> 1
                else -> 0
            }
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/internal/Linkify.kt
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Scans the provided spannable text and turns all occurrences
+     * of the link types into clickable links (Currently only support web urls).
+     *
+     * @param text Spannable whose text is to be marked-up with links.
+     * @return True if at least one link is found and applied.
+     */
+    @SuppressLint("RestrictedApi")
+    private fun addLinks(text: Spannable): Boolean {
+        val links: ArrayList<LinkSpec> = ArrayList<LinkSpec>()
```
leandroBorgesFerreira, line 78: Normally we use `listOf<T>()` and interfaces instead of concrete implementations (List instead of ArrayList).

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/internal/Linkify.kt
```diff
@@ -0,0 +1,212 @@
...
+        gatherLinks(
+            links, text, PatternsCompat.AUTOLINK_WEB_URL, arrayOf("http://", "https://", "rtsp://"),
+            Linkify.sUrlMatchFilter, null
+        )
+
+        pruneOverlaps(links, text)
+
+        if (links.isEmpty()) return false
+
+        for (link in links) {
```
leandroBorgesFerreira, line 88: You can use `forEach` and a `filter` instead of loops. Overall I think it is easier to read when the stream of operations becomes big.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/internal/Linkify.kt
```diff
@@ -0,0 +1,212 @@
...
+        val urlSpans: Array<URLSpan> = text.getSpans(0, text.length, URLSpan::class.java)
+        for (i in urlSpans.indices) {
+            val spec = LinkSpec(
+                markwonAddedSpan = urlSpans[i],
+                start = text.getSpanStart(urlSpans[i]),
+                end = text.getSpanEnd(urlSpans[i])
+            )
+            links.add(spec)
+        }
+        Collections.sort(links, COMPARATOR)
```
leandroBorgesFerreira, line 175: You can also use `links.sortWith(COMPARATOR)` for a more Kotlin way to do it.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/internal/Linkify.kt
```diff
@@ -0,0 +1,212 @@
...
+                val spec = LinkSpec(url = url, start = start, end = end)
+                links.add(spec)
+            }
+        }
+    }
+
+    private fun pruneOverlaps(links: ArrayList<LinkSpec>, text: Spannable) {
+        // Append spans added by Markwon to remove any overlap.
+        val urlSpans: Array<URLSpan> = text.getSpans(0, text.length, URLSpan::class.java)
+        for (i in urlSpans.indices) {
```
leandroBorgesFerreira, line 167: You can use e `map` in this case.

amitkma: I think `forEach` is better than map here.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/internal/Linkify.kt
amitkma: Can filtering and other operations be done in one iteration?

leandroBorgesFerreira: I personally prefer to do it separately, but fell free to use the one you prefer.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/Linkify.kt
```diff
@@ -0,0 +1,214 @@
...
+                break
+            }
+        }
+        if (!hasPrefix && prefixes.isNotEmpty()) {
+            url = prefixes[0] + url
+        }
+        return url
+    }
+
+    private fun gatherLinks(
```
leandroBorgesFerreira, line 140: Avoid to create functions that have `Unit` as the return type and apply a side effect to a parameter (`links` in tis case) and prefer to return the result of function. This makes it more functional and easier to reason about. 

You can just return the a `List` in this case.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/Linkify.kt
```diff
@@ -0,0 +1,214 @@
...
+        var len = links.size
+        var i = 0
+        while (i < len - 1) {
+            val a: LinkSpec = links[i]
+            val b: LinkSpec = links[i + 1]
+            var remove = -1
+            if (a.start <= b.start && a.end > b.start) {
+                when {
+                    b.end <= a.end -> {
+                        remove = i + 1
```
leandroBorgesFerreira, line 183: You can lift the attribution of `remove`. remove = when { [...] }

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/Linkify.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private data class LinkSpec(
+        val markwonAddedSpan: URLSpan? = null,
+        val url: String? = null,
+        val start: Int,
+        val end: Int,
+    )
+
+    private val DEFAULT_SPAN_FACTORY: (string: String?) -> URLSpan = { string: String? -> URLSpan(string) }
```
leandroBorgesFerreira, line 213: No need to wrap the constructor inside a function
```suggestion
    private val DEFAULT_SPAN_FACTORY: (string: String?) -> URLSpan = ::URLSpan
```

leandroBorgesFerreira (approved):
Some observations, but not mandatory, apply if as you prefer.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/Linkify.kt
amitkma: Then I will have to add `else` branch to it and not sure if it would do any benefit  (or make any logical sense to even have it) here.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/Linkify.kt
amitkma: It should be considered more as effect than side effect. The `gatherLinks` responsibility is to search the link in the string and add it to the list.  It is the only effect of this method.

UNRELEASED_CHANGELOG.md
```diff
@@ -75,6 +75,7 @@
 - Fixed attachments of camera. Now multiple videos and pictures can be taken from the camera.
 - Added the possibility to force light and dark theme. Set it in inside ChatUI to make all views, fragments and activity of the SDK light.
 - Fixed applying style to `SuggestionListView` when using it as a standalone component. You can modify the style using `suggestionListViewTheme` or `TransformStyle::suggestionListStyleTransformer`
+- Fixed markdown links rendering using custom linkify implementation.
```
adasiewiczr, line 78: Let's mention this fix in old ui components module

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_reply_view.xml
```diff
@@ -61,7 +61,6 @@
             android:id="@+id/replyText"
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:autoLink="web"
```
adasiewiczr, line 64: We also need to remove this from `TextView` used in old component module, right?

adasiewiczr (changes requested):

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_reply_view.xml
amitkma: Done.

UNRELEASED_CHANGELOG.md
amitkma: Done

adasiewiczr (approved):

PR #2207, opened 2021-08-12
[CAS-1202] Remove Gson from DB converters

zsmb13:
https://stream-io.atlassian.net/browse/CAS-1202

### 🎯 Goal

To be able to remove the Gson dependency for our clients once it's no longer needed in the client module either

### 🛠 Implementation details

Replaced Gson usages in DB converters with Moshi adapters

### 🧪 Testing

Run unit tests, smoke test the sample app

### ☑️ Checklist

- [ ] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/DateAdapter.kt
```diff
@@ -4,16 +4,18 @@ import com.squareup.moshi.JsonAdapter
...
 import com.squareup.moshi.JsonWriter
 import io.getstream.chat.android.client.utils.threadLocal
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
 import java.util.TimeZone
 
-internal class DateAdapter : JsonAdapter<Date>() {
+@InternalStreamChatApi
```
zsmb13, line 13: Exposed to be reused in the offline module

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/extensions/ChannelExtensionsTest.kt
```diff
@@ -11,13 +10,16 @@ import org.junit.jupiter.api.Assertions
...
+import kotlin.reflect.typeOf
 
 internal class ChannelExtensionsTest {
 
+    @OptIn(ExperimentalStdlibApi::class)
     @Test
     fun `When apply pagination Should not throw any exception`() {
         val channelsFile = File(this.javaClass.classLoader!!.getResource("channels.json").toURI())
-        val type = object : TypeToken<List<Channel>>() {}.type
+        val type = typeOf<List<Channel>>().javaType
```
zsmb13, line 22: Tiny improvement to use fewer Gson APIs in this test - will need to be revisited later to see how we can parse this JSON with Moshi while being in the offline module

JcMinarro (commented):
Looks right.
Only one doubt, are the json data (previously stored with Gson Parser) compatible with Moshi? If not, we will need to increase the DB version

zsmb13:
@JcMinarro In theory, it should be compatible (tests are passing). Might not be a bump the DB version just in case though. Should we do that?

buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
```diff
@@ -175,6 +175,7 @@ object Dependencies {
     const val mockitoKotlin = "com.nhaarman.mockitokotlin2:mockito-kotlin:${Versions.MOCKITO_KOTLIN}"
     const val moshi = "com.squareup.moshi:moshi:${Versions.MOSHI}"
     const val moshiCodegen = "com.squareup.moshi:moshi-kotlin-codegen:${Versions.MOSHI}"
+    const val moshiKotlin =  "com.squareup.moshi:moshi-kotlin:${Versions.MOSHI}"
```
ogkuzmin, line 178: I didn't want to mark our models with `@JsonClass(generateAdapter = true)` that's why I need an adapter for Kotlin

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -43,14 +43,17 @@ public data class Attachment(
 
     @IgnoreSerialisation
     @IgnoreDeserialisation
+    @Transient
```
ogkuzmin, line 46: To make this class serializable by Moshi we need to mark as non-serializable some fields

ogkuzmin (approved):

ogkuzmin:
@zsmb13 @JcMinarro  Nice to clean DB, let's bump DB version

ogkuzmin:
> @JcMinarro In theory, it should be compatible (tests are passing). Might not be a bump the DB version just in case though. Should we do that?

In order to be sure that you didn't destroy anything we need integration test suite (very basic) with some stored in DB cache

JcMinarro:
> @JcMinarro In theory, it should be compatible (tests are passing). Might not be a bump the DB version just in case though. Should we do that?

I don't think our tests check the previous "schema" stored on the DB.
As far as you are parsing with Moshi on both ways, we can't know if it keeps working with previous data.
At the end, because we only use DB "as run-time cache", it is safer to bump DB version everytime we change something on our DB Implementation

JcMinarro (approved):

PR #2228, opened 2021-08-16
[CAS-1198] Feature - Show avatar customization

leandroBorgesFerreira:
[CAS-1198](https://stream-io.atlassian.net/browse/CAS-1198)

### 🎯 Goal

Customize avatar visibility

### 🛠 Implementation details

Added function to show avatar based on user defined logic. 

Also, there's an avatar for mine messages too. The margins of the messages respects the new avatar added. 

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_20210816-121638 | Screenshot_20210816-121535 |

### 🧪 Testing

Use `setShowAvatarPredicate` to create different way to show the avatar in our `MessageListView`.

### ☑️ Checklist

- [ ] Affected documentation updated (docusaurus, tutorial, CMS (task created)). todo
- [ ] Reviewers added

### 🎉 GIF


stream-chat-android-ui-components/src/main/res/values/styles.xml
```diff
@@ -478,6 +478,7 @@
         <item name="android:layout_height">@dimen/stream_ui_avatar_size_small</item>
         <item name="android:layout_marginTop">8dp</item>
         <item name="android:layout_marginEnd">8dp</item>
+        <item name="android:layout_marginStart">8dp</item>
```
zsmb13, line 481: Not sure if this PR is ready to review (I was assuming so, as it's not a draft), but margins with the current approach are messed up in a lot of places.

image

I think the arbitrary predicate API in its current design is just not able to work correctly. 

The problem is that we have to know not only whether a message needs to show the avatar, but also whether it needs to show the margin to align with other messages in its message group that _do_ show an avatar.

Consider these two cases for messages that don't need to have avatars themselves:
- If the default predicate is used (avatar only for bottom position messages), all messages in non-bottom positions should still show the margin for the avatar.
- If we use a predicate that returns `false` in all cases, no messages should show the margin for the avatar, and they should just be lined up to the edge of the screen instead.

If the predicate can have arbitrary logic in it, you simply can't tell if a message needs the margin or not (consider `Random.nextBoolean()` for an extreme case).

We either have to also ask users if they want the margins to be present on the message (if they don't want the avatar), or have to restrict customizations options to hiding ours/theirs and showing for all / just bottom messages, or come up with some other approach that I didn't think of off the top of my head.

zsmb13 (changes requested):
See last comment for details - the rest were mostly just small observations about the code

stream-chat-android-ui-components/src/main/res/values/styles.xml
leandroBorgesFerreira: You're right. I did my tests with true/false and forgot about that 😵‍💫. Let's make the margins customisable so the users can decide how to apply then.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
```diff
@@ -29,11 +32,30 @@ internal class AvatarDecorator : BaseDecorator() {
...
+                avatarView.setUserData(data.message.user)
+                avatarView.isVisible = true
+            }
+        } else if (data.isTheirs && data.isTheirs && data.isBottomPosition()) {
             avatarView.setUserData(data.message.user)
+            avatarView.isVisible = true
+        }
+    }
+
+    private fun getAvatarView(holder: TextAndAttachmentsViewHolder, data: MessageListItem.MessageItem): AvatarView {
```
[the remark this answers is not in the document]
leandroBorgesFerreira: the `Position` of the `MessageItem` is also important. It is hard to tell which criteria the users will use before hand, we prefer to pass the whole message so they can decide the logic that they prefer.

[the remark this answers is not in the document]
zsmb13: This specific method is private and only uses `isMine` in its implementation.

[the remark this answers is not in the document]
leandroBorgesFerreira: I misunderstood what you meant.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
```diff
@@ -29,11 +32,30 @@ internal class AvatarDecorator : BaseDecorator() {
...
+        } else {
+            holder.binding.avatarView
+        }
+    }
+
+    private fun getAvatarView(holder: MessagePlainTextViewHolder, data: MessageListItem.MessageItem): AvatarView {
+        return if (data.isMine) {
+            holder.binding.avatarMineView
         } else {
-            avatarView.isVisible = false
```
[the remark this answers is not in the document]
leandroBorgesFerreira: Now I'm setting isVisible = false

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
```diff
@@ -4,23 +4,26 @@ import androidx.core.view.isVisible
 import com.getstream.sdk.chat.adapter.MessageListItem
 import com.getstream.sdk.chat.utils.extensions.isBottomPosition
 import io.getstream.chat.android.ui.avatar.AvatarView
+import io.getstream.chat.android.ui.message.list.MessageListView
 import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.GiphyViewHolder
 import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.MessagePlainTextViewHolder
 import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.TextAndAttachmentsViewHolder
 
-internal class AvatarDecorator : BaseDecorator() {
+internal class AvatarDecorator(
```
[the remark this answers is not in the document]
leandroBorgesFerreira: It should.

stream-chat-android-ui-components/src/main/res/values/styles.xml
leandroBorgesFerreira: Now it is possible to configure the margin of the start and the end of `MessageListView`. The draw back is that the user need to understand that he/she needs to create space for the avatar, but the good side is that it is now fully customisable.

leandroBorgesFerreira: I'll add the documentation for it tomorrow.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/MaxPossibleWidthDecorator.kt
```diff
@@ -40,7 +42,7 @@ internal class MaxPossibleWidthDecorator : BaseDecorator() {
     }
 
     companion object {
-        private const val MAX_POSSIBLE_WIDTH_FACTOR = .30f
+        private const val MAX_POSSIBLE_WIDTH_FACTOR = .25f
```
leandroBorgesFerreira, line 45: I increased it in 5% to accommodate the footnote when sending the attachment

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
```diff
@@ -600,6 +602,7 @@ public class MessageListView : ConstraintLayout {
             dateFormatter = messageDateFormatter,
             isDirectMessage = { channel.isDirectMessaging() },
             messageListViewStyle = requireStyle(),
+            showAvatarPredicate = this.showAvatarPredicate
```
zsmb13, line 605:
```suggestion
            showAvatarPredicate = this.showAvatarPredicate,
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/internal/MessageListItemDecoratorProvider.kt
```diff
@@ -19,14 +20,15 @@ internal class MessageListItemDecoratorProvider(
     dateFormatter: DateFormatter,
     isDirectMessage: () -> Boolean,
     messageListViewStyle: MessageListViewStyle,
+    showAvatarPredicate: MessageListView.ShowAvatarPredicate
```
zsmb13, line 23:
```suggestion
    showAvatarPredicate: MessageListView.ShowAvatarPredicate,
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
```diff
@@ -2,38 +2,77 @@ package io.getstream.chat.android.ui.message.list.adapter.viewholder.decorator.i
...
+
+    private fun getAvatarView(holder: MessagePlainTextViewHolder, isMine: Boolean): AvatarView {
+        return if (isMine) {
+            holder.binding.avatarView.isVisible = false
+            holder.binding.avatarMineView
+        } else {
+            holder.binding.avatarMineView.isVisible = false
+            holder.binding.avatarView
+        }
+    }
```
zsmb13, line 67: This might be worth de-duplicating, as we're doing the same thing three times here

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
```diff
@@ -2,38 +2,77 @@ package io.getstream.chat.android.ui.message.list.adapter.viewholder.decorator.i
...
+
+    private fun getAvatarView(holder: MessageDeletedViewHolder, isMine: Boolean): AvatarView {
+        return if (isMine) {
+            holder.binding.avatarView.isVisible = false
```
zsmb13, line 71: Something that just occurred to me (a bit late, I know): would it make our lives easier or harder if we used the avatars themselves to provide the margin, by just making them `INVISIBLE` instead of `GONE` when they're not needed? In that case, we'd need attributes to toggle all avatars on/off on each side of the message list, I suppose.

stream-chat-android-ui-components/src/main/res/values/attrs_message_list_view.xml
```diff
@@ -397,5 +397,8 @@
         <attr name="streamUiPinnedMessageIndicatorIcon" format="reference" />
 
         <attr name="streamUiPinnedMessageBackgroundColor" format="color" />
+
+        <attr name="streamUiMessageStartMargin" format="dimension|reference" />
+        <attr name="streamUiMessageEndMargin" format="dimension|reference" />
```
zsmb13, line 402: Out of curiosity, what does the message list look like when these values are:
- 0, but the predicate says avatars should be shown
- Half the default value, and predicate says avatars should show
- More than the default value

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/MessagePlainTextViewHolder.kt
```diff
@@ -52,6 +56,18 @@ internal class MessagePlainTextViewHolder(
...
+    private fun StreamUiItemMessagePlainTextBinding.configMargins(style: MessageListItemStyle) {
+        messageContainer.updateLayoutParams<ViewGroup.MarginLayoutParams> {
+            marginStart = style.messageStartMargin
+            marginEnd = style.messageEndMargin
+        }
+
+        footnote.updateLayoutParams<ViewGroup.MarginLayoutParams> {
+            marginStart = style.messageStartMargin
+            marginEnd = style.messageEndMargin
+        }
```
zsmb13, line 68: Should this be performed in a decorator perhaps?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
leandroBorgesFerreira: The binding views are not the same, so I don't think it is possible to make a method for all of then

zsmb13: I'm thinking something like `getAvatarView(myView: AvatarView, theirView: AvatarView, isMine: Boolean)`, we have similar things in some of our decorators

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
leandroBorgesFerreira: I thought about it. If we use a flag, we could just set a hard coded margin so the user wouldn't be able to provide a "wrong" margin... But then we would be taking away a big degree of customization, which I consider to be worse.

About the Invisible/Gone. The logic would be complex... For 0 avatars visible we would use Invisible, but for more than one, we would make Gone... We would need to keep track of the list. Set a custom margin is easier in the case

zsmb13: Let's check to be sure that something that's not 0 or the default value makes sense (as per my next comment), otherwise we can go with the custom margin approach

stream-chat-android-ui-components/src/main/res/values/attrs_message_list_view.xml
leandroBorgesFerreira: The idea is not to prevent the user from creating a margin that doesn't look good, but allowing him/her to customize it the way they prefer.

In the first case they would need to set a margin in the bottom to avoid text and avatar overlap. Same goes for the second. The third works fine, but they may end up with too little space in the screen for actual text

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/MessagePlainTextViewHolder.kt
leandroBorgesFerreira: I'll create a new one then. "MessageCo trainerMargin"

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
leandroBorgesFerreira: The avatar is not the one creating the margin anymore, so different values than default or 0 will be needed if the size of avatar changes (if the user want to keep the same margins from the avatar view)

UNRELEASED_CHANGELOG.md
```diff
@@ -71,6 +71,10 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+Now it is possible to customize when the avatar appears in the conversation. It is
+possible to use an avatar in messages fro other users and for message of the current user. Use
+`setShowAvatarPredicate` to change the default behaviour. Default: The avatar shows only
+for messages of other users, in the bottom of a group of messages (Position.Bottom).
```
zsmb13, line 77: We should update this to better describe the new APIs given the changes we've made in the PR since its first version (+ check for typos)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/MessageContainerMarginDecorator.kt
```diff
@@ -0,0 +1,43 @@
...
+import android.view.ViewGroup
+import androidx.core.view.updateLayoutParams
+import com.getstream.sdk.chat.adapter.MessageListItem
+import io.getstream.chat.android.ui.message.list.MessageListItemStyle
+import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.GiphyViewHolder
+import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.MessageDeletedViewHolder
+import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.MessagePlainTextViewHolder
+import io.getstream.chat.android.ui.message.list.adapter.viewholder.internal.TextAndAttachmentsViewHolder
+
+internal class MessageContainerMarginDecorator(private val style: MessageListItemStyle) : BaseDecorator() {
```
zsmb13, line 13: Nice abstraction!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/DefaultShowAvatarPredicate.kt
```diff
@@ -0,0 +1,9 @@
+package io.getstream.chat.android.ui.message.list
+
+import com.getstream.sdk.chat.adapter.MessageListItem
+
+public class DefaultShowAvatarPredicate : MessageListView.ShowAvatarPredicate {
```
zsmb13, line 5: Do we have to expose this as API?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/MessageDeletedViewHolder.kt
```diff
@@ -24,5 +26,9 @@ internal class MessageDeletedViewHolder(
         super.bindData(data, diff)
 
         style.textStyleMessageDeleted.apply(binding.deleteLabel)
+
+        binding.messageContainer.updateLayoutParams<ConstraintLayout.LayoutParams> {
+            horizontalBias = if (data.isTheirs) 0f else 1f
+        }
```
zsmb13, line 32: I think we need to do this for the footnote view as well, seeing things like this when running the app

image

docusaurus/docs/Android/03-ui/04-components/03-message-list.md
```diff
@@ -432,3 +432,16 @@ messageListView.setEmptyStateView(
...
+```
+messageListView.setShowAvatarPredicate(
+    ShowAvatarPredicate { messageItem ->
+        messageItem.positions.contains(MessageListItem.Position.BOTTOM) && messageItem.isTheirs
+    }
+)
+```
+:::note
+To avoid overlap between the avatar and the messages of the chat, remember to use `streamUiMessageStartMargin` and `streamUiMessageEndMargin` to create space for the avatar of the messages.
+:::
```
zsmb13, line 447: We should show a concrete example here of how these attributes can be used. For example, based on these docs, I have no idea what value to set the margin to if I start showing avatars for my own messages so that they show up correctly.

It'd be nice to have something like

```
If you set a predicate that shows avatars for your own messages as well, use this value:

streamUiMessageEndMargin="???"

If your predicate doesn't show avatars for your own messages (this is the default behavior), remove the end margin:

streamUiMessageEndMargin="0"
```

docusaurus/docs/Android/03-ui/04-components/03-message-list.md
```diff
@@ -432,3 +432,16 @@ messageListView.setEmptyStateView(
 This code will display the following empty state:
 
 ![The custom empty state in action](../../assets/message_lis_custom_empty_state.png)
+## Configure When Avatar Appears
+Is it possible to configure when the avatar for messages appears. You can use `MessageListView.setShowAvatarPredicate` and pass a predicate to define when the avatar is going to be shown. The example implements the default behaviour:
+
+```
```
zsmb13, line 438:
```suggestion
```kotlin
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/MessageDeletedViewHolder.kt
leandroBorgesFerreira: o.O

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/DefaultShowAvatarPredicate.kt
leandroBorgesFerreira: Not at all, good catch.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/MessageContainerMarginDecorator.kt
leandroBorgesFerreira: Thanks!

UNRELEASED_CHANGELOG.md
```diff
@@ -67,10 +67,29 @@
 
 ## stream-chat-android-ui-components
 ### 🐞 Fixed
-
+Fix for position of deleted messages for other users
 ### ⬆️ Improved
 
 ### ✅ Added
+Now it is possible to customize when the avatar appears in the conversation. It is
```
zsmb13, line 74: One last observation, leaving this comment on a random line so that it can be used for a discussion.

In message options, hiding the avatar makes it so that the option items no longer line up with the message bubble:

| Shown | Hidden |
| --- | --- |
| image  | image |

The design for our/their message options shows the correct avatar visible/invisible handling nicely (as one shows, and one hides the avatar):

image

The images above show the problem on _their_ side, but I assume it'd also be misaligned if we showed avatars for our own messages, and then opened message options.

---

Bonus: what's that gap between the message and the options below it? Blank space for the footnote? We should clean that up, but as it exists on `develop` too, it can go into another issue / PR.

leandroBorgesFerreira: Good eye. Let me take a look.

leandroBorgesFerreira: Done. The margin of the options follows the margin for the message

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/options/message/internal/MessageOptionsDialogFragment.kt
```diff
@@ -214,9 +214,21 @@ internal class MessageOptionsDialogFragment : FullScreenDialogFragment() {
...
+                } else {
+                    marginStart = style.itemStyle.messageStartMargin + optionsOffset()
+                }
+            }
         }
     }
 
+    private fun optionsOffset(): Int {
+        return requireContext().getDimension(R.dimen.stream_ui_spacing_medium)
+    }
```
zsmb13, line 230: Looks like this could be a local variable

UNRELEASED_CHANGELOG.md
```diff
@@ -73,10 +73,29 @@
...
+```
+streamUiMessageEndMargin=">@dimen/stream_ui_message_viewholder_avatar_missing_margin"
+```
+
+If your predicate doesn't show avatars for your own messages (this is the default behavior), remove the end margin:
+
+```
+streamUiMessageEndMargin="0dp"
+```
+
```
zsmb13, line 98: - We use bullet points for items in the changelog
  - In this case, we could probably just link to the newly added docs for details instead of duplicating content

docusaurus/docs/Android/03-ui/04-components/03-message-list.mdx
```diff
@@ -434,3 +434,28 @@ messageListView.setEmptyStateView(
...
+    }
+)
+```
+:::note
+To avoid overlap between the avatar and the messages of the chat, remember to use `streamUiMessageStartMargin` and `streamUiMessageEndMargin` to create space for the avatar of the messages.
+:::
+
+If you set a predicate that shows avatars for your own messages as well, use this value:
+
+```
```
zsmb13, line 453:
```suggestion
```xml
```

docusaurus/docs/Android/03-ui/04-components/03-message-list.mdx
```diff
@@ -434,3 +434,28 @@ messageListView.setEmptyStateView(
...
+streamUiMessageEndMargin=">@dimen/stream_ui_message_viewholder_avatar_missing_margin"
+```
+
+If your predicate doesn't show avatars for your own messages (this is the default behavior), remove the end margin:
+
+```
```
zsmb13, line 459:
```suggestion
```xml
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/options/message/internal/MessageOptionsDialogFragment.kt
```diff
@@ -214,9 +214,21 @@ internal class MessageOptionsDialogFragment : FullScreenDialogFragment() {
                     isMessagePinned = message.pinned
                 )
             }
+
+            updateLayoutParams<ViewGroup.MarginLayoutParams> {
+                if (messageItem.isMine) {
+                    marginEnd = style.itemStyle.messageEndMargin + optionsOffset()
```
zsmb13, line 220: If I add end margin on my items and set a predicate that displays avatars on my own messages, it'll now add a margin onto the message options as well, but not display my avatar there:

image

leandroBorgesFerreira: We add `Position.Bottom` to the `MessageItem` when we create it for this dialog... that's why the avatar doesn't appear in this case. 

Since the change the `MessageItem` of the dialog, I don't think we can include  messages in this dialog it in the Predicate. The way I see to deal with it is to create a separate configuration to control the avatar show in this specific case...

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
```diff
@@ -2,38 +2,57 @@ package io.getstream.chat.android.ui.message.list.adapter.viewholder.decorator.i
...
+        val shouldShow = showAvatarPredicate.shouldShow(data)
+
+        avatarView.isVisible = shouldShow
+
+        if (shouldShow) {
             avatarView.setUserData(data.message.user)
+        }
+    }
+
+    private fun getAvatarView(myAvatar: AvatarView, theirAvatar: AvatarView, isMine: Boolean): AvatarView {
```
ogkuzmin, line 49: IMO, It's not a clean method. You return a view, but also hide another. Shouldn't hiding be a part of `setupAvatar` method?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/DefaultShowAvatarPredicate.kt
```diff
@@ -0,0 +1,9 @@
+package io.getstream.chat.android.ui.message.list
+
+import com.getstream.sdk.chat.adapter.MessageListItem
+
+internal class DefaultShowAvatarPredicate : MessageListView.ShowAvatarPredicate {
```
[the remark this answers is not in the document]
leandroBorgesFerreira: But why are objects better than classes?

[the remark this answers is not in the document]
ogkuzmin: It doesn't take any significant place. It's like lambda not anymore (I bet lambda takes the same amount of memory). It's easy to set a singleton object than instantiate class instance and set. But it's more a matter of taste. It's up to you

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/AvatarDecorator.kt
leandroBorgesFerreira: Makes sense, there's a side effect in `getAvatarView`. Let me remove it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/DefaultShowAvatarPredicate.kt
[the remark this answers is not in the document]
leandroBorgesFerreira: In this case, I prefer to use a class because I like that is was just created.

PR #2233, opened 2021-08-17
2223 - Added SDK-specific version header

### LINKED ISSUE #2223: [Compose] - Set up Compose tracking headers for requests.
We need to track how many people are using our Compose SDK. To do that, we need to add a compose header to our requests.

filbabic:
Closes #2223.

### 🎯 Goal

We need a way to track if our customers use the Compose SDK or the core SDK to keep tabs on how many people use which SDK.

### 🛠 Implementation details

Exposed the `VERSION_PREFIX` property as public, but `@InternalStreamChatApi`. Also added an enum that describes which version headers we support.

Internally, we don't allow for header changes to values that we don't support - using `Delegates.vetoable()`.

### 🧪 Testing

We can test it by tracking out headers and using the Compose SDK. We should have the `X-Stream-Client` show up as `stream-chat-compose-4.16.0`

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
header

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -37,6 +37,7 @@ import io.getstream.chat.android.client.events.ConnectedEvent
 import io.getstream.chat.android.client.events.DisconnectedEvent
 import io.getstream.chat.android.client.extensions.ATTACHMENT_TYPE_FILE
 import io.getstream.chat.android.client.extensions.ATTACHMENT_TYPE_IMAGE
+import io.getstream.chat.android.client.header.VersionPrefixHeader
 import io.getstream.chat.android.client.helpers.QueryChannelsPostponeHelper
 import io.getstream.chat.android.client.logger.ChatLogLevel
 import io.getstream.chat.android.client.logger.ChatLogger
@@ -84,7 +85,7 @@ import java.nio.charset.StandardCharsets
 import java.util.Calendar
 import java.util.Date
 import java.util.concurrent.Executor
-import kotlin.jvm.Throws
+import kotlin.properties.Delegates
 
 /**
  * The ChatClient is the main entry point for all low-level operations on chat
@@ -1082,6 +1083,7 @@ public class ChatClient internal constructor(
         return api.sendEvent(eventType, channelType, channelId, extraData)
     }
 
+    @InternalStreamChatApi
     public fun getVersion(): String = VERSION_PREFIX + BuildConfig.STREAM_CHAT_VERSION
 
     @CheckResult
@@ -1576,7 +1578,12 @@ public class ChatClient internal constructor(
     }
 
     public companion object {
-        private const val VERSION_PREFIX = "stream-chat-android-"
+        @InternalStreamChatApi
+        @JvmStatic
+        public var VERSION_PREFIX: String by Delegates.vetoable(VersionPrefixHeader.CORE_ANDROID.prefix) { _, _, new ->
+            VersionPrefixHeader.values().any { it.prefix == new }
+        }
+
         private var instance: ChatClient? = null
 
         @JvmField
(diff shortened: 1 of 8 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1082,6 +1083,7 @@ public class ChatClient internal constructor(
         return api.sendEvent(eventType, channelType, channelId, extraData)
     }
 
+    @InternalStreamChatApi
     public fun getVersion(): String = VERSION_PREFIX + BuildConfig.STREAM_CHAT_VERSION
```
adasiewiczr, line 1087: Maybe we can make it `internal`. @JcMinarro what do you think?

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChannelActivity.kt
```diff
@@ -67,6 +67,7 @@ class ChannelActivity : AppCompatActivity() {
          * */
         setContent {
             ChatTheme {
+                ChatTheme.colors.appBackground
```
adasiewiczr, line 70: This change looks suspicious

filbabic: That's true, not sure what happened here :D

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/header/VersionPrefixHeader.kt
```diff
@@ -0,0 +1,10 @@
+package io.getstream.chat.android.client.header
+
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+
+@InternalStreamChatApi
+public enum class VersionPrefixHeader(public val prefix: String) {
+    CORE_ANDROID("stream-chat-android-"),
+    COMPOSE("stream-chat-compose-")
+    // TODO - @Rafal - we'll need to add a value for the old UI component too
```
JcMinarro, line 9: Could you add the one for `UI-Components` module?
The old UI should't need it, we will remove it soon.

I think you should be able to init it on `ChatUIInitializer` [class](https://github.com/GetStream/stream-chat-android/blob/main/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/internal/ChatUIInitializer.kt)

JcMinarro (changes requested):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: Is this method usefull for our customers?
I think we are using it into our Sample App to show the version we are using.
Maybe we could keep it public

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1576,7 +1578,12 @@ public class ChatClient internal constructor(
     }
 
     public companion object {
-        private const val VERSION_PREFIX = "stream-chat-android-"
+        @InternalStreamChatApi
+        @JvmStatic
+        public var VERSION_PREFIX: String by Delegates.vetoable(VersionPrefixHeader.CORE_ANDROID.prefix) { _, _, new ->
```
zsmb13, line 1583: This could just have the enum type instead of a `String`, would simplify a bunch of things

filbabic: Very good point - I'll update!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
filbabic: I'll keep it as public w/o the annotation!

zsmb13: I think public _with_ annotation is fine as we only want to use it in our own sample, and not have clients using it

JcMinarro (approved):
Good Job!!!!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
filbabic: @JcMinarro @adasiewiczr Should we keep the annotation then? I can see both sides at the moment, but I'll let you folks decide given that you have more experience with this.

PR #2244, opened 2021-08-18
Feature/2212 improve attachmentfactory api design

### LINKED ISSUE #2212: [Compose] - Improve AttachmentFactory API based on Marton's comments
Looking at the code around attachments:

AttachmentFactory(
    { state -> ImageAttachmentFactory(state) },
    { message -> message.attachments.all { it.type == "image" } }
),

Shouldn't ImageAttachmentFactory be a subclass of AttachmentFactory that already has this filter contained, and the thing that we call the ImageAttachmentFactory now be renamed to ImageAttachment, as it's just a composable function?

So we could end up with something like
public val defaultFactories: List<AttachmentFactory> = listOf(
    LinkAttachmentFactory(),
    ImageAttachmentFactory(), 
    FileAttachmentFactory(),
)

And then

public class ImageAttachmentFactory : AttachmentFactory(
    attachmentContent = { state -> ImageAttachment(state) },
    predicate = { message -> message.attachments.all { it.type == "image" } }
)

 - [x] We show generic file icons instead of thumbnails for image files
 - [x] Icons are cropped/cut in their corners, instead of showing up fully
 - [x] Download & open attachments when tapped (files for now)

filbabic:
Fixes #2212.

### 🎯 Goal

We had a lot of improvements and API changes here around the Attachments feature.

Some of the things we improved:

- [x] Updated the API to make it easier to override and implement custom attachment factories
- [x] We show generic file icons instead of thumbnails for image files
- [x] Icons are cropped/cut in their corners, instead of showing up fully
- [x] Download & open attachments when tapped (files for now)

### 🛠 Implementation details

We changed the `AttachmentFactory` to expose overrideable functions instead of properties.

We added general checks about file attachments and if they're of the "image" type.

Decoupled components for ease of use and easier navigation.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| before | after |

### 🧪 Testing

We can send and select various types of attachments without changes. We also see UI improvements for image thumbnails and file icons.

### ☑️ Checklist

- ~~[ ] New code is covered by unit tests~~
- [x] Reviewers added

### 🎉 GIF

fixed


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/LinkAttachmentFactory.kt
```diff
@@ -0,0 +1,134 @@
...
+ * build the UI for the message.
+ * */
+public class LinkAttachmentFactory : AttachmentFactory() {
+
+    override fun predicate(message: Message): Boolean {
+        return message.attachments.any { it.titleLink != null || it.ogUrl != null }
+    }
+
+    @Composable
+    public override fun Content(attachmentState: AttachmentState) {
```
zsmb13, line 41: Nitpick of nitpicks: visibility modifiers are redundant when overriding a `public` method (applies in other files too)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
```diff
@@ -0,0 +1,51 @@
...
+ * */
+public abstract class AttachmentFactory {
+    /**
+     * Returns if this specific factory can handle a specific message.
+     *
+     * @param message - The message to check.
+     * @return a boolean value, if we can consume the message and render UI.
+     * */
+    public fun canHandle(message: Message): Boolean {
+        return message.attachments.isNotEmpty() && predicate(message)
```
zsmb13, line 32: I'm thinking more and more that this should be the responsibility of the factory.

Checks like `attachments.any { it.type = "whatevsies" }` will automatically be `false` for messages that don't have collections.

And we can amend checks like `attachments.all { it.type = "type" }` with an `attachments.isNotEmpty() &&` at the start. That's part of the logic of whether the factory can render the given message's attachments.

(And perhaps on the use site we can also have an extra `if` that just doesn't check in with the factories if a message has no attachments.)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/components/FileAttachmentImage.kt
```diff
@@ -0,0 +1,46 @@
...
+import io.getstream.chat.android.compose.ui.util.MimeTypeIconProvider
+
+/**
+ * Represents the image that's shown in file attachments. This can be either an image/icon that represents the file type
+ * or a thumbnail in case the file type is an image.
+ *
+ * @param attachment - The attachment we use to show the image.
+ * */
+@Composable
+public fun FileAttachmentImage(attachment: Attachment) {
```
zsmb13, line 23: Could probably live in one file with `FileAttachmentContent`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -167,7 +167,7 @@ public fun DefaultMessageContainer(
                             DeletedMessageContent()
                         } else {
                             Column {
-                                attachmentFactory?.factory?.invoke(
+                                attachmentFactory?.Content(
```
zsmb13, line 170: `attachmentFactory?.Content` this looks uncomfortable, but I don't know if we can do anything to make the syntax nicer. I guess Composable functions are just not meant to be members? 😀 Are there examples of this in first-party APIs perhaps?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamShapes.kt
```diff
@@ -13,21 +13,24 @@ import androidx.compose.ui.unit.dp
...
 ) {
     public companion object {
         public val default: StreamShapes = StreamShapes(
             avatar = CircleShape,
             myMessageBubble = RoundedCornerShape(topStart = 16.dp, topEnd = 16.dp, bottomStart = 16.dp),
             otherMessageBubble = RoundedCornerShape(topStart = 16.dp, topEnd = 16.dp, bottomEnd = 16.dp),
             inputField = RoundedCornerShape(24.dp),
-            attachment = RoundedCornerShape(16.dp)
+            attachment = RoundedCornerShape(16.dp),
+            imageThumbnail = RoundedCornerShape(8.dp)
```
zsmb13, line 33: Lack of trailing commas in a bunch of places on this PR - it's why this file has more line changes than it needs to, they were already missing at the ends of both lists that are modified here.

We're awaiting the next release of ktlint that will hopefully place trailing commas more aggressively than the Studio formatter.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
```diff
@@ -222,11 +222,12 @@ private fun SelectedMessage(message: Message) {
...
                 } else {
                     Column {
-                        attachmentFactory?.factory?.invoke(
+                        attachmentFactory?.Content(
                             AttachmentState(
                                 modifier = Modifier.padding(4.dp),
-                                message = MessageItem(message, None)
-                            ) {}
+                                message = MessageItem(message, None),
+                                onLongItemClick = {}
```
zsmb13, line 229: Why the explicit empty value?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/FileAttachmentFactory.kt
```diff
@@ -0,0 +1,127 @@
...
+import io.getstream.chat.android.offline.ChatDomain
+
+/**
+ * An extension of the [AttachmentFactory] that validates attachments as files and uses [FileAttachmentContent] to
+ * build the UI for the message.
+ * */
+public class FileAttachmentFactory : AttachmentFactory() {
+
+    override fun predicate(message: Message): Boolean {
+        return message.attachments.any { it.type != "image" }
```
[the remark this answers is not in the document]
filbabic: That is basically the use case - however, thinking about the `isEmpty()` check and all, I think we can do it a bit differently.

We should trigger all attachment factories that make sense - if there is an image and a link, we should show both (link first I think). I'll try to work this out and improve these checks.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/LinkAttachmentFactory.kt
filbabic: Okay - it initially showed me an error as per the linting rules, will remove.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/components/FileAttachmentImage.kt
filbabic: It's also being used in `MessageInputFileAttachments`, which is why I decoupled it. Made more sense as it was reused

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
filbabic: Yeah this is very ugly - I generally prefer the parameter/function property approach.

I'm thinking of just going back to that because it's a bit nicer, even though this API might make more sense

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamShapes.kt
filbabic: Yeah it's been inconsistent for me and something I easily forget (not used to putting them honestly). I'll go through some of these files and add them where I see them missing

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
filbabic: Good point, removed

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
filbabic: Yeah I originally wanted for our users not to have to worry about this check, but if they implement a custom factory, it's sensible to expect and `isEmpty()` check or whatever makes sense for them.

I think it would be a better approach to try and find `all` factories that work for a given message and then just render them all.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/FileAttachmentFactory.kt
```diff
@@ -0,0 +1,127 @@
...
+ * */
+@Composable
+public fun FileAttachmentContent(attachmentState: AttachmentState) {
+    val (modifier, messageItem, _) = attachmentState
+    val (message, _) = messageItem
+
+    Column(
+        modifier = modifier
+            .wrapContentHeight()
+            .width(200.dp)
```
aldefy, line 64: im guessing since this is internal `message` UI so we don't allow customization , would this be useful  for later usecase

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/ImageAttachmentFactory.kt
```diff
@@ -0,0 +1,168 @@
...
+@OptIn(ExperimentalFoundationApi::class)
+@Composable
+public fun ImageAttachmentContent(attachmentState: AttachmentState) {
+    val (modifier, messageItem, onLongItemClick) = attachmentState
+    val (message, _) = messageItem
+    val context = LocalContext.current
+
+    Row(
+        modifier
+            .size(height = 200.dp, width = 250.dp)
```
aldefy, line 66: +1 on this for customization, trying to understand the decision to add similar lines  for new components

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/FileAttachmentFactory.kt
filbabic: Yeah this is an internal thing - the default UI/behavior we expose for attachments. We can still customize it through some ChatTheme params, like the `attachmentShape` and through the modifier.

We could expose more customization, but if people want completely custom items, they can just override the factory overall.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/ImageAttachmentFactory.kt
filbabic: Similar to the comment above - this is according to our own default UI, so if people want to customize it more, we'll probably expose things like `messageWidth` or just let people customize the entire item in an attachment, or just override the factory.

UNRELEASED_CHANGELOG.md
```diff
@@ -91,11 +91,18 @@
...
+- Fixed a bug where image file attachments weren't shown as thumbnails.
 - Added an overlay to the `ChannelInfo` that blocks outside clicks.
 - Updated the `ChannelInfoUserItem` to use the `UserAvatar`.
 
 ### ⬆️ Improved
 - Added default date and time formatting to Channel and Message items.
+- Improved attachments API by providing cleaner examples of attachment factories.
+- Updated documentation & examples.
+- Decoupled attachment content to specific attachment files.
+- Decouple message attachment content to a `MessageAttachmentsContent` component.
```
zsmb13, line 104:
```suggestion
- Decoupled message attachment content to a `MessageAttachmentsContent` component.
```

docusaurus/docs/Android/04-compose/02-general-customization/02-attachment-factory.mdx
```diff
@@ -3,42 +3,33 @@
...
-    private val predicate: (Message) -> Boolean,
+public abstract class AttachmentFactory(
+    public val predicate: (attachments: List<Attachment>) -> Boolean,
+    public val content: @Composable (AttachmentState) -> Unit,
 ) {
     // Validator to check if we can handle the given message
-    public fun canHandle(message: Message): Boolean {
-        return predicate(message)
+    public fun canHandle(attachments: List<Attachment>): Boolean {
+        return predicate(attachments)
```
zsmb13, line 12: Is this method still necessary, given that `predicate` is public?

docusaurus/docs/Android/04-compose/04-message-components/05-attachments-picker.mdx
```diff
@@ -67,12 +67,12 @@ AttachmentsPicker(
     ..., // State and UI customization
     onAttachmentsSelected = { attachments -> 
         // Dismiss the picker and store the attachments
-        attachmentsPickerViewModel.onShowAttachments(false)
-        composerViewModel.onAttachmentsSelected(attachments)
+        attachmentsPickerViewModel.changeAttachmentState(false)
```
zsmb13, line 70: `changeAttachmentState(false)` is quite cryptic, could either use a better method name or have a named parameter at the call site here

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/components/MessageAttachmensContent.kt
```diff
@@ -0,0 +1,45 @@
...
+    onLongItemClick: (Message) -> Unit,
+) {
+    val (message, _) = messageItem
+
+    if (message.attachments.isNotEmpty()) {
+
+        val (links, attachments) = message.attachments.partition { (it.ogUrl != null || it.titleLink != null) }
+        val linkFactory = ChatTheme.attachmentFactories.firstOrNull { links.isNotEmpty() && it.canHandle(links) }
+        val attachmentFactory =
+            ChatTheme.attachmentFactories.firstOrNull { attachments.isNotEmpty() && it.canHandle(attachments) }
```
zsmb13, line 31: In both calls here, it's odd to have the `isNotEmpty` calls inside the `firstOrNull` loop - that value is independent of the `attachmentFactories`.

We could do something like one of these approaches (I prefer the first):

```kotlin
val linkFactory = if (links.isNotEmpty()) {
    ChatTheme.attachmentFactories.firstOrNull { it.canHandle(links) }
} else {
    null
}
val attachmentFactory = attachments.takeIf { it.isNotEmpty() }?.let { attachments ->
    ChatTheme.attachmentFactories.firstOrNull { it.canHandle(attachments) }
}
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/components/MessageAttachmensContent.kt
```diff
@@ -0,0 +1,45 @@
+package io.getstream.chat.android.compose.ui.attachments.components
```
zsmb13, line 1: Typo in the filename 😀

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/components/MessageAttachmensContent.kt
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Represents the content that's shown in message attachments. We decide if we need to show link previews or other
+ * attachments.
+ *
+ * @param messageItem - The message that contains the attachments.
+ * @param onLongItemClick - Handler for long item taps on this content.
+ * */
+@Composable
+public fun MessageAttachmentsContent(
```
zsmb13, line 20: Giphy attachments have something weird going on after the changes on this PR.

|develop|this branch|
|---|---|
| image| image|

adasiewiczr: This is probably the same problem we had in core SDK - Giphy attachment has `titleLink != null` and we were rendering wrong view type

UNRELEASED_CHANGELOG.md
filbabic: Will apply!

docusaurus/docs/Android/04-compose/04-message-components/05-attachments-picker.mdx
filbabic: Will add named arguments!

docusaurus/docs/Android/04-compose/02-general-customization/02-attachment-factory.mdx
```diff
@@ -3,42 +3,29 @@
...
+    ImageAttachmentFactory(),
+    FileAttachmentFactory()
 )
 ```
 
-These factories perform specific checks, like if the `ogUrl` exists, giving us information that this is a link attachment, or if all of the message attachments are of the `"image"` type.
+These factories perform specific checks that you can explore by opening each specific factory.
 
-Each of these factories supplies a validator, as well as a factory composable lambda function that calls `LinkAttachmentFactory`, `ImageAttachmentFactory` or `FileAttachmentFactory`.
+Each of these factories supplies a predicate, as well as a content composable lambda function that calls `LinkAttachmentContent`, `ImageAttachmentContent` or `FileAttachmentContent`.
```
zsmb13, line 28: We don't list `GiphyAttachmentFactory` here, but maybe this is a lesson in how we shouldn't try listing things like this in docs 😀

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/GiphyAttachmentFactory.kt
```diff
@@ -0,0 +1,100 @@
...
+import com.getstream.sdk.chat.utils.extensions.imagePreviewUrl
+import io.getstream.chat.android.compose.R
+import io.getstream.chat.android.compose.state.messages.attachments.AttachmentState
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+
+/**
+ * An extensions of the [AttachmentFactory] that validates and shows Giphy attachments.
+ * */
+public class GiphyAttachmentFactory : AttachmentFactory(
+    canHandle = { attachments -> attachments.any { it.type == "giphy" } },
```
zsmb13, line 33: Nit: We have a couple different occurrances of the `"giphy"` constant string, we could extract to a constant somewhere

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/GiphyAttachmentFactory.kt
```diff
@@ -0,0 +1,100 @@
...
+@Composable
+public fun GiphyAttachmentContent(attachmentState: AttachmentState) {
+    val context = LocalContext.current
+    val (modifier, messageItem, onLongItemClick) = attachmentState
+    val (message, _) = messageItem
+    val attachment = message.attachments.firstOrNull { it.type == "giphy" }
+
+    requireNotNull(attachment) {
+        IllegalStateException("Missing Giphy attachment.")
+    }
```
zsmb13, line 54: `requireNotNull` will put whatever you hand it into an `IllegalArgumentException` itself, so 

```suggestion
    requireNotNull(attachment) { "Missing Giphy attachment." }
```

If you want to keep the `IllegalStateException` specifically, `checkNotNull` is your friend.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/GiphyAttachmentFactory.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    val previewUrl = attachment.titleLink ?: attachment.ogUrl
+
+    requireNotNull(previewUrl) {
+        IllegalStateException("Missing preview URL.")
+    }
```
zsmb13, line 60: Same here

```suggestion
    requireNotNull(previewUrl) { "Missing preview URL." }
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
```diff
@@ -0,0 +1,32 @@
...
+public object StreamAttachmentFactories {
+
+    /**
+     * Default attachment factories we provide, which can transform image, file and link attachments.
+     *
+     * Uses the functions below to display the UI.
+     * */
+    public val defaultFactories: List<AttachmentFactory> = listOf(
+        LinkAttachmentFactory(),
+        GiphyAttachmentFactory(),
```
zsmb13, line 16: Perhaps `Giphy` should go first here, to catch things before `LinkAttachmentFactory` would get involved. Then we don't need the `Giphy` check in that factory (though we might still want to keep it, just in case)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/StreamCoilImageLoader.kt
```diff
@@ -0,0 +1,60 @@
...
+import coil.ImageLoader
+import coil.ImageLoaderFactory
+import com.getstream.sdk.chat.coil.StreamImageLoaderFactory
+
+/**
+ * Provides a custom image loader that uses the [StreamImageLoaderFactory] to build the default loading settings.
+ *
+ * Gives support to load GIFs.
+ * */
+internal object StreamCoilImageLoader {
```
zsmb13, line 13: Let's review this later, but if it works for now, we can ship it (`internal` anyway)

zsmb13 (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/StreamCoilImageLoader.kt
filbabic: Yeah my goal would be to have the already `internal` API as public/reusable element, so that I don't have to C/P the solution here.

docusaurus/docs/Android/04-compose/02-general-customization/02-attachment-factory.mdx
filbabic: True, I've gone through the docs around to see if I can update it, but I've missed this point, I'll update accordingly.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/GiphyAttachmentFactory.kt
filbabic: Will use the `ModelType.attach_giphy` as that's already available!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/GiphyAttachmentFactory.kt
filbabic: Nice, thanks, will apply all this!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
filbabic: I think relying on the ordering is not the best approach, because we never know what the BE team does to the data - e.g. this whole change with the `titleLink`.

This way we can check for links, but unfortunately giphy also has links (even though it could use other properties), so we have to add this check.

PR #2249, opened 2021-08-19
Ensure channel loading view has no parent

amitkma:
### 🎯 Goal

It is done to ensure that the loading view set by the user has no parent to avoid crashes at runtime. 

### 🛠 Implementation details

Simply, remove the view from the parent if it has any parent. 

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.Gravity
 import android.view.View
+import android.view.ViewGroup
 import android.widget.FrameLayout
 import androidx.annotation.DrawableRes
 import androidx.core.view.isVisible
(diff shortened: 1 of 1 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -116,6 +117,11 @@ public class ChannelListView : FrameLayout {
      */
     @JvmOverloads
     public fun setLoadingView(view: View, layoutParams: LayoutParams = defaultChildLayoutParams) {
+        // Make sure the loading view has no parent.
+        view.parent?.let { parent ->
+            (parent as ViewGroup).removeView(view)
+        }
```
zsmb13, line 123: Good catch, we do need a check here.

Perhaps we could be more aggressive here and just throw an exception for this case. If clients write code where they pass in a View that has a parent, that's a clear bug. Us removing something randomly might be unexpected as well.

So something like

```kotlin
require(view.parent == null)
```

Also, we have custom empty views in addition to custom loading views, and we have this in other classes too. We should add checks everywhere.

amitkma: Yes, good idea. I will update with `require` instead of detaching from the parent.

amitkma:
Not very useful so closing now.

Issue #2132, opened 2021-08-03
[Compose] - General API improvements in visibility & customization


filbabic:
We need to revise which components we provide as "public" API.

Some considerations:
- DefaultMessageContainer and some of its children
- ReactionOptions
- MessageComposer children
- Attachment factories

PR #2252, opened 2021-08-19
2224 - Add basic date format customization

### LINKED ISSUE #2224: [Compose] - Date formatting - Improve date formatting and expose a way to customize the format.
We need to expose a way to customize the date format we use for the app's timestamps.

We also need to improve the default format to work like the core SDK.

filbabic:
Fixes #2224

### 🎯 Goal

This PR touches upon #2224 and #2132. The idea is to improve our general-app customization options where users can set up things like default time formatting, enforcing unique reactions and more.

In this case, we wanted to update the app to use the core SDK formatting for timestamps. We also provided a few small things that'll allow us to configure the app.

### 🛠 Implementation details

- Added `Timestamp` - component that formats the date you give it into a default time format that's app-wide.
- Added `DateFormatter` to `ChatTheme`.
- Updated relevant classes and call sites.

~~@zsmb13 @aldefy What do you think about these changes - `StreamConfiguration` specifically~~. We need a central place to group these global config options.

Also, what do you think about exposing these classes as `data class`es? This was it'll be easier for our users to create small mutations of the default behavior (using `copy()`. 

**LMK and once we decide on something, I'll update the changelog and the docs (in docus).** 

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -537,11 +536,7 @@ internal fun MessageFooter(
                 color = ChatTheme.colors.textLowEmphasis
             )
 
-            Text(
-                SimpleDateFormat.getTimeInstance().format(message.createdAt ?: Date()),
-                style = ChatTheme.typography.footnote,
-                color = ChatTheme.colors.textLowEmphasis
-            )
+            Timestamp(date = message.updatedAt ?: message.createdAt)
```
aldefy, line 539: [general] Do we not name smaller components like this example to be named `TimestampText` : cause i feel the current name could be cross referenced as a `model` layer

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -49,6 +49,7 @@ public class MessageListViewModel(
     public val chatDomain: ChatDomain,
     private val channelId: String,
     private val messageLimit: Int = 0,
+    private val enforceUniqueReactions: Boolean = true,
```
aldefy, line 52: curious , how is this set ? since its `private`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
```diff
@@ -29,6 +30,10 @@ private val LocalReactionTypes = compositionLocalOf<Map<String, Int>> {
     error("No reactions provided! Make sure to wrap all usages of Stream components in a ChatTheme.")
 }
 
+private val LocalStreamConfiguration = compositionLocalOf<StreamConfiguration> {
+    error("No stream configuration provided! Make sure to wrap all usages of Stream components in a ChatTheme.")
```
aldefy, line 34: Nitpicking , shouldn't `stream` be `Stream` ? branding and all

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/Timestamp.kt
```diff
@@ -0,0 +1,29 @@
...
+@Composable
+public fun Timestamp(
+    date: Date?,
+    modifier: Modifier = Modifier,
+) {
+    val formatter = ChatTheme.configuration.dateFormatter
+
+    Text(
+        modifier = modifier,
+        text = formatter.formatDate(date),
```
zsmb13, line 25: We also have a `formatTime` extension that we use in a couple places in the UI Components implementation. Might want to consider having two composables like this, or a parameter that decides how to format.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/Timestamp.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Date
+
+/**
+ * Represents a timestamp in the app, that's used primarily for channels and messages.
+ *
+ * @param date - The date to show in the timestamp.
+ * @param modifier - Modifier for styling.
+ * */
+@Composable
+public fun Timestamp(
```
zsmb13, line 17: Really love how this can encapsulate rendering and formatting a timestamp!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
zsmb13: As it's a class name, I'd go with
```suggestion
    error("No StreamConfiguration provided! Make sure to wrap all usages of Stream components in a ChatTheme.")
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamConfiguration.kt
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Central place for general app configuration options. Use this class to define how some of the features in the app
+ * work.
+ *
+ * @param dateFormatter - The formatter used for timestamps across the app.
+ * @param enforceUniqueReactions - Flag that enables or disables unique reactions on messages.
+ * */
+public class StreamConfiguration(
+    public val dateFormatter: DateFormatter,
+    public val enforceUniqueReactions: Boolean,
```
zsmb13, line 15: Let's consider what exactly should go here vs into separate parameters in ChatTheme. Would be nice to set guidelines for this somewhere.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
zsmb13: It's a constructor parameter

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
filbabic: So I wouldn't name this `TimestampText` because if we decide to add more components to it - e.g. an icon for some case, or something more, then it's no longer just a text.

I think if we're creating some `model` layer class, we'd name it a bit more specifically - like `MessageTimestamp`. or `TimestampItem` or `TimestampData` or something similar.

Does this make sense? Do you have any other idea of what we could name it that's not explicitly exposing a `Text` in it?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/Timestamp.kt
filbabic: Good point - I was thinking of exposing a formatting function or something, instead of the formatter itself, because that way we can have the same formatter, but override behavior anyway.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
filbabic: Will update!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamConfiguration.kt
filbabic: Do you think we can move the `reactionOptions` here?

E.g. the guidelines can be something like:
- Anything that's a theming option/direct UI impact has its own provider
- Anything that's state-based, business logic affecting goes here?

We can discuss more if this makese sense in the long run (e.g. if something is both UI & logic/state based)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
aldefy: For some odd reason i think `component` Suffix could be good alternative to our Compose widgets per say but i got the idea

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/Timestamp.kt
aldefy: +1 to Filip point , format function over a class

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
aldefy: this can be discussed later , resolving it for now @filbabic

aldefy (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamConfiguration.kt
```diff
@@ -0,0 +1,27 @@
...
+) {
+
+    public companion object {
+        /**
+         * Provides the default configuration for the app.
+         * */
+        public fun defaultConfiguration(context: Context): StreamConfiguration = StreamConfiguration(
+            DateFormatter.from(context),
+            true
```
zsmb13, line 24: Some named params would help here

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/DateFormat.kt
```diff
@@ -0,0 +1,42 @@
...
+ * Represents a timestamp in the app, that's used primarily for channels and messages.
+ *
+ * @param date - The date to show in the timestamp.
+ * @param modifier - Modifier for styling.
+ * @param formatter - The formatter that's used to format the date to a String.
+ * @param formatType - The type of formatting to provide - either a timestamp or a full date. We format the information
+ * using [DATE] by default, as it's the most common behavior.
+ * */
+@Composable
+public fun DateFormat(
```
zsmb13, line 25: The name we landed on for this component is weird for me. I keep thinking of [`SimpleDateFormat`](https://docs.oracle.com/javase/7/docs/api/java/text/SimpleDateFormat.html), I think this is something that contains info on how to format a date, and not something that displays it on the UI.

I think the original `Timestamp` (?) was probably better, but let's see what other ideas we have for this.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamConfiguration.kt
zsmb13: Not an easy question. For example, both `reactionTypes` and `attachmentFactories` have UI-related code directly in them, so according to those guidelines, they should not be inside the config. Maybe `dateFormatter` shouldn't be either 😅

PR #2255, opened 2021-08-19
[CAS-1229] Fixing state for anonymous user

leandroBorgesFerreira:
[CAS-1229](https://stream-io.atlassian.net/browse/CAS-1229)

### 🎯 Goal

Fix the state machine for anonymous user

### 🛠 Implementation details

Adding 3 new states. 

#### SocketState.Disconnected -> ClientStateEvent.ConnectionRequested -> SocketState.Pending
When the activity with our initialization SDK is killed, we go to the a disconnected state and we can request a connect from the state, not only `Idle` state. You can enable `Don't Keep Activities` to test this. Notice that we always go to the error state in the `setUser` method (ChatClient line 281).

#### UserState.Anonymous.AnonymousUserSet -> UserStateEvent.ConnectAnonymous -> UserState.Anonymous.Pending
The same situation as above, but with anonymous user. If the activity is recreated and `connectAnonymousUser` is called one more time, the user is already set (differently from connectUser, we don't check the state of `userStateService` and `socketStateService`) and we can request a connection although the user is already set. 

#### UserState.Anonymous.AnonymousUserSet -> UserStateEvent.UserUpdated -> UserState.Anonymous.AnonymousUserSet
It is possible to have an user updated also for the anonymous state. Without adding this state, the app will crash every time it comes back to onResume. 

### 🧪 Testing

Use:

```
 ChatClient.instance().connectAnonymousUser().enqueue()
```

instead of our current way to set user in our sample app inside `UserLoginViewModel` and check the crash. Also enable `Don't keep activities` to see different error.

### ☑️ Checklist

- ~[x] Comparison screenshots added for visual changes~
- [x] Reviewers added

### 🎉 GIF

giphy

### DIFF
--- a/UNRELEASED_CHANGELOG.md
+++ b/UNRELEASED_CHANGELOG.md
@@ -61,7 +61,7 @@
 
 ## stream-chat-android-ui-common
 ### 🐞 Fixed
-Bug related to anonymous users when app goes onPause and then comes back to onResume state. 
+- Fixed a bug in state handling for anonymous users. 
 ### ⬆️ Improved
 
 ### ✅ Added
--- a/UNRELEASED_CHANGELOG.md
+++ b/UNRELEASED_CHANGELOG.md
@@ -61,7 +61,7 @@
 
 ## stream-chat-android-ui-common
 ### 🐞 Fixed
-
+Bug related to anonymous users when app goes onPause and then comes back to onResume state. 
 ### ⬆️ Improved
 
 ### ✅ Added
@@ -82,7 +82,7 @@
   - `ChannelListActivity` - thin wrapper around `ChannelListFragment`
   - `MessageListFragment` - message list screen which internally contains `MessageListHeaderView`, `MessageListView`, `MessageInputView`.
   - `MessageListActivity` - thin wrapper around `MessageListFragment`
-    
+
 ### ⚠️ Changed
 
 ### ❌ Removed
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -259,7 +259,7 @@ public class ChatClient internal constructor(
         }
         val userState = userStateService.state
         when {
-            userState is UserState.UserSet && userState.user.id == user.id && (socketStateService.state == SocketState.Idle || socketStateService.state == SocketState.Disconnected) -> {
+            userState is UserState.UserSet && userState.user.id == user.id && socketStateService.state == SocketState.Idle -> {
                 userStateService.onUserUpdated(user)
                 tokenManager.setTokenProvider(tokenProvider)
                 connectionListener = listener
@@ -360,21 +360,26 @@ public class ChatClient internal constructor(
         level = DeprecationLevel.ERROR,
     )
     public fun setAnonymousUser(listener: InitConnectionListener? = null) {
-        socketStateService.onConnectionRequested()
-        userStateService.onSetAnonymous()
-        connectionListener = object : InitConnectionListener() {
-            override fun onSuccess(data: ConnectionData) {
-                notifySetUser(data.user)
-                listener?.onSuccess(data)
-            }
+        if (userStateService.state is UserState.NotSet) {
+            socketStateService.onConnectionRequested()
+            userStateService.onSetAnonymous()
+            connectionListener = object : InitConnectionListener() {
+                override fun onSuccess(data: ConnectionData) {
+                    notifySetUser(data.user)
+                    listener?.onSuccess(data)
+                }
 
-            override fun onError(error: ChatError) {
-                listener?.onError(error)
+                override fun onError(error: ChatError) {
+                    listener?.onError(error)
+                }
             }
+            config.isAnonymous = true
+            warmUp()
+            socket.connectAnonymously()
+        } else {
+            logger.logE("Failed to connect user. Please check you don't have connected user already")
+            listener?.onError(ChatError("User cannot be set until previous one is disconnected."))
         }
-        config.isAnonymous = true
-        warmUp()
-        socket.connectAnonymously()
     }
 
     @CheckResult
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -259,7 +259,7 @@ public class ChatClient internal constructor(
         }
         val userState = userStateService.state
         when {
-            userState is UserState.UserSet && userState.user.id == user.id && socketStateService.state == SocketState.Idle -> {
+            userState is UserState.UserSet && userState.user.id == user.id && (socketStateService.state == SocketState.Idle || socketStateService.state == SocketState.Disconnected) -> {
                 userStateService.onUserUpdated(user)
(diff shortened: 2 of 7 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -259,7 +259,7 @@ public class ChatClient internal constructor(
         }
         val userState = userStateService.state
         when {
-            userState is UserState.UserSet && userState.user.id == user.id && socketStateService.state == SocketState.Idle -> {
+            userState is UserState.UserSet && userState.user.id == user.id && (socketStateService.state == SocketState.Idle || socketStateService.state == SocketState.Disconnected) -> {
```
ogkuzmin, line 262: For me it's not clear why we need to set user when the socket is disconnected.
We check for the idle socket state, because for pushes and background job we could set user without WS connection, and when we open app we want to set user with full flow (setting WS connection). For that this condition.
But if we were disconnected, we just released the WS connection not to waste resources. For this case, we don't need to reset user again

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/SocketStateService.kt
```diff
@@ -54,6 +54,7 @@ internal class SocketStateService {
             state<SocketState.Disconnected> {
                 onEvent<ClientStateEvent.DisconnectedEvent> { _, _ -> stay() }
                 onEvent<ClientStateEvent.DisconnectRequestedEvent> { _, _ -> SocketState.Idle }
+                onEvent<ClientStateEvent.ConnectionRequested> { _, _ -> SocketState.Pending }
```
ogkuzmin, line 57: The same as described above we shouldn't reinit connection manually when we were disconnected

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/UserStateService.kt
```diff
@@ -44,6 +44,8 @@ internal class UserStateService {
             onEvent<UserStateEvent.UnsetUser> { _, _ -> UserState.NotSet }
         }
         state<UserState.Anonymous.AnonymousUserSet> {
+            onEvent<UserStateEvent.ConnectAnonymous> { _, _ -> UserState.Anonymous.Pending }
```
ogkuzmin, line 47: The same: there is no reason to connect user if we already set anonymous user

ogkuzmin (changes requested):

ogkuzmin:
We shouldn't rely on activity lifecycle to set user. It's not right. Activity could be recreated, but app keep state as foreground

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/SocketStateService.kt
leandroBorgesFerreira: If the app is killed by the OS, the websocket will lose the connection and enter in this state. A user that sets the connection in an activity will see a crash if he/she doesn't manage it in its own side. Maybe this is a state that shouldn't happen... but it seams too much that the app crashes if the user asks for another connection when the app comes back in the foreground. 

Maybe we should deal log it or deal with it in another way? Let's not throw an exception here.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: When the activity is recreated we could come in this if the user chooses to initialises our SDK in the activity instead of the `Application` class. He/she will go into the else case... For me it makes more sense to connect again, because in the end it was a second request made by the activity. 

In this case removing the check for `SocketState.Disconnected` is less problematic, because there's no crash and we log what's going on... But asking for a connection in the `onCreate` of an activity doesn't seam too crazy for me, I think it makes to support it. What do you think?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/UserStateService.kt
leandroBorgesFerreira: If the app is killed by the OS, the websocket will lose the connection and enter in this state. A user that sets the connection in an activity will see a crash if he/she doesn't manage it in its own side. Maybe this is a state that shouldn't happen... but it seams too much that the app crashes if the user asks for another connection when the app comes back in the foreground.

Maybe we should deal log it or deal with it in another way? Let's not throw an exception here.

ogkuzmin: If app got killed then nothing is kept in memory, state of this would be Idle

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
ogkuzmin: IMO, it's not appropriate usage of our SDK. UI lifecycle shouldn't impact set user flow @JcMinarro @adasiewiczr Guys, what do you think?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/clientstate/UserStateService.kt
leandroBorgesFerreira: Or maybe we do throw an exception, but provide a message, like: 

"You are trying to connection anonymously, but the user is already set, perhaps you are asking for a connection a second time? Please confirm if you are managing the lifecycle of your Activity correctly"

ogkuzmin: I think it's better to check state of `UserStateService` in `ChatClient::setAnonymousUser` similarly to `ChatClient::setUser`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
adasiewiczr: ``` if the user chooses to initialises our SDK in the activity instead of the Application class. He/she will go into the else case```
Are we setting `SocketState.Disconnected` during SDK initialization? I think it should be `disconnected` only if user was already set.
In our sample app, a user is connected in ViewModel and we don't have any problems with that.

leandroBorgesFerreira: @adasiewiczr If you enable "Don't keep activities" observe that we always go to the else branch. But yes, the app continue working just fine

leandroBorgesFerreira: Let's remove this than, no need to make this a super long discussion.

ogkuzmin (approved):
LGTM 👍

UNRELEASED_CHANGELOG.md
```diff
@@ -61,7 +61,7 @@
 
 ## stream-chat-android-ui-common
 ### 🐞 Fixed
-
+Bug related to anonymous users when app goes onPause and then comes back to onResume state.
```
zsmb13, line 64:
```suggestion
- Fixed a bug in state handling for anonymous users. 
```

Bullet points! (+ only useful details)

PR #2269, opened 2021-08-25
[CAS-1252] Fix/ws connection issues

JcMinarro:
### 🎯 Goal
Fix WS connection issues

### 🛠 Implementation details
There was some race-conditions between our `NetworkStateListener` and our `HealthMonitor` that opened/closed the WS connection when the shouldn't.

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
@@ -31,7 +31,7 @@ internal class NetworkStateProvider(private val connectivityManager: Connectivit
     private var isConnected: Boolean = isConnected()
 
     @Volatile
-    private var listeners: List<NetworkStateListener> = listOf()
+    private var listeners: Set<NetworkStateListener> = setOf()
 
     private val isRegistered: AtomicBoolean = AtomicBoolean(false)
 
@@ -56,8 +56,9 @@ internal class NetworkStateProvider(private val connectivityManager: Connectivit
 
     fun subscribe(listener: NetworkStateListener) {
         listeners = listeners + listener
-        connectivityManager.registerNetworkCallback(NetworkRequest.Builder().build(), callback)
-        isRegistered.set(true)
+        if (isRegistered.compareAndSet(false, true)) {
+            connectivityManager.registerNetworkCallback(NetworkRequest.Builder().build(), callback)
+        }
     }
 
     fun unsubscribe(listener: NetworkStateListener) {
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
@@ -42,7 +42,11 @@ internal class ChatSocketServiceImpl constructor(
     private val eventUiHandler = Handler(Looper.getMainLooper())
     private val healthMonitor = HealthMonitor(
         object : HealthMonitor.HealthCallback {
-            override fun reconnect() = this@ChatSocketServiceImpl.reconnect(connectionConf)
+            override fun reconnect() {
+                if (state is State.DisconnectedTemporarily) {
+                    this@ChatSocketServiceImpl.reconnect(connectionConf)
+                }
+            }
             override fun check() {
                 (state as? State.Connected)?.let {
                     sendEvent(it.event)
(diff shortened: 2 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -31,7 +31,7 @@ internal class NetworkStateProvider(private val connectivityManager: Connectivit
     private var isConnected: Boolean = isConnected()
 
     @Volatile
-    private var listeners: List<NetworkStateListener> = listOf()
+    private var listeners: Set<NetworkStateListener> = setOf()
```
JcMinarro, line 34: Use `Set` instead List allows us to don't have the same instance twices

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -56,8 +56,9 @@ internal class NetworkStateProvider(private val connectivityManager: Connectivit
 
     fun subscribe(listener: NetworkStateListener) {
         listeners = listeners + listener
-        connectivityManager.registerNetworkCallback(NetworkRequest.Builder().build(), callback)
-        isRegistered.set(true)
+        if (isRegistered.compareAndSet(false, true)) {
```
JcMinarro, line 59: Only register our callback if we didn't do it previously (And it is registered yet)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -42,7 +42,11 @@ internal class ChatSocketServiceImpl constructor(
     private val eventUiHandler = Handler(Looper.getMainLooper())
     private val healthMonitor = HealthMonitor(
         object : HealthMonitor.HealthCallback {
-            override fun reconnect() = this@ChatSocketServiceImpl.reconnect(connectionConf)
+            override fun reconnect() {
+                if (state is State.DisconnectedTemporarily) {
```
JcMinarro, line 46: `HealthMonitor` should only retry to reconnect if the current state is `DisconnectedTemporarily`, in another case it shouldn't

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -59,7 +63,10 @@ internal class ChatSocketServiceImpl constructor(
         }
 
         override fun onDisconnected() {
-            state = State.NetworkDisconnected
+            healthMonitor.stop()
+            if (state is State.Connected || state is State.Connecting) {
```
JcMinarro, line 67: When `Network` is disconnected, we should only update the `state` if it was `Connected` or `Connecting` into another case, we can keep the previous state

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocketServiceImpl.kt
```diff
@@ -59,7 +63,10 @@ internal class ChatSocketServiceImpl constructor(
         }
 
         override fun onDisconnected() {
-            state = State.NetworkDisconnected
+            healthMonitor.stop()
```
JcMinarro, line 66: As far as we don't have Network Connection, doesn't make sense to have `healthMonitor` working on the background

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/EventsParser.kt
```diff
@@ -44,8 +44,12 @@ internal class EventsParser(
     override fun onClosing(webSocket: WebSocket, code: Int, reason: String) { }
 
     override fun onClosed(webSocket: WebSocket, code: Int, reason: String) {
-        // Treat as failure and reconnect, socket shouldn't be closed by server
-        onFailure(ChatNetworkError.create(ChatErrorCode.SOCKET_CLOSED))
+        if (code == CODE_CLOSE_SOCKET_FROM_CLIENT) {
```
JcMinarro, line 47: If the connection was closed by the client, we doesn't need to forward an error.
On the other hand, because we closed it, we don't one to forward any new event we could receive from backend, that is because we update the flag

ogkuzmin (approved):
LGTM
This shows how we need to align socket logic and write integration tests for this layer

adasiewiczr (approved):
Great job!

PR #2277, opened 2021-08-26
[CAS-1240] Feature/release snapshot versions

JcMinarro:
### 🎯 Goal
Create an Snapshot version for every commits arrives to the `develop` branch

### 🛠 Implementation details
The Maven-Publish plugin we use does almost al the needed job for us. We only need to sufix the version with the world `-SNAPSHOT` and this version wouldn't be pusblished to Maven Cental but to the Maven Central Snapshot Repository.

### 🧪 Testing
You can check new snapshots version has been published [here](https://oss.sonatype.org/content/repositories/snapshots/io/getstream/) 

### ☑️ Checklist

- [x] Reviewers added


buildSrc/src/main/kotlin/io/getstream/chat/android/Configuration.kt
```diff
@@ -4,6 +4,10 @@ object Configuration {
     const val compileSdkVersion = 30
     const val targetSdkVersion = 30
     const val minSdkVersion = 21
-    const val versionName = "4.17.0"
+    const val majorVersion = 4
+    const val minorVersion = 17
+    const val pathVersion = 0
```
zsmb13, line 9:
```suggestion
    const val patchVersion = 0
```

Typo

scripts/publish-root.gradle
```diff
@@ -29,6 +39,8 @@ nexusPublishing {
             stagingProfileId = sonatypeStagingProfileId
             username = ossrhUsername
             password = ossrhPassword
+            group = Configuration.artifactGroup
```
zsmb13, line 42: We should remove this either here or in each module - I prefer it living in the module level config

stream-chat-android-compose/build.gradle
```diff
@@ -8,7 +8,11 @@ apply plugin: 'org.jlleitschuh.gradle.ktlint'
 
 ext {
     PUBLISH_GROUP_ID = Configuration.artifactGroup
-    PUBLISH_VERSION = Configuration.versionName + "-beta"
+    if (snapshot) {
+        PUBLISH_VERSION = Configuration.snapshotVersionName
+    } else {
+        PUBLISH_VERSION = Configuration.versionName + "-beta"
+    }
```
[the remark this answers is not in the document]
filbabic: @JcMinarro Yeah that makes sense - We'll plan for this!

PR #2278, opened 2021-08-26
[CAS-1130] Cleanup Channel's and User's extensions

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-1130

### 🎯 Goal

Simplify getting User's and Channel's `name` and `image`

### 🛠 Implementation details

Added `image` and `name` virtual properties to `Channel` and `User`.
`name` and `image` are still being stored inside the `extraData` but I introduced new properties so we don't have to add an additional import for extensions.
Creating `image` and `name` as real properties would be a breaking change. Consider the following code:
```
        val id = "id"
        val type = "messaging"
        val channel = Channel(id =id, type =type).apply { 
            name = "name"
            image = "image"
        }
        client.createChannel(id, type, channel.extraData)
```
We don't pass the whole `Channel` object when creating or updating the channel so the code snippet above won't be working if we convert those extensions into properties.
Moreover, the same applies if someone is using  `extraData["name"]` directly.

### 🧪 Testing

1. Check if images and names are displayed properly in the sample app
2. Try to create a group channel with a name and see if it works correctly.

### ☑️ Checklist

- [x] Reviewers added


bychkovdmitry (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Extensions.kt
```diff
@@ -28,28 +28,44 @@ public fun Channel.getUnreadMessagesCount(forUserId: String = ""): Int {
     }
 }
 
+@Deprecated(
+    message = "Use class member instead",
+    level = DeprecationLevel.WARNING,
+)
```
zsmb13, line 34: We can make this an `ERROR` immediately, as the member property will always shadow the extension of the same name. With the member being added, all of the extension imports will simply become unused, and the migration is just to remove them.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Extensions.kt
```diff
@@ -28,28 +28,44 @@ public fun Channel.getUnreadMessagesCount(forUserId: String = ""): Int {
...
 public var User.image: String
-    get() = getExternalField(this, EXTRA_IMAGE)
+    get() = image
```
zsmb13, line 36: These properties can now be simplified to 

```kotlin
public var User.image: String by User::image
```

Not that it matters, as they can't be invoked anymore anyway (except through Java static calls via `ContentUtils`).

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/User.kt
```diff
@@ -63,4 +63,16 @@ public data class User(
...
 
-) : CustomObject
+) : CustomObject {
+    var name: String
+        get() = getExternalField(this, EXTRA_NAME)
+        set(value) {
+            extraData[EXTRA_NAME] = value
+        }
+
+    var image: String
```
zsmb13, line 73: Have we created a followup task to make these real properties and move them into the constructor? (That will also remove the values from extraData, which is very breaking)

JcMinarro: I'm not sure we will be able to remove it completely, check this [thread](https://github.com/GetStream/stream-chat-android/pull/2278#discussion_r696788612)

zsmb13: We _can_ remove it, it will just break the code of anyone accessing these values through `extraData`, so we'll have to do it in a major version or something

adasiewiczr: Thanks for reminding me!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Extensions.kt
adasiewiczr: makes sense. I'll change it after the release

zsmb13 (approved):

PR #2280, opened 2021-08-26
[Compose] - 2218 adjust scroll to bottom logic and UI

### LINKED ISSUE #2218: [Compose] - Adjust scroll-to-bottom button to match design
We currently have a placeholder implementation for the scroll-to-bottom button in the message list
image

Replace this with a FAB as per the design.

filbabic:
Fixes #2218 

### 🎯 Goal

We needed to update the UI of the "scroll to bottom" component, as well as add logic to show the unread message count to the user.

### 🛠 Implementation details

- Change the UI of the component
- Added logic which compares currently visible messages and keeps the last seen message state in the ViewModel
- Added logic which calculates how many messages the user hasn't "read"

### 🎨 UI Changes

- Changed the UI of the "Scroll to bottom" button
- Updated "my messages" to match the design
- Fixed an issue with the message list flickering when sending a new message (everything re-drawing)

| Before | After |
| --- | --- |
| before | after |

### 🧪 Testing

- Open any Channel and scroll away from the bottom.
- Try sending a message with another user, to the same Channel.

### ☑️ Checklist

~~- [ ] Affected documentation updated (docusaurus, tutorial, CMS (task created))~~ (I'll update the docs as I go through the tutorial - the UI change affects many screenshots, so it's easier if I do it in one go!
- [x] Reviewers added

### 🎉 GIF

scroll


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
```diff
@@ -202,19 +209,22 @@ public fun Messages(
             }
         }
 
+        val firstVisibleItemIndex = currentListState.firstVisibleItemIndex
+
         when {
             newMessageState == MyOwn -> coroutineScope.launch {
-                currentListState.scrollToItem(5)
+                if (firstVisibleItemIndex > 5) {
```
aldefy, line 216: can this `5` be customizable ? 

Since this could be internal SDK thing but was wondering if there is a reason for having such a constant declared as is.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
```diff
@@ -223,26 +233,56 @@ public fun Messages(
...
+        }
+
+        if (unreadCount != 0) {
+            Surface(
+                modifier = Modifier
+                    .align(TopCenter),
+                shape = RoundedCornerShape(16.dp),
+                color = ChatTheme.colors.primaryAccent
+            ) {
+                Text(
```
aldefy, line 279: [Query] Any reason to wrap it inside a Surface or to get BG color as per L#277

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -236,6 +252,93 @@ public class MessageListViewModel(
...
 
+    /**
+     * Counts how many messages the user hasn't read already. This is based on what the last message they've seen is,
+     * and the current message state.
+     *
+     * @param newMessageState - The state that tells us if there are new messages in the list.
+     * @return - [Int] which describes how many messages come after the last message we've seen in the list.
+     * */
+    private fun getUnreadMessageCount(newMessageState: NewMessageState? = currentMessagesState.newMessageState): Int {
+        if (newMessageState == null || newMessageState == MyOwn) return 0
```
aldefy, line 263: since 0 is hardcoded , i feel we should declare const  immutable variable and use that, gets easy on code review and further logic changes to other parts of code in this function.

Please ignore if you feel otherwise.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -236,6 +252,93 @@ public class MessageListViewModel(
...
+
+        val messageItems = currentMessagesState.messageItems
+        val lastSeenMessagePosition =
+            getLastSeenMessagePosition(if (isInThread) lastSeenThreadMessage else lastSeenChannelMessage)
+        var unreadCount = 0
```
aldefy, line 268: . to const variable ^

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -236,6 +252,93 @@ public class MessageListViewModel(
...
+    }
+
+    /**
+     * Gets the list position of the last seen message in the list.
+     *
+     * @param lastSeenMessage - The last message we saw in the list.
+     * @return - [Int] list position of the last message we've seen.
+     * */
+    private fun getLastSeenMessagePosition(lastSeenMessage: MessageItem?): Int {
+        if (lastSeenMessage == null) return 0
```
aldefy, line 288: +1 here const imo for lastSeenMessagePosition = 0 can be declared

Please ignore this comment if you feel otherwise

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -236,6 +252,93 @@ public class MessageListViewModel(
...
+     * */
+    public fun updateLastSeenMessage(currentMessage: MessageItem) {
+        val lastSeenMessage = if (isInThread) lastSeenThreadMessage else lastSeenChannelMessage
+
+        if (lastSeenMessage == null) {
+            updateLastSeenMessageState(currentMessage)
+            return
+        }
+
+        val lastSeenMessageDate = lastSeenMessage.message.createdAt ?: Date()
```
aldefy, line 309: I have suggestion / enhancement

have a typealias for Date object  

As follows :  
_pseudo code from a scratch file_
```kotlin
import java.util.Date

typealias StreamDate = Date

data class MessageItem(
    val message : Message
)

data class Message(
    val createdAt : StreamDate?
)

lateinit var lastSeenMessage : MessageItem
val lastSeenMessageDate = lastSeenMessage.message.createdAt ?: Date()

// Suggestion
val lastSeenMessageDate = lastSeenMessage.message.createdAt ?: StreamDate()


// tomorrow we could
class SomeFancyDateLibClass {
    // Date logic
}
// go update 
typealias StreamDate = SomeFancyDateLibClass

```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -236,6 +252,93 @@ public class MessageListViewModel(
...
+
+    /**
+     * Updates the state of the last seen message. Based on if we're [isInThread] or not, it updates corresponding state.
+     *
+     * @param currentMessage - The current message the user sees.
+     * */
+    private fun updateLastSeenMessageState(currentMessage: MessageItem) {
+        if (isInThread) {
+            lastSeenThreadMessage = currentMessage
+
```
aldefy, line 333: blank line for readability ?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -402,18 +504,13 @@ public class MessageListViewModel(
...
+            MessageItem(
+                message = message,
+                groupPosition = position,
+                parentMessageId = parentMessageId,
+                isMine = user.id == currentUser?.id
             )
-        }
-
-        return items
+        }.reversed()
```
aldefy, line 513: [Query] probably more curious why is it reversed ?

aldefy (commented):
Minor questions to get SDK knowledge on these changes + few suggestions and few open questions 

feel free to resolve comments and reply offline

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
aldefy: Pros : we simply update one typealias , entire code where we have Date imports gets reduced to one typealias

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
filbabic: This should actually be changed to some form of offset. This was put in initially as a pseudo-random number, but will be updated!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
filbabic: Used a surface to make it a bit easier to allow for padding and a shape that wraps the text. We could add lots of these properties to the `Text` but it felt like it'd be too hard to follow the order of the modifiers.

LMK if this makes sense!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
filbabic: Honestly I don't think we should hardcoded a constant in this case, especially because it seems pretty clear what the 0 represents.

We're fetching the "unreadMessageCount", so if we return zero, then it's reasonable that there are no messages that are unread, right?

It is a pretty straightforward piece of logic in this case and I don't think a constant named e.g. `NO_UNREAD_MESSAGES` would make much of a difference.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
filbabic: As per the comments above, I think this is another case where it's pretty clear what we're doing. If there is no lastSeenMessage, then its position is `0`, or we're at the start of the list.

We could return -1 here or something potentially, but that would require extra checks for that specific case.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
filbabic: To me that is very against the open-closed principle. A typealias is just a way to name a type differently, for more clarity, or cleaner syntax.

If we take this approach and we change the `typealias StreamDate = Date` to e.g. a date from the ThreeTen lib, all the underlying code/calling code would be broken as we'd have to completely change the call sites.

If we were to abstract it behind an interface, that would make more sense, but our parsers are using Date currently and we'd have to change the models too.

LMK if this makes sense or if you have other suggestions!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
filbabic: yup!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
filbabic: We show message items in the reverse order, because we're reversing the message list too. This is the base way of showing a Chat, where the newest messages have to be at the bottom, rather than the top.

PR #2285, opened 2021-08-30
Fixing mute channel refresh

leandroBorgesFerreira:
Reverts GetStream/stream-chat-android#2219

## Explanation

We should not revert the extraData `isMuted`. Without this information in our `Channel` class, our diff never changes what causes our list of muted channel to not be refreshed when a mutes/unmutes a `Channel`. 

Although it works when the list is fully refreshed (user reloads `ChannelListView`'s activity), the adapter doesn't responds to just a change in the muted channels because it never sees a change in its data set, then no item of the adapter is reloaded and then this code never gets called: 

```
muteIcon.isVisible = ChatDomain.instance().user.value?.let(channel::isMutedFor) == true
```
 
We need to keep the extra data so the adapter knows when a change happened in the visibility of a channel


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -156,6 +160,17 @@ public class ChannelListViewModel(
 
     public data class State(val isLoading: Boolean, val channels: List<Channel>)
 
+    private fun parseMutedChannels(
+        channelsMap: List<Channel>,
+        channelMutesIds: List<String>?,
+    ): List<Channel> {
+        return channelsMap.map { channel ->
+            channel.copy().apply {
+                isMuted = channelMutesIds?.contains(channel.id) ?: false
```
ogkuzmin, line 169: We can reuse our extension `Channel::isMutedFor`, can't we?

ogkuzmin (commented):
Actually, I like this (old) approach better, since there is less logic in VH

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
leandroBorgesFerreira: We can, but in this case, we would need referential transparency because we would need the `User` class. Here we already have the list of muted channels.

ogkuzmin: But we also have  a user value there. Look where this method is invoked

ogkuzmin: Idea is simple. There is no sense having the same logic twice in our SDK. Here and as an extention

leandroBorgesFerreira: This method is called twice, one of them with currentUser available, but the other is not... I can call it with `ChatDomain.instance().user.value` but then I would be inserting nullability in the method =|. I just saw `channelMutesIds: List<String>,` doesn't need to be nullable

leandroBorgesFerreira: But... now taking a second look... maybe we don't need this at all:
```
stateMerger.addSource(queryChannelsController.mutedChannelIds.asLiveData())
```

Let me check this.

leandroBorgesFerreira: No, we still need to listen for the `mutedChannelIds` changes.

leandroBorgesFerreira: It would be also a bit odd that we listen for `mutedChannelIds`,  but we ignore its data... `User.channelMutes` can be out of sync with `QueryChannelsController.mutedChannelIds` for a super small period of time, but that would create a subtle bug in this method.

ogkuzmin (approved):
Yes, it fixes the problem.
But I also revealed another one. We have two sources of the same data currently. 
`User::channelMutes` and `QueryChannelsController::mutedChannelIds` means the same. IMO `QueryChannelsController::mutedChannelIds`is redundant source. We should update user values with latest values when mutes also got changed. I will add this to our backlog

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -156,6 +161,36 @@ public class ChannelListViewModel(
...
+                        extraData = channel.extraData.clone(EXTRA_DATA_MUTED, false)
+                    )
+
+                !channel.isMuted && channelMutesIds.contains(channel.id) ->
+                    channel.copy(
+                        extraData = channel.extraData.clone(EXTRA_DATA_MUTED, true)
+                    )
+
+                else -> channel
+            }
```
ogkuzmin, line 181:
```suggestion
  when {
                channel.isMuted != channelMutesIds.contains(channel.id) ->
                    channel.copy(extraData = channel.extraData.clone(EXTRA_DATA_MUTED, false))
                else -> channel
            }
```

ogkuzmin (approved):

PR #2294, opened 2021-08-31
[CAS-1201] Add support for new pagination

bychkovdmitry:
https://stream-io.atlassian.net/browse/CAS-1201

### 🎯 Goal

Add a new way to paginate through search message results using limit and next/previous values.

### 🛠 Implementation details

Using request parameters directly instead of encapsulating them in `SearchMessagesRequest`. Motivation:

1. Easier to deprecate the old method. If `SearchMessagesRequest` class were used, then the old method and the new one would be conflicting as only difference would be the result payload.
1. Consistent with what we already have. Most of the methods in `ChatClient` already use individual parameters instead of wrapper class.
1. Easier to document each parameter.

### 🎨 UI Changes

Screenshot 2021-09-01 at 03 18 49

### 🧪 Testing

1. Create an app with the new pagination enabled on the backend
2. Send 15 "Hi" messages and wait until they are properly indexed
3. Search for the messages containing "Hi" with the following query:
```kotlin
val firstPage = ChatClient.instance()
    .searchMessages(
        channelFilter = channelFilter,
        messageFilter = messageFilter,
        limit = 10,
        sort = QuerySort.desc("updated_at"),
    )
    .await()
```
4. Verify that the first chunk of messages has been loaded
5. Load the next page by using the following query:
```kotlin
ChatClient.instance()
    .searchMessages(
        channelFilter = channelFilter,
        messageFilter = messageFilter,
        limit = 10,
        next = firstPage.data().next, // the next page from previous query
    )
    .await()
```
6. Verify that the next 5 messages were loaded

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (8)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -715,11 +716,56 @@ public class ChatClient internal constructor(
...
+        return api.searchMessages(
+            channelFilter = channelFilter,
+            messageFilter = messageFilter,
+            offset = offset,
+            limit = limit,
+            next = next,
+            sort = sort
+        )
+    }
+
```
bychkovdmitry, line 768: Not a huge fan of placing `@JvmOverloads` everywhere

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/SearchMessagesRequest.kt
```diff
@@ -1,12 +1,42 @@
 package io.getstream.chat.android.client.api.models
 
 import com.google.gson.annotations.SerializedName
+import io.getstream.chat.android.client.models.Message
```
bychkovdmitry, line 4: One of the few "public" request classes.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/SearchWarning.kt
```diff
@@ -0,0 +1,23 @@
+package io.getstream.chat.android.client.models
+
```
bychkovdmitry, line 2: For example, if there are more than 500 channels that match the channel filter.

bychkovdmitry:
Merged develop and removed gson related entites. 👍  

Not sure about:
- Wrapped input parameters (`SearchMessagesRequest`) vs individual input parameters
- The naming of the result object: `SearchMessagesResult` vs `SearchMessagesResponse`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -715,11 +716,56 @@ public class ChatClient internal constructor(
...
     }
 
+    /**
+     * Search messages across channels. There are two ways to paginate through search results:
+     *
+     * * Using [limit] and [offset] parameters
+     * * Using [limit] and [next] parameters
+     *
+     * Pages of sort results will be returned with **next** and **previous** strings which
+     * can be supplied as a next parameter when making a query to get a new page of results.
```
adasiewiczr, line 736: It would be nice to mention when we can use `sort` parameter

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -715,11 +716,56 @@ public class ChatClient internal constructor(
...
+        if (offset != null && (sort != null || next != null)) {
+            return ErrorCall(ChatError("Cannot specify offset with sort or next parameters"))
+        }
+        return api.searchMessages(
+            channelFilter = channelFilter,
+            messageFilter = messageFilter,
+            offset = offset,
+            limit = limit,
+            next = next,
+            sort = sort
```
adasiewiczr, line 765:
```suggestion
            sort = sort,
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -715,11 +716,56 @@ public class ChatClient internal constructor(
...
+     *
+     * Pages of sort results will be returned with **next** and **previous** strings which
+     * can be supplied as a next parameter when making a query to get a new page of results.
+     *
+     * @param channelFilter channel filter conditions
+     * @param messageFilter message filter conditions
+     * @param offset pagination offset, cannot be used with sort or next
+     * @param limit the number of messages to return
+     * @param next pagination parameter, cannot be used with non-zero offset
+     * @param sort sort parameters, cannot be used with non-zero offset
```
adasiewiczr, line 743: This looks a little bit weird with the double `sort` word. Maybe we can add `-` after the parameter name? What do you think?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -692,10 +693,12 @@ internal class MoshiChatApi(
...
     override fun searchMessages(request: SearchMessagesRequest): Call<List<Message>> {
         val newRequest = io.getstream.chat.android.client.api2.model.requests.SearchMessagesRequest(
-            offset = request.offset,
-            limit = request.limit,
             filter_conditions = request.channelFilter.toMap(),
             message_filter_conditions = request.messageFilter.toMap(),
+            offset = request.offset,
+            limit = request.limit,
+            next = request.next,
+            sort = request.sort
```
adasiewiczr, line 701:
```suggestion
            sort = request.sort,
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/SearchWarning.kt
```diff
@@ -0,0 +1,23 @@
...
+public data class SearchWarning(
```
adasiewiczr, line 3: It would be nice to add KDoc here

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
bychkovdmitry: We have an internal doc where Marton explains why are no longer using dashes in such cases. I paraphrased the sentence a bit.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
bychkovdmitry: Nice catch!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
zsmb13: Yes, please don't add a dash there - it creates a weird looking list item as it's interpreted as Markdown.

image

zsmb13: On another note, @bychkovdmitry, can we please have full sentences for each of these comments, to adhere to the aforementioned guidelines? E.g.

```
     * @param sort The sort criteria applied to the result, cannot be used with non-zero offset.
```

bychkovdmitry: Yes, just a sec. This PR was created before the guidelines :)

adasiewiczr (approved):

PR #2298, opened 2021-09-01
[CAS-623] Remove Gson for good

zsmb13:
https://stream-io.atlassian.net/browse/CAS-623

### 🎯 Goal

The last step of our deprecation of Gson and migration to Moshi, its complete removal from our codebase.

### 🛠 Implementation details

See commit names for some details.

### 🧪 Testing

Smoke test sample apps, run unit tests.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/ClientConnectionTests.kt
```diff
@@ -84,19 +82,11 @@ internal class ClientConnectionTests {
...
         notificationsManager = mock()
         initCallback = mock()
-        api = GsonChatApi(
-            retrofitApi,
-            retrofitAnonymousApi,
-            UuidGeneratorImpl(),
-            fileUploader,
-            testCoroutines.scope
-        )
+        api = mock()
```
zsmb13, line 89: Mocking the entire `ChatApi` now instead of its contained Retrofit API interfaces, this way we don't have to create dummy DTOs and deal with mapping in our tests which would be very, very tedious

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser/FilterObjectTypeAdapterTest.kt
```diff
@@ -1,275 +0,0 @@
...
-import io.getstream.chat.android.test.randomBoolean
-import io.getstream.chat.android.test.randomInt
-import io.getstream.chat.android.test.randomString
-import org.junit.Assert
-import org.junit.jupiter.params.ParameterizedTest
-import org.junit.jupiter.params.provider.Arguments
-import org.junit.jupiter.params.provider.MethodSource
-import java.io.StringWriter
-
-internal class FilterObjectTypeAdapterTest {
```
JcMinarro, line 33: Shouldn't we use those tests with MoshiParser?

bychkovdmitry (changes requested):
Do we still need old API models like:
- io.getstream.chat.android.client.api.models.FlagResponse
- io.getstream.chat.android.client.api.models.FlagResponseUpdateUsersResponse
- many more...

They seem to be unused

bychkovdmitry:
What about moving  everything from `api2` to `api`. Of course if everything inside is internal.

zsmb13:
@bychkovdmitry Good call on both of those things!

>Do we still need old API models

Nope! I've now removed all the unused ones.

> moving everything from api2 to api

This can't be done at the moment, `api/models` still has things in it that can't be removed. Some of it is public API, some of it is for internal use (like in file uploading). In comparison, everything in `api2/models` are actual DTOs that get sent over the wire. I really, really don't want to mix these two (three) types of things together (like they were in the old implementation).

We should figure out how to get rid of all the still public things in `api/models` and then we could clean up the odd `api2` naming.

JcMinarro (approved):
Good to see we removed Gson!!!! :partying_face: :tada:

bychkovdmitry (approved):
Epic moment

PR #2300, opened 2021-09-01
Adding a constant size for giphy

leandroBorgesFerreira:
### 🎯 Goal

Fix glitch in scroll with many gifs in message list. 

https://user-images.githubusercontent.com/10619102/131756064-0d4022cf-e9d6-4a33-a2f3-54edf99e117a.mov

### 🛠 Implementation details

Now gifs have the same size and they are cropped, just like RN and Flutter SDKs do. 

### 🎨 UI Changes

The new scroll: 

https://user-images.githubusercontent.com/10619102/131756118-37be3634-9959-491c-b888-67346441a1c4.mov

### 🧪 Testing

Send many giphys and check the scroll with and without this PR. 
### ☑️ Checklist

- [x] Reviewers added. 

### 🎉 GIF

giphy


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -71,6 +71,14 @@ internal class MediaAttachmentView : ConstraintLayout {
             }
         }
 
+        if (attachment.type == ModelType.attach_giphy) {
+            val params = binding.imageView.layoutParams
+
+            binding.imageView.layoutParams = params.apply {
+                height = 200.dpToPx()
```
adasiewiczr, line 78: What do you think about making it customizable? Does it make sense?

leandroBorgesFerreira: I think that making it customizable makes total sense. But I think it should be only height so we don't conflict with the `MaxPossibleWidthDecorator` all messages are limited in width by this decorator and it would be confusing to make only this view as an exception.

leandroBorgesFerreira: About the fixed height of all the attachments. The pictures don't cause a glitch in our scroll because we don't need to load then using coil (or any image loader), so I would say that allow then to change its height accordingly with the picture is fine. What do you think?

adasiewiczr: I think we can keep the old logic for other attachments for now

adasiewiczr (approved):

bychkovdmitry (approved):

Issue #2232, opened 2021-08-17
[Compose] - Improve Attachments behavior


filbabic:
We have a few comments from Marton's testing, as well as things we've noticed so far.

- [x] Add attachment upload progress indicators
- [x] Preload images based on local file
- [x] Attachments picker can be opened without any mode selected
- ~~[ ] Download & open attachments when tapped (images)~~ This will be covered in the ImagePreviewActivity improvements

PR #2310, opened 2021-09-06
2232 - Image attachment improvements

filbabic:
Implements most of the changes from #2232 

### 🎯 Goal

We need to polish the look & feel of uploading images, to show indicators and pre-load items from the file system. This will give our users more information about what's happening with their messages.

### 🛠 Implementation details

Added a separate component that represents files being uploaded. Re-organized attachments files and API to group them more logically.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| N/A - We didn't have any progress indicators before. | uploading |

### 🧪 Testing

Attempt to upload a new image/file. You should see the new content with upload indicators.

### ☑️ Checklist

~~- [ ] Affected documentation updated (docusaurus, tutorial, CMS (task created))~~Will upload tutorial once a new version is out.
- [x] Reviewers added

### 🎉 GIF

upload


stream-chat-android-compose/api/stream-chat-android-compose.api
```diff
@@ -179,7 +179,7 @@ public final class io/getstream/chat/android/compose/state/messages/attachments/
 	public final fun copy (Landroidx/compose/ui/Modifier;Lio/getstream/chat/android/compose/state/messages/items/MessageItem;Lkotlin/jvm/functions/Function1;)Lio/getstream/chat/android/compose/state/messages/attachments/AttachmentState;
 	public static synthetic fun copy$default (Lio/getstream/chat/android/compose/state/messages/attachments/AttachmentState;Landroidx/compose/ui/Modifier;Lio/getstream/chat/android/compose/state/messages/items/MessageItem;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lio/getstream/chat/android/compose/state/messages/attachments/AttachmentState;
 	public fun equals (Ljava/lang/Object;)Z
-	public final fun getMessage ()Lio/getstream/chat/android/compose/state/messages/items/MessageItem;
+	public final fun getMessageItem ()Lio/getstream/chat/android/compose/state/messages/items/MessageItem;
```
zsmb13, line 182: Might be worth adding an entry into the changelog

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -557,7 +596,7 @@ private fun DeletedMessageFooter(
 
         Timestamp(
             modifier = Modifier.padding(8.dp),
-            date = message.updatedAt ?: message.createdAt
+            date = message.updatedAt ?: message.createdAt ?: Date()
```
zsmb13, line 599: Do we really need this `Date()` here?

stream-chat-android-compose/api/stream-chat-android-compose.api
filbabic: Good point!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
filbabic: In this case I put it in yeah - because when we upload images, the message _exists_ but doesn't really. There is no `updatedAt` or `createdAt` date, but we need to show a timestamp.

It's a weird case, which is why I put it just in here, instead of in the `Timestamp` for example.

zsmb13 (approved):
Ship it

PR #2319, opened 2021-09-08
Rename getUsers to getUsersExcludingCurrent

adasiewiczr:
### 🎯 Goal

Rename `getUsers` to more verbose `getUsersExcludingCurrent`

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/extensions/Channel.kt
```diff
@@ -1,11 +1,11 @@
...
 
-import io.getstream.chat.android.client.extensions.getUsers
+import io.getstream.chat.android.client.extensions.getUsersExcludingCurrent
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import io.getstream.chat.android.offline.ChatDomain
 
 @InternalStreamChatApi
-public fun Channel.isDirectMessaging(currentUserId: String = currentUserId()): Boolean = getUsers(currentUserId).size == 1
+public fun Channel.isDirectMessaging(currentUserId: String = currentUserId()): Boolean = getUsersExcludingCurrent(currentUserId).size == 1
```
zsmb13, line 9: Looks like this is the only place where we pass in a `currentUserId` to the method being renamed. And then the `isDirectMessaging` function is almost never called with a parameter either (just once as far as I see, and I don't know if even that is necessary. Let's see if we can eliminate these params if they're not actually needed somewhere.

adasiewiczr: We are using it inside `MessageInputView` to be sure it's initialized correctly if the view was created before a user is set. Let me check if can somehow improve that

zsmb13: Maybe we can check for that inside `getUsersExcludingCurrent`, if we don't have a user, just don't filter out anything?

adasiewiczr: 👍 makes sense. Then reinitialize the view if the user was set

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChannelExtension.kt
```diff
@@ -15,5 +15,5 @@ public fun Channel.isAnonymousChannel(): Boolean = id.isAnonymousChannelId()
 public fun Channel.isMutedFor(user: User): Boolean = user.channelMutes.any { mute -> mute.channel.cid == cid }
 
 @InternalStreamChatApi
-public fun Channel.getUsersExcludingCurrent(excludeUserId: String = ChatClient.instance().getCurrentUser()?.id ?: ""): List<User> =
-    members.map { it.user }.filterNot { it.id == excludeUserId }
+public fun Channel.getUsersExcludingCurrent(): List<User> =
+    members.map { it.user }.filterNot { it.id == ChatClient.instance().getCurrentUser()?.id ?: "" }
```
zsmb13, line 19: We do a bunch of comparisons to the empty string here in some cases, which is not something we want to do. How about this instead?

```suggestion
public fun Channel.getUsersExcludingCurrent(): List<User> {
    val users = members.map { it.user }

    val currentUserId = ChatClient.instance().getCurrentUser()?.id
    return if (currentUserId != null) {
        users.filterNot { it.id == currentUserId }
    } else {
        users
    }
}
```

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/extensions/Channel.kt
```diff
@@ -1,11 +1,8 @@
...
+import io.getstream.chat.android.client.extensions.getUsersExcludingCurrent
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
-import io.getstream.chat.android.offline.ChatDomain
 
 @InternalStreamChatApi
-public fun Channel.isDirectMessaging(currentUserId: String = currentUserId()): Boolean = getUsers(currentUserId).size == 1
-
-private fun currentUserId(): String = ChatDomain.instance().user.value?.id ?: ""
+public fun Channel.isDirectMessaging(): Boolean = getUsersExcludingCurrent().size == 1
```
zsmb13, line 8: It doesn't seem very efficient to create this list of users with a bunch of operations to just check for its size. How about

```suggestion
public fun Channel.isDirectMessaging(): Boolean {
    return members.size == 2 && includesCurrentUser()
}

private fun Channel.includesCurrentUser(): Boolean {
    val currentUserId = ChatClient.instance().getCurrentUser()?.id ?: return false
    return members.any { it.user.id == currentUserId }
}
```

zsmb13 (approved):

PR #2327, opened 2021-09-09
Use factory functions instead of subclasses for AttachmentFactories

zsmb13:
### 🎯 Goal

Add a workaround for weird compiler crashes related to AttachmentFactory subclasses.

### 🧪 Testing

Run the Compose app. Shouldn't crash on launch.

### ☑️ Checklist

- [x] ~Affected documentation updated (docusaurus, tutorial, CMS (task created))~
- [x] Reviewers added

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
@@ -1,15 +1,13 @@
 package io.getstream.chat.android.compose.ui.attachments
 
 import androidx.compose.runtime.Composable
-import com.getstream.sdk.chat.model.ModelType
 import io.getstream.chat.android.client.models.Attachment
 import io.getstream.chat.android.compose.state.messages.attachments.AttachmentState
-import io.getstream.chat.android.compose.ui.attachments.content.FileAttachmentContent
-import io.getstream.chat.android.compose.ui.attachments.content.GiphyAttachmentContent
-import io.getstream.chat.android.compose.ui.attachments.content.ImageAttachmentContent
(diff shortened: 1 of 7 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
```diff
@@ -39,7 +40,7 @@ public object StreamAttachmentFactories {
  * @param content - Composable function that allows users to define the content the [AttachmentFactory] will build using any given
  * [AttachmentState].
  */
-public abstract class AttachmentFactory(
+public open class AttachmentFactory @InternalStreamChatApi constructor(
```
filbabic, line 43: Do we need to have this as an `open` class? Since it's not going to be extended. Or do we want to keep it like this, just so that people _could_ extend it if they need to (or once the bug is fixed)

filbabic (commented):
@zsmb13 Can we point this to this [branch](https://github.com/GetStream/stream-chat-android/pull/2310), or merge the PR there and then clean up any conflicts we might have around this today? 

LMK if you want me to do it, it's just a matter of getting the PR approved and merged first!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
zsmb13: Yeah, I was thinking that this way we don't break anyone's code if they've already subclassed it (but we make it scary to subclass with the annotation)

filbabic: Works for me!

filbabic (commented):
Looks good!

PR #2337, opened 2021-09-13
[CAS-1305] Feature - Ktlint and unit tests for affected modules

leandroBorgesFerreira:
[CAS-1305](https://stream-io.atlassian.net/browse/CAS-1305)

### 🎯 Goal

Run unit tests and ktlint only for affected modules. 

### 🛠 Implementation details

This uses [dag-command](https://github.com/leandroBorgesFerreira/dag-command) to create the list of affected modules by a PR and then it uses [android-cli](https://github.com/GetStream/android-cli) to create a command for ktlint and unit tests based on the modules that were affected by a PR. 

**A module doesn't have to be directly changed to be affected by a PR, if it depends on something that was changed, it gets affected.**

### 🧪 Testing

Set this branch it to be the `defaultBranch`, and create a small change in one module. Try to run the unit tests only for the affected modules. You can follow this tutorial:

1. Set this branch in the configuration of dag-command:

```
dagCommand {
    filter = "all"
    defaultBranch = "origin/dagTest"
    outputType = "json"
    printModulesInfo = true
}
```
You can also just add a env var `export GITHUB_BASE_REF="develop"` in your machine too. 

2. Change some file in the project.
3. run `./gradlew dag-command`
4. Check that the file `affected-modules.json` represents only the affected modules by your change. 
5. Run `./gradlew selected-test-debug` and check that the command generated look as expected. You should also run it to confirm that it is correct. 

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
```diff
@@ -124,6 +125,7 @@ object Dependencies {
     const val coroutinesTest = "org.jetbrains.kotlinx:kotlinx-coroutines-test:${Versions.COROUTINES}"
     const val constraintLayout = "androidx.constraintlayout:constraintlayout:${Versions.CONSTRAINT_LAYOUT}"
     const val dexter = "com.karumi:dexter:${Versions.DEXTER}"
+    const val dagCommand = "com.github.leandroborgesferreira:dag-command:${Versions.DAG_COMMAND}"
```
leandroBorgesFerreira, line 128: Should we deploy it in our nexus before using it?

leandroBorgesFerreira:
One question is about the usage inside CI... Currently I'm inserting the android-cli jars inside the project. The automation can't be a `script.gradlew` because you can't run a gradle task inside another one.

buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
JcMinarro: As it is not a plugin was developed inside GetStream, I think it is fine to use without deploy into our nexus

build.gradle
```diff
@@ -53,6 +54,15 @@ allprojects {
     }
 }
 
+apply plugin: 'com.github.leandroborgesferreira.dag-command'
+
+dagCommand {
+    filter = "all"
+    defaultBranch = "origin/develop"
```
zsmb13, line 61: What happens on PRs where we're not targeting this branch? It's also odd that it includes `origin`, which doesn't make sense in the CI context (assuming that's the eventual goal, applying this there). People might have different names for the git remote locally as well.

zsmb13 (changes requested):
Can we hook up the Android CLI in some other way, without including the JAR directly? Especially having to commit the dependency JARs into this repo seems really clunky. Can't we move the script itself into the repo easily?

Could be a Kotlin script, could be a shell script, could be a Gradle task. We could use `buildSrc` to contain it. Doesn't seem like it's doing anything complex.

A bit worried in general about the additional custom config & tooling that this PR introduces. Our project is already non-trivial with things like git versioner etc, this adds a couple additional steps of complexity again for both usage and maintanence.

Take for example these lists of modules that live _in another repo_ that you'd potentially have to update if you change the module structure (e.g. create a new one) of the main repo https://github.com/GetStream/android-cli/blob/master/src/main/kotlin/commands/ktlint/KtlintSelection.kt#L25-L36

build.gradle
leandroBorgesFerreira: The PRs that don't target `origin/develop`, could have the normal unit test and Ktlint flow, the idea is to speed up only the merge with `develop` because this is where speed has value (ex: merge with `main` could take longer). About the remote, let's change it to just `develop`, so it doesn't become a problem with different remotes.

leandroBorgesFerreira:
The list of `modules` (optin and optout for ktlint and unit tests) can be just a parameter (like a file containing modules for ktlint and for unit tets) that lives inside this repository. Or we can parse which modules contain unit tests and which contain `ktlint` plugin applies (the second option is probably more robust).

The problem with `buildSrc` or `gradle` tasks is that one can't call another `gradle` task from those. One idea would be to create tasks that generate `.sh` with the gradle command and the CI executes it after generating it. What do you think?

leandroBorgesFerreira:
Now the commands are inside `buildSrc` and a command is generated in a `.sh` file. Then the CI can run this command.

buildSrc/src/main/kotlin/io/getstream/chat/android/command/utils/Modules.kt
```diff
@@ -0,0 +1,19 @@
+package io.getstream.chat.android.command.dag
+
+import com.squareup.moshi.Types
+import io.getstream.chat.android.command.utils.moshi
+import java.io.File
+
+fun parseModules(modulesFile: File): List<String> {
+    val json = modulesFile.readLines().joinToString()
+    return moshi()
+        .adapter<List<String>>(Types.newParameterizedType(List::class.java, String::class.java))
```
zsmb13, line 10:
```suggestion
        .adapter<List<String>>()
```

This works on its own if you `import com.squareup.moshi.adapter` and add `@OptIn(ExperimentalStdlibApi::class)`

buildSrc/src/main/kotlin/io/getstream/chat/android/command/utils/Tasks.kt
```diff
@@ -0,0 +1,12 @@
...
+import org.gradle.api.Action
+import org.gradle.api.Task
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.api.tasks.TaskProvider
+import org.gradle.kotlin.dsl.register
+
+inline fun <reified T : Task> TaskContainer.registerExt(
+    name: String,
+    configuration: Action<in T>,
+): TaskProvider<T> = this.register(name, T::class.java, configuration)
```
zsmb13, line 12: Neat!

buildSrc/src/main/kotlin/io/getstream/chat/android/command/dag/unittest/task/SelectedUnitTestsTask.kt
```diff
@@ -0,0 +1,35 @@
...
+
+        if (!changedModulesFile.exists()) {
+            changeModuleFileDoesNotExistInPath(config.changeModulesPath)
+        }
+
+        val command = parseModules(changedModulesFile).unitTestCommand(project)
+
+        writeFile(config.outputPath) { writer ->
+            writer.write(command)
+        }
```
zsmb13, line 30:
```suggestion
        File(config.outputPath).writeText(command)
```

Just some simplification again

zsmb13 (commented):
Really nice job moving this in-repo!! This is cool. It's easy to find, and easy to understand / maintain later on.

I left a bunch of comments but they're pretty much just small things about code style.

buildSrc/src/main/kotlin/io/getstream/chat/android/command/utils/Modules.kt
```diff
@@ -0,0 +1,19 @@
...
+        .fromJson(json)
+        ?: emptyList()
+}
+
+fun List<String>.generateGradleCommand(commandFunction: (String) -> String): String {
```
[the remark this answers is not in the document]
leandroBorgesFerreira: Some modules may have `testDebugUnitTest` and others may have only `testUnitTest` depending if they are a just a java library or an android library.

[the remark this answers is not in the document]
leandroBorgesFerreira: Which is the case of `stream-chat-android-code` I need to fix that, by the way.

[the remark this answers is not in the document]
zsmb13: Maybe we need to pass in the entire Project object then

buildSrc/build.gradle.kts
```diff
@@ -8,4 +8,5 @@ repositories {
 
 dependencies {
     implementation("com.squareup.moshi:moshi:1.12.0")
+    implementation("com.squareup.moshi:moshi-kotlin:1.12.0")
```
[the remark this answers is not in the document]
leandroBorgesFerreira: I kept the second one.

buildSrc/src/main/kotlin/io/getstream/chat/android/command/utils/Modules.kt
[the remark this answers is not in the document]
leandroBorgesFerreira: I used a generic type, so it gets more generic (haha).

zsmb13 (approved):
I can't try this locally atm because I don't have a shell environment where I could run Gradle tasks, but everything looks reasonable for me now - someone else please test it and give an additional approval 😃

PR #2343, opened 2021-09-14
[CAS-1317] Feature - Adding hard delete

leandroBorgesFerreira:
[CAS-1317](https://stream-io.atlassian.net/browse/CAS-1317)

### 🎯 Goal

Adding hard delete to ChatClient. If the flag is true, the message is deleted in the backend and the action can't be undone. 

### 🛠 Implementation details

The flag was added in the methods in the flow of message deletion with the default value as `false`

### 🧪 Testing

Set the flag to `true` and check that the messages gets removed instead of simplying showing "message deleted" when you clear the storage.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomain.kt
```diff
@@ -401,7 +401,7 @@ public sealed interface ChatDomain {
      * @see io.getstream.chat.android.offline.utils.RetryPolicy
      */
     @CheckResult
-    public fun deleteMessage(message: Message): Call<Message>
+    public fun deleteMessage(message: Message, hard: Boolean = false): Call<Message>
```
bychkovdmitry, line 404: Would be nice to update the KDoc above

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -519,11 +519,11 @@ public class MessageListViewModel(
      *
      * @param message Message to delete.
      */
-    public fun deleteMessage(message: Message) {
+    public fun deleteMessage(message: Message, hard: Boolean = false) {
```
bychkovdmitry, line 522: Same, KDoc above

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -872,8 +872,8 @@ public class ChatClient internal constructor(
     }
 
     @CheckResult
-    public fun deleteMessage(messageId: String): Call<Message> {
-        return api.deleteMessage(messageId)
+    public fun deleteMessage(messageId: String, hard: Boolean = false): Call<Message> {
```
bychkovdmitry, line 875: Sometimes @JvmOverloads produces lots of overloads and it feels messy to put it. But here I think it is fine. Only 1 overload to make the life of Java users easier.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -872,8 +872,8 @@ public class ChatClient internal constructor(
     }
```
bychkovdmitry, line 873: Did you verify that this param works as expected? From our docs I see:

```
// hard delete the message (works only server-side) 
await client.deleteMessage(messageID, true);
```

Strange, but several lines below I see that hard delete "Can be done client-side by users but be cautious this action is not recoverable"

bychkovdmitry (commented):
Could you add KDoc to public methods (ChatClient, ChannelController)?

bychkovdmitry:
So, are we supposed to display thread replies when parent message is deleted?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
adasiewiczr: Docs are outdated

UNRELEASED_CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 
 ### ⚠️ Changed
 - Deprecated `Channel#name`, `Channel#image`, `User#name`, `Ues#image` extension properties. Use class members instead.
+- 🚨 Breaking change: Now it is possible to hard delete messages. Insert a flag `hard = true` in the `ChatClient.deleteMessage` and it will be deleted in the backend. **This action can't be undone!**
```
adasiewiczr, line 36: Is this a breaking change? 🤔

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: Yep, that made me confused too.

UNRELEASED_CHANGELOG.md
leandroBorgesFerreira: I'll add the `@JvmOverloads` suggested by Dmitri and it won't be a breaking change. (It would be a breaking change for Java users)

leandroBorgesFerreira:
I'll have to check if show the thread of a deleted message or not... My guess for now is that we don't do this.

leandroBorgesFerreira:
For hard delete we don't show the thread, for soft delete we do.

bychkovdmitry (approved):

PR #2353, opened 2021-09-15
Force light mode fix

leandroBorgesFerreira:
### 🎯 Goal

Fix for light/dark mode force. 

### 🛠 Implementation details

Using `ContextThemeWrapper` instead of using `Context.createConfigurationContext`. `createConfigurationContext` creates a new context an Activity becomes a normal context when we call it... which causes this problem. 

`StreamUiEmptyTheme` is being used to create `ContextThemeWrapper` for `streamUiValidTheme`... it doesn't change the original theme the this theme is empty.

### 🧪 Testing

Force dark and light mode and test the clicks in MessageListView

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/Context.kt
```diff
@@ -63,18 +63,22 @@ internal val Context.streamThemeInflater: LayoutInflater
...
-            ContextThemeWrapper(configuredContext, typedValue.resourceId)
-        else -> ContextThemeWrapper(configuredContext, R.style.StreamUiTheme)
+        this.theme.resolveAttribute(R.attr.streamUiValidTheme, typedValue, true) ->
+            ContextThemeWrapper(this, R.style.StreamUiEmptyTheme)
+        this.theme.resolveAttribute(R.attr.streamUiTheme, typedValue, true) ->
+            ContextThemeWrapper(this, typedValue.resourceId)
+        else -> ContextThemeWrapper(this, R.style.StreamUiTheme)
+    }.apply {
+        applyOverrideConfiguration(
+            Configuration().apply { systemUiMode(uiMode)?.let { this.uiMode = it }}
```
JcMinarro, line 75: This new `Configuration` will loss info from the current configuration on the given Context

JcMinarro (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/Context.kt
```diff
@@ -63,18 +63,22 @@ internal val Context.streamThemeInflater: LayoutInflater
...
     val typedValue = TypedValue()
-    val configuredContext = configureUiMode(uiMode)
 
     return when {
-        configuredContext.theme.resolveAttribute(R.attr.streamUiValidTheme, typedValue, true) -> configuredContext
-        configuredContext.theme.resolveAttribute(R.attr.streamUiTheme, typedValue, true) ->
-            ContextThemeWrapper(configuredContext, typedValue.resourceId)
-        else -> ContextThemeWrapper(configuredContext, R.style.StreamUiTheme)
+        this.theme.resolveAttribute(R.attr.streamUiValidTheme, typedValue, true) ->
+            ContextThemeWrapper(this, R.style.StreamUiEmptyTheme)
```
JcMinarro, line 69: If it is a valid theme, we shouldn't wrap the context and apply a new theme on it

leandroBorgesFerreira: The problem is that we need `ContextThemeWrapper` to use `applyOverrideConfiguration`. At this moment I didn't find a other way to do this.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/Context.kt
leandroBorgesFerreira: Fixed, changed the constructor to add the original `Configuration`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
```diff
@@ -76,5 +76,5 @@ public object ChatUI {
             mimeTypeIconProviderOverride = value
         }
 
-    public var uiMode: UiMode = UiMode.SYSTEM
+    public var uiMode: UiMode = UiMode.LIGHT
```
JcMinarro, line 79: We shouldn't use `Light` as the default one

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/Context.kt
JcMinarro: What about to use the current theme?
Can't you do something like:
```
ContextThemeWrapper(this, this.theme)
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
leandroBorgesFerreira: Ops, this is a mistake.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/Context.kt
leandroBorgesFerreira: This requires API 23 😢

JcMinarro: We need to find another solution, if we apply an Empty Theme, we will override the valid theme and nothing will be configured with the theme

leandroBorgesFerreira: From what I read, the new theme doesn't override fields that are not the same. I did some tests and changing the base theme fields have effect in the sample app... but we would probably lose runtime changes in the theme. 

The other solution I see would be to pass a flag to our Style classes and parse the right color there. We could have literal colors (literal_white, literal_black, literal_blue...) and we choose between then. The draw back in that it will need more code and it is a bit harder to maintain... but on the good side it removes the need for the `StreamUiEmptyTheme` (which is super odd).

leandroBorgesFerreira: @JcMinarro What do you think? 👆

JcMinarro: So.... Whatever field doesn't exis on the new theme is asked to the "base theme" of the context?
On that case it should be fine. 

The second option you propose we should avoid, we shouldn't add any logic in our Style Code to decide which color to use, it should be handled by the system

JcMinarro (approved):

Issue #2120, opened 2021-08-02
[Compose] - ImagePreviewActivity improvements


filbabic:
We need to improve the experience of the `ImagePreviewActivity`. We have the basic behavior done, but we're missing some of the features to make it whole

Action points:
- [x] Open the specific image when clicking on message items
- [x] Implement header options toggle
- [x] Scroll to message option
- [x] Download image option
- [x] Reply to message option
- [x] Delete message option
- [x] Allow image resize, rotate, drag
- [x] Allow the image gallery option

PR #2356, opened 2021-09-16
[Compose] 2120 - image preview activity improvements

filbabic:
WIP More work done on #2120. Specifically with Image options that are shown in the header.

### 🎯 Goal

We're expanding and improving the set of features we support with the ImagePreviewActivity. We want to give the same experience as we do in the core Android SDK.

### 🛠 Implementation details

- Added a header toggle for images in the preview
- Added options to Download/Delete/Show in chat/reply to an image/message
- Added a way to open the image the user tapped on in the MessageList in the Preview

### 🎨 UI Changes

| Before | After |
| --- | --- |
| before | after |

The menus are also visible below.

| Before | After |
| --- | --- |
| options | gallery |

### 🧪 Testing

Open any image (or images) in MessageList. You should see the ellipsis (3 dots) in the header that toggle image options.

### ☑️ Checklist

~~- [] New code is covered by unit tests~~
- [x] Reviewers added

### 🎉 GIF

gallery


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/imagepreview/ImagePreviewAction.kt
```diff
@@ -0,0 +1,16 @@
+package io.getstream.chat.android.compose.state.imagepreview
+
+import io.getstream.chat.android.client.models.Message
+
+/**
+ * Represents the actions the user can take with images in the Image Preview.
+ */
+public sealed class ImagePreviewAction(public val message: Message)
```
zsmb13, line 8: Do these need to be public?

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/MessagesActivity.kt
```diff
@@ -111,6 +113,21 @@ class MessagesActivity : AppCompatActivity() {
...
+                                val message = listViewModel.getMessageWithId(result.messageId)
+
+                                if (message != null) {
+                                    composerViewModel.performMessageAction(Reply(message))
+                                }
+                            }
+
+                            ImagePreviewResultType.SHOW_IN_CHAT -> {
+                            }
+                            else -> Unit
```
zsmb13, line 129:
```suggestion
                            null -> Unit
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -73,26 +133,79 @@ public class ImagePreviewActivity : AppCompatActivity() {
...
+     */
+    @OptIn(ExperimentalAnimationApi::class)
     @Composable
-    private fun ImagePreviewContentWrapper(message: Message) {
+    private fun ImagePreviewContentWrapper(
+        message: Message,
+        initialAttachmentPosition: Int,
+    ) {
+        val pageCount = message.attachments.size
+        val startingPosition = if (initialAttachmentPosition >= pageCount) 0 else initialAttachmentPosition
```
zsmb13, line 157:
```suggestion
        val startingPosition = if (initialAttachmentPosition !in message.attachments.indices) 0 else initialAttachmentPosition
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -73,26 +133,79 @@ public class ImagePreviewActivity : AppCompatActivity() {
...
+                exit = fadeOut()
+            ) {
+                ImageGallery(
+                    pagerState = pagerState,
+                    modifier = Modifier.animateEnterExit(
+                        enter = slideInVertically(initialOffsetY = { height -> height / 2 }),
+                        exit = slideOutVertically(targetOffsetY = { height -> height / 2 })
+                    )
+                )
+            }
```
zsmb13, line 200: Great animations here!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -141,35 +265,245 @@ public class ImagePreviewActivity : AppCompatActivity() {
...
+                    .padding(16.dp)
+                    .width(150.dp)
+                    .wrapContentHeight()
+                    .align(Alignment.TopEnd),
+                shape = RoundedCornerShape(16.dp),
+                elevation = 4.dp,
+                color = ChatTheme.colors.barsBackground
+            ) {
+                Column(modifier = Modifier.fillMaxWidth()) {
+                    options.forEachIndexed { index, option ->
```
zsmb13, line 329: Not worth using a `LazyColumn` for? Wouldn't we get a nicer solution for the spacers between items if we used that?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -141,35 +265,245 @@ public class ImagePreviewActivity : AppCompatActivity() {
...
+     * @param attachments The attachments to show.
+     */
     @Composable
-    private fun ImagePreviewContent(pagerState: PagerState, message: Message) {
+    private fun ImagePreviewContent(pagerState: PagerState, attachments: List<Attachment>) {
         HorizontalPager(modifier = Modifier.background(ChatTheme.colors.appBackground), state = pagerState) { page ->
 
-            if (message.attachments.isNotEmpty()) {
-                val painter = rememberImagePainter(data = message.attachments[page].imagePreviewUrl)
+            if (attachments.isNotEmpty()) {
```
zsmb13, line 450: An early return could simplify this method `if (attachments.isEmpty()) return`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -141,35 +265,245 @@ public class ImagePreviewActivity : AppCompatActivity() {
...
+
+                val transformableState = rememberTransformableState { zoomChange, _, _ ->
+                    val newScale = (currentScale * zoomChange)
+                        .coerceAtLeast(1f)
+                        .coerceAtMost(3f)
+
+                    currentScale = newScale
+                }
+
+                val scale by animateFloatAsState(targetValue = currentScale)
```
zsmb13, line 463: This scaling works for an initial implementation, but we should revisit this and add nicer scale/pan control later on.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
```diff
@@ -200,24 +534,83 @@ public class ImagePreviewActivity : AppCompatActivity() {
...
+                title = stringResource(id = R.string.stream_compose_image_preview_reply),
+                titleColor = ChatTheme.colors.textHighEmphasis,
+                icon = Icons.Default.Reply,
+                iconColor = ChatTheme.colors.textHighEmphasis,
+                action = Reply(message)
+            ),
+            ImagePreviewOption(
+                title = stringResource(id = R.string.stream_compose_image_preview_show_in_chat),
+                titleColor = ChatTheme.colors.textHighEmphasis,
+                icon = Icons.Default.PanoramaFishEye,
```
zsmb13, line 575: This icon looks quite weird to me, we should replace it with the one in our design

zsmb13 (approved):
Left lots of comments, but almost all of them are minor. I'll leave it up to you which ones you want to address before merging and which ones you'll revisit later, we definitely don't have to take care of all these things before shipping these new things.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -117,10 +120,13 @@ public fun DefaultMessageContainer(
         )
     }
 
+    val backgroundColor by animateColorAsState(if (isFocused) ChatTheme.colors.highlight else ChatTheme.colors.appBackground)
```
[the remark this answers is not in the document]
filbabic: I'll change it to a `tween()` spec and increase the duration. You can LMK if it works once I update!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
```diff
@@ -208,6 +221,15 @@ public fun Messages(
                 }
             }
         }
+        val focusedItem = messages.firstOrNull { it.isFocused }
+
+        if (focusedItem != null) {
+            val focusedItemIndex = messages.indexOf(focusedItem)
+
```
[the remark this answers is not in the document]
filbabic: Good call - I tried both of the approaches and but figured having an item nullcheck woudl be more clear, however this does effectively prevent a second pass of items.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -594,4 +595,66 @@ public class MessageListViewModel(
...
+     * @param messageId The ID of the message.
+     */
+    public fun focusMessage(messageId: String) {
+        val messages = currentMessagesState.messageItems.map {
+            if (it.message.id == messageId) {
+                it.copy(isFocused = true)
+            } else {
+                it
+            }
+        }
```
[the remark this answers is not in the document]
filbabic: Sure - I was thinking about it while working on the snippet, but given that messages are not mutable, another option would be to remove an item from the list and add a new one that has the focused state as true.

I can try doing that and if it makes sense, we can go with that approach.

Edit: Given that the list is not mutable, we'd have to transform it to an ArrayList/MutableList and then perform the operations of removing and adding items, so I think this is not worth much at this point.

We could make the "isFocused" flag as a `var` and just change the target item, re-emit the list. Not sure if it's an improvement honestly. 🤷

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
filbabic: Thanks! It's just basic stuff that's default but flipped over for our use case!

However, I'm starting to add more of these to make the basic interactions a bit nicer ^^

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
filbabic: I think actually both an early return and a `finish()` call would be nicer, because the user can essentially remove all the images.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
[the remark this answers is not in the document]
zsmb13: Efficiency 💯 😄

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
filbabic: I think that it would be a bit of an overkill here, as we're just looking for a (generally) static list of a few items.

In terms of the spacers - these are dividers, so we'd have to do something similar here anyway.

So I figured it might've been confusing to have a LazyColumn and have all the items loaded eagerly.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
filbabic: Will add, also added a fix/improvement when trying to delete the last image in the message.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/imagepreview/ImagePreviewActivity.kt
filbabic: In this case it's kind of the default - zoom in 2x, then zoom in to 3x, then zoom out on taps. 

Or you mean the animation?

zsmb13: I mean we're just scaling into the middle of the image regardless of where the taps are happening, and then we can't pan around on the zoomed image

filbabic: Oh okay! That makes sense - I tried to keep it as close to the UI comp implementation which is fairly basic.

In Compose, we can easily add rotations, panning etc if needed.

PR #2357, opened 2021-09-16
Handle attachments with and without files separately

zsmb13:
### 🎯 Goal

Fix the problem where attachments with no files to upload get stuck forever without syncing to the backend.

### 🛠 Implementation details

Check whether an attachment has a file to upload (based on its `upload` property), and mark it synced immediately if there's no upload required.

### 🧪 Testing

Here's test code for non-file attachments (can be added in `ChatFragment`):

```kotlin
// Call in some listener
val attachment = Attachment(
    type = "location",
    extraData = mutableMapOf(
        "lat" to 111.1111,
        "lon" to -105.280184,
    ),
)
val message = Message(
    cid = args.cid,
    text = "Look at this attachment!",
    attachments = mutableListOf(attachment),
)
ChatDomain.instance().sendMessage(message).enqueue()

// Call on the MessageList
setAttachmentViewFactory(object:AttachmentViewFactory() {
    override fun createAttachmentView(
        data: MessageListItem.MessageItem,
        listeners: MessageListListenerContainer,
        style: MessageListItemStyle,
        parent: ViewGroup,
    ): View {
        val locAttachment  = data.message.attachments.firstOrNull { it.type == "location" }
        if (locAttachment != null) {
            return TextView(parent.context).apply {
                setText("Location is ${locAttachment.extraData["lat"]}, ${locAttachment.extraData["lon"]}")
            }
        }
        return super.createAttachmentView(data, listeners, style, parent)
    }
})
```

Test code for file attachments:

```kotlin
// Call in some listener
val file = File(requireContext().filesDir, "test.txt")
file.writeText("EMULATOR TEXT HERE")
val attachment = Attachment(
    type = "textfile",
    upload = file,
)
val message = Message(
    cid = args.cid,
    text = "Here's the file",
    attachments = mutableListOf(attachment),
)
ChatDomain.instance().sendMessage(message).enqueue()

// Call on the MessageList
setAttachmentViewFactory(object: AttachmentViewFactory() {
    override fun createAttachmentView(
        data: MessageListItem.MessageItem,
        listeners: MessageListListenerContainer,
        style: MessageListItemStyle,
        parent: ViewGroup,
    ): View {
        val textAttachment  = data.message.attachments.firstOrNull { it.type == "textfile" }
        if (textAttachment != null && textAttachment.uploadState in setOf(Attachment.UploadState.Success, null)) {
            return TextView(parent.context).apply {
                thread {
                    val fileText = URL(textAttachment.url).readText()
                    this.post {
                        setText("Text file says \"$fileText\"")
                    }
                }
            }
        }
        return super.createAttachmentView(data, listeners, style, parent)
    }
})
```

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
@@ -22,6 +22,13 @@ public data class Attachment(
     var name: String? = null,
     var fallback: String? = null,
 
+    /**
+     * The local file to upload when the attachment is sent. The [url] property
+     * will be populated with the URL of the uploaded file when done.
+     *
+     * Leaving this property empty means that there is no file to upload for
+     * this attachment.
+     */
     @Transient
     var upload: File? = null,
 
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
@@ -42,14 +42,23 @@ internal class MessageSendingService(
                     enrichWithCid(channelController.cid)
                 }
                 user = requireNotNull(domainImpl.user.value)
-                attachments.forEach { attachment ->
+
+                val (attachmentsToUpload, nonFileAttachments) = attachments.partition { it.upload != null }
+                attachmentsToUpload.forEach { attachment ->
                     attachment.uploadId = generateUploadId()
                     attachment.uploadState = Attachment.UploadState.InProgress
                 }
+                nonFileAttachments.forEach { attachment ->
+                    attachment.uploadState = Attachment.UploadState.Success
+                }
+
                 type = getMessageType(message)
                 createdLocallyAt = createdAt ?: createdLocallyAt ?: Date()
-                syncStatus =
-                    if (hasAttachments()) SyncStatus.AWAITING_ATTACHMENTS else if (domainImpl.isOnline()) SyncStatus.IN_PROGRESS else SyncStatus.SYNC_NEEDED
+                syncStatus = when {
+                    attachmentsToUpload.isNotEmpty() -> SyncStatus.AWAITING_ATTACHMENTS
+                    domainImpl.isOnline() -> SyncStatus.IN_PROGRESS
+                    else -> SyncStatus.SYNC_NEEDED
+                }
             }
         ).onSuccess { newMessage ->
             // Update flow in channel controller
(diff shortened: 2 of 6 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -22,6 +22,10 @@ public data class Attachment(
     var name: String? = null,
     var fallback: String? = null,
 
+    /**
+     * The local file to upload when the attachment is sent. The [url] will be populated with
+     * the URL of the uploaded file.
```
filbabic, line 27: Thank you so much!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/MessageSendingService.kt
```diff
@@ -42,14 +42,23 @@ internal class MessageSendingService(
                     enrichWithCid(channelController.cid)
                 }
                 user = requireNotNull(domainImpl.user.value)
-                attachments.forEach { attachment ->
+
+                val (attachmentsToUpload, nonFileAttachments) = attachments.partition { it.upload != null }
```
JcMinarro, line 46: Shouldn't we check the `UploadState`? :thinking: 
If the attachment is already uploaded, we shouldn't enqueue it either and be able to send it directly

zsmb13: As far as I see, this `MessageSendingService` is just the initial thing that sends messages when someone invokes the `SendMessage` use case through `ChatDomain`. Retries are handled in a different place in the code, they don't come to this service. So this is always the first time we are setting the `attachmentState` for something.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/AttachmentUploader.kt
```diff
@@ -75,8 +75,12 @@ internal class AttachmentUploader(
             mimeType = mimeType,
             url = url,
             uploadState = Attachment.UploadState.Success,
-            type = attachmentType.toString(),
         ).apply {
+            // If attachment type was not set, set it based on this value
+            // determined by the MIME type guessed from the file's extension
+            if (type == null) {
+                type = attachmentType.toString()
+            }
```
zsmb13, line 83: Previously we were overriding whatever custom `type` the user has set on the `Attachment` here

leandroBorgesFerreira (approved):

adasiewiczr (approved):

Issue #2346, opened 2021-09-15
Logical error: Messages with attachments should have another sync status


rupinderjeet:
**Describe the bug**

This is a crash report from Crashlytics (34 crashes by 7 users).
```
Fatal Exception: java.lang.IllegalArgumentException
Logical error. Messages with attachments should have another sync status!

io.getstream.chat.android.offline.ChatDomainImpl.retryMessagesWithoutAttachments (ChatDomainImpl.kt:748)
io.getstream.chat.android.offline.ChatDomainImpl.access$retryMessagesWithoutAttachments (ChatDomainImpl.kt:134)
io.getstream.chat.android.offline.ChatDomainImpl$retryMessagesWithoutAttachments$1.invokeSuspend (Unknown Source:14)
kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:106)
kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.kt:571)
kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask (CoroutineScheduler.kt:750)
kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker (CoroutineScheduler.kt:678)
kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt:665)
```

**SDK version**
- 4.17.2

**To Reproduce**
This happens when user launches the app. Upon launch, I initialize `ChatClient` in `Application` class. (I have doubts about doing it here though). I am unable to reproduce it on my end.

**Expected behavior**
ChatClient is `initialized`.

**Device:**
 - Vendor and model: HUAWEI Y6 Pro (2019) (armeabi-v7a)
 - Android version: 9

**Screenshots**
NA

Additional Information:
Huawei Y6 Pro (2019) has appeared in #2333 as well

#### Also, do you use something like work-manager to resend failed messages?

adasiewiczr:
hey @rupinderjeet,
thanks for reporting, we are going to check that.
We recommend initializing our SDK in the `Application` class.

samiuelson:
Hey 👋🏼 The new [release](https://github.com/GetStream/stream-chat-android/releases/tag/4.19.0) containing the fix is now available. Please check it and let us know if you have any issues.

PR #2361, opened 2021-09-17
[CAS-1325] Improve messages retry

adasiewiczr:
https://stream-io.atlassian.net/browse/CAS-1325
#2346 

### 🎯 Goal

Improve messages retry

### 🛠 Implementation details

There might be a situation when all attachments were sent but the message has `SYNC_NEDDED` status. In that case, we are calling `retryMessagesWithoutAttachments` that was checking if the message doesn't contain attachments.
I've changed the condition the check if all attachments are already uploaded.

### 🧪 Testing

- Try to send attachments
- Try to send attachments while being offline

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -739,9 +745,13 @@ internal class ChatDomainImpl internal constructor(
         return retriedMessages
     }
 
-    private suspend fun retryMessagesWithoutAttachments(): List<Message> {
+    private suspend fun retryMessagesWithSyncedAttachments(): List<Message> {
```
zsmb13, line 748: This still includes messages that have no attachments, right? Maybe it's worth documenting that either in the method name (which will get quite long, but it's be more descriptive), or with actual comments.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -739,9 +745,13 @@ internal class ChatDomainImpl internal constructor(
...
         val messages = repos.selectMessagesSyncNeeded()
-        require(messages.all { it.attachments.isEmpty() }) { "Logical error. Messages with attachments should have another sync status!" }
+        require(
+            messages.all {
+                it.attachments.all { attachment -> attachment.uploadState === Attachment.UploadState.Success }
+            }
+        ) { "Logical error. Messages without synchronized attachments should have another sync status!" }
```
zsmb13, line 754:
```suggestion
        ) { "Logical error. Messages with non-synchronized attachments should have another sync status!" }
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -722,7 +728,7 @@ internal class ChatDomainImpl internal constructor(
 
     @VisibleForTesting
     internal suspend fun retryMessages(): List<Message> {
-        return retryMessagesWithoutAttachments() + retryMessagesWithAttachments()
+        return retryMessagesWithSyncedAttachments() + retryMessagesWithAttachments()
```
zsmb13, line 731: Could we do some further renaming here to clearify what the two categories we're handling are?

```suggestion
        return retryMessagesWithSyncedAttachments() + retryMessagesWithPendingAttachments()
```

zsmb13 (approved):
LGTM, just one suggestion to clarify the code a bit

PR #2362, opened 2021-09-17
Only skip progress updates if logging interceptors are added

zsmb13:
### 🎯 Goal

Display upload progress correctly.

### 🛠 Implementation details

- If logging is enabled, both loggers will log, and these 2 reads of the request body will be skipped
- If logging is disabled, we don't need to skip anything

### 🧪 Testing

Upload attachments in the UI Components and Compose sample apps, both in debug and release builds.

### ☑️ Checklist

- [x] Reviewers added


amitkma (approved):
Looks good to me

JcMinarro (approved):
So fragile implementation, we should try to find a better solution, in other case whenver a new interactor is added, our uploader file process could fail.
Maybe we can handle it properly, but if we plan to open HttpClient Builder to our customers, we could have some issues there

PR #2365, opened 2021-09-17
[CAS-1310] Adding custom api call logger interceptor

leandroBorgesFerreira:
[CAS-1310](https://stream-io.atlassian.net/browse/CAS-1310)

### 🎯 Goal

Add a custom interceptor for our API calls. This way a customer can choose to log the API calls the way he/she prefers. 

### 🛠 Implementation details

Adding a `Interceptor` for the builder of `ChatClient`. If a customer sets him/her own interceptor, it substitutes our `HttpLoggingInterceptor` to avoid two interceptors for loggin.

### 🧪 Testing

Use a different logging interceptor and check if it works for you.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -5,7 +5,7 @@ This document lists deprecated constructs in the SDK, with their expected time �
 | API / Feature | Deprecated (warning) | Deprecated (error) | Removed | Notes |
 | --- | --- | --- | --- | --- |
 | ChatClient#cdnTimeout | 2021.10.12<br/>4.20.0 | 2021.10.26 ⌛ | 2021.11.09 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
-| ChatClient#baseTimeout | 2021.09.28<br/>4.19.0 | 2021.10.15 ⌛ | 2021.11.15 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
+| ChatClient#baseTimeout | 2021.10.12<br/>4.20.0 | 2021.10.26 ⌛ | 2021.11.09 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
 | `DeviceRegisteredListener` <br/>*client* | 2021.09.28<br/>4.19.0 | 2021.09.28<br/>4.19.0 | 2021.10.12 ⌛ | This class is not used anymore |
 | `ViewReactionsViewStyle#bubbleBorderColor` <br/>*client* | 2021.09.28<br/>4.19.0 | 2021.10.12 ⌛ | 2021.10.26 ⌛ | Use bubbleBorderColorMine instead  |
 | `NotificationConfig` attributes <br/>*client* | 2021.09.28<br/>4.19.0 | 2021.09.28<br/>4.19.0 | 2021.10.12 ⌛ | Some attributes are not needed anymore |
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -4,7 +4,7 @@ This document lists deprecated constructs in the SDK, with their expected time �
 
 | API / Feature | Deprecated (warning) | Deprecated (error) | Removed | Notes |
 | --- | --- | --- | --- | --- |
-| ChatClient#cdnTimeout | 2021.09.28<br/>4.19.0 | 2021.10.15 ⌛ | 2021.11.15 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
+| ChatClient#cdnTimeout | 2021.10.12<br/>4.20.0 | 2021.10.26 ⌛ | 2021.11.09 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
 | ChatClient#baseTimeout | 2021.09.28<br/>4.19.0 | 2021.10.15 ⌛ | 2021.11.15 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
 | `DeviceRegisteredListener` <br/>*client* | 2021.09.28<br/>4.19.0 | 2021.09.28<br/>4.19.0 | 2021.10.12 ⌛ | This class is not used anymore |
 | `ViewReactionsViewStyle#bubbleBorderColor` <br/>*client* | 2021.09.28<br/>4.19.0 | 2021.10.12 ⌛ | 2021.10.26 ⌛ | Use bubbleBorderColorMine instead  |
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -4,6 +4,8 @@ This document lists deprecated constructs in the SDK, with their expected time �
 
 | API / Feature | Deprecated (warning) | Deprecated (error) | Removed | Notes |
 | --- | --- | --- | --- | --- |
+| ChatClient#cdnTimeout | 2021.09.28<br/>4.19.0 | 2021.10.15 ⌛ | 2021.11.15 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
+| ChatClient#baseTimeout | 2021.09.28<br/>4.19.0 | 2021.10.15 ⌛ | 2021.11.15 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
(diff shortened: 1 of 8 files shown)


UNRELEASED_CHANGELOG.md
```diff
@@ -27,6 +27,15 @@
 - Fixed sending messages using `ChatClient::sendMessage` without explicitly specifying the sender user id.
 
 ### ⬆️ Improved
+- It is now possible to add a interceptor for API calls of Stream API. You can add it in the Build of ChatClient:
```
zsmb13, line 30: This PR will further complicate this one https://github.com/GetStream/stream-chat-android/pull/2362, let's make sure we figure out how to make the two work together

leandroBorgesFerreira: 🤔 Ok. Let's merge the #2362 and I'll adapt this PR to it.

zsmb13: That PR is now merged. However, I'm not sure how we can make this work with arbitrary interceptors added by our clients. If they ever read the body of the request, they'll break our upload progress reporting code.

See https://github.com/GetStream/stream-chat-android/pull/2362#pullrequestreview-758671037

leandroBorgesFerreira:
Closing this in favor of #2386

leandroBorgesFerreira:
#2386 Making this PR viable

zsmb13 (commented):
If we're exposing OkHttp anyway, how about we just expose the `OkHttpClient.Builder` that we use? We could provide API like this (just a high-level example):

```kotlin
val client = ChatClient.Builder("apikey", context)
    .okHttpConfig { builder ->
        builder
            .connectTimeout(10, TimeUnit.SECONDS)
            .addInterceptor { chain ->
                chain.proceed(chain.request())
            }
    }
    .build()
```

We could apply the provided lambda to our builder after everything but the `ProgressInterceptor` is added. As clients would have direct access to the builder, we could also deprecate the various timeout options available on the builder, and ask them to use this API instead.

[Coil does something similar](https://coil-kt.github.io/coil/recipes/#using-a-custom-okhttpclient), though they actually let the user provide the entire client. We could also allow that, in that case we could just use their provided instance as the base client in `BaseChatModule`:

```kotlin
private val baseClient: OkHttpClient by lazy { customClient ?: OkHttpClient() }
```

And the public API could be:

```kotlin
val client = ChatClient.Builder("apikey", context)
    .okHttpClient(client)
    .build()
```

ogkuzmin:
I agree with Marton. IMO there is no sense to provide only interceptors for logging. It's better to let users inject their own interceptors or provide the whole `OkHttpClient` or builder for `OkHttpClient`.

leandroBorgesFerreira:
Ok. Let me allow it to be more broadly customizable and let users provide the whole HTTP client. That would even improve their performance as they would have only one HTTP client for everything.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1591,6 +1594,7 @@ public class ChatClient internal constructor(
             return this
         }
 
+        @Deprecated("Use customizeOkHttpClient to set the timeouts")
```
zsmb13, line 1597: I think we should also deprecate `cdnTimeout` and the corresponding properties in `ChatClientConfig`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1605,6 +1609,15 @@ public class ChatClient internal constructor(
             warmUp = false
         }
 
+        public fun customOkHttpClient(okHttpClient: OkHttpClient): Builder = apply {
+            this.customOkHttpClient = okHttpClient
+        }
```
zsmb13, line 1614: Seems to me like this API on its own would be enough for all use cases, and we don't need the second new method if we provide this one. I'd also name it `okHttpClient` to keep it simpler. Everything else around here looks all good to me!

DEPRECATIONS.md
```diff
@@ -4,6 +4,8 @@ This document lists deprecated constructs in the SDK, with their expected time 
 
 | API / Feature | Deprecated (warning) | Deprecated (error) | Removed | Notes |
 | --- | --- | --- | --- | --- |
+| ChatClient#cdnTimeout | 2021.09.28<br/>4.19.0 | 2021.10.15 ⌛ | 2021.11.15 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
```
adasiewiczr, line 7:
```suggestion
| ChatClient#cdnTimeout | 2021.10.12<br/>4.20.0 | 2021.10.26 ⌛ | 2021.11.09 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
```

DEPRECATIONS.md
```diff
@@ -4,6 +4,8 @@ This document lists deprecated constructs in the SDK, with their expected time 
...
+| ChatClient#baseTimeout | 2021.09.28<br/>4.19.0 | 2021.10.15 ⌛ | 2021.11.15 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
```
adasiewiczr, line 8:
```suggestion
| ChatClient#baseTimeout | 2021.10.12<br/>4.20.0 | 2021.10.26 ⌛ | 2021.11.09 ⌛ | Use ChatClient.okHttpClient() to set the timeouts  |
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -148,14 +149,14 @@ internal open class BaseChatModule(
             .addInterceptor(HeadersInterceptor(getAnonymousProvider(config, isAnonymousApi)))
             .apply {
                 if (config.loggerConfig.level != ChatLogLevel.NOTHING) {
-                    addInterceptor(HttpLoggingInterceptor())
```
adasiewiczr, line 151: Why are we removing this interceptor?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
zsmb13: (The question about why we need both new methods is still open)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
leandroBorgesFerreira: No, small mistake

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: Ops, I removed only the setter of the builder for httpClientConfig

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -128,7 +129,7 @@ internal open class BaseChatModule(
     }
 
     // Create Builders from a single client to share threadpools
-    private val baseClient: OkHttpClient by lazy { OkHttpClient() }
+    private val baseClient: OkHttpClient by lazy { customOkHttpClient ?: OkHttpClient() }
```
zsmb13, line 132: I think the `clientBuilder` method will override the timeouts set on this `OkHttpClient` in all cases

zsmb13 (changes requested):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
leandroBorgesFerreira: Not it only overrides the timeout if the baseClient is not the custom one

zsmb13: Neat solution!

zsmb13 (approved):
I added docs on the new method, good to go as far as I'm concerned

PR #2376, opened 2021-09-21
[CAS-1102] Feature/dismiss notification when go into a channel

JcMinarro:
### 🎯 Goal
All notifications related to a Channel are dismissed after the user going into the channel conversation 

### 🛠 Implementation details
When the channel is load (In our UI Components or Compose implementation) we search all notifications related to this channel and dismiss them

### ☑️ Checklist

- [x] Reviewers added


CHANGELOG.md
```diff
@@ -68,6 +69,7 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Notifications are dismissed after the user go into the channel conversation
```
adasiewiczr, line 72: It would be nice to mention this will happen only if you are using our `MessageListViewModel`.
Should we also describe this somewhere in docs? 🤔

zsmb13: +1, We should definitely document this somewhere around notifications

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/ChatNotifications.kt
```diff
@@ -85,6 +86,14 @@ internal class ChatNotificationsImpl constructor(
         handler.onDismissNotification(notificationId)
     }
 
+    /**
+     * Dismiss notification associated to the [channelType] and [channelId] received on the params
```
filbabic, line 90: I think we should try to improve the documentation for our internal (and public) code. 

I think @zsmb13 prepared some guidelines on what they should look like, but generally, we should list out params using `@param`, we should make sure we finish all sentences with a full-stop (so that we know the comment is done) etc.

In general, we'll look back to these functions and having a bit more structure or description on how the function works will be helpful to internal devs too.

LMK what you think!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/ChatNotificationHandler.kt
```diff
@@ -233,6 +240,15 @@ public open class ChatNotificationHandler @JvmOverloads constructor(
         .joinToString { it.name }
         .takeIf { it.isNotEmpty() }
 
+    private fun dismissSummaryNotification(notificationSummaryId: Int) {
```
filbabic, line 243: Would love to have docs here too!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -206,7 +206,10 @@ public class MessageListViewModel(
                         newState
                     }
                     lastLoadedMessage = newLastMessage
-                    setChanel(controller.toChannel())
+                    controller.toChannel().let { channel ->
```
filbabic, line 209: I think it might be easier to just say:

```kotlin
val channel = controller.toChannel()
...
```

filbabic (commented):
Left just a few comments. I think we should try to improve the docs a bit (as per the other comments too), so that the API is more clear to users and us! :]

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -99,7 +99,7 @@ public class ChatClient internal constructor(
     public val config: ChatClientConfig,
     private val api: ChatApi,
     private val socket: ChatSocket,
-    private val notifications: ChatNotifications,
+    @InternalStreamChatApi public val notifications: ChatNotifications,
```
adasiewiczr, line 102: We need to make it public for testing purposes, right? Maybe we can leave it for now but it would be nice to improve this in the future

adasiewiczr (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: Yep, it is needed because on the tests side we need to override this collaborator and it is not accessible in another case.
Hope AGP having support  for `java-test-fixtures` plugin soon and we can avoid all this ugly changes

PR #2378, opened 2021-09-21
[Compose] - 2355 - Improve MessageListHeader.kt API

### LINKED ISSUE #2355: [Compose] - Add default values to MessageListHeader
### Context

Add default values for lambda params in `MessageListHeader`

### Change Proposal

Make it easier to call this function, for example I shouldn't have to provide the last param here

```kotlin
MessageListHeader(
    channel = listViewModel.channel,
    currentUser = user,
    isNetworkAvailable = true,
    messageMode = listViewModel.messageMode,
    modifier = Modifier
        .fillMaxWidth()
        .height(56.dp),
    onBackPressed = onBackPressed,
    onHeaderActionClick = {},
)
```

### Acceptance criteria

- [ ] Parameters of `MessageListHeader` can be omitted where it makes sense

filbabic:
Fixes #2355 

### 🎯 Goal

We're improving the `MessageListHeader` API so that it's more customizable and easier to use.

### 🛠 Implementation details

- We added more default value parameters so that the user can easily use the API.
- We also added more parameters that allow for customization of specific parts of the header component.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
update


docusaurus/docs/Android/04-compose/04-message-components/02-message-list-header.mdx
```diff
@@ -1,12 +1,9 @@
 # MessageListHeader
 
-The `MessageListHeader` component is a clean, stateless component, that doesn't require a `ViewModel`. It consists of three elements:
+The `MessageListHeader` component is a clean, stateless component, that doesn't require a `ViewModel`. It consists of three internal components by default and it
+allows you to customize the data, design style, action handlers and the internal components.
```
zsmb13, line 4: > consists of three internal components by default

Why do we include this in the docs if we no longer list what these are? Does this give the reader any information?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -38,42 +38,54 @@ import io.getstream.chat.android.compose.ui.util.getDisplayName
...
+            modifier = Modifier
+                .size(36.dp),
+            imageVector = Icons.Default.ArrowBack,
+            onBackPressed = onBackPressed,
+        )
+    },
 
+    titleContent: @Composable RowScope.() -> Unit = {
+        DefaultMessageHeaderTitle(
+            modifier = Modifier.weight(6f),
```
zsmb13, line 73: I'm curious, does `6f` matter here as the value? Would `1f` have the same effect?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -86,55 +98,58 @@ public fun MessageListHeader(
...
-            )
+            titleContent()
 
-            ChannelAvatar(
-                modifier = Modifier.size(36.dp),
-                channel = channel,
-                currentUser = currentUser,
-                contentDescription = channel.name
-            )
+            trailingContent()
```
zsmb13, line 105: Love how clean the implementation of this function is now 😄

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -86,55 +98,58 @@ public fun MessageListHeader(
...
+        channelName
+    } else {
+        stringResource(id = R.string.stream_compose_thread_title)
+    }
+
+    val subtitle = if (messageMode == Normal) {
+        stringResource(id = R.string.stream_compose_channel_members, memberCount, onlineCount)
+    } else {
+        stringResource(id = R.string.stream_compose_thread_subtitle, channelName)
+    }
```
zsmb13, line 143: We can make the `else` case more obvious with `when`s here

```suggestion
    val title = when (messageMode) {
        Normal -> channelName
        is Thread -> stringResource(id = R.string.stream_compose_thread_title)
    }

    val subtitle = when (messageMode) {
        Normal -> stringResource(id = R.string.stream_compose_channel_members, memberCount, onlineCount)
        is Thread -> stringResource(id = R.string.stream_compose_thread_subtitle, channelName)
    }
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -86,55 +98,58 @@ public fun MessageListHeader(
...
-    subtitle: String,
-    isNetworkAvailable: Boolean,
+    messageMode: MessageMode = Normal,
+    onHeaderActionClick: (Channel) -> Unit = {},
+    isNetworkAvailable: Boolean = true,
 ) {
+    val memberCount = channel.memberCount
+    val onlineCount = channel.members.count { it.user.online }
+
+    val channelName = channel.getDisplayName()
```
zsmb13, line 131: We could probably inline this value to wherever it's used - I think it's never used more than once

docusaurus/docs/Android/04-compose/04-message-components/02-message-list-header.mdx
filbabic: That's a good point. It used to describe the component and it felt a bit weird without having it here.

Maybe rephrasing this to something like: "It allows you to customize the data, design style, action handlers and the internal components that represent the leading and trailing actions and the title content."

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
filbabic: That's a good point - `1f` works too as it fills the remaining space.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
filbabic: Yeah it's really nice! :]

And we can customize different parts of it to a greater extent which is cool

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
filbabic: Channel name is used in two places, but the counts are not, so I can inline those two!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
zsmb13: The suggestion here was specifically to have `is Thread` as the condition on the second branches. Otherwise, it's "what to do in normal mode and what to do otherwise", but it might not be immediately clear what the "otherwise" case means if you don't know this code very well.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
zsmb13: Only one of those two usages will actually happen though (one is in the Normal, the other in the Thread case), so even if we inline it we'll only ever fetch the value once

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
filbabic: Ohh okay that makes more sense now. I'll update it accordingly!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
filbabic: That's true - I had figued pulling it out is okay in the sense that if we want to customize the display name or anything it'll be easier to update it in one place.

But thinking about it, there's really not much need for that!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/MessagesScreen.kt
```diff
@@ -61,7 +62,7 @@ public fun MessagesScreen(
     showHeader: Boolean = true,
     enforceUniqueReactions: Boolean = true,
     onBackPressed: () -> Unit = {},
-    onHeaderActionClick: () -> Unit = {},
+    onHeaderActionClick: (Channel) -> Unit = {},
```
zsmb13, line 65: Small tip for later: you can name these params and it will help a bit with autocompletion on the caller side:

```suggestion
    onHeaderActionClick: (channel: Channel) -> Unit = {},
```

zsmb13 (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/MessagesScreen.kt
filbabic: @zsmb13 Yeah that's a good idea - do you think we should do this for all of our lambdas?

zsmb13: Yeah, I think it's good practice. Wish there was a lint warning to enable for this 😃 It's a super minor thing though, really just helps the IDE tooling a tiny bit on the call site. Doesn't have any practical effect otherwise.

PR #2383, opened 2021-09-22
Add script for closing issues without response

adasiewiczr:
### 🎯 Goal

Add script for closing issues without response

### 🛠 Implementation details

Added a Github action for closing issues marked with `waiting for response` label that didn't receive any response from the author for at least 7 days.
If the original author comes back and gives more information, the label is removed and the issue is reopened, if necessary

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/.github/workflows/no-response.yml
+++ b/.github/workflows/no-response.yml
@@ -1,0 +1,25 @@
+name: No Response
+
+on:
+  issue_comment:
+    types: [created]
+  schedule:
+    # Schedule at the end of every day
+    - cron: '0 0 * * *'
+
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@v0.5.0
+        with:
+          token: ${{ github.token }}
+          daysUntilClose: 7
+          responseRequiredLabel: waiting for response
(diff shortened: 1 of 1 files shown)


.github/workflows/no-response.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  noResponse:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lee-dohm/no-response@v0.5.0
+        with:
+          token: ${{ github.token }}
+          daysUntilClose: 7
+          responseRequiredLabel: waiting for response
+          responseRequiredColor: D28960
```
zsmb13, line 19: Feels like we could just create the label on the UI, and then we don't need the hex code in the config here 😃

zsmb13 (approved):
LGTM, we can adjust it later if this turns out to be too aggressive, but really helpful otherwise

.github/workflows/no-response.yml
adasiewiczr: The label is already created and you're right - we don't need it in the config

amitkma (approved):
LGTM

JcMinarro (approved):

PR #2384, opened 2021-09-22
Refactor ProgressTracker implementation 

amitkma:
https://stream-io.atlassian.net/browse/CAS-1199

### 🎯 Goal
The current implementation of how attachment upload progress is propagated isn't the best. We have a static instance of the progress tracker stored inside ProgressTrackerFactory. It's always easy to have memory leaks.

### 🛠 Implementation details

`ProgressTrackerFactory` is not used and `UploadState` is updated with the help of Callback. 

### ☑️ Checklist

- [x] Reviewers added


amitkma:
@filbabic Could you please check how we can do the same for Compose?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
```diff
@@ -32,7 +32,7 @@ internal class UploadAttachmentsWorker(private val appContext: Context) {
                 Result.success(Unit)
             } else {
                 val hasPendingAttachment = message.attachments.any { attachment ->
-                    attachment.uploadState == Attachment.UploadState.InProgress
+                    attachment.uploadState == Attachment.UploadState.InProgress(0)
```
zsmb13, line 35: Shouldn't this be

```suggestion
                    attachment.uploadState is Attachment.UploadState.InProgress
```

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/ChatDomainImplTest.kt
```diff
@@ -79,7 +79,7 @@ internal class ChatDomainImplTest {
             val awaitingAttachmentsMessage = randomMessage(
                 syncStatus = SyncStatus.AWAITING_ATTACHMENTS,
                 attachments = mutableListOf(
-                    randomAttachment { uploadState = Attachment.UploadState.InProgress },
+                    randomAttachment { uploadState = Attachment.UploadState.InProgress(10) },
```
zsmb13, line 82: Is this `10` on purpose?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/ProgressRequestBody.kt
```diff
@@ -16,7 +14,6 @@ internal class ProgressRequestBody(
     private val callback: ProgressCallback
 ) : RequestBody() {
 
-    private val handler = Handler(Looper.getMainLooper())
```
zsmb13, line 19: How come we don't need this anymore?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -689,6 +695,19 @@ public class ChannelController internal constructor(
         }
     }
 
+    private fun updateAttachmentUploadState(messageId: String, uploadId: String, newState: Attachment.UploadState) {
+        val message = _messages.value[messageId]
+        if (message != null) {
+            val attachment = message.attachments.firstOrNull { it.uploadId == uploadId }
+            if (attachment != null) {
+                val newAttachments = message.attachments - attachment + attachment.copy(uploadState = newState)
```
zsmb13, line 703: Will this change the order of attachments, always putting the one that we've just updated to the end of the list?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -41,7 +41,7 @@ public data class Attachment(
 ) : CustomObject {
 
     public sealed class UploadState {
-        public object InProgress : UploadState()
+        public data class InProgress(val currentProgress: Long) : UploadState()
```
zsmb13, line 44: `currentProgress` here is a percentage value from 0 to 100, right? We should probably document that.

Also, did we give up on adding two parameters here, a current bytes + a total bytes value?

stream-chat-android-client/api/stream-chat-android-client.api
```diff
@@ -1954,7 +1954,14 @@ public final class io/getstream/chat/android/client/models/Attachment$UploadStat
...
 public final class io/getstream/chat/android/client/models/Attachment$UploadState$InProgress : io/getstream/chat/android/client/models/Attachment$UploadState {
-	public static final field INSTANCE Lio/getstream/chat/android/client/models/Attachment$UploadState$InProgress;
+	public fun <init> (J)V
+	public final fun component1 ()J
+	public final fun copy (J)Lio/getstream/chat/android/client/models/Attachment$UploadState$InProgress;
+	public static synthetic fun copy$default (Lio/getstream/chat/android/client/models/Attachment$UploadState$InProgress;JILjava/lang/Object;)Lio/getstream/chat/android/client/models/Attachment$UploadState$InProgress;
+	public fun equals (Ljava/lang/Object;)Z
+	public final fun getCurrentProgress ()J
+	public fun hashCode ()I
+	public fun toString ()Ljava/lang/String;
```
zsmb13, line 1964: Changing `InProgress` from an object to a data class is a breaking change, should definitely be in the changelogs

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/AttachmentUploader.kt
```diff
@@ -54,11 +45,12 @@ internal class AttachmentUploader(
                     it
                 }
             }
-
-            progressTracker?.setComplete(true)
+            attachment.uploadState = Attachment.UploadState.Success
+            progressCallback?.onSuccess(file.absolutePath)
```
zsmb13, line 49: This would be a good time to document the methods of `ProgressCallback`, as it's not obvious what `String` you'll receive here as the parameter.

Actually, now that I'm commenting on this, it'd make more sense to me to get the URL of the successful upload than the local file path.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
```diff
@@ -230,31 +226,18 @@ private class FileAttachmentViewHolder(
...
-                            binding.progressBar,
-                            attachment,
-                            uploadedBytes,
-                            isComplete,
-                            totalValue
-                        )
-                    }
-                }
-            }
+            val uploadState = (attachment.uploadState as Attachment.UploadState.InProgress)
```
zsmb13, line 229: We could take `uploadState` as a parameter to this method to avoid this downcasting

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
```diff
@@ -230,31 +226,18 @@ private class FileAttachmentViewHolder(
...
+            val progress = uploadState.currentProgress
+            val fileLength = attachment.upload?.length() ?: 0
+            val totalValue = MediaStringUtil.convertFileSizeByteCount(fileLength)
+            val nominalProgress = progress * fileLength / 100f
+
+            fileSizeView.text =
```
zsmb13, line 235: Not specific to this line, but the entire message item slides up and down a bit vertically as all of these progress updates are happening, which wasn't the case with the previous implementation.

zsmb13 (commented):
Promising stuff, hope we can make this approach work, would be much nicer than ProgressTrackerFactory!

Left a bunch of comments about things that are not quite clear for me yet.

stream-chat-android-client/api/stream-chat-android-client.api
```diff
@@ -2865,10 +2872,6 @@ public abstract interface class io/getstream/chat/android/client/uploader/FileUp
 	public abstract fun sendImage (Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/io/File;Lio/getstream/chat/android/client/utils/ProgressCallback;)Lio/getstream/chat/android/client/utils/Result;
 }
 
-public final class io/getstream/chat/android/client/uploader/ProgressTrackerKt {
-	public static final fun toProgressCallback (Lio/getstream/chat/android/client/uploader/ProgressTracker;)Lio/getstream/chat/android/client/utils/ProgressCallback;
```
adasiewiczr, line 2869: This is also a breaking change. I'm not sure if we can avoid it but definitely, we should mention it in the CHANGELOG and provide a migration guide

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/UploadAttachmentsWorker.kt
amitkma: Oops, you are right.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/ChatDomainImplTest.kt
amitkma: Just for testing, I tried to use some middle value instead of 0 or 100 to ensure if there is any weird behaviour.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/ProgressRequestBody.kt
amitkma: Because we are not updating progress directly on UI. As the progress is updating in the attachment model and that is observed on the UI thread in ViewModel.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
amitkma: Yes, we need to document the progress but I am little unsure about if we keep percentage here or two parameters, `bytesRead`, `totalBytes`.  I ditched the two parameters idea for the moment as ProgressCallback is public class and we will have to change the arguments of `onProgress` method also. That will be a breaking change. What do you suggest here?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/AttachmentUploader.kt
amitkma: I was not sure about the method of ProgressCallback if it accepts file path or URL. We should definitely document such things, especially it is a public class.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
amitkma: Not sure to understand this.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
amitkma: Yes, It's a known UI glitch. Working on it.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
zsmb13: We could add a second parameter to the `handleInProgressAttachment`, which has the `InProgress` type, and we could pass in that value at the call site. That way we wouldn't have to do this downcast (from the `UploadState` type to the more specific `InProgress` type) which could be unsafe if we call this method from an incorrect place in the code.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
zsmb13: Perhaps we could add a new method in `ProgressCallback` with two parameters, give it a default empty implementation. We could deprecate the old one, telling people to implement the new one instead.

Perhaps we could have the default implementation call back into the old one as its default implementation.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/attachment/AttachmentUploader.kt
ogkuzmin: It was done before Amit. By @zsmb13 ))) 
I mean this PR suppose to improve flow of upload progress. We can discuss and improve later this side of `ProgressCallback`. I believe no one currently has clear vision why it was done so

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/ProgressTracker.kt
```diff
@@ -26,17 +24,3 @@ public class ProgressTracker {
 
     public fun isComplete(): StateFlow<Boolean> = isComplete
 }
-
```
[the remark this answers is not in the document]
amitkma: I kept it there as compose is using it in some places. It is internal class so I would prefer to remove it as soon as compose migrates to this new approach.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/ProgressRequestBody.kt
zsmb13: Looking at the code that calls into `StreamFileUploader`, we could have existing clients who call `ChatClient#sendFile` and pass in a progress callback to it. If they previously relied on this being called on the UI thread, we'd now break their code.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -682,13 +688,28 @@ public class ChannelController internal constructor(
...
+    private fun updateAttachmentUploadState(messageId: String, uploadId: String, newState: Attachment.UploadState) {
+        val message = _messages.value[messageId]
+        if (message != null) {
+            val newAttachments = message.attachments.fold(emptyList<Attachment>()) { acc, item ->
+                if (item.uploadId == uploadId) {
+                    acc + item.copy(uploadState = newState)
+                } else {
+                    acc + item
+                }
+            }
```
zsmb13, line 706: This creates a new list for each iteration, so going through 100 messages creates 100 new list instances. 

Wouldn't a `map` operation work to achieve what we need here, modifying a single item, and keeping all others? That creates just a single new list and pre-allocates it to the appropriate size as well.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/TextAndAttachmentsViewHolder.kt
```diff
@@ -97,22 +88,18 @@ internal class TextAndAttachmentsViewHolder(
...
-            val scope = CoroutineScope(DispatcherProvider.Main)
-            this.scope = scope
-
-            scope.launch {
-                trackFilesSent(context, uploadIdList, binding.sentFiles)
-            }
-        } else {
+        val totalAttachmentsCount = data.message.attachments.size
+        var completedAttachmentsCount =
+            data.message.attachments.filter { it.uploadState == null || it.uploadState == Attachment.UploadState.Success }.size
```
zsmb13, line 93: Small simplification

```suggestion
            data.message.attachments.count { it.uploadState == null || it.uploadState == Attachment.UploadState.Success }
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
```diff
@@ -194,8 +188,11 @@ private class FileAttachmentViewHolder(
     }
 
     private fun subscribeForProgressIfNeeded(attachment: Attachment) {
-        if (attachment.uploadState is Attachment.UploadState.InProgress) {
-            handleInProgressAttachment(binding.fileSize)
+        if (attachment.uploadState is Attachment.UploadState.Idle) {
+            handleInProgressAttachment(binding.fileSize, 0L, attachment.upload?.length() ?: 0)
+        } else if (attachment.uploadState is Attachment.UploadState.InProgress) {
+            val uploadState = attachment.uploadState as Attachment.UploadState.InProgress
+            handleInProgressAttachment(binding.fileSize, uploadState.bytesRead, uploadState.totalBytes)
```
zsmb13, line 195: I think this is a bit cleaner for the same thing:

```suggestion
        val uploadState = attachment.uploadState
        if (uploadState is Attachment.UploadState.Idle) {
            handleInProgressAttachment(binding.fileSize, 0L, attachment.upload?.length() ?: 0)
        } else if (uploadState is Attachment.UploadState.InProgress) {
            handleInProgressAttachment(binding.fileSize, uploadState.bytesRead, uploadState.totalBytes)
```

zsmb13 (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
zsmb13: I'm still seeing this problem with the latest version of this PR

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -1498,6 +1519,27 @@ public class ChannelController internal constructor(
...
+        ProgressCallback {
+        override fun onSuccess(attachmentUrl: String?) {
+            updateAttachmentUploadState(messageId, uploadId, Attachment.UploadState.Success)
+        }
+
+        override fun onError(error: ChatError) {
+            updateAttachmentUploadState(messageId, uploadId, Attachment.UploadState.Failed(error))
+        }
+
+        override fun onProgress(progress: Long) = Unit
```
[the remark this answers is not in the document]
amitkma: If we give it a default implementation, users won't know if they have to implement it or not as IDE won't complain. New `onProgress` also has a default implementation.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
amitkma: Map operation won't work here because the state flow won't emit the new value as it will already be modified. That's why the creation of a new list is necessary unless we modify our StateFlow for such cases.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/FileAttachmentsView.kt
zsmb13: https://user-images.githubusercontent.com/12054216/135230527-fe349c8c-033a-46e0-99a8-d7ff36d14013.mp4


https://user-images.githubusercontent.com/12054216/135230539-35463d29-cfc2-4ae3-889f-441c1039d22e.mp4

zsmb13: One of the issues is that the ViewHolder resizes vertically really quickly, dropping down below the Message Input briefly. Then it tries to animate / climb back up to its normal position, but I think the rapid updates happening to the list keep interrupting that animation, hence the laggy movement upwards.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileUploadContent.kt
```diff
@@ -87,39 +85,41 @@ public fun FileUploadItem(attachment: Attachment) {
...
-                        style = ChatTheme.typography.footnote,
-                        color = ChatTheme.colors.textLowEmphasis
-                    )
                 }
             }
         }
     }
 }
+
+@Composable
```
filbabic, line 108: I'd just add KDocs here and maybe a named argument for `modifier` for the `Row`.

Otherwise, looks good!

filbabic (changes requested):
Left just one comments to clean up a bit, the rest looks good (Compose side)! :]

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileUploadContent.kt
zsmb13: Added!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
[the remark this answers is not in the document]
zsmb13: Ah, it's a difficult call. We'll have to break something at some point here in the API here. For new users using this API, the good thing would be having a default implementation for the old method and the new one not having a default implementation, forcing them to implement the new, correct one.

What do you think about going for that approach?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/TextAndAttachmentsViewHolder.kt
```diff
@@ -97,22 +88,18 @@ internal class TextAndAttachmentsViewHolder(
...
-            clearScope()
-            val scope = CoroutineScope(DispatcherProvider.Main)
-            this.scope = scope
-
-            scope.launch {
-                trackFilesSent(context, uploadIdList, binding.sentFiles)
-            }
-        } else {
+        val totalAttachmentsCount = data.message.attachments.size
+        var completedAttachmentsCount =
```
zsmb13, line 92:
```suggestion
        val completedAttachmentsCount =
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -40,9 +40,28 @@ public data class Attachment(
...
-        public object InProgress : UploadState()
+        /**
+         * Idle state before attachment starts to upload.
+         */
+        public object Idle : UploadState()
+
+        /**
+         * State representing attachment upload progress.
+         */
+        public data class InProgress(val bytesRead: Long, val totalBytes: Long) : UploadState()
```
zsmb13, line 55: We should also review the param names here, especially `bytesRead` is very off

amitkma: Not sure how I did that? I came up with `bytesUploaded` 🤔

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 🐞 Fixed
 
 ### ⬆️ Improved
+- `ProgressTrackerFactory` and `ProgressTracker` is no more used and the attachment uploaded state is updated instead.
```
filbabic, line 6: I'd maybe rephrase this instead, it's a bit hard to understand what's going on here.

Something like:

- Removed/deprecated 'ProgressTrackerFactory` and `ProgressTracker`
- Added `UploadState` to `Attachment` for easier access when uploading items

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ProgressCallback.kt
```diff
@@ -2,8 +2,37 @@ package io.getstream.chat.android.client.utils
...
+
+    /**
+     * Called when the attachment could not be uploaded due to cancellation, network problem or timeout etc with an [error].
+     *
+     * @see ChatError
+     */
     public fun onError(error: ChatError)
+
+    /**
+     * Called when the attachment upload is in progress with [progress] in percentage.
```
filbabic, line 23: Is this comment true? We do pass the `progress` as a `Long` value, not sure how that's the percentage.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/uploader/StreamFileUploaderTest.kt
```diff
@@ -33,7 +33,7 @@ internal class StreamFileUploaderTest {
     private val url = "url"
 
     private val progressCallback = object : ProgressCallback {
-        override fun onSuccess(file: String) = Unit
+        override fun onSuccess(attachmentUrl: String?) = Unit
```
filbabic, line 36: Just url as per the comment above

filbabic (commented):
Left a few comments - looks good on the Compose side, the rest I'm also supporting Marton's side, with a few additions of what I think we should improve.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ProgressCallback.kt
amitkma: Yeah, we used to do the same earlier. Of course, it doesn't make sense.

CHANGELOG.md
amitkma: UploadState was already part of Attachment and was used but it wasn't used to update progress.

zsmb13: Something like...

> `Attachment#uploadState` is now updated in real-time during uploads.

?

filbabic: @amitkma What I'm trying to say is that even though I know what this PR is for I still don't understand what this changelog comment is saying - it's somewhat confusing.

Something like Marton mentioned would be nicer for users, as they'd know what changed and what to use next more easily.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/ProgressRequestBody.kt
amitkma: Added a warning for this in Changelog

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Attachment.kt
```diff
@@ -40,9 +40,28 @@ public data class Attachment(
...
-        public object InProgress : UploadState()
+        /**
+         * Idle state before attachment starts to upload.
+         */
+        public object Idle : UploadState()
+
+        /**
+         * State representing attachment upload progress.
+         */
+        public data class InProgress(val bytesUploaded: Long, val totalBytes: Long) : UploadState()
```
zsmb13, line 55: I'm gonna nitpick here, it's weird that we have `bytesUploaded` and then `totalBytes`.

It should either be `bytesUploaded` and `bytesTotal` (sounds awkward) or `uploadedBytes` and `totalBytes` (more natural imo). We should also sync it with the param names in `ProgressCallback.

amitkma: Actually, the `bytesUploaded` and `totalBytes` are more natural and verbal. The first one tells us the number of bytes uploaded and the second one total bytes. And this is the general convention I have seen used across many projects. 

1. https://github.com/apache/eagle/blob/df0e91de65cf9428f0f735065eba580e37183030/eagle-core/eagle-app/eagle-app-base/src/main/java/org/apache/eagle/app/environment/impl/StormSubmitter.java#L310
2. https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/Progress.java#L124
4. https://github.com/cloudinary/cloudinary_kotlin/blob/03df38fc5334d0acb2e8c2b5a01cce294842a21f/uploader/src/main/kotlin/com/cloudinary/http/HttpClient.kt#L5

zsmb13: Alright, we can keep these names then. Let's make them consistent in `ProgressCallback`, and then this PR should be good to go.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/Utils.kt
```diff
@@ -10,7 +10,8 @@ private val COMMAND_PATTERN = Pattern.compile("^/[a-z]*$")
 internal fun getMessageType(message: Message): String {
     val hasAttachments = message.attachments.isNotEmpty()
     val hasAttachmentsToUpload = message.attachments.any { attachment ->
-        attachment.uploadState is Attachment.UploadState.InProgress
+        attachment.uploadState is Attachment.UploadState.InProgress ||
```
ogkuzmin, line 13: We can replace it by using extension `Message::hasPendingAttachments`

ogkuzmin (approved):
Wow, great job! I'm so glad that we could get rid of whole `ProgressTrackerFactory` ❤️

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/message/Utils.kt
amitkma: Done.

zsmb13 (approved):
🛳️🛳️🛳️🛳️🛳️🛳️

filbabic (approved):
Looks good! :]

PR #2342, opened 2021-09-14
Attachment upload fixes

zsmb13:
### 🎯 Goal

Make attachment uploads show progress nicer in both SDKs.

### 🛠 Implementation details

See commit messages for small, step-by-step changes.

### 🎨 UI Changes

Note: the Compose SDK's "after" recording doesn't include the fix for attachment counts in the footer yet.

<table>
<thead>
<tr>
<th></th>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
UI Components
</td>
<td>
<video src="https://user-images.githubusercontent.com/12054216/133408446-7996c362-6278-4647-baa4-6afd6a19c163.mp4" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/12054216/133408431-f8228aea-a477-4f72-9f95-ad801aa6e4d0.mp4" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
Compose UI Components
</td>
<td>
<video src="https://user-images.githubusercontent.com/12054216/133408441-f9cc233e-255d-425f-b36e-2cc9fbf97430.mp4" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/12054216/133408454-ebdb3f35-7266-4239-bfa6-e0b1ec36309c.mp4" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Build the sample apps, upload various (especially large-ish) files.

### ☑️ Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] PR targets the `develop` branch
- [ ] Changelog is updated with client-facing changes
- [ ] ~New code is covered by unit tests~
- [x] Comparison screenshots added for visual changes
- [ ] ~Affected documentation updated (docusaurus, tutorial, CMS (task created))~
- [x] Reviewers added

PR #2386, opened 2021-09-22
Use tagging and an interceptor to add progress tracking to upload requests

zsmb13:
### 🎯 Goal

Eliminate the brittle workaround added previously in https://github.com/GetStream/stream-chat-android/pull/2342/ to avoid triggering progress updates from logging interceptors.

### 🛠 Implementation details

Instead of wrapping the body in a `ProgressRequestBody` that contains a `ProgressCallback` before passing it to Retrofit, we pass in the callback as a Tag (see [OkHttp docs](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-request/-builder/tag/)/[Retrofit docs](https://square.github.io/retrofit/2.x/retrofit/retrofit2/http/Tag.html)).

Then we identify the tagged request at the end of the interceptor chain in a new interceptor, and wrap the body into a `ProgressRequestBody` there, after all other interceptors are done doing their work (which for logging interceptors included triggering writes of the request body).

### 🧪 Testing

Send images / files, see progress working. Both in debug and release builds.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/StreamFileUploader.kt
```diff
@@ -41,17 +41,15 @@ internal class StreamFileUploader(
...
         return retrofitCdnApi.sendFile(
-            channelType,
-            channelId,
-            part,
-            connectionId
+            channelType = channelType,
+            channelId = channelId,
+            file = part,
+            connectionId = connectionId,
+            progressCallback = null,
```
zsmb13, line 52: It would be nice to have just a single `sendFile` method here that takes a nullable `ProgressCallback` as a parameter that is `null` by default. However, this would be a breaking change for our clients that have existing implementations of `StreamFileUploader`.

leandroBorgesFerreira (approved):

adasiewiczr (approved):

PR #2388, opened 2021-09-22
[CAS-1307] Fixing isFirstRequest problem

leandroBorgesFerreira:
[CAS-1307](https://stream-io.atlassian.net/browse/CAS-1307)

### 🎯 Goal

Fix shimmer effect. We are showing the shimmer effect if we don't change `channelOffset` of `QueryChannelsController` because our online request may not return any new channel, so it behaves like it was the first request.

We can't use channelOffset to check that we are in the firstPage because `updateOnlineChannels` doesn't run when there's no internet and that would cause all offline requests to behave like the firstPage. 

### 🛠 Implementation details

Now the first page is tracked in `QueryChannelsController`. 

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/134380328-5282f2f9-13e7-4f82-a7ec-11e4f8559107.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/134380532-7a1692fa-c10e-4b49-bc0e-9434280c9fca.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Follow the videos: 
- Go to the `develop` branch. Disable the internet and request new pages. Check that the shimmer effect appears many times. 
- Go to this branch. Disable the internet and request new pages. Check that the shimmer effect appears only one time.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

### DIFF
--- a/stream-chat-android-offline/api/stream-chat-android-offline.api
+++ b/stream-chat-android-offline/api/stream-chat-android-offline.api
@@ -443,6 +443,7 @@ public final class io/getstream/chat/android/offline/querychannels/QueryChannels
 	public final fun getNewChannelEventFilter ()Lkotlin/jvm/functions/Function3;
 	public final fun getRecoveryNeeded ()Z
 	public final fun getSort ()Lio/getstream/chat/android/client/api/models/QuerySort;
+	public final fun isFirstRequest ()Z
 	public final fun query (IIILkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun query$default (Lio/getstream/chat/android/offline/querychannels/QueryChannelsController;IIILkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
 	public final fun refreshChannel (Ljava/lang/String;)V
--- a/stream-chat-android-offline/api/stream-chat-android-offline.api
+++ b/stream-chat-android-offline/api/stream-chat-android-offline.api
@@ -443,7 +443,6 @@ public final class io/getstream/chat/android/offline/querychannels/QueryChannels
 	public final fun getNewChannelEventFilter ()Lkotlin/jvm/functions/Function3;
 	public final fun getRecoveryNeeded ()Z
 	public final fun getSort ()Lio/getstream/chat/android/client/api/models/QuerySort;
-	public final fun isFirstRequest ()Z
 	public final fun query (IIILkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun query$default (Lio/getstream/chat/android/offline/querychannels/QueryChannelsController;IIILkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
 	public final fun refreshChannel (Ljava/lang/String;)V
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
@@ -647,7 +647,7 @@ internal class ChatDomainImpl internal constructor(
             )
             val response = queryChannelController.runQueryOnline(pagination)
             if (response.isSuccess) {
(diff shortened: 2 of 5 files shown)


stream-chat-android-offline/api/stream-chat-android-offline.api
```diff
@@ -443,6 +443,7 @@ public final class io/getstream/chat/android/offline/querychannels/QueryChannels
 	public final fun getNewChannelEventFilter ()Lkotlin/jvm/functions/Function3;
 	public final fun getRecoveryNeeded ()Z
 	public final fun getSort ()Lio/getstream/chat/android/client/api/models/QuerySort;
+	public final fun isFirstRequest ()Z
```
zsmb13, line 446: Does this have to be public API?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
```diff
@@ -59,6 +59,9 @@ public class QueryChannelsController internal constructor(
     ) : this(filter, sort, client, domain as ChatDomainImpl)
 
     private var channelOffset = INITIAL_CHANNEL_OFFSET
+    internal var isFirstRequest: Boolean = true
```
ogkuzmin, line 62: I wouldn't add this new state since it makes current spaghetti logic more complicated. Can we solve this issue without it on UI side? In VM for example?
Also one possible solution that we shouldn't trigger any new load more request in offline mode (until we got online result of previous request)

leandroBorgesFerreira: The controller is the one changing the value of `isFirstRequest`, then we would create a bond between the two because ViewModel would need to update this field when update the channel list. We would also need to track `channelOffset` and `isFirstRequest` in two different places... that would also be complicated. 

Making online data wait before fetching offline data wouldn't solve the problem =|. The problem is that:
 `channelOffset` is only updated when more online data is successfully fetched. That may or may not happen, as the request can fail. If the request fails, we pick channels from the database and populate our channels... in this case, `channelOffset` is still 0, but our channel list is populated. Then the next request is treated as it was the first-page request because it is the first page **online** request.

In resume, waiting for the online that doesn't work because the request can fail =|.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -639,15 +639,15 @@ internal class ChatDomainImpl internal constructor(
...
+                QuerySort(),
                 INITIAL_CHANNEL_OFFSET,
                 CHANNEL_LIMIT,
                 MESSAGE_LIMIT,
                 MEMBER_LIMIT
             )
             val response = queryChannelController.runQueryOnline(pagination)
             if (response.isSuccess) {
-                queryChannelController.updateOnlineChannels(response.data(), pagination.isFirstPage)
+                queryChannelController.updateOnlineChannels(response.data(), queryChannelController.isFirstRequest)
```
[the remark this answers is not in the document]
leandroBorgesFerreira: 🤔  Makes sense. What about `queryChannelController.updateOnlineChannels(response.data(), true)`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
ogkuzmin: I didn't mean expose some additional data from `QueryChannelsController` to VM. I think your issue can be solved easily in UI part only. I prepared this PR #2404. As I understood your problem it will solve. Can you have a look and test it a little bit?

leandroBorgesFerreira:
Let's continue with #2404

PR #2404, opened 2021-09-29
[CAS-1307] Do not show shimmer if we have some channels already

ogkuzmin:
### 🎯 Goal

Alternative solution for #2388 

### 🛠 Implementation details

Do not show loading view if we have already some channels in state of VM

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
@@ -640,7 +640,7 @@ internal class ChatDomainImpl internal constructor(
             .take(3)
         for (queryChannelController in queriesToRetry) {
             val pagination = QueryChannelsPaginationRequest(
-                QuerySort<Channel>(),
+                QuerySort(),
                 INITIAL_CHANNEL_OFFSET,
                 CHANNEL_LIMIT,
                 MESSAGE_LIMIT,
@@ -648,7 +648,7 @@ internal class ChatDomainImpl internal constructor(
             )
             val response = queryChannelController.runQueryOnline(pagination)
             if (response.isSuccess) {
-                queryChannelController.updateOnlineChannels(response.data(), pagination.isFirstPage)
+                queryChannelController.updateOnlineChannels(response.data(), true)
                 updatedChannelIds.addAll(response.data().map { it.cid })
             }
         }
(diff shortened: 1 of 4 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -640,15 +640,15 @@ internal class ChatDomainImpl internal constructor(
...
+                QuerySort(),
                 INITIAL_CHANNEL_OFFSET,
                 CHANNEL_LIMIT,
                 MESSAGE_LIMIT,
                 MEMBER_LIMIT
             )
             val response = queryChannelController.runQueryOnline(pagination)
             if (response.isSuccess) {
-                queryChannelController.updateOnlineChannels(response.data(), pagination.isFirstPage)
+                queryChannelController.updateOnlineChannels(response.data(), true)
```
leandroBorgesFerreira, line 651: the `true` doesn't make difference, but I believe it makes it more obvious.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
```diff
@@ -306,14 +306,14 @@ public class QueryChannelsController internal constructor(
...
      *
      * @param channels The list of channels to update.
-     * @param isFirstPage If it's the first page we set/replace the list of results. if it's not the first page we add to the list.
+     * @param isFirstOnlinePage If it's the first page we set/replace the list of results. if it's not the first page we add to the list.
      *
      */
     internal suspend fun updateOnlineChannels(
         channels: List<Channel>,
-        isFirstPage: Boolean,
+        isFirstOnlinePage: Boolean,
```
leandroBorgesFerreira, line 314: This is not really the `isFirstPage`, it is actually the first online page... this name makes it easier to understand.

leandroBorgesFerreira (approved):
Good idea to use value of _channels instead of keeping the state in the QUerryChannelsController, in the end _channels already holds the information about `isFirstPage`.

I made changes to the PR, because just using _channels is enough... no need to touch the viewmodels.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
leandroBorgesFerreira: I changed my mind, it is already implicit that this is online by the name of the method updateOnlineChannels

PR #2407, opened 2021-09-29
[Compose] - Overall Attachments API improvements and cleanup

### LINKED ISSUE #2237: [Compose] Editing a message removes attachments
**Describe the bug**

Editing a message removes its attachments.

**SDK version**
- 4.16.0

**To Reproduce**
1. Send a message containing attachments
2. Start an edit
3. Modify the text
4. Send the new message
5. Message no longer has its attachments

**Expected behavior**
Attachments are not modified

filbabic:
Fixes #2237.

Also covers more general improvements that we planned around attachments.

### 🎯 Goal

We wanted to make the Attachments API as easy and intuitive to use, so we removed multiple instances where attachments required extra structure or weren't exposing parameters/properties in a way that suits Composable functions.

We also had to fix bugs with editing messages, where the user would be able to "add/remove" attachments from a given message, which ultimately is not yet supported in our API.

### 🛠 Implementation details

- Removed integrations when the user is editing messages
- Fixed a bug where editing a message with attachments causes it to lose attachments
- Improved various smaller parts of AttachmentsState and AttachmentFactory API

### 🎨 UI Changes

| Before | After |
| --- | --- |
| before | after |

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,12 +67,20 @@
 
 ## stream-chat-android-compose
 ### 🐞 Fixed
+- Fixed a bug where attachments weren't properly stored when editing a message
 
 ### ⬆️ Improved
+- General improvements in the Attachments API and the way we build different attachments
+- Allowed for better long clicks on attachments
 
 ### ✅ Added
 
 ### ⚠️ Changed
+- Removed AttachmentPicker option when editing messages
+- Removed Attachment previews when editing messages with attachments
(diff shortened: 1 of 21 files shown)


filbabic:
This seems like a chonky PR but it's mostly smaller changes across many attachment related files.

samiuelson (approved):

PR #2409, opened 2021-09-29
[CAS-1304] POC - Attachment recycling

leandroBorgesFerreira:
[CAS-1304](https://stream-io.atlassian.net/browse/CAS-1304)

### 🎯 Goal

POC for attachment recycling. This doesn't cover all the cases, as it is just a POC.

### 🛠 Implementation details

Instead of adding programmatically in our MessageItem, now message items have a recycler view to hold our attachments. Attachments are handled as AttachmentGroup, because we can separate them into groups (like an image gallery) or show them one by one (like our file attachments).

The recycler views share the same RecycledViewPool. 

Users can implement their own view holders and handle attachments in their own way, making it customizable. The view holders have `bind` and `unbind` methods to handle recycling as we do for other views. 

### 🧪 Testing

Check the attachments. It contains small UI problems, which can be fixed in the final version of this proposal. 

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
giphy


zsmb13:
Converted to a draft so that we don't accidentally think to merge the POC 😄

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/TextAndAttachmentsViewHolder.kt
```diff
@@ -80,10 +86,25 @@ internal class TextAndAttachmentsViewHolder(
...
 
     private fun setupAttachment(data: MessageListItem.MessageItem) {
-        with(binding.attachmentsContainer) {
-            removeAllViews()
-            addView(attachmentViewFactory.createAttachmentView(data, listeners, style, binding.root))
+        attachmentViewHolderFactory.setUp(data.message.attachments)
+        adapter = FileAttachmentsAdapter(attachmentViewHolderFactory)
+
+        binding.run {
+            attachmentsRecycler.layoutManager = LinearLayoutManager(context)
```
zsmb13, line 93: This might be useful to set somewhere around here https://developer.android.com/reference/androidx/recyclerview/widget/LinearLayoutManager#setRecycleChildrenOnDetach(boolean)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
```diff
@@ -108,8 +112,9 @@ public open class MessageListItemViewHolderFactory {
             decoratorProvider.decorators,
             listenerContainer,
             markdown,
-            attachmentViewFactory,
-            style
+            attachmentViewHolderFactory,
+            mediaAttachmentsRecycledViewPool,
+            fileAttachmentsRecycledViewPool
```
zsmb13, line 117: This doesn't look like it'd scale very well, users would have to add new RecycledViewPools for their own custom types of attachments, wouldn't they?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/attachment/AttachmentViewHolderFactory.kt
```diff
@@ -0,0 +1,18 @@
+package io.getstream.chat.android.ui.message.list.adapter.viewholder.attachment
+
+import android.view.ViewGroup
+import io.getstream.chat.android.client.models.Attachment
+import io.getstream.chat.android.ui.common.adapters.SimpleListAdapter
+
+public interface AttachmentViewHolderFactory {
```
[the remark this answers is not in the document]
zsmb13: I don't see how this interface with two methods for media and files would allow any flexibility for clients to add their own types of attachments

zsmb13:
We'll move ahead with #2490 instead of this one

PR #2415, opened 2021-09-30
Update Kotlin version to 1.5.30, Compose to 1.0.3

zsmb13:
### 🎯 Goal

Update dependencies to the latest available version.

### 🧪 Testing

Smoke test the app.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/MessageRepository.kt
```diff
@@ -66,6 +66,7 @@ internal class MessageRepositoryImpl(
                 Pagination.LESS_THAN -> {
                     return messageDao.messagesForChannelOlderThan(cid, messageLimit, messageTime)
                 }
+                null -> TODO("Is this just invalid?")
```
zsmb13, line 69: Looking for some input on this one

JcMinarro: Can't we do `messageFilterDirection` property "Not Nullable"? :thinking:

zsmb13: Looks like we use its `null` value in other places, so that doesn't seem like a possibility.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
```diff
@@ -1121,6 +1136,22 @@ public class ChannelController internal constructor(
...
+            is ConnectingEvent,
+            is DisconnectedEvent,
+            is ErrorEvent,
+            is GlobalUserBannedEvent,
+            is GlobalUserUnbannedEvent,
+            is HealthEvent,
+            is NotificationMutesUpdatedEvent,
+            is UnknownEvent,
+            is UserDeletedEvent,
+            -> Unit // Ignore these events
```
filbabic, line 1154: Oh why 😢

zsmb13: I didn't want to add an `else` case here, because we handle *most* events, just ignore some specific ones. This way if we add a new one, we're forced to handle it explicitly here.

filbabic (approved):
Looks good to me overall!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/channel/ChannelController.kt
filbabic: Yeah that's good - it's better to be specific, it's just that we really have a lot of different events and it's hard to keep track of them.

adasiewiczr (approved):

PR #2416, opened 2021-09-30
[CAS-1348] Use `cancel-in-progress` option into our GH Actions

JcMinarro:
### 🎯 Goal
Cancel redundant in progress Workflows

### ☑️ Checklist

- [x] Reviewers added


.github/workflows/apk-s3-distribute.yml
```diff
@@ -4,6 +4,10 @@ on:
   release:
     types: [released]
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
```
zsmb13, line 10: We probably don't need to apply these to the distribute workflows, right?

.github/workflows/publish-snapshot.yml
```diff
@@ -5,6 +5,10 @@ on:
     branches:
       - develop
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
```
zsmb13, line 11: I'd also skip using these for these kinds of publishing jobs.

Maybe we should only do this for PR checks? It doesn't seem super useful for other workflows, and if a release/publish kind of workflow is cancelled while it's in the middle of execution (I don't know how cancellation works exactly), that could leave things in a weird state.

.github/workflows/apk-s3-distribute.yml
JcMinarro: As far as we only want the last compilation, we can add it.
If there is a new commit that arrives to this branch, the current job is not needed anymore and we want the new one to be run ASAP

.github/workflows/publish-snapshot.yml
JcMinarro: It is canceling the previous job (The same name) on the current branch, but always run the last one. So we shouldn't leave things in a weird state

.github/workflows/apk-s3-distribute.yml
zsmb13: For example with releases to Sonatype, we might create a repo and start uploading files to it. If we cancel the job and the process gets killed off, we'd leave non-closed staging repos that have some random amount of release files in them, wouldn't we?

Plus it seems like it wouldn't be a huge win to cancel something like a snapshot release. The big gains would come from pretty much just the PR workflows (maybe the `develop` builds, but those merges can't really happen quickly enough to cancel each other because there has to be at least one PR workflow run inbetween anyway). Isn't it enough to add cancellation for that workflow?

JcMinarro: They shouldn't go so quickly, so, it shouldn't be canceled never.
BTW, I am going to remove the one from the publish-snapshot, because keep the repository opened could gives us some issues (following snapshots not published, Manually close it....)
For the rest of them I think it should be fine. Maybe it is never canceled, but if it is canceled no inconsistent data is generated

zsmb13: I'd prefer adding it only where we see that it actually makes sense and provides a benefit. Otherwise simpler > better in this case, isn't it?

I can imagine similar weird leftover things in other cases, in Firebase app distribution as well for example.

zsmb13 (approved):

PR #2429, opened 2021-10-04
 Add Kdocs on `ChatClient.Builder` methods, deprecate `cdnUrl`

zsmb13:
### 🎯 Goal

Make `ChatClient.Builder` methods more obvious.

More info on `cdnUrl` deprecation: [Slack discussion](https://getstream.slack.com/archives/CE5N802GP/p1633338690250200)

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -1703,6 +1703,11 @@ public class ChatClient internal constructor(
             return this
         }
 
+        @Deprecated(
+            message = "Do not use this method for file upload customization. Instead, implement the FileUploader interface and use the fileUploader method of this builder.",
+            level = DeprecationLevel.ERROR,
+            replaceWith = ReplaceWith("this.fileUploader(CustomFileUploader())")
+        )
         public fun cdnUrl(value: String): Builder {
             var cdnUrl = value
             if (cdnUrl.startsWith("https://")) {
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1574,21 +1580,61 @@ public class ChatClient internal constructor(
...
             this.notificationsHandler = notificationsHandler
             return this
         }
 
+        /**
+         * Sets a custom file uploader implementation that will be used by the client
+         * to upload files and images.
+         *
+         * The default implementation uses Stream's own CDN to store these files,
+         * which has a 20 MB upload size limit.
```
bychkovdmitry, line 1632: We need to review this ASAP, I'm almost sure clients will still not be able to send files via UI components as this limit is hardcoded there.

bychkovdmitry (approved):
Great job!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
zsmb13: I believe this is handled with the [`streamUiAttachmentMaxFileSizeMb`](https://github.com/GetStream/stream-chat-android/blob/6e6354373b6011dedc4b152f4aac04bd68fe05f5/stream-chat-android-ui-components/src/main/res/values/attrs_message_input_view.xml#L94) attribute of `MessageInputView`. That value gets passed into `MessageInputFieldView`, which seems to use it correctly.

Am I missing something else? Should we also handle this somewhere else in the code?

bychkovdmitry: Sorry, you are right. Just always get confused about this line in our adapter:
```
internal var attachmentMaxFileSize: Long = AttachmentConstants.MAX_UPLOAD_FILE_SIZE
```

zsmb13: Ah yes, that's just by default, but it should be updated correctly when it's overridden.

PR #2433, opened 2021-10-04
Add possibility to specify newChannelEventFilter in ChannelListView constructor

adasiewiczr:
### 🎯 Goal

Allow overriding `QueryChannelController::newChannelEventFilter`

### 🛠 Implementation details

Added `newChannelEventFilter` parameter to `ChannelListViewModel` (null by default).
**NOTE:** Ideally we should make `newChannelEventFilter` immutable and pass it to `QueryChannelController` by the constructor, but it's not an easy change and produces a breaking change so I think we should postpone it until we merge ChatDomain into LLC

### 🧪 Testing

Try to create `ChannelListViewModel `with  `newChannelEventFilter` (for example `{_, _ -> false}`). You shouldn't be able to see now channels after creating them.

### ☑️ Checklist

- [x] Reviewers added


JcMinarro (approved):

ogkuzmin (changes requested):
Let me think about it. I have ideas before, maybe we can do better things here

PR #2435, opened 2021-10-04
[CAS-1344] Show error logs when our SDK is instantiated twice

JcMinarro:
### 🎯 Goal
Show error logs when our SDK is instantiated twice

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -1729,7 +1729,13 @@ public class ChatClient internal constructor(
             this.callbackExecutor = callbackExecutor
         }
 
-        public override fun buildChatClient(): ChatClient {
+        @InternalStreamChatApi
+        @Deprecated(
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1729,12 +1730,20 @@ public class ChatClient internal constructor(
             this.callbackExecutor = callbackExecutor
         }
 
-        public override fun buildChatClient(): ChatClient {
+        @InternalStreamChatApi
+        @Deprecated(
```
adasiewiczr, line 1734: The method name has changed so, I think we don't need this Deprecated annotation here.

zsmb13: We're adding this to scare people from picking this method from autocompletion results (deprecation is visible there immediately, while the `@InternalStreamChatApi` annotation only produces an error when already selected and added to the code.

zsmb13 (approved):

PR #2436, opened 2021-10-04
[CAS-1366] Redraw attachments only when there's a change in diff

leandroBorgesFerreira:
[CAS-1366](https://stream-io.atlassian.net/browse/CAS-1366)

### 🎯 Goal

Don't reload giphys when the attachments don't change

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/135907395-d8cd8615-1b9a-43b3-8145-dbdc53be2460.mov
" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/135907277-f6cd65cf-12ce-49d2-8ca1-0e51597a2b08.mov
" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Make reaction in a giphy

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF


CHANGELOG.md
```diff
@@ -65,12 +65,13 @@
 ## stream-chat-android-ui-components
 ### 🐞 Fixed
 - Fixed bug related to scroll of messages.
-
+- Updating attachments view holder only when att
```
adasiewiczr, line 68:
```suggestion
```

CHANGELOG.md
```diff
@@ -65,12 +65,13 @@
...
+- Updating attachments view holder only when attachments have changed. This fixes a problem with reloading gifs when reactions are added or removed in then.
```
adasiewiczr, line 69:
```suggestion
- Updating attachments view holder only when attachments have changed. This fixes a problem with reloading gifs when reactions are added or removed.
```

adasiewiczr (approved):
We need to review all view holders and make sure we are using diff payloads properly. I'll add a task for that

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/TextAndAttachmentsViewHolder.kt
```diff
@@ -75,7 +75,10 @@ internal class TextAndAttachmentsViewHolder(
         binding.messageText.isVisible = data.message.text.isNotEmpty()
         markdown.setText(binding.messageText, data.message.text)
 
-        setupAttachment(data)
+        if (diff?.attachments == true) {
```
ogkuzmin, line 78: Seems like if diff is null we don't draw attachments. Is it right? Maybe we can make diff not-nullable passing some default empty value for example

leandroBorgesFerreira: The default empty value wouldn't be the original/correct data, which can be confusing. Fixing the nullability of the diff is outside of the scope of this PR. Let's create another PR an investigate why the diff can be null. 

About the draw with null diff, what about drawing it if the diff is null and only not drawing when it is false? This is closer from the current behaviour

bychkovdmitry (approved):

ogkuzmin (approved):

PR #2437, opened 2021-10-05
Task/improve setting new channels filter alternative

ogkuzmin:
Alternative solution for #2433

### 🎯 Goal
Provide the possibility to inject logic of events handling for the query channels flow

### 🛠 Implementation details

Introduce new interface `ChannelEventsHandler` 

### 🧪 Testing

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
@@ -5,6 +5,7 @@ import io.getstream.chat.android.client.api.models.FilterObject
 import io.getstream.chat.android.client.api.models.QuerySort
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.livedata.ChatDomainImpl
+import io.getstream.chat.android.offline.querychannels.ChannelEventsHandler
 import io.getstream.chat.android.offline.querychannels.QueryChannelsController
 
 /**
@@ -32,13 +33,18 @@ public sealed interface QueryChannelsController {
      * Return true to add the channel, return false to ignore it.
      * By default it will simply add every channel for which this event is received
      */
+    // TODO deprecate
     public var newChannelEventFilter: (Channel, FilterObject) -> Boolean
 
     /**
      * When ChannelUpdatedEvent is triggered, if it is true a new query to the server is done to check if the update
      * on the channel match the filter to be added or deleted from the list of channels
      */
+    // TODO deprecate
     public var checkFilterOnChannelUpdatedEvent: Boolean
+
+    // TODO add docs
+    public var channelEventsHandler: ChannelEventsHandler?
 
     /**
      * If the API call failed and we need to rerun this query
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChannelEventsHandler.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChannelEventsHandler.kt
@@ -1,0 +1,62 @@
+package io.getstream.chat.android.offline.querychannels
+
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.client.api.models.FilterObject
+import io.getstream.chat.android.client.api.models.QueryChannelsRequest
+import io.getstream.chat.android.client.call.await
+import io.getstream.chat.android.client.events.ChannelUpdatedByUserEvent
+import io.getstream.chat.android.client.events.ChannelUpdatedEvent
+import io.getstream.chat.android.client.events.HasChannel
+import io.getstream.chat.android.client.events.NotificationAddedToChannelEvent
+import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.Filters
+import io.getstream.chat.android.client.utils.map
+
+internal class DefaultChannelEventsHandler(
+    private val client: ChatClient,
+    private val filter: FilterObject,
+) : BaseChannelEventsHandler() {
+    internal var checkFilterOnChannelUpdatedEvent: Boolean = false
+
+    internal var newChannelEventFilter: suspend (Channel, FilterObject) -> Boolean = { channel, filter ->
+        client.queryChannels(
+            QueryChannelsRequest(
+                filter = Filters.and(
+                    filter,
+                    Filters.eq("cid", channel.cid)
+                ),
+                offset = 0,
+                limit = 1,
+                messageLimit = 0,
+                memberLimit = 0,
+            )
+        ).await()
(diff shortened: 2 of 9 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChannelEventsHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+import io.getstream.chat.android.client.utils.map
+
+internal class DefaultChannelEventsHandler(
+    private val client: ChatClient,
+    private val filter: FilterObject,
+) : BaseChannelEventsHandler() {
+    internal var checkFilterOnChannelUpdatedEvent: Boolean = false
+
+    internal var newChannelEventFilter: suspend (Channel, FilterObject) -> Boolean = { channel, filter ->
+        client.queryChannels(
```
ogkuzmin, line 22: taken from `QueryChannelsController`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChannelEventsHandler.kt
```diff
@@ -0,0 +1,62 @@
...
+import io.getstream.chat.android.client.events.NotificationAddedToChannelEvent
+import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.Filters
+import io.getstream.chat.android.client.utils.map
+
+internal class DefaultChannelEventsHandler(
+    private val client: ChatClient,
+    private val filter: FilterObject,
+) : BaseChannelEventsHandler() {
+    internal var checkFilterOnChannelUpdatedEvent: Boolean = false
```
adasiewiczr, line 19: I think we can remove it and just not check channels on those events by default

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
```diff
@@ -120,9 +109,8 @@ public class QueryChannelsController internal constructor(
     }
 
     internal suspend fun updateQueryChannelSpec(channel: Channel) {
-        if (newChannelEventFilter(channel, filter)) {
+        if (defaultChannelEventsHandler.newChannelEventFilter(channel, filter)) {
```
adasiewiczr, line 112:
```suggestion
        if (eventsHandler.newChannelEventFilter(channel, filter)) {
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
```diff
@@ -58,25 +53,19 @@ public class QueryChannelsController internal constructor(
...
-            )
-        ).await()
-            .map { channels -> channels.any { it.cid == channel.cid } }
-            .let { it.isSuccess && it.data() }
-    }
 
-    public var checkFilterOnChannelUpdatedEvent: Boolean = false
     public var recoveryNeeded: Boolean = false
+    private val defaultChannelEventsHandler: DefaultChannelEventsHandler = DefaultChannelEventsHandler(client, filter)
+    public var channelEventsHandler: ChannelEventsHandler? = null
```
adasiewiczr, line 59: Can we try to make it immutable?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
```diff
@@ -32,14 +33,19 @@ public sealed interface QueryChannelsController {
...
 
     /**
      * When ChannelUpdatedEvent is triggered, if it is true a new query to the server is done to check if the update
      * on the channel match the filter to be added or deleted from the list of channels
      */
+    // TODO deprecate
     public var checkFilterOnChannelUpdatedEvent: Boolean
 
+    // TODO add docs
+    public var channelEventsHandler: ChannelEventsHandler?
```
adasiewiczr, line 47: Is it possible to make it immutable?

adasiewiczr (commented):
I like the solution - looks cleaner than the previous one.
I've added some comments and have one more question - are we ok with exposing the `suspend` function?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
ogkuzmin: It's not right. This old channel event filter is only kept now in `DefaultChannelEventsHandler `.  Maybe we should delete this method completely

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChannelEventsHandler.kt
ogkuzmin: But then the behaviour will be changed in a new version. It keeps old logic. I think we should keep it until we delete `QueryChannelsController:: checkFilterOnChannelUpdatedEvent`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
ogkuzmin: No, since it would require changes in `ChatDomain::queryChannels`. In future maybe we can iterate over it

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/controller/QueryChannelsController.kt
ogkuzmin: The same reason as above

ogkuzmin:
Speaking about suspend functions. Let's have agreement. I don't see problem in exposing it for users, maybe only for Java customers it can be complicated

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChannelEventsHandler.kt
adasiewiczr: I would be surprised if anyone is using those flags/newEventFilter at the moment because it wasn't easy to access them but I'm okay with removing it later

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsController.kt
ogkuzmin: Not exactly. It is used in ChatDomain here (I believe we can use here `addChannel`):
image

And we use old event filter here (IDK what we have to do here):
image

Issue #2405, opened 2021-09-29
Reaction view gets cut


YasserHossam:
**Describe the bug**
In the chat message, If i add more than 6 reactions, The reaction view is getting cut and prevent from showing more than 6 reactions at the message, I think if we can set a custom anchor offset it will be fixed

**SDK version**
- 4.19.0

![image](https://user-images.githubusercontent.com/12577690/135265319-d6ba1aab-7b90-44e3-8592-209e27849ce5.png)

![image](https://user-images.githubusercontent.com/12577690/135265420-dfb99f4a-9bf4-4e9f-9d54-e00d0debf5a3.png)

bychkovdmitry:
Hello @YasserHossam 

Thanks for reaching out.  Indeed, out design is not flexible enough to accommodate 6+ reactions for a particular message. We are going to discuss possible solutions. I'll get back to you shortly with more information.

YasserHossam:
@bychkovdmitry 
There is a proposed solution to this issue, it might help you with it

Here's how the fixed version of the reactions view looks like

<img width="496" alt="Screen Shot 2021-10-06 at 7 28 20 PM" src="https://user-images.githubusercontent.com/12577690/136254361-bf717fe9-eb32-42cd-a030-22664d8af806.png">

PR #2449, opened 2021-10-06
Fix ViewReactionView offset

AhmedNTS:
This is a proposed solution

### 🎯 Goal

Current ViewReactionView does not support reactions up to 9 reaction (which is our case)

### 🛠 Implementation details

What in the PR is that I calculated the total expected width of the ReactionView and compared it with its container view 
if the expected width is bigger than the container view we calculate the difference and this is the dynamic-offset that should be used to apply margin to ReactionSpace view 

### 🎨 UI Changes

UI changes can be found in this PR related opened issue [here](https://github.com/GetStream/stream-chat-android/issues/2405
)

### ☑️ Checklist

- [✅] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [✅] PR targets the `develop` branch
- [✅] Comparison screenshots added for visual changes
- [ ] Reviewers added

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/ReactionsDecorator.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/decorator/internal/ReactionsDecorator.kt
@@ -1,8 +1,11 @@
 package io.getstream.chat.android.ui.message.list.adapter.viewholder.decorator.internal
 
+import android.graphics.Rect
+import android.util.Log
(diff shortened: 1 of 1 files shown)


AhmedNTS:
@samiuelson this PR is targeting an enhancement in ViewReactionView(the bubble over the message in the message-list view) not EditReactionView

samiuelson:
> @samiuelson He is not trying to fix our `EditReactionView` and he is not adding the problem that you mentioned. We can merge this and work on a fix for the problem that you mentioned.

Sure!

Issue #2548, opened 2021-10-29
[Compose] - Add localization support for Channels set of features


filbabic:
### Context

We need to add localization support for our Channels features, for the existing components.

### Change Proposal

Compare and snyc up all the strings between the UI components and the Compose components for Channels. Make sure to add all the languages supported currently in UI components to Compose.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Sync up strings between UI sdks
- [ ] Add missing languages to Compose strings
- [ ] Test to see if these languages work and if the UI behaves correctly

PR #2556, opened 2021-11-01
[Compose] - 2548 - Add localization support for channels set of features

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2548

### 🎯 Goal

Provide translations for the string used in the "channels" set of features.

### 🛠 Implementation details

- Keeping all the strings in a single resource file as there are very few of them
- Not prefixing common strings
- In contrast to the view-based UI components, giving more independence to the inner components (e.g. strings from channel actions dialog are prefixed with the `channel_info` prefix and not with `channel_list` prefix)

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/ChannelsScreen.kt
```diff
@@ -152,9 +152,11 @@ public fun ChannelsScreen(
         if (activeAction is LeaveGroup) {
             SimpleDialog(
                 modifier = Modifier.padding(16.dp),
```
bychkovdmitry, line 154: A bit of duplication here, but that's basically how we want to do it: ensure that strings are customizable independently, for example, for "Leave group" button and a confirmation dialog with "Leave group" title.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/header/ChannelListHeader.kt
```diff
@@ -170,8 +170,8 @@ internal fun DefaultChannelListHeaderAction(
     ) {
         Icon(
```
bychkovdmitry, line 171: Edit seems weird. We cant anything without a context (a selected channel). That's why renamed to "New chat" button.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/EmptyView.kt
```diff
@@ -39,7 +39,7 @@ public fun EmptyView(modifier: Modifier = Modifier) {
         Spacer(Modifier.size(16.dp))
 
         Text(
-            text = stringResource(id = R.string.stream_compose_empty_state_title),
+            text = stringResource(id = R.string.stream_compose_channel_list_empty_title),
             style = ChatTheme.typography.bodyBold,
```
bychkovdmitry, line 43: As discussed, this will be improved later. We need a slightly different wording for this `EmptyView` depending on the place where it is used.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -134,8 +135,8 @@ public fun DefaultMessageHeaderTitle(
     }
 
     val subtitle = when (messageMode) {
```
bychkovdmitry, line 137: This is a bit verbose unfortunately

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelUtils.kt
```diff
@@ -28,7 +28,7 @@ public fun Channel.getDisplayName(): String {
         ?: getUsersExcludingCurrent()
             .joinToString { it.name }
             .takeIf { it.isNotEmpty() }
-        ?: stringResource(id = R.string.stream_compose_channel_list_untitled_channel)
+        ?: stringResource(id = R.string.stream_compose_untitled_channel)
 }
```
bychkovdmitry, line 32: Reused in the message list header. No need to duplicate the string as it is the presentation of an unnamed channel, regardless of the component (channel list, message list header, etc.)

stream-chat-android-compose/src/main/res/values-en/strings.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
```
bychkovdmitry, line 2: Too fee string resources to split into different resource files. We can do it later.

stream-chat-android-compose/src/main/res/values-en/strings.xml
```diff
@@ -0,0 +1,80 @@
...
+    <plurals name="stream_compose_channel_members">
+        <item quantity="one">%1d Member, %2d Online</item>
+        <item quantity="other">%1d Members, %2d Online</item>
+    </plurals>
+    <string name="stream_compose_ok">Yes</string>
+    <string name="stream_compose_cancel">Cancel</string>
+    <string name="stream_compose_waiting_for_network">Waiting for network</string>
+    <string name="stream_compose_untitled_channel">Channel without name</string>
+
+    <string name="stream_compose_message_label">Send a message</string>
```
bychkovdmitry, line 35: Not channel list related strings below, will be sorted our later.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/ChannelsScreen.kt
filbabic: That makes sense to me and I'm happy with it - if by design we want to change what these are, we can do it separately! :]

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/header/ChannelListHeader.kt
filbabic: That makes sense, it's kind of a weird situation currently, because the button doesn't do much at the moment, as we don't support creating a new channel or editing or anything. 

We could always keep it as hidden by default but still expose the slot. 

What do you think?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/common/EmptyView.kt
filbabic: Yeah do make sure to check with Design if we want completely different UI for these placeholders or not. If we do - we can split them up in UI, otherwise we can just expose string/image params for this component!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
filbabic: That's fine, the framework doesn't support quantity resource grabbing so we have to do it like this

stream-chat-android-compose/src/main/res/values-en/strings.xml
filbabic: Also fine - I love that you've split tehm up in regions at least, so it's easier to navigate them in general.

Looks great! :]

filbabic (approved):
Looks great! Just left 1-2 comments with questions, but we're fine as-is

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/header/ChannelListHeader.kt
bychkovdmitry: Yes, every button without default action should be hidden: the "new chat" button in the toolbar, the "view info" button in the channel info dialog, etc.

Issue #2456, opened 2021-10-07
[Compose] - Add proper placeholder for the case when search results are empty


bychkovdmitry:
### Context

We use the same "let's start chatting" placeholder for 2 different cases: 
1. When that are no channels that match our filter
1. When there are no search results for a particular search query

The second use case is inappropriate for this placeholder.

### Change Proposal

Make a separate component for search result list with appropriate empty placeholder.
OR
Introduce a new placeholder in the `ChannelList` component which shows appropriate placeholder when search results are empty.

### Acceptance criteria

When there are no channels that match the query, then an appropriate "no results" placeholder is shown to the user.

PR #2562, opened 2021-11-02
[Compose] - 2456 - Review placeholders

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2456

### 🎯 Goal

Implement a distinct placeholder for the case, when there are no channels matching the search query.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
```diff
@@ -261,6 +264,27 @@ public fun Messages(
...
 }
 
+/**
+ * The default empty placeholder that is displayed when there are no messages in the channel.
+ *
+ * @param modifier Modifier for styling.
+ */
+@Composable
+internal fun DefaultMessageListEmptyView(modifier: Modifier) {
+    Column(
```
filbabic, line 274: I don't think we need a column given that we're showing just one item - maybe a `Box` here makes more sense/is a bit less confusing. As with a column you'd expect more elements.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channel/ChannelListViewModel.kt
```diff
@@ -96,6 +96,7 @@ public class ChannelListViewModel(
         viewModelScope.launch {
             searchQuery.combine(queryConfig) { query, config -> query to config }
                 .collectLatest { (query, config) ->
+                    channelsState = channelsState.copy(searchQuery = query)
```
filbabic, line 99: Does it make sense to maybe push this towards the happy path - `is QueryChannelsController.ChannelsState.Result -> {}` in `observeChannels`?

Because this way we might have two state updates and UI prompts (compose should optimize for us but still).

filbabic (changes requested):
Left a couple of comments! There is one thing you might want to change, so please LMK if you decide to do so, so I can drop another review! :]

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channel/ChannelListViewModel.kt
bychkovdmitry: Nice catch, was also thinking about that. I will also cover not so happy path when the network in not available and we get `QueryChannelsController.ChannelsState.OfflineNoResults`

filbabic (approved):
Looks good now! Nice work! :]

Down the line, we could write some more docs about this, but I think it's fine if we don't do it immediately.

Issue #2564, opened 2021-11-03
[Compose] - Add state for Muted channels


filbabic:
### Context

We need to support some UI and state for muted channels, that we don't currently have available.

### Change Proposal

Add the state to the `DefaultChannelItem` when it's muted, to show that to the user.

Update the UI based on the Channel's muted status.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Research UI for muted channels (design)
- [ ] Build new UI for muted channel state
- [ ] Apply the UI when the channel is muted

PR #2570, opened 2021-11-04
[Compose] - 2564 - Add state for muted channels

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2564

### 🎯 Goal

Add corresponding indicator to the muted channels in the channel list.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/list/ChannelItem.kt
```diff
@@ -148,12 +150,14 @@ public fun DefaultChannelItem(
...
  * @param currentUser The currently logged in user, used for data handling.
+ * @param isMuted If the channel is muted for the current user.
  * @param modifier Modifier for styling.
  */
 @Composable
 public fun ChannelDetails(
     channel: Channel,
     currentUser: User?,
+    isMuted: Boolean = false,
     modifier: Modifier = Modifier,
```
filbabic, line 161: We should put the modifier as the first non-optonal argument here if we want to keep the `isMuted` as false by def!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/list/ChannelItem.kt
```diff
@@ -171,7 +175,15 @@ public fun ChannelDetails(
 
         val lastMessageText = channel.getLastMessagePreviewText(currentUser)
 
-        if (lastMessageText.isNotEmpty()) {
+        if (isMuted) {
+            Text(
+                text = "Channel muted",
```
filbabic, line 180: I think we have some UI/Design for this so it would be great if we can pull and see if this is what we want (I think this is what iOS does) and then have it translated in resources.

Also - Do we want to show the mute icon for this?

Generally - let's see what the design has in store for us and what we can do with that!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channel/ChannelListViewModel.kt
```diff
@@ -122,26 +122,28 @@ public class ChannelListViewModel(
...
-                        endOfChannels = controller.endOfChannels.value
+                        channels = emptyList()
                     )
+                    is QueryChannelsController.ChannelsState.Result -> {
+                        channelsState.copy(
+                            isLoading = false,
+                            channels = state.channels,
+                            isLoadingMore = false,
+                            endOfChannels = controller.endOfChannels.value,
+                            mutedChannelIds = mutedChannelIds.toSet()
```
filbabic, line 142: Nice work!

filbabic (commented):
Looks good to me in general!

As per our discussion, seems like the extra data approach might be cleaner, evne though it's easy to miss - so if we can document it somehow in `Channel`, that would be great

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channel/ChannelListViewModel.kt
```diff
@@ -122,26 +122,28 @@ public class ChannelListViewModel(
...
-                QueryChannelsController.ChannelsState.Loading,
-                ->
-                    channelsState.copy(isLoading = true)
-                QueryChannelsController.ChannelsState.OfflineNoResults -> channelsState.copy(
-                    isLoading = false,
-                    channels = emptyList()
-                )
-                is QueryChannelsController.ChannelsState.Result -> {
-                    channelsState.copy(
+        controller.mutedChannelIds.combine(controller.channelsState) { mutedChannelIds, state -> mutedChannelIds to state }
```
adasiewiczr, line 125: Keep in mind I'm going to deprecate `mutedChannelIds` this week (probably tomorrow). It's better to use `user.channelMutes` instead, but I'm not sure if we are updating it properly. I'll let you know once I'll finish the task

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channel/ChannelListViewModel.kt
```diff
@@ -122,26 +123,27 @@ public class ChannelListViewModel(
      * It connects the 'loadingMore', 'channelsState' and 'endOfChannels' properties from the [controller].
      */
     private suspend fun observeChannels(controller: QueryChannelsController) {
```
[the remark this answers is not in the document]
filbabic: Should we mention this in the commnets by any chance, so we don't forget? Link a ticket or something?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channel/ChannelListViewModel.kt
```diff
@@ -253,4 +255,21 @@ public class ChannelListViewModel(
...
+    /**
+     * Uses [Channel.isMuted] property to store additional flag for each channel if the channel is muted
+     * for the current user.
+     *
+     * @param mutedChannelIds The list of channel IDs that represent channels muted for the current user.
+     * @return The list of augmented channels.
+     *
+     * @see [Channel.isMuted]
+     */
+    private fun List<Channel>.combine(mutedChannelIds: List<String>): List<Channel> {
```
filbabic, line 268: I would make this a regular function with two parameters and a more concrete name. 

This can be basically used with any `List<Channel>`and any `List<String>`. I know it's private, but still, the name doesn't say much.

I'd propose something like:

```
private fun enrichMutedChannels(channels: List<Channel>, mutedChannelIds: List<String>): List<Channel>{}
```

I know the signature above might be easier to use and all, but this would be much more clear and easier to understand if a person is coming from outside of this domain (e.g. new people)

filbabic (changes requested):
Overall looks good, but I do have two comments I'd like us to address before merging!

filbabic (approved):

Issue #2565, opened 2021-11-03
[Compose] - Add a Channel display name factory customization option to the app


filbabic:
### Context

We support a way to customize what's shown in the channel name display, in other SDK variants.

The idea is to support a factory/function that transforms a Channel to a String to display the name in a custom way.

### Change Proposal

Add a new property called `channelNameFormatter` and CompositionLocal that provides it. We can build a small class/interface with the type: `ChannelNameFormatter` (similar to `DateFormatter`) and provide a default implementation that just uses the `channel.displayName`.

But our users will be able to provide their own if they want.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Provide a new class/interface used to format the Channel name.
- [ ] Provide a default implementation that uses the `channel.displayName()`
- [ ] Provide the formatter through a `ChatTheme` parameter and give it a default value
- [ ] Expose the formatter through a CompositionLocal

PR #2573, opened 2021-11-04
[Compose] - 2565 - Add channel display name factory

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2565

### 🎯 Goal

Add `Channel` display name factory customization option to the app.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

giphy (2)


filbabic (approved):
Looks good to me! :]

filbabic (changes requested):
One thing I did notice is - we're missin Kdocs for the new provider and parameter in ChatTheme as well as docusaurus.

Can you please update those too! (sorry if I missed them initially)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelNameFormatter.kt
```diff
@@ -0,0 +1,68 @@
...
+ * - If the channel is distinct, then a comma-separated list of member names is returned
+ * - Otherwise, the placeholder text defined in [R.string.stream_compose_untitled_channel] is returned
+ *
+ * @param context The context to load string resources.
+ * @param maxMembers The maximum number of members used to generate a name for a distinct channel.
+ */
+private class DefaultChannelNameFormatter(
+    private val context: Context,
+    private val maxMembers: Int,
+) : ChannelNameFormatter {
```
skydoves, line 53: I'm not sure but how about adding the `@Composable` annotation to the `format` method and removing the `context `argument? We may get the context from the `LocalContext.current` instead.

```kotlin
@Composable
override fun format(channel: Channel): String {
    val context: Context = LocalContext.current
    return channel.name.takeIf { it.isNotEmpty() }
        ?: channel.getUsersExcludingCurrent()
            .joinToString(limit = maxMembers) { it.name }
            .takeIf { it.isNotEmpty() }
        ?: context.getString(R.string.stream_compose_untitled_channel)
}
```

skydoves (commented):
Totally looks good to me! I left one suggestion. :)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelNameFormatter.kt
bychkovdmitry: That's actually a very good question. I actually don't have a strong opinion about that.

filbabic: @skydoves @bychkovdmitry Really nice discussion.

Generally, adding @Composable to functions that return values is not that common and should serve more less only for functions that really do depend on the scope and context of the Compose tree.

E.g. if you're observing state that changes and the function needs to recompose. Or if you're, based on the compose state, lauching extra lifecycle things that affect the Compose tree in some way, like system back click handlers updating the navigation or the active screen.

In this case, although we depend on the Context, nothing mutates within the Compose tree to affect our state or result, as this result will always be the same, regardless of the context object (it's just used to fetch the resource).

So in that sense, I prefer it this way and I think it's good, maybe we could split it up into a smaller function that might be more descriptive, but it's also fine this way!

filbabic (approved):
Looks good!

skydoves (approved):

PR #2581, opened 2021-11-08
[CAS-1303] Make retryPolicy immutable and set via Builder

amitkma:
https://stream-io.atlassian.net/browse/CAS-1303

### 🎯 Goal

The goal of this PR is to make `retryPolicy` immutable and that should be set only with the Builder.

### 🛠 Implementation details

All mutable vars of `retryPolicy` are converted into `val`. A new function `retryPolicy()` is added to Builders to customize the retry policy while creating ChatDomain instance.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
@@ -91,7 +91,7 @@ public sealed interface ChatDomain {
     public val banned: LiveData<Boolean>
 
     /** The retry policy for retrying failed requests */
-    public var retryPolicy: RetryPolicy
+    public val retryPolicy: RetryPolicy
 
     /**
      * Updates about currently typing users in active channels. See [TypingEvent].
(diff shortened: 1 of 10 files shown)


ogkuzmin (commented):
Nice job. One thing. Can you add retryPolicy to `io.getstream.chat.android.offline.experimental.plugin.Config` and set retry policy to the `ChatDomain` in `OfflinePlugin::init`

ogkuzmin (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -91,7 +91,7 @@ public sealed interface ChatDomain {
     public val banned: LiveData<Boolean>
 
     /** The retry policy for retrying failed requests */
-    public var retryPolicy: RetryPolicy
```
zsmb13, line 94: Is there a good reason why we're not just deprecating the setter first, and introducing a breaking change instead?

ogkuzmin: We don't provide any documentation for that. We don't even use it everywhere for retries inside. It's weird also to reset it after ChatDomain is configured already. I have doubts that it's used by anyone currently.

zsmb13: Given that setting that property was previously the only way to set this, I'd be happier if we just deprecated it at first as we introduced the new API that replaces it, but if you feel like this'd be a super niche feature, I'm okay with doing the breaking change too, given that it's clearly highlighted in the changelog

zsmb13 (approved):

PR #2594, opened 2021-11-09
Init ToggleService by default in ChatClient.Builder

ogkuzmin:
### 🎯 Goal

Avoid NPE when SDK users don't init ToggleService

### 🛠 Implementation details

Init by default in ChatClient.Builder

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -88,6 +88,7 @@ import io.getstream.chat.android.client.user.storage.UserCredentialStorage
 import io.getstream.chat.android.client.utils.ProgressCallback
 import io.getstream.chat.android.client.utils.Result
 import io.getstream.chat.android.client.utils.TokenUtils
+import io.getstream.chat.android.client.utils.internal.toggle.ToggleService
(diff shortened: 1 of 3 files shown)


amitkma (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1823,6 +1824,10 @@ public class ChatClient internal constructor(
                 loggerConfig = ChatLogger.Config(logLevel, loggerHandler),
             )
 
+            if (ToggleService.isInitialized().not()) {
```
leandroBorgesFerreira, line 1827: `MessageListViewModel` can be initialized before this line if called. Pehaps can we wait to the initialilzation in a lazy way using an observable?

leandroBorgesFerreira (changes requested):
It still crashes is MessageListViewModel is initialised too soon.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: Nope, this observation doesn't make sense.

leandroBorgesFerreira (approved):
🚢 🚢 🚢

PR #2599, opened 2021-11-10
[CAS-1265] Add replayEventsForActiveChannels extension function

amitkma:
https://stream-io.atlassian.net/browse/CAS-1265

### 🎯 Goal

To merge ChatDomain in ChatClient

### 🛠 Implementation details

Added an extension function `replayEventsForActiveChannels` to the ChatClient so end users don't need ChatDomain.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -144,6 +144,11 @@ public sealed interface ChatDomain {
      * @return Executable async [Call] responsible for obtainging list of historical [ChatEvent] objects.
      */
     @CheckResult
+    @Deprecated(
+        message = "replayEventsForActiveChannels is deprecated. Use extension function ChatClient::replayEventsForActiveChannels instead",
+        replaceWith = ReplaceWith("ChatClient.replayEventsForActiveChannels()"),
```
skydoves, line 149: How about updating the `expression` of `ReplaceWith` explicitly and adding an `import` parameter like the below?
We might prevent breaking client codes when using the `Replace with` on IDE.

```kotlin
replaceWith = ReplaceWith(
            expression = "ChatClient.instance().replayEventsForActiveChannels(cid)",
            imports = arrayOf("io.getstream.chat.android.client.ChatClient")),
```

### replace with
replace

### before
I can see still breaking imports and parameters (cid) after using the `Replace with`.

before

### after
we can resolve this by adding imports and updating the `expression` explicitly with parameters.

after

amitkma: I absolutely agree here. We should have it for all our deprecated functions.

ogkuzmin (approved):

DEPRECATIONS.md
```diff
@@ -4,6 +4,7 @@ This document lists deprecated constructs in the SDK, with their expected time 
 
 | API / Feature | Deprecated (warning) | Deprecated (error) | Removed | Notes |
 | --- | --- | --- | --- | --- |
+| `ChatDomain#replayEventsForActiveChannels` | 2021.11.10<br/>4.21.0 | 2021.11.10<br/>⌛ | 2021.11.10 ⌛ | Use `ChatClient#replayEventsForActiveChannels` instead |
```
ogkuzmin, line 7: If you introduce deprecation you also should define periods when it gets deprecated with error and removed completely. Usually it 1 month, but can be two weeks. Read more [here](https://github.com/GetStream/InternalAndroidGuidelines#deprecations)

skydoves (approved):
Looks good to me!

DEPRECATIONS.md
amitkma: Yes, but it has to be updated when we do a release. Isn't it?

PR #2605, opened 2021-11-11
[Compose] Improve the design of ChannelInfo dialog

bychkovdmitry:
### 🎯 Goal

Improve the design of `ChannelInfo` dialog

### 🛠 Implementation details

Tiny fix to the design of the channel actions dialog.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| before-one | after-one |
|before-group | after-group|

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
```diff
@@ -51,58 +47,63 @@ import io.getstream.chat.android.compose.state.channel.list.LeaveGroup
...
-    }
 
     val canLeaveChannel = !selectedChannel.isDistinct()
-
-    val canDeleteChannel = selectedChannel.members
-        .firstOrNull { it.user.id == user?.id }
+    val canDeleteChannel = channelMembers.firstOrNull { it.user.id == currentUser?.id }
         ?.role
         ?.let { it == "admin" || it == "owner" }
         ?: false
```
bychkovdmitry, line 78: We can later extract this logic and reuse it for `MessageListHeader` subtitle.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
```diff
@@ -111,15 +112,15 @@ public fun ChannelInfo(
             ChannelOption(
                 title = stringResource(id = R.string.stream_compose_channel_info_delete_conversation),
                 titleColor = ChatTheme.colors.errorAccent,
```
bychkovdmitry, line 114: This should be `painterResource`. If minor breaking change to the `ChannelOption` class is possible at this stage, then better to do it now.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/UserUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+        return context.getString(R.string.stream_compose_user_status_online)
+    }
+
+    return (lastActive ?: updatedAt ?: createdAt)?.let {
+        if (it.isInLastMinute()) {
+            context.getString(R.string.stream_compose_user_status_last_seen_just_now)
+        } else {
+            context.getString(
+                R.string.stream_compose_user_status_last_seen,
+                DateUtils.getRelativeTimeSpanString(it.time).toString()
```
bychkovdmitry, line 30: This line copied from UI components and looks suspicious to me. There can be a mixture of locales. For example: "Last seen 5 минут назад"

stream-chat-android-compose/src/main/res/drawable/stream_compose_ic_person_remove.xml
```diff
@@ -0,0 +1,22 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="25dp"
+    android:height="24dp"
+    android:viewportWidth="25"
+    android:viewportHeight="24"
+    >
+    <path
+        android:fillColor="#000000"
```
bychkovdmitry, line 8: Need to decide, which is preferable for us:
- Black icons + required tint set externally with `iconTint`
- Icons with embedded color

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
filbabic: You mean we should replace the icon with a painter? Yeah sure, you can do that, but make sure to add it to the changelog! :]

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
```diff
@@ -154,42 +155,55 @@ public fun ChannelInfo(
...
-                    horizontalArrangement = Arrangement.spacedBy(12.dp),
-                    contentPadding = PaddingValues(start = 16.dp, end = 16.dp)
-                ) {
-                    items(selectedChannel.members) { member ->
-                        ChannelInfoUserItem(member = member)
-                    }
-                }
+                ChannelMembers(otherMembers)
 
                 ChannelOptions(channelOptions, onChannelOptionClick)
```
filbabic, line 172: @bychkovdmitry One things I noticed is missing is "muting channels" - should we add this option to our list of channel options?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
```diff
@@ -154,42 +155,55 @@ public fun ChannelInfo(
...
+        modifier = modifier
+            .fillMaxWidth()
+            .padding(vertical = 24.dp),
+        horizontalArrangement = Arrangement.Center,
+        contentPadding = PaddingValues(start = 16.dp, end = 16.dp)
+    ) {
+        items(members) { member ->
+            ChannelInfoUserItem(
+                modifier = Modifier
+                    .width(80.dp)
```
filbabic, line 199: Pull out into ChatTheme.dimens maybe?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
```diff
@@ -204,15 +218,13 @@ private fun ChannelInfoUserItem(
...
-            .width(48.dp),
+        modifier = modifier,
         verticalArrangement = Arrangement.Center,
         horizontalAlignment = Alignment.CenterHorizontally
     ) {
 
         UserAvatar(
-            modifier = modifier
-                .size(48.dp),
+            modifier = Modifier.size(64.dp),
```
filbabic, line 227: Dimens as per comment above

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/UserUtils.kt
filbabic: Yeah that makes sense - any way to figure out if we can support the language that the DateUtils supports?

Any way to resolve this problem, in the case of a language missing?

stream-chat-android-compose/src/main/res/drawable/stream_compose_ic_person_remove.xml
filbabic: All icons should have an `iconTint` set externally, because we're using our ChatTheme.

That being said, for the purpose of the icon being visible here and so that we can inspect it, we put the color as black

filbabic (commented):
Added a few comments, but overall, looks great! :]

stream-chat-android-compose/src/main/res/drawable/stream_compose_ic_person_remove.xml
bychkovdmitry: Ok, for view-based UI components we decided to embed tint so that the icons can be set externally by the client (can even be a png and not a vector drawable).

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
bychkovdmitry: This looks like a something new. I wasn't able to find this option in any SDK. But yes, we should do it. Will do it later in another PR as this looks like is a big change (handle action in viewmodel, reflect the muted state in the UI with different labels "mute channel"/"unmute channel").

filbabic: Cool!

I'll open a ticket for this and we can focus on it in another PR!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/UserUtils.kt
bychkovdmitry: Don't have a solution at the moment.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
```diff
@@ -51,58 +46,63 @@ import io.getstream.chat.android.compose.state.channel.list.LeaveGroup
...
  *
  * @param selectedChannel The channel the user selected.
- * @param user Currently logged-in user data.
+ * @param currentUser The currently logged-in user data.
  * @param onChannelOptionClick Handler for when the user selects a channel option.
  * @param modifier Modifier for styling.
  * @param shape The shape of the component.
  */
 @Composable
 public fun ChannelInfo(
```
skydoves, line 63: I'm curious about the reason we are using `Box` instead of `BottomSheetScaffold` to show up `ChannelInfo` on the `ChannelsScreen`.

skydoves (commented):
Not related to this PR, I found some weird bug. 
In the `ChannelsScreen`, the `selectChannel` method on `onChannelLongClick` block seems not to work correctly. If we press the back button when the `ChannelInfo` is showing up, the app will be closed instead of invoking `selectChannel(null)` method on `SystemBackPressedHandler`. 

```kotlin
onChannelLongClick = {
  Log.e("Test", "onChannelLongClick: ${it.cid}") // correct
  listViewModel.selectChannel(it)
   Log.e("Test", "selectedChannel: $selectedChannel") // print null
}
```

Should we change the type of the `selectedChannel` in `ChannelListViewModel` as `MutableState<Channel?>` for observing the value?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
bychkovdmitry: Yes, also curious. We need to review attachment picker dialog and channel info dialog to see if we can use `BottomSheetScaffold`.

bychkovdmitry:
"Not related to this PR, I found some weird bug."

Thanks! Looking into the issue...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
filbabic: Generally - I'm trying to avoid using things like the BottomSheets and special types of Scaffolds etc, because I feel like it's very contraining.

With a Box that fills the entire screen as an overlay, as well as the ChannelInfo component (internal content), our users can choose to use modifiers to position the sheet in the center of the screen, at the top or bottom etc.

It's not constraining, while the BottomSheet is and it entails more complexity to the screen than a simple Box/Card/Surface combination.

For the Screen type of components, it doesn't matter much what we use I guess, but if we make the ChannelInfo a BottomSheet and require it to be used within a Scaffold or something, then we'll potentially have limitations for our users.

Within the Screen components however, we could try using higher-level components to make it easier, but we'd have to do it in a way that doesn't force our users to use a specific pattern!

filbabic (approved):

skydoves (approved):

skydoves (approved):
Perfect! 💯

PR #2609, opened 2021-11-11
Fix unit tests

ogkuzmin:
### 🎯 Goal

Fix broken tests in offline module

### 🧪 Testing

Run tests and be sure they are passing

### ☑️ Checklist

- [x] Reviewers added


amitkma (changes requested):
This file should not be presented.

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/ChatDomainImplCreateChannelTest.kt
```diff
@@ -205,25 +201,6 @@ internal class ChatDomainImplCreateChannelTest {
             }
         }
 
-    @Test
-    fun `Given failed network response When create distinct channel Should return failed response And do not insert any channel to DB`() =
```
JcMinarro, line 209: Why this test has been removed?
Have we change our logic?
Can't we configure this tests to follow the same approach?
If the new logic produce new results, can't we verify the new results are as expected?

amitkma (approved):

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/ChatDomainImplCreateChannelTest.kt
ogkuzmin: We completely removed this method from offline module. And test was a kind of dummy. Check if network request failed we didn't do anything and returned failed result.

JcMinarro (approved):

PR #2610, opened 2021-11-11
Fix unit test script to return test command exit code

amitkma:
### 🎯 Goal

To set GitHub action's status according to Gradle tests command output.

### 🛠 Implementation details

Returned from the script with exit code of tests command. 

### 🧪 Testing

Make sure GitHub action fails for failing unit tests. 

### ☑️ Checklist

- [ ] Reviewers added


stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/integration/BaseConnectedIntegrationTest.kt
```diff
@@ -71,10 +71,10 @@ internal open class BaseConnectedIntegrationTest : BaseDomainTest() {
             backgroundSyncEnabled,
             context,
             offlinePlugin = plugin,
+            retryPolicy = NoRetryPolicy()
```
JcMinarro, line 74: Why this change is needed?

scripts/ci-unit-tests.sh
```diff
@@ -4,4 +4,10 @@ command_file="unit-tests-command.sh"
 chmod +x $command_file && \
 ./$command_file
 
+# Exit code of gradle test command output
+exitCode=$?
+
 rm $command_file
+
+# Exit with test command output to set the GH action's check run status
+exit $exitCode
```
JcMinarro, line 13: We don't really need to store the previous command exit code. 
As far as we are concatenated command on a single line and using `&&`, whenever a command doesn't return `0` (It was run properly without errors), the following commands are not run
```suggestion
./$command_file && \
rm $command_file
```

JcMinarro (changes requested):

scripts/ci-unit-tests.sh
amitkma: Ah good suggestions. Then the last command to delete the temporary `unit-tests-command` won't be deleted. Will it be a problem?

JcMinarro: It shouldn't. It will be there on the docker machine that is created, but because on every CI-Job the docker machine is a new one, it is not going to affect

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/integration/BaseConnectedIntegrationTest.kt
amitkma: Not needed anymore as it will be done by Oleg's PR.

JcMinarro (approved):

ogkuzmin (approved):

PR #2612, opened 2021-11-11
PoC ComposerViewModel merge

filbabic:
Improvements around #1423

### 🎯 Goal

More work done on the Compose part of the migration, to integrate the new ViewModel instead of the old one.

### 🛠 Implementation details

- Removed the previous MessageComposerViewModel and replaced it with the new, shared, MCVM.
- Updated the accessors and mutability of the properties within the MessageComposerViewModel.
- Exposed more state for the Compose UI to listen to.
- Fixed an issue with the InternalStreamChatApi being applied to the entire MCVM class, instead of the constructor
- Updated the MessageComposer signature to be more respective of the new changes and easier to use as a customer

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/MessagesActivity.kt
```diff
@@ -164,15 +164,15 @@ class MessagesActivity : AppCompatActivity() {
                 .wrapContentHeight(),
             viewModel = composerViewModel,
             integrations = {},
-            input = {
+            input = { value, attachments, action ->
```
filbabic, line 167: Instead of directly observing state within the component, we can propagate it down to make it fully stateless.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessageComposerViewModel.kt
```diff
@@ -14,14 +14,16 @@ import io.getstream.chat.android.client.models.Attachment
...
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 
 /**
  * ViewModel responsible for handling the composing and sending of messages.
  */
-@InternalStreamChatApi
-public open class MessageComposerViewModel(
+
+public open class MessageComposerViewModel @InternalStreamChatApi constructor(
```
samiuelson, line 26: We'll probably need to provide access to the constructor but remove the `open` modifier. Devs will need to create view model factory to pass data in the constructor (if they aren't using screen components).

samiuelson: And we won't inherit from the view model anywhere hopefully 🤞🏼

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -60,20 +62,24 @@ public fun MessageComposer(
         DefaultComposerIntegrations(onAttachmentsClick)
     },
     label: @Composable () -> Unit = { DefaultComposerLabel() },
-    input: @Composable RowScope.() -> Unit = {
+    input: @Composable RowScope.(String, List<Attachment>, MessageAction?) -> Unit = { value, attachments, action ->
```
samiuelson, line 65: Looks like a good direction to me 👍🏼

samiuelson: Maybe we could make `MesasgeComposer` stateless too, extract `viewModel` param outside, and pass `State` type params instead? I think it's working nice in ui-components that we don't inject view models into ui components and we just bind view model with view inside Fragment, or Activity.

samiuelson:
Maybe, as advised by @zsmb13, it would be safe approach to have separate implementations of view models in both `compose` and `ui-components` modules. 

This way, we could inject a shared class (call it `MessageComposerController`) from `ui-common` into both view models. This shared class would hold current business logic and expose data similarly to the current view model. The two, SDK-specific view models, would act as a wrappers - exposing states and passing actions to the controller instance.

This way we would:
* Avoid breaking changes related to imports
* We would also avoid inheritance in favour of composition
* Write a bit more code then previously planned

filbabic:
@samiuelson @zsmb13 The way I see it, it seems hard not to introduce breaking changes.

If we go with the single ViewModel option (that seems like not something we want), we'd have to introduce import changes and potentially extra state for one of the UI SDKs.

If we go with the option of a Controller, with 2 ViewModels, then we have a ton of duplicate code that we have to introduce, as both of the ViewModels have to either:
- Expose the Controller to users and have them use those functions directly (breaking change)
- Expose all the functions and state as before to not break functionality, relaying to the Controller, which is duplicating code and error prone in case one of the SDKs decides to mutate the behavior/state in a way that's different from what it should be

If we go with Option 1, then we'd mostly be using the Controller and very little the ViewModel, which is confusing as to when you should use which.

If we go with Option 2, then we're just abstracing/reusing a small bit of behavior and state, but we're also duplicating everything, which generally feels like why would we have the controller in the first place.

samiuelson (approved):

PR #2616, opened 2021-11-12
CAS-1423 `ComposerViewModel` code sharing

samiuelson:
https://stream-io.atlassian.net/browse/CAS-1434

### 🎯 Goal

We want to share `ComposerViewModel` code to add new features quickly in both Compose and ui-components 

### 🛠 Implementation details

Business logic of `ComposerViewModel` is extracted into `MessageComposerController` class and moved into `common` module. 

### 🧪 Testing

Compose message input view should work without issues

### ☑️ Checklist

- [x] Reviewers added


filbabic (approved):
This looks good to me overall, I'd love to get a bit more insight from @zsmb13 for example, but I'll approve on my end!

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessageComposerController.kt
```diff
@@ -0,0 +1,246 @@
...
+ * It acts as a central place for both the core business logic and state required to create and send messages, handle
+ * attachments, message actions and more.
+ *
+ * If you require more state and business logic, compose this Controller with your code and apply the necessary changes.
+ *
+ * @param channelId The ID of the channel we're chatting in.
+ * @param chatClient The client used to communicate to the API.
+ * @param chatDomain The domain used to communicate to the API and store data offline.
+ */
+public class MessageComposerController constructor(
```
bychkovdmitry, line 30: This looks more like `MessageComposerViewModelDelegate` to me, but `MessageComposerController` is also fine 👍

bychkovdmitry (approved):
Great job!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -60,20 +62,24 @@ public fun MessageComposer(
         DefaultComposerIntegrations(onAttachmentsClick)
     },
     label: @Composable () -> Unit = { DefaultComposerLabel() },
-    input: @Composable RowScope.() -> Unit = {
+    input: @Composable RowScope.(String, List<Attachment>, MessageAction?) -> Unit = { value, attachments, action ->
```
zsmb13, line 65: > `String, List<Attachment>, MessageAction?`

I'd prefer wrapping these into a data class where they can have proper names, especially a `String` on its own looks really odd.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -23,12 +23,12 @@ import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
+import com.getstream.sdk.chat.state.MessageMode
+import com.getstream.sdk.chat.state.Normal
+import com.getstream.sdk.chat.state.Thread
```
zsmb13, line 28: Any chance we can rename `Thread` at some point? It's very hard to import this, as `java.lang.Thread` is imported for Kotlin users by default.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -23,12 +23,12 @@ import androidx.compose.ui.platform.LocalContext
...
 import io.getstream.chat.android.client.models.Channel
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.compose.R
-import io.getstream.chat.android.compose.state.messages.MessageMode
-import io.getstream.chat.android.compose.state.messages.Normal
-import io.getstream.chat.android.compose.state.messages.Thread
```
zsmb13, line 31: We use this import in our tutorial, let's make sure we roll out a speedy update of the tutorial after these changes get shipped in a new version

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageComposerViewModel.kt
```diff
@@ -149,46 +87,24 @@ public class MessageComposerViewModel(
...
-            dismissMessageActions()
-            sendMessageCall.enqueue()
-        }
-        clearData()
-    }
+    public fun sendMessage(message: Message): Unit = messageComposerController.sendMessage(message)
 
     /**
-     * Builds a new [Message] to send to our API. If [isInEditMode] is true, we use the current
+     * Builds a new [Message] to send to our API. Based on the [MessageComposerController] state, we use the current
```
zsmb13, line 104: `MessageComposerController` should probably not show up in any of the docs of this class. It's an implementation detail, users should not know that it exists as far as I understand.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessagesViewModelFactory.kt
```diff
@@ -5,14 +5,15 @@ import android.content.Context
...
+import com.getstream.sdk.chat.viewmodel.MessageComposerController
 import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.compose.handlers.ClipboardHandlerImpl
 import io.getstream.chat.android.compose.ui.util.StorageHelperWrapper
 import io.getstream.chat.android.offline.ChatDomain
 
 /**
  * Holds all the dependencies needed to build the ViewModels for the Messages Screen.
- * Currently builds the [MessageComposerViewModel], [MessageListViewModel] and [AttachmentsPickerViewModel].
+ * Currently builds the [MessageComposerController], [MessageListViewModel] and [AttachmentsPickerViewModel].
```
zsmb13, line 16: We could probably stick with the old wording here for the docs - the controller doesn't need to be mentioned publicly, and the `MessageComposerViewModel` is still being built just like before

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessageComposerController.kt
zsmb13: Isn't this supposed to be `@InternalStreamChatApi`? Do we ever want users to instantiate or call methods of this class?

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/state/MessageMode.kt
```diff
@@ -1,4 +1,4 @@
-package io.getstream.chat.android.compose.state.messages
+package com.getstream.sdk.chat.state
```
zsmb13, line 1: We really need to fix these packages names in the common module at some point. Perhaps we should already have these in an `io.getstream.*` package even if old code in the common module is in `com.getstream`. At least not to add new code under the old package name.

samiuelson: Should the base package name in the common module (for `MessageComposerController`, `MessageMode`, and `MessageAction` classes)  be `io.getstream.chat.android` or `io.getstream.chat.android.common`? cc @filip

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
samiuelson: Yup, it forces manual import. Maybe it could be a `TreadReply`?  cc @filbabic

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessageComposerController.kt
filbabic: To call methods perhaps - they can reuse the core behavior if they want, without having to use the client directly I guess. 

They might also choose to not use ViewModels (as in Jetpack), so having a Controller they can wrap in their own presenter or Repo could be useful.

If you feel like it's not supposed to be exposed, we can always block it off!

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/state/MessageMode.kt
filbabic: Yeah that makes sense, we can move all the code to io.getstream.*, I'll update this and then we can review

zsmb13: I think `io.getstream.chat.android.ui` or `io.getstream.chat.android.ui.common` would be appropriate for the package name here. The former is simpler, maybe we can go with that.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/viewmodel/MessageComposerController.kt
zsmb13: The fewer things we can expose as public API, the happier I am 😃 I'd like to keep this internal for the time being, and only expose it if we really, really need to. Integrating against the ViewModel should be enough for users. If they need low-level, they can use `ChatClient`. This feels like we're providing them too many options, so it'll just get complicated for them to choose between. Plus it's not consistent, we don't have such controllers available for other components.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -12,44 +12,51 @@ import io.getstream.chat.android.client.models.Channel
...
 import kotlinx.coroutines.launch
 import java.util.Date
 import java.util.concurrent.TimeUnit
 
 /**
  * ViewModel responsible for handling all the business logic & state for the list of messages.
+ *
+ * @param chatClient Used to connect to the API.
+ * @param chatDomain Used to connect to the API and fetch the domain status.
+ * @param channelId Used to connect to the required channel.
```
zsmb13, line 56: Lines like this feel like documenting implementation details instead of helping users with what they need to provide. 

As a user, something along the lines of "The ID of the channel to load messages for" would be more useful.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/components/MessageInput.kt
```diff
@@ -89,24 +90,21 @@ public fun MessageInputOptions(
...
  * @param onValueChange Handler when the value changes.
  * @param onAttachmentRemoved Handler when the user removes a selected attachment.
  * @param modifier Modifier for styling.
  * @param label Composable function that represents the label UI, when there's no input/focus.
  */
 @Composable
 public fun MessageInput(
-    value: String,
-    attachments: List<Attachment>,
-    activeAction: MessageAction?,
```
zsmb13, line 104: I feel like we're losing something here by making the stateless component take a single state parameter instead of the individual ones. It was neater the other way imo (having `value` / `onValueChange` especially).

stream-chat-android-ui-common/src/main/java/io/getstream/chat/android/common/state/MessageListAction.kt
```diff
@@ -1,4 +1,4 @@
-package io.getstream.chat.android.compose.state.messages.list
+package io.getstream.chat.android.common.state
```
zsmb13, line 1: - These new `io.getstream.chat.android.common` files are under the `java` source folder instead of `kotlin`
- For this specific file, maybe the `MessageAction` name would be more appropriate, to match the name of the class

Issue #2626, opened 2021-11-15
[Compose] - Show correct information in the MessageListHeader subtitle


bychkovdmitry:
### Context

The subtitle in the `MessageListHeader` shows incorrect information about the members of the channel. 

### Change Proposal

The subtitle in the `MessageListHeader` should show the exact same information as in the `ChannelInfo` component: either member count for a group channel, or user presence status for 1-to-1 conversation.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] `MessageListHeader` subtitle shows information about the channel members according to the design.

PR #2627, opened 2021-11-15
[Compose] - 2626 - Show correct information in MessageListHeader subtitle

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2626

### 🎯 Goal

Show either channel members for group channel or member presence for 1-to-1 conversation.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| photo_2021-11-15 17 01 27 | photo_2021-11-15 17 01 24 |

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelUtils.kt
```diff
@@ -57,3 +59,47 @@ public fun Channel.getReadStatuses(userToIgnore: User?): List<Date> {
...
+ * @param context The context to load string resources.
+ * @param currentUser The currently logged in user.
+ * @return The text that represent the member status of the channel.
+ */
+public fun Channel.getMembersStatusText(context: Context, currentUser: User?): String {
+    val otherMembers = members.filter { it.user.id != currentUser?.id }
+
+    return when {
+        otherMembers.isEmpty() -> ""
+        isOneToOne(currentUser) -> {
```
filbabic, line 91: I think we can remove the {} as it's only ever going to be one statement!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelUtils.kt
```diff
@@ -57,3 +59,47 @@ public fun Channel.getReadStatuses(userToIgnore: User?): List<Date> {
...
+            otherMembers.first()
+                .user
+                .getLastSeenText(context)
+        }
+        else -> {
+            context.resources.getQuantityString(
+                R.plurals.stream_compose_channel_members,
+                otherMembers.count(),
```
filbabic, line 99: Why do we have to use `count()` twice here? Also, could we pull it out in a variable if we really need to fetch the number twice?

filbabic (changes requested):
Two minor comments!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelUtils.kt
bychkovdmitry: Doesn't agree here as the readability suffers. Anyway, committed the change.

zsmb13: +1 on readability, I'd also have kept the branch, but it's fine either way, let's not get lost in the detail 😃

filbabic: We can totally keep the braces if you feel it's more readable, I felt differently from my PoV, but it's 2 against one 😄

filbabic (approved):

Issue #2351, opened 2021-09-15
Expose DefaultMessageContainer child components as public


filbabic:
### Context

If the user wants to build a custom **MessageItem**, they need to use the **ThreadParticipants** and **MessageReactions** components. But these are currently _private_ or _internal_.

### Change Proposal

Make all the smaller components of the **DefaultMessageContainer**, such as **ThreadParticipants**, **MessageReactions**, **MessageFooter** and others public if applicable.

Some components might still be private/internal because they are very simple, or the user needs to define what happens in a specific case when it's shown by default, like the **DeletedMessageContent**.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Research which components can be used by the user for custom UI.
- [x] Expose **ThreadParticipants**.
- [x] Expose **MessageReactions**.
- [ ] Expose various Footer components.
- [ ] Document these components as **Utility components** in docusaurus

filbabic:
Merge given that this is covered in #2742

PR #2633, opened 2021-11-16
[Compose] - 2351 - Expose the MessageReactions component

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2351

### 🎯 Goal

Expose the `MessageReactions` component.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -499,13 +501,44 @@ public fun DefaultMessageItemContent(
...
+public fun MessageReactions(
+    message: Message,
+    modifier: Modifier = Modifier,
+) {
+    val ownReactions = message.ownReactions
+    val supportedReactions = ChatTheme.reactionTypes
+
+    val reactions = message.reactionCounts
+        .map { it.key }
+        .filter { supportedReactions[it] != null }
```
filbabic, line 519: Maybe we can filter for the reactions that are supported within the app and then have the `isNotEmpty()` check, after which we map them to the pair?

Also - I think reactions should be a `Pair<String, Painter>`, as that way users can load custom emoji reactions (e.g. like in Slack), as that way they can provide us with any type of Painter to load the image.

Finally - I don't really like that we have 2 components with the name `MessageReactions` - can we put this logic into one component as to me this is a bit confusing.

We should also make this customizable by exposing a lambda that represents the content of each reaction item (the icon in our case) so that our users can override what they look like (and potentially apply their own logic). 

If we can, we could also remove the default hardcoded modifiers as our users might want to have the `background` as transparent.

Can you please apply these changes?

filbabic (changes requested):
Left a comment - I think we can improve the customizability of the reactions component a lot by providing a few more parameters and or changing the API by a little.

This would let our users customize them more and reuse reactions in their own UI.

filbabic (approved):
Looks really good! Now people can use our component and customize it, or just build something of their own if they want!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -360,21 +363,31 @@ public fun DefaultMessageItemHeaderContent(
...
-                reactions = reactions
-            )
-        }
+        val ownReactions = message.ownReactions
+        val reactionCounts = message.reactionCounts
+        val supportedReactions = ChatTheme.reactionTypes
+
+        if (reactionCounts.isEmpty()) return
+
+        message.reactionCounts
```
skydoves, line 374: Just a small suggestion, how about this way?

```kotlin
val reactionCounts = message.reactionCounts.ifEmpty { return }
reactionCounts
      .filter { supportedReactions.containsKey(it.key) }
```

skydoves (approved):
Looks good to me!
I left one suggestion. :)

PR #2634, opened 2021-11-16
Fix/query channels spec fix

ogkuzmin:
### 🎯 Goal

We need to specify every query channels request with not only filter object but also with query sort. Previously we also calculated the hash of Filter not correctly (`Filters.in("members", listOf("userId1"))` and `Filters.nin("members", listOf("userId1"))` had the same hash)

### 🛠 Implementation details

1. Fix hash and equal of FilterObject
2. Add to `QueryChannelsSpec` query sort and remove id (keep id only on database level)
3. Test fixes and cleanups

### 🧪 Testing

You can checkout the branch `bug/trellSampleShuffling` and paginate one list and then another, there should not be shufflings

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/models/FilterObjectTests.kt
```diff
@@ -0,0 +1,25 @@
...
+        filterObject1 `should not be equal to` filterObject2
+    }
+
+    @Test
+    fun `Two different filter objects Should be equal`() {
+        val filterObject1 = Filters.`in`("members", listOf("userId1", "userId2"))
+        val filterObject2 = Filters.`in`("members", listOf("userId1", "userId2"))
+
+        filterObject1 `should be equal to` filterObject2
+    }
```
adasiewiczr, line 24: Nice! Maybe it would be worth it to add a test case for sample from description? 
```Filters.in("members", listOf("userId1")) and Filters.nin("members", listOf("userId1")) ```

File: stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/QueryChannelsSpec.kt [hunk 2]
```diff
@@ -1,11 +1,12 @@
...
+import io.getstream.chat.android.client.models.Channel
 
 internal data class QueryChannelsSpec(
     val filter: FilterObject,
-    var cids: Set<String> = emptySet()
+    val querySort: QuerySort<Channel>
 ) {
-    val id: String
-        get() = filter.hashCode().toString()
+    var cids: Set<String> = emptySet()
```
adasiewiczr, line 11: Why are we excluding it from the constructor?

ogkuzmin: Kotlin compare data classes by fields from constructor, fields from the body of class aren't considered when equals and hashCode are running. That was my motivation move mutable cids from constructor to body

adasiewiczr (approved):

PR #2644, opened 2021-11-17
[CAS-1364] Improvement - Resizable gifs

leandroBorgesFerreira:
[CAS-1364](https://stream-io.atlassian.net/browse/CAS-1364)

### 🎯 Goal

Making Giphys autosizable instead of having a pre defined size. 

### 🛠 Implementation details

Width and Height is take from API and both `messageContainer` and `imageView` are set accordingly. If giphy is too wide, then the giphy is resized accordingly. 

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_20211117-192429 | Need to update  |

### 🧪 Testing

Send many gifs to the chat and check the auto size. 

### ☑️ Checklist

- [x] Reviewers added.

### 🎉 GIF
giphy


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -21,12 +26,17 @@ import io.getstream.chat.android.ui.common.extensions.internal.streamThemeInflat
...
 import io.getstream.chat.android.ui.message.list.adapter.view.MediaAttachmentViewStyle
+import kotlin.math.min
 
 internal class MediaAttachmentView : ConstraintLayout {
     var attachmentClickListener: AttachmentClickListener? = null
     var attachmentLongClickListener: AttachmentLongClickListener? = null
     var giphyBadgeEnabled: Boolean = true
 
+    private val maxMediaAttachmentWidth: Int by lazy {
+        (Resources.getSystem().displayMetrics.widthPixels * MAX_WIDTH_PERCENTAGE).toInt()
```
zsmb13, line 37: `Resources.getSystem().displayMetrics.widthPixels` is the width of the entire screen, right? Should (could) we take the width of the current `View` instead? For example, what happens if there's something else next to it on one of the sides?

stream-chat-android-ui-common/api/stream-chat-android-ui-common.api
```diff
@@ -180,6 +180,25 @@ public final class com/getstream/sdk/chat/model/AttachmentMetaData {
 	public fun toString ()Ljava/lang/String;
 }
 
+public final class com/getstream/sdk/chat/model/GiphyInfo {
```
zsmb13, line 183: Do we need to expose anything new in this module to clients? I feel like there should be no changes in this `.api` file on this PR, unless I'm missing something.

zsmb13 (commented):
A couple comments from my perspective, I'll let someone else do a full review after this

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
leandroBorgesFerreira: At this moment the view is not drawn yet, so the width and height are 0 =|, but we should use `messageMaxWidthFactorMine` and `messageMaxWidthFactorTheirs`, I forgot to update this to use those. 

About view next to MessageListView... That's a complicated one, because we don't know the size of the view before it is actually drawn. Maybe a configuration to set a max possible width?

stream-chat-android-ui-common/api/stream-chat-android-ui-common.api
leandroBorgesFerreira: No need to make `GiphyInfo`. Let me check for updates in the API

leandroBorgesFerreira: 🤔  I think it makes sense to change the API to remove `streamUiMediaAttachmentGiphyHeight` as now giphys are auto sizeable... or we can keep it so users can choose to have it the same size always.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentsGroupView.kt
```diff
@@ -30,22 +30,30 @@ import io.getstream.chat.android.ui.message.list.adapter.viewholder.decorator.in
...
         context.createStreamThemeWrapper(),
         attrs,
         defStyleAttr,
         defStyleRes
     )
 
+    fun setMessageContainerView(containerView: View) {
+        this.containerView = containerView
+    }
+
```
skydoves, line 56: I just wondering if we could integrate the `setMessageContainerView` into the `setupMediaAttachmentView`. The `setMessageContainerView` must be called before `setupMediaAttachmentView` and it doesn't seem to be used independently.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -139,7 +166,38 @@ internal class MediaAttachmentView : ConstraintLayout {
...
     }
 
+    private fun Attachment.giphyInfo(field: GiphyInfoType): GiphyInfo? {
+        val giphyInfoMap = (extraData["giphy"] as Map<String, Any>?)?.get(field.value) as Map<String, String>?
+
+        return giphyInfoMap?.let { map ->
+            GiphyInfo(
+                url = map["url"] ?: this.thumbUrl ?: "",
+                width = map["width"]?.toInt() ?: GIPHY_INFO_DEAFULT_WIDTH,
+                height = map["height"]?.toInt() ?: GIPHY_INFO_DEAFULT_HEIGHT,
```
skydoves, line 176: Don't we need to use `dpToPx()` for the `GIPHY_INFO_DEAFULT_WIDTH` and `GIPHY_INFO_DEAFULT_HEIGHT`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -139,7 +166,38 @@ internal class MediaAttachmentView : ConstraintLayout {
...
+    }
+
+    internal enum class GiphyInfoType(val value: String) {
+        ORIGINAL("original")
+    }
+
+    internal data class GiphyInfo(
+        val url: String,
+        val width: Int,
+        val height: Int,
```
skydoves, line 190: It would be great to be marked with `@Px` annotation.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/MediaAttachmentViewStyle.kt
```diff
@@ -83,18 +82,13 @@ public data class MediaAttachmentViewStyle(
                     a.getDrawable(R.styleable.MediaAttachmentView_streamUiMediaAttachmentPlaceHolderIcon)
                         ?: context.getDrawableCompat(R.drawable.stream_ui_picture_placeholder)!!
 
-                val giphyHeight =
-                    a.getDimensionPixelSize(R.styleable.MediaAttachmentView_streamUiMediaAttachmentGiphyHeight,
-                        DEFAULT_GIPHY_HEIGHT.dpToPx())
```
skydoves, line 88: Maybe we can also remove the `DEFAULT_GIPHY_HEIGHT` const in `MediaAttachmentViewStyle` class.

skydoves (commented):
I left several comments! 😄

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentsGroupView.kt
leandroBorgesFerreira: Makes sense

leandroBorgesFerreira:
I removed the size limitation of giphy based on the screen size. There's already a limitation for the message container

leandroBorgesFerreira:
Nevermind, if we take it take, it can grow to be much bigger than the screen.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -21,12 +25,20 @@ import io.getstream.chat.android.ui.common.extensions.internal.streamThemeInflat
...
 import io.getstream.chat.android.ui.message.list.adapter.view.MediaAttachmentViewStyle
+import kotlin.math.min
 
 internal class MediaAttachmentView : ConstraintLayout {
     var attachmentClickListener: AttachmentClickListener? = null
     var attachmentLongClickListener: AttachmentLongClickListener? = null
     var giphyBadgeEnabled: Boolean = true
 
+    internal val displayMetrics: DisplayMetrics
+        inline get() = Resources.getSystem().displayMetrics
```
skydoves, line 36: Would it be better to move this into the `common.extensions.internal` package?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentsGroupView.kt
```diff
@@ -30,22 +30,30 @@ import io.getstream.chat.android.ui.message.list.adapter.viewholder.decorator.in
 internal class MediaAttachmentsGroupView : ConstraintLayout {
     var attachmentClickListener: AttachmentClickListener? = null
     var attachmentLongClickListener: AttachmentLongClickListener? = null
-    val maxMediaAttachmentHeight: Int by lazy {
+    private val maxMediaAttachmentHeight: Int by lazy {
         (Resources.getSystem().displayMetrics.heightPixels * MAX_HEIGHT_PERCENTAGE).toInt()
     }
```
skydoves, line 35: One little tweak:  we can reuse the `displayMetrics` property here.

stream-chat-android-ui-common/api/stream-chat-android-ui-common.api
zsmb13: I think it's fine to keep it as an optional attribute. If we do that, let's make sure we document it in the attrs file properly.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
zsmb13: Are you sure the width of the View wouldn't be available by the time we need to calculate this value? In that case, perhaps we should wait until the View is layed out, and only then run any code that lays out the Gif contained within?

zsmb13: It just feels incorrect to base this sizing on the display size, as that's not what we actually want to do - we want to size based on the containing View's size.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -73,14 +82,36 @@ internal class MediaAttachmentView : ConstraintLayout {
         }
 
         if (attachment.type == ModelType.attach_giphy) {
-            binding.imageView.updateLayoutParams {
-                height = style.giphyHeight
+            attachment.giphyInfo(GiphyInfoType.ORIGINAL)?.let { giphyInfo ->
```
ogkuzmin, line 85: Can we extract this part to another method like `setupLayoutForGiphy`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -73,14 +82,36 @@ internal class MediaAttachmentView : ConstraintLayout {
...
+                    }
+
+                    binding.imageView.updateLayoutParams {
+                        height = parseHeight(giphyInfo) - DEFAULT_MARGIN_FOR_CONTAINER_DP.dpToPx()
+                        width = parseWidth(giphyInfo) - DEFAULT_MARGIN_FOR_CONTAINER_DP.dpToPx()
+                    }
+
+                    showImageByUrl(url) {
+                        showMore()
+                        showGiphyLabel()
```
ogkuzmin, line 101: I see three doublings. Can we extract it to something like `showImageByUrlDefault(url: String)`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/MediaAttachmentViewStyle.kt
```diff
@@ -83,18 +80,13 @@ public data class MediaAttachmentViewStyle(
                     a.getDrawable(R.styleable.MediaAttachmentView_streamUiMediaAttachmentPlaceHolderIcon)
                         ?: context.getDrawableCompat(R.drawable.stream_ui_picture_placeholder)!!
 
-                val giphyHeight =
```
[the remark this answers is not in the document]
leandroBorgesFerreira: At the moment I'm just testing the code, let me put this back to draft

Issue #2621, opened 2021-11-15
[Compose] - Implement limitations when sending new messages


filbabic:
### Context

Our API has certain limitations when sending messages, that disable the user from sending their content if the limits are exceeded.

We need to implement these limitations and make sure our users don't break the API.

### Change Proposal

Implement the following limitations:
- 5000 character message length
- 10 message attachments limit
- 5MB (investigate if correct) per-attachment limit.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Research more on our exact limitations and API design
- [ ] Implement our limitations and disable the send message button in case the user breaks any of them
- [ ] Test the new limitations with respective file and message length sizes, as well as different number of attachments
- [ ] Update the documentation to let our users know about this

PR #2651, opened 2021-11-19
[Compose] - 2621 - Add message length limitation

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2621

### 🎯 Goal

Add message length limitation.

### 🛠 Implementation details

This PR only introduces message length limitation. The rest of the limitation will be covered in different PRs.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| device-2021-11-19-182347 | device-2021-11-19-182238 |
| device-2021-11-19-181733 | device-2021-11-19-181921 |

### 🧪 Testing

Modify the message limitation in the dashboard or test with the default value - 5000 chars.

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageInputState.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageInputState.kt
@@ -9,9 +9,11 @@ import io.getstream.chat.android.common.state.MessageAction
  * @param inputValue The current text value that's within the input.
  * @param attachments The currently selected attachments.
  * @param action The currently active [MessageAction].
+ * @param maxMessageLength The maximum allowed message length.
  */
 public data class MessageInputState(
     val inputValue: String = "",
     val attachments: List<Attachment> = emptyList(),
     val action: MessageAction? = null,
+    val maxMessageLength: Int = Integer.MAX_VALUE,
 )
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageInputState.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageInputState.kt
@@ -4,6 +4,11 @@ import io.getstream.chat.android.client.models.Attachment
 import io.getstream.chat.android.common.state.MessageAction
 
 /**
+ * The default allowed number of characters in a message.
+ */
+private const val DEFAULT_MAX_MESSAGE_LENGTH: Int = 5000
+
+/**
  * Represents the state within the message input.
  *
  * @param inputValue The current text value that's within the input.
@@ -15,5 +20,5 @@ public data class MessageInputState(
     val inputValue: String = "",
     val attachments: List<Attachment> = emptyList(),
     val action: MessageAction? = null,
-    val maxMessageLength: Int = Integer.MAX_VALUE,
+    val maxMessageLength: Int = DEFAULT_MAX_MESSAGE_LENGTH,
 )
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageInputState.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageInputState.kt
@@ -2,11 +2,7 @@ package io.getstream.chat.android.compose.state.messages.composer
 
 import io.getstream.chat.android.client.models.Attachment
 import io.getstream.chat.android.common.state.MessageAction
-
-/**
- * The default allowed number of characters in a message.
- */
-private const val DEFAULT_MAX_MESSAGE_LENGTH: Int = 5000
+import io.getstream.chat.android.common.state.ValidationError
 
 /**
  * Represents the state within the message input.
@@ -14,11 +10,11 @@ private const val DEFAULT_MAX_MESSAGE_LENGTH: Int = 5000
  * @param inputValue The current text value that's within the input.
  * @param attachments The currently selected attachments.
  * @param action The currently active [MessageAction].
- * @param maxMessageLength The maximum allowed message length.
+ * @param validationErrors The list of validation errors.
  */
 public data class MessageInputState(
     val inputValue: String = "",
     val attachments: List<Attachment> = emptyList(),
     val action: MessageAction? = null,
-    val maxMessageLength: Int = DEFAULT_MAX_MESSAGE_LENGTH,
+    val validationErrors: List<ValidationError> = emptyList(),
 )
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
@@ -14,7 +14,7 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
-import androidx.compose.ui.Alignment.Companion.CenterVertically
+import androidx.compose.ui.Alignment.Companion.Bottom
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
@@ -66,6 +66,7 @@ public fun MessageComposer(
         MessageInput(
             modifier = Modifier
                 .fillMaxWidth()
+                .padding(vertical = 8.dp)
                 .weight(1f),
             label = label,
             messageInputState = inputState,
@@ -77,6 +78,7 @@ public fun MessageComposer(
     val value by viewModel.input.collectAsState()
     val selectedAttachments by viewModel.selectedAttachments.collectAsState()
     val activeAction by viewModel.lastActiveAction.collectAsState(null)
+    val maxMessageLength by viewModel.maxMessageLength.collectAsState()
 
     MessageComposer(
         modifier = modifier,
@@ -87,7 +89,12 @@ public fun MessageComposer(
         },
         integrations = integrations,
         input = input,
-        messageInputState = MessageInputState(value, selectedAttachments, activeAction),
+        messageInputState = MessageInputState(
+            inputValue = value,
+            attachments = selectedAttachments,
+            action = activeAction,
+            maxMessageLength = maxMessageLength
+        ),
         shouldShowIntegrations = true,
         onCancelAction = onCancelAction
     )
@@ -115,7 +122,7 @@ public fun MessageComposer(
     integrations: @Composable RowScope.() -> Unit,
     input: @Composable RowScope.(MessageInputState) -> Unit,
 ) {
-    val (value, attachments, activeAction) = messageInputState
+    val (value, attachments, activeAction, maxMessageLength) = messageInputState
 
     Surface(
         modifier = modifier,
@@ -137,24 +144,21 @@ public fun MessageComposer(
             }
 
             Row(
-                Modifier
-                    .fillMaxWidth(),
-                verticalAlignment = CenterVertically
+                Modifier.fillMaxWidth(),
+                verticalAlignment = Bottom
             ) {
 
                 if (shouldShowIntegrations && activeAction !is Edit) {
                     integrations()
                 } else {
                     Spacer(
-                        modifier = Modifier
-                            .size(48.dp)
-                            .align(CenterVertically)
+                        modifier = Modifier.size(48.dp)
                     )
                 }
 
                 input(messageInputState)
 
-                val isInputValid = value.isNotEmpty() || attachments.isNotEmpty()
+                val isInputValid = (value.isNotEmpty() || attachments.isNotEmpty()) && value.length < maxMessageLength
 
                 IconButton(
                     enabled = isInputValid,
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
@@ -152,7 +152,7 @@ public fun MessageComposer(
                     integrations()
                 } else {
                     Spacer(
-                        modifier = Modifier.size(48.dp)
+                        modifier = Modifier.size(16.dp)
                     )
                 }
 
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
@@ -78,7 +78,7 @@ public fun MessageComposer(
     val value by viewModel.input.collectAsState()
     val selectedAttachments by viewModel.selectedAttachments.collectAsState()
     val activeAction by viewModel.lastActiveAction.collectAsState(null)
-    val maxMessageLength by viewModel.maxMessageLength.collectAsState()
+    val validationErrors by viewModel.validationErrors.collectAsState()
 
     MessageComposer(
         modifier = modifier,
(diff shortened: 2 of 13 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -77,6 +78,7 @@ public fun MessageComposer(
     val value by viewModel.input.collectAsState()
     val selectedAttachments by viewModel.selectedAttachments.collectAsState()
     val activeAction by viewModel.lastActiveAction.collectAsState(null)
+    val maxMessageLength by viewModel.maxMessageLength.collectAsState()
```
bychkovdmitry, line 81: Should we expose `MessageInputState` in the ViewModel?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageComposerViewModel.kt
```diff
@@ -122,4 +127,12 @@ public class MessageComposerViewModel(
...
      */
     public fun leaveThread(): Unit = messageComposerController.leaveThread()
+
+    /**
+     * Disposes the inner [MessageComposerController].
+     */
+    override fun onCleared() {
+        super.onCleared()
+        messageComposerController.onCleared()
+    }
```
bychkovdmitry, line 137: How do you feel about disposing the inner controller here?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageInputState.kt
```diff
@@ -9,9 +9,11 @@ import io.getstream.chat.android.common.state.MessageAction
  * @param inputValue The current text value that's within the input.
  * @param attachments The currently selected attachments.
  * @param action The currently active [MessageAction].
+ * @param maxMessageLength The maximum allowed message length.
  */
 public data class MessageInputState(
     val inputValue: String = "",
     val attachments: List<Attachment> = emptyList(),
     val action: MessageAction? = null,
+    val maxMessageLength: Int = Integer.MAX_VALUE,
```
filbabic, line 18: Does our BE support this kind of a limit? Maybe it's better to set it to a default (5000) just to make sure the message API will work when the users try to send a message.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
filbabic: I was thinking about it, but there's no pretty way to do it - we'd have to combine multiple flows and use `combine/combineLatest` on them to form a composite flow that creates and updates the `MessageInputState`.

If you feel like this is something that's worthwhile, we can do it in this PR and then update our samples & docs.

WDYT?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -137,24 +144,21 @@ public fun MessageComposer(
             }
 
             Row(
-                Modifier
-                    .fillMaxWidth(),
-                verticalAlignment = CenterVertically
+                Modifier.fillMaxWidth(),
+                verticalAlignment = Bottom
```
filbabic, line 148: Is there a reason for this alignment chagne?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -137,24 +144,21 @@ public fun MessageComposer(
...
             ) {
 
                 if (shouldShowIntegrations && activeAction !is Edit) {
                     integrations()
                 } else {
                     Spacer(
-                        modifier = Modifier
```
filbabic, line 149: Can we remove this block actually - Marton and I were discussing this and we feel like it's better to have the input expand to the edge of the screen (also per the design I think). 

Maybe we can remove this or resize the spacer to have it symmetrical with the right side

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/components/ComposerIntegrations.kt
```diff
@@ -20,13 +18,12 @@ import io.getstream.chat.android.compose.ui.theme.ChatTheme
  * @param modifier Modifier for styling.
  */
 @Composable
-internal fun RowScope.DefaultComposerIntegrations(
+internal fun DefaultComposerIntegrations(
```
filbabic, line 21: Why isn't this a part of the RowScope anymore? This way the user can no longer affect if the integrations property needs to have a `weight()` for example. Or if it needs to be centered/bottom/top oriented.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/components/MessageInput.kt
```diff
@@ -109,6 +109,7 @@ public fun MessageInput(
     InputField(
         modifier = modifier,
         value = value,
+        maxLines = 6,
```
filbabic, line 112: I feel like this should be customizable and a constant somewhere - like a parameter of the MessageInput, that is `DEFAULT_MESSAGE_INPUT_LINE_LIMIT`, that we define within our SDK

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
```diff
@@ -43,6 +54,27 @@ public class MessageComposerController(
      */
     public val selectedAttachments: MutableStateFlow<List<Attachment>> = MutableStateFlow(emptyList())
 
+    /**
+     * Represents the maximum allowed message length.
+     */
+    public val maxMessageLength: MutableStateFlow<Int> = MutableStateFlow(Integer.MAX_VALUE)
```
filbabic, line 60: Same here - maybe have this as a default limit (as we mentioned before)

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
```diff
@@ -43,6 +54,27 @@ public class MessageComposerController(
...
+     * Sets up the data loading operations such as observing the maximum allowed message length.
+     */
+    init {
+        scope.launch {
+            val result = chatDomain.watchChannel(channelId, 0).await()
+
+            if (result.isSuccess) {
+                val channelController = result.data()
+                channelController.channelConfig.collect {
+                    maxMessageLength.value = it.maxMessageLength
```
filbabic, line 72: Will this be updated in the runtime, based on the dashboard config? If not, we can just collect it once, update the value and then stop collecting.

That might help us as we won't have to collect the values and cancel/clean up the scoep.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageComposerViewModel.kt
filbabic: That works - but do check out my comments below, so that we can see if it's needed at all!

filbabic (changes requested):
Left some comments for us to take a look at and potentially clean up! :]

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/components/ComposerIntegrations.kt
bychkovdmitry: This is an internal implementation which no longer requires the `RowScope`. The weight can be set externally here:

```
    integrations: @Composable RowScope.() -> Unit = {
    },
```

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
bychkovdmitry: Unfortunately, I anticipate some other use cases like observing channel members for suggestions, where we will need to observe the changes. However, you are right here. Support for updating this value in real time not needed.

bychkovdmitry: Also, the state in the controller is not updated instantly:
`internal val channelConfig: MutableStateFlow<Config> = MutableStateFlow(Config())`
Thinking about a hack which implies subscribing to something else that 100% happens after the network request and then collecting the channel config value.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -137,24 +144,21 @@ public fun MessageComposer(
...
-                            .size(48.dp)
-                            .align(CenterVertically)
+                        modifier = Modifier.size(16.dp)
                     )
                 }
 
                 input(messageInputState)
```
bychkovdmitry, line 160: Just disabling the "send" button for now. Will show a toast or a snackbar in the next PR as like with suggestion popup it is hard to display it above the composer.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
```diff
@@ -43,6 +60,32 @@ public class MessageComposerController(
...
+     */
+    init {
+        scope.launch {
+            val result = chatDomain.watchChannel(channelId, 0).await()
+
+            if (result.isSuccess) {
+                val channelController = result.data()
+                channelController.channelConfig.collect {
+                    maxMessageLength = it.maxMessageLength
+                }
```
bychkovdmitry, line 84: This is a black box, it can emit couple of times the default value, then a value from the server. Need to discuss this.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
filbabic: Yeah sounds good!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
filbabic: Thank you so much! :]

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/components/ComposerIntegrations.kt
filbabic: Alright, yeah good point, thanks for making sure!

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
```diff
@@ -243,4 +292,56 @@ public class MessageComposerController(
...
+                add(
+                    ValidationError.AttachmentCountExceeded(
+                        attachmentCount = attachmentCount,
+                        maxAttachmentCount = maxAttachmentCount
+                    )
+                )
+            }
+
+            val attachments: List<Attachment> = selectedAttachments.value
+                .filter { it.fileSize > maxAttachmentSize }
```
filbabic, line 329: One thing that we need to note here - I think the max attachment size od 5MB is a BE limitation. So we should make the check that the file size also isn't larger than that.

Something like:

`it.fileSize > maxAttachmentSize || it.fileSize > MAX_DEFAULT_FILE_SIZE`. It kind of seems ugly but it's the way things should work

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
```diff
@@ -43,6 +60,32 @@ public class MessageComposerController(
...
 
+    /**
+     * Represents the list of validation errors for the current text input and the currently selected attachments.
+     */
+    public val validationErrors: MutableStateFlow<List<ValidationError>> = MutableStateFlow(emptyList())
+
+    /**
+     * Represents the maximum allowed message length in the message input.
+     */
+    private var maxMessageLength: Int = DEFAULT_MAX_MESSAGE_LENGTH
```
filbabic, line 71: Any reason why this is not in the constructor? Cannot access the object constant?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
filbabic: Might be the same for number of attachments - as I'm not sure that messages can have more than 10 attachments

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
filbabic: Is it possible to show some kind of errors within either the attachments (as we discussed) or the composer? Dedicate a small slot?

We can create this as an extra ticket, for UI/UX improvement

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
filbabic: I think we can observe it and all, but we should definitely follow up with BE to see what the issue here is for inconsistency

filbabic (commented):
Left a few comments!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
bychkovdmitry: Yes, I will create a separate ticket for validation errors. Needs to be checked with the design team.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
bychkovdmitry: Because this value is coming from the backed. Do we need a way to override the backend value?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
bychkovdmitry: https://github.com/GetStream/stream-chat-android/issues/2653

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
bychkovdmitry: So, the following are the backend limitations:
- 20mb for Stream CDN, any file size for a custom CDN.
- 10 attachments per message. This looks like a value hardcoded on the backend, I can limit enforce the upper limit to 10 items. Don't want to crash if a client specifies something like 20 attachments per message.
- 5000 chars per message, but can be set to a bigger value via dashboard.

filbabic: This all sounds good and if we can observe things here using channel config, that's great. If not, we can put these default limits and add support for dashboard customization later on!

filbabic (approved):

PR #2652, opened 2021-11-19
[CAS-1421] Improvement - Avatar Customization

leandroBorgesFerreira:
[CAS-1421](https://stream-io.atlassian.net/browse/CAS-1421)

### 🎯 Goal

Improve customization of avatar. Add more possibilities for customization. Now it is possible: 

- Change shape of avatar. Rounded square was added. 

### 🛠 Implementation details

New options added in AvatarStyle

### 🎨 UI Changes

| Before - images from BE | After - avatar tye = color |
| --- | --- |
| Screenshot_20211119-165851 | Screenshot_20211122-132154 |

### 🧪 Testing

Customize avatar in many different ways. 

### ☑️ Checklist

- [x] Comparison screenshots added for visual changes. todo
- [x] Reviewers added

### 🎉 GIF
giphy

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
@@ -116,15 +116,15 @@ public class AvatarView : AppCompatImageView {
             AvatarType.PICTURE -> {
                 if (drawable != null) {
                     super.onDraw(canvas)
-                    drawBorder(canvas)
-                    drawOnlineStatus(canvas)
+                    drawBorder(canvas, avatarStyle, borderPaint)
+                    drawOnlineStatus(canvas, isOnline, avatarStyle)
                 }
             }
             AvatarType.COLOR -> {
-                drawColor(canvas, avatarStyle)
+                drawColor(canvas, avatarStyle, backgroundPaint)
                 drawInitials(canvas, avatarInitials)
-                drawBorder(canvas)
-                drawOnlineStatus(canvas)
+                drawBorder(canvas, avatarStyle, borderPaint)
+                drawOnlineStatus(canvas, isOnline, avatarStyle)
             }
         }
     }
@@ -144,7 +144,7 @@ public class AvatarView : AppCompatImageView {
         backgroundPaint.color = avatarStyle.avatarColor
     }
 
-    private fun drawOnlineStatus(canvas: Canvas) {
+    private fun drawOnlineStatus(canvas: Canvas, isOnline: Boolean, avatarStyle: AvatarStyle) {
         if (isOnline && avatarStyle.onlineIndicatorEnabled) {
             val cx: Float = when (avatarStyle.onlineIndicatorPosition) {
                 OnlineIndicatorPosition.TOP_LEFT,
@@ -184,7 +184,7 @@ public class AvatarView : AppCompatImageView {
         )
     }
 
-    private fun drawColor(canvas: Canvas, avatarStyle: AvatarStyle) {
+    private fun drawColor(canvas: Canvas, avatarStyle: AvatarStyle, backgroundPaint: Paint) {
         if (avatarStyle.avatarShape == AvatarShape.SQUARE) {
             canvas.drawRoundRect(
                 0F,
@@ -205,7 +205,7 @@ public class AvatarView : AppCompatImageView {
         }
     }
 
-    private fun drawBorder(canvas: Canvas) {
+    private fun drawBorder(canvas: Canvas, avatarStyle: AvatarStyle, borderPaint: Paint) {
         if (avatarStyle.avatarBorderWidth != 0) {
             if (avatarStyle.avatarShape == AvatarShape.SQUARE) {
                 canvas.drawRoundRect(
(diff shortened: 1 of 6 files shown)


stream-chat-android-ui-components/src/main/res/values/styles.xml
```diff
@@ -229,6 +229,9 @@
         <item name="streamUiThreadAvatarStyle">@style/StreamUi.Thread.Avatar</item>
         <item name="streamUiReplyAvatarStyle">@style/StreamUi.Reply.Avatar</item>
         <item name="streamUiMessageOptionsAvatarStyle">@style/StreamUi.MessageOptions.Avatar</item>
+        <item name="streamUiAvatarInitialsTextColor">@color/stream_ui_white</item>
+        <item name="streamUiAvatarInitialsTextSize">20sp</item>
+        <item name="streamUiAvatarColor">@color/stream_ui_black</item>
```
skydoves, line 234: I was wondering if it would be better to provide the same default values when we initialize the custom attributes on `AvatarStyles.` For example, we're initializing with `stream_ui_accent_blue` color for the `AvatarColor` on `AvatarStyle`. And same for the `streamUiGlobalAvatarBorderColor`, `streamUiAvatarInitialsTextSize`, and `streamUiAvatarBorderWidth`.

skydoves (commented):
Looks good overall! I left a few comments. 😃

stream-chat-android-ui-components/src/main/res/values/styles.xml
leandroBorgesFerreira: I aligned the initializations, but `streamUiAvatarBorderWidth` has many different values. I think we should leave as is so we avoid uncessary changes in the API.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
```diff
@@ -69,10 +112,18 @@ public class AvatarView : AppCompatImageView {
...
+                if (drawable != null) {
+                    super.onDraw(canvas)
+                    drawBorder(canvas)
+                    drawOnlineStatus(canvas)
+                }
+            }
+            AvatarType.COLOR -> {
+                drawColor(canvas, avatarStyle)
+                drawInitials(canvas, avatarInitials)
+            }
```
skydoves, line 126: I am just wondering if we don't need to draw the border and online status for this type.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
```diff
@@ -114,30 +166,89 @@ public class AvatarView : AppCompatImageView {
         }
     }
 
-    private fun drawBorder(canvas: Canvas) {
-        if (avatarStyle.avatarBorderWidth != 0) {
+    private fun avatarShape(style: AvatarStyle): StreamImageLoader.ImageTransformation {
```
skydoves, line 169: Can we use the global `avatarStyle` property instead of passing the parameter for this and `drawColor` functions?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
leandroBorgesFerreira: ops, we do. My mistake

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarView.kt
leandroBorgesFerreira: We can, but then we lose [referential transparency.](https://www.sitepoint.com/what-is-referential-transparency/) . Right not, this functional is a method that could be used in any place that can pass `AvatarStyle` as an argument, but we if use the class property, this method gets coupled to this class. 

Using parameters instead makes the code easier to reason about (in my option) because the state that the method has to deal with is only the state of the parameters passed to it instead of the state of the whole class

skydoves: That's a great point! Then should we refactor the `drawOnlineStatus` to receive the `avatarStyle ` for consistency?

skydoves (approved):
Looks good to me! 👍

leandroBorgesFerreira:
I just took broader look at our code and looks that maybe a good part of this PR is not necessary as it can probably be implemented using AvatarBitmapFactory. Testing to see if it possible or not.

skydoves:
@leandroBorgesFerreira Hmm.. I think so. It would be better to implement initials by using `AvatarBitmapFactory`.
It seems segmentation of images with initials for a multi-user channel is impossible in this RP.

leandroBorgesFerreira:
Keeping avatar share as is adds new behaviour, but removing avatar type as we can always bitmap and to way to customize is to override`AvatarBitmapFactory`

leandroBorgesFerreira:
@skydoves Now AvatarView only accepts shape change as it is something not possible to change with `AvatarBitmapFactory` (the factory only cretes the Bitmap, but it doesn't change its shape, this is done by Coil inside AvatarView)

samiuelson (approved):

skydoves (approved):

Issue #2656, opened 2021-11-22
[Compose] - Don't rely on the channel members to show message list header subtitle


bychkovdmitry:
### Context

We rely on the `Channel.members` field to show the number of members in a channel. Also, we use it to conut and display the number online users. But the `Channel.members` is unreliable as it only contains the first 30 members of a channel. 

### Change Proposal

Use `Channel.member_count` and `Channel.watcher_count` fields to display the `MessageListHeader` subtitle.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] The `member_count` field is used to display the number channel members in the `MessageListHeader` 
- [ ] The `watcher_count` field is user to display the number of "online" members in the `MessageListHeader`

PR #2657, opened 2021-11-22
[Compose] - 2656 - Do not rely on channel members to display message list header subtitle

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2656

### 🎯 Goal

We rely on the `Channel.members` field to show the number of members in a channel. Also, we use it to count and display the number online users. But the `Channel.members` is unreliable as it only contains the first 30 members of a channel, that's why we need to find another way.

### 🛠 Implementation details

We need to use `Channel.member_count` and `Channel.watcher_count` fields to display the `MessageListHeader` subtitle.

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
@@ -51,6 +51,7 @@ import io.getstream.chat.android.compose.ui.theme.ChatTheme
 import io.getstream.chat.android.compose.ui.util.getMembersStatusText
 import io.getstream.chat.android.compose.ui.util.isDistinct
 import io.getstream.chat.android.compose.ui.util.isMuted
+import io.getstream.chat.android.compose.ui.util.isOneToOne
 
 /**
  * Shows special UI when an item is selected.
@@ -78,8 +79,11 @@ public fun ChannelInfo(
         ?.role
         ?.let { it == "admin" || it == "owner" }
         ?: false
-
-    val otherMembers = channelMembers.filter { it.user.id != currentUser?.id }
+    val membersToDisplay = if (selectedChannel.isOneToOne(currentUser)) {
+        channelMembers.filter { it.user.id != currentUser?.id }
+    } else {
+        channelMembers
+    }
 
     val channelOptions = listOfNotNull(
         ChannelOption(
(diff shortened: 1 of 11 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
```diff
@@ -78,8 +79,11 @@ public fun ChannelInfo(
         ?.role
         ?.let { it == "admin" || it == "owner" }
         ?: false
-
-    val otherMembers = channelMembers.filter { it.user.id != currentUser?.id }
+    val membersToDisplay = if (selectedChannel.isOneToOne(currentUser)) {
+        channelMembers.filter { it.user.id != currentUser?.id }
+    } else {
+        channelMembers
+    }
```
bychkovdmitry, line 86: Displaying the current user in the list of channel members like other SDKs. Also, makes sense as we, for example, have a caption that the channel contains 3 members, and we actually have 3 use avatars below (including the current user).

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelUtils.kt
```diff
@@ -84,21 +84,26 @@ public fun Channel.isOneToOne(currentUser: User?): Boolean {
...
-                count,
-                count,
-                otherMembers.count { it.user.online }
+            val memberCountString = context.resources.getQuantityString(
+                R.plurals.stream_compose_member_count,
+                memberCount,
+                memberCount
             )
+
+            return if (watcherCount > 0) {
```
bychkovdmitry, line 98: Discussed with the backend team, and they claim that we can rely on the `watcherCount` field, which actually represents online users (not only the users who are watching the channel, but also connected to the WS). We might have some inconsistencies for some time as we had permissions disabled in the dashboard.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfo.kt
filbabic: Makes total sense, nicely done!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelUtils.kt
filbabic: I've seen the discussion - I think if they stand behind it, then that's completely okay. Especially if we can show the correct number of people that are in the channel and not just the ones we paginated for

filbabic (approved):
Looks good!

PR #2659, opened 2021-11-22
[Compose] - 2658 - Make interfaces functional

bychkovdmitry:
https://github.com/orgs/GetStream/projects/6/views/4

### 🎯 Goal

Make interfaces functional where possible

### 🛠 Implementation details

- Made ChannelNameFormatter interface is functional
- Made ClipboardHandler interface is functional
- Made MessagePreviewFormatter interface is functional

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MessagePreviewFormatter.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MessagePreviewFormatter.kt
@@ -1,15 +1,15 @@
 package io.getstream.chat.android.compose.ui.util
 
-import android.annotation.SuppressLint
 import android.content.Context
-import androidx.compose.runtime.Composable
 import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.buildAnnotatedString
 import io.getstream.chat.android.client.models.Attachment
 import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.models.User
-import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.compose.ui.attachments.AttachmentFactory
+import io.getstream.chat.android.compose.ui.theme.StreamTypography
 
 /**
  *  An interface that allows to generate a preview text for the given message.
@@ -23,7 +23,6 @@ public fun interface MessagePreviewFormatter {
      * @param currentUser The currently logged in user.
      * @return The formatted text representation for the given message.
      */
-    @Composable
     public fun formatMessagePreview(message: Message, currentUser: User?): AnnotatedString
 
     public companion object {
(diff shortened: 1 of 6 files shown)


skydoves (approved):
👍

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MessagePreviewFormatter.kt
```diff
@@ -97,6 +99,7 @@ private class DefaultMessagePreviewFormatter(
     /**
      * Appends the message text to the [AnnotatedString].
      */
+    @SuppressLint("ComposableNaming")
     @Composable
```
filbabic, line 103: I think we should try and remove the Composable annotation from these functions as for the most part, these do not rely on Compose at all.

The only Compose requirement here are the fontStyles and the AttachmentFactories, which we can pass to the messageFormatter as parameters.

Here's an example of what it would look like for `appendAttachmentText()`:

```
    /**
     * Appends a string representations of [attachments] to the [AnnotatedString].
     */
    private fun AnnotatedString.Builder.appendAttachmentText(
        attachments: List<Attachment>,
        attachmentFactories: List<AttachmentFactory>,
        attachmentTextFontStyle: FontStyle,
    ) {
        if (attachments.isNotEmpty()) {
            attachmentFactories
                .firstOrNull { it.canHandle(attachments) }
                ?.textFormatter
                ?.let { textFormatter ->
                    attachments.mapNotNull { attachment ->
                        textFormatter.invoke(attachment)
                            .let { previewText ->
                                previewText.ifEmpty { null }
                            }
                    }.joinToString()
                }?.let { attachmentText ->
                    val startIndex = this.length
                    append(attachmentText)

                    addStyle(
                        SpanStyle(
                            fontStyle = attachmentTextFontStyle
                        ),
                        start = startIndex,
                        end = startIndex + attachmentText.length
                    )
                }
        }
    }
```

Notice how with just a couple of parameters we managed to make this a regular function. The base function signature would be the following:

```
    public fun formatMessagePreview(
        message: Message,
        currentUser: User?,
        textFontStyle: FontStyle,
        senderNameFontStyle: FontStyle,
        attachmentTextFontStyle: FontStyle,
        attachmentFactories: List<AttachmentFactory>,
    ): AnnotatedString
```
Which makes sense, given that this is an implementation detail of our SDK and if our users want to override it or display something completely different, they probably won't use our `ChannelDetails`.

When we add the Composable annotation, Compose adds a new entry to its tree that represents the value of that Composable - in this case a UI-less, value-less function, that appends some text to a builder.

Since these Composables produce absolutely no observable result -we don't return a value that's powered by a `remember` block, we don't produce UI with this, we don't rely on anything that's likely to change in the runtime (factories and typography don't change often), we're adding unnecessary entries to the tree that might cause performance overhead.

I'm not 100% sure of the impact of this, but it's also easier to have the SAM/functional interface approach, if the functions are not Composables.

LMK what you think - I can make the proposed changes if you want, as I have them stashed locally.

filbabic (changes requested):
Left one comment for us to discuss as I think we can improve our MessageFormatter!

bychkovdmitry:
Great job removing unnecessary composable! Approved (I can't approve my own PR)

filbabic (approved):
Approved from Dmitrii!

PR #2649, opened 2021-11-19
Refac/cas 1267 thread controller in offline plugin

ogkuzmin:
### 🎯 Goal

Refactor `ThreadController` and move logic to `OfflinePlugin`

### 🛠 Implementation details

1. Move data to `ThreadMutableState`
2. Move logic to `ThreadLogic`
3. Wrap side effects in listeners and invoke them when `ChatClient::getReplies` is invoked
4. Fix tests

### ☑️ Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] PR targets the `develop` branch
- [ ] Changelog is updated with client-facing changes
- [ ] New code is covered by unit tests
- [ ] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (docusaurus, tutorial, CMS (task created))
- [x] Reviewers added

PR #2660, opened 2021-11-22
Refac/thread offline plugin in vm

ogkuzmin:
This PR should be merged after [#2649]

### 🎯 Goal
Use thread state from `OfflinePlugin` in VMs

### 🛠 Implementation details

1. Use ThreadState in compose
2. Use ThreadState in UI components

### 🧪 Testing

Run debug version and try to enable/disable and see that thread are shown correctly

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/channel/thread/RepliesQueryReference.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/channel/thread/RepliesQueryReference.kt
@@ -28,7 +28,8 @@ public class RepliesQueryReference(
 
     /**
      * Returns [ThreadState] for the thread replies of a message with id equal to [messageId]. And fill it by data from
-     * [ChatClient.getReplies].
+     * [ChatClient.getReplies]. We use a coroutine instead of a simple enqueue() call, to have control over the
+     * lifecycle of the request.
      *
      * @param scope Coroutine scope where initial data filling action is being invoked.
      */
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/channel/thread/RepliesQueryReference.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/channel/thread/RepliesQueryReference.kt
@@ -1,0 +1,39 @@
+package io.getstream.chat.android.offline.experimental.channel.thread
+
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.client.call.Call
+import io.getstream.chat.android.client.call.await
+import io.getstream.chat.android.client.models.Message
+import io.getstream.chat.android.core.ExperimentalStreamChatApi
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+import io.getstream.chat.android.offline.experimental.channel.thread.state.ThreadState
+import io.getstream.chat.android.offline.experimental.extensions.state
+import io.getstream.chat.android.offline.experimental.plugin.QueryReference
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+
+@ExperimentalStreamChatApi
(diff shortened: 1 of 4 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/channel/thread/RepliesQueryReference.kt
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * Returns [ThreadState] for the thread replies of a message with id equal to [messageId]. And fill it by data from
+     * [ChatClient.getReplies].
+     *
+     * @param scope Coroutine scope where initial data filling action is being invoked.
+     */
+    override fun asState(scope: CoroutineScope): ThreadState {
+        scope.launch { get().await() }
```
filbabic, line 36: What happens with the list of messages here? I imagine they get updated somewhere internally, but to me it's super weird seeing a coroutine launched to get a list of data and then not doing anything with it.

filbabic (commented):
Just left one comment - kind of confusing to fetch items and not do anything with them in the use case.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/channel/thread/RepliesQueryReference.kt
ogkuzmin: The logic of our offline support is so that we expose some state objects with reactive data and under the hood, we run a request to cache and BE to obtain data and fill those reactive sources with it. So `ThreadState` that is returned here provides you a possibility to bind your UI to reactive messages source.

filbabic: I understand it's updating data internall and that the ThreadState gives you the option of using the state.

However I don't understand why we launch a coroutine and await it, if the result value is never used (it just updates internal value).

Can we do a simple `enqueue()` here?

ogkuzmin: Yes, we can `enqueue`. But passing scope you can have control flow of request lifecycle. I was thinking it's as use the full feature. If you think it's not needed we can remove it

filbabic: I think we can keep it - but just document it a bit more if possible.

E.g.

"We fetch the data in a coroutine and await its result because we want to update the internal data based on the result. We use a coroutine instead of a simple `enqueue()` call, to have control over the lifecycle of the request."

I think that should be good enough so that people know what's going on!

ogkuzmin: But it says pretty the same, doesn't it?

`@param scope Coroutine scope where initial data filling action is being invoked.`

We can change `action` to `coroutine` for example. What do you think?

filbabic: Hmmm yeah it's mostly the same - I think `action` can stay.

I think maybe just adding: "We use a coroutine instead of a simple enqueue() call, to have control over the lifecycle of the request." would be helpful then so that we know why we chose one over the other.

If you feel like this doesn't make sense, we can just resolve this convo and move on with the merging process! :]

ogkuzmin: For me your feedback is valuable. And I'm glad to find such description that make code reading easy

filbabic (approved):

adasiewiczr (approved):

Issue #2669, opened 2021-11-24
Incapsulate the data required to render channel list items in ChannelItem class


bychkovdmitry:
### Context

We use a list of `Channel` objects to render channels on the channel list screen. The `Channel` model limits the data that we can display. We are already passing the `isMuted` flag in `Channel`'s extra data. We might also want to pass typing users.

### Change Proposal

Introduce the `ChannelItem` class and use it in the `ChannelsState` class:

```
public data class ChannelsState(
    val isLoading: Boolean = true,
    val isLoadingMore: Boolean = false,
    val endOfChannels: Boolean = false,
    val channels: List<ChannelItem> = emptyList(),
    val searchQuery: String = "",
)
```

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Create `ChannelItem` class which contains `channel` and `isMuted` fields
- [ ] Remove `isMuted` extension method
- [ ] Make required changes to the channels set of components so that they are able to work with the newly introduced `ChannelItem`.

PR #2672, opened 2021-11-24
[2669] Introduce ChannelItem class

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2669

### 🎯 Goal

Introduce `ChannelItem` class that represents an item on the channel list screen.

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChannelActivity.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChannelActivity.kt
@@ -142,7 +142,7 @@ class ChannelActivity : AppCompatActivity() {
             },
             detailsContent = {
                 Text(
-                    text = ChatTheme.channelNameFormatter.formatChannelName(channelItem.channel),
+                    text = ChatTheme.channelNameFormatter.formatChannelName(it.channel),
                     style = ChatTheme.typography.bodyBold,
                     color = ChatTheme.colors.textHighEmphasis
                 )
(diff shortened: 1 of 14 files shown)


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChannelActivity.kt
```diff
@@ -202,6 +203,7 @@ class ChannelActivity : AppCompatActivity() {
                         .wrapContentHeight()
                         .align(Alignment.Center),
                     shape = RoundedCornerShape(16.dp),
+                    isMuted = listViewModel.isChannelMuted(selectedChannel!!.cid),
```
filbabic, line 206: Let's not do this. I know it's a bit uglier, but I'd rather capture a value and then check that, than have `!!` in code.

Something like:

```
val currentSelectedChannel = selectedChannel

if (currentSelectedChannel != null) {
 // ...
}
```

The state delegate isn't the best in this case, but I think this is okay too.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/ChannelsScreen.kt
```diff
@@ -141,6 +141,7 @@ public fun ChannelsScreen(
                     .align(Alignment.BottomCenter),
                 selectedChannel = selectedChannel!!,
                 currentUser = user,
+                isMuted = listViewModel.isChannelMuted(selectedChannel!!.cid),
```
filbabic, line 144: Same here. Worst case, we could use one of the capturing functions such as let, but I'd still prefer a simple capturing value.

filbabic (changes requested):
Would love to get rid of the !! operator, but the rest of it

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChannelActivity.kt
bychkovdmitry: Yes, works for me. Implemented like that at fist, but this produces IDE warning. Done, captured the value in a temp `currentSelectedChannel` value.

filbabic (approved):
Looks good! :]

PR #2673, opened 2021-11-24
[CAS-1364] Fix - resizable gifs with new view holder

leandroBorgesFerreira:
[CAS-1364](https://stream-io.atlassian.net/browse/CAS-1364)

### 🎯 Goal

Making Giphys autosizable instead of having a pre defined size.

### 🛠 Implementation details

A new view holder was added, just for giphy instead of sharing TextAndAttachmentViewHolder for everything. 

This makes the implementation way easier to understand and debug. Using the old ViewHolder for everything makes it super hard to manage widths because all of the attachments get affected by it. It is necessary to redraw all the attachments to their original intended layout and all of them have multiple layers of views that might show problems with the unwanted shrink/expantion. In the end, a redraw in Giphy should only affect a Giphy and not a whole chain of views. 

On the good side, it simplifies TextAndAttachmentViewHolder as now it is not responsible for giphys anymore. And if makes the resize of Giphys a lot easier as only the ViewHolders of giphy get affected by it which removes the need to worry about giphy resize affecting other attachments.

On the bad size... well it is a big change to our API. Our users will now have to care about one more ViewHolder... Although this is a tricky change, I believe it pays off as customers would be able to change way less code if they only want to change the behaviour of giphys in the SDK, they won't have to re-implement both media and file logic. 

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_20211124-154700 | Screenshot_20211124-154506|

### 🧪 Testing

Customize the giphys in a bunch of different ways and check giphys in a bunch of differents ways... Reply, thread, reply a message with a giphy, reply a giphy with another giphy... 

### ☑️ Checklist

- [x] PR targets the `develop` branch.
- [x] Affected documentation updated (docusaurus, tutorial, CMS (task created)). to check
- [x] Reviewers added.

### 🎉 GIF
giphy
giphy (1)
_That was me trying to fix the problem of giphy's resize interfering with other attachments..._


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -16,17 +18,23 @@ import com.google.android.material.shape.ShapeAppearanceModel
...
 
 internal class MediaAttachmentView : ConstraintLayout {
     var attachmentClickListener: AttachmentClickListener? = null
     var attachmentLongClickListener: AttachmentLongClickListener? = null
     var giphyBadgeEnabled: Boolean = true
 
+    private val maxMediaAttachmentWidth: Int by lazy(LazyThreadSafetyMode.NONE) {
+        (displayMetrics().widthPixels)
+    }
+
```
leandroBorgesFerreira, line 37: This is something that I didn't find a solution for. If I wait for the views to be drawn to calculate the limit of the width based of the container of this View, then I start to have inconsistent results for views in the binding (I may get the current one of the next of the previous, accordingly to the direction of my scroll and the speed). So using methods like `doOnPreDraw` are not an option as it creates many inconsistencies in the final size of the giphy and it also creates many glitches in the scroll. 

At the moment this is a hard limit that only stops the giphy to be bigger then entire screen... Another solution would be to provide a way for customers to customize it using up of a percentage of the screen... Using percentage of the screen looks a bit odd, but dp would create problems for different sizes of screen of the final client. 

Well... this is the problem here and I'm open to tips.

leandroBorgesFerreira: Another observation is that the problem of too big gifs may be a problem that happens so rarely that perhaps can be ignored...

skydoves (commented):
I think this is not the best solution but we can replace the display metrics with measuring; Once we can measure the width of the parent view, and then we can cache the measured width on the ViewHolder and reuse it. 😃 

### GiphyAttachmentViewHolder
```kotlin
private var maxWidth: Int = -1

override fun bindData(data: MessageListItem.MessageItem, diff: MessageListItemPayloadDiff?) {
  ...
  val showGiphy = {
      binding.mediaAttachmentView.showGiphy(
        attachment = data.message.attachments.first(),
        container = binding.messageContainer,
        maxWidth = maxWidth
    )
  }

  if (maxWidth == -1) {
      parent.post {
          maxWidth = parent.width
          showGiphy()
       }
    } else {
         showGiphy()
    }
}
```

### MediaAttachmentView
Added a new `maxWidth` parameter on few functions:
- fun showGiphy(attachment: Attachment, container: View, maxWidth: Int)
- private fun giphyLayoutUpdate(attachment: Attachment, containerView: View, maxWidth: Int)
- private fun parseWidth(giphyInfo: GiphyInfo, maxWidth: Int): Int
- private fun parseHeight(giphyInfo: GiphyInfo, maxWidth: Int): Int

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
leandroBorgesFerreira: Solved using post in parent.

CHANGELOG.md
```diff
@@ -74,6 +74,8 @@ Consider migrating to `stream-chat-android-ui-components` or `stream-chat-androi
 ### 🐞 Fixed
 - Removed ripple effect for attachements in message options.
 ### ⬆️ Improved
+- Giphys are now auto sizable, there's no need to set a predefined size for then as they will follow the size of the sent giphy, respecting the limits of the message size. You can also choose to stay with constant size by setting `streamUiMediaAttachmentConstantSize` to true and setting the desired height with `streamUiMediaAttachmentGiphyHeight`
+- streamUiMediaAttachmentGiphyHeight removed as it is no longer necessary.
```
filbabic, line 78: Why do we remove this if the above comment says you can still use it?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
```diff
@@ -141,6 +144,12 @@ public open class MessageListItemViewHolderFactory {
         )
     }
 
+    protected fun createGiphyAttachmentViewHolder(
```
filbabic, line 147: Comments!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/internal/MessageListItemViewTypeMapper.kt
```diff
@@ -32,12 +35,18 @@ internal object MessageListItemViewTypeMapper {
     }
 
     private fun messageItemToViewType(messageItem: MessageListItem.MessageItem): Int {
+        val message = messageItem.message
+
+        val (linksAndGiphy, _) = message.attachments.partition { attachment -> attachment.hasLink() }
+        val (giphy, _) = linksAndGiphy.partition { attachment -> attachment.type == ModelType.attach_giphy }
```
filbabic, line 41: Can we just use filter here?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/internal/MessageListItemViewTypeMapper.kt
```diff
@@ -32,12 +35,18 @@ internal object MessageListItemViewTypeMapper {
...
         return when {
-            messageItem.message.isError() -> ERROR_MESSAGE
-            messageItem.message.isSystem() -> SYSTEM_MESSAGE
-            messageItem.message.deletedAt != null -> MESSAGE_DELETED
-            messageItem.message.isGiphyEphemeral() -> GIPHY
-            messageItem.message.attachments.isNotEmpty() -> TEXT_AND_ATTACHMENTS
+            message.isError() -> ERROR_MESSAGE
+            message.isSystem() -> SYSTEM_MESSAGE
+            message.deletedAt != null -> MESSAGE_DELETED
+            message.isGiphyEphemeral() -> GIPHY
```
filbabic, line 47: maybe change this to GIPHY_EPHEMERAL rather htan just giphy

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -61,37 +64,63 @@ internal class MediaAttachmentView : ConstraintLayout {
...
-            }
+
+        setOnClickListener { attachmentClickListener?.onAttachmentClick(attachment) }
+        setOnLongClickListener {
+            attachmentLongClickListener?.onAttachmentLongClick()
+            true
         }
+    }
+
+    fun showGiphy(attachment: Attachment, containerView: View, maxWidth: Int) {
```
filbabic, line 81: docs

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/MessageListItemViewHolderFactory.kt
adasiewiczr: + mention this in CHANGELOG. This is a breaking change

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentView.kt
```diff
@@ -139,7 +168,40 @@ internal class MediaAttachmentView : ConstraintLayout {
         }
     }
 
+    private fun Attachment.giphyInfo(field: GiphyInfoType): GiphyInfo? {
```
adasiewiczr, line 171: It would be really nice to do parsing earlier and put those parameters directly in the `Attachment` object but we would have to make them nullable. Maybe `Attachment` model can be improved in the future

filbabic:
Closing this one as #2932 will take over this work!

PR #2689, opened 2021-11-25
[CAS-1454] [WIP] Basic MessageComposer implementation 

samiuelson:
https://stream-io.atlassian.net/browse/CAS-1454

### 🎯 Goal

Implement basic `MessageComposer` component allowing deep customization of its contents using slot API. For now it's marked as `@ExperimentalStreamChatApi` until all the features are implemented.

### 🛠 Implementation details

1. Created `MessageComposerViewModel` wrapping `MessageComposerController`
2. Implemented base `MessageComposerView`
3. Imoplemented binding - `MessageComposerView::bindView`

### 🎨 UI Changes

https://user-images.githubusercontent.com/4527432/143437110-1f827ecf-0655-4908-9cc5-891b76301a6d.mp4

### 🧪 Testing

In `ChatFragment` replace `MessageInputView` with `MessageComposerView` and `MessageInputViewModel` with `MessageComposerViewModel`. See if it's possible to send a plain text message.

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+    private lateinit var binding: StreamUiMessageComposerBinding
+
+    /**
+     * @property messageInputState The current state of the [MessageComposerView] distributed using [Flow] API.
+     *
+     * Whenever the value of this property changes updated the [MessageComposerView] re-renders automatically.
+     * If you are using custom [View] implementation for [leadingContent], [centerContent], or [trailingContent] you may
+     * want to collect [messageInputState] items to update them when the state changes.
+     */
+    public val messageInputState: MutableStateFlow<MessageInputState> = MutableStateFlow(MessageInputState())
```
filbabic, line 37: Should this potentially be within the ViewModel? I know Views commonly have state they expose, but if users want to access it, they should be able to do it through the ViewModel.

Also - we could expose something like this through the `Controller` instead and just observe it here.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+     */
+    public var onInputTextChanged: (String) -> Unit = {}
+
+    private val defaultLeadingContent: View by lazy {
+        val binding = StreamUiMessageComposerDefaultLeadingContentBinding.inflate(
+            streamThemeInflater,
+            binding.trailingContent,
+            false
+        )
+        binding.attachmentsButton.isVisible = false
```
filbabic, line 60: maybe add this to an `apply()`block and just return the result.root?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+     * @property onCancelAction Handler invoked when the user cancels the active action (in Reply or Edit modes).
+     */
+    public var onCancelAction: () -> Unit = {}
+
+    /**
+     * @property onInputTextChanged Handler invoked when the user enters text in message input field.
+     */
+    public var onInputTextChanged: (String) -> Unit = {}
+
+    private val defaultLeadingContent: View by lazy {
```
filbabic, line 54: Do we need this default leading content as a property or can it be somehow lazily added to the UI when the user starts setting things up?

It seems it's not "lazy" at all, given that we immediately create the `leadingContent` within the `init()` block and create this property. Maybe it can all be inlined there?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+
+    private val defaultCenterContent: View by lazy {
+        val binding = StreamUiMessageComposerDefaultCenterContentBinding.inflate(
+            streamThemeInflater,
+            binding.trailingContent,
+            false
+        )
+        binding.messageEditText.addTextChangedListener(object : TextWatcherAdapter() {
+            override fun afterTextChanged(s: Editable) {
+                onInputTextChanged(s.toString())
```
filbabic, line 87: What happens if the user changes `onInputTextChanged` but keeps the default content - do we have any possible scoping/capturing issues here, where the old listener is bound and we have wrong behavior?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+            }
+        })
+        binding.clearCommandButton.isVisible = false
+        binding.selectedAttachmentsContainer.isVisible = false
+        binding.commandBadge.isVisible = false
+        binding.messageReplyView.isVisible = false
+        CoroutineScope(Dispatchers.Main).launch {
```
filbabic, line 94: Why do we collect the state here, inside the default content? This seems like more of an `init()` block. In this case, if the user doesn't collect the state properly if they override the UI, then it won't update if I understood it correctly

Also - do we cancel this dispatcher if the user replaces the content? We don't have access to the Scope or the Job to call cancel properly.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+
+    private val defaultTrailingContent: View by lazy {
+        val binding = StreamUiMessageComposerDefaultTrailingContentBinding.inflate(
+            streamThemeInflater,
+            binding.trailingContent,
+            false
+        ).apply {
+            CoroutineScope(Dispatchers.Main).launch {
```
filbabic, line 129: Same here - we're launching multiple scopes and coroutines in the Main thread to collect the same state - would it be easier to just collect all the state from the ViewModel in one place and update all the areas which care about the state?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+    public constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context,
+        attrs,
+        defStyleAttr) {
+        init()
+    }
+
+    private fun init() {
+        binding = StreamUiMessageComposerBinding.inflate(streamThemeInflater, this)
+        binding.leadingContent.apply {
+            removeAllViews()
```
filbabic, line 169: Should we maybe check if there are existing views and everything is set up, before calling removing all views and re-adding them?

In case of configuration changes, this will just reset all the state, right? That's also why having the MessageInputState in the ViewModel is kind of better, because it's persisted across orientation changes.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+        }
+        binding.centerContent.apply {
+            removeAllViews()
+            addView(centerContent())
+        }
+        binding.trailingContent.apply {
+            removeAllViews()
+            addView(trailingContent())
+        }
+        invalidate()
```
filbabic, line 180: Why do we invalidate at the end?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -0,0 +1,195 @@
...
+}
+
+/**
+ * Represents the state within the message input.
+ *
+ * @param inputValue The current text value that's within the input.
+ * @param attachments The currently selected attachments.
+ * @param action The currently active [MessageAction].
+ */
+public data class MessageInputState(
```
filbabic, line 191: We could just reuse this too and put it in ui-common

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
```diff
@@ -0,0 +1,24 @@
+package io.getstream.chat.android.ui.message.composer
+
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.coroutineScope
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
+
+public fun MessageComposerViewModel.bindView(view: MessageComposerView, lifecycleOwner: LifecycleOwner) {
+
+    view.onSendMessageAction = {
```
filbabic, line 10: So basically if the user decides to bind the view to the ViewModel, we override the `onSendMessageAction`? As well as `onInputTextChanged`. 

How do our users customize these then? If they want to intercept the changes and all.

Do they first have to bind the view and then override the changes?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
```diff
@@ -0,0 +1,24 @@
...
+        val text = input.value
+        val attachments = selectedAttachments.value
+        val message = buildNewMessage(text, attachments)
+        sendMessage(message)
+    }
+
+    view.onInputTextChanged = { setMessageInput(it) }
+
+    lifecycleOwner.lifecycle.coroutineScope.launch {
```
filbabic, line 19: why not `lifecycleOwner.lifecycleScope`? Do we have access to that?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
```diff
@@ -0,0 +1,24 @@
...
+        input.collect { input ->
+            view.messageInputState.value = view.messageInputState.value.copy(inputValue = input)
```
filbabic, line 21: This wouldn't be required if the state came from the ViewModel - could just be a single line to set the input for the ViewModel

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_composer.xml
```diff
@@ -0,0 +1,44 @@
...
+            android:layout_height="wrap_content"
+            android:layout_gravity="center"
+            android:layout_weight="1"
+            app:layout_constraintLeft_toRightOf="@+id/leadingContent"
+            app:layout_constraintRight_toLeftOf="@+id/trailingContent"
+            app:layout_constraintTop_toTopOf="parent"
+            />
+
+        <FrameLayout
+            android:id="@+id/trailingContent"
```
filbabic, line 33: What happens if the Center content is too big? Because of `wrap_content` it could fall outside of bounds, or push the trailing content out of the screen. Sme goes for leading content

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_composer_default_center_content.xml
```diff
@@ -0,0 +1,106 @@
...
+        tools:text="giphy"
+        />
+
+    <androidx.appcompat.widget.AppCompatEditText
+        android:id="@+id/messageEditText"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:background="@null"
+        android:inputType="textCapSentences|textMultiLine"
+        android:maxLines="7"
```
filbabic, line 79: Any way to customize all of these hardcoded numbers/attributes?

filbabic (changes requested):
Hey! Would love to discuss this more and see an example of using and customizing this component in the UI as I feel like with some of the API design decisions, it might be hard to customize it as a user or we might run into edge case problems and we should try to avoid those as soon as possible!

Can you take a look at the comments and we can chat more about it next week? :]

Thanks!

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_composer_default_center_content.xml
samiuelson: My plan was to use the TransformStyle and style attrs - the same API type that's used in `MessageInputView` and apply those style transformations only to the default contents.

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_composer.xml
samiuelson: Currently the `centerContent` is configured to take all the remaining horizontal space in the container. IMO, this should be adjusted so that it also has minimal width and height set. Also, we should configure leading and trailing containers to have minimal sizes too. Not sure how to do it properly.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
samiuelson: Yup, view model could expose `messageInputState`. Would it work with compose as well or should it be XML-specific?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
samiuelson: The motivation here was to allow user access to `messageInputState` in the scope of `MessageComposerView` when they override `MessageComposerView.trailingContent` for example. But maybe we could force them to subscribe to ViewModel and bind the `trailingContent: View` on their own and finally pass it to `MessageComposer` 🤔

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
samiuelson: `lazy` will make sense only when user will be able to decide to hide one of content views using style or attrs-based API. Style will be parsed in init() too. Otherwise we are not able to make it effectively lazy.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
samiuelson: Yes, this would be good. Default content definition would land `MessageComposerViewModelBinding` then `MessageComposerView` implementation would be cleaner.

samiuelson: I've refactored `messageInputState` in the 8dfa367 commit 👇🏼 
* `messageInputState` is now exposed by `MessageComposerViewModel.messageInputState: StateFlow<MessageInputState>`
* Default content is defined inside `MessageComposerViewModel::bindView` 
* `MessageComposerView` is stateless now

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
samiuelson: For now, I've exposed it in ui-components `MessageComposerViewModel` only.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
samiuelson: I've simplified this in be6f49d. Letting the custom/default content elements communicate with view model directly. This will simplify actions flow a lot.

filbabic: I think it's not a matter of simplification - we just need to expose the ability for our users to override the UI actions and intercept them - it's what we've been exposing for a while. 

The issue here was that the user can set those actions before calling `bindView()` and then call bind and those actions would be overriden.

We can build the logic in a way that we allow for the callbacks to be overriden and then we can set up default behavior for users so they don't have to do anything by default.

stream-chat-android-ui-components/src/main/res/layout/stream_ui_message_composer.xml
filbabic: We could always use weights or something like that, constraints alternatively. Since the Composer is a Constraint layout, we can use proper constraints and constraint chain attributes to make it work as it's intended.

filbabic (changes requested):
Left a few comments - we can discuss this and see if we can make the API a bit more flexible for our users.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModel.kt
```diff
@@ -0,0 +1,146 @@
...
+public class MessageComposerViewModel(
+    private val messageComposerController: MessageComposerController,
+) : ViewModel() {
+
+    private val _messageInputState: MutableStateFlow<MessageInputState> = MutableStateFlow(MessageInputState())
+
+    /**
+     * UI state of the message input component.
+     */
+    public val messageInputState: StateFlow<MessageInputState> = _messageInputState
```
samiuelson, line 37: I'd like to move this to `MessageComposerController` in a separate PR as this will potentially touch compose module

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/state/MessageInputState.kt
```diff
@@ -0,0 +1,16 @@
...
+import io.getstream.chat.android.client.models.Attachment
+
+/**
+ * Represents the state within the message input.
+ *
+ * @param inputValue The current text value that's within the input.
+ * @param attachments The currently selected attachments.
+ * @param action The currently active [MessageAction].
+ */
+public data class MessageInputState(
```
samiuelson, line 12: Just moved to common. I'd prefer refactoring compose usages in a separate PR

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/DefaultCenterContent.kt
```diff
@@ -0,0 +1,53 @@
...
+import android.text.Editable
+import android.util.AttributeSet
+import android.widget.FrameLayout
+import androidx.core.view.isVisible
+import com.google.android.material.internal.TextWatcherAdapter
+import io.getstream.chat.android.common.state.MessageInputState
+import io.getstream.chat.android.ui.common.extensions.internal.streamThemeInflater
+import io.getstream.chat.android.ui.databinding.StreamUiMessageComposerDefaultCenterContentBinding
+
+internal class DefaultCenterContent : FrameLayout {
```
samiuelson, line 13: Serving default content as custom `View` implementation will allow to easily pass style, and attrs from the `MessageComposerView` in the future

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -1,56 +1,31 @@
...
-     */
-    public var leadingContent: MessageComposerView.(ViewGroup) -> View = { emptyView() }
-        set(value) {
-            field = value
-            binding.leadingContent.apply {
-                removeAllViews()
-                addView(value(this))
-            }
-        }
+    public var onSendMessage: () -> Unit = {}
```
filbabic: Should we expose a message as a parameter here, so that our users can potentially override it/add extra data?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -1,56 +1,31 @@
...
-     */
-    public var trailingContent: MessageComposerView.(ViewGroup) -> View = { emptyView() }
-        set(value) {
-            field = value
-            binding.trailingContent.apply {
-                removeAllViews()
-                addView(value(this))
-            }
-        }
+    public var onClearInput: () -> Unit = {}
```
filbabic: Are we missing `onDismissMessageAction`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
```diff
@@ -65,31 +40,125 @@ public class MessageComposerView : ConstraintLayout {
...
+    }
+
+    public fun renderState(state: MessageInputState) {
+        (binding.trailingContent.children.first() as? DefaultTrailingContent)?.renderState(state)
+        (binding.centerContent.children.first() as? DefaultCenterContent)?.renderState(state)
+        (binding.leadingContent.children.first() as? DefaultLeadingContent)?.renderState(state)
+    }
+}
+
+internal class DefaultLeadingContent : FrameLayout {
```
filbabic: Should these  be inner classes, but public maybe? What if our users want to use them but customize the behavior a bit?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModel.kt
```diff
@@ -47,6 +51,15 @@ public class MessageComposerViewModel(
      */
     public val lastActiveAction: Flow<MessageAction?> = messageComposerController.lastActiveAction
 
+    init {
+        viewModelScope.launch {
+            input.collect {
+                val oldState = _messageInputState.value
```
filbabic, line 57: why not just update the messageInputState whenever we call `setMessageInput`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
```diff
@@ -1,80 +1,30 @@
...
-        )
-        binding.apply {
-            attachmentsButton.isVisible = false
-            commandsButton.isVisible = false
-        }.root
+public fun MessageComposerViewModel.bindView(
+    view: MessageComposerView,
+    lifecycleOwner: LifecycleOwner,
+) {
+    view.onSendMessage = {
```
filbabic: We still don't allow the user to customize these actions if they want, we just override them

filbabic (changes requested):
Left a few more comments - I think we're going in the positive direction, but still a few more polish points left to clear up!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
samiuelson: My understanding is this would enforce making `MessageComposerView` stateful. If `MessageComposerView` was responsible for instantiating `Message` it would need to store state internally 🤔

samiuelson: IMO, this is something we can provide via `MessageComposerViewModel::bindView` and in case `MessageComposerView` is used without the default view model it's the responsibility of the external controller to hold the state and allow instantiating the `Message`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerView.kt
samiuelson: Yes. This makes sense to me

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/composer/MessageComposerViewModelBinding.kt
samiuelson: Added possibility for customizing them in `bindView` function's params

filbabic (approved):
To me this looks really good and very promising for our users! :]

We can continue with our plan - have a feature/working branch where we'd point all these changes and prepare them until they are fully ready. 

I would also love to have a branch that features this implementation as an example, so that we can see how the tests work and how the API is used on the user side!

Issue #2721, opened 2021-11-30
[Compose] - Improve naming of Messages components and state


filbabic:
### Context

Currently, our components and state are named the same and it can be confusing when it comes to imports and usage.

We should rename our state classes to feature the "State" suffix so that we can clearly distinguish them.

E.g. 

`MessageOptions` -> UI component that shows a list of message options
`MessageOptionState` -> The class that holds the state required to render a singular message option.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Go through the message components and see where it makes sense to update the components/state relations.

PR #2726, opened 2021-12-01
General improvements & API cleanup

filbabic:
#2721 

### 🎯 Goal

As per our discussion, we're renaming components and state to make more sense and to be easily discoverable.

I've also cleaned up the SelectedMessageOverlay API a little and fixed some UI issues. Generally, let me know if any of the names here don't make sense an we can clean them up.

My goal is for us to resolve this before we push out the "stable" compose version - we can keep this PR around, batch the changes and then merge it after the next release, so that we don't introduce breaking changes every release.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF
name


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamDimens.kt
```diff
@@ -60,7 +66,10 @@ public data class StreamDimens(
             attachmentsContentFileUploadWidth = 250.dp,
             threadSeparatorVerticalPadding = 8.dp,
             threadSeparatorTextVerticalPadding = 2.dp,
-            messageOverlayActionItemHeight = 40.dp
+            messageOverlayActionItemHeight = 40.dp,
+            messageOptionsMaxWidth = 200.dp,
+            messageOptionsMaxHeight = 300.dp,
```
bychkovdmitry, line 71: Setting this to a smaller value to allow scroll in a popup will look weird. I think for both cases scroll should not be possible:
- The message options overlay as it is now
- Our future implementation of message options in a bottom sheet

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
```diff
@@ -187,24 +201,36 @@ internal fun ReactionOptionItem(
  * in the list, as those are not as important.
  *
  * @param message Message to show.
+ * @param isMine If the message is owned by the current user.
```
bychkovdmitry, line 204: Modifier KDoc

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/list/MessageItemState.kt
```diff
@@ -1,26 +1,26 @@
-package io.getstream.chat.android.compose.state.messages.items
+package io.getstream.chat.android.compose.state.messages.list
```
bychkovdmitry, line 1: 👍

bychkovdmitry (approved):
Great job renaming stuff and improving the UI of the message options overlay!

samiuelson (approved):
LGTM

PR #2728, opened 2021-12-01
2702 - Added message pinning

### LINKED ISSUE #2702: [Compose] - Support Message Pin action & Pinned message UI
### Context

We are missing the pin message action and the pinned message UI, so we need to add it to the list of our message options.

### Acceptance criteria

- [ ] Add pin message action to the message options in channels.
- [ ] Add the pinned message UI for pinned messages.

filbabic:
Fixes #2702 

### 🎯 Goal

Adding message pinning as well as pinned message UI to message items.

### 🛠 Implementation details

Added a new option to the list of default message options, added appropriate UI to the default UI, added message list item UI as well and the business logic required to pin and unpin messages..

### 🎨 UI Changes

Previously we didn't have pinned messages, as compared to now:

| Before | After |
| --- | --- |
| Screenshot_1638364104 | Screenshot_1638364144 |

We also have the following options:

| Unpin | Pin |
| --- | --- |
| Screenshot_1638364099 | Screenshot_1638364111 |

### 🧪 Testing

Open any channel and long tap on an item - you should see options to pin/unpin messages.

You should also see different UI for messages which are pinned - the background should be highlighted and it should give you information on who pinned the message.

### ☑️ Checklist

~~- [ ] PR targets the `develop` branch~~ -> Will point to the naming branch to avoid conflicts
- [x] Reviewers added

### 🎉 GIF

pinhead


skydoves (approved):
Looks great to me! 👍

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/list/MessageFocusState.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Represents the state when the message is currently being focused.
+ */
+public object MessageFocused : MessageFocusState()
+
+/**
+ * Represents the state when we've removed the focus from the message.
+ */
+public object FocusRemoved : MessageFocusState()
```
skydoves, line 16: This is just a suggestion, I think if we have the same prefix (Message-) would be better to maintain consistency. 😄

bychkovdmitry (approved):
Works really well!

PR #2734, opened 2021-12-02
[2733] Enable scrolling behavior in SelectedMessageOverlay

MarinTolic:
### 🎯 Goal

Enable scrolling behavior in `SelectedMessageOverlay` in order to avoid clipping the message or clipping or fully excluding `MessageOptions` from the screen resulting in an inability to perform any actions over messages.

### 🛠 Implementation details

Enabled vertical Scrolling in `SelectedMessageOverlay` by applying `.verticalScroll(ScrollState)` to the `Box` `Modifier` inside `SelectedMessageOverlay`. 

Switched from implementing a `LazyColumn` for displaying `MessageOptionItem`s to using a `Column` so that nested scrolling would be avoided. Used key utility Composable to optimize recomposition since the Composables were called from the same call site.

### 🎨 UI Changes

Enabled vertical scrolling in `SelectedMessageOverlay`

| Before | After |
| --- | --- |
| before | scroll_short |

### 🧪 Testing

Testing can be done manually by long pressing on a really long message and opening the aforementioned overlay. After that action is accomplished one should scroll down until all of `MessageOptions` is visible and test it by performing an action such as replying to a message, editing a message or deleting a message.

### ☑️ Checklist

- [x] Reviewers added

### 🎉 GIF

scrolling

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
@@ -17,15 +17,17 @@ import androidx.compose.foundation.layout.sizeIn
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.layout.wrapContentWidth
-import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyRow
 import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.foundation.verticalScroll
 import androidx.compose.material.Icon
 import androidx.compose.material.Surface
 import androidx.compose.material.Text
 import androidx.compose.material.ripple.rememberRipple
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.key
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Alignment.Companion.End
@@ -78,6 +80,8 @@ public fun SelectedMessageOverlay(
     onMessageAction: (MessageAction) -> Unit,
     onDismiss: () -> Unit,
 ) {
+    val boxScrollState = rememberScrollState()
+
     val ownReactions = message.ownReactions
 
     val reactionOptions = reactionTypes.entries
@@ -93,6 +97,7 @@ public fun SelectedMessageOverlay(
         modifier = Modifier
             .fillMaxSize()
             .background(ChatTheme.colors.overlay)
+            .verticalScroll(boxScrollState)
             .clickable(
                 onClick = onDismiss,
                 indication = null,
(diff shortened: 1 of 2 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
```diff
@@ -281,19 +286,21 @@ public fun MessageOptions(
...
-                        .fillMaxWidth()
-                        .height(0.5.dp)
-                        .background(color = ChatTheme.colors.borders)
-                )
+        Column(modifier) {
+            options.forEach { option ->
+                key(option) {
+                    Spacer(
+                        modifier = Modifier
+                            .fillMaxWidth()
```
skydoves, line 294: Just for curious, can we use [matchParentSize](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/BoxScope#(androidx.compose.ui.Modifier).matchParentSize()) instead of the `fillMaxWidth()`?

skydoves (approved):
Looks good to me! 👍

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
MarinTolic: This is a great question Jaewoong and it's something I've been thinking about yesterday.

Unfortunately [ColumnScope](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/ColumnScope) doesn't have [.mathParentSize()](https://developer.android.com/reference/kotlin/androidx/compose/foundation/layout/BoxScope#(androidx.compose.ui.Modifier).matchParentSize()) or similar width related behavior, but the same effect can be achieved through the use of [Intrinsic measurements](https://developer.android.com/jetpack/compose/layouts/intrinsic-measurements). I agree with you that simply filling the available space inside the parent would be a much more organic solution that implementing using `.fillMaxWidth()` and then using hard dimensions to cut down on size.

Were you possibly thinking of wrapping the elements inside a `Box` or is there something I'm missing?

skydoves: Oops, sorry I was mis pointed the line. I wanted to point the line 98 in this class. 😄

filbabic: We could put the matchParentSize to the column (as it's within a box) and then match the parent/wrap the items?

Alternatively, what Marin suggested makes sense to me - but if it's more work and it's overkill for this (given that we'll migrate to new UX for the component) then we can just leave it as-is.

filbabic (approved):
Looks good to me - as I mentioned, we can have these improvements for the current iteration of the component, but we're likely to change all this, this month, so maybe we can shift our focus on doing the UI/UX update instead of the work to make this component better? :]

You can take the initiative on this one if you'd like, I feel like that would be a great "first task" for a more wholesome improvement!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/overlay/SelectedMessageOverlay.kt
MarinTolic: I've tried making it work by using the `matchParentSize()` approach but unfortunately it doesn't result in the desired outcome as it aligns the content to the top and `MessageOptions` fill maximum available horizontal space.

I already have a working solution for my proposition using `IntrinsicSize` but understand that a migration is planned, so any outcome you decide is okay by me.

I would just like a clear input on whether we will:

1. Keep this PR as is and merge it
2. Add a commit using `IntrinsicSize` and review it
3. Sync up and see if we can make it work using `.matchParentSize()`
4. Some other proposition

filbabic: Let's merge as-is and then move on with the newly proposed UI/UX in the next few weeks!

bychkovdmitry (approved):

PR #2735, opened 2021-12-02
[CAS-1482] Fix - Query controller

leandroBorgesFerreira:
[CAS-1482](https://stream-io.atlassian.net/browse/CAS-1482)

### 🎯 Goal

Fix of channels not being updated when the filter is requesting channels that the user is **not** a member. 

### 🛠 Implementation details

Removing or adding the user depends on the filter, not the event that arrived. If the channel fails if the filter, it must be removed, if it passes, it must be added. 

### 🎨 UI Changes

Left side - Normal filtering. 
Right side - Displaying only channels where the user is not a member
<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/144480750-5b3ce670-815f-4752-adef-e8e4b0b6d5e9.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/144480611-b859b562-1d86-41d5-b767-0bdbbaf16919.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Change the filter to 
```
filter = Filters.and(
    Filters.nin("members", listOf(ChatClient.instance().getCurrentUser()?.id ?: "")),
    Filters.or(Filters.notExists("draft"), Filters.eq("draft", false)),
)
```
instead of:
```
filter = Filters.and(
    Filters.`in`("members", listOf(ChatClient.instance().getCurrentUser()?.id ?: "")),
    Filters.or(Filters.notExists("draft"), Filters.eq("draft", false)),
)
```

Please try other filters aswell.
### ☑️ Checklist

- [x] Comparison screenshots added for visual changes.
- [x] Reviewers added.

### 🎉 GIF

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
@@ -405,8 +405,9 @@ public data class NotificationRemovedFromChannelEvent(
     override val cid: String,
     override val channelType: String,
     override val channelId: String,
+    override val channel: Channel,
     override val member: Member,
-) : CidEvent(), HasMember
+) : CidEvent(), HasMember, HasChannel
 
 /**
  * Triggered when a message reaction is deleted
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandler.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandler.kt
@@ -6,6 +6,7 @@ import io.getstream.chat.android.client.api.models.QueryChannelsRequest
 import io.getstream.chat.android.client.call.await
 import io.getstream.chat.android.client.events.ChannelUpdatedByUserEvent
 import io.getstream.chat.android.client.events.ChannelUpdatedEvent
+import io.getstream.chat.android.client.events.HasChannel
 import io.getstream.chat.android.client.events.NotificationAddedToChannelEvent
 import io.getstream.chat.android.client.events.NotificationMessageNewEvent
 import io.getstream.chat.android.client.events.NotificationRemovedFromChannelEvent
@@ -47,12 +48,16 @@ internal class DefaultChatEventHandler(private val client: ChatClient, private v
     override fun handleNotificationAddedToChannelEvent(
         event: NotificationAddedToChannelEvent,
         filter: FilterObject,
-    ): EventHandlingResult = fireRequestIfChannelIsAbsent(event.channel, filter)
+    ): EventHandlingResult {
+        return handleMemberUpdate(event, event.cid, filter)
+    }
 
     override fun handleChannelUpdatedByUserEvent(
         event: ChannelUpdatedByUserEvent,
         filter: FilterObject,
-    ): EventHandlingResult = EventHandlingResult.Skip
+    ): EventHandlingResult {
+        return handleMemberUpdate(event, event.cid, filter)
+    }
 
     override fun handleChannelUpdatedEvent(event: ChannelUpdatedEvent, filter: FilterObject): EventHandlingResult =
         EventHandlingResult.Skip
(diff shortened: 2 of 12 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandler.kt
```diff
@@ -47,12 +48,16 @@ internal class DefaultChatEventHandler(private val client: ChatClient, private v
...
+    ): EventHandlingResult {
+        return handleMemberUpdate(event, event.cid, filter)
+    }
 
     override fun handleChannelUpdatedByUserEvent(
         event: ChannelUpdatedByUserEvent,
         filter: FilterObject,
-    ): EventHandlingResult = EventHandlingResult.Skip
+    ): EventHandlingResult {
+        return handleMemberUpdate(event, event.cid, filter)
```
ogkuzmin, line 59: We can't do this. `ChannelUpdatedByUserEvent` is often update and this logic isn't generic for all possible use-cases (like messaging and live stream. messaging requires a user being a member of channel and live stream doesn't). 
If we do such requests always out-of-box we will throttle our BE and customers reach their rate limit too soon. We can suggest such change to particular customers that works fine in their case. They can inject their own `ChatEventHandler` into `ChannelListViewModel`. We have docs about it [here](https://getstream.io/chat/docs/sdk/android/ui/components/channel-list/#chateventhandler)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandler.kt
```diff
@@ -47,12 +48,16 @@ internal class DefaultChatEventHandler(private val client: ChatClient, private v
     override fun handleNotificationAddedToChannelEvent(
         event: NotificationAddedToChannelEvent,
         filter: FilterObject,
-    ): EventHandlingResult = fireRequestIfChannelIsAbsent(event.channel, filter)
+    ): EventHandlingResult {
+        return handleMemberUpdate(event, event.cid, filter)
```
ogkuzmin, line 52: In the previous version it did the same. Right?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
```diff
@@ -405,8 +405,9 @@ public data class NotificationRemovedFromChannelEvent(
     override val cid: String,
     override val channelType: String,
     override val channelId: String,
+    override val channel: Channel,
     override val member: Member,
-) : CidEvent(), HasMember
+) : CidEvent(), HasMember, HasChannel
```
ogkuzmin, line 410: Nice catch 👍

ogkuzmin (commented):
@adasiewiczr Nice to have your thoughts here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandler.kt
leandroBorgesFerreira: In this case it can remove the user in an Add events, accordingly with the filter, so the behaviour is a bit different now

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
leandroBorgesFerreira: ty!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandler.kt
leandroBorgesFerreira: I took a look at the code and the update wouldn't happen many times, only once. The when case: 
```
when {
                filterPassed && !hasChannel -> EventHandlingResult.Add(channel)

                !filterPassed && hasChannel -> EventHandlingResult.Remove(cid)

                else -> EventHandlingResult.Skip
            }
```
Will once go to the `else` condition because the channel will be updated, so it won't ask to add or remove a user without need.

leandroBorgesFerreira: It is filter that does many queries, not the add/remove. Ok, let's add a skip in this event.

ogkuzmin (approved):

PR #2736, opened 2021-12-02
[CAS-1449] Send image attachment dimensions

leandroBorgesFerreira:
[CAS-1449](https://stream-io.atlassian.net/browse/CAS-1449)
### 🎯 Goal

Check a way to send images dimensions. 

### 🛠 Implementation details

Adding dimensions to AttachmentMetaData and using it instead of Pair<File, String>

### 🧪 Testing

Check the logs of the attachment sent to backend.

### ☑️ Checklist

- [ ] Reviewers added

### 🎉 GIF

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/model/AttachmentMetaData.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/model/AttachmentMetaData.kt
@@ -20,6 +20,8 @@ public data class AttachmentMetaData(
     var isSelected: Boolean = false
     var selectedPosition: Int = 0
     var videoLength: Long = 0
+    var width: Int = 0
+    var heigth: Int = 0
 
     public constructor(attachment: Attachment) : this(
         type = attachment.type,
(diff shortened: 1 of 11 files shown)


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/model/AttachmentMetaData.kt
```diff
@@ -20,6 +20,8 @@ public data class AttachmentMetaData(
     var isSelected: Boolean = false
     var selectedPosition: Int = 0
     var videoLength: Long = 0
+    var width: Int = 0
+    var heigth: Int = 0
```
samiuelson, line 25: 1. Changelog should be updated about this change.
2. I'd appreciate having docs explaining what state of those props user can expect when the Attachment is not an image (e.g. mp3 file)
3. Shouldn't we add @JvmOverloads to the constructor given we have bunch of default values set here?

leandroBorgesFerreira: Makes sense 👍 . For now let's hold this PR a bit until we add support for autosizable images. This is more to show that it is possible to send the dimensions than the final solution.

leandroBorgesFerreira:
We won't need this PR, as the backend will take care of image dimensions.

Issue #2638, opened 2021-11-16
Implement also send to channel checkbox


bychkovdmitry:
### Context

We need to implement the "Also send to channel" checkbox which should be shown when the message composer is in "thread" mode.

### Change Proposal

Add the corresponding checkbox to the `MessageComposer`.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] The "Also send to channel" checkbox is shown in the `MessageComposer` in thread mode.
- [ ] A different text is shown for one-to-one channels and group channels: "Also send to channel" or "Also send as direct message".
- [ ] The strings are translated to all the supported languages.
- [ ] The message can be found in the channel when the checkbox is checked and the thread reply is sent.
- [ ] Confirm with the design team if we need a way to navigate from the reply sent to channel to the corresponding  thread.

PR #2746, opened 2021-12-07
[2638] Add support for also sent to channel

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2638

### 🎯 Goal

Add the "Also send to channel" checkbox to the message composer.

### 🎨 UI Changes

https://user-images.githubusercontent.com/9600921/145020675-e64085ac-f5da-4f38-8b00-42030ea2c1d3.mov

### ☑️ Checklist

- [x] Reviewers added


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -544,9 +550,10 @@ public class MessageListViewModel(
      * [MessageItemGroupPosition.Bottom] or [MessageItemGroupPosition.None] if the message isn't in a group.
      *
      * @param messages The messages we need to group.
+     * @param isInThread If we are in inside a thread.
      * @return A list of [MessageListItemState]s, each containing a position.
      */
-    private fun groupMessages(messages: List<Message>): List<MessageListItemState> {
+    private fun groupMessages(messages: List<Message>, isInThread: Boolean): List<MessageListItemState> {
```
bychkovdmitry, line 556: A small fix. Thread and channel messages are observed and combined at the same time, therefore we need to pass this parameter explicitly.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageComposerState.kt
```diff
@@ -25,4 +28,6 @@ public data class MessageComposerState(
     val mentionSuggestions: List<User> = emptyList(),
     val commandSuggestions: List<Command> = emptyList(),
     val cooldownTimer: Int = 0,
+    val messageMode: MessageMode = MessageMode.Normal,
+    val showInChannel: Boolean = false,
```
filbabic, line 32: This is a bit vague as to what is `showInChannel` - what will we show? The message, the composer etc.

Maybe renaming to `alsoSendToChannel` might be more clear.

But I'm not 100% sure as it's the same component for 2 places where we have messages.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -82,6 +90,19 @@ public fun MessageComposer(
     onCancelAction: () -> Unit = { viewModel.dismissMessageActions() },
     onMentionSelected: (User) -> Unit = { viewModel.selectMention(it) },
     onCommandSelected: (Command) -> Unit = { viewModel.selectCommand(it) },
+    onShowInChannelChecked: (Boolean) -> Unit = { viewModel.setShowInChannel(it) },
```
filbabic, line 93: Maybe `selected` instead of `checked`? What do you think?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -82,6 +90,19 @@ public fun MessageComposer(
...
+            onCancelAction = onCancelAction
+        )
+    },
+    footerContent: @Composable ColumnScope.(MessageComposerState) -> Unit = {
+        DefaultMessageComposerFooterContent(
+            messageComposerState = it,
+            onShowInChannelChecked = onShowInChannelChecked
+        )
+    },
     mentionPopupContent: @Composable (List<User>) -> Unit = {
```
filbabic, line 106: Maybe we could also add the slot API for the send button, given that we're updating the composer? :]

filbabic (commented):
Just a few comments!

Looks good :]

CHANGELOG.md
```diff
@@ -96,6 +96,7 @@ Consider migrating to `stream-chat-android-ui-components` or `stream-chat-androi
 - Added support for Giphy command.
 - Added message pinning to the list of message options
 - Added pinned message UI
+- Added a checkbox to the `MessageComposer` component, that allows to display a thread message in the parent channel.
```
bielikb, line 99: I see that [unlike iOS](https://github.com/GetStream/stream-chat-swift/blob/main/CHANGELOG.md#-changed), actually you arent referencing to the pull request that provided the change. 
Treat this as a comment, rather than an action to take within this PR :)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageComposerState.kt
bielikb: or the other variation could be -> `sendsMessageToChannel` :)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -265,6 +300,66 @@ public fun MessageComposer(
...
+ *
+ * @param messageComposerState The state of the message composer.
+ * @param onShowInChannelChecked Handler when the user checks the also send to channel checkbox.
+ */
+@Composable
+public fun DefaultMessageComposerFooterContent(
+    messageComposerState: MessageComposerState,
+    onShowInChannelChecked: (Boolean) -> Unit,
+) {
+    if (messageComposerState.messageMode is MessageMode.MessageThread) {
```
bielikb, line 338: This is a nitpick and feel free to disregard ;)

For better readability, could we "guard" against this statement here? ```(messageComposerState.messageMode is MessageMode.MessageThread)```

Dunno wha'ts the most fancy way to do it in Kotlin, but here's the logic that I have on my mind:

```
if !(messageComposerState.messageMode is MessageMode.MessageThread) return
```

This way the reader no longer needs to account for this condition when reading the body of this function.

bielikb:
This work looks terrific 🔥. 
I left a couple of minor comments, treat them as suggestions rather than "the work that needs to be done."

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/composer/MessageComposerState.kt
bychkovdmitry: Yeah, that's true `showInChannel` is indeed misleading here. Will use a more verbose alternative here.

CHANGELOG.md
bychkovdmitry: I'm moving this discussion to Slack

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
filbabic: In compose - it's actually a practice to declare when a UI component will be emitted, rather than when it's not

That way we know that in case a condition isn't met, it won't be called and we'll have this empty call

it's a bit different in swiftUI I guess because you do have objects and references, so you can always refer to stuff

filbabic (approved):

adasiewiczr (approved):

PR #2747, opened 2021-12-07
[CAS-1482] Fix - Handling remove add for watching channels

leandroBorgesFerreira:
[CAS-1482](https://stream-io.atlassian.net/browse/CAS-1482)
### 🎯 Goal

Remove and add channels correctly - Now there is an implementation for when the app lists channels where the user is not a member.

### 🛠 Implementation details

Adding a factory so users can choose more easily between the desired implementation of ChatEventHandler and added `NonMemberChatEventHandler` so there's already an implmentation for that. 

### 🎨 UI Changes

Left side - Normal filtering. 
Right side - Displaying only channels where the user is not a member
<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/144480750-5b3ce670-815f-4752-adef-e8e4b0b6d5e9.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/144480611-b859b562-1d86-41d5-b767-0bdbbaf16919.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Merge this branch with `bug/trellSampleShuffling` and use trelltest to test removal and addition of channels. 

### ☑️ Checklist

- [x] Changelog is updated with client-facing changes. (it was changed in another PR)
- [x] Comparison screenshots added for visual changes. 
- [x] Reviewers added

### 🎉 GIF
giphy


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/ChatEventHandler.kt
```diff
@@ -101,6 +115,8 @@ public abstract class BaseChatEventHandler : ChatEventHandler {
     public open fun handleCidEvent(event: CidEvent, filter: FilterObject): EventHandlingResult {
         return when (event) {
             is ChannelHiddenEvent -> EventHandlingResult.Remove(event.cid)
+            is MemberRemovedEvent -> EventHandlingResult.Remove(event.cid)
```
ogkuzmin, line 118: Do you mean maybe `handleMemberRemovedEvent`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/ChatEventHandler.kt
```diff
@@ -64,6 +66,18 @@ public abstract class BaseChatEventHandler : ChatEventHandler {
...
     ): EventHandlingResult
 
+    /** Handles [MemberAddedEvent] event. It runs in background. */
+    public abstract fun handleMemberAddedEvent(
+        event: MemberAddedEvent,
+        filter: FilterObject,
+    ): EventHandlingResult
+
+    /** Handles [MemberRemovedEvent] event. It runs in background. */
+    public abstract fun handleMemberRemovedEvent(
```
ogkuzmin, line 76: We can't make it abstract. Because we break current implementation of our customers. We can provide default implementation with `SKIP` result

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandler.kt
```diff
@@ -41,14 +49,17 @@ internal class DefaultChatEventHandler(private val client: ChatClient, private v
...
     override fun handleNotificationAddedToChannelEvent(
         event: NotificationAddedToChannelEvent,
         filter: FilterObject,
-    ): EventHandlingResult = handleMemberUpdate(event, event.cid, filter)
+    ): EventHandlingResult = handleMemberUpdate(event.channel, event.cid, filter)
+
+    override fun handleMemberAddedEvent(
+        event: MemberAddedEvent,
+        filter: FilterObject,
+    ): EventHandlingResult = handleMemberUpdate(event.cid, filter)
```
adasiewiczr, line 62: I think we shouldn't do additional requests here. Actually, we should get rid of `query channels` requests from this class and replace them with default behavior which will add or remove channels based on membership

leandroBorgesFerreira: The problem is that we don’t know if an Add really means Add or Remove in the query, it depends on the filter. Example: 
- I’m filtering channel that I’m **not** a member.
- I receive an event NotificationAddedToChannel
- That channel will fail in my filter now, because now I’m a member.
- As a result that channel has to be removed from the list, although the event is NotificationAddedToChannel.

So, we have: 
 Added to channel -> Filter Passes -> Added to the list
 Added from channel -> Filter Fails -> Removed from the list
 Removed from the channel -> Filter Passes -> Added to the list
 Removed from the channel -> Filter Fails -> Removed to the list
 
 If we don’t use the filter, there’s no way to know which is the correct behaviour. This class would have the wrong logic without it =|.

adasiewiczr: Unfortunately, we don't have a good way to solve it right now and that's why we decided to leave this decision to integrators. They are providing the filter, so they should know what to do in such a case. It's not the best possible solution, but an approach with querying channels is not safe.
For now, we should have a solid default implementation with good docs and we'll revisit this case once we have a better solution

leandroBorgesFerreira: as we discussed, I added a default implementation with no request to channels. And there's a factory to use different implementations of the event handler and an ChannelEventHandler to to list channels that we user is not a member. (ChannelEventHandler)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
```diff
@@ -0,0 +1,139 @@
...
+                filter = Filters.and(
+                    filter,
+                    Filters.eq("cid", cid)
+                ),
+                offset = 0,
+                limit = 1,
+                messageLimit = 0,
+                memberLimit = 0,
+            )
+        ).await()
```
ogkuzmin, line 53: I see doubling here. What if we could extract it to some object or high level function?

```kotlin
internal object ChannelFilterRequest {
    suspend operator fun invoke(client: ChatClient, cid: String, filter: FilterObject): Result<List<Channel>> =
        client.queryChannelsInternal(
            QueryChannelsRequest(
                filter = Filters.and(
                    filter,
                    Filters.eq("cid", cid)
                ),
                offset = 0,
                limit = 1,
                messageLimit = 0,
                memberLimit = 0,
            )
        ).await()
}

```

Then we can reuse this object in both event handlers

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
```diff
@@ -0,0 +1,139 @@
...
+import io.getstream.chat.android.client.utils.map
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Default implementation of [ChatEventHandler] which is more generic than [MessagingChatEventHandler]. It skips updates
+ * and makes an API request if a channel wasn't yet handled before when receives [NotificationAddedToChannelEvent],
+ * [NotificationMessageNewEvent], [NotificationRemovedFromChannelEvent].
+ */
+public class NonMemberChatEventHandler(
```
ogkuzmin, line 26: Why do you call it `NoMember...` if for member events it makes request?

leandroBorgesFerreira: I updated the documentation now. This is the ChatEventHanlder to be used when filter is asking for channels which the current user is not a member.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
```diff
@@ -0,0 +1,139 @@
...
+import io.getstream.chat.android.client.utils.Result
+import io.getstream.chat.android.client.utils.map
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Default implementation of [ChatEventHandler] which is more generic than [MessagingChatEventHandler]. It skips updates
+ * and makes an API request if a channel wasn't yet handled before when receives [NotificationAddedToChannelEvent],
+ * [NotificationMessageNewEvent], [NotificationRemovedFromChannelEvent].
+ */
```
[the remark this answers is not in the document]
leandroBorgesFerreira: Updated the documentation, this was incorrect

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/ChannelFilterRequest.kt
```diff
@@ -0,0 +1,25 @@
+package io.getstream.chat.android.offline.querychannels
+
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.client.api.models.FilterObject
+import io.getstream.chat.android.client.api.models.QueryChannelsRequest
+import io.getstream.chat.android.client.call.await
+import io.getstream.chat.android.client.models.Channel
+import io.getstream.chat.android.client.models.Filters
+import io.getstream.chat.android.client.utils.Result
+
```
ogkuzmin, line 10: Nice to add KDOC here

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
```diff
@@ -0,0 +1,108 @@
...
+import io.getstream.chat.android.client.events.MemberRemovedEvent
+import io.getstream.chat.android.client.events.NotificationAddedToChannelEvent
+import io.getstream.chat.android.client.events.NotificationMessageNewEvent
+import io.getstream.chat.android.client.events.NotificationRemovedFromChannelEvent
+import io.getstream.chat.android.client.models.Channel
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Implementation of [ChatEventHandler] that handles events when ChannelListViewModel has a filter to show channels which
```
ogkuzmin, line 17: I would avoid here mentioning `ChannelListViewModel`. Because this module is not aware of any UI part. It's better to substitute it by `a query channels request has a filter to include channels where the current user is not a member`

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/querychannels/QueryChannelsControllerIntegrationTest.kt
```diff
@@ -160,50 +157,6 @@ internal class QueryChannelsControllerIntegrationTest : BaseConnectedMockedTest(
             sut.channels.value shouldBeEqualTo listOf(data.channel1, data.channel2, data.channel3)
         }
 
-    @Test
-    fun `Given channel in the list and default events handler When handle NotificationMessageNewEvent Should not make request to API`(): Unit =
```
ogkuzmin, line 164: I wouldn't just delete these tests. If logic gets changed we should reflect it in our test. Now, `DefaultChatEventHandler` adds a channel from this event if it's absent what about checking it? If you has questions how to write such tests I can assist you

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Handles [MemberRemovedEvent]. It makes a request to API to define outcome of handling.
+     *
+     * @param event Instance of [MemberRemovedEvent] that is being handled.
+     * @param filter [FilterObject] which is used to define an outcome.
+     */
+    override fun handleMemberRemovedEvent(
+        event: MemberRemovedEvent,
+        filter: FilterObject,
+    ): EventHandlingResult = handleMemberUpdate(event.cid, filter)
```
adasiewiczr, line 63: I think we don't have to do this request if we are assuming that channel should be visible event if you are not the member

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    override fun handleNotificationAddedToChannelEvent(
+        event: NotificationAddedToChannelEvent,
+        filter: FilterObject,
+    ): EventHandlingResult = EventHandlingResult.Skip
+
+    override fun handleMemberAddedEvent(
+        event: MemberAddedEvent,
+        filter: FilterObject,
+    ): EventHandlingResult = handleMemberUpdate(event.cid, filter)
```
adasiewiczr, line 33: Do we need to do a request here? Shouldn't we just add a channel to the list?

leandroBorgesFerreira: We actually just need to remove it (the logic is inverse here).

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
leandroBorgesFerreira: We need to add the channel because the user is no longer a member. If the user was previously there, it need to be removed.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
leandroBorgesFerreira: We need to make a request for Channel because we there's no Channel in the event `MemberRemovedEvent`

leandroBorgesFerreira: Our API to add channels doesn't work with cid, only the whole channel

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/querychannels/QueryChannelsControllerIntegrationTest.kt
leandroBorgesFerreira: I added unit tests in the new classes that cover the new cases added

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandlerTest.kt
```diff
@@ -0,0 +1,165 @@
...
+import io.getstream.chat.android.test.randomString
+import kotlinx.coroutines.flow.MutableStateFlow
+import org.amshove.kluent.`should be equal to`
+import org.junit.jupiter.api.Test
+import java.util.Date
+
+internal class DefaultChatEventHandlerTest {
+
+    @Test
+    fun `When received NotificationAddedToChannelEvent and the channel is not present, it should be added`() {
```
ogkuzmin, line 22: Nice tests! 👍 
One thing. We have [naming convention](https://github.com/GetStream/InternalAndroidGuidelines#testing) for tests. According to it this test should have the following name:
`Given the channel is not present When handle NotificationAddedToChannelEvent Should add this channel`

stream-chat-android-offline/src/test/java/io/getstream/chat/android/offline/querychannels/DefaultChatEventHandlerTest.kt
```diff
@@ -0,0 +1,165 @@
...
+            cid = cid,
+            channelType = randomString(),
+            channelId = randomString(),
+            channel = channel,
+            totalUnreadCount = randomInt(),
+            unreadChannels = randomInt(),
+        )
+    }
+
+    private fun randomNotificationRemovedFromChannelEvent(
```
ogkuzmin, line 133: We have `Mother` file with top-level creating functions that can be reshared across the test codebase. Would you mind to move these functions there?

ogkuzmin (commented):
Few last comments

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/querychannels/NonMemberChatEventHandler.kt
```diff
@@ -0,0 +1,93 @@
...
+import io.getstream.chat.android.client.events.MemberRemovedEvent
+import io.getstream.chat.android.client.events.NotificationAddedToChannelEvent
+import io.getstream.chat.android.client.events.NotificationMessageNewEvent
+import io.getstream.chat.android.client.events.NotificationRemovedFromChannelEvent
+import io.getstream.chat.android.client.models.Channel
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Implementation of [ChatEventHandler] that handles events when the of channels shows which
```
ogkuzmin, line 17: `when the of channels` - some misspelling here

ogkuzmin (approved):
LGTM

docusaurus/docs/Android/03-ui/04-components/02-channel-list.mdx
```diff
@@ -446,3 +446,41 @@ public sealed class EventHandlingResult {
...
+public class ChannelListViewModelFactory @JvmOverloads constructor(
+    private val filter: FilterObject? = null,
+    private val sort: QuerySort<Channel> = ChannelListViewModel.DEFAULT_SORT,
+    private val limit: Int = 30,
+    private val messageLimit: Int = 1,
+    private val chatEventHandlerFactory: ChatEventHandlerFactory = ChatEventHandlerFactory(),
+) 
+```
+
+There's a default implementation of `ChatEventHandler` for a ChannelListView listing channels that the current user is not a member. You can override the factory and inject it:
```
adasiewiczr, line 461:
```suggestion
A default `ChatEventHandlerFactory` implementation provides a `ChatEventHandler` that updates the `ChannelListView` based on membership - the channel will be added if the current user is a member and it will be removed otherwise.  You can override the factory, like in the example below:
```

docusaurus/docs/Android/03-ui/04-components/02-channel-list.mdx
```diff
@@ -446,3 +446,41 @@ public sealed class EventHandlingResult {
...
+```
+public open class NonMemberChatEventHandlerFactory : ChatEventHandlerFactory() {
+
+    public override fun chatEventHandler(channels: StateFlow<List<Channel>>): ChatEventHandler {
+        return NonMemberChatEventHandler(ChatClient.instance(), channels = channels)
+    }
+}
+```
+
+Then use the factory in you ChannelListViewModel
```
adasiewiczr, line 472:
```suggestion
Make sure to pass the factory to the `ChannelListViewModel`:
```

adasiewiczr (commented):
Just small docs improvements and looks good besides that. Let's mention in the CHANGELOG that `ChannelListViewModel`'s constructor is changed and it takes `ChatEventHandlerFactory` instead `ChatEventHandler`

CHANGELOG.md
```diff
@@ -85,7 +85,7 @@ Consider migrating to `stream-chat-android-ui-components` or `stream-chat-androi
   You can customize its appearance using `streamUiMediaActivityHeader`, `streamUiMediaActivityHeaderLeftActionButtonStyle` and `streamUiMediaActivityHeaderTitleStyle` attributes.
 
 ### ⚠️ Changed
-
+- Constructor of `ChannelListViewModel` and `ChannelListViewModelFactory` changed. Now they ask for `ChatEventHandlerFactory` instead `ChatEventHandler`, so users can use `StateFlow<List<Channel>>` in their implementations of `ChatEventHandler`, which can make implementation smarter with resources (don't try to add a channel that is already there, for example)
```
adasiewiczr, line 88:
```suggestion
- Constructor of `ChannelListViewModel` and `ChannelListViewModelFactory` changed. Now they ask for `ChatEventHandlerFactory` instead `ChatEventHandler`, so users can use `StateFlow<List<Channel>>` in their implementations of `ChatEventHandler`, which can make implementation smarter with resources (don't try to add a channel that is already there, for example) [#2747](https://github.com/GetStream/stream-chat-android/pull/2747)
```

adasiewiczr (approved):
- [] Bugs validated (bugfixes)

Issue #2742, opened 2021-12-06
[Compose] - Restructure components in the project


filbabic:
### Context

We're looking to improve our navigation and structure of the project, to make it easier to find components and link them to the corresponding docs.

We proposed the following structure:

- `common` -> `components`: we'd rename the current common package and have it hold all smaller building block components. We'd keep everything that's not of the highest level here. **Each component would have its own file, for the sake of discoverability.**
- `messages`, `channels` -> These main feature packaged should the Screen components in the root level and smaller packages for the feature and respective high level bound/stateless components. 
- high-level bound/stateless component packages (e.g. `attachments`, `list`) - these packages should hold components that match the given feature. E.g. AttachmentsPicker, MessageComposer, MessageList etc. Most of these components should only have one file instead of one file for each component piece, given that these are mostly self-sustained components.

### Acceptance criteria

To accept this task we need to complete the following items:

- [x] Rename the common package to components.
- [x] Move components which make sense to be reused/building blocks to the components package from other files
- [x] Split up components into their respective files for discoverability

PR #2752, opened 2021-12-08
2742 - Restructure Compose components and packaging

filbabic:
Work on #2742 

### 🎯 Goal
An absolutely mad amount of restructuring and renaming. A few smaller UI/logic improvements and warning cleanup.

### ☑️ Checklist

~~- [ ] Affected documentation updated (docusaurus, tutorial, CMS (task created))~~
- [x] Reviewers added


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/UserAvatar.kt
```diff
@@ -110,23 +110,23 @@ public enum class OnlineIndicatorAlignment(public val alignment: Alignment) {
 @Preview
 @Composable
 private fun UserWithImageAvatarPreview() {
-    buildUserAvatarPreview(PreviewUserData.userWithImage)
```
MarinTolic, line 113: I think we should extract `OnlineIndicator(modifier: Modifier = Modifier)` **(line 88)** into a separate file as it's used both in UserAvatar and ChannelAvatar and is a publicly exposed function. 

This would run more inline with the new organization.

filbabic: Yeah definitely - I didn't dive into the components already in the components package, should've ran through them

filbabic:
@MarinTolic yeah that's the general idea. I hope I managed to wrap this up for all components and files we have,

The previews could come at the end, the position of them doesn't matter much.

MarinTolic (approved):

PR #2758, opened 2021-12-09
Preprocess ownReactions for currentUser when event arrives

ogkuzmin:
### 🎯 Goal

Update `Message::ownReactions` for current user when WS event arrives

### 🛠 Implementation details

Set field as merge of lates reactions for this user and own cached reactions

### 🧪 Testing

UI Comp Sample App:

1. Enter channel X
2. Open message overlay and leave a reaction (everything is fine)
3. After that open the message overlay and react using a different reaction without removing the previous one. 
4. Reactions should be shown accordingly

### ☑️ Checklist

- [x] Reviewers added

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
@@ -54,6 +54,7 @@ import io.getstream.chat.android.client.models.ChannelUserRead
 import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.models.User
 import io.getstream.chat.android.offline.ChatDomainImpl
+import io.getstream.chat.android.offline.extensions.mergeReactions
 import io.getstream.chat.android.offline.extensions.setMember
 import io.getstream.chat.android.offline.extensions.updateReads
 import kotlinx.coroutines.launch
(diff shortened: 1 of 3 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
```diff
@@ -452,10 +453,13 @@ internal class EventHandlerImpl(
     }
 
     private fun Message.enrichWithOwnReactions(batch: EventBatchUpdate, user: User?) {
-        if (user != null && domainImpl.user.value?.id != user.id) {
-            ownReactions = batch.getCurrentMessage(id)?.ownReactions ?: mutableListOf()
+        ownReactions = if (user != null && domainImpl.user.value?.id != user.id) {
```
adasiewiczr, line 456: 🤔 Shouldn't we always but empty list in that case?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/extensions/Reaction.kt
```diff
@@ -4,14 +4,33 @@ import io.getstream.chat.android.client.models.Reaction
...
+ *
+ * @return Collection of reactions where cached data is substituted by more recent one if they have same [Reaction.type].
+ */
+internal fun mergeReactions(
+    recentReactions: Collection<Reaction>,
+    cachedReactions: Collection<Reaction>,
+): Collection<Reaction> {
+    return (
+        cachedReactions.associateBy(Reaction::type) +
+            recentReactions.associateBy(Reaction::type)
```
adasiewiczr, line 34: What happens if the cached reaction has the same type as the recent reaction but different `syncStatus` or `score`? I think we should get more recent one

ogkuzmin: I have test for that. They should be substituted by more recent

adasiewiczr (approved):

PR #2770, opened 2021-12-14
2713 - Added logic to enable/disable date separators and system messa…

### LINKED ISSUE #2713: [Compose] - Fix the message grouping logic
### Context

We have a bug in the message grouping logic where the date separators _aren't_ taken into account as previous messages and such don't affect the way we add positions to our items.

In a real-world use case, separators provide a visually separating item, which means we _should_ take them into account in our grouping logic.

image

### Acceptance criteria

- [x] Improve the algorithm when checking previous messages to take date separators into account.
- [x] Also include logic around system messages.
- [x] Allow the user to enable/disable separators and system messages.

filbabic:
Fixes #2713 

### 🎯 Goal

Our grouping logic wasn't taking into account if we'll be adding date separators before the next item. 

This caused weird UI as messages were marked off in a "middle" position, even though there's a visual (date) separator.

In this fix, we've improved the logic to take into account the next message, as well as added logic to disable System and Date separator items in the list, if the user wants.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| before | after |

We also added a way to disable date separator/system message items.

disabled separators

### 🧪 Testing

By opening any chat that has a date separator in the middle of one person's message group, these separators should now split up the group visually.

#### General
- [ ] Thread with the PR link started in a respective Slack channel (#android-team or #compose-chat-sdk-team) (required)

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

group


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -274,15 +278,20 @@ public class MessageListViewModel(
...
      * @param messages List of all messages.
      * @return Filtered messages.
      */
-    private fun filterDeletedMessages(messages: List<Message>): List<Message> {
+    private fun filterMessagesToShow(messages: List<Message>): List<Message> {
         val currentUser = user.value
 
-        return messages.filter { !(it.user.id != currentUser?.id && it.deletedAt != null) }
+        return messages.filter {
+            val isNotDeletedByOtherUser = !(it.user.id != currentUser?.id && it.deletedAt != null)
```
MarinTolic, line 290: Possibly just a small optimization, but would it not be prudent to write this condition as:

```kotlin
!(it.deletedAt != null && it.user.id != currentUser?.id &&)
```
The reasoning behind this is as follows:

Fewer messages are deleted than are owned by others.  Logical operator `&&` will make sure that whatever is to the right of it does not get evaluated if whatever is to the left of it is `false` since it already knows the condition cannot be satisfied.

This would save unnecessary evaluation.

filbabic: That's actually really good - if the message is not deleted, we save 1 comparison. 

Nicely done! :]

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -589,13 +598,17 @@ public class MessageListViewModel(
...
             val user = message.user
             val previousMessage = messages.getOrNull(index - 1)
+            val nextMessage = messages.getOrNull(index + 1)
 
             val previousUser = previousMessage?.user
-            val nextUser = messages.getOrNull(index + 1)?.user
+            val nextUser = nextMessage?.user
+
+            val willSeparateNextMessage =
+                nextMessage?.let { shouldAddDateSeparator(message, it) } ?: false
```
MarinTolic, line 607: Not directly related to this PR but the function `shouldAddDateSeparator()` uses a hardcoded time unit and value.

``kotlin
return timeDifference > TimeUnit.HOURS.toMillis(4)
```

Since we're giving users the ability to show or not show the date separators, would it also be nice to extract this time difference in a way that lets the user have more fine grained control over the date separators?

MarinTolic (approved):
### ☑️Reviewer Checklist

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
filbabic: Good point - I'll add a way to customize this logic

MarinTolic (approved):

bychkovdmitry (approved):
LGTM

Issue #2676, opened 2021-11-25
Channel list doesn't autoscroll to the top


bychkovdmitry:
### Context

If the channel list is scrolled to the top and some channel is bumped to the top, the user needs to scroll manually to the updated channel above.

### Change Proposal

Autoscroll the channel list to top. Possible solutions which come to my mind:

- Adding a "dummy" item with 0 height to the top of the channel list
- Remembering the previous scroll state and comparing it with the new one.

### Acceptance criteria

To accept this task we need to complete the following items:

- [x] The channel list scroll position stays at the top when some channel is bumped to the first position in the list.

PR #2887, opened 2022-01-05
[2676] Fix channel list autoscroll

bychkovdmitry:
https://github.com/GetStream/stream-chat-android/issues/2676

### 🎯 Goal

Out of the box, the Compose implementation of the `LazyColumn` component doesn't autoscroll the list to the top when an item is added (or moved) above the first item in the list (when using keys). This makes the experience on the channel list screen really bad:

-  We first get channels from the cache and then shuffled channels from the network. The user is forced to scroll to the top manually.
- When somebody sends a message to a channel, it is bumped to the top. But from the client's perspective it just disappears as the autoscroll to the top doesn't work out of the box.

### 🛠 Implementation details

Tried to implement autoscroll similarly to what we have on the message list screen. Unfortunately, it was even more glitchy as when a channel with a new message is bumped to the top it is first removed from the list, then added above outside of the viewport. Next, we need to autoscroll manually to the item above.

Therefore, for now, I've implemented a workaround with a dummy item. It won't work in all cases (like the list with a few pixel offset from the top), but it's still better than what we have at the moment.

### 🎨 UI Changes

https://user-images.githubusercontent.com/9600921/148212872-558c7c0d-db49-43ba-9baf-60e5d44443cf.mov

#### General

#### Code & documentation
~~- [ ] Affected documentation updated (KDocs, docusaurus, tutorial)~~

### ☑️Reviewer Checklist


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/Channels.kt
```diff
@@ -60,3 +69,19 @@ public fun Channels(
...
+ * If the list is scrolled to the top and a channel new item is added or moved
+ * to the position above, then the list will automatically autoscroll to it.
+ */
+@Composable
+private fun DummyFirstChannelItem() {
+    Box(
+        modifier = Modifier
+            .height(1.dp)
+            .fillMaxWidth()
+            .background(color = ChatTheme.colors.appBackground)
```
MarinTolic, line 85: I wouldn't put a background color on this as we can't guarantee that users won't change the background color of the `Channels` component or that their app background won't be a non `ChatTheme` color for some reason.

In general I think it can only cause headaches with no benefit.

I don't know if this was introduced to solve a specific issue though.

bychkovdmitry: Hoisted the background color to the channels screen

filbabic (approved):
Looks good to me! :]

### ☑️Reviewer Checklist

Issue #2912, opened 2022-01-06
Selected message menu/overlay allows scrolling after long tap to show it


filbabic:
### Context & proposal

When we long tap on messages in the list, the click still persists and allows scrolling/gestures to go back.

### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Consume the long tap in the message items after showing the overlay.
- [ ] Make sure that we don't allow scrolling after the overlay is shown

PR #2916, opened 2022-01-07
Add small fix to intercept scrolling when popups are shown

filbabic:
Proposed fix for #2912 

### 🎯 Goal

Our users have reported that long tapping on an item in the list both opens a dialog and allows them to scroll while holding the press. This shouldn't happen.

### 🛠 Implementation details

Added a check in our ScrollHelper that checks if we have any `DialogFragment`s active. If we do, it manually calls stop scrolling on the list to prevent further scroll, which releases the long tap/press.

### 🧪 Testing

- Open any messages screen (any channel). 
- Long tap on a message and keep holding the click.
- Attempt to scroll the list behind by holding and pulling the click up or down (while the dialog is active).

Before this change: The list kept scrolling.
After this change: The list should be static.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

scrolling

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/internal/MessageListScrollHelper.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/internal/MessageListScrollHelper.kt
@@ -1,11 +1,13 @@
 package io.getstream.chat.android.ui.message.list.internal
 
 import androidx.core.view.isVisible
+import androidx.fragment.app.DialogFragment
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import com.getstream.sdk.chat.adapter.MessageListItem
 import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.ui.common.extensions.internal.dpToPx
+import io.getstream.chat.android.ui.common.extensions.internal.getFragmentManager
 import io.getstream.chat.android.ui.common.extensions.internal.safeCast
 import io.getstream.chat.android.ui.common.extensions.isDeleted
 import io.getstream.chat.android.ui.message.list.adapter.BaseMessageItemViewHolder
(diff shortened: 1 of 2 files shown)


bychkovdmitry:
### ☑️Reviewer Checklist
~~- [ ] UI Changes correct (before & after images)~~
-~~ [ ] New feature tested and works~~
~~- [ ] All code we touched has new or updated KDocs~~

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/internal/MessageListScrollHelper.kt
```diff
@@ -42,6 +44,19 @@ internal class MessageListScrollHelper(
...
         recyclerView.addOnScrollListener(
             object : RecyclerView.OnScrollListener() {
+
+                /**
+                 * Checks if we currently have a popup shown over the list.
+                 *
+                 * @param recyclerView The list that we're observing.
+                 * @param newState The scroll state of the list.
+                 */
+                override fun onScrollStateChanged(recyclerView: RecyclerView, newState: Int) {
```
bychkovdmitry, line 54: I still don't get why long click listener returning `true` doesn't consume the click.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/internal/MessageListScrollHelper.kt
```diff
@@ -76,6 +91,20 @@ internal class MessageListScrollHelper(
...
+     * Checks if we have any popups shown over the list and stops scrolling in case we do.
+     *
+     * @param recyclerView The list that's being observed for long taps and scrolling.
+     */
+    private fun stopScrollIfPopupShown(recyclerView: RecyclerView) {
+        val fragmentManager = recyclerView.context.getFragmentManager() ?: return
+        val hasDialogsShown = fragmentManager.fragments.any { it is DialogFragment }
+
+        if (hasDialogsShown) {
+            recyclerView.stopScroll()
```
bychkovdmitry, line 104: This works pretty well. An alternative solution would be to use `LayoutManager.canScrollVertically`, but it is more verbose and requires introducing state. So, I think your solution is pretty neat.

bychkovdmitry (approved):
LGTW, works as expected 99/100 times. Nice job!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/internal/MessageListScrollHelper.kt
filbabic: Yeah I'm not 100% sure about that either. We can play around with more research, maybe we can find a better solution in the future

PR #2950, opened 2022-01-18
Intercepted the default click listener to disable opening uploading files

### LINKED ISSUE #2875: Error when trying to open a file that is being uploaded
### Context & proposal
Error is being shown because `url` is not available. We should disable clicks on files that are not uploaded yet

filbabic:
Fixes #2875 

### 🎯 Goal

We can't open uploading files as the URL doesn't exist for them yet.

### 🛠 Implementation details

Checked if the currently selected message has _any_ uploading attachments and if so, we disable/intercept the click on them.

### 🧪 Testing

Try uploading an attachment and clicking on the uploading file. It shouldn't open the preview.

#### General

#### Code & documentation

### ☑️Reviewer Checklist


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
```diff
@@ -358,6 +363,9 @@ public class MessageListView : ConstraintLayout {
 
     private val DEFAULT_ATTACHMENT_CLICK_LISTENER =
         AttachmentClickListener { message, attachment ->
+            val hasInvalidAttachments = message.attachments.any { it.upload != null || (it.uploadState != null && it.uploadState != Attachment.UploadState.Success) }
```
filbabic, line 366: There are a few things to do here:

A) Do what this implementation does - check that everything in a message has been successfully uploaded.
B) check that only the clicked attachment is uploaded (then we can open the screen, but just for this attachment)
C) Check if we can simplify this check by just making sure `uploadState` is not null (just thought of this)
D) Try to intercept this click listener somewhere in the Upload UI, but this seems like it'll be buried in the implementation.

filbabic: I'll add corresponding docs once we agree on the given behavior.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
```diff
@@ -358,6 +363,11 @@ public class MessageListView : ConstraintLayout {
 
     private val DEFAULT_ATTACHMENT_CLICK_LISTENER =
         AttachmentClickListener { message, attachment ->
+            val hasInvalidAttachments = message.attachments.any { it.uploadState != null }
```
filbabic, line 366: This feels much better to me.

But we still have the issue of: do we block opening until _all_ attachments are uploaded vs blocking only the uploaded attachments. 

This could mess up the preview too if there are images, so I propose we just don't allow users to open anything if all the attachments aren't uploaded.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
MarinTolic: To me, it seems that we don't support partial uploads as even when `uploadState == Attachment.UploadState.Success` evaluates to true, we can't open the file because `attachment.url` is null. Same applies to `attachments.ogUrl`. Thanks for pointing out in a private chat that we need to support open graph backed content as well.

It seems like we could achieve the desired effect with

```kotlin
 if(attachment.url == null && attachment.ogUrl) return@AttachmentClickListener
 ```

This would allow us to iterate per attachment and not open any attachments without valid URLs. If we ever implemented support for partially uploaded messages the previews for ones with valid URLs would automatically work.

Comments on this from the core team would be great as they are stronger in this domain.

MarinTolic: Other than that, both the more complex logical evaluation and the simpler one `it.uploadState != null` seem to be equivalent in their effect.

adasiewiczr: Not really. You can always create a custom attachment and any of its fields during creation. Moreover, I think we are using more than `url` and `ogUrl` for displaying different attachment types.
`message.attachments.any { it.uploadState != null }` should work fine in our use case but we should consider another approach in the future.

MarinTolic (approved):
### ☑️Reviewer Checklist
- [ ] ~~UI Changes correct (before & after images)~~

PR #2955, opened 2022-01-20
[2911] Fix attachment support

MarinTolic:
### 🎯 Goal

Due to the different way that different platforms handle mime types, we cannot rely on simply expecting to have the correct mime type and judging the attachment format by it.

For that reason we have to do extra work when handling attachment mime types in order to be flexible and guarantee compatibility with other SDKs.

### 🛠 Implementation details

Implement a fallback option algorithm that will first look for a valid media mime type and open an audio/ video using the correct `MediaAttachmentPreviewHandler.`. If the file does not have a full and valid mime type in `attachment.MimeType`, try to look for information about the file type in `attachment.type`. 

Sometimes platforms send no type and an incomplete mime type containing only the subtype, so if both previous options fail, try to open the video based only on subtype.

### 🎨 UI Changes

**NOTE:** Screenshots are from the AttachmentsComposeSample app

| Before | After |
| --- | --- |
| before | after |

### 🧪 Testing

Easiest way to test this is by switching to the tutorial environment as the `leandro` user and opening the "attachment test" channel. From there, you can send a bunch of different files and file types and make sure that these are visible and/or downloadable/clickable (based on the file type) in the message list.

To switch to the environment, you can use the Patch that's visible when expanding the section below.

<details>
<summary>Patch file to apply these changes for testing</summary>

```
Index: stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatApp.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatApp.kt b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatApp.kt
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatApp.kt	(revision 8cea64466323cd45a1a388bef2f8586f7cbccf19)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatApp.kt	(date 1643102488327)
@@ -41,7 +41,7 @@
     private fun setupStreamSdk() {
         val offlinePlugin = OfflinePlugin(Config(userPresence = true, persistenceEnabled = true))
 
-        ChatClient.Builder("qx5us2v6xvmh", applicationContext)
+        ChatClient.Builder("9vdff5qum64q", applicationContext)
             .logLevel(ChatLogLevel.ALL)
             .notifications(createNotificationConfig(), createNotificationHandler())
             .withPlugin(offlinePlugin)
@@ -51,13 +51,13 @@
     private fun connectUser() {
         ChatClient.instance().connectUser(
             user = User(
-                id = "filip",
+                id = "leandro",
                 extraData = mutableMapOf(
-                    "name" to "Filip Babić",
-                    "image" to "https://ca.slack-edge.com/T02RM6X6B-U022AFX9D2S-f7bcb3d56180-128",
+                    "name" to "Tutorial Droid",
+                    "image" to "https://bit.ly/2TIt8NR",
                 ),
             ),
-            token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoiZmlsaXAifQ.WKqTjU6fHHjtFej-sUqS2ml3Rvdqn4Ptrf7jfKqzFgU"
+            token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoibGVhbmRybyJ9.9zN8Z9k9S2mLc-M5c8OSwh4Qw3jk7LlYXOty3icrXto"
         ).enqueue()
     }
 
```
</details>

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

files


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/factory/FileAttachmentFactory.kt
```diff
@@ -15,7 +16,11 @@ import io.getstream.chat.android.compose.ui.theme.ChatTheme
  */
 @Suppress("FunctionName")
 public fun FileAttachmentFactory(): AttachmentFactory = AttachmentFactory(
-    canHandle = { attachments -> attachments.any { it.uploadId != null || it.upload != null } },
+    canHandle = { attachments ->
+        attachments.any {
+            it.uploadId != null || it.upload != null || it.type == ModelType.attach_file || it.type == ModelType.attach_video || it.type == ModelType.attach_video
```
MarinTolic, line 21: The formatting on this is supper ugly and I am open to suggestions as personally I cannot find a way to make it prettier.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/filepreview/MediaAttachmentPreviewHandler.kt
```diff
@@ -42,10 +40,30 @@ public class MediaAttachmentPreviewHandler(private val context: Context) : Attac
         )
     }
 
-    private companion object {
-        private const val MIME_TYPE_VIDEO = "video"
-        private const val MIME_TYPE_AUDIO = "audio"
-        private const val MIME_SUBTYPE_MP4 = "mp4"
-        private const val MIME_SUBTYPE_QUICKTIME = "quicktime"
-    }
+    private fun buildSubTypeList() = listOf(
```
MarinTolic, line 43: This is a fallback option and as such is not guaranteed to cover all formats in all variations as the list of such would be prohibitively long. 

We might partially mitigate this if we better structure mime types inside `ModelType` as previously mentioned in our discussions.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/filepreview/MediaAttachmentPreviewHandler.kt
```diff
@@ -16,20 +16,18 @@ public class MediaAttachmentPreviewHandler(private val context: Context) : Attac
         val type = attachment.type
 
         if (assetUrl.isNullOrEmpty()) return false
-        if (mimeType.isNullOrEmpty()) return false
+
+        // Check if the base of Attachment.mimeType is audio or video
+        if (!mimeType.isNullOrEmpty() && (mimeType.contains(ModelType.attach_audio) || mimeType.contains(ModelType.attach_video))) return true
```
MarinTolic, line 21: This in certain cases causes false positives such as in the case of `.avi` files which can be opened by us but not played by ExoPlayer. 

This should still be better than causing false negatives by unwittingly restricting mime types.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/factory/FileAttachmentFactory.kt
MarinTolic: Also the reason why we do this is because at least in one case we have received a video attachment containing no mime type and a file type.

filbabic: Why do we check that `it.type == ModelType.attach_video` twice?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/filepreview/MediaAttachmentPreviewHandler.kt
```diff
@@ -16,20 +16,18 @@ public class MediaAttachmentPreviewHandler(private val context: Context) : Attac
...
-            MIME_SUBTYPE_MP4,
-            MIME_SUBTYPE_QUICKTIME
-        )
+        // Fallback in case we receive an incomplete mime type and both previous checks fail
+        val supportedMimeSubTypes = buildSubTypeList()
 
-        return supportedMimeTypes.any { mimeType.contains(it) } ||
-            type == ModelType.attach_audio ||
-            type == ModelType.attach_video
+        return supportedMimeSubTypes.any { mimeType?.contains(it) ?: false }
```
filbabic, line 30: Maybe do something like:

```
return if (mimeType == null) {
  false
} else {
  val supportedMimeSubTypes = buildSubTypeList()
  supportedMimeSubTypes.any { subtype -> subtype in mimeType }
}
```

This way we don't necessarily go through the entire list of subtypes to check with a null mime type that all values are null and false.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/filepreview/MediaAttachmentPreviewHandler.kt
filbabic, line 43: docs missing

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MimeTypeIconProvider.kt
```diff
@@ -3,6 +3,9 @@ package io.getstream.chat.android.compose.ui.util
 import com.getstream.sdk.chat.model.ModelType
 import io.getstream.chat.android.compose.R
 
+/**
+ * Provides icons for files attachments.
```
filbabic, line 7: `file attachments`

filbabic (commented):
Looks good to me - I've left a couple of comments so please check them out!


### ☑️Reviewer Checklist

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/factory/FileAttachmentFactory.kt
MarinTolic: That's a mistake that, well spotted.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/filepreview/MediaAttachmentPreviewHandler.kt
MarinTolic: Good suggestion, I've implemented it.

filbabic (approved):

Issue #2697, opened 2021-11-26
Leak detect by strict mode


spvince:
**Describe the bug**
Strict mode detect a leak

**SDK version**
- 4.22.0

**Leak**
```
D/StrictMode: StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
        at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1924)
        at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:303)
        at android.database.sqlite.SQLiteDatabase.dispose(SQLiteDatabase.java:344)
        at android.database.sqlite.SQLiteDatabase.finalize(SQLiteDatabase.java:328)
        at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:291)
        at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:278)
        at java.lang.Daemons$Daemon.run(Daemons.java:139)
        at java.lang.Thread.run(Thread.java:923)
     Caused by: java.lang.Throwable: Explicit termination method 'close' not called
        at dalvik.system.CloseGuard.openWithCallSite(CloseGuard.java:259)
        at dalvik.system.CloseGuard.open(CloseGuard.java:230)
        at android.database.sqlite.SQLiteDatabase.openInner(SQLiteDatabase.java:919)
        at android.database.sqlite.SQLiteDatabase.open(SQLiteDatabase.java:898)
        at android.database.sqlite.SQLiteDatabase.openDatabase(SQLiteDatabase.java:762)
        at android.database.sqlite.SQLiteDatabase.createInMemory(SQLiteDatabase.java:958)
        at android.database.sqlite.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:368)
        at android.database.sqlite.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:316)
        at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper$OpenHelper.getWritableSupportDatabase(FrameworkSQLiteOpenHelper.java:145)
        at androidx.sqlite.db.framework.FrameworkSQLiteOpenHelper.getWritableDatabase(FrameworkSQLiteOpenHelper.java:106)
[...]

PR #2974, opened 2022-01-24
[Issue 2697] Leak in database

leandroBorgesFerreira:
issue: [2697](https://github.com/GetStream/stream-chat-android/issues/2697) 
### 🎯 Goal

Fix memory leak in Room in the SDK

### 🛠 Implementation details

Creating `RepositoryFacade` lazily, avoiding initiating it too soon and creating an in-memory database because the user is not set yet. 

Before we were creating the database 2 times (1 for in-memory and 1 for the normal one). I believe that Room doesn't close the first database in this situation, which causes the memory leak. 

### 🧪 Testing

Adding this to the Application class of the sample app:
```
import android.app.Application
 import android.os.Build
+import android.os.StrictMode
 import coil.Coil
 import coil.ImageLoader
 import coil.decode.GifDecoder
@@ -20,6 +21,7 @@ class App : Application() {

     override fun onCreate() {
         super.onCreate()
+        strictMode()
         chatInitializer.init(getApiKey())
         instance = this
         DebugMetricsHelper.init()
@@ -36,6 +38,21 @@ class App : Application() {
         initializeToggleService()
     }

+    private fun strictMode() {
+        StrictMode.setThreadPolicy(StrictMode.ThreadPolicy.Builder()
+            .detectDiskReads()
+            .detectDiskWrites()
+            .detectNetwork() // or .detectAll() for all detectable problems
+            .penaltyLog()
+            .build())
+        StrictMode.setVmPolicy(StrictMode.VmPolicy.Builder()
+            .detectLeakedSqlLiteObjects()
+            .detectLeakedClosableObjects()
+            .penaltyLog()
+            .penaltyDeath()
+            .build())
+    }
+
     private fun getApiKey(): String {
         val user = userRepository.getUser()
         return if (user != SampleUser.None) {
```

And check that it crashes in develop branch, but not in this one.
#### General
- [ ] Thread with the PR link started in a respective Slack channel (#android-team or #compose-chat-sdk-team) (required)

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (2)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -182,7 +182,13 @@ internal class ChatDomainImpl internal constructor(
 
     @VisibleForTesting
     val defaultConfig: Config = Config(connectEventsEnabled = true, muteEnabled = true)
-    internal var repos: RepositoryFacade = createNoOpRepos()
+    internal var repos: RepositoryFacade
```
filbabic, line 185: Could we add some docs here to make this change more visible for the future.

Otherwise - this makes absolute sense to me, getting an instance lazily is usually the way to go!

filbabic (approved):
Looks good - this also applies to Compose, so it's a nice 2 for 1! :]

### ☑️Reviewer Checklist

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
JcMinarro: Why don't we create a "real NoOp"  instance?
The problem here is that we are calling "NoOp" to an implementation that use the inMemory DB.
We can create a `RepositoryFacadeNoOp` class that does nothing, then we don't need to create the InMemoryDB.

leandroBorgesFerreira: @JcMinarro That would be super complex. We would need to create a NoOp version of all our DAO, also `ChatDatabase` extends `RoomDatabase` which should be created using `databaseBuilder` or `inMemoryDatabaseBuilder`. Take a look at its documentation for `RoomDatabase`:

> Creates a RoomDatabase.
You cannot create an instance of a database, instead, you should acquire it via Room.databaseBuilder(Context, Class, String) or Room.inMemoryDatabaseBuilder(Context, Class).

One thing that we could do is not create an in-memory database if the user is not set and simply return an error instead to avoid that customers initialize database twice without intention.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -182,7 +182,13 @@ internal class ChatDomainImpl internal constructor(
...
+        get() = _repos ?: createNoOpRepos()
```
skydoves, line 186: This getter seems to create `createNoOpRepos` whenever we access the `repos` property.
Don't we need to initialize the `_repos` backing property?

```kotlin
get() {
    if (_repos == null) {
        _repos = createNoOpRepos()
    }
    return _repos!!
}
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
JcMinarro: Ok, I don't want to block this PR more because of it.
We can iterate over it later.
BTW, Could you add some docs about why it is done that way to be sure it is not refactored (and reintroducing the bug again) in the future?

leandroBorgesFerreira: Done

adasiewiczr (approved):
### ☑️Reviewer Checklist

PR #2979, opened 2022-01-25
[ISSUE-2697] Disposing Coil requests

leandroBorgesFerreira:
extension of issue: [2697](https://github.com/GetStream/stream-chat-android/issues/2697) 

### 🎯 Goal

Disposing Coil requests. If Coil requests are kept in the SDK, a reference to the views will be kept, then the garbage collection won't be able to collect our Views, Activitie and Fragments. 

### 🛠 Implementation details

Adding a Disposable class that our Views call before that are killed. 

### 🧪 Testing

Very hard to test. Check for memory leaks using strict mode. 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (3)


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/disposable/DisposableList.kt
```diff
@@ -0,0 +1,18 @@
+package com.getstream.sdk.chat.disposable
+
+public class DisposableList {
+
+    private val disposables: MutableList<Disposable> = mutableListOf()
```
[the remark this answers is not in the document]
leandroBorgesFerreira: The problem is that Disposable is an interface, so it is hard to know how the hash of the Set will be generated. Users may create an undesired collision it two different objects end up with the same hash. That would be super hard to debug, if happens

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/disposable/DisposableExtension.kt
```diff
@@ -0,0 +1,10 @@
+package com.getstream.sdk.chat.disposable
+
+/**
+ * Helper function to add a new disposable to disposableList
+ *
+ * @param disposableList [DisposableList]
+ */
+public fun Disposable.addToDisposableList(disposableList: DisposableList) {
```
JcMinarro, line 8: Do we really need this extension?
And do we need to expose it to our customers?

It is a bit weird that a variable that is received as a parameter is the one that is updated.

I have seen it is used to have some sugar API when a method return a disposable that needs to be added.
```
val disposableList = DisposableList()
createDisposable().addToDisposableList(disposableList)
```

It can be switch by:
```
val disposableList = DisposableList()
disposableList.add(createDisposable())
```
Or if you would need to concatenate new actions, you could use:
```
val disposableList = DisposableList()
createDisposable().also(disposableList::add)
```

leandroBorgesFerreira:
@MarinTolic  Showed me that Coil already stops calls respecting lifecycler. We can add `Disposable` for cases out of lifecycle context, but there's no need to manually call dispose like this PR does. 

Source: https://coil-kt.github.io/coil/getting_started/#cancelling-requests

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/gallery/overview/internal/MediaAttachmentAdapter.kt
```diff
@@ -48,7 +48,7 @@ internal class MediaAttachmentAdapter(
             )
 
             val user = attachmentGalleryItem.user
-            if (user != null && showUserAvatars) {
+            if (showUserAvatars) {
```
leandroBorgesFerreira, line 51: Just a suggestion from AS, it will never be null anyways.

filbabic:
Hey!

Just wanted to check if this is still viable or if we need to close down the PR? LMK if this works and can be integrated in our SDK or we should scrap it!

leandroBorgesFerreira:
This should be merged. Users using our SDK may want to cancel Coil requests if the request is not related to the android lifecycler.

filbabic:
Approved - as soon as the comments from @skydoves are resolved/approved and you resolve the conflicts we can merge!

PR #3000, opened 2022-01-31
Merge fork of branch `GiphyResizeAux`

MarinTolic:
Continued working on resizeable Giphys.

Added:

- Styled attributes for `GiphyMediaAttachmentView`
- Additional capabilities such as constant size and giphy scaling logic
- Kdocs

Note:

Setting the following items in `StreamUi.MessageList.GiphyMediaAttachment` will result in the users seeing the UI as it was previously and visually avodiing introducing a breaking change. 

```xml
<item name="streamUiGiphyMediaAttachmentConstantSize">true</item>
<item name="streamUiGiphyMediaAttachmentGiphyMaxHeight">250dp</item>
<item name="streamUiGiphyMediaAttachmentScaleType">centerCrop</item>
```

The new system with autosizeable Giphy containers is much better however I believe this should be communicated to the users in case they want to keep the old system for some reason .

You can also keep the old square dimensions but have non cropped Giphys by changing to `fitCenter` scaling method such as this.

```xml
<item name="streamUiGiphyMediaAttachmentScaleType">fitCenter</item>
```

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/GiphyMediaAttachmentViewStyle.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/GiphyMediaAttachmentViewStyle.kt
@@ -1,0 +1,83 @@
+package io.getstream.chat.android.ui.message.list.adapter.view
+
+import android.content.Context
+import android.graphics.drawable.Drawable
+import android.util.AttributeSet
+import android.widget.ImageView
+import androidx.annotation.ColorInt
+import androidx.annotation.DimenRes
+import io.getstream.chat.android.ui.R
+import io.getstream.chat.android.ui.common.extensions.internal.dpToPx
+import io.getstream.chat.android.ui.common.extensions.internal.getColorCompat
+import io.getstream.chat.android.ui.common.extensions.internal.getDrawableCompat
+import io.getstream.chat.android.ui.common.extensions.internal.getEnum
+import io.getstream.chat.android.ui.common.extensions.internal.use
+import io.getstream.chat.android.ui.message.list.adapter.view.internal.GiphyMediaAttachmentView
+
+/**
+ * Sets the style for [io.getstream.chat.android.ui.message.list.adapter.view.internal.GiphyMediaAttachmentView].
+ */
+public class GiphyMediaAttachmentViewStyle(
+    public val progressIcon: Drawable,
+    public val giphyIcon: Drawable,
+    public val placeholderIcon: Drawable,
+    @ColorInt public val imageBackgroundColor: Int,
+    public val giphyConstantSizeEnabled: Boolean,
+    @DimenRes public val giphyHeight: Int,
+    public val giphyType: GiphyMediaAttachmentView.Companion.GiphyInfoType,
+    public val scaleType: ImageView.ScaleType
+) {
+    internal companion object {
+        private const val DEFAULT_HEIGHT_DP = 200
+
+        operator fun invoke(context: Context, attrs: AttributeSet?): GiphyMediaAttachmentViewStyle {
+            context.obtainStyledAttributes(
+                attrs,
+                R.styleable.GiphyMediaAttachmentView,
+                R.attr.streamUiMessageListMediaAttachmentStyle,
+                R.style.StreamUi_MessageList_GiphyMediaAttachment
+            ).use { attributes ->
+                val progressIcon =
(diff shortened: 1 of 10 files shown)


stream-chat-android-ui-components/src/main/res/values/styles.xml
```diff
@@ -498,6 +498,17 @@
         <item name="streamUiMediaAttachmentMoreCountTextStyle">normal</item>
     </style>
 
+    <style name="StreamUi.MessageList.GiphyMediaAttachment" parent="StreamUi">
+        <item name="streamUiGiphyMediaAttachmentProgressIcon">@drawable/stream_ui_rotating_indeterminate_progress_gradient</item>
+        <item name="streamUiGiphyMediaAttachmentGiphyIcon">@drawable/stream_ui_giphy_label</item>
+        <item name="streamUiGiphyMediaAttachmentPlaceHolderIcon">@drawable/stream_ui_picture_placeholder</item>
+        <item name="streamUiGiphyMediaAttachmentImageBackgroundColor">@color/stream_ui_literal_transparent</item>
+        <item name="streamUiGiphyMediaAttachmentConstantSize">false</item>
+        <item name="streamUiGiphyMediaAttachmentGiphyMaxHeight">200dp</item>
```
MarinTolic, line 507: We could consider providing users with a way of setting width as well. It's easy to implement however the issue is that this would become irrelevan when `streamUiGiphyMediaAttachmentConstantSize` is set to `false` so it might mislead.

Currently what happens is that if `streamUiGiphyMediaAttachmentConstantSize` is set to `true` the Giphy container is a square where `heighet == width ==  streamUiGiphyMediaAttachmentGiphyMaxHeight.`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/GiphyMediaAttachmentView.kt
```diff
@@ -40,18 +50,65 @@ internal class GiphyMediaAttachmentView : ConstraintLayout {
...
+            )
+        }
+    }
+
+    /**
+     * Displays the Giphy image inside of a resizeable container.
+     */
+    private fun showResizeableGiphy(url: String) {
+        binding.mediaAttachmentContent.updateLayoutParams {
+            this.height = style.giphyMaxHeight
```
MarinTolic, line 107: Updating layour params both on the actual `ShapeableImageView` and the container allows us to dynamically resize the container without having container sizes shifting during or after load.

The only edgecase noticed is when we have an image that is smaller than the actual container size. 

i.e. try setting this in styles:

```xml
<item name="streamUiGiphyMediaAttachmentGiphyMaxHeight">250dp</item>
<item name="streamUiGiphyMediaAttachmentGiphyType">fixedHeight</item>
```

You will see the container "jump" after loading.
Where as this:

```xml
<item name="streamUiGiphyMediaAttachmentGiphyMaxHeight">250dp</item>
<item name="streamUiGiphyMediaAttachmentGiphyType">original</item>
```

Will not result in it jumping since the loaded size is larger than the container's pre loaded size.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/GiphyMediaAttachmentViewStyle.kt
```diff
@@ -0,0 +1,100 @@
...
+        operator fun invoke(context: Context, attrs: AttributeSet?): GiphyMediaAttachmentViewStyle {
+            context.obtainStyledAttributes(
+                attrs,
+                R.styleable.GiphyMediaAttachmentView,
+                R.attr.streamUiMessageListMediaAttachmentStyle,
+                R.style.StreamUi_MessageList_GiphyMediaAttachment
+            ).use { attributes ->
+                val progressIcon =
+                    attributes.getDrawable(R.styleable.GiphyMediaAttachmentView_streamUiGiphyMediaAttachmentProgressIcon)
+                        ?: context.getDrawableCompat(R.drawable.stream_ui_rotating_indeterminate_progress_gradient)!!
```
MarinTolic, line 56: I would prefer not forcing '!!' here but this is a code style that we've widely adopted when fetching styled attributes so I did not want to break it.

stream-chat-android-ui-components/src/main/res/values/styles.xml
filbabic: width is set as the maximum width allowed for messages, in the parent. It's kind of hard to do both and expect working results in all scenarios

Issue #2897, opened 2022-01-05
Eliminate ChatDomain::deleteMessage


adasiewiczr:
### Acceptance criteria

To accept this task we need to complete the following items:

- [ ] Side effects of edit message are moved to OfflinePlugin
- [ ] Calls replaced by ChatClient::deleteMessage
- [ ] Method is deprecated

PR #3013, opened 2022-02-02
[ISSUE-2897] ChatDomain remove delete message

leandroBorgesFerreira:
issue: https://github.com/GetStream/stream-chat-android/issues/2897

### 🎯 Goal

Remove/Deprecate `ChatDomain.deleteMessage` and use `ChatClient.deleteMessage
`
### 🛠 Implementation details

Adding a new method `ChatClient.deleteMessageAndUpdateLocalData`. This method had to be create because `ChatClient.deleteMessage` uses message.id instead of the whole message, which won't work for `ChannelLogic`. One option would be to change the original method, I'm open to other solutions as well. 

### 🧪 Testing

Test that deleting a message works the same way in `develop` and this branch.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePlugin.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePlugin.kt
@@ -8,6 +8,7 @@ import io.getstream.chat.android.client.api.models.QueryChannelsRequest
 import io.getstream.chat.android.client.errors.ChatError
 import io.getstream.chat.android.client.experimental.plugin.Plugin
 import io.getstream.chat.android.client.experimental.plugin.listeners.ChannelMarkReadListener
+import io.getstream.chat.android.client.experimental.plugin.listeners.DeleteMessageListener
 import io.getstream.chat.android.client.experimental.plugin.listeners.EditMessageListener
 import io.getstream.chat.android.client.experimental.plugin.listeners.QueryChannelListener
 import io.getstream.chat.android.client.experimental.plugin.listeners.QueryChannelsListener
@@ -45,7 +46,8 @@ public class OfflinePlugin(
     QueryChannelListener,
     ThreadQueryListener,
     ChannelMarkReadListener,
-    EditMessageListener {
+    EditMessageListener,
+    DeleteMessageListener {
 
     internal constructor() : this(Config())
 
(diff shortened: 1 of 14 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePlugin.kt
```diff
@@ -213,6 +263,10 @@ public class OfflinePlugin(
         )
     }
 
+    private fun requireRepos(): RepositoryFacade {
+        return repos ?: throw IllegalStateException("RepositoryFace was set too late in OfflinePlugin")
```
MarinTolic, line 267: Should "RepositoryFace" be "RepositoryFacade"?

Also do you think that something like the following might be more clear to the users?

```suggestion
        return repos ?: throw IllegalStateException("RepositoryFacade cannot be null")
```

This is fully up to your discretion :D

leandroBorgesFerreira: I prefer to say that it was set too late, because this will happen in case this plugin was initialised too fast. But I'll fix the `RepositoryFacade`

leandroBorgesFerreira:
I added a change in this PR to be able to insert RepositoryFacade in it.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePlugin.kt
```diff
@@ -78,6 +85,49 @@ public class OfflinePlugin(
         logic = LogicRegistry(state)
     }
 
+    /**
+     * Method called when a request for message edition happens. Use it to update database, update messages in the SDK,
+     * update the UI when a message occurs...
+     *
+     * @param message [Message].
```
skydoves, line 92:
```suggestion
     * @param messageId Id of the [Message].
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePlugin.kt
```diff
@@ -69,7 +73,10 @@ public class OfflinePlugin(
             retryPolicy(config.retryPolicy.toLiveDataRetryPolicy())
         }.build()
 
-        initState(io.getstream.chat.android.offline.ChatDomain.instance as ChatDomainImpl, chatClient)
+        val chatDomainImpl = io.getstream.chat.android.offline.ChatDomain.instance as ChatDomainImpl
```
skydoves, line 76: Maybe should we use `ChatDomain.instance()`?

leandroBorgesFerreira:
Closed in favor of: #3013

PR #3018, opened 2022-02-03
Bugfix/3003 pdf preview

filbabic:
This was leftover from previous PR as the commit wasn't properly pushed due to AS issues


MarinTolic (approved):
Checklist not needed as this was a leftover commit not pushed initially by AS.

PR #3041, opened 2022-02-04
Compose 1.1.0-rc03 test PR

MarinTolic:
This is a test PR meant to highlight what changes we need to make in order to for our apps to run Compose 1.1.0-rc03 and what benefits we can get from upgrading.


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/view/internal/MediaAttachmentsGroupView.kt
```diff
@@ -185,6 +185,9 @@ internal class MediaAttachmentsGroupView : ConstraintLayout {
                 stateCopy.viewThree.setImageShapeByCorners(0f, 0f, 0f, bottomLeftCorner)
                 stateCopy.viewFour.setImageShapeByCorners(0f, 0f, bottomRightCorner, 0f)
             }
+            else -> {
+                // has to be exhaustive or ktLint will fail
```
MarinTolic, line 189: I think this is a consequence of running Kotlin 1.6.10:

https://kotlinlang.org/docs/whatsnew16.html#stable-exhaustive-when-statements-for-enum-sealed-and-boolean-subjects

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -101,39 +104,36 @@ public fun Messages(
...
-            coroutineScope.launch {
+        LaunchedEffect(key1 = messagesState) {
+            if (focusedItemIndex != -1) {
                 currentListState.scrollToItem(focusedItemIndex)
             }
-        }
 
-        when {
-            !currentListState.isScrollInProgress && newMessageState == Other && firstVisibleItemIndex < 3 -> {
-                coroutineScope.launch {
```
MarinTolic, line 112: This refactor is a consequence of mitigating the warning "Calls to launch should happen inside a LaunchedEffect and not composition" that is now displayed when running Compose 1.1.0-rc03.

The official take is that `rememberCoroutineScope` should be used only if you are launching a Coroutine outside of Composition but want that Coroutine to share the same Lifecycle as the Composable.

The refactor above function mostly the same but it has a notable flaw that we have encountered previously which is failing to kep up with the influx of new messages in a fast live stream enviroment.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -92,10 +93,11 @@ import io.getstream.chat.android.compose.ui.util.isUploading
  */
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
-public fun MessageItem(
+public fun LazyItemScope.MessageItem(
     messageItem: MessageItemState,
     onLongItemClick: (Message) -> Unit,
-    modifier: Modifier = Modifier,
+    modifier: Modifier = Modifier
+        .animateItemPlacement(),
```
MarinTolic, line 100: As a result of the way we create functions according to the `DefaultComposables` template that do not expose a modifier we have to make all of our Composables that want to be animated in a LazyList an extension function of `LazyItemScope`.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/MessagesActivity.kt
```diff
@@ -84,6 +85,7 @@ class MessagesActivity : AppCompatActivity() {
         }
     }
 
+    @OptIn(ExperimentalFoundationApi::class)
```
MarinTolic, line 88: `Modifier.animateItemPlacement()` is still experimental

buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
```diff
@@ -65,7 +65,7 @@ object Versions {
     internal const val MOCKITO_KOTLIN = "2.2.0"
     internal const val MOCKITO = "3.11.2"
     // Be careful with updating Moshi to 1.13.0 as it's shipped as MR-jar which causes problems with Jetifier. Check [this](https://github.com/square/moshi/issues/1424) issue for more details
-    internal const val MOSHI = "1.12.0"
```
MarinTolic, line 68: From what I gather this problem has been fixed with AGP 7.2.0 and upwards. However anyone using an older AGP might have issues with Jetifier.

buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
```diff
@@ -65,7 +65,7 @@ object Versions {
...
+    internal const val MOSHI = "1.13.0"
```
MarinTolic, line 68: Had to bump up Moshi because Compose 1.1.0-rc03 requires Kotlin 1.6.10 and  Moshi 1.12.0 fails to parse metadata if we are using Kotlin 1.6.10

filbabic:
Closing this one as we'll create a new PR with the stable upgrade.

PR #3051, opened 2022-02-07
[2969] Disable commands when text or attachments are present and disa…

MarinTolic:
### 🎯 Goal

Attachments should not be mixed with commands because the two are incompatible.
In order to signal that to the user we aim to remove the presence of commands suggestions and disable the commands buttom when attachments are present.

The user is also now unable to attach new content until they have dismissed the suggestions popup.

### 🛠 Implementation details

Introduced a new `SuggestionListControllerListener` to notify 'MessageInputView' when commands, mentions, and visibility changes occur.

Check for atttachments in `selectedAttachmentsCountListener` inside `MessageInputView` and enable/ disable command popups accordingly.

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| with_attachment_before | with_attachments_after |
| with_commands_before | with_commands_after |
| with_mentions_before | with_mentions_after |

### 🧪 Testing

This is a permutation based improvement so please follow the given instructions and feel free to improvise scenarios beyond them as it is easy to miss an edge case here

Scenario one:

1. Select an attachments or attachment 
2. Type in "/"

Expected result: commands popup should not be visible 

Scenario Two:

1. Type in "/"

Expected result: Commands popup should be visible, attachment button disabled

Scenario Three:

1. Type in "@"

Expected result: Mentions popup should be visible, attachment button disabled

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

experience


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -527,6 +549,27 @@ public class MessageInputView : ConstraintLayout {
...
+                isSuggestionListUiVisible = isVisible
+                binding.attachmentsButton.isEnabled = !isSuggestionListUiVisible && !inputContainsCommands
+            }
+
+            override fun containsCommands(doesContain: Boolean) {
+                inputContainsCommands = doesContain
+                binding.attachmentsButton.isEnabled = !isSuggestionListUiVisible && !inputContainsCommands
+            }
+
+            override fun containsMentions(doesContain: Boolean) {}
```
MarinTolic, line 570: Not relevant in our case but it might be handy for users so I decided to put it in the `SuggestionListControllerListener` interface.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestion/list/SuggestionListControllerListener.kt
```diff
@@ -0,0 +1,22 @@
+package io.getstream.chat.android.ui.suggestion.list
+
+/**
+ * A listener meant to be used for [SuggestionListController].
+ */
+public interface SuggestionListControllerListener {
+
+    /**
+     * Called when [SuggestionListUi] changes visibility.
```
filbabic, line 9: Could we document the SuggestionListUi too?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestion/list/SuggestionListController.kt
```diff
@@ -54,21 +105,34 @@ public class SuggestionListController(
...
     public fun isSuggestionListVisible(): Boolean {
         return suggestionListUi.isSuggestionListVisible()
     }
 
     private suspend fun renderSuggestions(suggestions: Suggestions) = withContext(DispatcherProvider.Main) {
         suggestionListUi.renderSuggestions(suggestions)
+        suggestionListControllerListener?.onSuggestionListUiVisibilityChanged(suggestionListUi.isSuggestionListVisible())
     }
 
     private fun String.isCommandMessage() = COMMAND_PATTERN.matcher(this).find()
```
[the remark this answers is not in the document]
filbabic: It kind of is and isn't.

UX-wise, client side, we cannot know that this is a command, because it shouldn't be.

The reason BE parses it is probably if it splits based on "/" and then gets the entire second part as one string. But htis can be treated as a command + another word, as a non-command, or as a two word command.

I think our regex is okay and users can trick it and push it to BE but nothing should happen I think.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -527,6 +549,27 @@ public class MessageInputView : ConstraintLayout {
         }
     }
 
+    /**
+     * Creates an instance of [SuggestionListControllerListener].
+     *
+     * Used to modify attachment button visibility to discourage mixing
```
filbabic, line 555: I would rephrase the comment here.

```
Used to disable integration buttons, based on if commands or attachments are added to the message.
```

Discouraging makes sense but at the same time implies it can be broken and ignored. We shouldn't really _encourage_ such behavior and just be straight up with what this does.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
filbabic: Sounds good!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -527,6 +549,27 @@ public class MessageInputView : ConstraintLayout {
...
+     * commands and attachments since they are not compatible.
+     */
+    private fun createSuggestionsListControllerListener(): SuggestionListControllerListener =
+        object : SuggestionListControllerListener {
+            override fun onSuggestionListUiVisibilityChanged(isVisible: Boolean) {
```
filbabic, line 560: Could we potentially wrap this in a class that's called `DefaultSuggestionListControllerListener` and then hold the values/properties `isSuggestionListUiVisible` and `inputContainsCommands` there?

It seems like we're adding two variables to the message input view just to keep them around to disable the UI, but I'd much rather receive an emission from the listener here that's a composite of the checks we have:

E.g. 
```
          override fun onSuggestionListStateChanged(shouldEnableAttachments: Boolean) {
                binding.attachmentsButton.isEnabled = shouldEnableAttachments
            }
```

Just have a single function (or a lambda param) that emits if the attachments should be enabled or disabled. This way we don't have to calculate anything ourselves and the logic is featured in one place.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -622,7 +665,17 @@ public class MessageInputView : ConstraintLayout {
 
                     refreshControlsState()
                     handleKeyStroke()
-                    messageInputDebouncer?.submitSuspendable { suggestionListController?.onNewMessageText(messageText) }
+
+                    // Debouncing when clearing the input will cause the suggestion list
```
filbabic, line 669: I would use the /**/ format of comments here if possible

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -622,7 +665,17 @@ public class MessageInputView : ConstraintLayout {
...
+                    // popup to appear briefly after clearing the input in certain cases.
+                    if (messageText == String.EMPTY) {
```
filbabic, line 671: could we just say .isEmpty() or .isEmptyOrBlank()

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestion/list/SuggestionListController.kt
```diff
@@ -12,32 +12,83 @@ import kotlinx.coroutines.withContext
...
+     * Enables or disables mention suggestions.
+     *
+     * Reactively recomputes suggestions upon setting.
+     */
     public var mentionsEnabled: Boolean by Delegates.observable(true) { _, _, _ -> computeSuggestions() }
+
+    /**
+     * Enables or disables commands suggestions.
+     *
+     * Reactively recomputes suggestions upon setting.
```
filbabic, line 57: Really nice work

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/suggestion/list/SuggestionListController.kt
filbabic, line 138: Also slight comment here

filbabic (commented):
I did leave a few comments to take a look at, but for the most part this is great! :]

Once we clean those up, I can take another look, approve and merge!

### ☑️Reviewer Checklist

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
MarinTolic: Yes, certainly.

Sometimes I do these things because of code consistency. `String.EMPTY` is used elsewhere so I simply continued the chain.

I'll change the other implementations too.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -527,6 +539,16 @@ public class MessageInputView : ConstraintLayout {
...
 
+    /**
+     * Creates an instance of [SuggestionListControllerListener].
+     *
+     * Used to disable integration buttons, based on if commands or attachments are added to the message.
+     */
+    private fun createSuggestionsListControllerListener(): DefaultSuggestionListControllerListener =
+        DefaultSuggestionListControllerListener { shouldEnable ->
+            binding.attachmentsButton.isEnabled = shouldEnable
+        }
```
filbabic, line 550:
```suggestion
        DefaultSuggestionListControllerListener { shouldEnableAttachments ->
            binding.attachmentsButton.isEnabled = shouldEnableAttachments
        }
```

filbabic (approved):

PR #3066, opened 2022-02-09
Offline plugin refactor

leandroBorgesFerreira:
### 🎯 Goal

Structural changes in `OfflinePlugin`

### 🛠 Implementation details

In the code. 

### 🧪 Testing

Test offline plugin and check that the behaviour doesn't change.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -192,8 +190,10 @@ public class ChatClient internal constructor(
             }
         }
         logger.logI("Initialised: " + getVersion())
+    }
 
-        plugins.forEach { it.init(appContext, this) }
+    internal fun addPlugins(plugins: List<Plugin>) {
+        this.plugins = plugins
```
leandroBorgesFerreira, line 196: It would be nice to have `plugins` in the constructor, but OfflinePlugin needs an initialization logic which must happen after `ChatClient` initialization. Then plugins are added by the builder after `ChatClient` is initialized.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1937,8 +1936,8 @@ public class ChatClient internal constructor(
 
         @InternalStreamChatApi
         @ExperimentalStreamChatApi
-        public fun withPlugin(plugin: Plugin): Builder = apply {
-            plugins += plugin
+        public fun withPlugin(plugin: PluginFactory): Builder = apply {
```
leandroBorgesFerreira, line 1939: Factory instead of the actual implementation, this way the initialization of the Plugin belongs to the factory, and then the plugin is not aware of its creation/initialization.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/experimental/persistence/OfflinePlugin.kt
```diff
@@ -0,0 +1,24 @@
...
+import io.getstream.chat.android.client.experimental.plugin.listeners.GetMessageListener
+import io.getstream.chat.android.client.experimental.plugin.listeners.HideChannelListener
+import io.getstream.chat.android.client.experimental.plugin.listeners.MarkAllReadListener
+import io.getstream.chat.android.client.experimental.plugin.listeners.QueryChannelListener
+import io.getstream.chat.android.client.experimental.plugin.listeners.QueryChannelsListener
+import io.getstream.chat.android.client.experimental.plugin.listeners.ThreadQueryListener
+import io.getstream.chat.android.core.ExperimentalStreamChatApi
+
+@ExperimentalStreamChatApi
+public interface OfflinePlugin :
```
leandroBorgesFerreira, line 15: Now the `OfflinePlugin`  doesn't violate the Open/Close principle anymore. If we need to change/extend the behavior of `OfflinePlugin`, we can change the implementation of one of its dependencies, but we don't need to modify the code of the `OfflinePlugin`. This way, changes happen in much more specific and smaller places, which makes changes easier to reason about. 

Dependency inversion is also respected. Now the class doesn't instantiate its own dependencies and just receive the interfaces that if needs to perform its job.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -143,10 +141,10 @@ internal class ChatDomainImpl internal constructor(
     override var userPresence: Boolean = false,
     internal var backgroundSyncEnabled: Boolean = false,
     internal var appContext: Context,
-    private val offlinePlugin: OfflinePlugin,
     internal val uploadAttachmentsNetworkType: UploadAttachmentsNetworkType = UploadAttachmentsNetworkType.NOT_ROAMING,
     override val retryPolicy: RetryPolicy = DefaultRetryPolicy(),
-) : ChatDomain {
+    private val globalState: GlobalMutableState = GlobalMutableState.getOrCreate()
```
leandroBorgesFerreira, line 146: GlobalMutableState must be a single instance, otherwise, we may end up with multiple different states if we have more than one instance of it.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -193,49 +194,6 @@ internal class ChatDomainImpl internal constructor(
 
     private var _repos: RepositoryFacade? = null
 
-    private val globalState: GlobalMutableState = offlinePlugin.globalState.toMutableState()
```
leandroBorgesFerreira, line 196: No need to manually delegate functions... let's use Kotlin's sugar syntax.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -480,8 +439,8 @@ internal class ChatDomainImpl internal constructor(
         val cid = "%s:%s".format(channelType, channelId)
         if (!activeChannelMapImpl.containsKey(cid)) {
             val channelController = ChannelController(
-                mutableState = offlinePlugin.state.channel(channelType, channelId).toMutableState(),
```
leandroBorgesFerreira, line 483: I didn't see the need for a container for `state` and `logic`. It would just add another class to SDK without much value, keeping the properties inside `ChatDomainImpl` seemed to be the best approach.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -0,0 +1,68 @@
...
+    override fun getOrCreate(): OfflinePlugin {
+        return instance ?: createOfflinePlugin().also {
+            instance = it
+        }
+    }
+
+    private fun createOfflinePlugin(): OfflinePlugin {
+        // This can only be called after ChatClient was instantiated!!
+        val chatClient = ChatClient.instance()
+        ChatDomain.Builder(appContext, chatClient).apply {
```
leandroBorgesFerreira, line 43: This is necessary while the merge is not complete.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePlugin.kt
```diff
@@ -1,288 +0,0 @@
...
-
-    /**
-     * Global state of this plugin.
-     */
-    // TODO: Move to StateRegistry when we remove ChatDomain.
-    public val globalState: GlobalState = GlobalMutableState()
-
-    override val name: String = MODULE_NAME
-
-    override fun init(appContext: Context, chatClient: ChatClient) {
```
leandroBorgesFerreira, line 80: Plugins shouldn't be statefull. Initialization can be done by its Factory.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePlugin.kt
```diff
@@ -1,288 +0,0 @@
...
-        logic.channel(channelType, channelId).onHideChannelRequest(channelType, channelId, clearHistory)
-
-    override suspend fun onHideChannelResult(
-        result: Result<Unit>,
-        channelType: String,
-        channelId: String,
-        clearHistory: Boolean,
-    ): Unit = logic.channel(channelType, channelId).onHideChannelResult(result, channelType, channelId, clearHistory)
-
-    internal fun clear() {
```
leandroBorgesFerreira, line 277: Plugins are stateless. If a dependency of a plugin contains state, then we should deal with this state in the dependency and not bring the state to the whole plugin.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/persistance/OfflineSupport.kt
```diff
@@ -0,0 +1,17 @@
...
+import io.getstream.chat.android.offline.experimental.global.GlobalState
+import io.getstream.chat.android.offline.experimental.plugin.OfflinePluginImpl
+import io.getstream.chat.android.offline.experimental.plugin.logic.LogicRegistry
+import io.getstream.chat.android.offline.experimental.plugin.state.StateRegistry
+
+/**
+ * The entry point of all offline state ([OfflinePluginImpl.state]) and behavior ([OfflinePluginImpl.logic]).
+ */
+@ExperimentalStreamChatApi
+internal class OfflineSupport(
```
adasiewiczr, line 13: Do we need this class? I think we have other ways of accessing different States/Logic

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/OfflinePluginImpl.kt
```diff
@@ -0,0 +1,52 @@
...
+    ThreadQueryListener by threadQueryListener,
+    ChannelMarkReadListener by channelMarkReadListener,
+    EditMessageListener by editMessageListener,
+    GetMessageListener by getMessageListener,
+    HideChannelListener by hideChannelListener,
+    MarkAllReadListener by markAllReadListener {
+
+    override val name: String = MODULE_NAME
+
+    companion object {
```
adasiewiczr, line 46: Can be private

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -0,0 +1,68 @@
...
+import io.getstream.chat.android.offline.experimental.plugin.listener.HideChannelListenerImpl
+import io.getstream.chat.android.offline.experimental.plugin.listener.MarkAllReadListenerImpl
+import io.getstream.chat.android.offline.experimental.plugin.listener.QueryChannelListenerImpl
+import io.getstream.chat.android.offline.experimental.plugin.listener.QueryChannelsListenerImpl
+import io.getstream.chat.android.offline.experimental.plugin.listener.ThreadQueryListenerImpl
+import io.getstream.chat.android.offline.experimental.plugin.logic.LogicRegistry
+import io.getstream.chat.android.offline.experimental.plugin.state.StateRegistry
+
+@ExperimentalStreamChatApi
+public class StreamOfflinePluginFactory(
```
adasiewiczr, line 27: Can we add KDocs?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -0,0 +1,68 @@
...
+            if (config.persistenceEnabled) offlineEnabled() else offlineDisabled()
+            if (config.userPresence) userPresenceEnabled() else userPresenceDisabled()
+            recoveryEnabled()
+            retryPolicy(config.retryPolicy.toLiveDataRetryPolicy())
+        }.build()
+
+        val stateRegistry =
+            StateRegistry.getOrCreate(io.getstream.chat.android.offline.ChatDomain.instance as ChatDomainImpl)
+        val logic = LogicRegistry.getOrCreate(stateRegistry)
+        val globalStateRegistry = GlobalMutableState.getOrCreate()
```
adasiewiczr, line 54: Can we move `GlobalState` to `StateRegistry`? 🤔

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/logic/LogicRegistry.kt
```diff
@@ -100,4 +100,16 @@ internal class LogicRegistry internal constructor(private val stateRegistry: Sta
...
+    internal companion object {
+        private var instance: LogicRegistry? = null
+
+        internal fun getOrCreate(stateRegistry: StateRegistry): LogicRegistry {
+            return instance ?: LogicRegistry(stateRegistry).also { logicRegistry ->
+                instance = logicRegistry
+            }
+        }
+
+        internal fun get(): LogicRegistry? = instance
```
adasiewiczr, line 113: Do we need this or we can only keep `getOrCreate` instead?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/state/StateRegistry.kt
```diff
@@ -68,4 +71,16 @@ public class StateRegistry internal constructor(
...
+    internal companion object {
+        private var instance: StateRegistry? = null
+
+        internal fun getOrCreate(chatDomainImpl: ChatDomainImpl): StateRegistry {
+            return instance ?: StateRegistry(chatDomainImpl).also { stateRegistry ->
+                instance = stateRegistry
+            }
+        }
+
+        internal fun get(): StateRegistry? = instance
```
adasiewiczr, line 84: Do we need this?

adasiewiczr (commented):
Great job! We are going into the right direction

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: I would prefer to avoid this `addPlugins` as well.
If we have the `PluginFactory`, why don't we use it to create plugins when they are needed? With it we will have `chatClient` already instanciated and be able to inject into `OfflinePlugin`.
Am I missing anything? :thinking:

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro, line 1939:
```suggestion
        public fun withPlugin(pluginFactory: PluginFactory): Builder = apply {
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -129,10 +130,6 @@ public class ChatClient internal constructor(
     private val userStateService: UserStateService = UserStateService(),
     private val tokenUtils: TokenUtils = TokenUtils,
     internal val scope: CoroutineScope,
-    private val appContext: Context,
```
JcMinarro, line 132: Really good we are removeing `Context` dependency on our `ChatClient`!!!! :partying_face: :tada:

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: @JcMinarro. If initialize then only when we need it, then it also won't go in the constructor. We can have the factories in the constructor and initialize the plugins when they are needed. I didn't want to expose the factories in the constructor of ChatClient, but this could be a way to initialize the plugins lazilly. But let me see if I can remove ChatDomain from OfflinePlugin what happens then.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/experimental/persistence/OfflinePlugin.kt
leandroBorgesFerreira: This interface is not necessary 🤦 . But the documentation applies to the OfflinePluginImpl as well.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/global/GlobalMutableState.kt
```diff
@@ -57,6 +57,18 @@ internal class GlobalMutableState : GlobalState {
...
+    internal companion object {
+        private var instance: GlobalMutableState? = null
+
+        internal fun getOrCreate(): GlobalMutableState {
+            return instance ?: GlobalMutableState().also { globalState ->
+                instance = globalState
+            }
+        }
+
+        internal fun create(): GlobalMutableState = GlobalMutableState()
```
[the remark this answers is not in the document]
leandroBorgesFerreira: We may want a method to force the creation of a new GlobalState. In our case, we use it in our unit tests, otherwise the state will be shared between tests.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/persistance/OfflineSupport.kt
leandroBorgesFerreira: Yep. That's just noise from interaction with the code. I should have deleted.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/state/StateRegistry.kt
leandroBorgesFerreira: Same thing with the above explanation.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: 🎉 🎉 🎉

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
leandroBorgesFerreira: Now that I took a better look at this class, I'm thinking... Do we even need it? It is a class aggregating the state of the whole SDK... But I feel this could be just an interface implemented by ChatClient, which actually have the state of the SDK already.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: So... Removing `ChatDomain` from `StateRegistry` is easy, but to remove it from `StreamOfflinePluginFactory.createOfflinePlugin` is much harder because much of the initialization of the SDK happens inside `ChatDomain`. 

We need to remove the initialization of `RepositoryFacade (this shouldn't be there anyway...), a way to provide the scopes and place to put and provide `lastUsers`. Let's talk about it over chat, I prefer this as a separated task.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
adasiewiczr: 🤔 currently `ChatClient` doesn't know anything about the "state". All the `StateFlows` related to channel or query channels belong to `OfflinePlugin`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/logic/LogicRegistry.kt
adasiewiczr: 👍 we can probably move `requireNotNull` check here and return the non-nullable `LogicRegistry` instead. If we do so, we should also mark this method with `@Throws` annotation

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/state/StateRegistry.kt
adasiewiczr: 👍 consider moving `nonNull` check here as in comment above

adasiewiczr (approved):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
leandroBorgesFerreira: Yep

Issue #3072, opened 2022-02-11
Create RepositoryFacade outside ChatDomain


leandroBorgesFerreira:
`RepositoryFacade` is initialized inside ChatDomainImpl, which causes many classes to be indirectly dependent on `ChatDomainImpl`. We should remove RepositoryFacade out of `ChatDomainImpl` and create it independently.

PR #3080, opened 2022-02-15
[ISSUE-3072] Repositories facade refactor

leandroBorgesFerreira:
[issue](https://github.com/GetStream/stream-chat-android/issues/3072)

### 🎯 Goal

Refactor RepositoryFacade. 

### 🛠 Implementation details

- Removing methods from Repositories that are not used
- Making some methods more generic in repositories. (Substituting methods like **selectPendingMessage** for **selectMessageByStatus**)
- Stop initializing FacadeRepository inside ChatDomain and move it to the Builder of ChatDomain and ChatClient. This way creation logic don't happen inside the classes not related to creation. To make initialization correctly and simpler, there's the introduction of `InitilizationCoordinator` with allows factories to listen for events in the creation of the SDK (user is connected, database it created...) and run its initialization only after mandatory events already happen. (Example: You can't initialize `RepositoryFacade` before the database is created)
- `RepositoryFacade` is now a Singleton. Although it is annoying to manage the state of a Singleton, RepositoryFacade should have only one instance, otherwise, we would fragment its cache and generate performance drops (Same values cached more than once or cache not hitting when it should...)

### 🧪 Testing

Test the persistence of the SDK. 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/RepositoryFacade.kt
```diff
@@ -155,4 +157,49 @@ internal class RepositoryFacade constructor(
...
+         */
+        internal fun refreshAndGet(factory: RepositoryFactory, scope: CoroutineScope, defaultConfig: Config) =
+            create(factory, scope, defaultConfig).also { facade -> instance = facade }
+
+        internal fun get(): RepositoryFacade = instance ?: throw IllegalStateException(
+            "RepositoryFacade must be initialised first"
+        )
+
+        @VisibleForTesting
+        internal fun create(factory: RepositoryFactory, scope: CoroutineScope, defaultConfig: Config): RepositoryFacade {
```
leandroBorgesFerreira, line 182: It is nice for unit test to be able to create an instance of it, but not as a singleton. Otherwise, the state is shared between unit tests which creates a mess.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -150,8 +151,7 @@ public class ChatClient internal constructor(
         }
     )
 
-    public val disconnectListeners: MutableList<(User?) -> Unit> = mutableListOf()
-    public val preSetUserListeners: MutableList<(User) -> Unit> = mutableListOf()
+    private val initializationCoordinator = InitializationCoordinator.getOrCreate()
```
adasiewiczr, line 154: Does it make sense to move it to a constructor?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2080,7 +2088,9 @@ public class ChatClient internal constructor(
                 module.userStateService,
                 scope = module.networkScope,
                 retryPolicy = retryPolicy,
-            )
+            ).also {
+                configureInitializer(it)
```
adasiewiczr, line 2092: We can get rid of this once we remove ChatDomain, right?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
```diff
@@ -0,0 +1,42 @@
+package io.getstream.chat.android.client.setup
+
+import io.getstream.chat.android.client.models.User
+
+// Make it internal once ChatDomain is merged with ChatClient.
+public class InitializationCoordinator private constructor() {
```
adasiewiczr, line 6: Can we add KDocs to this class and it's public methods?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/global/GlobalState.kt
```diff
@@ -98,4 +98,9 @@ public interface GlobalState {
      * @return True if initialized otherwise False.
      */
     public fun isInitialized(): Boolean
+
+    /**
+     * Clears the state of GlobalState
```
adasiewiczr, line 103:
```suggestion
     * Clears the state of [GlobalState].
```

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -45,17 +43,7 @@ public class StreamOfflinePluginFactory(
...
-        ChatDomain.Builder(appContext, chatClient).apply {
-            if (config.backgroundSyncEnabled) enableBackgroundSync() else disableBackgroundSync()
-            if (config.persistenceEnabled) offlineEnabled() else offlineDisabled()
-            if (config.userPresence) userPresenceEnabled() else userPresenceDisabled()
-            recoveryEnabled()
-        }.build()
-
-        val userStateFlow = MutableStateFlow<User?>(null)
-        chatClient.preSetUserListeners.add { user -> userStateFlow.value = user }
+        val userStateFlow = MutableStateFlow(ChatClient.instance().getCurrentUser())
```
adasiewiczr, line 46: Can we add a note somewhere to KDocs that there is a guarantee the plugin won't be created before current user is set?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    private val userDisconnectedListeners: MutableList<(User?) -> Unit> = mutableListOf()
+    private val userConnectedListeners: MutableList<(User) -> Unit> = mutableListOf()
+    private val databasedCreatedListeners: MutableList<() -> Unit> = mutableListOf()
```
adasiewiczr, line 10: As discussed on Slack - let's add TODO

stream-chat-android-offline/src/main/java/io/getstream/chat/android/livedata/ChatDomain.kt
```diff
@@ -613,7 +616,16 @@ public sealed interface ChatDomain {
             }
             val offlineChatDomain = offlineChatDomainBuilder.build()
             instance = buildImpl(offlineChatDomain)
-            return instance()
+
+            return instance().also {
+                RepositoryFacadeBuilder {
```
adasiewiczr, line 621: Shouldn't we try to move this initialization to `OfflinePlugin`? 🤔

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/creation/builder/RepositoryFacadeBuilder.kt
```diff
@@ -53,28 +55,18 @@ internal class RepositoryFacadeBuilder {
...
-        val userRepository = factory.createUserRepository()
-        val getUser: suspend (userId: String) -> User = { userId ->
-            requireNotNull(userRepository.selectUser(userId)) { "User with the userId: `$userId` has not been found" }
-        }
+        return RepositoryFacade.refreshAndGet(factory, scope, config)
+    }
 
-        val messageRepository = factory.createMessageRepository(getUser)
-        val getMessage: suspend (messageId: String) -> Message? = messageRepository::selectMessage
+    fun initialise() {
```
adasiewiczr, line 63: Why do we need an extra method for initialization here? Can't we just use `build`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/RepositoryFacade.kt
```diff
@@ -155,4 +157,49 @@ internal class RepositoryFacade constructor(
             }
         }
     }
+
+    internal companion object {
+        private var instance: RepositoryFacade? = null
+
+        internal fun initialise(factory: RepositoryFactory, scope: CoroutineScope, defaultConfig: Config) {
```
adasiewiczr, line 164: Does it make sense to replace it's usages with `refreshAndGet`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/RepositoryFacade.kt
```diff
@@ -155,4 +157,49 @@ internal class RepositoryFacade constructor(
...
+        /**
+         * Use it get create and refresh the singleton of [RepositoryFacade]. This is needed when we currentUser of the
+         * SDK is changes.
+         *
+         * Todo: Improve documentation
+         */
+        internal fun refreshAndGet(factory: RepositoryFactory, scope: CoroutineScope, defaultConfig: Config) =
+            create(factory, scope, defaultConfig).also { facade -> instance = facade }
+
+        internal fun get(): RepositoryFacade = instance ?: throw IllegalStateException(
```
adasiewiczr, line 177: Can we mark it with @Throw() ?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/RepositoryFacade.kt
```diff
@@ -26,7 +27,7 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
 import java.util.Date
 
-internal class RepositoryFacade constructor(
+internal class RepositoryFacade @VisibleForTesting internal constructor(
```
adasiewiczr, line 30: Do we still need an internal constructor here?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/RepositoryFacade.kt
```diff
@@ -155,4 +157,49 @@ internal class RepositoryFacade constructor(
...
+
+        internal fun initialise(factory: RepositoryFactory, scope: CoroutineScope, defaultConfig: Config) {
+            instance = create(factory, scope, defaultConfig)
+        }
+
+        /**
+         * Use it get create and refresh the singleton of [RepositoryFacade]. This is needed when we currentUser of the
+         * SDK is changes.
+         *
+         * Todo: Improve documentation
```
adasiewiczr, line 172: Would be nice to add KDocs to all those functions here

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt
```diff
@@ -58,6 +60,16 @@ class ChatInitializer(private val context: Context) {
             .withPlugin(offlinePlugin)
             .build()
 
+        InitializationCoordinator.getOrCreate().addUserConnectedListener {
```
adasiewiczr, line 63: Do we need to do the same change in compose sample?

leandroBorgesFerreira: At the end, there's no need for this code.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/RepositoryFacade.kt
leandroBorgesFerreira: We use it in the `BaseRepositoryFacadeTest`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
leandroBorgesFerreira: In the end, there's no need for `databasedCreatedListeners`. The initialization moved to `StreamOfflinePluginFactory`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: Now this listens for `addUserConnectedListener`. Listen for user connection is necessary because it is an asynchronous event.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -54,13 +60,19 @@ public class StreamOfflinePluginFactory(
...
-        val userStateFlow = MutableStateFlow<User?>(null)
-        chatClient.preSetUserListeners.add { user -> userStateFlow.value = user }
+        val scope = CoroutineScope(DispatcherProvider.IO)
+
+        val repos = RepositoryFacadeBuilder {
+            context(appContext)
+            scope(scope)
+            defaultConfig(io.getstream.chat.android.client.models.Config(connectEventsEnabled = true, muteEnabled = true))
+            ChatClient.instance().getCurrentUser()?.let(this::currentUser)
+            setOfflineEnabled(true)
```
adasiewiczr, line 70: We should make this configurable by some config, but we can handle that in a separate task

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -284,18 +275,6 @@ internal class ChatDomainImpl internal constructor(
...
             setUser(current)
         }
-        // past behaviour was to set the user on the chat domain
-        // the new syntax is to automatically pick up changes from the client
-        // listen to future user changes
-        client.preSetUserListeners.add {
-            setUser(it)
-        }
-        // disconnect if the low level client disconnects
-        client.disconnectListeners.add {
```
adasiewiczr, line 294: I think this part is missing in the new implementation. We should clear the state after disconnecting the user

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
leandroBorgesFerreira: But does it make sense to disable offline support now that we are using the `OfflinePlugin`? The way to disable it now should be not setting the plugins.

leandroBorgesFerreira: I think we could remove `setOfflineEnabled ` from RepositoryFacade, WDYT?

leandroBorgesFerreira: As decided over chat, let's remove it in a future moment, but keep it for now.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/experimental/plugin/handler/StateHandler.kt
```diff
@@ -0,0 +1,8 @@
+package io.getstream.chat.android.client.experimental.plugin.handler
+
+public interface StateHandler {
```
adasiewiczr, line 3: Can we add KDocs to this interface and public methods?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -76,6 +77,13 @@ public class StreamOfflinePluginFactory(
         val logic = LogicRegistry.getOrCreate(stateRegistry)
         val globalStateRegistry = GlobalMutableState.getOrCreate()
 
+        val stateHandler = StateHandlerImpl().apply {
+            registerClearStateListener {
+                logic.clear()
+                globalStateRegistry.clearState()
```
adasiewiczr, line 87: Should we also clear `stateRegistry`? 🤔

amitkma:
@leandroBorgesFerreira I noticed that offline is not working as expected. Two major things I found: 
1. Channels are not sorted (with latest message time) as we have them when we are online.
2. There are only 5-10 messages inside a channel when offline even though you scroll up and load more messages which should be available in database. 

Could you please confirm them?

leandroBorgesFerreira:
@amitkma I checked and the second problem is present both in develop and this branch... I'm not sure if the first one is one present in this branch. 

I'll add a fix for the second problem in this branch.

leandroBorgesFerreira:
I noticed that the sort order may vary a big both in develop and this branch and also previous versions of the SDK. Let's create an issue to take a look exclusively at that, we need to make sure we are storing our state consistently. But it doesn't seams related to this PR.

leandroBorgesFerreira:
I did some changes in the initialization of OfflinePlugin. The plugin should not be aware of user connection, as it is an offline plugin. Let's remove this dependency in the future. For now, it is initialized once the user is set, instead of connected.

leandroBorgesFerreira:
I found a bug related to user switch. I need to investigate it.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -76,8 +117,9 @@ public class StreamOfflinePluginFactory(
             deleteReactionListener = DeleteReactionListenerImpl(
                 logic = logic,
                 globalState = globalStateRegistry,
-                repos = chatDomainImpl.repos,
+                repos = repos,
             ),
+            stateHandler = stateHandler
```
amitkma, line 122: Why do we have to pass this stateHandler to OfflinePlugin?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/experimental/plugin/handler/StateHandler.kt
```diff
@@ -0,0 +1,20 @@
+package io.getstream.chat.android.client.experimental.plugin.handler
+
+/**
+ * Handles the state of stateful classes in the SDK. It can be used to clear the state of objects when a certain event
+ * occurs.
+ */
+public interface StateHandler {
```
amitkma, line 7: Isn't it possible to move this class to OfflinePlugin? It is not used in this module I believe.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/factory/StreamOfflinePluginFactory.kt
leandroBorgesFerreira: Not really, taking a look, StateHandler is not really necessary

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/experimental/plugin/handler/StateHandler.kt
leandroBorgesFerreira: Removed StateHandler.

leandroBorgesFerreira:
The problem indeed happens currently in `develop` branch. This may be a problem of API data overriding local data.

PR #3106, opened 2022-02-21
Allowing mutiple error handlers

leandroBorgesFerreira:
### 🎯 Goal

Proposal for change of API of ErrorHandler. This is not supposed to be merged, it only shows a concept.

### Implementation
- Instead of allowing only one handler, multiple error handlers can be allowed in ChatClient, this way a user can choose the handle error the way they prefer, with or without changing our default behavior. This way the customers can customize error handlers in a more granular way and without side effects (adding an error handler won't break the default behavior, it will just add a second fallback). 

- With error handlers returning `Call<T>` instead of `Result<T>` we keep the same structure of all Call modifiers. Just like `retry` extension, `doOnStart`, `doOnResult`, `precondition` we should always return Call<T> for the call modifies, then we end up with a consistent API which also looks nicer are Calls chain together in a nicer way.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -161,6 +163,7 @@ public class ChatClient internal constructor(
         PushNotificationReceivedListener { _, _ -> }
 
     public lateinit var plugins: List<Plugin>
+    public lateinit var errorHandlers: List<ErrorHandler>
```
leandroBorgesFerreira, line 165: We can have multiple error handlers, just like plugins. This way, if an error handler fails, more fallbacks can be used.

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/Call.kt
```diff
@@ -104,11 +104,10 @@ public fun <T : Any> Call<T>.withPrecondition(
  * @param function Function that returns data in the case of error.
  */
 @InternalStreamChatApi
-public fun <T : Any> Call<T>.onErrorReturn(
+public fun <T: Any> Call<T>.onErrorReturn(
     scope: CoroutineScope,
     function: suspend (originalError: ChatError) -> Result<T>,
-): Call<T> =
-    ReturnOnErrorCall(this, scope, function)
+): ReturnOnErrorCall<T> = ReturnOnErrorCall<T>(this, scope, function)
```
leandroBorgesFerreira, line 110: It would be nice it `onErrorReturn` returns `ReturnOnErrorCall` so this function can't be used just as a `map` of Call.

adasiewiczr (approved):

PR #3137, opened 2022-02-25
[3135] Allow passing long click handler to `MessageText`

MarinTolic:
### 🎯 Goal

Currently long pressing on any part of the text containing a link message will block it from opening message options. This is because 'Modifier.pointerInput()' inside `ClickableText` (foundation component) consumes all pointer events and does not back propagate them to the `Composable`s parent.

### 🛠 Implementation details

Trying to not consume all pointer events required too much low level re-writting so the following was implemented:

- Created a modified copy of `ClickableText` containing a long press handler
- Propagated a lambda to the previously mentioned long press handler that opens message options

### 🎨 UI Changes

**No UI changes**

### 🧪 Testing

Either find the long message containing both text and a link in the channel "Unknown Base" or create one of your own. Template should be plain text message + link.

- Launch Compose Sample App from the main or develop branch and long press on any part of the text of said message
- Launch Compose Sample App from the branch in this PR and long press on any part of the text of said message

Expected results: main/ develop branch versions will not open up message options while the version of the sample app from this branch will

#### General

#### Code & documentation
- [ ] ~~Comparison screenshots added for visual changes~~ (no visual changes)

### ☑️Reviewer Checklist

### 🎉 GIF

options
ll request_


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageText.kt
```diff
@@ -28,11 +35,13 @@ import io.getstream.chat.android.compose.ui.theme.ChatTheme
  *
  * @param message Message to show.
  * @param modifier Modifier for styling.
+ * @param onLongItemClick Handler used for long pressing on the message text.
  */
 @Composable
 public fun MessageText(
     message: Message,
     modifier: Modifier = Modifier,
+    onLongItemClick: (Message) -> Unit,
```
MarinTolic, line 44: This change is currently breaking in case anyone is using `MessageText` directly.
However I made this purposefully taking into consideration 2 thing.

1. We have a chance to do this before stable
2. It signals the issue to the customers and hopefully triggers them to read the release notes

It could easily be made non-breaking by introducing a default argument but I was worried that this would only hide the issue for our customers and not make them address it if they have it.

Also this isn't my favorite solution to the issue. I wanted to backpropagate events to the parent composable but doing that would require too much low level work as unfortunately `Modifier.pointerInput()` does not allow signaling that pointer events were not consumed.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
```diff
@@ -122,7 +122,10 @@ internal fun DefaultMessageContent(
         )
 
         if (message.text.isNotEmpty()) {
-            DefaultMessageTextContent(message = message)
+            DefaultMessageTextContent(
+                message = message,
+                onLongItemClick = onLongItemClick
```
MarinTolic, line 127: I kept our naming scheme for the sake of not deviating but I think it is not good here because:

1. `onLongItemClick` semantically says that the item is long and not the click
2. Not everything is an item, I think this should only apply to members of the Lazys

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageText.kt
```diff
@@ -140,3 +150,49 @@ private fun buildAnnotatedMessageText(message: Message): AnnotatedString {
         }
     }
 }
+
+/**
+ * A spin-off of a Foundation component that allows calling long press handlers.
+ * Contains only one additional parameter.
+ *
+ * @param onLongPress Handler called on long press.
```
MarinTolic, line 158: I did not want to simply copy paste the docs from the original implementation in 'foundation' or reinvent them.

If you think they are necessary I will do so.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
filbabic: 1) I don't think that's what people think personally, `onItemLongClick` isn't much different. It can also mean once the item performs a long click if you think about it. 

2) Item can be anything really. A timestamp can be an "item" within the container.

Should be fine IMHO

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageText.kt
filbabic: That's totally fine, people can browse the foundation impl

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageText.kt
filbabic: Sounds good - make sure to add it to the changelog as a "change"

filbabic (approved):
### ☑️Reviewer Checklist

PR #3147, opened 2022-02-28
[Proof of Concept] Synchronize chat events

MarinTolic:
This is a simple proof of concept PR used to validate the idea and should not be treated as a merge-ready PR.

### 🎯 Goal

Currently due to launching multiple coroutines our chat events can come in asynchronously which can cause race conditions.

The aim of this PR is to use `MutableSharedFlow` as a simple and effective synchronization mechanism while providing buffer capabilities for events that replace the previous events (connect events).

### 🛠 Implementation details

- Introduce relevant `MutableSharedFlows`
- Introduce collect and job hygiene (cancellation) functions

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
@@ -356,7 +356,9 @@ internal class ChatDomainImpl internal constructor(
     private fun startListening() {
         if (eventSubscription.isDisposed) {
             eventSubscription = client.subscribe {
-                eventHandler.handleEvents(listOf(it))
+                scope.launch {
+                    eventHandler.handleEvents(listOf(it))
+                }
             }
         }
     }
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
@@ -76,7 +76,7 @@ internal class EventHandlerImpl(
     private val client: ChatClient,
 ) {
     private val eventBuffer = MutableSharedFlow<List<ChatEvent>>()
-    private val connectEventBuffer = MutableSharedFlow<List<ChatEvent>>(
+    private val connectEventBuffer = MutableSharedFlow<ChatEvent>(
         replay = 0,
         extraBufferCapacity = 1,
         onBufferOverflow = BufferOverflow.DROP_OLDEST
@@ -88,7 +88,7 @@ internal class EventHandlerImpl(
     private var firstConnect = true
 
     internal suspend fun handleEvents(events: List<ChatEvent>) {
-        connectEventBuffer.emit(events)
+        bufferConnectEvents(events)
         eventBuffer.emit(events)
     }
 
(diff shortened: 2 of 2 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
```diff
@@ -65,65 +65,109 @@ import io.getstream.chat.android.offline.experimental.extensions.state
...
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.launch
 
 internal class EventHandlerImpl(
     private val domainImpl: ChatDomainImpl,
     private val client: ChatClient,
 ) {
+    private val eventBuffer = MutableSharedFlow<List<ChatEvent>>()
+    private val connectEventBuffer = MutableSharedFlow<ChatEvent>(
```
MarinTolic, line 79: This is optional, but the idea behind this is that these events should replace one another if they come in quick succession (e.g. lost connection but quickly reconnected). This way in the **probably rare** case of a busy consumer (collector) there will not be a case where we are still processing old and irrelevant values.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
```diff
@@ -65,65 +65,109 @@ import io.getstream.chat.android.offline.experimental.extensions.state
...
+import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.launch
 
 internal class EventHandlerImpl(
     private val domainImpl: ChatDomainImpl,
     private val client: ChatClient,
 ) {
+    private val eventBuffer = MutableSharedFlow<List<ChatEvent>>()
```
MarinTolic, line 78: `MutableSharedFlow` promises syncronization regardless of what thread the value was emitted from and its default strategy is to suspend when the consumer (collector) is busy. This way we do not lose any events.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/EventHandlerImpl.kt
```diff
@@ -65,65 +65,109 @@ import io.getstream.chat.android.offline.experimental.extensions.state
...
+        with(jobs) {
+            add(collectGeneralEvents())
+            add(collectConnectEvents())
         }
     }
 
-    private fun handleConnectEvents(sortedEvents: List<ChatEvent>) {
-        // send out the connect events
-        for (event in sortedEvents) {
+    internal fun stopCollectingEvents() {
```
MarinTolic, line 106: Hygiene point for cancelling jobs.
Unsure where this could be used.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/ChatDomainImpl.kt
```diff
@@ -356,7 +356,9 @@ internal class ChatDomainImpl internal constructor(
     private fun startListening() {
         if (eventSubscription.isDisposed) {
             eventSubscription = client.subscribe {
-                eventHandler.handleEvents(listOf(it))
+                scope.launch {
```
MarinTolic, line 359: We could possibly avoid using Flows as synchronization mechanisms if we made subscriptions suspending functions which would allow us to subscribe inside of a single coroutine body.

However I do not know if this may come at a cost for our users who do not use Coroutines. (would need further evaluation)

PR #3196, opened 2022-03-11
Improve creating channel offline

adasiewiczr:
### 🎯 Goal
Improve creating channel offline.

### 🛠 Implementation details
Channel's data like `name` and `image` wasn't properly filled when creating offline because the `ChatClient::createChannel` method receives `memberIds` instead of `members` but the `Channel` object requires a list of `Member` objects. In the previous implementation, we were creating members under the hood using `Member(User(id))` but this way we are losing information like `name`, `image` etc. There are two solutions for this problem:
1. Change the  `ChatClient::createChannel`  signature and pass `memberList` instead of `memberIds` - the problem is every FE SDK has almost the same signature and requires only ids so our implementation would diverge
2. Take users from cache (if exists) and fallback to `Member(User(id))`. We'll still have an edge case when the user is not cached in the DB yet, but this should work in most of the cases.

### 🎨 UI Changes

| Before | After |
| --- | --- |
|Screenshot_1646997982| Screenshot_1646997927|

### 🧪 Testing

Try to create a Group Channel offline and observe the photo will be properly filled if the user was already cached.

#### General
- [ ] ~PR is linked to the GitHub issue it resolves~

#### Code & documentation

### ☑️Reviewer Checklist


amitkma (approved):
### ☑️Reviewer Checklist

filbabic (approved):
Looks good! Still a small part where if a user isn't cached we have some issues, but that's fairly edge case and shouldn't be commonly reproduced.

### ☑️Reviewer Checklist

Issue #3032, opened 2022-02-04
Review ChatClientReferenceAdapter


adasiewiczr:
### Context & proposal
Our new approach uses [ChatClientReferenceAdapter](https://github.com/GetStream/stream-chat-android/blob/develop/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/adapter/ChatClientReferenceAdapter.kt) for exposing a state after doing a query.
Take a look at current implementation and think about possible improvements

PR #3201, opened 2022-03-14
[ISSUE-3032] Review chat client reference adapter2

leandroBorgesFerreira:
issue #3032 

### 🎯 Goal

Changing the API of ChatClientReferenceAdapter

### 🛠 Implementation details

Removing one extra call. Now the intermediate class always provides State.

### 🧪 Testing

Test a flow the uses the States

#### General

#### Code & documentation
- ~[ ] Changelog is updated with client-facing changes˜v 

### ☑️Reviewer Checklist

### 🎉 GIF
giphy


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/adapter/ChatClientStateCalls.kt
```diff
@@ -6,38 +6,46 @@ import io.getstream.chat.android.client.api.models.QueryChannelsRequest
...
         val (channelType, channelId) = cid.cidToTypeAndId()
         val userPresence = runCatching { ChatDomain.instance().userPresence }.getOrDefault(false)
         val request = QueryChannelPaginationRequest(limit).toWatchChannelRequest(userPresence)
         return queryChannel(channelType, channelId, request)
     }
 
     /** Reference request of the get thread replies query. */
-    public fun getReplies(messageId: String, limit: Int = DEFAULT_MESSAGE_LIMIT): RepliesQueryReference =
-        RepliesQueryReference(messageId, limit, chatClient)
+    internal fun getReplies(messageId: String, limit: Int = DEFAULT_MESSAGE_LIMIT): ThreadState {
```
adasiewiczr, line 45: Does it make sense to add one more parameter `scope` and then call:
```
scope.launch{
chatClient.getReplies(messageId, limit).await()
}
```
instead? This way calls could be easily cancelled 🤔

adasiewiczr, line 45: Does it make sense to add one more parameter `scope` and then call:
```
scope.launch{
chatClient.getReplies(messageId, limit).await()
}
```
instead? This way calls could be easily cancelled 🤔

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/extensions/ChatClient.kt
```diff
@@ -33,4 +37,18 @@ public val ChatClient.globalState: GlobalState
...
 
 @InternalStreamChatApi
-public fun ChatClient.asReferenced(): ChatClientReferenceAdapter = ChatClientReferenceAdapter(this)
+internal fun ChatClient.requestsAsState(): ChatClientStateCalls = ChatClientStateCalls(this, state)
+
+public fun ChatClient.queryChannelsAsState(request: QueryChannelsRequest): QueryChannelsState {
+    return requestsAsState().queryChannels(request)
+}
+
+public fun ChatClient.watchChannelAsState(cid: String, limit: Int = DEFAULT_MESSAGE_LIMIT): ChannelState {
```
adasiewiczr, line 46: What do you think about removing the default `limit` parameter here and in `getRepliesAsState `? There are cases when we don't need any messages (for example `MessageInputView`) and it's easy forgetting to change the default limit. Making this parameter required would force everyone to think if they really need any messages

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/extensions/ChatClient.kt
```diff
@@ -33,4 +37,18 @@ public val ChatClient.globalState: GlobalState
     get() = GlobalMutableState.getOrCreate()
 
 @InternalStreamChatApi
-public fun ChatClient.asReferenced(): ChatClientReferenceAdapter = ChatClientReferenceAdapter(this)
+internal fun ChatClient.requestsAsState(): ChatClientStateCalls = ChatClientStateCalls(this, state)
+
+public fun ChatClient.queryChannelsAsState(request: QueryChannelsRequest): QueryChannelsState {
```
adasiewiczr, line 42: Could we add KDocs to all new methods here?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/experimental/plugin/adapter/ChatClientStateCalls.kt
leandroBorgesFerreira: Thinking about it, it does make sense, because simply class Call.enqueue will cause memory leaks, because the calls won't be canceled =|. But how customers using Java will use this API?

filbabic: I think in this case enqueue does the same thing - we could have a scope to potentially launch a coroutine, but not much would be gained if this starts an async/parallel request anyway.

filbabic (approved):
This looks much nicer to me, nice work!

As far as the coroutines part goes - we could have it wrapped in a coroutine and then try to cancel it, but if the await operation or the coroutne is already launched, based on how quick we try to "cancel" it, it might not cancel at all before returning data.

So it could be an improvement, but could also not really do much 🤷 would have to test more intensively and that's hard to replicate

adasiewiczr (approved):

PR #3273, opened 2022-03-30
[13] Create internal logging tool

kanat:
### 🎯 Goal

Ticket: https://github.com/GetStream/android-internal-board/issues/13

Introduce `StreamLog` - a new parameterized file logging.

### 🛠 Implementation details

The logging is getting enabled by default in debug builds only, cause `stream-logger-android` module is getting linked as a dependency using `debugApi`.

`stream-logger-android` module contains `StreamLogFileProvider` class which is getting invoked right before `Application.onCreate()`. This class initializes the `StreamLog` with android & file logging.

### 🧪 Testing

- Cut new debug build of any sample app.
- Launch app 
- send in terminal `adb shell am startservice -a io.getstream.logging.android.CLEAR`
- you should see the toast message `Logs cleared`
- open any channel and send a few messages
- send in terminal `adb shell am startservice -a io.getstream.logging.android.SHARE`
- you should see a file-sharing dialog chooser.
- share log file
- check it's not empty & contains log content.

##### Option 1:
```kotlin
private const val TAG = "SomeViewModel"
// ...
StreamLog.d(TAG) { "[<method_name>] arg1: $arg1, arg2: $arg2, arg3: $arg3" }
```
##### Option 2:
```kotlin
val logger = StreamLog.getLogger("SomeViewModel")
// ...
logger.d { "[<method_name>] arg1: $arg1, arg2: $arg2, arg3: $arg3" }
```

#### General

#### Code & documentation
- [ ] ~~Changelog is updated with client-facing changes~~
- [ ] ~~New code is covered by unit tests~~
- [ ] ~~Comparison screenshots added for visual changes~~

### ☑️Reviewer Checklist

### 🎉 GIF


libraries/logging/stream-logger-android/build.gradle
```diff
@@ -0,0 +1,77 @@
...
+apply plugin: 'kotlin-android'
+apply plugin: 'de.mannodermaus.android-junit5'
+
+ext {
+    PUBLISH_GROUP_ID = Configuration.artifactGroup
+    PUBLISH_ARTIFACT_ID = 'stream-logger-android'
+    PUBLISH_VERSION = rootVersionName
+}
+
+apply from: "${rootDir}/scripts/publish-module.gradle"
```
adasiewiczr, line 14: Do we want to publish those libraries?

libraries/logging/stream-logger-android/detekt-baseline.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>EmptyFunctionBlock:ActivityLifecycleCallbacks.kt$ActivityLifecycleCallbacks${}</ID>
+    <ID>SwallowedException:StreamLogFileProvider.kt$StreamLogFileProvider$e: PackageManager.NameNotFoundException</ID>
```
adasiewiczr, line 6: Can we fix those issues before merging?

libraries/logging/stream-logger-android/src/main/java/io/getstream/logging/android/file/StreamLogFileManager.kt
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.logging.android.file
+
+import io.getstream.logging.android.file.StreamLogFileManager.ClearManager
+import io.getstream.logging.android.file.StreamLogFileManager.ShareManager
+
+public object StreamLogFileManager {
```
adasiewiczr, line 22: Can we add KDocs to this object, methods, etc?

libraries/logging/stream-logger-android/src/main/java/io/getstream/logging/android/file/impl/LifecycleAwareLogFileManager.kt
```diff
@@ -0,0 +1,68 @@
...
+import android.content.Intent
+import android.os.Handler
+import android.os.Looper
+import android.widget.Toast
+import androidx.core.content.FileProvider
+import io.getstream.logging.android.file.StreamLogFileManager
+import io.getstream.logging.file.FileStreamLogger
+import java.io.File
+
+internal class LifecycleAwareLogFileManager(
```
adasiewiczr, line 29: Let's add KDoc here

libraries/logging/stream-logger/src/main/java/io/getstream/logging/CompositeStreamLogger.kt
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.logging
+
+public class CompositeStreamLogger(
```
adasiewiczr, line 19: Let's add KDoc

libraries/logging/stream-logger/src/main/java/io/getstream/logging/StreamLog.kt
```diff
@@ -0,0 +1,213 @@
...
+
+package io.getstream.logging
+
+import io.getstream.logging.StreamLogger.Level.DEBUG
+import io.getstream.logging.StreamLogger.Level.ERROR
+import io.getstream.logging.StreamLogger.Level.INFO
+import io.getstream.logging.StreamLogger.Level.VERBOSE
+import io.getstream.logging.StreamLogger.Level.WARN
+
+public object StreamLog {
```
adasiewiczr, line 25: Missing KDoc

stream-chat-android-core/build.gradle
```diff
@@ -31,6 +31,8 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
 }
 
 dependencies {
+    api project(":libraries:logging:stream-logger")
```
adasiewiczr, line 34: Shouldn't we include it only as `debugApi`?

stream-chat-android-client/build.gradle
```diff
@@ -58,6 +58,7 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
 
 dependencies {
     api project(':stream-chat-android-core')
+    debugApi project(":libraries:logging:stream-logger-android")
```
adasiewiczr, line 61: I think we don't need to include this here as it was already added in `core` module

adasiewiczr (commented):
Great job overall! I left some small comments and questions.
Would love to get second review from @JcMinarro

adasiewiczr (commented):
One more thing - we should update `.github/CODEOWNERS` file

stream-chat-android-core/build.gradle
kanat: `core` contains a dependency on a `stream-logger` which is just an entry point and does not contain any logging behind.

`StreamLog` is disabled by default

stream-chat-android-client/build.gradle
kanat: This one brings an additional dependency `stream-logger-android` which contains `Logcat` wrapper, `File` wrapper and also enables the logging

libraries/logging/stream-logger-android/build.gradle
JcMinarro: I think we will need it as well we need it on the `core` module.
It won't be a "direct" dependency our customers will use, but a transitive dependency that could be needed

libraries/logging/stream-logger-android/src/main/java/io/getstream/logging/android/file/impl/ActivityLifecycleCallbacks.kt
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+package io.getstream.logging.android.file.impl
+
+import android.app.Activity
+import android.app.Application
+import android.os.Bundle
+
+internal abstract class ActivityLifecycleCallbacks : Application.ActivityLifecycleCallbacks {
```
JcMinarro, line 23: Do we really need this abstract class? :thinking:

libraries/logging/stream-logger-android/src/main/java/io/getstream/logging/android/file/impl/LifecycleAwareLogFileManager.kt
```diff
@@ -0,0 +1,68 @@
...
+    private val fileLogger: FileStreamLogger,
+) : ActivityLifecycleCallbacks(), StreamLogFileManager.ShareManager, StreamLogFileManager.ClearManager {
+
+    private val handler = Handler(Looper.getMainLooper())
```
JcMinarro, line 33: Instead of been using thread, couldn't we use coroutines?

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
```diff
@@ -0,0 +1,173 @@
...
+
+    private var currentFile: File? = null
+    private var currentWriter: Writer? = null
+
+    init {
+        executor.execute {
+            init()
+        }
+    }
+
```
JcMinarro, line 66: Perform logic, and on this case background tasks, is not a good idea.
As an example, if we would need to mock/spy this class in our tests, we won't be able to avoid this logic, that maybe is not needed for our tests.
Could you considere to extract it outside of the class??
One way could be to have a private constructor that receive all needed arguments, and having a companion method to build this an instance of this class

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
```diff
@@ -0,0 +1,173 @@
...
+private const val INTERNAL_FILE_1: String = "internal_1.txt"
+
+/**
+ * The [StreamLogger] implementation with log file persistence.
+ */
+public class FileStreamLogger(
+    private val config: Config,
+) : StreamLogger {
+
+    private val executor: Executor = Executors.newSingleThreadExecutor()
```
JcMinarro, line 49: Could we inject this collaborator by constructor?
On the other hand, could we move from "Threads" to "Coroutines"?

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
```diff
@@ -0,0 +1,173 @@
...
+        val internalFile: File = when {
+            !internalFile0.exists() || !internalFile1.exists() -> internalFile0
+            internalFile0.lastModified() > internalFile1.lastModified() -> internalFile0
+            else -> internalFile1
+        }
+        this.currentFile = internalFile
+        this.currentWriter = internalFile.fileWriter()
+    }
+
+    private fun swapFiles() = runCatching {
```
JcMinarro, line 112: Could you add the return type of this function?

libraries/logging/stream-logger/detekt-baseline.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>EmptyFunctionBlock:StreamLogger.kt$SilentStreamLogger${}</ID>
+    <ID>MaxLineLength:StreamLog.kt$IsLoggableValidator$*</ID>
+    <ID>MaxLineLength:StreamLog.kt$StreamLog$*</ID>
+    <ID>MaxLineLength:StreamLogger.kt$StreamLogger$*</ID>
```
JcMinarro, line 8: Can we fix those issues before merging?

libraries/logging/stream-logger/src/main/java/io/getstream/logging/StreamLog.kt
```diff
@@ -0,0 +1,213 @@
...
+
+    @PublishedApi
+    internal var logger: StreamLogger = SilentStreamLogger
+    @PublishedApi
+    internal var validator: IsLoggableValidator = IsLoggableValidator { _, _ -> false }
+
+    /**
+     * Sets custom [StreamLogger] implementation.
+     */
+    public fun init(
```
JcMinarro, line 35: It doesn't seem an `init()` but a `set()` method.
An `init()` method is expected to be called at the beginning and only once, and if by some reason it is called twice, notify about it.
WDYT about to rename it?

JcMinarro: On the other hand, we we need to update both params in one shot?
Maybe we can only use `var` fields and allow to update it from outside

libraries/logging/stream-logger/src/main/java/io/getstream/logging/StreamLogger.kt
```diff
@@ -0,0 +1,61 @@
...
+        public const val DEBUG: Priority = 3
+        /** Priority constant for the [log] method; use [StreamLog.i]. */
+        public const val INFO: Priority = 4
+        /** Priority constant for the [log] method; use [StreamLog.w]. */
+        public const val WARN: Priority = 5
+        /** Priority constant for the [log] method; use [StreamLog.e]. */
+        public const val ERROR: Priority = 6
+        /** Priority constant for the [log] method. */
+        public const val ASSERT: Priority = 7
+    }
```
JcMinarro, line 39: What about to have an enum class for those type?

libraries/logging/stream-logger/src/main/java/io/getstream/logging/helper/stringify.kt
```diff
@@ -0,0 +1,48 @@
...
+}
+
+public fun Priority.stringify(): String = when (this) {
+    StreamLogger.VERBOSE -> "V"
+    StreamLogger.DEBUG -> "D"
+    StreamLogger.INFO -> "I"
+    StreamLogger.WARN -> "W"
+    StreamLogger.ERROR -> "E"
+    StreamLogger.ASSERT -> "E"
+    else -> "?"
```
JcMinarro, line 37: With an enum, this else wouldn't be needed

libraries/logging/stream-logger/src/main/java/io/getstream/logging/kotlin/KotlinStreamLogger.kt
```diff
@@ -0,0 +1,49 @@
...
+    private val dateFormat: DateFormat = SimpleDateFormat("yyyy-MM-dd HH:mm:ss''SSS", Locale.ENGLISH),
+) : StreamLogger {
+
+    override fun log(priority: Priority, tag: String, message: String, throwable: Throwable?) {
+        val now = dateFormat.format(now())
+        val thread = Thread.currentThread().run { "$name:$id" }
+        val composed = "$now ($thread) [${priority.stringify()}/$tag]: $message"
+        val finalMessage = throwable?.let {
+            "$composed\n${it.stringify()}"
+        } ?: composed
```
JcMinarro, line 41: This logic seems to be duplicated across the PR, maybe we could extract it

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/handler/internal/EventHandlerImpl.kt
```diff
@@ -541,7 +547,7 @@ internal class EventHandlerImpl(
 
     private suspend fun handleEventsInternal(events: List<ChatEvent>, isFromSync: Boolean) {
         events.forEach { chatEvent ->
-            logger.logD("Received event: $chatEvent")
+            logger.v { "[handleEventsInternal] chatEvent: $chatEvent" }
```
JcMinarro, line 550:
```suggestion
            logger.d { "[handleEventsInternal] chatEvent: $chatEvent" }
```

JcMinarro (changes requested):

libraries/logging/stream-logger-android/detekt-baseline.xml
kanat: Those 2 are produced because of the following things:

1. **EmptyFunctionBlock**
```kotlin
internal abstract class ActivityLifecycleCallbacks : Application.ActivityLifecycleCallbacks {

    override fun onActivityCreated(activity: Activity, bunlde: Bundle?) {}

    override fun onActivityStarted(activity: Activity) {}

    override fun onActivityResumed(activity: Activity) {}

    override fun onActivityPaused(activity: Activity) {}

    override fun onActivityStopped(activity: Activity) {}

    override fun onActivitySaveInstanceState(activity: Activity, bunlde: Bundle) {}

    override fun onActivityDestroyed(activity: Activity) {}
}
```

2. **SwallowedException**
```
private fun Context.getVersionName(): String = try {
    packageManager?.getPackageInfo(packageName, 0)?.versionName ?: ""
} catch (e: PackageManager.NameNotFoundException) {
    ""
}
```

libraries/logging/stream-logger-android/src/main/java/io/getstream/logging/android/file/impl/ActivityLifecycleCallbacks.kt
kanat: Wanna keep `LifecycleAwareLogFileManager` clean from redundant functions

libraries/logging/stream-logger-android/src/main/java/io/getstream/logging/android/AndroidStreamLogger.kt
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The [StreamLogger] implementation for android projects.
+ */
+public class AndroidStreamLogger : StreamLogger {
+
+    override fun log(priority: Priority, tag: String, message: String, throwable: Throwable?) {
+
+        val androidPriority = priority.toAndroidPriority()
+        val androidTag = tag.takeIf { it.length > MAX_TAG_LEN && !isNougatOrHigher() }
```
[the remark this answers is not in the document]
kanat: Yup, that's true

Screenshot 2022-04-22 at 13 49 49

libraries/logging/stream-logger/src/main/java/io/getstream/logging/StreamLogger.kt
kanat: I keep as `Int` to make logic like
```kotlin
val validator = IsLoggableValidator { priority: Int, tag: String ->  
    priority > StreamLogger.INFO && tag in validTags
}
```

libraries/logging/stream-logger/src/main/java/io/getstream/logging/helper/stringify.kt
kanat: Answered above why I keep `Int`

libraries/logging/stream-logger/src/main/java/io/getstream/logging/kotlin/KotlinStreamLogger.kt
kanat: the `composed` is different in diff Loggers

**AndroidLogger**: 
```kotlin
val composed = "($thread) $message"
```

**StreamLogger**: 
```kotlin
val composed = "$now ($thread) [${priority.stringify()}/$tag]: $message"
```

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
kanat: There is a specified implementation for Tests named `KotlinStreamLogger`

libraries/logging/stream-logger-android/src/main/java/io/getstream/logging/android/file/impl/LifecycleAwareLogFileManager.kt
kanat: It's a fully android specific class and the idea is to keep this class off 3rd party libraries as coroutines or rxjava and use android SDK only.

libraries/logging/stream-logger-android/detekt-baseline.xml
JcMinarro: If we really need those empty methods (Maybe we don't really need this empty class and move the implementation to the class that implement this abstract class), we should add a comment like `// no-op`, with it we are sure it was thought to not do anything and it is not because we forgot to implement it

For the second one, as far as we are not using the exception, it can be called `_` (default Kotlin's name for ignored arguments) or contain `ignored` as a part of the name

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
kanat: I wanna keep this module clean from any 3rd party libs

kanat: yes we can move it constructor

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
JcMinarro: It is fine that we have another implementation that could be used on tests, but, what about classes that explicitly ask for a `FileStreamLoger`? It can't be mocked without running this logic.
Keeping `init()` block outside of our classes are good practiced

libraries/logging/stream-logger/detekt-baseline.xml
kanat: `EmptyFunctionBlock` comes from
```kotlin
public object SilentStreamLogger : StreamLogger {

    override fun log(priority: Priority, tag: String, message: String, throwable: Throwable?) {}
}
```

kanat: Others are fixed

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
JcMinarro: We can use a default params with the `Executors.newSingleThreadExecutor()` value, but allowing to inject a different one on the case the one that will use this class needs it

libraries/logging/stream-logger/src/main/java/io/getstream/logging/StreamLogger.kt
JcMinarro: You can have almost the same logic with enum:
```
enum class Priority(val level: Int) {
    VERBOSE(2),
    DEBUG(3),
    INFO(4),
    WARN(5),
    ERROR(6),
    ASSERT(7),
}

val validator = IsLoggableValidator { priority: Priority, tag: String ->
    priority.level > Priority.INFO.level
}
```
We can create a function into our enum class to return if the level is bigger

JcMinarro: Using a "real type", other devs can know which values they can use without digging into the class to see which values are valid and which one not, allowing us to refactor it in the future without breaking our customers logic

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
kanat: `FileStreamLoger` is used at `StreamLogFileProvider` only. That ContentProvider will be included in Debug builds only. There is no need to mock `FileStreamLoger`.

libraries/logging/stream-logger-file/src/main/java/io/getstream/logging/file/FileStreamLogger.kt
kanat: But I don't wanna allow passing any diff Executor by design to restrict it to be always a SingleThread one.

libraries/logging/stream-logger/src/main/java/io/getstream/logging/StreamLogger.kt
kanat: What is a big advantage of turing it into `enum`?
I won't use enum as it is cause I still will be converting names to:
```kotlin
 VERBOSE -> "V"
```

JcMinarro: What is the advantage of using `typealias Priority = Int`? 
It does nothing because we don't limit what you can use to define the log level.
Keeping a good type structure on our code helps us to reduce bugs in the future.
If you use Int, and for some reason, we need to introduce an intermediate value, it will break all our customer's apps.
Instead of that, if we have a clear data structure (In this case, an enum type), it won't break our customer's apps because they would be using "an Int value" that doesn't make sense

libraries/logging/stream-logger-android/detekt-baseline.xml
JcMinarro: It is ok you create the `ActivityLifecycleCallbacks` class, but it needs to follow our Detek Rules

JcMinarro (approved):

adasiewiczr (approved):

PR #3321, opened 2022-04-11
Fix Detekt Issues on Compose module

JcMinarro:
### 🎯 Goal
Fix majority of the issues we had on Compose Module 
Fixes https://github.com/GetStream/android-internal-board/issues/18

#### General
- ~[ ] Thread with the PR link started in a respective Slack channel (#android-chat-core or #android-chat-ui)~

#### Code & documentation

### ☑️Reviewer Checklist


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/AttachmentFactory.kt
```diff
@@ -28,7 +28,8 @@ import io.getstream.chat.android.compose.ui.util.previewText
  * @param canHandle Checks the message and returns if the factory can consume it or not.
  * @param previewContent Composable function that allows users to define the content the [AttachmentFactory] will build,
  * using any given [AttachmentState], when the message is displayed in the message input preview, before sending.
- * @param content Composable function that allows users to define the content the [AttachmentFactory] will build using any given
+ * @param content Composable function that allows users to define the content the [AttachmentFactory] will build
+ * using any given
  * [AttachmentState], when the message is displayed in the message list.
```
filbabic, line 33: can we pull this up in the previous line? or format a bit nicer

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -403,12 +405,10 @@ public class MessageListViewModel(
             return
         }
 
-        if (message.id == lastSeenMessage.id) return
```
filbabic, line 406: The reason why we have this condition here is that we don't want to fetch/allocate any variables in case the IDs are the same. By doing it later, we fetch the 2 properties and potentially create 2 date instances, without having to do so.

To me this is not an optimization it's just making this more confusing and less optimized

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messageoptions/MessageOptions.kt
```diff
@@ -129,90 +129,159 @@ public fun defaultMessageOptionsState(
     val isMessageFailed = selectedMessage.syncStatus == SyncStatus.FAILED_PERMANENTLY
 
     return listOfNotNull(
```
filbabic, line 131: I think this change is also less optimized - as we'll have X more composable slots in the UI tree that recompose every time the state changes, rather than one slot that recomposes everything.

But I'll have to check and test

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/handler/MediaAttachmentPreviewHandler.kt
```diff
@@ -31,21 +31,14 @@ public class MediaAttachmentPreviewHandler(private val context: Context) : Attac
...
-        } else {
-            val supportedMimeSubTypes = buildMimeSubTypeList()
-            supportedMimeSubTypes.any { subtype -> subtype in mimeType }
+        return when {
+            assetUrl.isNullOrEmpty() -> false
+            mimeType?.contains(ModelType.attach_audio) == true -> true
+            mimeType?.contains(ModelType.attach_video) == true -> true
+            type?.contains(ModelType.attach_audio) == true -> true
+            type?.contains(ModelType.attach_video) == true -> true
+            buildMimeSubTypeList().any { subtype -> mimeType?.contains(subtype) == true } -> true
```
filbabic, line 40: Again, less optimized - we build a list that allocates memory and then we have an O(N) check if the type exists. It's much more optimized to just return false if the mimeType is `null`.

filbabic (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/handler/MediaAttachmentPreviewHandler.kt
JcMinarro: This condition can be added before building the list of subtypes (As you did on the PR ;) )
The idea with this rule is not "only" optimized the logic, but to reduce the number of different conditions a method can run. Having only one return condition, it is easier to modify it later by adding/removing some of them

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messageoptions/MessageOptions.kt
filbabic: It depends on how it adds the item to the UI tree - but it's the same, yes, tested! :]

It's just a matter of 1 function being a composable and X of them, but in the end it seems to be optimized anyway

filbabic (approved):

adasiewiczr (approved):

PR #3365, opened 2022-04-20
Repository factory refactor

leandroBorgesFerreira:
This PR should be merge after #3358 

### 🎯 Goal

Refactoring the initialization of `RepositoryFacade` and `RepositoryFactory`.

### 🛠 Implementation details

Adding the possibility for users to customize the repositories and use their own version of repositories in OfflinePlugin. This allows the users to have their own database system as long it implements our interfaces.

### 🧪 Testing

Test persistence in the SDK

#### General
- ~[ ] PR is linked to the GitHub issue it resolves~

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/factory/RepositoryProvider.kt
```diff
@@ -0,0 +1,33 @@
...
+
+        private var instance: RepositoryProvider? = null
+
+        public fun changeRepositoryFactory(repositoryFactory: RepositoryFactory) {
+            instance = RepositoryProvider(repositoryFactory)
+        }
+
+        public fun get(): RepositoryProvider? = instance
+    }
+}
```
kanat, line 33: Hey @leandroBorgesFerreira,

I can see we set a `RepositoryFactory` instance though `changeRepositoryFactory`, but we never use `RepositoryProvider.get`. So I'm wondering what is the purpose of this class.

Do we really need to provide a singleton which holds `RepositoryFactory` reference?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/factory/RepositoryProvider.kt
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.persistance.repository.factory
+
+public class RepositoryProvider private constructor(
+    private val repositoryFactory: RepositoryFactory,
```
kanat, line 20: Are KDocs missing here?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/factory/RepositoryFactory.kt
```diff
@@ -0,0 +1,52 @@
...
+    public fun createMessageRepository(
+        getUser: suspend (userId: String) -> User,
+    ): MessageRepository
+
+    public fun createReactionRepository(getUser: suspend (userId: String) -> User): ReactionRepository
+
+    public fun createSyncStateRepository(): SyncStateRepository
+
+    public fun createAttachmentRepository(): AttachmentRepository
+}
```
kanat, line 52: Are KDocs missing here?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -76,6 +81,12 @@ public class StreamOfflinePluginFactory(
      */
     override fun get(user: User): Plugin = createOfflinePlugin(user)
 
+    private var repositoryFactory: RepositoryFactory? = null
+
+    public fun setRepositoryFactory(repositoryFactory: RepositoryFactory) {
+        this.repositoryFactory = repositoryFactory
+    }
+
```
kanat, line 89: Could we potentially pass something like `RepositoryFactoryCreator` into `StreamOfflinePluginFactory` constructor with a default implementation provided? What do you think?

leandroBorgesFerreira: We already default to `DatabaseRepositoryFactory`, if the user doesn't set a custom factory. That's the behavior that I expect to happen in most cases. 
We can ask that `RepositoryFactory` is passed in the constructor of `StreamOfflinePluginFactory`, but that would complicate this class, as users of Java can't use default values and they will be obligated to pass the `RepositoryFactory` inside `StreamOfflinePluginFactory`. In the current way, they would simply use the default factory.

kanat: Just curious if we could use `StreamOfflinePluginFactory @JvmOverloads constructor(...)`.
What do you think? :)

leandroBorgesFerreira: `@JvmOverloads` Would indeed solve this problem. Another problem that I see is that `DatabaseRepositoryFactory` need `User` which is only obtained when `StreamOfflinePluginFactory.get` is called, so it isn't available when the constructor is called

kanat: Yup, that's why I thought we might have something like `RepositoryFactoryCreator` and pass it to `StreamOfflinePluginFactory` constructor with `@JvmOverloads`:
```kotlin
interface RepositoryFactoryCreator {
    fun createRepositoryFactory(
            scope: CoroutineScope,
            context: Context,
            user: User?,
            offlineEnabled: Boolean,
    ): RepositoryFactory
}
```

Just thinking out loud :)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/factory/RepositoryFactory.kt
```diff
@@ -0,0 +1,88 @@
...
+import io.getstream.chat.android.client.persistance.repository.ReactionRepository
+import io.getstream.chat.android.client.persistance.repository.SyncStateRepository
+import io.getstream.chat.android.client.persistance.repository.UserRepository
+
+/**
+ * Factory that creates all repositories of SDK.
+ */
+public interface RepositoryFactory {
+
+    public fun <T : Any> get(classz: Class<T>): T
```
kanat, line 35: Would it be possible to use `KClass` instead?

leandroBorgesFerreira: I thought about `KClass`, but wouldn't this be worse for Java users?

kanat (approved):
LGTM

PR #3377, opened 2022-04-22
Fixes attachment upload crash

### LINKED ISSUE #3360: Crash when switching user if there is any upload work pending
### Context & proposal
When a user is disconnected, the LogicRegistry, StateRegistry, repos, etc are cleared. So if at that moment, UploadAttachmentWorker tries to upload an attachment, it crashes because it internally uses `LogicRegistry` instance which is null.
We should cancel any pending work if a user is disconnected. Also, we should check in `doWork` of the upload worker and ensure that we start uploading only when user is connected.

### Acceptance criteria

- [ ] There is no crash when switching users even if there are pending attachments.
- [ ] Any pending attachment should not upload once the user logout.

amitkma:
Closes #3360 
### 🎯 Goal

When user is disconnected, we clear `LogicRegistry` and `StateRegistry`. So if there is any pending upload work that requires LogicRegistry etc would crash. This PR is intended to fix this issue. 

### 🛠 Implementation details

- Work id is stored and the work is canceled using that id when the user disconnects.

### 🧪 Testing

- In the sample app, Start uploading some attachments.
- When attachments are being uploaded or before any attachment upload starts, sign-out that user, and there should be no crash/exception in logs.

#### General

#### Code & documentation

### ☑️Reviewer Checklist


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/interceptor/internal/SendMessageInterceptorImpl.kt
```diff
@@ -232,21 +233,25 @@ internal class SendMessageInterceptorImpl(
             Result.success(messageToBeSent.copy(type = Message.TYPE_REGULAR))
         } else {
             Result.error(ChatError("Could not upload attachments, not sending message with id ${newMessage.id}"))
+        }.also {
+            uploadIds.remove(newMessage.id)
```
leandroBorgesFerreira, line 237: I believe we need to remove the id from `uploadIds` only if the result is successful. Otherwise, the worker will still exist and retry sending the attachment and we won't be able to cancel it, if needed

amitkma: I think this `Result.error` is not from worker but when all attachments of this message have `upload failed state`. That should happen once a worker has finished his job. So it doesn't seem right to keep this work id stored.

filbabic (approved):
### ☑️Reviewer Checklist

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/interceptor/internal/SendMessageInterceptorImpl.kt
kanat: Based on `UploadAttachmentsAndroidWorker` & `UploadAttachmentsWorker` it seems like worker waits until all attachments get processed and it fails if one of those attachment got failed.

It looks like we're good to remove an `uploadId` here in both cases

kanat (approved):
Hey @amitkma 👋

As a future improvement:
Can we combine `jobs` & `uploadIds` to keep tracking the uploading process in a single collection?

### ☑️Reviewer Checklist

PR #3405, opened 2022-04-26
Disable Detekt rules on Docs Module

JcMinarro:
### 🎯 Goal
Disable Detekt rules on Docs Module


bychkovdmitry (approved):
LGTM, `./gradlew detektBaseline` no longer dumps warning for the module.

PR #3464, opened 2022-05-04
Api request analyser

leandroBorgesFerreira:
### 🎯 Goal

Propose a plugin to allow users to debug our SDK using `MessageListView`. Only one request was added at the moment because this is a proposal that should be discussed by the team, so this code only illustrates the concept.

### 🛠 Implementation details

Added `ApiRequestsAnalyser` that registers the calls and can be used to bump the requests providing information about when the API was called and with parameters we used. 

### 🧪 Testing

Use the command `request_dump`

### 🎉 GIF

giphy (1)
_dump_


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2368,6 +2376,10 @@ internal constructor(
             userCredentialStorage = credentialStorage
         }
 
+        public fun withApiRequestAnalyser(apiRequestsAnalyser: ApiRequestsAnalyser): Builder = apply {
```
skydoves, line 2379: What do you think about adding a parameter to check if this only works in Debug mode like below?

```kotlin
public fun withApiRequestAnalyser(apiRequestsAnalyser: ApiRequestsAnalyser, debugOnly: Boolean = true): Builder = apply {
     if (debugOnly && BuildConfig.DEBUG) {
         this.apiRequestsAnalyser = apiRequestsAnalyser
     }
}
```

We need to prevent being enabled this option by mistake and clients don't need to split up their initialization codes to check whether it is a debug mode.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugins/requests/ApiRequestsDumper.kt
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Date
+import java.util.Locale
+
+private const val NOT_FOUND = "not found"
+
+internal class ApiRequestsDumper(
+    private val dateFormat: DateFormat = SimpleDateFormat("HH:mm:ss.SSS", Locale.getDefault())
+) : ApiRequestsAnalyser {
+
+    private val requestsDataMap: MutableMap<String, MutableList<RequestData>> = mutableMapOf()
```
skydoves, line 14: What about adding a `clearDataCacheByName` or `clearAllDataCache` interface/method for the `ApiRequestsAnalyser`? The message cache can be huge and that makes it difficult to track specific requests.

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
```diff
@@ -37,7 +37,7 @@ class App : Application() {
         super.onCreate()
         chatInitializer.init(getApiKey())
         instance = this
-        DebugMetricsHelper.init()
+        // DebugMetricsHelper.init()
```
skydoves, line 40: Out of curiosity, why do we need to remove this?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
```diff
@@ -286,10 +286,19 @@ public class MessageInputView : ConstraintLayout {
     }
 
     public fun setCommands(commands: List<Command>) {
-        suggestionListController?.commands = commands
+        suggestionListController?.commands = commands + requestDump()
```
skydoves, line 289: Perhaps we need to check if the `apiRequestsAnalyser` was registered?

skydoves (commented):
I left several comments :)
BTW, This analyzer looks really helpful to log requests.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -177,6 +178,7 @@ internal constructor(
     internal val retryPolicy: RetryPolicy,
     private val initializationCoordinator: InitializationCoordinator = InitializationCoordinator.getOrCreate(),
     private val appSettingsManager: AppSettingManager,
+    public val apiRequestsAnalyser: ApiRequestsAnalyser?
```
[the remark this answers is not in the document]
leandroBorgesFerreira: I preferred to delete it in the end. Let's not add this to the constructor o ChatClient. As this is a debug tool, users can use its Factories static method

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/input/MessageInputView.kt
leandroBorgesFerreira: This got out dated

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
leandroBorgesFerreira: Ops, my mistake 🤦

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2336,6 +2339,19 @@ internal constructor(
             userCredentialStorage = credentialStorage
         }
 
+        public fun withApiRequestAnalyser(
```
skydoves, line 2342: It seems to missed KDocs here. Also, It would be great if we could add @JvmOverloads. 😄

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2336,6 +2339,19 @@ internal constructor(
...
+            apiRequestsAnalyser: ApiRequestsAnalyser,
+            debugOnly: Boolean = true,
+        ): Builder = apply {
+            if (debugOnly && BuildConfig.DEBUG) {
+                this.apiRequestsAnalyser = apiRequestsAnalyser
+            }
+        }
+
+        public fun debugRequests(shouldDebug: Boolean): Builder = apply {
```
skydoves, line 2351: Same here, KDocs are missing.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugins/requests/ApiRequestsAnalyser.kt
```diff
@@ -0,0 +1,42 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.plugins.requests
+
+public interface ApiRequestsAnalyser {
```
skydoves, line 19: It seems to missed KDocs here.

skydoves (commented):
New changes look good! I left some minor comments.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: I deleted this method. The `debugRequests` is enough.

skydoves (approved):
Looks good to me! 🙌

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatClientConfig.kt
```diff
@@ -18,13 +18,15 @@ package io.getstream.chat.android.client.api
...
 
+@Suppress("LongParameterList")
 public class ChatClientConfig(
     public val apiKey: String,
     public var httpUrl: String,
     public var cdnHttpUrl: String,
     public var wssUrl: String,
     public val warmUp: Boolean,
     public val loggerConfig: ChatLogger.Config,
+    public val debugRequests: Boolean
```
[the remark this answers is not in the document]
kanat: So would be nice to add `debugRequests` param to that KDoc as well

kanat:
Hey @leandroBorgesFerreira ,

I wonder if you want to add some description about this flag to Docusarus :)

kanat (approved):
LGTM

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/home/HomeFragment.kt
```diff
@@ -79,13 +82,27 @@ class HomeFragment : Fragment() {
...
             setOnActionButtonClickListener {
                 navigateSafely(R.id.action_homeFragment_to_addChannelFragment)
             }
             setOnUserAvatarClickListener {
                 binding.drawerLayout.openDrawer(GravityCompat.START)
             }
             if (BuildConfig.DEBUG) {
+                setOnTitleClickListener {
+                    if (ApiRequestsAnalyser.isInitialized()) {
+                        Log.d("ApiRequestsAnalyser", ApiRequestsAnalyser.get().dumpAll())
```
bychkovdmitry, line 95: I would add a toast here as well

bychkovdmitry (approved):
LGTM from the UI side

PR #3466, opened 2022-05-05
[3364] Get rid of InitConnectionListener usage in ChatClient

kanat:
### 🎯 Goal

`InitConnectionListener` has been converted into an `internal` interface which makes its usage in `ChatClient` useless.

The goal is to get rid of it from `ChatClient`.

### 🧪 Testing

Run 
- offline/online swiches
- logout & login within the same sessions

Make sure ChatClinet reconnects successfully.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -467,50 +440,58 @@ internal constructor(
...
     }
 
-    private fun setGuestUser(userId: String, username: String, listener: InitConnectionListener? = null) {
-        getGuestToken(userId, username).enqueue { result ->
+    @CheckResult
+    public fun connectGuestUser(userId: String, username: String): Call<ConnectionData> {
+        return CoroutineCall(scope) {
+            logger.logD("[connectGuestUser] userId: '$userId', username: '$username'")
+            val result = getGuestToken(userId, username).await()
             if (result.isSuccess) {
```
leandroBorgesFerreira, line 480: You can use mapSuspend to avoid the if/else. 

```
getGuestToken(userId, username).await()
                .mapSuspend { guestUser ->
                    setUser(guestUser.user, ConstantTokenProvider(guestUser.token))
                }
                .data()
                .also { finalResult ->
                    logger.logV(
                        "[connectAnonymousUser] completed: ${
                            finalResult.stringify { "ConnectionData(connectionId=${it.connectionId})" }
                        }"
                    )
                }
```

Not mandatory, do if you see an improvement.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -325,49 +289,49 @@ internal constructor(
...
                 userState.user.id == user.id &&
                 socketStateService.state == SocketState.Idle -> {
+                logger.logV("[setUser] user is Set & socket.state is Idle")
                 userStateService.onUserUpdated(user)
                 tokenManager.setTokenProvider(cacheableTokenProvider)
-                connectionListener = listener
                 socketStateService.onConnectionRequested()
                 socket.connect(user)
                 initializationCoordinator.userConnected(user)
+                waitConnection.first()
```
leandroBorgesFerreira, line 316: Can we guarantee that this value will always the non null?

kanat: This contract comes from the deleted `InitConnectionListener `:
```kotlin
internal open class InitConnectionListener {

     open fun onSuccess(data: ConnectionData) {
     }

     open fun onError(error: ChatError) {
     }

     data class ConnectionData(val user: User, val connectionId: String)
}
```

So that's why `waitConnection` do not accept null values:
```kotlin
private val waitConnection = MutableSharedFlow<Result<ConnectionData>>()
```

JcMinarro (commented):
Look Great!!!
Do you think we can add any test on this part to verify we don't break this critical path?

kanat:
> Look Great!!!
> Do you think we can add any test on this part to verify we don't break this critical path?

I've been relying on existing UT, cause this change does not introduce any new logic, however, I've added `ConnectUsersTest` in addition.

Issue #3542, opened 2022-05-18
BiDirectional infinite scroll from MessageListView


leandroBorgesFerreira:

leandroBorgesFerreira:
When the messages in MessageListView become non-linear (example: Searching for a message) the view should start to fetch messages when scrolling bottom, not only top, until it becomes linear again.

PR #3615, opened 2022-05-27
Infinite scroll to bottom

leandroBorgesFerreira:
issue: https://github.com/GetStream/stream-chat-android/issues/3542

### 🎯 Goal

Adding requests for new messages at the bottom when there's a gap in `MessageListView`.

### 🛠 Implementation details

- Added flag to notify that when a request can start a gap in the message list. This can happen, for example, when a search happens. 
- Not allowing to request many times for new messages at the bottom. Otherwise, the request would happen twice and a merge would happen with the same request. 
- ChannelLogic helps the messages at the bottom of the gap. When an overlap happens, the gap is considered closed. 

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/170691725-93b5e2af-fda7-4fcd-8721-f88a7ae8c802.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/170691044-25689360-8ed4-4523-b106-305aa8e3279d.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Search a message and scroll bottom. I used "secret message" in my account =P.

#### General

#### Code & documentation
- [ ] Affected documentation updated (KDocs, docusaurus, tutorial). Todo?

### ☑️Reviewer Checklist

### 🎉 GIF


leandroBorgesFerreira:
The PR is not considering messages coming from the database, which makes insistencies when evaluating the gap. There's still more to be done

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/MessagesGapInfo.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.offline.plugin.logic.channel.internal
+
+public data class MessagesGapInfo(
+    val messagesAboveGap: List<Long>,
+    val messagesBellowGap: List<Long>
+)
```
kanat, line 22: What do those parameters mean in the historical axis?
Does `Above` imply either `older` or `newer` messages (same for `Below`)?

kanat:
Hey @leandroBorgesFerreira,

Do we have any ticket for this PR?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/ChannelRequest.kt
```diff
@@ -22,6 +22,7 @@ internal interface ChannelRequest<T : ChannelRequest<T>> {
     var state: Boolean
     var watch: Boolean
     var presence: Boolean
+    var canCreateGap: Boolean
```
kanat, line 25: Can we add a KDoc here?
It's hard to grasp what this property is for.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -124,6 +124,13 @@ internal class ChannelLogic(
     private val attachmentUrlValidator: AttachmentUrlValidator = AttachmentUrlValidator(),
 ) : QueryChannelListener {
 
+    private val messageIdsBellowGap = mutableListOf<Long>()
+    private val messageIdsAboveGap = mutableListOf<Long>()
```
kanat, line 128: Can we rename it in respect of the historical axis?

kanat:
I'm still able to see GAPs in history.
Message to search: `Secret message`
Tested channel: `Skinny Secret`
Date where GAP appeared: `April 27-28`

| Here is the screenshot from `InifiteScrollBranch` (found the message and was scrolling down to the newer messages) | Here is the screenshot from `AppTester` build `1.4111.0.e61d4f3` (opened channel and was scrolling up to older messages) |
| --- | --- |
| InfiniteScrollToBottom | AppTester_1_4111_0_e61d4f3 |

leandroBorgesFerreira:
Hi Kanat. The logic has a window to trigger the request for new messages. It works like this to a avoid requesting too many messages in a row if the user scrolls all the way down at once. So you can still create a gap, if the scrolling happens too fast... That was probably your case. 

I'm working in a second PR that makes MessageListView aware of gaps scroll both up and down, which should fix this problem. But I created this PR as an intermediate one, because the logic is complicated and it was taking too long to merge something. Besides there are some breaking changes that need to evaluated.

The solution is still not 100%, but I think it already improves the SDK

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/MessagesGapInfo.kt
leandroBorgesFerreira: above is older and bellow is newer. The problem is that older messages than the gap may be added when loading newer messages and newer messages may be added when loading older messages... At first I thought about naming then older and newer messages than gap, but in the end I thought is was more confusing than clarifying.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
leandroBorgesFerreira: I thought about naming then older and newer. But this got confusing. This part of the logic is very oriented by the direction and location of the messages and not so much about history. I would get confused using names related to historical axis.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/MessagesGapInfo.kt
kanat: I mean the names do give me no clue if the Above is for newer or not.
It's more like a description in UI positioning like Above something.

kanat: Let's imagine we (or customers) changed the orientation of the history axis. 
And now `Above` implies newer, while `Below` - older.

leandroBorgesFerreira:
🤔  Let me check something then...

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/MessagesGapInfo.kt
leandroBorgesFerreira: Let's use `Forward` and `Backward` then, it still uses direction.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/MessageListView.kt
```diff
@@ -624,6 +647,38 @@ public class MessageListView : ConstraintLayout {
         }
     }
 
+    public fun gapInMessage(hasGap: Boolean, gapInfo: MessagesGapInfo?) {
```
TGazica, line 650: Please add documentation for the function.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/view/EndlessMessageListScrollListener.kt
```diff
@@ -59,20 +71,61 @@ public class EndlessMessageListScrollListener(
...
+            }
+
+            dy < 0 -> {
+                handleScrollUp(layoutManager, recyclerView)
+            }
+        }
+    }
+
+    private fun canRequestNow(): Boolean {
+        return (Date().time - lastRequest > bottomRequestTriggerDelay)
```
TGazica, line 94: shouldn't the check be if loading is in progress return false, otherwise true, why are we using a delay?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -246,10 +327,19 @@ internal class ChannelLogic(
...
+        canCreateGap: Boolean,
+    ): Result<Channel> {
+        return runChannelQuery(
+            olderWatchChannelRequest(limit = messageLimit, baseMessageId = baseMessageId).apply {
+                this.canCreateGap = canCreateGap
+            }
+        )
     }
 
+    // Colocar se as mensagens são mais novas ou mais velhas
```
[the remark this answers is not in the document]
leandroBorgesFerreira: ops, I need to delete that 🤦

leandroBorgesFerreira:
Closed in favor of: https://github.com/GetStream/stream-chat-android/pull/3654

Issue #3571, opened 2022-05-23
Improve logs for QuerySort and its usage


leandroBorgesFerreira:
We should improve the logs for our sorting logic then it will be easier to debug problems when the sort doesn't happen as expected

PR #3624, opened 2022-05-30
Query sort by map

leandroBorgesFerreira:
Related ticket: https://github.com/GetStream/stream-chat-android/issues/3571

### 🎯 Goal

Provide a solution for sorting without using reflection as there are performance and reliability problems with it. 

### 🛠 Implementation details

- Making QuerySort as an interface so users can choose between `QuerySortByReflection` or `QuerySortByMap`. 
- Adding `QuerySortByMap` as a implementation for QuerySort. Using the sorter a user can only sort fields that implement the interface `QueryableByMap` and the class must provide a way to access its `Comparable<*>` fields as a map. This way there's no need to use reflection. There's a good performance gain because the order of complexity to get a property goes from O(N) to O(1) where N is the number of properties that a class being sorted has.

**Note: This PR creates breaking changes!**

### 🧪 Testing

Check the sorting of the SDK with and without internet. 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


docusaurus/docs/Android/02-client/06-guides/08-optimizations.mdx
```diff
@@ -22,3 +22,11 @@ Call<T>.forceNewRequest(): Call<T>
...
 
 
+## QuerySorter
+To sort your object and present then in a desired order, tt is possible to choose between 2 implementations of QuerySorter: `QuerySort` and `QuerySortByMap`. 
+
+`QuerySort` uses reflection to find the fields that should be used to compare when sorting. This avoids the need for extra implementation to make a class able to be sorted. 
+
+The drawback is that reflection is an expensive operation and this implementation will have a lower performance compared to `QuerySortByMap`.
+
+On the other hand, `QuerySortByMap` needs the class to be sorted to extend from `QueryableByMap` and provide the fields to be used in the comparation in a map. This implementation doesn't use reflection, so it will run faster.
```
adasiewiczr, line 32:
```suggestion
You can also choose to not use reflection and gain performance by using `QuerySortByMap` instead but this requires some extra work. Your custom class needs to implement the `QueryableByMap` interface and provide the names of the fields used to sort the list.
```

adasiewiczr: Let's also add something like this below:
```
:::note
Models included in the SDK already implement the `QueryableByMap` interface and include all the fields together with `extraData`!
:::
```

adasiewiczr: Can we also add an example of how to implement it? Maybe we can copy Channel, Message or any other model?

docusaurus/docs/Android/02-client/06-guides/08-optimizations.mdx
```diff
@@ -22,3 +22,11 @@ Call<T>.forceNewRequest(): Call<T>
 The returned `Call` will be forced to make a new request to API. 
 
 
+## QuerySorter
+To sort your object and present then in a desired order, tt is possible to choose between 2 implementations of QuerySorter: `QuerySort` and `QuerySortByMap`.
```
adasiewiczr, line 26:
```suggestion
You can choose between two QuerySorter implementations: (`QuerySort` and `QuerySortByMap`) to sort and present a list of objects (for example Channels) in the desired order.
```

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@
 - Added `EventHandlerSequential` to support a sequential event processing. [3604](https://github.com/GetStream/stream-chat-android/pull/3604)
 
 ### ⚠️ Changed
-
+- 🚨 Breaking change: QuerySort is now an interface and the user can choose between implementations that use reflection or not. [3624](https://github.com/GetStream/stream-chat-android/pull/3624)
```
adasiewiczr, line 37: We should update this part

docusaurus/docs/Android/02-client/06-guides/08-optimizations.mdx
adasiewiczr: We should also add a note which one is default below this sentence!

adasiewiczr (commented):
Looks great overall. I added some comments and I spotted most of the things don't have any KDocs. Can we add them? Can we also resolve conflicts? @amitkma @kanat @bychkovdmitry @MarinTolic  can you please do additional reviews? It's quite a change

kanat (commented):
Hey @leandroBorgesFerreira ,

I like the idea of getting rid of reflection.

However, I wouldn't say the `ByMap` approach converts `O(N)` into `O(1)`, cause each `customObject` has a finite number of properties, which in approximation can be treated as a constant. I mean the number of the defined properties in those classes does not depend on any input factor of N. The only property which can be increased is `customObject.extraData`, but as long as the `extraData` is behind the `Map` implementation it also has `O(1)`.

Probably the biggest concern I've got about `ByMap` implementation is the memory usage. It can be seen that each `compare` call produces 2 map instances (for each object to compare). See the picture below
Screenshot 2022-06-02 at 15 15 45


I created another implementation named `QuerySortByField`, which is not using a reflection (more readable & less side-effects) and it is still memory friendly.
PR is here: https://github.com/GetStream/stream-chat-android/pull/3632

leandroBorgesFerreira:
Kanat, I meant that we used to use a search for fields before, instead of a accessing the fields without any search. Example: 
A class with 10 fields would need a search of 10 fields in the worse case scenario, when the approach using `when` or `map` finds the field with only 1 interaction (no search is required), the more fields are added to the class (or the bigger it is), the slower the search will be... in case of the new implementation, the search with big or small classes it the same. 

That's only valid for the case when using `FieldNameSortAttribute`, because `FieldSortAttribute` doesn't search for fields, it also does only one interaction... But reflection is still slower.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Message.kt
```diff
@@ -196,9 +197,13 @@ public data class Message(
...
     var threadParticipants: List<User> = emptyList(),
-) : CustomObject {
+) : CustomObject, ComparableFieldProvider {
     public companion object {
         public const val TYPE_REGULAR: String = "regular"
         public const val TYPE_EPHEMERAL: String = "ephemeral"
     }
+
+    override fun getComparableField(fieldName: String): Comparable<*>? {
+        TODO("Not yet implemented")
```
leandroBorgesFerreira, line 207: Remember: Implement this tomorrow

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/querysort/QuerySortByReflection.kt
```diff
@@ -0,0 +1,178 @@
...
+import io.getstream.chat.android.client.models.CustomObject
+import io.getstream.logging.StreamLog
+import kotlin.jvm.internal.Reflection
+import kotlin.reflect.KClass
+import kotlin.reflect.KProperty1
+
+/**
+ * Sort specification for api queries. You can specify QuerySort by referencing kotlin class property or passing field
+ * name as string instance.
+ * QuerySort.asc(Channel::memberCount) and QuerySort.asc<Channel>("member_count") mean the same.
```
adasiewiczr, line 33: We should update this KDoc and missing ones to the public methods

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/querysort/IQuerySort.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.api.models.querysort
+
+public interface IQuerySort<T : Any> {
```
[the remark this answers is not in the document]
adasiewiczr: Let's also add missing KDocs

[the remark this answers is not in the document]
adasiewiczr: Let's also add missing KDocs

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/querysort/ComparableFieldProvider.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.api.models.querysort
+
+public interface ComparableFieldProvider {
```
adasiewiczr, line 19: Missing KDocs

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/querysort/BaseQuerySort.kt
```diff
@@ -0,0 +1,71 @@
...
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.api.models.querysort
+
+import io.getstream.chat.android.client.api.models.querysort.internal.CompositeComparator
+import io.getstream.chat.android.client.api.models.querysort.internal.SortAttribute
+import io.getstream.chat.android.client.api.models.querysort.internal.SortSpecification
+
+public abstract class BaseQuerySort<T : Any> : IQuerySort<T> {
```
adasiewiczr, line 23: Missing KDocs

adasiewiczr, line 23: Missing KDocs

adasiewiczr (commented):
Let's add missing KDocs.

@kanat can you also review this? Did you have an agreement on performance?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/querysort/IQuerySort.kt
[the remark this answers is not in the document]
leandroBorgesFerreira: We can have it named as "QuerySorter", "QuerySort", "IQuerySort" or something else. QuerySort would have the same name for the interface and the class, so we would have 2 different `QuerySort`. I personally like `QuerySorter`... `IQuerySort` is another option to avoid that the interface and class have the same name. Which do you prefer? Both are ok to me.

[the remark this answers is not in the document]
adasiewiczr: I don't like `I` prefix as it's more like a C# convention. `QuerySorter` sounds much better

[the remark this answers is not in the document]
kanat: Ideally, we should get rid of the existing `QuerySort`, cause it can be replaced by the new `QuerySortByReflection`.

We had 2 options:
1.  have both `class` & `interface` named `QuerySort` and resolve those by using the `as` keyword in kotlin. Customers are using a `class`, not an `interface`. So it's safe
2.  use a different name for `interface` until we delete the `class` version.

@leandroBorgesFerreira and I were discussing that, and just randomly picked up the name `IQuerySort`. 

P.S.
My personal preference is the option `1` to avoid massive changes in the future when we delete the `class` version and we will have to rename `IQuerySort` to `QuerySort`.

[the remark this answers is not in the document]
kanat: @adasiewiczr we agreed to use `QuerySortByField` as a default implementation.
It is space & time friendly in terms of complexity.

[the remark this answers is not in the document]
leandroBorgesFerreira: Let's use `QuerySorter` as decided in our daily.

kanat (approved):
LGTM

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/querysort/QuerySortByField.kt
```diff
@@ -0,0 +1,112 @@
...
+
+import io.getstream.chat.android.client.api.models.querysort.internal.SortAttribute
+import io.getstream.chat.android.client.api.models.querysort.internal.SortSpecification
+import io.getstream.chat.android.client.api.models.querysort.internal.compare
+import io.getstream.chat.android.client.extensions.snakeToLowerCamelCase
+import io.getstream.logging.StreamLog
+
+/**
+ * Implementation of QuerySorter for fields that implements `ComparableFieldProvider`. This QuerySorter doesn't
+ * use reflection and it will, be more performatic than `QuerySortByReflection`.
```
filbabic, line 27:
```suggestion
 * use reflection and it's more performant than `QuerySortByReflection`.
```

PR #3637, opened 2022-06-03
[3563] Add a typing update pruning mechanism

MarinTolic:
### 🎯 Goal

Implement a pruning mechanism for typing updates that automatically cancels typing events after a start typing event has expired.

### 🛠 Implementation details

- Create a start typing wrapper that automatically invokes a lambda upon completing a timed job
- Create a pruning class that keeps track of typing start events and removes any that have not been removed after a set time
- Introduce the pruning class into `ChannelStateLogicImplTest`

### 🎨 ~~UI Changes~~

No UI changes, only behavioral changes.

### 🧪 Testing

Use the following patch 

<details>

<summary> The following patch disables sending the stop typing updates from the client in both Compose and UI Components </summary>

```
Index: stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt	(revision 2751419db6eccefcedf14f7aacbc408d84abba9c)
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt	(date 1658501105184)
@@ -16,6 +16,7 @@
 
 package io.getstream.chat.android.ui.channel.list.viewmodel
 
+import android.util.Log
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.MutableLiveData
@@ -55,7 +56,9 @@
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.isActive
 
@@ -144,6 +147,12 @@
     private var queryChannelsState: StateFlow<QueryChannelsState?> = MutableStateFlow(null)
 
     init {
+        chatClient.globalState.typingChannels.onEach { typingChannels ->
+            val userChannelPairList =  typingChannels.map { Pair(it.value.channelId, it.value.users.map { it.id }) }
+
+            Log.d("currentlyTypingChannels", userChannelPairList.toString())
+        }.launchIn(viewModelScope)
+
         if (filter == null) {
             viewModelScope.launch {
                 val filter = buildDefaultFilter().first()
Index: stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt	(revision 2751419db6eccefcedf14f7aacbc408d84abba9c)
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt	(date 1658501070497)
@@ -16,6 +16,7 @@
 
 package io.getstream.chat.android.compose.viewmodel.channels
 
+import android.util.Log
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
@@ -50,7 +51,9 @@
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.filterNotNull
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
 
 /**
@@ -178,6 +181,12 @@
      * Makes the initial query to request channels and starts observing state changes.
      */
     private suspend fun init() {
+        chatClient.globalState.typingChannels.onEach { typingChannels ->
+            val userChannelPairList =  typingChannels.map { Pair(it.value.channelId, it.value.users.map { it.id }) }
+
+            Log.d("currentlyTypingChannels", userChannelPairList.toString())
+        }.launchIn(viewModelScope)
+
         logger.d { "Initializing ChannelListViewModel" }
 
         searchQuery.combine(queryConfigFlow) { query, config -> query to config }
Index: stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/typing/DefaultTypingUpdatesBuffer.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/typing/DefaultTypingUpdatesBuffer.kt b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/typing/DefaultTypingUpdatesBuffer.kt
--- a/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/typing/DefaultTypingUpdatesBuffer.kt	(revision 2751419db6eccefcedf14f7aacbc408d84abba9c)
+++ b/stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/utils/typing/DefaultTypingUpdatesBuffer.kt	(date 1658501105158)
@@ -108,7 +108,7 @@
         if (isTyping) {
             isTyping = false
         }
-        onTypingStopped()
+       // onTypingStopped()
     }
 
     /**
@@ -125,7 +125,7 @@
                 delay(DEFAULT_SEND_TYPING_UPDATES_INTERVAL)
             }
 
-            onTypingStopped()
+            //onTypingStopped()
         }
     }
 
Index: stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt b/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt	(revision 2751419db6eccefcedf14f7aacbc408d84abba9c)
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt	(date 1658501105124)
@@ -16,6 +16,7 @@
 
 package io.getstream.chat.android.offline.plugin.logic.channel.internal
 
+import android.util.Log
 import io.getstream.chat.android.client.api.models.QueryChannelRequest
 import io.getstream.chat.android.client.events.TypingStartEvent
@@ -207,6 +208,7 @@
      * @param event The start typing event or null if user stops typing.
      */
     override fun setTyping(userId: String, event: TypingStartEvent?) {
+        Log.d("SetTyping", event.toString())
         if (userId != clientState.user.value?.id) {
             typingEventPruner.processEvent(userId, typingStartEvent = event)
         }
@@ -224,6 +226,7 @@
     ) {
         mutableState.updateTypingEvents(eventsMap = rawTypingEvents, typingEvent = typingEvent)
         globalMutableState.tryEmitTypingEvent(cid = mutableState.cid, typingEvent = typingEvent)
+        Log.d("PrunerUpdated", rawTypingEvents.toString())
     }
 
     /**
```

</details>

Test scenario 1a:

1. Apply the given patch
2. Open the app on at least two devices with different users
3. Open the same channel and start typing from device A
4. Observe the changes on device B

Expected: The typing status should be visible and will be removed after ~7 seconds.

Test scenario 1b:

The following scenario involves tracking global typing updates via Logcat.

You can regex for: 

- "SetTyping" -> Here you can track every call to `ChannelStateLogicImpl.setTyping()`
- "PrunerUpdated" -> Here you can track every time the pruner emits an update
- "currentlyTypingChannels" -> Here you can track every time `globalMutableState.TypingChannels` emits.

1. Apply the given patch
2. Log in with different users on at least two devices
3. Apply one of the Logcat filters mentioned above
4. Start typing on preferably multiple devices on different channels

Expected: If you are tracking "currentlyTypingChannels" you should see every currently active channel (typing wise) **that the current user is a member of** with all available typing updates. Consequently each individual update should be cleaned after 7 seconds elapses and the event is considered stale.

Test Scenario 2: 

**Please do not apply the patch in this test scenario.**

1. Open the app on at least two devices with different users
3. Open the same channel and start typing from device A
6. Keep typing for around 10 seconds

Expected: The typing status should be visible for the full duration of the typing.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

giphy


adasiewiczr (commented):
Looks good overall. I added some suggestions for naming and small improvements.
@kanat can you take a look as well?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,217 @@
...
+     * Processes the incoming chat event accordingly.
+     *
+     * @param event The chat event to be processed.
+     */
+    override fun processEvent(event: ChatEvent) {
+        if (event is TypingStartEvent) {
+            processTypingStartEvent(event)
+        } else if (event is TypingStopEvent) {
+            processTypingStopEvent(event)
+        }
```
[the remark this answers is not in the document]
MarinTolic: Good suggestion. I'll add an exception here because passing other events by mistake here will leave them unprocessed an our regression testing should catch any of those.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/ChatEventCache.kt
```diff
@@ -0,0 +1,42 @@
...
+     * needs to be fired after a timer.
+     */
+    fun processEvent(event: ChatEvent)
+
+    /**
+     * Registers a lambda that should be
+     * invoked when an event has been processed
+     * and needs to be fired.
+     */
+    fun setOnEventFired(onEvent: (event: ChatEvent) -> Unit)
```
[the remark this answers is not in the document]
MarinTolic: Removed it.

It was honestly only used to signal that there should be a way to get a callback once an event has been processed. However you are right, that can be left to the implementation.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,217 @@
...
+ * of stale jobs.
+ * @param cleanStaleEventMs The time it takes before a [TypingStartEvent] is
+ * considered stale and should be "cleaned".
+ * @param onEventFired The lambda called after a typing event has been processed
+ * and cached if needed.
+ */
+internal class TypingEventCache(
+    private val coroutineScope: CoroutineScope = CoroutineScope(DispatcherProvider.IO),
+    private val cleanStaleEventMs: Long = EphemeralStartTypingEvent.DEFAULT_DELAY_TIME_MS,
+    private var onEventFired: (event: ChatEvent) -> Unit,
```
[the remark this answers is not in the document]
MarinTolic: Made it immutable.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,217 @@
...
+ * it performs is simple and short lived.
+ * @param typingStartEvent The event that needs to be wrapped and "cleaned"
+ * after a period of time.
+ * @param delayTimeMs The period of time it takes before the event is "cleaned".
+ * @param onDeath The lambda called when the stale typing event is "cleaned".
+ */
+private data class EphemeralStartTypingEvent(
+    private val coroutineScope: CoroutineScope,
+    private val typingStartEvent: TypingStartEvent,
+    private val delayTimeMs: Long = DEFAULT_DELAY_TIME_MS,
```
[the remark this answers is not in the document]
MarinTolic: It's references in multiple (2) places so I'd leave it as is if it's not a big deal to have it here.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * @param coroutineScope The coroutine scope used for timed cleaning
+ * of stale jobs.
+ * @param cleanStaleEventMs The time it takes before a [TypingStartEvent] is
+ * considered stale and should be "cleaned".
+ * @param onEventFired The lambda called after a typing event has been processed
+ * and cached if needed.
+ */
+internal class TypingEventCache(
+    private val coroutineScope: CoroutineScope = CoroutineScope(DispatcherProvider.IO),
```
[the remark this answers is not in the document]
MarinTolic: This is a good observation.

My idea was to expose this so that we could pass in a `CoroutineScope` at the instantiation site. However one is not directly available in `EventParser.kt`. I would be much happier if this scope was passed in.

I wanted to avoid unnecessary scope instantiation inside of `EphemeralStartTypingEvent` because these events could be plentiful and all of them should live either shorter lives or lives that are as long as the life of `TypingEventCache` but no longer so an outer scope would prevent unnecessary instantiation.

I've added a cancel function, but I don't think that's an answer to this situation because I'm not sure if there's a good candidate that we could use to trigger it.

Advice would be appreciated here.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * @param coroutineScope The coroutine scope used for timed cleaning
+ * of stale jobs. Does not use a default value by design because the job
+ * it performs is simple and short lived.
+ * @param typingStartEvent The event that needs to be wrapped and "cleaned"
+ * after a period of time.
+ * @param delayTimeMs The period of time it takes before the event is "cleaned".
+ * @param onDeath The lambda called when the stale typing event is "cleaned".
+ */
+private data class EphemeralStartTypingEvent(
```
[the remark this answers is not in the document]
MarinTolic: I've changed the name and made it more exact, although I'm still not sure if it's a good name.
This class is a bit difficult to name.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,217 @@
...
+    private val coroutineScope: CoroutineScope,
+    private val typingStartEvent: TypingStartEvent,
+    private val delayTimeMs: Long = DEFAULT_DELAY_TIME_MS,
+    private val onDeath: (TypingStopEvent) -> Unit,
```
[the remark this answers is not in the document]
MarinTolic: Changed the name to `onTypingStopEvent`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,234 @@
...
+        _currentlyTypingUsers.remove(keyValue)
+
+        // Signal that the event has been processed
+        onEventFired(typingStopEvent)
+    }
+
+    /**
+     * Cancels [coroutineScope].
+     */
+    fun cancel() {
```
adasiewiczr, line 129: It might not be needed if we decide to pass the same scope we are using in `OfflinePlugin` as it's already being canceled on user disconnect but let's wait for @kanat  input

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,234 @@
...
+                processTypingStopEvent(event)
+            }
+            else -> {
+                throw IllegalArgumentException(
+                    "This class should be only used to cache typing events. " +
+                        "All other events will remain unprocessed."
+                )
+            }
+        }
+    }
```
kanat, line 83:
```suggestion
    override fun processEvent(event: ChatEvent): Boolean {
        return when (event) {
            is TypingStartEvent -> {
                processTypingStartEvent(event)
                true
            }
            is TypingStopEvent -> {
                processTypingStopEvent(event)
                true
            }
            else -> false
        }
    }
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/EventsParser.kt
```diff
@@ -36,6 +39,13 @@ internal class EventsParser(
     private val logger = ChatLogger.get("Events")
     private var closedByClient = true
 
+    /**
+     * Processes typing events and "cleans" stale [TypingStartEvent]s
+     * by firing out their [TypingStopEvent] counterparts that hold the
+     * same important data, e.g. user, cid, .. etc
+     */
+    private val typingEventCache = TypingEventCache { event -> onEvent(event) }
```
kanat, line 47: `EventsParser` should not look like a good place to keep `typingEventCache` based on its name.
Would it be possible to move `typingEventCache` into the layers above?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/EventsParser.kt
```diff
@@ -100,7 +110,13 @@ internal class EventsParser(
...
                 }
             } else {
-                onEvent(event)
+                // Don't call onEvent() immediately, instead
+                // let the caching class decide when it needs to be called
+                if (event is TypingStartEvent || event is TypingStopEvent) {
+                    typingEventCache.processEvent(event)
+                } else {
+                    onEvent(event)
+                }
```
kanat, line 119: What if customers rely on `TypingStartEvent` to keep coming.
Shouldn't we just fire `TypingStopEvent` on timeout and not prevent `TypingStartEvent` from being passed further to avoid any breaking changes?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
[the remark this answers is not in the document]
kanat: Would be nice if we could link `TypingEventCache` to the socket's lifecycle.
Meaning that whenever the socket dies we cancel `TypingEventCache`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/EventsParser.kt
MarinTolic: That is essentially what is done [here](https://github.com/GetStream/stream-chat-android/pull/3637/files/cf3e25d9cceb470877d3c90539d6b60214bd6c82#diff-d1cedbe1ab0f946234b94756db4e441dc2fc5b7f712964c1941f9cc839f6f184R77).

You'll notice that the event is processed by basically getting registered and stored and then straight fired back out.

The reason why we are processing it is not to hijack it but rather to register it so we can time it's `TypingStopEvent` counterpart.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
```diff
@@ -0,0 +1,234 @@
...
+ * Used to provide key value equality for distinct typing events.
+ * A single value such as a [User] is not a good qualifier
+ * because a single user could type in multiple channels across
+ * the time span of a few seconds.
+ *
+ * @param user The user currently the typing event is tied to.
+ * @param cid The channel the the typing event is tied to.
+ */
+@InternalStreamChatApi
+internal data class TypingEventKeyValue(
```
filbabic, line 144: this is a very strange name for a class. Maybe TypingEventData?

Also - do we even need this class? both of the typing events (start, stop) have the same data as this, and we could just destructure it or format a string instead to use as a key. Something like: `userId:cid` can be the key.

Because it doesn't make sense - in the code, you transform both the Stop and Start events using `toStartTypingEventKeyValue`, which is confusing.

I'd just rather have a function:

```
fun typingEventToKey(user, cid) = "${user.id}:$cid"
```
or something,  than create a new class that does confusing transformations and then is used (as an object) as a key. Strings are much more flexible when it comes to keys and we can store any kind of information in them, nicely formatted, while being easier on the memory (roughly).

filbabic (commented):
Left some comments!

JcMinarro:
We shouldn't fire "fake events" based on some delay time hardcoded value on the SDK, we should always delegate it to backend.
I know we could have some network issues and lose some events, but we should handle it differently.
At the end it will be only "an UI Issue" if, by some reason, `StopTypingEvent` is not received. What we can do is filter `StartTypingEvent` on `ChannelMutableState` and only get them that has been done on the least X seconds.
By this approach no extra events are fired and we have a fallback on the case an `StopTypingEvent` is lost by any network error.
We only need to update `ChannelMutableState::typing` by something like:
```
override val typing: StateFlow<TypingEvent> = _typing
        .map { typingMap ->
            val dateFilter = Date(Date().time - 5000)
            val userList = typingMap.values
                .asSequence()
                .filter { dateFilter.before(it.createdAt)}
                .sortedBy(ChatEvent::createdAt)
                .mapNotNull { event ->
                    when (event) {
                        is TypingStartEvent -> event.user
                        else -> null
                    }
                }
                .toList()

            TypingEvent(channelId, userList)
        }
        .stateIn(scope, SharingStarted.Eagerly, TypingEvent(channelId, emptyList()))
```

MarinTolic:
Draftet so a team discussion can be had as there were differing opinions on how to proceed with this feature.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
```diff
@@ -0,0 +1,246 @@
...
+        ) -> Unit,
+    ) {
+        this.onUpdated = onUpdated
+    }
+
+    /**
+     * Clears all existing typing updates and posts
+     * the resulting updated lists.
+     */
+    fun clear() {
```
MarinTolic, line 187: Not used anywhere other than in a test currently, but it will be useful when https://github.com/GetStream/stream-chat-android/issues/3931 is implemented.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
```diff
@@ -348,7 +361,7 @@ internal class ChannelStateLogicImpl(
      * @param deleteDate The date when the channel was deleted.
      */
     override fun deleteChannel(deleteDate: Date) {
-        mutableState.setChannelData(mutableState.channelData.value?.copy(deletedAt = deleteDate))
+        mutableState.setChannelData(mutableState.channelData.value.copy(deletedAt = deleteDate))
```
MarinTolic, line 364: Same as: https://github.com/GetStream/stream-chat-android/pull/3637/files#r927443018

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/state/global/internal/GlobalMutableState.kt
```diff
@@ -215,8 +215,14 @@ public class GlobalMutableState private constructor(
 
     override fun tryEmitTypingEvent(cid: String, typingEvent: TypingEvent) {
         _typingChannelsOld.tryEmit(typingEvent)
+
         val typingChannelsCopy = _typingChannels.value.toMutableMap()
-        typingChannelsCopy[cid] = typingEvent
+
+        if (typingEvent.users.isEmpty()) {
```
MarinTolic, line 221: Previously we haven't removed keys for typing events that contained no active typing events.
This aims to fix that and not populate the map with keys that contain empty lists.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/factory/StreamStatePluginFactory.kt
```diff
@@ -135,15 +135,15 @@ public class StreamStatePluginFactory(
...
-                chatClient
-            )
+        val logic = LogicRegistry.create(
+            stateRegistry = stateRegistry,
+            globalState = globalState,
+            userPresence = config.userPresence,
+            repos = repositoryFacade,
+            client = chatClient,
+            clientState = clientState,
+            coroutineScope = scope,
```
MarinTolic, line 145: We currently don't cancel children of the scope anywhere, however this will be addressed here: https://github.com/GetStream/stream-chat-android/issues/3931

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/caching/TypingEventCache.kt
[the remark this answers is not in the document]
MarinTolic: Planned on doing so here: https://github.com/GetStream/stream-chat-android/issues/3931

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Initializes the pruner by passing in and setting
+     * the ID of the channel that the pruner is responsible for.
+     *
+     * @param channelId The ID of the channel the pruner is responsible for.
+     */
+    fun initialize(channelId: String) {
+        this.channelId = channelId
+    }
```
JcMinarro, line 75: `channelId` looks like it will be always the same (At least `initialize` name told me it should only be called on the initialization process), Why can't we inject it by constructor? :thinking:  It looks as a simple string that we should have before the `TypingEventPruner` is created

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Triggered when typing events have been updated.
+     *
+     * Use [setOnTypingEventsUpdatedListener] to set a listener externally.
+     */
+    private var onUpdated: (
+        rawTypingEvents: Map<String, TypingStartEvent>,
+        typingEvent: TypingEvent,
+    ) -> Unit = { _, _ -> }
```
JcMinarro, line 57: On the same way than the previous message, does it make sense to use a `TypingEventPruner` without an `OnTypingEventsUpdatedListener`? If not, why don't we inject this listener within the constructor?

stream-chat-android-state/src/test/java/io/getstream/chat/android/offline/internal/TypingEventPrunerTest.kt
```diff
@@ -0,0 +1,168 @@
...
+            val secondTypingStartEvent = randomTypingStartEvent(user = randomUser(id = "User2"))
+
+            typingEventPruner.processEvent(firstTypingStartEvent.user.id, firstTypingStartEvent)
+            typingEventPruner.processEvent(secondTypingStartEvent.user.id, secondTypingStartEvent)
+
+            // Time is advanced by slightly more than it takes for the pruner to remove stale events
+            advanceTimeBy(prunerDelay + 1)
+
+            typingEventPruner.getRawTyping() shouldBeEqualTo mapOf()
+            typingEventPruner.getTypingEvent() shouldBeEqualTo TypingEvent(channelId, listOf())
```
JcMinarro, line 124: Can't we "verify" over the `onUpdated` listener?
It should be the same, and we hide the internal implementation

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
MarinTolic: You are right, this is a leftover from a previous strategy where I thought initialization might look different.
It should be optimized.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
MarinTolic: You are right, this is a leftover from a previous strategy where I thought initialization might look different.
It should be optimized.

stream-chat-android-state/src/test/java/io/getstream/chat/android/offline/internal/TypingEventPrunerTest.kt
MarinTolic: Yes, that's a good idea, we can do that. :smile:

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
```diff
@@ -28,20 +33,67 @@ public class InitializationCoordinator private constructor() {
...
 
+    /**
+     * Notifies when the socket connection state has been changed.
+     *
+     * @param chatEvent The socket state change event. Pass in only the following event types:
+     * [ConnectedEvent], [ConnectingEvent], [DisconnectedEvent].
+     */
+    internal fun socketConnectionStateChanged(chatEvent: ChatEvent) {
+        val socketConnectionState: SocketConnectionState = when (chatEvent) {
+            is ConnectedEvent -> SocketConnectionState.CONNECTED
```
MarinTolic, line 70: Currently we only react to `SocketConnectionState.DISCONNECTED` inside `TypingEventsPruner`, however this might be useful in the future and there doesn't seem to be any real performance penalty to processing all connection state changes rather than just `DISCONNECTED.`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
```diff
@@ -28,20 +33,67 @@ public class InitializationCoordinator private constructor() {
...
+
+        onSocketConnectionStateChangedListener.forEach { function ->
+            function(socketConnectionState)
+        }
+    }
+
+    /**
+     * Used to represent socket connection events.
+     */
+    public enum class SocketConnectionState {
```
adasiewiczr, line 91: Do we need a different state for that? Maybe we can reuse `io.getstream.chat.android.client.models.ConnectionState`? 🤔

MarinTolic: Good suggestion, will change.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/state/internal/ClientStateImpl.kt
```diff
@@ -57,6 +58,9 @@ internal class ClientStateImpl(
         _initialized.value = false
         _connectionState.value = ConnectionState.OFFLINE
         _user.value = null
+
+        InitializationCoordinator.getOrCreate()
+            .removeSocketConnectionStateListeners()
```
MarinTolic, line 63: Remove the listeners when we clear state so that there are no stale listeners lying around.

kanat: Hey @MarinTolic 

Would it be possible to move this out of `ClientStateImpl`?
This is because, from my point of view, `ClientStateImpl.clearState()` is expected to clear the state's data only without calling any side-effect Singleton functions.

MarinTolic: @kanat 

I guess there are two ways that I see this being possible. One is to manually call it after `ClientState.clearState()` is called, wherever it is called. It's a sort of a roundabout way of doing it and we would have to be mindful of doing it in the future if anything changes, but it would get the job of removing it from the function done. 

The other one is a suggestion from @adasiewiczr to use `WeakReference` for the listeners in order to allow the GC to collect the listeners once individual instnaces of `ChannelStateLogicImpl` are removed from the heap.

I personally prefer going the explicit route of calling remove, but either way you think is the better way of doing this is fine by me.

Of course, if you have any alternative suggestions I'm open to them as well.

kanat: By `exposting` I meant that `clearState` should be an internal function.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/event/handler/internal/EventHandlerImpl.kt
```diff
@@ -134,7 +134,7 @@ internal class EventHandlerImpl(
             val initJob = scope.async {
                 syncManager.updateAllReadStateForDate(currentUser.id, Date())
                 syncManager.loadSyncStateForUser(currentUser.id)
-                replayEventsForAllChannels(currentUser)
+                replayEventsForAllChannels()
```
kanat, line 137: Hey @MarinTolic 

Do you mind reverting it back? It may cause merge conflict with changes, which touch Syncing & Event handling

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/event/handler/internal/EventHandlerImpl.kt
```diff
@@ -193,7 +193,7 @@ internal class EventHandlerImpl(
             }
     }
 
-    private suspend fun replayEventsForAllChannels(user: User) {
+    private suspend fun replayEventsForAllChannels() {
```
kanat, line 196: Same here (revert). I would appreciate it

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/state/internal/ClientStateImpl.kt
adasiewiczr: We can remove listeners in `ChatClient::disconnect`. 
 `ClientState::clear` should be internal indeed as someone might call it by mistake and will leave our SDK in a weird state. We should change that in a separate PR

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/event/handler/internal/EventHandlerImpl.kt
MarinTolic: Reverted :smile:

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/event/handler/internal/EventHandlerImpl.kt
MarinTolic: Reverted :smile:

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/state/internal/ClientStateImpl.kt
MarinTolic: Switched to removing listeners in `ChatClient::disconnect` as per suggestion.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
MarinTolic: Done :smile:

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
MarinTolic: Done :smile:

stream-chat-android-state/src/test/java/io/getstream/chat/android/offline/internal/TypingEventPrunerTest.kt
MarinTolic: Done :smile:

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
```diff
@@ -0,0 +1,226 @@
...
+ * @param delayTimeMs The period of time it takes before an individual stale
+ * typing event is removed.
+ * @param onUpdated Triggered when typing events have been updated.
+
+ */
+internal class TypingEventPruner(
+    private val channelId: String,
+    private val coroutineScope: CoroutineScope,
+    private val delayTimeMs: Long = DEFAULT_DELAY_TIME_MS,
+    initializationCoordinator: InitializationCoordinator = InitializationCoordinator.getOrCreate(),
```
JcMinarro, line 46: Why do you need this instance?
I can see you needed to create a new method to provide the current `ConnectionState`, but from the place where `TypingEventPruner` is created you have access to the `ClientState` that provides a property `connectionState` with the same value.
Can't we use it instead?
On the other hand, I can see it is used to stop previous launched job, but, Is it right?
Let me show you an example:
-. We receive a typing indicator from another user.
-. Our device switch from one network to another (There is an small disconnection)
-. Why did we clear typing indicators?

MarinTolic: I think we could use `ClientState`, and since `connectionState` is a `StateFlow` we could use `collectLatest` with a slight delay (100ms let's say) which would means that small interruptions to the connectivity should not clear the typing indicators.


Btw. eagerly clearing typing indicators isn't such a huge deal because in the current mode of operation, they are sent every three seconds for as long as someone is typing.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
```diff
@@ -28,20 +34,65 @@ public class InitializationCoordinator private constructor() {
 
     private val userDisconnectedListeners: MutableList<(User?) -> Unit> = mutableListOf()
 
+    /**
+     * Listeners which will get invoked when the web socket goes through connection changes.
+     */
+    private val onSocketConnectionStateChangedListener: MutableList<(ConnectionState) -> Unit> = mutableListOf()
```
filbabic, line 40:
```suggestion
    private val onSocketConnectionStateChangedListeners: MutableList<(ConnectionState) -> Unit> = mutableListOf()
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
```diff
@@ -28,20 +34,65 @@ public class InitializationCoordinator private constructor() {
...
+
     /**
      * Adds a listener to user disconnection.
      */
     public fun addUserDisconnectedListener(listener: (User?) -> Unit) {
         userDisconnectedListeners.add(listener)
     }
 
+    /**
+     * Adds a listener used to listen to socket connection changes
```
filbabic, line 50: Unfinished docs, missing param

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
```diff
@@ -0,0 +1,226 @@
...
+    private val onUpdated: (
```
filbabic, line 47: can we inline this?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
MarinTolic: This will be removed as per our slack discussion here: https://getstream.slack.com/archives/C02AA5LP0SK/p1659107529380459

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/setup/InitializationCoordinator.kt
MarinTolic: Same as here: https://github.com/GetStream/stream-chat-android/pull/3637#discussion_r934353431

stream-chat-android-state/src/test/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImplTest.kt
```diff
@@ -24,31 +24,40 @@ import io.getstream.chat.android.client.models.User
...
 import org.mockito.kotlin.verify
 import org.mockito.kotlin.whenever
 import java.util.Date
 
+@OptIn(ExperimentalCoroutinesApi::class)
 internal class ChannelStateLogicImplTest {
 
+    @Rule
+    val testCoroutines: TestCoroutineRule = TestCoroutineRule()
+
```
JcMinarro, line 60: We are mixing jUnit4 and jUnit5.
On this test-sWe are mixing jUnit4 and jUnit5.
On this test-suite we are using `org.junit.jupiter.api.Test` (jUnit5) and adding a `org.junit.Rule` (jUnit4).
Instead of that, we should use a `org.junit.jupiter.api.extension.ExtensionContext` (jUnit5) and in our code we have this one `TestCoroutineExtension`uite we are using `org.junit.jupiter.api.Test` (jUnit5) and adding a `org.junit.Rule` (jUnit4).
Instead of that, we should use a `org.junit.jupiter.api.extension.ExtensionContext` (jUnit5) and in our code we have this one `TestCoroutineExtension`

stream-chat-android-state/src/test/java/io/getstream/chat/android/offline/reactions/DeleteReactionsTests.kt
```diff
@@ -52,6 +54,9 @@ import org.mockito.kotlin.whenever
 @ExperimentalCoroutinesApi
 internal class DeleteReactionsTests {
 
+    @Rule
+    val testCoroutines: TestCoroutineRule = TestCoroutineRule()
+
```
JcMinarro, line 59: Same here

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
```diff
@@ -0,0 +1,210 @@
...
+        onUpdated(getRawTyping(), getTypingEvent())
+    }
+
+    /**
+     * Produces an up to date raw typing map.
+     *
+     * Used to set a value to
+     * [io.getstream.chat.android.offline.plugin.state.channel.internal.ChannelMutableState.rawTyping].
+     */
+    private fun getRawTyping(): Map<String, TypingStartEvent> = typingEvents.mapValues { it.value.typingStartEvent }
```
filbabic, line 132:
```suggestion
    private fun getRawTypingEvents(): Map<String, TypingStartEvent> = typingEvents.mapValues { it.value.typingStartEvent }
```

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
```diff
@@ -0,0 +1,210 @@
...
+ * @param delayTimeMs The period of time it takes before the event is removed.
+ * @param removeTypingEvent The lambda called when the stale typing event should be removed.
+ */
+@InternalStreamChatApi
+internal data class TimedTypingStartEvent(
+    private val coroutineScope: CoroutineScope,
+    internal val typingStartEvent: TypingStartEvent,
+    private val userId: String,
+    private val delayTimeMs: Long,
+    private val removeTypingEvent: (userId: String) -> Unit,
```
filbabic, line 185: can we inline this too?

filbabic (approved):
Left a few comments - from the UI/UX perspective, this looks great.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
MarinTolic: Decided to keep due to naming consistency with `ChannelMutableState.rawTyping`.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/TypingEventPruner.kt
MarinTolic: inlined

stream-chat-android-state/src/test/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImplTest.kt
MarinTolic: Switched to Junit5 extensions

stream-chat-android-state/src/test/java/io/getstream/chat/android/offline/reactions/DeleteReactionsTests.kt
MarinTolic: Switched to Junit5 extensions

JcMinarro (approved):

PR #3656, opened 2022-06-08
Fixing concurrency problems with unread count

leandroBorgesFerreira:
issue: https://github.com/GetStream/stream-chat-android/issues/3657

### 🎯 Goal

Fix race condition that makes count for unread messages be incremented twice. 

### 🛠 Implementation details

Adding a synchronized block to prevent multiple threads to access logic to increment data. 

### 🧪 Testing

It is a race condition hard to reproduce. We can check that the changes work with a customer that presented the problem. 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -399,34 +399,34 @@ internal class ChannelLogic(
      *
      * @param message [Message].
      */
-    internal fun incrementUnreadCountIfNecessary(message: Message) {
+    private fun incrementUnreadCountIfNecessary(message: Message) {
         val currentUserId = globalMutableState.user.value?.id ?: return
+        val unreadCount: Int = mutableState._unreadCount.value
```
leandroBorgesFerreira, line 404: Here I'm dealing with race conditions with immutability instead of locks. I don't pick the value of unreadCount **after** the `shouldIncrementUnreadCount` evaluation. So even if the change in the `unreadCount` happens twice, if will happen for the same value, because the snapshot of the data already happened.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/event/handler/internal/EventHandlerProvider.kt
```diff
@@ -16,19 +16,29 @@
...
      */
     internal var eventHandler: EventHandler
-        get() = _eventHandler ?: throw IllegalStateException("EventHandler was not set in the EventHandlerProvider. Looks like there's a initialisation problem")
+        get() = _eventHandler
+            ?: throw IllegalStateException("EventHandler was not set in the EventHandlerProvider. Looks like there's a initialisation problem")
         set(value) {
+            if (_eventHandler != null) {
+                logger.logE("EventHandlerProvider was initialized twice. We shouldn't happen, you may be initializing " +
+                    "the SDK twice. Stop listening the old EventHandler...")
+                _eventHandler?.stopListening()
```
leandroBorgesFerreira, line 45: If, for some reason, this is being initialized for the second time, avoid the previous connection to keep open.

kanat:
Hey @leandroBorgesFerreira,

I guess this issue is also can be fixed by applying an `EventHandlerSequential` which excludes the case of applying events in parallel from different threads.

kanat (approved):
LGTM

PR #3679, opened 2022-06-14
avoiding vals in companion objects

leandroBorgesFerreira:
### 🎯 Goal

Reduce the memory usage of the SDK.

### 🛠 Implementation details

Removing `val`s as they can't be collected by the garbage collector. Only `const val`s should be allowed to live inside companion objects. 

### 🎨 UI Changes

`develop` memory sage with app in background: 
Screenshot 2022-06-14 at 13 53 52

this branch memory sage with app in background: 
Screenshot 2022-06-14 at 13 51 20

### 🧪 Testing

Use the profiler 

#### General
- ~[ ] PR is linked to the GitHub issue it resolves~

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/ExtraDataValidator.kt
```diff
@@ -166,7 +166,7 @@ internal class ExtraDataValidator(
 
     private companion object {
 
-        private val reservedInChannel = arrayOf(
```
kanat, line 169: I would move it inside the class instead of having functions because we gonna produce that array on each API call which is getting validated

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/ExtraDataValidator.kt
```diff
@@ -180,23 +180,23 @@ internal class ExtraDataValidator(
             "config",
         )
 
-        private val reservedInMessage = arrayOf(
```
kanat, line 183: same here

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/ExtraDataValidator.kt
```diff
@@ -180,23 +180,23 @@ internal class ExtraDataValidator(
...
+        private fun reservedInMessage() = arrayOf(
             "id",
             "cid",
             "created_at",
             "updated_at",
             "deleted_at",
         )
 
-        private val reservedInUser = arrayOf(
```
kanat, line 191: same here

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/errors/ChatErrorCode.kt
```diff
@@ -61,14 +61,14 @@ public enum class ChatErrorCode(public val code: Int, public val description: St
...
+        private fun authenticationErrors() = setOf(
             AUTHENTICATION_ERROR_CODE,
             TOKEN_EXPIRED_ERROR_CODE,
             TOKEN_NOT_VALID_ERROR_CODE,
             TOKEN_DATE_INCORRECT_ERROR_CODE,
             TOKEN_SIGNATURE_INCORRECT_ERROR_CODE,
         )
 
-        public fun isAuthenticationError(code: Int): Boolean = authenticationErrors.contains(code)
+        public fun isAuthenticationError(code: Int): Boolean = authenticationErrors().contains(code)
```
kanat, line 72: We could replace it with something like this:

```suggestion
        public fun isAuthenticationError(code: Int): Boolean = when (code) {
            AUTHENTICATION_ERROR_CODE,
            TOKEN_EXPIRED_ERROR_CODE,
            TOKEN_NOT_VALID_ERROR_CODE,
            TOKEN_DATE_INCORRECT_ERROR_CODE,
            TOKEN_SIGNATURE_INCORRECT_ERROR_CODE -> true
            else -> false
        }
```

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
```diff
@@ -692,12 +692,12 @@ public class MessageComposerController(
         /**
          * The regex pattern used to check if the message ends with incomplete mention.
          */
-        private val MentionPattern = Pattern.compile("^(.* )?@([a-zA-Z]+[0-9]*)*$", Pattern.MULTILINE)
+        private fun mentionPattern() = Pattern.compile("^(.* )?@([a-zA-Z]+[0-9]*)*$", Pattern.MULTILINE)
```
kanat, line 695: Hey @leandroBorgesFerreira ,

I'm a bit concerned about this change.

Based on Pattern docs:
```
repeated matches it is less efficient since it does not allow the compiled pattern to be reused.
```

I think if that would be a problem Google didn't add static compiled Patterns at `android.util.Patterns`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/ExtraDataValidator.kt
leandroBorgesFerreira: Indeed, the companion object is unnecessary in this class

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
leandroBorgesFerreira: You're right, this can happen very often. What about moving this to the class instead of the `companion object`?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
```diff
@@ -355,6 +355,6 @@ public open class AvatarBitmapFactory(private val context: Context) {
         /**
          * Marker object to detect whether methods have been implemented by custom subclasses.
          */
-        private val NOT_IMPLEMENTED_MARKER = Bitmap.createBitmap(1, 1, Bitmap.Config.ALPHA_8)
+        private fun notImplementedMarker() = Bitmap.createBitmap(1, 1, Bitmap.Config.ALPHA_8)
```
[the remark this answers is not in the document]
leandroBorgesFerreira: The point of stopping using `val` that doesn't have `const` inside companion objects, because garbage collector won't ever collect them. Each object takes very little memory, but its aggregation can make a difference in the end. 

It doesn't have to be a function, we can also move the variable inside the class instead of being an static reference

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/SwipeViewHolder.kt
```diff
@@ -156,6 +156,6 @@ public abstract class SwipeViewHolder(itemView: View) : BaseChannelListItemViewH
     }
 
     private companion object {
-        private val SWIPE_THRESHOLD = 16.dpToPxPrecise()
+        private fun swipeThreshold() = 16.dpToPxPrecise()
```
kanat, line 159: I wonder why we cannot keep it here, cause it's Float primitive.
So in the end its gonna be converted into:
```
private static final float SWIPE_THRESHOLD = IntKt.dpToPxPrecise(16);
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/background/MessageBackgroundFactoryImpl.kt
```diff
@@ -152,23 +152,23 @@ public open class MessageBackgroundFactoryImpl(private val style: MessageListIte
...
     public companion object {
-        private val MESSAGE_OTHER_USER_BACKGROUND = R.color.stream_ui_white
-        private val MESSAGE_CURRENT_USER_BACKGROUND = R.color.stream_ui_grey_gainsboro
-        private val SMALL_CARD_VIEW_CORNER_RADIUS = 2.dpToPxPrecise()
+        private fun messageOtherUserBackground() = R.color.stream_ui_white
+        private fun messageCurrentUserBackground() = R.color.stream_ui_grey_gainsboro
+        private fun smallCardViewCornerRadius() = 2.dpToPxPrecise()
 
-        internal val DEFAULT_CORNER_RADIUS = 16.dpToPxPrecise()
+        internal fun defaultCornerRadius() = 16.dpToPxPrecise()
```
kanat, line 171: Didn't get why do we need to turn those static primitives into functions
```java
   private static final int MESSAGE_OTHER_USER_BACKGROUND;
   private static final int MESSAGE_CURRENT_USER_BACKGROUND;
   private static final float SMALL_CARD_VIEW_CORNER_RADIUS;
   private static final float DEFAULT_CORNER_RADIUS;
```

leandroBorgesFerreira: That's bugging me. `val` without const should be `private final`, but not static. Are you sure they'll become static values?

leandroBorgesFerreira: Moved to class variable

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/DateDividerViewHolder.kt
```diff
@@ -53,11 +53,11 @@ internal class DateDividerViewHolder(
...
         style.textStyleDateSeparator.apply(binding.dateLabel)
 
-        binding.dateLabel.background = ShapeAppearanceModel.Builder().setAllCornerSizes(DEFAULT_CORNER_RADIUS).build()
+        binding.dateLabel.background = ShapeAppearanceModel.Builder().setAllCornerSizes(defaultCornerRadius()).build()
             .let(::MaterialShapeDrawable).apply { setTint(style.dateSeparatorBackgroundColor) }
     }
 
     private companion object {
-        private val DEFAULT_CORNER_RADIUS = 16.dpToPxPrecise()
+        private fun defaultCornerRadius() = 16.dpToPxPrecise()
```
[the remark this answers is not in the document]
leandroBorgesFerreira: Moved as class variable.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/SwipeViewHolder.kt
leandroBorgesFerreira: Moved as class variable.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/avatar/AvatarBitmapFactory.kt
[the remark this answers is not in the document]
leandroBorgesFerreira: Moved as class variable

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
leandroBorgesFerreira: Moved to class variable

filbabic:
I've tested this with an emulator and two devices:

Emulator Pixel API 29

With optimizations
(Channels) 132.8 -> 136.6 (up to 147) (Messages)

around 135 MB of RAM for a base use case of channels/messages (after)

126 -> 134 MB -> 128 (before), GC kicks in and all

It seems like without optimizations things are better here 🤷 
--------------

Google Pixel 6 Pro Android 12 (😢)
307 MB in Channels
380MB in messages (powerless radio as JC, some images and messages)
Forcing GC pushes down to 330MB

Without optimizations
330MB in Channels
391MB in Messages (powerless radio as JC, some images and messages)
Forcing GC pushes down to 339MB

In this case it seems that we have ~7% improvements in channels (23MB down) and ~3% improvements in messages (11MB down), which is good! :]

----
Google Pixel 5

Without optimizations
215MB channels
252MB messages
GC drops to 203MB

With optimizations
220MB channels
255MB messages
GC drops to 205MB

Here the results seem to be more less the same, with the "without optimizations" part being slightly more performant.

I'm not sure if we actually achieved any optimizations or not, it can be very device specific and depend on the API version. Maybe @kanat @JcMinarro @bychkovdmitry or someone else can test on other devices and we can see if in average this is an improvement or not.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/internal/ChannelListItemAdapter.kt
```diff
@@ -37,16 +37,16 @@ internal class ChannelListItemAdapter(
...
 
     override fun onBindViewHolder(holder: BaseChannelListItemViewHolder, position: Int, payloads: MutableList<Any>) {
         val diff = (
             payloads
                 .filterIsInstance<ChannelListPayloadDiff>()
                 .takeIf { it.isNotEmpty() }
-                ?: listOf(FULL_CHANNEL_LIST_ITEM_PAYLOAD_DIFF)
-            ).fold(EMPTY_CHANNEL_LIST_ITEM_PAYLOAD_DIFF, ChannelListPayloadDiff::plus)
+                ?: listOf(fullChannelListItemPayloadDiff())
+            ).fold(emptyChannelListItemPayloadDiff(), ChannelListPayloadDiff::plus)
```
JcMinarro, line 49: Doesn't it decrease the performance of our `ChannelListView`???
Per every bind we will need to create, at least` one `emptyChannelListItemPayloadDiff`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/adapter/viewholder/SwipeViewHolder.kt
```diff
@@ -26,6 +26,11 @@ import kotlin.math.absoluteValue
 
 public abstract class SwipeViewHolder(itemView: View) : BaseChannelListItemViewHolder(itemView) {
 
+    /**
+     * Swipe threshold
+     */
+    private val swipeThreshold = SWIPE_THRESHOLD.dpToPxPrecise()
```
JcMinarro, line 32: Having it here will increase the memory used, because we are going to have one property per `SwipeViewHolder` instance

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/mention/list/viewmodel/MentionListViewModel.kt
```diff
@@ -39,14 +39,14 @@ public class MentionListViewModel : ViewModel() {
         val isLoading: Boolean,
     )
 
+    private val initialState = State(
```
JcMinarro, line 42: Do we really need this property?
Maybe we can add it to the initialization of `_state`, then, when the state is updated, it is recollected by GC

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/adapter/viewholder/internal/DateDividerViewHolder.kt
```diff
@@ -40,6 +40,8 @@ internal class DateDividerViewHolder(
     ),
 ) : DecoratedBaseMessageItemViewHolder<MessageListItem.DateSeparatorItem>(binding.root, decorators) {
 
+    private val defaultCornerRadius = DEFAULT_CORNER_RADIUS.dpToPxPrecise()
```
JcMinarro, line 43: Same here, the memory size will be increase because `DateDividerViewHolder` is used multiple times on our SDK

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/message/list/reactions/view/internal/ViewReactionsBubbleDrawer.kt
```diff
@@ -31,6 +31,8 @@ internal class ViewReactionsBubbleDrawer(
     private val viewReactionsViewStyle: ViewReactionsViewStyle,
 ) {
 
+    private val multipleReactionsBaselineOffset = MULTIPLE_REACTIONS_BASELINE_OFFSET.dpToPx()
```
JcMinarro, line 34: Same here

PR #3745, opened 2022-06-23
Fix connecting user in sample apps

adasiewiczr:
### 🎯 Goal

Fix connecting users in sample apps.

### 🛠 Implementation details

We don't need to connect a user if they were already connected.

### 🧪 Testing

1. Open a sample app
2. Go to bg
3. Send a message from another device
4. Tap on push notification
5. You should be able to see messages and app should be in connected state

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/login/UserLoginActivity.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/login/UserLoginActivity.kt
@@ -78,10 +78,8 @@ class UserLoginActivity : AppCompatActivity() {
                             // login screen then we need to reinitialize the SDK with our API key.
                             ChatHelper.initializeSdk(applicationContext, userCredentials.apiKey)
                         }
-                        ChatHelper.connectUser(
-                            userCredentials = userCredentials,
-                            onSuccess = ::openChannels,
-                        )
+                        ChatHelper.connectUser(userCredentials = userCredentials)
+                        openChannels()
                     },
                     onCustomLoginClick = ::openCustomLogin
                 )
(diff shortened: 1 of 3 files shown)


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/login/UserLoginActivity.kt
```diff
@@ -78,10 +78,8 @@ class UserLoginActivity : AppCompatActivity() {
                             // login screen then we need to reinitialize the SDK with our API key.
                             ChatHelper.initializeSdk(applicationContext, userCredentials.apiKey)
                         }
-                        ChatHelper.connectUser(
-                            userCredentials = userCredentials,
-                            onSuccess = ::openChannels,
-                        )
+                        ChatHelper.connectUser(userCredentials = userCredentials)
+                        openChannels()
```
kanat, line 82: Hey @adasiewiczr,

Shouldn't we wait for the successful result before opening Channels?

filbabic: not in the samples - as we want ot be able to log in offline

adasiewiczr: we won't get success if we are offline. We could either set the timeout, but then we'll have to block UI and don't allow clicking multiple times, or move forward without checking the result.
This behavior was aligned with XML, we might changed it in the future

JcMinarro: On our sample app it is safe because we are using static token that should be valid always without any expiration time neither become invalid at any point. So, as far as we connect the user, the SDK will handle all needed reconnection when the device is online again.

TGazica (approved):
### ☑️Reviewer Checklist

PR #3756, opened 2022-06-23
Fixing reply inside notification

leandroBorgesFerreira:
### 🎯 Goal

Fix reply inside notification

### 🛠 Implementation details

When there are no attachments to be sent, return a `Result.success` instead of `Result.error`. An error doesn't make sense in this case, because nothing wrong happens, there were just no attachments to be sent. This way it is possible to send messages without attachments even when the socket is not connected. 

### 🎨 UI Changes

_Add relevant screenshots_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/175353664-294061c6-c0ac-4da8-aef4-f3fa9a15a8b5.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/175353629-0da059c9-5792-4417-b3ea-6bc0faef7773.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

</details>

### 🧪 Testing

Reply messages using the notification

#### General
- [ ] ~PR is linked to the GitHub issue it resolves~. **This was found while running tests in the SDK.**

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (2)
_I need my reply!!_

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/interceptor/internal/SendMessageInterceptorImpl.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/interceptor/internal/SendMessageInterceptorImpl.kt
@@ -23,7 +23,6 @@ import io.getstream.chat.android.client.extensions.internal.hasPendingAttachment
 import io.getstream.chat.android.client.extensions.internal.populateMentions
 import io.getstream.chat.android.client.extensions.uploadId
 import io.getstream.chat.android.client.interceptor.SendMessageInterceptor
-import io.getstream.chat.android.client.logger.ChatLogger
 import io.getstream.chat.android.client.models.Attachment
 import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.persistance.repository.AttachmentRepository
@@ -62,7 +61,6 @@ internal class SendMessageInterceptorImpl(
     private var jobsMap: Map<String, Job> = emptyMap()
     private val uploadIds = mutableMapOf<String, UUID>()
 
-    private val logger = ChatLogger.get("MessageSendingService")
     override suspend fun interceptMessage(
         channelType: String,
         channelId: String,
@@ -96,7 +94,11 @@ internal class SendMessageInterceptorImpl(
         channelId: String,
     ): Result<Message> {
         val preparedMessage = prepareNewMessage(message, channelType, channelId)
-        return uploadAttachments(preparedMessage, channelType, channelId)
+        return if (preparedMessage.hasPendingAttachments()) {
+            uploadAttachments(preparedMessage, channelType, channelId)
+        } else {
+            Result.success(preparedMessage)
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/interceptor/internal/SendMessageInterceptorImpl.kt
```diff
@@ -173,19 +173,17 @@ internal class SendMessageInterceptorImpl(
...
 
-            message.hasPendingAttachments() -> {
+            !globalState.isOnline() && message.hasPendingAttachments() -> {
                 // We enqueue attachments upload here if user is offline but an error is returned so message is not sent right away.
                 enqueueAttachmentUpload(message, channelType, channelId)
                 Result(ChatError("Chat is offline, not sending message with id ${message.id} and text ${message.text}"))
             }
 
+            !message.hasPendingAttachments() -> Result.success(message.copy(type = Message.TYPE_REGULAR))
+
```
kanat, line 186: Should we check not only the WS connection but also the network availability then?
It seems like we should keep returning errors in the case when the mobile is offline.

kanat:
Is there any ticket related to this PR, just to understand where it comes from? :)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/interceptor/internal/SendMessageInterceptorImpl.kt
adasiewiczr: good call. I think this change is not so easy as we should separate socket and network state first

leandroBorgesFerreira: I think it would be nice to check both cases, there's no reason to stop the message if only the socket is disconnected and there is a connection available. When both socket and network states are separated, I think we should do that. 

I refactored the code to avoid `if` inside the `when`, but the logic for the cases is the same, the only real addition is:
```
!message.hasPendingAttachments() -> Result.success(message.copy(type = Message.TYPE_REGULAR))
```

I think we should return `succcess` when there's no attachment to be sent. Otherwise, we are sending a failure for something that doesn't have to be done =|. If we return `success` we allow the flow of sending a message to continue, because it will be saved in the database and it will be sent if there's a connection because we will make a rest call in the `ChatClient`.

leandroBorgesFerreira:
ticket added

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/interceptor/internal/SendMessageInterceptorImpl.kt
leandroBorgesFerreira: But now checking, the `else` case should be removed in this code...

adasiewiczr: as discussed on Slack: let's leave it as is and improve in the future after completing #3795

adasiewiczr (approved):
LGTM

PR #3772, opened 2022-06-27
[3529 & 3770] Optimize calls to mark read 

### LINKED ISSUE #3529: Repeated calls to `MessageListViewModel.updateLastSeenMessageState()` when a message fails to send
### Context & proposal

If a message fails to send, Compose will make repeated calls to the method mentioned in the title.

In the logs you will notice the following:

```
 I/Chat:ChannelMarkReadHelper: Last message date [null] is not after last read event [Tue May 17 13:17:21 GMT+02:00 2022]; no need to update.
 ```
 
 **Note**: The failed message has to be in view for this to happen

### Acceptance criteria

- [ ] Fix the issue

### LINKED ISSUE #3770: The channel is marked as read too early
### Context & proposal

Instead of marking the channel as read when the reader reaches the most recent item, compose will mark the channel as unread when it reaches the first unread message item.

Theoretically, the reader could be n items away from the most recent message and the channel could be marked as read.

### Acceptance criteria

- [ ] Mark the channel as read only when the most recent item has been reached.

### PR DESCRIPTION
**Pull request by MarinTolic**
### 🎯 Goal

closes #3529
closes #3770 

### 🛠 Implementation details

#3529

Extracted marking as read into a separate function that triggers a lambda when a new item is visible. 

The benefit of this is that only the newly visible item will call mark as read, where as previously any recomposed item would call mark as read.

#3770

Compare the IDs of the messages that are marked as read, marking the channel as read only when the last item is reached.

### 🧪 Testing

#3529 

#### on `develop`:

1. Log in as Marton
2. Enter a channel named `Marton is banned here`
3. Send a few messages (you need at least two failed messages in order for compose to enter a loop)
4. Observe Logcat

Result: Compose will enter a recomposition loop and start making repeated unnecessary calls to mark read

#### on the branch featured in this PR:

1. Log in as Marton
2. Enter a channel named `Marton is banned here`
3. Send a few messages (you need at least two failed messages in order for compose to enter a loop)
4. Observe Logcat

Result: No vicious loop entered.

**Note**:  Since this also touches upon functionalities such as focusing on a message, etc, please also test scenarios such as:

- Navigating to a quoted message
- Opening a push notification
- Other similar actions you can think of

#3770

#### on `develop`: 

1. Launch the app on two separate devices while logging in with two different users
2. Observe the same channel with both users
3. Using the Observer user (user A), scroll up so that you can no longer see the first 5 or 6 messages
4. Using the Poster user (User A), post many messages (it's best to number up or down,  making the individual message text such as 10, 9, 8...)
5. Slowly scroll down using the Observer user and notice when the channel gets marked as read on the Poster user's app

Result: The channel will get marked as read when it hits the first visible (not the most recent) message

#### on the branch featured in this PR:

1. Launch the app on two separate devices while logging in with two different users
2. Observe the same channel with both users
3. Using the Observer user (user A), scroll up so that you can no longer see the first 5 or 6 messages
4. Using the Poster user (User A), post many messages (it's best to number up or down,  making the individual message text such as 10, 9, 8...)
5. Slowly scroll down using the Observer user and notice when the channel gets marked as read on the Poster user's app

Result: The channel will get marked as read when it hits the most recent message

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

catread

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
@@ -94,8 +94,9 @@ public fun Messages(
     val density = LocalDensity.current
 
     /** Marks the bottom most item as read every time it changes. **/
-    OnLastVisibleItemChanged(lazyListState) {
-        val message = messagesState.messageItems[it]
+    OnLastVisibleItemChanged(lazyListState) { messageIndex ->
+        val message = messagesState.messageItems[messageIndex]
+
         if (message is MessageItemState) {
             onLastVisibleMessageChanged(message.message)
         }
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
@@ -95,7 +95,7 @@ public fun Messages(
 
     /** Marks the bottom most item as read every time it changes. **/
     OnLastVisibleItemChanged(lazyListState) { messageIndex ->
-        val message = messagesState.messageItems[messageIndex]
+        val message = messagesState.messageItems.getOrNull(messageIndex)
 
         if (message is MessageItemState) {
             onLastVisibleMessageChanged(message.message)
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
@@ -93,6 +93,14 @@ public fun Messages(
     var parentSize by remember { mutableStateOf(IntSize(0, 0)) }
     val density = LocalDensity.current
 
+    /** Marks the bottom most item as read every time it changes. **/
+    OnLastVisibleItemChanged(lazyListState) {
+        val message = messagesState.messageItems[it]
+        if (message is MessageItemState) {
+            onLastVisibleMessageChanged(message.message)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt [hunk 1]
```diff
@@ -535,7 +542,7 @@ public class MessageListViewModel(
         val cid = channelState.value?.cid
         if (cid == null || chatClient.globalState.isOffline()) return
 
-        chatClient.loadMessageById(cid, message.id, DefaultMessageLimit, DefaultMessageLimit).enqueue()
```
MarinTolic, line 538: This was newly deprecated.
It should be replaced in XML too but I didn't want to litter this PR.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -93,6 +93,15 @@ public fun Messages(
     var parentSize by remember { mutableStateOf(IntSize(0, 0)) }
     val density = LocalDensity.current
 
+    /** Marks the bottom most item as read every time it changes. **/
+    OnLastVisibleItemChanged(lazyListState) { messageIndex ->
```
MarinTolic, line 97: One possible consequence I can think of here is having an out of bounds error but brief torture tests did not manage to cause it.

If you can think of a scenario, please test it.

bychkovdmitry (approved):
LGTM

bychkovdmitry (changes requested):
Unread count badge is now broken

bychkovdmitry (approved):

PR #3807, opened 2022-06-29
Refactor/call implementation

### LINKED ISSUE #3761: Refactor `Call` implementation to run `away() properly and be able to cancel them
The current implementation of `Call` interface doesn't provide a method to run `away()` properly and be able to be canceled on the case the `Call` is canceled.
There is an extension function over `Call` interface that simulates an `away()` method, but it is not using the current coroutineScope injected on some `Call` implementations, or it is using a "continuation" approach that could generate some deadlock if it is not used properly.

Because of that, we need to implement a proper `away()` implementation on every `Call` subclass that allows us to run it in a suspend function.

JcMinarro:
### 🎯 Goal
Fix #3761
Implement proper `await()` implementation to our `Call` subclasses

### 🛠 Implementation details
-. Add a new `Call::await` suspend function to our `Call` interface
-. Implement it on every `Call` subclass 
-. Refactor the behavior of every `Call` subclass to be able to be canceled and suspend future iterations
-. Add tests for all of our `Call` implementation classes

#### General

#### Code & documentation

### ☑️Reviewer Checklist


TGazica (approved):
Looks good from ui side

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/DistinctCall.kt
```diff
@@ -105,4 +107,8 @@ internal class DistinctCall<T : Any>(
     private companion object {
         private const val TAG = "Chat:DistinctCall"
     }
+
+    override suspend fun await(): Result<T> = withContext(DispatcherProvider.IO) {
+        execute()
+    }
```
kanat, line 113: Hey @JcMinarro,

Can we turn `await` into the way of using `suspendCancellableCoroutine`:
```suggestion
    override suspend fun await(): Result<T> {
        StreamLog.d(TAG) { "[await] uniqueKey: $uniqueKey" }
        return suspendCancellableCoroutine { continuation ->
            enqueue { result ->
                StreamLog.v(TAG) { "[await] completed($uniqueKey)" }
                continuation.resume(result)
            }
            continuation.invokeOnCancellation {
                StreamLog.v(TAG) { "[await] canceled($uniqueKey)" }
                delegate.get()?.cancel()
            }
        }
    }
```

and replace `execute` with the following snippet:
```kotlin
    override fun execute(): Result<T> = runBlocking {
        await()
    }
```

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -32,24 +35,23 @@ public class CoroutineCall<T : Any>(
...
+        canceled.set(true)
         job?.cancel()
     }
 
-    override fun execute(): Result<T> {
-        return runBlocking(block = suspendingTask)
-    }
-
     override fun enqueue(callback: Call.Callback<T>) {
+        println("[JcLog] CoroutineCall::enqueue")
```
kanat, line 51: Is this something to be deleted?

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/DoOnStartCall.kt
```diff
@@ -29,20 +34,35 @@ internal class DoOnStartCall<T : Any>(
...
+    private val canceled = AtomicBoolean(false)
 
-    override fun execute(): Result<T> = runBlocking {
-        sideEffect()
-        originalCall.execute()
-    }
+    override fun execute(): Result<T> = runBlocking { await() }
 
     override fun enqueue(callback: Call.Callback<T>) {
+        println("[JcLog] DoOnStartCall::enqueue")
```
kanat, line 42: Same here.

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -32,24 +35,23 @@ public class CoroutineCall<T : Any>(
...
         job = scope.launch {
-            val result = suspendingTask()
+            val result = await()
+            yield()
             withContext(DispatcherProvider.Main) {
                 callback.onResult(result)
             }
```
kanat, line 57: Should we check that the call is not canceled, cause we might have switched the thread?
```suggestion
            withContext(DispatcherProvider.Main) {
                if (!canceled.get()) {
                    callback.onResult(result)
                }
            }
```

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/CoroutineCall.kt
```diff
@@ -32,24 +35,23 @@ public class CoroutineCall<T : Any>(
...
-    override fun execute(): Result<T> {
-        return runBlocking(block = suspendingTask)
-    }
-
     override fun enqueue(callback: Call.Callback<T>) {
+        println("[JcLog] CoroutineCall::enqueue")
         job = scope.launch {
-            val result = suspendingTask()
+            val result = await()
+            yield()
```
kanat, line 54: Do we need `yield` here, since the next step is about to switch the context(thread)?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/RetryCall.kt
```diff
@@ -41,25 +44,32 @@ internal class RetryCall<T : Any>(
...
-            val result = callRetryService.runAndRetry {
-                originalCall
-            }
+            val result = await()
+            yield()
             withContext(DispatcherProvider.Main) {
                 callback.onResult(result)
             }
         }
     }
```
kanat, line 59: Should we check a cancellation scenario here?
```kotlin
    override fun enqueue(callback: Call.Callback<T>) {
        if (canceled.get()) {
            callback.onResult(Result.error(ChatError("The call was canceled before its execution")))
            return
        }
        scope.launch {
            if (canceled.get()) {
                callback.onResult(Result.error(ChatError("The call was canceled before complete its execution")))
                return@launch
            }
            val result = await()
            yield()
            withContext(DispatcherProvider.Main) {
                if (!canceled.get()) {
                    callback.onResult(result)
                }
            }
        }
    }
```

kanat: Not sure if we need to omit callback calls on the cancellation.
I guess if users cancel a call they might expect silence from the callback 🤔

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/RetryCall.kt
```diff
@@ -41,25 +44,32 @@ internal class RetryCall<T : Any>(
 ) : Call<T> {
 
     private var job: Job? = null
```
kanat, line 46: `job` is never set.

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/DoOnStartCall.kt
```diff
@@ -29,20 +34,35 @@ internal class DoOnStartCall<T : Any>(
...
             sideEffect()
-            originalCall.enqueue(callback)
+            yield()
+            originalCall.enqueue {
+                if (!canceled.get()) {
+                    scope.launch(DispatcherProvider.Main) { callback.onResult(it) }
+                }
+            }
         }
     }
```
kanat, line 52: What do you think about the following version? 
```kotlin
    override fun enqueue(callback: Call.Callback<T>) {
        if (canceled.get()) {
            callback.onResult(Result.error(ChatError("The call was canceled before its execution")))
            return
        }
        job = scope.launch {
            sideEffect()
            yield()
            originalCall.enqueue {
                if (!canceled.get()) {
                    scope.launch(DispatcherProvider.Main) {
                        if (!canceled.get()) {
                            callback.onResult(it)
                        }
                    }
                }
            }
        }
    }
```

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/ReturnOnErrorCall.kt
```diff
@@ -35,27 +38,42 @@ public class ReturnOnErrorCall<T : Any>(
...
+    override fun execute(): Result<T> = runBlocking { await() }
 
     override fun enqueue(callback: Call.Callback<T>) {
-        originalCall.enqueue { originalResult ->
-            if (originalResult.isSuccess) callback.onResult(originalResult)
-            else job = scope.launch {
-                val result = onErrorReturn(originalResult.error())
-                withContext(DispatcherProvider.Main) {
-                    callback.onResult(result)
+        scope.launch {
```
kanat, line 46: It seems like there is no way to cancel the first `launch`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/RetryCall.kt
JcMinarro: If it is canceled, we don't want to be notified on the callback.
Imagine the callback is being executed into an activity that is already destroyed, we don't want this code arrives there after we cancel the `Call`

JcMinarro: About to check the cancelation excenario, it should be done with the `yield()`, but as you catch on previous message, we are not storing the job, so it is not canceled :facepalm: 
Let me fix it

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/RetrofitCall.kt
```diff
@@ -40,83 +46,78 @@ internal class RetrofitCall<T : Any>(
...
 
     override fun enqueue(callback: Call.Callback<T>) {
-        enqueue(call) {
-            if (!canceled.get()) {
-                callback.onResult(it)
+        scope.launch {
+            enqueue(call) { result ->
+                scope.launch { notifyResult(result) { callback.onResult(it) } }
             }
         }
```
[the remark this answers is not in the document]
JcMinarro: `notifyResult()` method already checks if the `Call` is already canceled.

kanat, line 56: Hey @JcMinarro,

Since we already use `awaitResponse` which converts callback style into a suspendable coroutine I wonder if we can use `await` here instead of invoking `enqueue` of child call

skydoves: I agree. Retrofit provides some Kotlin extensions such as `awaitResponse`, so we can keep the Coroutines style with those extensions and `try-catch` expression.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/RetrofitCall.kt
```diff
@@ -40,83 +46,78 @@ internal class RetrofitCall<T : Any>(
...
-    }
+    override fun execute(): Result<T> = runBlocking { await() }
 
     override fun enqueue(callback: Call.Callback<T>) {
-        enqueue(call) {
-            if (!canceled.get()) {
-                callback.onResult(it)
+        scope.launch {
+            enqueue(call) { result ->
+                scope.launch { notifyResult(result) { callback.onResult(it) } }
```
skydoves, line 54: What do you think about removing the duplicated `scope.launch` by changing the `callback` as a suspend function like the below?

```kotlin
private suspend fun enqueue(call: retrofit2.Call<T>, callback: suspend (Result<T>) -> Unit)
private suspend fun notifyResult(result: Result<T>, callback: suspend (Result<T>) -> Unit)
private suspend fun notifyResult(result: Result<T>, callback: suspend (Result<T>) -> Unit) =
        withContext(DispatcherProvider.Main) {
            result.takeUnless { canceled.get() }?.let { callback(it) }
        }

```

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/BlockedCall.kt
```diff
@@ -0,0 +1,70 @@
...
+import io.getstream.chat.android.client.utils.Result
+import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+import java.util.concurrent.atomic.AtomicBoolean
+
+public class BlockedCall<T : Any>(private val result: Result<T>) : Call<T> {
```
skydoves, line 29: What do you think about marking this, `BlockedRetrofit2Call`, and `BlockedTask` with `@InternalStreamChatApi` annotation?

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/MapCall.kt
```diff
@@ -16,45 +16,41 @@
...
-            if (it.isSuccess) {
-                val data = mapper(it.data())
-                callback.onResult(Result(data))
-            } else {
-                val error = it.error()
-                callback.onResult(Result(error))
-            }
-        }
+    override suspend fun await(): Result<K> = withContext(DispatcherProvider.IO) {
+        call.await()
```
skydoves, line 50: Just out of my curiosity, is there any reason that we don't use the injected coroutine scope for `MapCall` and `ZipCall`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/RetrofitCall.kt
JcMinarro: I am just simplifying it ;)

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/MapCall.kt
JcMinarro: We are not injecting the scope because it will complicate how the extensions to "map" or "zip" calls are used.
Currently we do something like that:
```
val callA: Call<Int> = [...] 
val callB: Call<String> = callA.map { "$it" }
```
Maybe from this part is not easy to access to and inject the proper coroutine scope

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/DistinctCall.kt
JcMinarro: This class has been refactored to follow the same approach than the rest of implementations

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/ReturnOnErrorCall.kt
JcMinarro: Cancel the first launch will do nothing, because inside of this coroutine we only enqueue the inner call.
BTW, the `originalCall` is cancelled, and when `enqueue()` is used, no `callback.onResult()` is invoked, so second job won't be created

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/call/RetrofitCall.kt
JcMinarro: Already updated

skydoves (approved):
Totally looks great to me! 💯 

### ☑️Reviewer Checklist

PR #3818, opened 2022-06-30
[3727] Compose bi directional pagination

TGazica:
### 🎯 Goal

Add bi-directional pagination to the Compose SDK with scroll to selected message and scroll to bottom.

### 🛠 Implementation details

Updated `ChannelLogic` which determines if the pagination has reached either end or start of the messages list. Also added checks when receiving or sending new messages, so that the new message isn't added to the list if the newest messages are not loaded to avoid breaking pagination of newer messages.

Introduced new pagination handler for reaching the end of new messages to `MessageList` composable. Also added `ScrollToPositionState` inside `MessagesState` which notifies if we are loading the focused messages or the newest messages if the user wants to scroll to the bottom of the list, or if we should scroll to the focused message or bottom of the list. Added scroll to bottom handler to `MessageList`.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/38032787/176719148-72f40164-bf9b-433e-9f5e-190f44cdc6ea.mp4" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/38032787/176719135-254b3323-06c0-4064-9f9a-c2bba4cf9157.mp4" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

1. Build this branch and enter a channel that has more than one page
2. Scroll n number of pages and reply to a message
4. Exit the channel and enter it again, the message history will be reset
5. Click on the quoted message which should load the message and a page that surrounds the message
6. Scroll up and down to ensure that the bi directional pagination is working
7. Try sending messages from the current user and from a different device, messages shouldn't be added to the bottom of the list, otherwise pagination will break
8. Click on scroll to bottom button, if the newest messages aren't present they should be loaded and the list should scroll to the bottom

Test that the ui-components sdk still has functioning two way pagination since there were changes how the end of list is determined. Scroll to bottom will come in a separate pr.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

giphy


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -264,6 +280,10 @@ internal class ChannelLogic(
         return runChannelQuery(aroundIdWatchChannelRequest(aroundMessageId))
     }
 
+    internal suspend fun loadNewestMessages(limit: Int): Result<Channel> {
```
adasiewiczr, line 283: The implementation looks almost the same as a `watch`. Can we reuse it? 🤔

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -394,7 +414,15 @@ internal class ChannelLogic(
         repos.insertMessages(messages)
     }
 
-    internal fun upsertMessage(message: Message) = upsertMessages(listOf(message))
+    internal fun upsertMessage(message: Message) {
+        /* Do not upsert message if the newest messages are not loaded to avoid breaking pagination. If the message
+        * is inside the list and was updated in some kind of way, update it in the list. */
+        if (mutableState._endOfNewerMessages.value ||
+            mutableState._messages.value.containsKey(message.id)
```
adasiewiczr, line 421: This change might introduce unexpected behavior and we need to test it carefully.
Also, I'm not sure if we should do this check here:
1. Why shouldn't we upsert one message if it's not present in the state but have different behavior for a list of message?
2. Rather an edge case but what will happen if I already watch a channel (so I can receive events) or send messages but the `_endOfNewerMessages ` value was not updated yet?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelRequest.kt
```diff
@@ -93,6 +93,10 @@ public open class QueryChannelRequest : ChannelRequest<QueryChannelRequest> {
         return keys.contains(Pagination.AROUND_ID.toString())
     }
 
+    public fun isFilteringMessages(): Boolean {
+        return messages.keys.any { key -> Pagination.values().map { it.toString() }.contains(key) }
```
adasiewiczr, line 97: What do you think about replacing it with:
```Pagination.values().map { it.toString() }.fold(false) { currentValue, paginationKey -> currentValue || messages.keys.contains(paginationKey) }```?
It's not a big performance improvement but we avoid iterating through `messages.keys` and `Pagination.values()` mapping on every iteration

File: stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt (hunk 2 above)
Comment (TGazica) [reply stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt:421]: 1. If we click on a quoted message and it load n pages away, the newest messages will not be inside the list. The `_endOfNewerMessages` is here to safeguard that we do not insert any new messages that are received to the list. If we add them to the bottom it would break pagination (n number of pages would be missing). The contains ensures that if the messages is contained in the list, any action that modifies it will be propagated (edit, reactions) and will be visible in the list.
2. This could happen yes, if we are initially loading the list, until the call finishes the boolean will be false. But if we have a boolean that checks whether the initial loading is taking place, we can upsert newer messages.

If this kind of safeguard makes more sense elsewhere we can discuss where to move it.

File: stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelRequest.kt (hunk 3 above)
Comment (TGazica) [reply stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelRequest.kt:97]: Yes, this works.

Comment (filbabic) [reply stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelRequest.kt:97]: Could we also try to intersect the two sets of keys and then see if the intersection isn't empty?

File: stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/MessagesState.kt [hunk 4]
```diff
@@ -45,13 +49,14 @@ public data class MessagesState(
     val newMessageState: NewMessageState? = null,
     val parentMessageId: String? = null,
     val unreadCount: Int = 0,
+    val startOfMessages: Boolean = false,
```
filbabic, line 52: missing docs param for startOfMessages

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/ScrollToPositionState.kt
```diff
@@ -0,0 +1,77 @@
...
+import kotlinx.coroutines.flow.StateFlow
+
+/**
+ * Determines if the list is loading data to prepare for scroll to a certain part of the list or is currently scrolling
+ * to it.
+ */
+public sealed class ScrollToPositionState
+
+/**
+ * State when the focused message is not in the current list and that the data is being loaded form the API.
```
filbabic, line 30:
```suggestion
 * State when the focused message is not in the current list and the data is being loaded from the API.
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelRequest.kt
adasiewiczr: good idea!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -82,15 +84,28 @@ public fun Messages(
     onMessagesStartReached: () -> Unit,
     onLastVisibleMessageChanged: (Message) -> Unit,
     onScrolledToBottom: () -> Unit,
+    onMessagesEndReached: (String) -> Unit,
```
filbabic, line 87: These are a breaking change, we should duplicate this function without these params and then just kind of disable the new functionality in older component alternatives, while deprecating it

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -82,15 +84,28 @@ public fun Messages(
...
-    val (_, isLoadingMore, endOfMessages, messages) = messagesState
+    val messages = messagesState.messageItems
+    val endOfMessages = messagesState.endOfMessages
+    val startOfMessages = messagesState.startOfMessages
+    val isLoadingMoreNewMessages = messagesState.isLoadingMoreNewMessages
+    val isLoadingMoreOldMessages = messagesState.isLoadingMoreOldMessages
+    val scrollToPositionState = messagesState.scrollToPositionState
 
-    var parentSize by remember { mutableStateOf(IntSize(0, 0)) }
+    var parentSize = remember { IntSize(0, 0) }
```
filbabic, line 108: Why isn't this wrapped with a state?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -197,43 +234,57 @@ internal fun BoxScope.DefaultMessagesHelperContent(
 
     val offset = messagesState.focusedMessageOffset.collectAsState()
```
filbabic, line 235: if we add `by` here, we won't have to use `offset.value`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -197,43 +234,57 @@ internal fun BoxScope.DefaultMessagesHelperContent(
...
+        if (scrollToStartState == ScrollToNewestMessages) {
+            lazyListState.scrollToItem(0)
+        }
+
         when {
-            !lazyListState.isScrollInProgress && newMessageState == Other &&
-                firstVisibleItemIndex < 3 -> coroutineScope.launch {
-                lazyListState.animateScrollToItem(0)
+            focusedItemIndex == -1 &&
+                !lazyListState.isScrollInProgress &&
```
filbabic, line 254: This is extremely complex - can we pull this out in a function and then call this value something meaningful?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -197,43 +234,57 @@ internal fun BoxScope.DefaultMessagesHelperContent(
...
+                coroutineScope.launch {
+                    lazyListState.animateScrollToItem(0)
+                }
             }
 
-            !lazyListState.isScrollInProgress && newMessageState == MyOwn -> coroutineScope.launch {
-                if (firstVisibleItemIndex > 5) {
-                    lazyListState.scrollToItem(5)
+            focusedItemIndex == -1 &&
+                !lazyListState.isScrollInProgress &&
```
filbabic, line 265: same here

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -197,43 +234,57 @@ internal fun BoxScope.DefaultMessagesHelperContent(
...
+                    lazyListState.animateScrollToItem(0)
                 }
-                lazyListState.animateScrollToItem(0)
             }
         }
     }
 
-    if (abs(firstVisibleItemIndex) >= 3) {
+    if ((messagesState.parentMessageId == null && (abs(firstVisibleItemIndex) >= 3 || !areNewestMessagesLoaded)) ||
+        (messagesState.parentMessageId != null && abs(firstVisibleItemIndex) >= 3)
```
filbabic, line 280: Kind of same here - maybe 2 values and we can have `firstValue || secondValue` (obviously not htose names, something meaningful

filbabic: Also to mention - having thread and messages state in the same UI, we might want to split it up in the future

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -170,7 +178,6 @@ public class MessageListViewModel(
      * Holds the current [MessageMode] that's used for the messages list. [MessageMode.Normal] by default.
      */
     public var messageMode: MessageMode by mutableStateOf(MessageMode.Normal)
-        private set
```
filbabic, line 173: why did we remove this?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -280,6 +287,9 @@ public class MessageListViewModel(
                                 messageItems = emptyList(),
                             )
                         is io.getstream.chat.android.offline.plugin.state.channel.MessagesState.Result -> {
+
+                            val scrollingState = shouldScrollToBottom(channelState.endOfNewerMessages.value)
```
filbabic, line 291: I'd rename this: getScrollingState() or something - shouldScrollToBottom is a name if we'd expect a boolean - true or false, not a complex data type

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -311,16 +325,16 @@ public class MessageListViewModel(
...
                         }
 
-                        messagesState.messageItems.firstOrNull {
-                            it is MessageItemState && it.message.id == scrollToMessage?.id
-                        }?.let {
-                            focusMessage((it as MessageItemState).message.id)
+                        if (messagesState.scrollToPositionState == LoadFocusedMessageData && scrollToMessage != null) {
+                            messagesState.messageItems.firstOrNull {
+                                it is MessageItemState && it.message.id == scrollToMessage?.id
+                            }?.let { focusMessage((it as MessageItemState).message.id) }
```
filbabic, line 337: do we have to have this check? can't we attempt to focusMessage(scrollToMessage.id) and if it's null, nothing happens?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -394,7 +413,15 @@ internal class ChannelLogic(
         repos.insertMessages(messages)
     }
 
-    internal fun upsertMessage(message: Message) = upsertMessages(listOf(message))
+    internal fun upsertMessage(message: Message) {
+        /* Do not upsert message if the newest messages are not loaded to avoid breaking pagination. If the message
+        * is inside the list and was updated in some kind of way, update it in the list. */
+        // if (mutableState._endOfNewerMessages.value ||
+        //     mutableState._messages.value.containsKey(message.id)
```
filbabic: should we remove the commented out code?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -754,6 +781,7 @@ internal class ChannelLogic(
         if (changedMessages.isNotEmpty()) {
             upsertMessages(changedMessages)
         }
+        mutableState._loading
```
filbabic, line 784: this is useless, right?

filbabic (changes requested):
Left a few comments - let's continue improvements here as there's a few bits we can polish out before merging

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
TGazica: no we should not, this should have been reverted

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
TGazica: yes, must have left it by accident after testing

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
TGazica: No need to wrap it with state, it is not observed anywhere, only passed so that we can calculate the offset. There is no need for it to be a state.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
TGazica: forgot to add it back, was testing how it would work with a controller which is to be implemented later.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
TGazica: We do not, but to me it does not make sense to iterate through the whole list each time data updates if we do not have to.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
TGazica: Also we should check where it makes sense to block upserting a list of messages sense to prevent breaking pagination.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -182,64 +351,154 @@ private fun OnLastVisibleItemChanged(lazyListState: LazyListState, onChanged: (f
...
-    if (abs(firstVisibleItemIndex) >= 3) {
+    if (isScrollToBottomButtonVisible(isMessageInThread, firstVisibleItemIndex, areNewestMessagesLoaded)) {
         MessagesScrollingOption(
             unreadCount = messagesState.unreadCount,
             modifier = Modifier.align(Alignment.BottomEnd),
             onClick = {
-                coroutineScope.launch {
-                    if (firstVisibleItemIndex > 5) {
-                        lazyListState.scrollToItem(5)
+                scrollToBottom {
```
filbabic, line 408: This is also a UX breaking change - we had the implementation of scrollign to the 5th item from the bottom then all the way to the end, as a nice easing scroll, rather than just jumping to the bottom. Can we keep that?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -504,14 +515,53 @@ public class MessageListViewModel(
...
+            messagesState = messagesState.copy(isLoadingMore = true, isLoadingMoreOldMessages = true)
+            chatClient.loadOlderMessages(channelId, messageLimit).enqueue()
+        }
+    }
+
+    /**
+     * Loads newer messages of a channel following the currently newest loaded message. In case of threads this will
+     * do nothing.
+     *
+     * @param messageId The id of the most new [Message] inside the messages list.
```
filbabic, line 542:
```suggestion
     * @param messageId The id of the newest [Message] inside the messages list.
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
TGazica: we can keep it, will revert.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -289,7 +297,11 @@ public class MessageListViewModel(
                                 ),
                                 isLoadingMore = false,
                                 endOfMessages = channelState.endOfOlderMessages.value,
+                                startOfMessages = channelState.endOfNewerMessages.value,
                                 currentUser = user,
+                                isLoadingMoreNewMessages = false,
+                                isLoadingMoreOldMessages = false,
+                                unreadCount = unreadCount ?: messagesState.unreadCount
```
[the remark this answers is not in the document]
TGazica: By adding the unread count to combine, it should take that value which would be null, so that should not happen.

adasiewiczr (commented):
Great job!

PR #3939, opened 2022-07-22
[3936] channel list filter change

TGazica:
### 🎯 Goal

Fix channel list population when switching filters. 

### 🛠 Implementation details

Changed how the channel list is parsed and set to the view along with pagination states.

### 🧪 Testing
<details>

```
Index: stream-chat-android-ui-components-sample/src/main/res/layout/fragment_channels.xml
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-components-sample/src/main/res/layout/fragment_channels.xml b/stream-chat-android-ui-components-sample/src/main/res/layout/fragment_channels.xml
--- a/stream-chat-android-ui-components-sample/src/main/res/layout/fragment_channels.xml	(revision cb859d12352b23ea9f3b8d33a1372d5cec2612c8)
+++ b/stream-chat-android-ui-components-sample/src/main/res/layout/fragment_channels.xml	(date 1658498206186)
@@ -55,4 +55,32 @@
         app:layout_constraintTop_toBottomOf="@id/searchInputView"
         />
 
+    <TextView
+        android:id="@+id/allChannels"
+        android:text="All"
+        app:layout_constraintEnd_toStartOf="@id/switchFilter"
+        app:layout_constraintTop_toTopOf="@id/switchFilter"
+        />
+
+    <androidx.appcompat.widget.SwitchCompat
+        android:id="@+id/switchFilter"
+        app:layout_constraintEnd_toStartOf="@id/singleChannel"
+        />
+
+    <TextView
+        android:id="@+id/singleChannel"
+        android:text="Single"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="@id/switchFilter"
+        />
+
 </androidx.constraintlayout.widget.ConstraintLayout>
Index: stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt	(revision cb859d12352b23ea9f3b8d33a1372d5cec2612c8)
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt	(date 1658498206184)
@@ -29,7 +29,7 @@
 import androidx.fragment.app.viewModels
 import androidx.navigation.findNavController
 import com.getstream.sdk.chat.utils.Utils
-import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.extensions.isAnonymousChannel
 import io.getstream.chat.android.client.models.Filters
@@ -41,27 +41,46 @@
 import io.getstream.chat.ui.sample.R
 import io.getstream.chat.ui.sample.application.App
 import io.getstream.chat.ui.sample.common.navigateSafely
-import io.getstream.chat.ui.sample.data.user.SampleUser
 import io.getstream.chat.ui.sample.databinding.FragmentChannelsBinding
 import io.getstream.chat.ui.sample.feature.common.ConfirmationDialogFragment
 import io.getstream.chat.ui.sample.feature.home.HomeFragmentDirections
 
 class ChannelListFragment : Fragment() {
 
-    private val viewModel: ChannelListViewModel by viewModels {
+    var isPinnedMode: Boolean = false
+        set(value) {
+            field = value
+            viewModel.setFilters(filter)
+        }
+
+    val pinned = listOf(
+        // "messaging:sample-app-channel-109",
+        // "messaging:c92c5e2d-a547-4337-a3bd-2d0c29b71130",
+        // "messaging:sample-app-channel-6",
+        "messaging:!members-QlpghUGsUksIlxfJHiq3mkIjsbPCK7Peipy767bzjnw",
+        // "messaging:!members-HmbzW3-bq4cZ76SWuXt9OFIn17cOB6HGi8i0uwCUDag",
+        // "essaging:1e570561-5f5e-4606-af27-c0153adb8962",
+        // "messaging:sample-app-channel-57",
+        // "messaging:8d29d9ee-445a-43f6-841c-e00d91d32c42"
+    )
+
+    val filter: FilterObject
+        get() = if (isPinnedMode) {
         val user = App.instance.userRepository.getUser()
-        val userId = if (user == SampleUser.None) {
-            ChatClient.instance().getCurrentUser()?.id ?: ""
-        } else {
-            user.id
-        }
+        Filters.and(
+            Filters.`in`("members", listOf(user.id)),
+            Filters.`in`("cid", pinned),
+        )
+    } else {
+        val user = App.instance.userRepository.getUser()
+        Filters.and(
+            Filters.`in`("members", listOf(user.id)),
+        )
+    }
 
+    private val viewModel: ChannelListViewModel by viewModels {
         ChannelListViewModelFactory(
-            filter = Filters.and(
-                Filters.eq("type", "messaging"),
-                Filters.`in`("members", listOf(userId)),
-                Filters.or(Filters.notExists("draft"), Filters.eq("draft", false)),
-            ),
+            filter = filter,
         )
     }
@@ -85,6 +104,10 @@
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        binding.switchFilter.setOnCheckedChangeListener { _, isChecked ->
+            isPinnedMode = isChecked
+        }
+
         setupOnClickListeners()
         viewModel.bindView(binding.channelsView, viewLifecycleOwner)
         searchViewModel.bindView(binding.searchResultListView, this)
```

</details>

1. Build this branch with the patch above
2. Click on the filter switch
3. Login as Oleg
4. On `All` all channels should be visible and on  `Single` only a single channel should be visible.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

giphy (2)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -368,10 +368,22 @@ public class ChannelListView : FrameLayout {
         this.loadingView.isVisible = true
     }
 
+    @Deprecated(
+        message = "Deprecated in favor of new logic for setChannels. To set the loading item call set channels with" +
+            "the loading item already in the list. Show loading more function is race condition prone, which can" +
+            "cause the list to apply an old state of channels.",
+        level = DeprecationLevel.WARNING
```
filbabic, line 375: Since it's deprecated with a warning but we still have the old behavior, can we, within the body, just do what the message says?

Just add the list of items with the loading item?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/ChannelListView.kt
```diff
@@ -368,10 +368,22 @@ public class ChannelListView : FrameLayout {
...
+    )
     public fun showLoadingMore() {
         this.simpleChannelListView.showLoadingMore(true)
     }
 
+    @Deprecated(
+        message = "Deprecated in favor of new logic for setChannels. To set the loading item call set channels with" +
+            "the loading item already in the list. Show loading more function is race condition prone, which can" +
+            "cause the list to apply an old state of channels.",
```
filbabic, line 384: Message here doesn't make sense - we should set the list _without_ the item, right?

filbabic: Also references `showLoadingMore`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/internal/SimpleChannelListView.kt
```diff
@@ -164,9 +164,20 @@ internal class SimpleChannelListView @JvmOverloads constructor(
...
+        }
     }
 
+    @Deprecated(
+        message = "Deprecated in favor of new logic for setChannels. To set the loading item call set channels with" +
+            "the loading item already in the list. Show loading more function is race condition prone, which can" +
+            "cause the list to apply an old state of channels.",
+        level = DeprecationLevel.WARNING
+    )
     fun showLoadingMore(show: Boolean) {
```
filbabic, line 181: Same here, we could migrate to the new behavior

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
```diff
@@ -192,22 +196,23 @@ public class ChannelListViewModel(
         /**
          * We clean up any previous loads to make sure the current one is the only one running.
          */
-        if (queryJob != null) {
-            queryJob?.cancel()
+        queryJob?.cancel()
+        val queryJob = SupervisorJob(viewModelScope.coroutineContext.job).also {
```
filbabic, line 200: Why do we specifically create a supervisor here with a parent job context? A supervisor job should override the parent job anyway so it doesn't do anything. We can still pass in the context of the ViewModelScope.

Alternatively, the `ViewModelScope`'s context should already be using a SupervisorJob, so we scan just use that (probably).

Either way - unless we use the job to cancel the children at some point in time, we don't need a supervisor.

If we use it just for that, then we can always have a live SupervisorJob, and instead of canceling the job, we can cancel its children and reuse the job.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/LiveData.kt
```diff
@@ -34,3 +41,19 @@ internal fun <P, Q> Pair<LiveData<P>, LiveData<Q>>.observeTogether(
         block(this.first.value, it)
     }
 }
+
+/**
+ * Adds Flow as a Source to this MediatorLiveData
```
filbabic, line 46: missing full stop at the end

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/LiveData.kt
```diff
@@ -34,3 +41,19 @@ internal fun <P, Q> Pair<LiveData<P>, LiveData<Q>>.observeTogether(
...
+ */
+internal fun <T, S> MediatorLiveData<T>.addFlow(
+    context: CoroutineContext = EmptyCoroutineContext,
```
filbabic, line 49: In this case if we pass in a SupervisorJob that has only a Job as a context parameter, we're not not doing anything, we should pass in the correct context that uses a valid dispatcher etc.

filbabic (commented):
Left some comments!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
kanat: On the one hand, we use `viewModelScope.coroutineContext.job` as parent for this `SupervisorJob` to ensure this `SupervisorJob ` will be cancelled with `viewModelScope`.

On the other hand, filtering has a shorter lifecycle, because it can be changed with a new `setFilters` method. So this `queryJob` is also passed to all `asLiveData()` methods as well to cancel those. This is because we kept listening to the previous `QueryChannelsState` changes.
For that purpose, a new extension function is being created named `addFlow`.

kanat: I firstly wanted to reuse a single instance of `QueryChannelsState`. However, in this case,` invokeOnCompletion` won't be called in the snippet below:
```kotlin
internal fun <T, S> MediatorLiveData<T>.addFlow(
    context: CoroutineContext = EmptyCoroutineContext,
    source: Flow<S>,
    onChanged: Observer<in S>
) {
    source.asLiveData(context).also { liveData ->
        addSource(liveData, onChanged)
        context[Job]?.invokeOnCompletion {
            removeSource(liveData)
        }
    }
}
```

filbabic: On the first point - good point, I missed how the parent job is being used (I'm dumb).

On the second - if we have a supervisor job and we pass it around to create new coroutines (children) then we can just cancelChildren right?

If we don't need that, we don't need a supervisor right, we can just pass in a regular job whose parent context is the VM context - then by cancelling the job, we clear up its coroutines and the parent job is not canceled since it's also a supervisor.

So the proposed strucutre would be:

Supervisor Job (VM Scope)
  - Job (query)
      - Any other coroutines started with it
      
Rather than

SupervisorJob (VM Scope)
  - Supervisor (Query)
    - Other coroutines

Does this make sense? I might be missing something but I think we can just rely on a simple job here or use the supervisor in a reusable manner rather than recreating it

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/common/extensions/internal/LiveData.kt
filbabic: I think that should be enough - sorry for the misinterpretation of what the SupervisorJob "constructor" was doing

kanat: Sure, no worries, man 👌

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModel.kt
kanat: Makes sense. 
Thanks for the feedback, man 🍻

Replaced `SupervisorJob` with a regular `Job`.

adasiewiczr (approved):
Looks good from core perspective

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModelBinding.kt
```diff
@@ -37,25 +39,23 @@ public fun ChannelListViewModel.bindView(
...
+            view.setPaginationEnabled(!it.endOfChannels && !it.loadingMore)
         }
-    }
 
-    paginationState.observe(lifecycleOwner) { paginationState ->
-        view.setPaginationEnabled(!paginationState.endOfChannels && !paginationState.loadingMore)
+        var list: List<ChannelListItem> = state?.channels?.map(ChannelListItem::ChannelItem) ?: emptyList()
+        if (paginationState?.loadingMore == true) {
+            list = list + ChannelListItem.LoadingMoreItem
+        }
```
bychkovdmitry, line 50: I logged whats going on and seems that loading a new page produced redundant emit. Given 30 loaded channels and the list is scrolled to the end:

1. 30 channels + 1 loading more item
2. 60 channels + 1 loading more item <- autoscrolling here to the 60th channel
3. 60 channels 

I think the list should be scrolled to the 30th item (or 31 item in case the progress bar is seamlessly replaced with the 31 item). Could you check?

bychkovdmitry (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/channel/list/viewmodel/ChannelListViewModelBinding.kt
TGazica: Yes, seems that the loadingMore flag was not set to false before so the indicator would be added and scrolled to. Can you chek out the fix?

bychkovdmitry (approved):
Looks good from the UI side!

PR #3948, opened 2022-07-26
Feature/3929 bidirectional pagination

TGazica:
### 🎯 Goal

Rerevert of #3818 

### 🛠 Implementation details

Updated `ChannelLogic` which determines if the pagination has reached either end or start of the messages list. Also added checks when receiving or sending new messages, so that the new message isn't added to the list if the newest messages are not loaded to avoid breaking pagination of newer messages.

Introduced new pagination handler for reaching the end of new messages to `MessageList` composable. Added scroll to bottom handler to `MessagesList`.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/38032787/176719148-72f40164-bf9b-433e-9f5e-190f44cdc6ea.mp4" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/38032787/176719135-254b3323-06c0-4064-9f9a-c2bba4cf9157.mp4" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

1. Build this branch and enter a channel that has more than one page
2. Scroll n number of pages and reply to a message
4. Exit the channel and enter it again, the message history will be reset
5. Click on the quoted message which should load the message and a page that surrounds the message
6. Scroll up and down to ensure that the bi directional pagination is working
7. Try sending messages from the current user and from a different device, messages shouldn't be added to the bottom of the list, otherwise pagination will break
8. Try adding reactions to messages you are scrolled to, replying to them and deleting them to insure that the messages are properly updated
9. Click on scroll to bottom button, if the newest messages aren't present they should be loaded and the list should scroll to the bottom
10. Background the app when newest messages are not loaded and send new messages from another device, the messages should not be added to the list.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

giphy


stream-chat-android-state/.env
```diff
@@ -0,0 +1,3 @@
+STREAM_LOG_LEVEL=ALL
```
adasiewiczr, line 1: Why do we need this file? 🤔

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
```diff
@@ -1197,23 +1227,50 @@ public class MessageListViewModel(
...
+     */
+    public fun scrollToBottom(messageLimit: Int = DefaultMessageLimit, scrollToBottom: () -> Unit) {
+        if (isInThread) {
+            scrollToBottom()
+        } else {
+            if (currentMessagesState.startOfMessages) {
+                scrollToBottom()
+                return
+            }
+            chatClient.loadNewestMessages(channelId, messageLimit).enqueue {
```
adasiewiczr, line 1266: Do we have a guarantee the messages are not loaded already?

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/ChatClient.kt
```diff
@@ -330,3 +331,35 @@ private suspend fun ChatClient.loadMessageByIdInternal(
...
+ * @return Executable async [Call] responsible for loading the newest messages.
+ */
+public fun ChatClient.loadNewestMessages(cid: String, messageLimit: Int, userPresence: Boolean = true): Call<Channel> {
+    return CoroutineCall(state.scope) {
+        loadNewestMessagesInternal(cid, messageLimit, userPresence)
+    }
+}
+
+@CheckResult
+internal suspend fun ChatClient.loadNewestMessagesInternal(
```
adasiewiczr, line 352: Can we move this logic to `loadNewestMessages ` instead?

adasiewiczr (commented):
Great job! Added a few comments/questions.
@kanat @leandroBorgesFerreira please take a look on this PR as well

stream-chat-android-state/.env
TGazica: Probably not, not sure where and when this was introduced, will remove it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/MessageListViewModel.kt
TGazica: yes we do, but the function is not implemented properly, might have left it like this after some testing. This should be inside else since the check above checks whether the newest messages are loaded or not. Will fix this.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/ChatClient.kt
```diff
@@ -116,9 +116,10 @@ public fun ChatClient.watchChannelAsState(
     cid: String,
     messageLimit: Int,
     coroutineScope: CoroutineScope = CoroutineScope(DispatcherProvider.IO),
+    skipMessages: Boolean = false
```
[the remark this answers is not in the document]
TGazica: Also, we are sending `skipMessages = true` within the `MessageComposerViewModel` and `MessageHeaderViewModel` but debouncing is done using checking if two `QueryChannelRequest`s are the same. I updated it's equals function to take into account `skipMessages`.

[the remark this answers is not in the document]
adasiewiczr: What if instead of introducing an additional parameter we can manipulate with `messageLimit`. We could expose the limit via constructor in each view model and set it to `0` if `messageId != 0`. What do you think?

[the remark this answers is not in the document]
TGazica: Hmm, in stead of exposing `messageLimit` maybe we can expose `messageId` and then set the limit to 0 if we find a non-null `messageId`. Introducing `messageLimit` to composer and header might be confusing and in the case of list would mean that we take it as param for pagination and loading of data. How does this sound?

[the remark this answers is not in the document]
adasiewiczr: Makes sense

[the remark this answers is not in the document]
TGazica: Maybe we don't have to pass `messageId` to all of the VMs. The only VM that is using the message list is the `MessageListViewModel`. We can pass `messageLimit = 0` to all others vms and let them refresh channel statate without messages. Then we can just do the check inside `MessageListViewModel`. The message list query won't be debounced due to different `messageLimit`.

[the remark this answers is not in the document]
filbabic: I think we can just pass in the messageId to the ListViewModel as the "observed/focused message". 
----

It might be useful to pass in the messageId to all VMs in the future, e.g. if someone wants to outline in the header which message is being observed, or alternatively in the Composer which message is being jumped to and edited.

But these seem like custom use cases and will most likely be confusing to our customers if it doesn't exist as default behavior they can relate to. Hence why I propose to just add it where it makes sense now.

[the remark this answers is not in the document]
TGazica: @adasiewiczr refactored so we use `messageLimit` to control upserting of messages on itial loads.

bychkovdmitry:
Not related to this PR. Message highlight animation in Compose is broken

bychkovdmitry:
There's a small problem with paginating to newer messages. The scroll instantly jumps to the bottom of the new chunk

bychkovdmitry:
1. Scroll up until scroll to bottom FAB is displayed
2. Enter thread
3. FAB remains in the screen

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -446,6 +469,18 @@ internal class ChannelMutableState(
         _messages.value = messages.associateBy(Message::id)
     }
 
+    private fun cacheMessages() {
+        cachedMessages.value = mapOf()
+        cachedMessages.value = cachedMessages.value + sortedVisibleMessages.value.associateBy(Message::id)
```
bychkovdmitry, line 474: Won't `cachedMessages.value` be always empty here?

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/sync/internal/SyncManager.kt
```diff
@@ -390,7 +390,7 @@ internal class SyncManager(
 
                 foundChannels.forEach { channel ->
                     val channelLogic = logicRegistry.channel(channel.type, channel.id)
-                    channelLogic.updateDataFromChannel(channel)
+                    channelLogic.updateDataFromChannel(channel, channel.messages.size)
```
bychkovdmitry, line 393: Could you explain this line?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
```diff
@@ -107,8 +110,8 @@ public class MessageComposerController(
      */
     public val channelState: Flow<ChannelState> = chatClient.watchChannelAsState(
         cid = channelId,
-        messageLimit = DefaultMessageLimit,
-        coroutineScope = scope
+        messageLimit = if (messageId != null) 0 else DefaultMessageLimit,
```
bychkovdmitry, line 113: Good, so in case no message is focused, then composer and message list will have the very same request which will be executed only once?

stream-chat-android-ui-components/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
```diff
@@ -217,10 +246,14 @@ internal class MessageListItemLiveData(
      * @return Full list of [MessageListItem] to represent the state.
      */
     private fun buildItemsList(): List<MessageListItem> {
-        return getLoadingMoreItems() + messageItemsWithReads + typingItems
+        return if (loadingMoreOlderItems) {
```
bychkovdmitry, line 249: I haven't seen loading indicators in the message list for quite a while. Are they handled in the UI?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/MessagesState.kt
```diff
@@ -34,6 +31,9 @@ import kotlinx.coroutines.flow.StateFlow
  * @param newMessageState The state that represents any new messages.
  * @param parentMessageId The id of the parent message - if we're in a thread.
  * @param unreadCount The count of messages we haven't read yet.
+ * @param startOfMessages If we're ate the start of messages (to stop pagination).
```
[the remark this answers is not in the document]
TGazica: Makes sense, will rename it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
```diff
@@ -50,18 +49,20 @@ import io.getstream.chat.android.compose.viewmodel.messages.MessageListViewModel
...
  * @param onReactionsClick Handler when the user taps on message reactions and selects them.
- * @param onMessagesStartReached Handler for pagination.
+ * @param onMessagesStartReached Handler for pagination when the end of the oldest messages has been reached.
  * @param onLastVisibleMessageChanged Handler that notifies us when the user scrolls and the last visible message
  * changes.
  * @param onScrollToBottom Handler when the user reaches the bottom.
  * @param onGiphyActionClick Handler when the user clicks on a giphy action such as shuffle, send or cancel.
  * @param onQuotedMessageClick Handler for quoted message click action.
  * @param onMediaGalleryPreviewResult Handler when the user selects an option in the Media Gallery Preview screen.
+ * @param onMessagesEndReached Handler for pagination when the end of newest messages have been reached.
```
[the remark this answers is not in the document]
TGazica: Agreed, this should be `loadNextPage` or something along those lines.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -181,65 +212,160 @@ private fun OnLastVisibleItemChanged(lazyListState: LazyListState, onChanged: (f
  * Represents the default scrolling behavior and UI for [Messages], based on the state of messages and the scroll state.
  *
  * @param messagesState The state of messages, current message list, thread, user and more.
- * @param lazyListState The scrolling state of the list, used to manipulate and trigger scroll events.
+ * @param messagesLazyListState The scrolling state of the list, used to manipulate and trigger scroll events.
+ * @param scrollToBottom Handler when the user requests to scroll to the bottom of the messages list.
  */
 @Composable
 internal fun BoxScope.DefaultMessagesHelperContent(
```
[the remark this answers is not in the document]
TGazica: I can check it, but my guess it is visible on start since `areNewestMessagesLoaded` is probably false before the list comes in.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Channel.kt
```diff
@@ -53,6 +53,9 @@ import java.util.Date
  * in the events.
  * @param membership Represents relationship of the current user to this channel.
  * @param extraData A map of custom fields for the channel.
+ * @param cachedMessages The list of cached messages if the regular list does not contain the newest loaded messages.
```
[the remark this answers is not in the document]
TGazica: Makes sense for cached messages. As for inside search it has some ties with it but we can see to rename it.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/state/channel/internal/ChannelMutableState.kt
TGazica: yes, we shuld just set the value to the soretdVisibility values.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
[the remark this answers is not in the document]
filbabic: Here's the thing: if we don't trigger "loading of the next page", the `loadNextPage` naming doesn't make sense - hence why we named it the "endReached" so our users can trigger actions when they've reached the end.

But the naming might want to imply the end of the current page, so something like: "onMessagesPageEndReached" rathern than the very end

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/sync/internal/SyncManager.kt
TGazica: This updates the channel data on sync and updates the newest messages. we pass messages size as `messageLimit` since if `messageLimit == 0` we ignore the messages. So if there are messages they will get added to the list.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/common/composer/MessageComposerController.kt
TGazica: Correct

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/Channel.kt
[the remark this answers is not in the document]
TGazica: I would go with `cachedLatestMessages` so that implies it is a cache and might be cleared at some point. For `latestMessages` i would expect them always to return something.

stream-chat-android-ui-components/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
TGazica: it seems we do not have a ViewHolder for the loading spinners.

stream-chat-android-state/.env
```diff
@@ -0,0 +1,2 @@
+STREAM_API_KEY=b67pax5b2wdq
```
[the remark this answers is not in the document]
TGazica: removed

stream-chat-android-ui-components/src/main/kotlin/com/getstream/sdk/chat/viewmodel/messages/MessageListItemLiveData.kt
bychkovdmitry: Created a ticket for that:
https://github.com/GetStream/stream-chat-android/issues/4241

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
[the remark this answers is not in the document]
bychkovdmitry: I mean "start" and "end" was a bit confusing for me at first glance. I thought that "start" is the first chunk of loaded messages. But that's only me. I'm more concerned with `isInsideSearch` variable, as it has nothing to do with search. Maybe `missingLatestMessages`?

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -181,20 +188,42 @@ internal class ChannelLogic(
     }
 
     private suspend fun runChannelQueryOffline(request: QueryChannelRequest): Channel? {
+        /* It is not possible to guarantee that the next page of newer messages is the same of backend,
```
MarinTolic, line 191:
```
...is the same of backend
```

"the same as the one in the backend" probably fits better here

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogic.kt
```diff
@@ -347,11 +359,54 @@ internal class ChannelStateLogic(
...
+        val currentMessages = mutableState.cachedLatestMessages.value
+        return currentMessages + attachmentUrlValidator.updateValidAttachmentsUrl(messages, currentMessages)
+            .filter { newMessage -> isMessageNewerThanCurrent(currentMessages[newMessage.id], newMessage) }
+            .associateBy(Message::id)
+    }
+
+    /**
+     * @param isNotificationUpdate Whether the data is updating due to a new notification.
+     * @param isInsideSearch Whether we are inside search or not.
+     * @param isScrollUpdate Whether the update is dut to a scroll update, meaning pagination.
```
MarinTolic, line 396:
```suggestion
     * @param isScrollUpdate Whether the update is due to a scroll update, meaning pagination.
```

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -83,6 +83,11 @@ internal class ChannelMutableState(
     /** the data to hide messages before */
     var hideMessagesBefore: Date? = null
 
+    /**
+     * Messages to show on the channels list when the state is inside search to show the latest message in the preview.
+     */
+    internal val cachedLatestMessages: MutableStateFlow<Map<String, Message>> = MutableStateFlow(emptyMap())
```
MarinTolic, line 89: Sometimes when you navigate to an older message via a quoted reply and then go back, the preview message of the channel you were in when you navigated to said message is different from the latest one and the channel is misplaced in the list (Affects both XML and Compose). This happens even when you are not in the search state.

I wonder if this could be the cause of it.

MarinTolic (changes requested):
Found a few issues that need to be ironed out, otherwise an awesome improvement on the previous mechanism.

Issues communicated via Slack.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/state/channel/internal/ChannelMutableState.kt
TGazica: Could be, will investigate this.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -148,16 +169,26 @@ public fun Messages(
...
-                    if (!endOfMessages && index == messages.lastIndex &&
+                    if (!endOfMessages &&
+                        index == messages.lastIndex &&
                         messages.isNotEmpty() &&
                         lazyListState.isScrollInProgress
                     ) {
                         onMessagesStartReached()
                     }
+
+                    val newestMessageItem = (messages.firstOrNull { it is MessageItemState } as? MessageItemState)
```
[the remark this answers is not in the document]
TGazica: This could possibly be an issue if all of the loaded messages are not `MessageItemState`. What do you propose we do instead?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -181,65 +212,160 @@ private fun OnLastVisibleItemChanged(lazyListState: LazyListState, onChanged: (f
...
+private fun shouldScrollToBottomOnNewMessage(
+    focusedItemIndex: Int,
+    firstVisibleItemIndex: Int,
+    newMessageState: NewMessageState?,
+    areNewestMessagesLoaded: Boolean,
+    isScrollInProgress: Boolean,
+): Boolean {
+    newMessageState ?: return false
+
+    return focusedItemIndex == -1 &&
```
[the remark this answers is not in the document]
TGazica: Since it is an index and if we search we acquire by `indexOfFirst` if the index is not found `-1` is returned. This check just checks that the focused message is not inside the list.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessagesLazyListState.kt
```diff
@@ -0,0 +1,140 @@
...
+ */
+public data class MessagesLazyListState(
+    public val lazyListState: LazyListState,
+    private val messageOffsetHandler: MessageOffsetHandler = defaultOffsetHandler,
+) {
+
+    /**
+     * By default tracks the size of [Messages] composable. Used to calculate [focusedMessageOffset].
+     */
+    private val _parentSize: MutableStateFlow<IntSize> = MutableStateFlow(IntSize.Zero)
```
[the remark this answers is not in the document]
TGazica: No, but then we have `var`, dont they also mark the class as unsable?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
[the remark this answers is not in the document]
TGazica: Unread counts will need to be sorted out in core, here is the [pr](https://github.com/GetStream/stream-chat-android/pull/4122).

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/state/channel/internal/ChannelMutableState.kt
TGazica: Fixed the logic

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessagesLazyListState.kt
```diff
@@ -43,33 +42,49 @@ public data class MessagesLazyListState(
...
+    }
+
+    /**
+     * Calculates the offset of the focused message on the screen.
+     *
+     * @param parentSize The size of the parent view.
+     * @param focusedMessageSize The size of the message view we wish to focus to on the screen.
+     */
+    private fun calculateMessageOffset(parentSize: IntSize, focusedMessageSize: IntSize) {
+        _focusedMessageOffset.value = messageOffsetHandler.calculateOffset(parentSize, focusedMessageSize)
```
MarinTolic, line 87: Maybe `updateMessageOffset` would be the better naming scheme here in order to make the inner workings of the function more explicit and it seems to be a shorthand for doing so.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Added `isFilteringMessages` check on request. [#3948](https://github.com/GetStream/stream-chat-android/pull/3948)
```
MarinTolic, line 21: What request specifically? Maybe being a bit more descriptive here if necessary could  be good.

MarinTolic, line 21: What request specifically? Maybe being a bit more descriptive here if necessary could be good.

CHANGELOG.md
```diff
@@ -91,6 +97,7 @@
 
 ## stream-chat-android-compose
 ### 🐞 Fixed
+- Fixed pagination when the newest messages arent loaded with newer messages pagination. Fixed scroll to bottom if the newest messages aren't loaded. [#3948](https://github.com/GetStream/stream-chat-android/pull/3948)
```
MarinTolic, line 100:
```suggestion
- Fixed pagination when the newest messages aren't loaded with newer messages pagination. Fixed scroll to bottom if the newest messages aren't loaded. [#3948](https://github.com/GetStream/stream-chat-android/pull/3948)
```

Also the first sentence is a bit hard to read and understand.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -181,65 +208,160 @@ private fun OnLastVisibleItemChanged(lazyListState: LazyListState, onChanged: (f
...
+    val lazyListState = messagesLazyListState.lazyListState
+
+    val messages = messagesState.messageItems
+    val newMessageState = messagesState.newMessageState
+    val areNewestMessagesLoaded = messagesState.newestMessageLoaded
+    val isMessageInThread = messagesState.parentMessageId != null
+
     val coroutineScope = rememberCoroutineScope()
 
     val firstVisibleItemIndex = lazyListState.firstVisibleItemIndex
```
MarinTolic, line 229: I see we removed the switch to derived state here.
I know the linter wants this to be derived so I'm wondering why we reverted.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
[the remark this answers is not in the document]
MarinTolic: Could we make it into a constant then?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessagesLazyListState.kt
```diff
@@ -0,0 +1,157 @@
...
```
MarinTolic, line 87: Maybe `updateMessageOffset` would be the better naming scheme here in order to make the inner workings of the function more explicit and it seems to be a shorthand for doing so.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
TGazica: added derived back in. Removed it when fixing a bug reported by Dmitrii, had to remove remember since it was rememberins an old instance of lazyListState and was causing bugs.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessagesLazyListState.kt
TGazica: I would leave it calculate since we are doing some calculations with the parameters we send, we are not just applying the passed value to the offset.

TGazica: https://github.com/GetStream/stream-chat-android/pull/3948#discussion_r998537998

We can but then it should be a constant we use for every check if the item is not insde the list. I would leave it as is.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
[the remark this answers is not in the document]
TGazica: We can but then it should be a constant we use for every check if the item is not insde the list. I would leave it as is.

MarinTolic (approved):
There are a few issues with the unread count badge such as:

XML: 

- Unread count badge not showing unread message count after navigating to a message via reply

These have been addressed in the message list controller PR seen [here](https://github.com/GetStream/stream-chat-android/pull/4157) so it's unnecessary to insist on them being fixed.

### ☑️Reviewer Checklist

PR #3954, opened 2022-07-26
Fixing message preview

leandroBorgesFerreira:
Fixing message preview.


stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -137,10 +138,15 @@ internal class ChannelLogic(
...
             repos.insertChannelConfig(ChannelConfig(channel.type, channel.config))
-            storeStateForChannel(channel)
-        }
-            .onSuccess { channel -> channelStateLogic.propagateChannelQuery(channel, request) }
-            .onError(channelStateLogic::propagateQueryError)
+
+            if (!mutableState.insideSearch.value && !request.isFilteringAroundIdMessages()) {
+                storeStateForChannel(channel)
+            } else {
+                delay(300)
```
leandroBorgesFerreira, line 145: This is just fixing a race condition that makes a bug in the jump to the message.

TGazica (approved):
Looks good

PR #3960, opened 2022-07-27
Cancel children jobs when ChatClient is disconnected

JcMinarro:
### 🎯 Goal
Fix: #3814 
When `ChatClient.disconnect()` is invoked, any jobs started within `ChatClient` is not needed anymore.
Canceling the "root coroutine scope" used within `ChatClient` and the rest of our SDK's Components all jobs are canceled.

### 🧪 Testing
1-. Apply the following patch.
2-. Open the Sample App, interact with some channels and sign-out the current users.
3-. On the logcat you should see some logs that show you started jobs within the `ChatClient` and after canceling all children, only one job is alive (It should be the one that is canceling the rest of jobs)

<details>

```
Index: stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt	(revision a0b47ea3032c50286db22d28ee058b85e0ea0dbe)
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt	(date 1658917981735)
@@ -165,6 +165,7 @@
 import io.getstream.logging.StreamLog
 import io.getstream.logging.android.AndroidStreamLogger
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableSharedFlow
@@ -1092,8 +1093,12 @@
      */
     private fun postponeCancelScope() {
         scope.launch {
+            println("Current children: ${scope.coroutineContext[Job]?.children?.count()}")
+            scope.coroutineContext[Job]?.children?.forEach { println("Job: $it") }
             delay(100)
             scope.coroutineContext.cancelChildren()
+            println("Children after that: ${scope.coroutineContext[Job]?.children?.count()}")
+            scope.coroutineContext[Job]?.children?.forEach { println("Job: $it") }
         }
     }
 

```

</details>

#### General

#### Code & documentation

### ☑️Reviewer Checklist


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1082,6 +1086,17 @@ internal constructor(
...
+    /**
+     * Cancel all jobs on the ChatClient Scope.
+     * This method postpone a new coroutine to cancel all children jobs.
+     */
+    private fun postponeCancelScope() {
+        scope.launch {
+            delay(100)
+            scope.coroutineContext.cancelChildren()
+        }
+    }
```
JcMinarro, line 1098: We need to schedule and delay this process to be able to pass our tests because in another case the test itself is canceled, and not pass our test suite

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1082,6 +1086,17 @@ internal constructor(
         disconnect(true).execute()
     }
 
+    /**
+     * Cancel all jobs on the ChatClient Scope.
+     * This method postpone a new coroutine to cancel all children jobs.
+     */
+    private fun postponeCancelScope() {
+        scope.launch {
+            delay(DELAY_TIME_TO_CANCEL_CHILDREN)
```
filbabic, line 1095: Any reason why we postpone here?

filbabic: Just saw the previous comment - if we could do it in a non.delay manner, it would be great, but I guess this is fine for now.

JcMinarro: As I commented on the previous comment, it is needed to be able to pass our tests, because when `connectUser()` fails, `disconnect()` method is called (To keep a clean state in our `ChatClient`), but because we are using the same coroutineScope and it is canceled, our tests fail

adasiewiczr: It should be fine to keep it for now. Later we can investigate if we can somehow get rid of it

adasiewiczr (approved):
LGTM

kanat (approved):
LGTM.

It would be to have a linked ticket to investigate it and eliminate this workaround.

Issue #3238, opened 2022-03-22
Revert #3237 after bumping Kotlin to 1.6.20


adasiewiczr:
### Context & proposal
#3237 was introduced in order to fix Kotlin 1.6.0 compiler [issue](https://youtrack.jetbrains.com/issue/KT-49793) that should be fixed in `1.6.20` release. 
Revert the workaround and make sure everything works correctly by following Testing steps from #3237

PR #3965, opened 2022-07-28
Update Compose to 1.2.0, Kotlin to 1.7.0, and target SDK to 32

skydoves:
### 🎯 Goal

Update Compose to 1.2.0, Kotlin to 1.7.0, and target SDK to 32. Also updated dependencies below along with the Compose 1.2.0 and Kotlin 1.7.0.

- accompanist `0.25.0` stable.
- Dokka `1.7.0`.
- KSP `1.7.0`.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

gif


filbabic:
Should we also cover this since we're pushing the Kotlin version up: https://github.com/GetStream/stream-chat-android/issues/3238

filbabic:
@adasiewiczr Compose and Kotlin have to go hand-in-hand - in this case we're updating to 1.7.0 due to Compose 1.2.0, do we think it's okay, are we okay with the upgrade without a breaking change?

filbabic (approved):
From my standpoint, everything looks and works good! :]

I couldn't find any issues and the code updates are great!

The only concern I had was pushing to 1.7.0 of Kotlin, but we have to do that regardless because of Compose. I'm not sure if it's a breaking change for our users and how it'll work compatibility-wise.

But I think there were no major API or functionality changes between 1.6.21 and 1.7

buildSrc/src/main/kotlin/io/getstream/chat/android/Configuration.kt
```diff
@@ -1,8 +1,8 @@
 package io.getstream.chat.android
 
 object Configuration {
-    const val compileSdk = 31
-    const val targetSdk = 31
+    const val compileSdk = 32
+    const val targetSdk = 32
```
adasiewiczr, line 5: We shouldn't bump compile and target SDK versions for now as it's a breaking change and we are not ready to do that as we didn't adapt to new permissions system, etc.

adasiewiczr (changes requested):

buildSrc/src/main/kotlin/io/getstream/chat/android/Configuration.kt
JcMinarro: I was concern about that as well, but reviewing documentation [SDK 32 is Android 12L](https://developer.android.com/studio/releases/platforms#12), the new permissions system is added on [SDK 33 (Android 13)](https://developer.android.com/about/versions/13/behavior-changes-13)

adasiewiczr: @JcMinarro nice catch! We should be good to bump both then

adasiewiczr (approved):

PR #4011, opened 2022-08-09
[3176] Apply fixed image and link dimensions in order to avoid items jumping upon loading

MarinTolic:
### 🎯 Goal

Closes #3176 

The second iteration of our CDN provides us with fixed sizes for images. This allows us to prevent jumping images changing their sizes abruptly while loading, which had previously lead to focused items such as when navigating to original messages the reply was made to jumping out of focus.

### 🛠 Implementation details

- Eagerly set the dimension ratio and dimension of the image attachment containers in `ImageAttachmentGroupView` if we are displaying a single image
- Set a fixed size (dimen ratio) for the link previews containers

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| before | after |

### 🧪 Testing

Repeat the same steps for develop and this current branch
**Important:** Uninstall and reinstall the app or flush the storage & cache before switching branches. Otherwise images might remain loaded in the memory and the results might get muddied. 

1. Log in as Filip, Amit, Jc, Thierry or Tommasso
2. Enter channels `Fixed image link test` and `Fixed image sizes test channel`
3. Scroll upwards and view the links/ images

Result: The images and links should "jump" on `develop` but not on the PR branch.

1. Log in as Filip, Amit, Jc, Thierry or Tommasso
2. Enter channel `WS Events
3. Using the message `Use this for testing` sent by JC, navigate to the original message the reply was made to

Result: On `develop`, the message should be bumped out of focus by loading neighboring images/ links. On the PR branch, it should remain in focus.

#### General
- [x] Assigned a person / code owner group (required)
- [x] Thread with the PR link started in a respective Slack channel (#android-chat-core or #android-chat-ui) (required)
- [x] PR targets the `develop` branch
- [x] PR is linked to the GitHub issue it resolves

#### Code & documentation
- [x] Changelog is updated with client-facing changes
- [x] New code is covered by unit tests
- [x] Comparison screenshots added for visual changes
- [ ] Affected documentation updated (KDocs, docusaurus, tutorial)

### ☑️Reviewer Checklist
- [ ] UI Components sample runs & works
- [ ] Compose sample runs & works
- [ ] UI Changes correct (before & after images)
- [ ] Bugs validated (bugfixes)
- [ ] New feature tested and works
- [ ] Release notes and docs clearly describe changes
- [ ] All code we touched has new or updated KDocs

### 🎉 GIF

[...]

PR #4013, opened 2022-08-09
[3176] Apply fixed image dimensions

MarinTolic:
# Please review #4011 before reviewing this PR

The core changes from #4011 have been cherry picked into this PR, so ideally we shouldn't be asking for an unnecessary core review but rather merge the other one first and then have no core diff with develop.

### 🎯 Goal

Previously our image previews inside the message list in Compose were static, meaning they would retain the same dimensions regardless of the aspect ratio of the image they were displaying.

We can achieve a nicer looking UI by resizing the images according to their aspect ratio.

### 🛠 Implementation details

- Deprecate `StreamDimens.attachmentsContentImageHeight` in favor of `StreamDimens.attachmentsContentImageMaxHeight`
- Calculate the ratio of the image attachments inside `ImageAttachmentContent` and apply them while respecting the max given height of an image attachment preview dictated by `StreamDimens.attachmentsContentImageMaxHeight`

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| Screenshot_1660128326 | Screenshot_1660128262 |

### 🧪 Testing

1. Launch the Compose Sapmle app
2. Log in as Amit, Jc, Filip, Thierry or Tommaso
3. Enter the channel `Fixed image sizes test channel`
4. Scroll and observe all photos, report if any photo is exceeding dimensions or being rendered in a poor way
5. Feel free to post photos with various aspect ratios to see how they load
6. Read the changelog to make sure all the changes are understandable to the customer.

<details>

<summary>Patch that lowers the maximum height of an image attachment to 200.dp </summary>

```
Index: stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamDimens.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamDimens.kt b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamDimens.kt
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamDimens.kt	(revision 0eae187d4fbe12b96bb7abc547034f084b3db96a)
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamDimens.kt	(date 1660128687709)
@@ -262,7 +262,7 @@
             quotedMessageAttachmentEndPadding = 0.dp,
             groupAvatarInitialsXOffset = 1.5.dp,
             groupAvatarInitialsYOffset = 2.5.dp,
-            attachmentsContentImageMaxHeight = 600.dp,
+            attachmentsContentImageMaxHeight = 200.dp,
         )
     }
 }
```

</details>

1. Apply the patch
2. Repeat the steps above and notice that the photos are now displayed with a different maximum height
3. Report if anything is rendered wrong

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

[!gif](https://media3.giphy.com/media/OPMdqmdAdnSBcEbiMa/giphy.gif?cid=ecf05e47efwsaxv9jby6megq3ssnlggnhguoz6wld8vvsemo&rid=giphy.gif&ct=g)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamDimens.kt
```diff
@@ -162,7 +163,8 @@ public data class StreamDimens(
             quotedMessageAttachmentStartPadding = 8.dp,
             quotedMessageAttachmentEndPadding = 0.dp,
             groupAvatarInitialsXOffset = 1.5.dp,
-            groupAvatarInitialsYOffset = 2.5.dp
+            groupAvatarInitialsYOffset = 2.5.dp,
+            attachmentsContentImageMaxHeight = 600.dp,
```
MarinTolic, line 167: `attachmentsContentImageHeight` would be deprecated in this scenario. 

The idea would no longer be that users set a fixed height, but rather a maximum height that the images could expand to since now they adapt themselves according to the aspect ratio.

filbabic: I'm not 100% sure we receive height/width for legacy images, i.e. ones uploaded before the switch - could you check that.

If that is the case, then having a max height still could be useful.

MarinTolic: We don't, hence why [here](https://github.com/GetStream/stream-chat-android/pull/4013/files/0eae187d4fbe12b96bb7abc547034f084b3db96a#diff-26ca363b87ad5e9cb52b46ce1d0ae50ab0910042d042e3ca7cca6d051c74c21eR110) Ithe ratio is nullable and the fallback option is to square the photo using a ratio of `1:1`.

Essentially, this way our images follow the pattern of our group images so we're safe, even if slightly less aesthetically pleasing.

DEPRECATIONS.md
```diff
@@ -4,6 +4,7 @@ This document lists deprecated constructs in the SDK, with their expected time 
 
 | API / Feature | Deprecated (warning) | Deprecated (error) | Removed | Notes |
 | --- | --- | --- | --- | --- |
+| `StreamDimens` constructor containing params: `attachmentsContentImageHeight`  | 2022.08.16 <br/>5.8.0 | 2022.08.30 ⌛ | 2022.09.13 ⌛ | This constructor has been deprecated. Use the constructor that does not contain these parameters. |
```
MarinTolic, line 7: Worded as a plural because the next PR will deprecate two more parmeters regarding Giphys when auto-sizing Giphys is implemented. 

They will be added here.

filbabic: We should probably update this given that we're shelvign giphy improvements for a bit

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/ImageAttachmentContent.kt
```diff
@@ -60,6 +62,7 @@ import io.getstream.chat.android.uiutils.extension.hasLink
  * and the onLongItemClick handler.
  */
 @OptIn(ExperimentalFoundationApi::class)
+@Suppress("MagicNumber", "LongMethod")
```
filbabic, line 65: Instead of suppressing, we should generate the baseline to include this.

I would also just pull out the aspect ratio "magic numbers" to constants, as it will be a bit more descriptive

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/ImageAttachmentContent.kt
```diff
@@ -86,9 +89,25 @@ public fun ImageAttachmentContent(
...
+                    }
+                }
+            }
+
             ImageAttachmentContentItem(
                 attachment = attachment,
-                modifier = Modifier.weight(1f),
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .aspectRatio(ratio ?: 1f),
```
filbabic, line 110: e.g. here we can say that the ratio is "equal ratio" or something

`const val EQUAL_RATIO = 1f` for example

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/ImageAttachmentContent.kt
```diff
@@ -98,7 +117,7 @@ public fun ImageAttachmentContent(
             Column(
                 modifier = Modifier
                     .weight(1f, fill = false)
-                    .fillMaxHeight(),
+                    .aspectRatio(0.5f),
```
filbabic, line 120: I'm not 100% sure what `0.5f` ratio means - so a constant would make sense here too

filbabic (approved):
Approved on the face of it, looks and works great.

I've left some minor commnets about getRekt stuff, so please resolve those!

### ☑️Reviewer Checklist

adasiewiczr (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/ImageAttachmentContent.kt
MarinTolic: Each column takes up 0.5x of the total width of the message bubble, so making it twice the size equals to a square when taking both into account-

PR #4018, opened 2022-08-10
[ISSUE-3972] switch user method

leandroBorgesFerreira:
### 🎯 Goal

Add switchUser method

### 🛠 Implementation details

Adding the methods and making some changes to disconnect method to avoid UI freeze and avoid cancelation of coroutines while the user is trying to connect.

The disconnection doesn't fully complete when the new connection starts,  we need to fix this problem. I created a ticket for this already, as it gets out of the scope of this PR. 

### 🎨 UI Changes

https://user-images.githubusercontent.com/10619102/184008266-389b7b58-4987-47a7-900b-4e769fcc3895.mov

### 🧪 Testing

Which users with and without internet

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (6)
_This was harder than I anticipated..._

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
@@ -46,23 +46,18 @@ internal class StreamLifecycleObserver(
         }
     }
 
-    suspend fun dispose() {
+    suspend fun dispose(inThread: Boolean) {
         logger.d { "[dispose] no args" }
         if (isObserving) {
-            withContext(DispatcherProvider.Main) {
+            if (inThread) {
                 lifecycle.removeObserver(this@StreamLifecycleObserver)
                 logger.v { "[dispose] unsubscribed" }
+            } else {
+                withContext(DispatcherProvider.Main) {
+                    lifecycle.removeObserver(this@StreamLifecycleObserver)
+                    logger.v { "[dispose] unsubscribed" }
+                }
             }
-        }
-        isObserving = false
-        recurringResumeEvent = false
-    }
-
-    fun disposeInThread() {
-        logger.d { "[dispose] no args" }
-        if (isObserving) {
-            lifecycle.removeObserver(this@StreamLifecycleObserver)
-            logger.v { "[dispose] unsubscribed" }
         }
         isObserving = false
         recurringResumeEvent = false
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
@@ -58,6 +58,16 @@ internal class StreamLifecycleObserver(
         recurringResumeEvent = false
     }
 
+    fun disposeInThread() {
+        logger.d { "[dispose] no args" }
+        if (isObserving) {
+            lifecycle.removeObserver(this@StreamLifecycleObserver)
+            logger.v { "[dispose] unsubscribed" }
+        }
+        isObserving = false
+        recurringResumeEvent = false
+    }
+
     override fun onResume(owner: LifecycleOwner) {
         logger.d { "[onResume] owner: $owner" }
         // ignore event when we just started observing the lifecycle
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
@@ -16,8 +16,6 @@
 
 package io.getstream.chat.android.client.socket
 
-import android.os.Handler
-import android.os.Looper
 import androidx.annotation.VisibleForTesting
 import io.getstream.chat.android.client.clientstate.DisconnectCause
 import io.getstream.chat.android.client.errors.ChatError
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
@@ -57,7 +57,6 @@ internal open class ChatSocket constructor(
     private var eventsParser: EventsParser? = null
     private var socketConnectionJob: Job? = null
     private val listeners = mutableSetOf<SocketListener>()
-    private val eventUiHandler = Handler(Looper.getMainLooper())
     private val healthMonitor = HealthMonitor(
         coroutineScope = coroutineScope,
         reconnectCallback = {
@@ -127,6 +126,7 @@ internal open class ChatSocket constructor(
                     connectionConf = null
                     networkStateProvider.unsubscribe(networkStateListener)
                     healthMonitor.stop()
+
                     callListeners { it.onDisconnected(DisconnectCause.UnrecoverableError(newState.error)) }
                 }
             }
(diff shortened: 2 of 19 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
```diff
@@ -57,7 +55,6 @@ internal open class ChatSocket constructor(
     private var eventsParser: EventsParser? = null
     private var socketConnectionJob: Job? = null
     private val listeners = mutableSetOf<SocketListener>()
-    private val eventUiHandler = Handler(Looper.getMainLooper())
```
leandroBorgesFerreira, line 60: This is useless... the events are already async. This is only making the SDK slower.

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/FlatMapCall.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.getstream.chat.android.client.utils.map
+import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+import java.util.concurrent.atomic.AtomicBoolean
+
+internal class FlatMapCall<T : Any, K : Any>(
+    private val call: Call<T>,
+    private val mapper: (T) -> Call<K>,
+) : Call<K> {
```
kanat, line 32: I wonder how `FlatMapCall` differs from `MapCall`

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/utils/Result.kt
```diff
@@ -282,6 +283,14 @@ public fun <T : Any, R : Any> Result<T>.flatMap(func: (T) -> Result<R>): Result<
     }
 }
 
+public fun <T : Any, R : Any> Result<T>.flapMapCall(func: (T) -> Call<R>): Result<R> {
+    return if (isSuccess) {
+        func(data()).execute()
```
kanat, line 288: `execute` is blocking which means we gonna block `enqueue` callback function

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/FlatMapCall.kt
```diff
@@ -0,0 +1,59 @@
...
+        canceled.set(true)
+        call.cancel()
+    }
+
+    override fun execute(): Result<K> = runBlocking { await() }
+
+    override fun enqueue(callback: Call.Callback<K>) {
+        call.enqueue { result ->
+            result.takeUnless { canceled.get() }
+                ?.flapMapCall(mapper)
```
kanat, line 46: Here `flapMapCall ` will block `this` callback until

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/FlatMapCall.kt
kanat, line 32: I do believe `FlatMapCall` requires a scope to be passed.

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/FlatMapCall.kt
```diff
@@ -0,0 +1,59 @@
...
+                ?.let(callback::onResult)
+        }
+    }
```
kanat, line 49:
```suggestion
    override fun enqueue(callback: Call.Callback<K>) {
        call.enqueue { result ->
            result.takeUnless { canceled.get() }
                ?.flapMapCall(mapper) { finalResult ->
                    finalResult.takeUnless { canceled.get() }
                        ?.let(callback::onResult)
                }
        }
    }
```

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/utils/Result.kt
```diff
@@ -282,6 +283,14 @@ public fun <T : Any, R : Any> Result<T>.flatMap(func: (T) -> Result<R>): Result<
...
+    } else {
+        Result.error(error())
+    }
+}
```
kanat, line 292:
```suggestion
public fun <T : Any, R : Any> Result<T>.flapMapCall(func: (T) -> Call<R>, callback: Call.Callback<R>) {
    if (isSuccess) {
        func(data()).enqueue {
            callback.onResult(it)
        }
    } else {
        callback.onResult(Result.error(error()))
    }
}
```

kanat: We also need somehow to cancel this child `Call<R>` if parent `Call<T>` gets cancelled

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/ThreadCall.kt
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    override fun enqueue(callback: Call.Callback<T>) {
+        thread = Thread { callback.onResult(execute()) }
+        thread?.start()
+    }
+
+    override suspend fun await(): Result<T> {
+        return execute()
+    }
```
kanat, line 46: it is blocking and not suspending. I would expect here a wrapper around `enqueue` instead

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/ThreadCall.kt
```diff
@@ -0,0 +1,51 @@
...
+) : Call<T> {
+
+    private var thread: Thread? = null
+
+    @Suppress("TooGenericExceptionCaught")
+    override fun execute(): Result<T> = try {
+        Result.success(task())
+    } catch (e: Throwable) {
+        Result.error(e)
+    }
```
kanat, line 37: as per design we're supposed to return `Call.callCanceledError()`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/StreamLifecycleObserver.kt
```diff
@@ -46,12 +46,17 @@ internal class StreamLifecycleObserver(
...
                 lifecycle.removeObserver(this@StreamLifecycleObserver)
                 logger.v { "[dispose] unsubscribed" }
+            } else {
+                withContext(DispatcherProvider.Main) {
+                    lifecycle.removeObserver(this@StreamLifecycleObserver)
+                    logger.v { "[dispose] unsubscribed" }
+                }
             }
         }
         isObserving = false
```
kanat, line 62: Can we check if this method is invoked from the `Main` thread to avoid passing any additional flags?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
```diff
@@ -297,9 +295,7 @@ internal open class ChatSocket constructor(
 
     private fun callListeners(call: (SocketListener) -> Unit) {
         synchronized(listeners) {
-            listeners.forEach { listener ->
-                eventUiHandler.post { call(listener) }
-            }
+            listeners.forEach(call)
         }
```
kanat, line 299: This might be a breaking change. In other words, this may affect those who use `ChatClient.subscribeFor...`. Customers may expect those events coming within the `Main` thread

stream-chat-android-core/src/main/java/io/getstream/chat/android/client/call/FlatMapCall.kt
leandroBorgesFerreira: 🤔  But the code is inside `enqueue{}`, which is already async. The blocking will be the same as a normal `map` function.

leandroBorgesFerreira:
Maybe ThreadCall is not necessary at all. Let me check and we can remove many of the code with comments.

leandroBorgesFerreira:
Closing in favor or #4026. If we fix the problem related to disconnection ascynchrounously, there's no need to most of the code of this PR.

PR #4026, opened 2022-08-11
[ISSUE-3972] Switch User 2

leandroBorgesFerreira:
### 🎯 Goal

Add function to switch users. `ChatClient.switchUser`.

### 🛠 Implementation details

Events of disconnection are being handled synchronously instead of async. Then the connection of a new user only happens when the SDK already fully disconnected with the previous user.

### 🎨 UI Changes

https://user-images.githubusercontent.com/10619102/184191580-2bda483d-02d3-43be-aaa8-d68fcec431fc.mov

### 🧪 Testing

Switch users with or without internet. 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -282,21 +282,6 @@ internal constructor(
...
-                        is DisconnectCause.UnrecoverableError -> {
-                            userStateService.onSocketUnrecoverableError()
-                            if (ToggleService.isSocketExperimental().not()) {
-                                socketStateService.onSocketUnrecoverableError()
-                            }
-                        }
-                    }
-                    clientState.toMutableState()?.setConnectionState(ConnectionState.OFFLINE)
-                }
                 is NewMessageEvent -> {
```
kanat, line 285: I wonder why this was extracted to a separate observer? 🤔

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
```diff
@@ -295,10 +300,14 @@ internal open class ChatSocket constructor(
         socket = null
     }
 
-    private fun callListeners(call: (SocketListener) -> Unit) {
+    private fun callListeners(forceSync: Boolean = false, call: (SocketListener) -> Unit) {
```
kanat, line 303: Maybe we could create 2 separate methods to make this more transparent?

Kinda
```kotlin
private fun callListeners(call: (SocketListener) -> Unit) {
    synchronized(listeners) {
        listeners.forEach(call)
    }
}
```
and 
```kotlin
private fun callListenersOnUiThread(call: (SocketListener) -> Unit) {
    handler.post {
        callListeners(call)
    }
}
```

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/data/user/UserRepository.kt
```diff
@@ -52,6 +52,10 @@ class UserRepository(context: Context) {
         prefs.edit().clear().apply()
     }
 
+    fun clearUserCommit() {
```
kanat, line 55: Can we add a `commit` or `async` flag to the existing `clearUser` function with a predefined default value to keep previous behavior?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/home/HomeFragmentViewModel.kt
```diff
@@ -42,6 +43,17 @@ class HomeFragmentViewModel : ViewModel() {
             user = ChatClient.instance().getCurrentUser() ?: unauthorizedUser,
```
kanat, line 43: I believe this could be replaced now with 
`ChatClient.instance().clientState.user.value ?: unauthorizedUser.`

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/data/user/UserRepository.kt
leandroBorgesFerreira: Actually we don't need this funtion

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1093,10 +1167,13 @@ internal constructor(
...
-        appSettingsManager.clear()
         _repositoryFacade = null
-        val currentJob = currentCoroutineContext()[Job]
-        scope.coroutineContext.cancelChildrenExcept(currentJob)
+        appSettingsManager.clear()
+
+        clientState.toMutableState()?.clearState()
+        clientState.toMutableState()?.setInitializionState(InitializationState.NOT_INITIALIZED)
+
+        logger.v { "[disconnectSuspend] completed('$userId')" }
```
JcMinarro, line 1176: Why did we rearranged this method? :thinking:

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
```diff
@@ -297,9 +302,13 @@ internal open class ChatSocket constructor(
...
-            listeners.forEach { listener ->
-                eventUiHandler.post { call(listener) }
-            }
+            listeners.forEach(call)
+        }
+    }
+
+    private fun callListenersOnUiThread(call: (SocketListener) -> Unit) {
+        eventUiHandler.post {
+            callListeners(call)
```
JcMinarro, line 311: Why some of the events are notified on the UI thread and other no?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: To set `NOT_INITIALIZED` only when the disconnection is fully complete.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
leandroBorgesFerreira: It is not necessary indeed.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
kanat: Tbh, I cannot understand how it may guarantee that if Socket notifies about coming events using `Handler`:
```
eventUiHandler.post { call(listener) }
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
```diff
@@ -297,8 +302,8 @@ internal open class ChatSocket constructor(
 
     private fun callListeners(call: (SocketListener) -> Unit) {
         synchronized(listeners) {
-            listeners.forEach { listener ->
-                eventUiHandler.post { call(listener) }
+            eventUiHandler.post {
+                listeners.forEach(call)
             }
         }
```
kanat, line 308: `synchronized` should be moved into the thread where iteration runs, otherwise, it won't work:
```kotlin
        eventUiHandler.post {
            synchronized(listeners) {
                listeners.forEach(call)
            }
        }
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: When I wrote the reply there were notifications with and without `Handler` and this one was not using then. But you're right, we can't use `Handler` if we would like to make sure that the notification has already happened.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
leandroBorgesFerreira: Fixed

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
leandroBorgesFerreira: Sorry, it is actually necessary. Using `eventUiHandler.post {` forces the listeners to be async. This can't happen in the disconnection, because that would cause that we return `Result.success` before the socket is completed disconnected... the listeners would run after the disconnection is complete, which would allow users to try to connect before they are allowe to.

JcMinarro: Instead of using `eventUiHandler.post {}`, you can use `coroutineScope.launch(DispatchersProvider.Main) { }` and, because we are cancelling all coroutineScopes of our SDK, it will be cancelled when `ChatClient.disconnect()` is called.

leandroBorgesFerreira: Indeed, this works. I tested in several situations and the disconnection always happens after this events are completed.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: I did some testing and using `subscribeSuspend` makes the socket listener to run after the Result.success is complete. But that doesn't happen when using `subscribe`, using a coroutine call makes it wait to the listener to complete.

leandroBorgesFerreira:
I kept the post using handler inside `socket/ChatSocket` to avoid too many changes in this PR. This socket will be substituted, then the new implementation will use cotourines anyway.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Show rounded avatars on Push Notification when `MessagingStyleNotificationHandler` is used. [#4059](https://github.com/GetStream/stream-chat-android/pull/4059)
 
 ### ✅ Added
+- Method to switch between users `ChatClient.switchUser`. Can user use when switching between user to simplify code for disconnecting and connecting to the SDK. [#4018](https://github.com/GetStream/stream-chat-android/pull/4018)
```
kanat, line 20:
```suggestion
- Method to switch between users `ChatClient.switchUser`. Can be used for switching between users to simplify code for disconnecting and connecting to the SDK. [#4018](https://github.com/GetStream/stream-chat-android/pull/4018)
```

kanat (approved):
LGTM

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/home/HomeViewModel.kt
```diff
@@ -92,11 +92,15 @@ class HomeViewModel(
         when (action) {
             is UiAction.LogoutClicked -> {
                 viewModelScope.launch {
-                    chatClient.disconnect(false).await()
+                    ChatClient.instance().disconnect(false).await()
```
filbabic, line 95: Any reason why we don't just use the chatClient like before?

filbabic (approved):
From the UI side it works fine! Just left a couple of comments :]

### ☑️Reviewer Checklist

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/home/HomeViewModel.kt
leandroBorgesFerreira: No, let me revert this.

PR #4028, opened 2022-08-11
[ISSUE-3710] Thread query listener impl separation

leandroBorgesFerreira:
### 🎯 Goal

Prepare ThreadQueryListenerImpl to handle both state and/or persistence accordingly with how users configure the SDK.

### 🛠 Implementation details

- `ThreadLogic` is no longer the implementation of `ThreadQueryListener`. There's already a specific class for that which is `ThreadQueryListenerImpl` (or `ThreadQueryListenerFull`) now.  `ThreadLogic` should only handle the logic of threads instead of also being a lister of ChatClient requests. This way each class becomes more specialized. 

- `ThreadQueryListenerFull` handles both state and persistence, but its properties are optional.

I don't believe this class should be separated into database and state completely because the code uses persistence to update the state if the information lives only inside the database at the moment the listener is called. Separating it would cause more requests to API, because the state won't be aware that information can be fetched from persistence. 

As a solution, the properties that handle each case are now optional and the class adapts itself to the configuration of the SDK. If no persistence is provided by the user, this class only interacts with the state and updates it with the data that comes from the API. If only the database is provided, the state is not propagated and only the database receives updates. If none of the dependencies is provided, this class doesn't do anything and becomes a NOOP listener. 

### 🧪 Testing

Check the threads inside the channels.

#### General

#### Code & documentation
- ~[ ] Changelog is updated with client-facing changes~. There's no client facing changes
- [ ] New code is covered by unit tests. **Let's discuss the approach first**.

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (7)
_I didn't find a good giphy for this pr..._

### DIFF
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerFull.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerFull.kt
@@ -18,8 +18,10 @@ package io.getstream.chat.android.offline.plugin.listener.internal
 
 import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.errors.ChatError
+import io.getstream.chat.android.client.extensions.internal.users
 import io.getstream.chat.android.client.models.Message
-import io.getstream.chat.android.client.persistance.repository.RepositoryFacade
+import io.getstream.chat.android.client.persistance.repository.MessageRepository
+import io.getstream.chat.android.client.persistance.repository.UserRepository
 import io.getstream.chat.android.client.plugin.listeners.ThreadQueryListener
 import io.getstream.chat.android.client.utils.Result
 import io.getstream.chat.android.client.utils.onSuccessSuspend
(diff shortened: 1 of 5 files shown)


stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerFull.kt
```diff
@@ -0,0 +1,135 @@
...
+        } else {
+            val result = chatClient.getMessage(messageId).await()
+            if (result.isSuccess) {
+                val message = result.data()
+                threadLogic?.upsertMessage(result.data())
+                userRepository?.insertUsers(message.users())
+                messageRepository?.insertMessage(message)
+                Result.success(Unit)
+            } else {
+                Result(result.error())
```
kanat, line 81: There results are not needed

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerFull.kt
```diff
@@ -0,0 +1,135 @@
...
+    override suspend fun onGetRepliesRequest(messageId: String, limit: Int) {
+        val threadLogic = logic?.thread(messageId)
+
+        threadLogic?.setLoading(true)
+        val messages = messageRepository?.selectMessagesForThread(messageId, limit)
+        val parentMessage = messages?.firstOrNull { it.id == messageId } ?: threadLogic?.getMessage(messageId)
+
+        if (parentMessage != null) {
+            messages?.let { threadLogic?.upsertMessages(it) }
+            Result.success(Unit)
```
kanat, line 71: same here

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerFull.kt
```diff
@@ -0,0 +1,135 @@
...
+        } else {
+            Result.success(Unit)
+        }
+    }
+
+    override suspend fun onGetRepliesRequest(messageId: String, limit: Int) {
+        val threadLogic = logic?.thread(messageId)
+
+        threadLogic?.setLoading(true)
+        val messages = messageRepository?.selectMessagesForThread(messageId, limit)
```
kanat, line 66: It feels like in the case of State of this Listener we should look for the message inside the `ChannelState` first.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerFull.kt
```diff
@@ -0,0 +1,135 @@
...
+ * @param logic [LogicRegistry] Optional class to handle state updates
+ * @param messageRepository [MessageRepository] Optional to handle database updates related to messages
+ * @param userRepository [UserRepository]  Optional to handle database updates related to user
+ * @param chatClient [ChatClient]
+ */
+internal class ThreadQueryListenerFull(
+    private val logic: LogicRegistry?,
+    private val messageRepository: MessageRepository?,
+    private val userRepository: UserRepository?,
+    private val chatClient: ChatClient,
```
kanat, line 45: I wonder if we a pass just a suspending function here to avoid having unnecessary dependency:
```kotlin
getRemoteMessage: (messageId: String) -> Message
```

kanat (approved):
LGTM

PR #4056, opened 2022-08-17
[ISSUE-3702]  Separate HideChannelListenerImpl

leandroBorgesFerreira:
### 🎯 Goal

Separate HideChannelListenerImpl into database and state.

### 🛠 Implementation details

2 new classes and onde aggregator that should be only temporary

### 🧪 Testing

Hide some channels

#### General

#### Code & documentation
- [ ] ~Changelog is updated with client-facing changes~. No client-facing changes
- [ ] ~Affected documentation updated (KDocs, docusaurus, tutorial)~

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (1)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -166,6 +151,29 @@ public class StreamOfflinePluginFactory(
         ).also { offlinePlugin -> cachedOfflinePluginInstance = offlinePlugin }
     }
 
+    private fun getHideChannelListener(chatClient: ChatClient, statePlugin: StatePlugin): HideChannelListener {
+        val hideChannelListenerDatabase = HideChannelListenerDatabase(
+            channelRepository = chatClient.repositoryFacade,
+            messageRepository = chatClient.repositoryFacade
```
kanat, line 157: Shouldn't we pass the exact repositories instead of `repositoryFacade`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/factory/StreamOfflinePluginFactory.kt
```diff
@@ -190,6 +198,21 @@ public class StreamOfflinePluginFactory(
...
     }
 
+    private fun getDeleteMessageListenerDatabase(
+        chatClient: ChatClient,
+        statePlugin: StatePlugin,
+    ): DeleteMessageListener {
+        val deleteMessageListenerDatabase = DeleteMessageListenerDatabase(
+            clientState = chatClient.clientState,
+            messageRepository = chatClient.repositoryFacade,
+            userRepository = chatClient.repositoryFacade
```
kanat, line 208: same here

kanat (approved):
LGTM

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/factory/StreamOfflinePluginFactory.kt
JcMinarro: The `DeleteMessageListenerDatabase` constructor requires the concrete repository, it should be fine to provide the `RepositoryFacade` here, because it implement all of our `Repositories`.
We don't have a way to extract a concrete repository from our `RepositoryFacade`

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/HideChannelListenerState.kt
```diff
@@ -37,7 +33,7 @@ internal class HideChannelListenerImpl(
     ): Result<Unit> = validateCidWithResult(Pair(channelType, channelId).toCid()).toUnitResult()
 
     override suspend fun onHideChannelRequest(channelType: String, channelId: String, clearHistory: Boolean) {
-        logic.channel(channelType, channelId).setHidden(true)
+        logic.channel(channelType, channelId).stateLogic().setHidden(true)
```
adasiewiczr, line 36: Can we change `setHidden` method name? I think it's misleading because it either hides or shows the channel based on the flag. Maybe it would be better to call it `changeHiddenValue`?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/factory/StreamOfflinePluginFactory.kt
kanat: The thing is that @leandroBorgesFerreira has already decomposed `RepositoryFacade` calls into a set of separate repositories calls. That's why it has no sense to pass the `RepositoryFacade` instance because otherwise, it will lead to duplicate calls to DB.

leandroBorgesFerreira: 🤔  I think it won't make duplicated class to DB. In the end, we are just calling RepositoryFacade in different alias. I need to use the Repositories instead of the RepositoryFacade because I need interfaces that are inside LLC module instead of a implementation that live inside a more specific module

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/HideChannelListenerState.kt
leandroBorgesFerreira: I changed to toggleHidden

adasiewiczr (approved):

kanat (approved):

PR #4072, opened 2022-08-22
[4070] Decrement online count

### LINKED ISSUE #4070: Online count doesn't get decremented
**Describe the bug**
 
The online member count doesn't get decremented when a member leaves the channel.

We rely on the `watcher_count` field to display the number of online members on the chat screen. When somebody leaves a group we receive `member.removed` and `channel.updated` events which don't contain the updated field.

**To Reproduce**
1. Go to the chat screen
2. Observe the information in the toolbar (e.g. 3 members, 3 online)
3. From another client log in as a different user and leave the channel

**Expected behavior**
"2 members, 2 online"

**Actual behavior**
"2 members, 3 online"

bychkovdmitry:
### 🎯 Goal

Closes #4070 

The online member count doesn't get decremented when a member leaves the channel.

### 🧪 Testing

Go to the chat screen:
1. Observe the information in the toolbar (e.g. 3 members, 3 online)
2. From another client log in as a different user and leave the channel

Expected behavior:
"2 members, 2 online"

The behavior before the fix:
"2 members, 3 online"

#### General

#### Code & documentation

### ☑️Reviewer Checklist


stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
```diff
@@ -346,12 +346,19 @@ internal class ChannelStateLogicImpl(
...
+        val user = member.user
+
+        mutableState.rawMembers = mutableState.rawMembers - user.id
 
         mutableState.setMembersCount(mutableState.membersCount.value - 1)
+
+        if (user.online) {
+            deleteWatcher(user)
+            setWatcherCount(mutableState.watcherCount.value - 1)
+        }
```
kanat, line 361: Hey @bychkovdmitry,

I wonder if we could use a `member` object available in `mutableState.rawMembers`:
```kotlin
    override fun deleteMember(userId: String) {
        val member = mutableState.rawMembers[userId]
        val user = member.user
        ...
```

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
```diff
@@ -346,12 +346,19 @@ internal class ChannelStateLogicImpl(
...
+     * @param member The member to be removed.
      */
-    override fun deleteMember(userId: String) {
-        mutableState.rawMembers = mutableState.rawMembers - userId
+    override fun deleteMember(member: Member) {
+        val user = member.user
+
+        mutableState.rawMembers = mutableState.rawMembers - user.id
 
         mutableState.setMembersCount(mutableState.membersCount.value - 1)
```
JcMinarro, line 356: Why do we have a separate count for members?
Shouldn't we delegate it to the size of the collection?

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
JcMinarro, line 361: Same here, If we already have a list of watcher, we shouldn't need to have a separate counter for something our collection already know.
On the other hand, as far as we are using `StateFlow`, this property (The counter) could be a "map" of the other property (The collection) and it will be always up-to-date

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
bychkovdmitry: There can be 10 loaded members for a channel with 100 members, so we should have a separate counter.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/internal/ChannelStateLogicImpl.kt
bychkovdmitry: I think the list of watchers is also paginated. There can be 1000 watchers for a live stream channel, but I don't think all the watchers will be returned for the channel.

bychkovdmitry: I'm not 100% sure that the current user will be in the first page of members

adasiewiczr: There is no guarantee for that

adasiewiczr (approved):
### ☑️Reviewer Checklist

PR #4102, opened 2022-08-26
[ISSUE-4060] Fixing event time precision

leandroBorgesFerreira:
### 🎯 Goal

Increase precision of time sent to backend for the `/sync` endpoint.

### 🛠 Implementation details

Added ExactDate and ExactDateAdapter to internally serialise the String of date inside Json to both Date and keep the original String. 

### 🎨 UI Changes
<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/187240796-c211805f-3e90-4608-8b07-9a3032ee9be7.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/187240774-aecefe24-1176-4347-b122-3373d6d8e6b1.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Check that sync is emitting events with the exact same time of the last sync. 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy
_Yes, we should!_


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -386,5 +387,8 @@ internal interface ChatApi {
     @CheckResult
     fun getSyncHistory(channelIds: List<String>, lastSyncAt: Date): Call<List<ChatEvent>>
 
+    @CheckResult
+    fun getSyncHistory(channelIds: List<String>, lastSyncAt: ExactDate): Call<List<ChatEvent>>
```
JcMinarro, line 391: `ChatApi` is an internall interface, we don't need to support an extra method here. 
We can change method signature to get an String instead with the `lastSyncAt`.
If we do this change, we don't really need `ExactDate` neither `ExactDataUpstream` and its adapters

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/requests/SyncHistoryRequest.kt
```diff
@@ -17,10 +17,10 @@
...
 
 import com.squareup.moshi.JsonClass
-import java.util.Date
+import io.getstream.chat.android.client.api2.model.dto.utils.internal.ExactDateUpstream
 
 @JsonClass(generateAdapter = true)
 internal data class SyncHistoryRequest(
     val channel_cids: List<String>,
-    val last_sync_at: Date,
+    val last_sync_at: ExactDateUpstream,
```
JcMinarro, line 25: It can be an String, and it simplify our development

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
```diff
@@ -59,6 +61,8 @@ internal class MoshiChatParser : ChatParser {
     private val moshi: Moshi by lazy {
         Moshi.Builder()
             .addAdapter(DateAdapter())
+            .addAdapter(ExactDateAdapter())
+            .addAdapter(ExactDateUpstreamAdapter())
```
JcMinarro, line 65: This adapter shouldn't be needed if we remove `ExactDateUpstream` class

leandroBorgesFerreira: We need it for downstream. In order to create the `SyncState` with both `lastSyncedAt` and `rawLastSyncedAt` we need a different adapter that covers both `Date` and `String` otherwise we would be forced to choose between one of then.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
leandroBorgesFerreira: We still need `ExactDate` for downstream

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -384,7 +384,7 @@ internal interface ChatApi {
     fun translate(messageId: String, language: String): Call<Message>
 
     @CheckResult
-    fun getSyncHistory(channelIds: List<String>, lastSyncAt: Date): Call<List<ChatEvent>>
+    fun getSyncHistory(channelIds: List<String>, lastSyncAt: String): Call<List<ChatEvent>>
```
kanat, line 387: This seems to be a breaking change in public API.
Can we add an old function version with the `Date` param as well?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2476,12 +2480,39 @@ internal constructor(
...
+     * @return Executable async [Call] responsible for obtaining missing events.
+     */
+    @CheckResult
+    public fun getSyncHistory(
+        channelsIds: List<String>,
+        lastSyncAt: String,
+    ): Call<List<ChatEvent>> {
+        val parsedDate = streamDateFormatter.parse(lastSyncAt) ?: throw IllegalArgumentException(
+            "The string for data: $lastSyncAt could not be parsed for format: ${streamDateFormatter.datePattern} "
+        )
```
kanat, line 2508: maybe we could make `getSyncHistory` with a `String` param as an `internal` function to avoid having this validation by parsing the input date? This is because we can internally guarantee we always gonna pass a valid `rawDate`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
leandroBorgesFerreira: This API is not public. There's no problem in keeping both methods, but this is conflicting with @JcMinarro's comment. Let's decide which one we prefer.

kanat: My bad, cool then 👍

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2476,12 +2480,39 @@ internal constructor(
...
+     * @return Executable async [Call] responsible for obtaining missing events.
+     */
+    @CheckResult
+    internal fun getSyncHistory(
+        channelsIds: List<String>,
+        lastSyncAt: String,
+    ): Call<List<ChatEvent>> {
+        val parsedDate = streamDateFormatter.parse(lastSyncAt) ?: throw IllegalArgumentException(
+            "The string for data: $lastSyncAt could not be parsed for format: ${streamDateFormatter.datePattern} "
+        )
```
kanat, line 2508: I'm wondering if we could replace exception throwing with `ErrorCall`
```suggestion
        val parsedDate = streamDateFormatter.parse(lastSyncAt) ?: return ErrorCall(scope, ChatError(
            "The string for data: $lastSyncAt could not be parsed for format: ${streamDateFormatter.datePattern}"
        ))
```

leandroBorgesFerreira: 🤔

leandroBorgesFerreira: We need to parse to have the variable as a parameter first. Let's return in the ErrorCall

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
leandroBorgesFerreira: It is not possible to make it internal because this is uses in state module, but I put the annotation of `InternalStreamChatApi`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/utils/internal/ExactDate.kt
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.api2.model.dto.utils.internal
+
+import java.util.Date
+
+/**
+ * DTO to keep serialized date and also the original Date as Stirng as sent by backend.
```
adasiewiczr, line 22: Can we add params description? Especially the format of `rawDate` so it's easy to find in the future

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 ### 🐞 Fixed
 
 ### ⬆️ Improved
+- Improving precision in time in the endpoint that syncs information between SDK and Stream's backend to make sure that undesired events are not coming due to a incorrect round down in time or desired events are not being ignored due to a incorrect round up of time when serializing/desirializing time in the SDK. [#4102](https://github.com/GetStream/stream-chat-android/pull/4102)
```
adasiewiczr, line 17: I would also mention adding `ChannelEvent:: rawCreatedAt` as a separate point

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2476,12 +2479,43 @@ internal constructor(
...
+
+        return api.getSyncHistory(channelsIds, stringDate)
             .withPrecondition(scope) {
                 checkSyncHistoryPreconditions(channelsIds, lastSyncAt)
             }
     }
 
+    /**
+     * Returns all events that happened for a list of channels since last sync (while the user was not
+     * connected to the web-socket). LastSyncAt is in the format yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'. Use this version of
```
adasiewiczr, line 2492:
```suggestion
     * connected to the web socket). [lastSyncAt] is in _yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'_ format.  Use this version of
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2476,12 +2479,43 @@ internal constructor(
...
+     * getSyncHistory when high precision is necessary.
+     *
+     * @param channelsIds The list of channel CIDs. Cannot be empty.
+     * @param lastSyncAt The last time the user was online and in sync. Shouldn't be later than 30 days.
+     *
+     * @return Executable async [Call] responsible for obtaining missing events.
+     */
+    @CheckResult
+    @InternalStreamChatApi
```
adasiewiczr, line 2501: Does it make sense to make it public? 🤔 Maybe we should also think about deprecating the old `getSynHistory` with `Date`?

leandroBorgesFerreira: making it public can be useful for customers that want to use a greater precision. About deprecating the old one... I would say that using both is preferable, because the string version may be considered hard to use if the customer only wants to generate the time locally.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
```diff
@@ -605,6 +647,7 @@ public data class ConnectedEvent(
 public data class ConnectingEvent(
     override val type: String,
     override val createdAt: Date,
+    override val rawCreatedAt: String? = null,
```
adasiewiczr, line 650: Do we need the default value here and below?

adasiewiczr (approved):
LGTM
### ☑️Reviewer Checklist

PR #4126, opened 2022-09-01
[ISSUE-4109] Using Repositories implementations instead of facade

leandroBorgesFerreira:
### 🎯 Goal

Using implementation of repositories instead of `RepositoryFacade`.

### 🧪 Testing

Test the persistence of the SDK

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (1)
_A facade_

### DIFF
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/factory/StreamStatePluginFactory.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/factory/StreamStatePluginFactory.kt
@@ -135,7 +135,11 @@ public class StreamStatePluginFactory(
         }
 
         val stateRegistry = StateRegistry.create(
-            scope.coroutineContext.job, scope, clientState.user, repositoryFacade, repositoryFacade.observeLatestUsers()
+            scope.coroutineContext.job,
+            scope,
+            clientState.user,
+            repositoryFacade.messageRepository,
+            repositoryFacade.observeLatestUsers()
         )
         val logic = LogicRegistry.create(
             stateRegistry = stateRegistry,
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/RepositoryFacade.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/RepositoryFacade.kt
@@ -37,14 +37,14 @@ import java.util.Date
 @InternalStreamChatApi
 @SuppressWarnings("LongParameterList")
 public class RepositoryFacade private constructor(
-    private val userRepository: UserRepository,
-    private val configsRepository: ChannelConfigRepository,
-    private val channelsRepository: ChannelRepository,
-    private val queryChannelsRepository: QueryChannelsRepository,
-    private val messageRepository: MessageRepository,
-    private val reactionsRepository: ReactionRepository,
-    private val syncStateRepository: SyncStateRepository,
-    private val attachmentRepository: AttachmentRepository,
+    public val userRepository: UserRepository,
+    public val configsRepository: ChannelConfigRepository,
+    public val channelsRepository: ChannelRepository,
+    public val queryChannelsRepository: QueryChannelsRepository,
+    public val messageRepository: MessageRepository,
+    public val reactionsRepository: ReactionRepository,
+    public val syncStateRepository: SyncStateRepository,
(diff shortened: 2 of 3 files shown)


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/factory/StreamStatePluginFactory.kt
```diff
@@ -135,7 +135,11 @@ public class StreamStatePluginFactory(
         }
 
         val stateRegistry = StateRegistry.create(
-            scope.coroutineContext.job, scope, clientState.user, repositoryFacade, repositoryFacade.observeLatestUsers()
+            scope.coroutineContext.job,
+            scope,
+            clientState.user,
+            repositoryFacade.messageRepository,
+            repositoryFacade.observeLatestUsers()
```
skydoves, line 142: It would be better if we could define the argument names explicitly here.

skydoves (approved):
Looks good to me!

JcMinarro (changes requested):
The `RepositoryFacade` was created because when we insert some entities like `channel` (eg: `ChannelRepository::insertChannel`), this entity contains other entities that need to be inserted in a different repository, like for example `user` (eg: `UserReposotory::insertUsers`).
If we use the repositories injected into our `RepositoryFacade` instead of the facade itself, we lose this logic, and it needs to be duplicated every time we use those methods.

Our `RepositoryFacade` implements all our Repository Interfaces, so when a class asks for a concrete repository and we use the `RepositoryFacade`, the class will only be able to access the methods of the concrete repository, and not the rest of our `RepositoryFacade`. There is no issue to use our `RepositoryFacade` there.

leandroBorgesFerreira:
Hi @JcMinarro. 

One problem that we have with the facade is that we use some overrides in the original implementations that cause duplication when using the interfaces. For example: 
```
userRepository.insertUsers(messageToBeDeleted.users())
messageRepository.insertMessage(messageToBeDeleted, true)
```

This code would call insert user twice because of the implementation of insertMessage of RepositoryFacade: 
```
override suspend fun insertMessage(message: Message, cache: Boolean) {
        insertUsers(message.users())
        messageRepository.insertMessage(message, cache)
}
```

Using `RepositoryFacade` brings this problem with it. Another solution, instead of using the implementations inside the facade would be to erase this method and make sure the other classes are not having this problem (pehaps clean all override), but it is a bigger change. 

What do you think?

JcMinarro:
> Hi @JcMinarro.
> 
> One problem that we have with the facade is that we use some overrides in the original implementations that cause duplication when using the interfaces. For example:
> 
> ```
> userRepository.insertUsers(messageToBeDeleted.users())
> messageRepository.insertMessage(messageToBeDeleted, true)
> ```
> 
> This code would call insert user twice because of the implementation of insertMessage of RepositoryFacade:
> 
> ```
> override suspend fun insertMessage(message: Message, cache: Boolean) {
>         insertUsers(message.users())
>         messageRepository.insertMessage(message, cache)
> }
> ```
> 
> Using `RepositoryFacade` brings this problem with it. Another solution, instead of using the implementations inside the facade would be to erase this method and make sure the other classes are not having this problem (pehaps clean all override), but it is a bigger change.
> 
> What do you think?

Just that is the case.
Whenever a channel/message is added, we need to add all related users with this entity.
If we forgot to insert them in some place, our SDK could crash when the entity is recovered (And indeed, it was crashing in the past until we implemented the facade and handled this case)

If someone inserts twice the same user, it doesn't create a duplicate.
The code you posted is totally fine.

kanat:
Hey @JcMinarro 👋

The main issue is that we're doing unnecessary `IO` operations, at least 2 times more than it actually requires 🤷‍♂️
So I suppose by `duplicate` @leandroBorgesFerreira meant not the entities but `IO` operations.

kanat (approved):
LGTM

JcMinarro:
> Hey @JcMinarro wave
> 
> The main issue is that we're doing unnecessary `IO` operations, at least 2 times more than it actually requires man_shrugging So I suppose by `duplicate` @leandroBorgesFerreira meant not the entities but `IO` operations.

Those unnecessary `IO` operations can be avoided by removing the `insertUsers()` call from outside of the `repositoryFacade` invocation, because they are not needed.
But we can't allow other users to insert corrupt data to our DB

PR #4133, opened 2022-09-07
[4130] Implement file sharing

### LINKED ISSUE #4130: A file downloading procedure should be implemented in order to share files
### Context & proposal

We need to download videos in order to share them, however we currently don't have an API to do so.

### Acceptance criteria

- [ ] Implement a file download API.
- [ ] Use the new download API to implement a user friendly method to share files inside `StreamFileUtil`

MarinTolic:
### 🎯 Goal

closes #4130

### 🛠 Implementation details

- Added a new API specialized for file downloads
- Created a function for downloading files to cache in `StreamFileUtils`
- Created a way of clearing the cache

### 🧪 Testing

<details>

<summary>Patch that switches image sharing to the file sharing method</summary>

```
Index: stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/ImagePreviewActivity.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/ImagePreviewActivity.kt b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/ImagePreviewActivity.kt
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/ImagePreviewActivity.kt	(revision 0869ffdefd36fa328b2c7ca7e05befb9b54f7e00)
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/ImagePreviewActivity.kt	(date 1663258893681)
@@ -21,6 +21,7 @@
 import android.content.Intent
 import android.net.Uri
 import android.os.Bundle
+import android.widget.Toast
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
@@ -738,15 +739,19 @@
      */
     private fun onShareImageClick(attachment: Attachment) {
         lifecycleScope.launch {
-            val uri = StreamImageLoader.instance().loadAsBitmap(
+            val result = StreamImageLoader.instance().loadAsBitmap(
                 context = applicationContext,
                 url = attachment.imagePreviewUrl!!
             )?.let {
-                StreamFileUtil.writeImageToSharableFile(applicationContext, it)
+                StreamFileUtil.writeFileToShareableFile(applicationContext, attachment)
             }
 
-            if (uri != null) {
-                shareImage(uri)
+            result?.also {
+                if (it.isSuccess) {
+                    shareImage(it.data())
+                } else {
+                    Toast.makeText(applicationContext, result.error().cause?.message, Toast.LENGTH_SHORT).show()
+                }
             }
         }
     }
@@ -902,6 +907,14 @@
         }
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+        val result = StreamFileUtil.clearStreamCache(context = applicationContext)
+        if(result.isError){
+            Toast.makeText(applicationContext, result.error().cause?.message, Toast.LENGTH_SHORT).show()
+        }
+    }
+
     public companion object {
         /**
          * The column count used for the image gallery.
```

</details>

1. Apply the patch from above
2. Open the compose app
3. Open the image gallery by clicking on a single or a group of image attachments (not file attachments)
4. Click share (bottom left corner)
5. Open the device explorer in Android Studio and navigate to the sample app folder
6. Check for existence of a folder called `stream_cache` inside the normal  app cache folder
7. Share a few files, the folder should start having more cached files
8. Close the image gallery
9. Check `stream_cache`, it should now be empty but not deleted.

#### General

#### Code & documentation
- [ ] ~~Changelog is updated with client-facing changes~~ (No client facing changes)

### ☑️Reviewer Checklist

### 🎉 GIF

gif


stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +58,54 @@ public object StreamFileUtil {
...
+     * @param attachment the attachment to be downloaded.
+     *
+     * @return The [Uri] that represents the path to the downloaded file.
+     */
+    public suspend fun writeFileToShareableFile(context: Context, attachment: Attachment): Uri? {
+        val result = runCatching {
+            val attachmentName = (attachment.url ?: attachment.assetUrl)?.hashCode()
+            val fileName = attachmentName.toString() + attachment.name
+
+            val file = File(context.cacheDir, fileName)
```
MarinTolic, line 78: When do we clean this cache?
I see a couple of ways of doing it.

### 1. Clean the cache when the gallery activity closes

This way we clean it regularly, and if we miss a lifecycle event it's not a bit deal because the chances are we'll clean it up the next time.

Con is that going back and reopening the gallery will make the file have to be re downloaded.

### 2. Expose cache directories to users and warn users about cleaning them in the docs

The pros are that we are absolved of having to manage cache manually, which can be tricky for an SDK library and that we give the users the option to clean these manually.

The con is obvious, with the fact that they have the option comes the responsibility to do so and the lack of a plug'n'play feeling.

MarinTolic: We could have a subfolder in the cache directory called `StreamCache` that would be exposed for cleaning.

JcMinarro: The first option is a really bad experience for the final user.
The second option aport almost nothing to our implementation, because by default we are not exposing any "clearign mechanism"

adasiewiczr: I like @JcMinarro idea, exposing `CacheManager` or `CachePolicy` with the default implementation sounds like a good plan

MarinTolic: @JcMinarro  @adasiewiczr  

Yup, I think we are all on the same page here.

That was essentially the initial idea, but I expressed it poorly here. Not to expose the path itself but rather cache actions through normal functions such as `saveFile`, `deleteCache`, etc.

As discussed in today's meeting, we have a relatively clear heading.

Thanks for the input guys! <3

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +59,127 @@ public object StreamFileUtil {
...
+            val streamCacheDir = getOrCreateStreamCacheDir(context, onError)
+            val attachmentName = (attachment.url ?: attachment.assetUrl)?.hashCode()
+            val fileName = attachmentName.toString() + attachment.name
+
+            val file = File(streamCacheDir, fileName)
+
+            // When File.createNewFile returns false it means that the file already exists.
+            // We then check the hash name equality to confirm it's the same file and check file size
+            // equality to make sure we've completed the download successfully.
+            if (!file.createNewFile() &&
```
MarinTolic, line 145: Unfortunately, the BE currently resizes images without updating the attachment file size, this means that this method is not reliable.

However, they have been made aware of this issue and will update the field in the future.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +59,127 @@ public object StreamFileUtil {
...
+            onError(e)
+            null
+        }
+    }
+
+    /**
+     * Deletes all the content contained within the
+     * Stream cache directory.
+     *
+     *  @param context The [Context] necessary to perform
```
adasiewiczr, line 93: Something weird happened with formatting here

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +59,127 @@ public object StreamFileUtil {
...
+     * @return The Stream cache directory if it exists or its creation
+     * was successful, false otherwise.
+     */
+    @Suppress("TooGenericExceptionCaught")
+    private fun getOrCreateStreamCacheDir(
+        context: Context,
+        onError: (exception: Exception) -> Unit = {},
+    ): File? {
+        return try {
+            val streamCacheDir = File(context.cacheDir, streamCacheDirName)
```
adasiewiczr, line 79: Could we call `mkdir` in also?
Something like this
```
return try {
     File(context.cacheDir, streamCacheDirName).also { streamCacheDir -> 
         streamCacheDir.mkdir()
     }
}
```

adasiewiczr (commented):
Added two small comments but looks good besides that!

### ☑️Reviewer Checklist

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +59,127 @@ public object StreamFileUtil {
...
+            streamCacheDir.mkdir()
```
JcMinarro, line 80:
```suggestion
            streamCacheDir.mkdirs()
```

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +59,127 @@ public object StreamFileUtil {
...
+            directory.listFiles()?.forEach {
+                it.delete()
+            }
+
+            true
+        } catch (e: Exception) {
+            onError(e)
+            false
+        }
+    }
```
JcMinarro, line 116: public fun clearStreamCache(context: Context): Boolean = File(context.cacheDir, streamCacheDirName).deleteRecursively()

JcMinarro (commented):
It works as expected on Xiaomi Mi Mix 2S

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
MarinTolic: Implemented the suggestion.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
MarinTolic: Implemented.

That's a really good suggestion.
I was not aware of the existence of `mkdirs()` and the difference between the two.

Thank you!

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
MarinTolic: Fixed.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
MarinTolic: I don't necessarily want to delete the folder as well.
It's a folder that might end up being used relatively often and I'd prefer it stay here even when it's empty.

If you see any downside with that approach, please let me know.

JcMinarro: If for some reason a folder is created in this folder and it is not empty, this clear method will fails and it won¡t be able to clean the cache at all.

JcMinarro: Reviewing our code, I can see we are silencing the exception.
On this method, you are catching whatever Exception (Which exception could happen here? Why don't we have a more clear one?) but, because the `onError` function has a default empty implementation we are not handling it.
Whenever we have an error on this method, we are not going to be notified and think all works fine.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +59,126 @@ public object StreamFileUtil {
...
+        context: Context,
+        attachment: Attachment,
+        onError: (exception: Exception) -> Unit = {},
+    ): Uri? {
+        val result = runCatching {
+            val streamCacheDir = getOrCreateStreamCacheDir(context, onError)
+            val attachmentName = (attachment.url ?: attachment.assetUrl)?.hashCode()
+            val fileName = attachmentName.toString() + attachment.name
+
+            val file = File(streamCacheDir, fileName)
```
JcMinarro, line 139: If `getOrCreateStreamCacheDir()` fails, we will be notified twice on `onError` callback, one because it couldn't be created (But the execution wasn't stopped) and another one when the File will be trying to be created (There is no parent folder....)
I don't see any advantage of having `onError` functions (Even when it is empty implementation)
We should:
1 -. Throw exception (We should be clear about which exceptions can be throw here)
2-. Return an Either type with the `Uri` value or `Exception`

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
MarinTolic: Adopted `deleteRecursively()` and a switch to using `Result<T>` in order to make exception handling clearer, as discussed in the meeting.

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
MarinTolic: Removed the `onError` functions and adopted using `Result<T>` in order to make exception handling clearer, as discussed in the meeting.

JcMinarro (approved):
LGTM

stream-chat-android-ui-common/src/main/kotlin/com/getstream/sdk/chat/StreamFileUtil.kt
```diff
@@ -54,4 +61,144 @@ public object StreamFileUtil {
...
+        context: Context,
+        attachment: Attachment,
+    ): Result<Uri> {
+        val runCatching = kotlin.runCatching {
+            val getOrCreateCacheDirResult = getOrCreateStreamCacheDir(context)
+            if (getOrCreateCacheDirResult.isError) return Result(error = getOrCreateCacheDirResult.error())
+
+            val streamCacheDir = getOrCreateCacheDirResult.data()
+
+            val attachmentName = (attachment.url ?: attachment.assetUrl)?.hashCode()
```
filbabic, line 149: we should name this: nameHashCode or attachmentHash or something.

It doens't make sense to put this in the same line:
```
 val fileName = CACHED_FILE_PREFIX + attachmentName.toString() + attachment.name
```

bychkovdmitry (approved):
Works as described!

Issue #4136, opened 2022-09-07
Refactor ThreadMutableState logic


JcMinarro:
### Context & proposal
`ThreadMutableState` shouldn't expose its internal maps/mutableStateFlow

### Acceptance criteria

To accept this task we need to complete the following items:

- [x] Internal MutableStateFlow shouldn't be exposed, and only update then by methods that perform the need logic
- [x] Remove unneeded `ThreadMutableState` interface
- [x] Remove `toMutableState()` extension

Issue #3710, opened 2022-06-19
Separate ThreadQueryListenerImpl in state and database 


leandroBorgesFerreira:
# Context & proposal
Separate the state and database parts of ThreadQueryListenerImpl and move them to separate locations.

# Acceptance criteria
The new implementations of ThreadQueryListenerImpl should not have both state and database logic together.

leandroBorgesFerreira:
There no need to change any code. This class is already pure state.

leandroBorgesFerreira:
**Correction:** This class is not just pure state, there is also database change in the ThreadLogic.

leandroBorgesFerreira:
I'll wait for the PR: https://github.com/GetStream/stream-chat-android/pull/3959 to be complete to change this one

PR #4149, opened 2022-09-12
Refactor/thread mutable state

JcMinarro:
### 🎯 Goal
Not expose `MutableStateFlow` outside of the class that is using it and perform update logic within the class that contains this `MutableStateFlow`
Fix: #4136 
### 🛠 Implementation details
Remove any public access to the mutable states and refactor the updating logic of those states.

#### General
- [X] PR is linked to the GitHub issue it resolves

#### Code & documentation

### ☑️Reviewer Checklist


leandroBorgesFerreira (changes requested):
Removing the interfaces would recreate the issue [#3710](https://github.com/GetStream/stream-chat-android/issues/3710). Some new interfaces would need to be created later. We can reduce the current interface instead of completely deleting it.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/logic/channel/thread/internal/ThreadStateLogic.kt
```diff
@@ -16,47 +16,73 @@
...
+import io.getstream.chat.android.client.utils.SyncStatus
 import io.getstream.chat.android.offline.plugin.state.channel.thread.internal.ThreadMutableState
 
 /**
  * The logic of the state of a thread. This class contains the logic of how to
  * update the state of the thread in the SDK.
+ *
+ * @property mutableState [ThreadMutableState]
  */
-internal interface ThreadStateLogic {
```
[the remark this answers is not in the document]
JcMinarro: As far as this Interface is Internal and only has one implementation, it is unnecessary.
When we will plan to use it from a different module and we really need the interface from there, we can pull method members up to an interface and put the interface on the proper module it needs to live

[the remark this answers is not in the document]
leandroBorgesFerreira: The internal interface was only to avoid keeping it public before the time... but it won't be hard to create a different interface when it is needed, so no need to stuck this PR over this, let's move on.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/state/channel/thread/internal/ThreadMutableState.kt
```diff
@@ -18,33 +18,54 @@ package io.getstream.chat.android.offline.plugin.state.channel.thread.internal
...
+    fun setEndOfOlderMessages(isEnd: Boolean) {
+        _endOfOlderMessages.value = isEnd
+    }
+
+    fun setOldestInThread(message: Message?) {
+        _oldestInThread.value = message
+    }
+
+    fun deleteMessage(message: Message) {
+        _messages.value -= message.id
```
leandroBorgesFerreira, line 65: I'm not 100% sure that this will actually delete the message, because this returns a new map, without the key. Although it may remove the key also in the original map (I'm not sure) it think this way would be safer: 
```
_messages.value.toMutableMap().remove(message.id)
```

leandroBorgesFerreira: The same goes for `upsertMessages`... using `+=` to make a side effect looks non obvious to me.

leandroBorgesFerreira: I understand that this came form the original code and not introduced in this PR, but I think it would be nice to fix it here, since it is a very small change.

leandroBorgesFerreira: Nevermind, we are reatributing the value, this is necessary.

PR #4202, opened 2022-09-22
Removing event handling result

leandroBorgesFerreira:
### 🎯 Goal

Moving `EventHanlingResult` as it is not necessary to live in the LLC module.


JcMinarro (changes requested):
Link in the docs needs to be update as well.
Previously we had this [broken link that is fixed](https://github.com/GetStream/stream-chat-android/pull/4200)

JcMinarro (approved):

adasiewiczr (approved):

PR #4208, opened 2022-09-27
[ISSUE-4166] Separating ThreadQueryListenerFull

leandroBorgesFerreira:
### 🎯 Goal

Separate ThreadQueryListener into database and state.

### 🛠 Implementation details

In comments

### 🧪 Testing

Test threads in the sample app. 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF
giphy (1)
_No, man, it is Thread not Threat, jesus..._

### DIFF
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerState.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerState.kt
@@ -17,13 +17,10 @@
 package io.getstream.chat.android.offline.plugin.listener.internal
 
 import io.getstream.chat.android.client.errors.ChatError
-import io.getstream.chat.android.client.extensions.internal.users
 import io.getstream.chat.android.client.models.Message
 import io.getstream.chat.android.client.persistance.repository.MessageRepository
-import io.getstream.chat.android.client.persistance.repository.UserRepository
 import io.getstream.chat.android.client.plugin.listeners.ThreadQueryListener
 import io.getstream.chat.android.client.utils.Result
-import io.getstream.chat.android.client.utils.onSuccessSuspend
 import io.getstream.chat.android.offline.plugin.logic.channel.thread.internal.ThreadLogic
 import io.getstream.chat.android.offline.plugin.logic.internal.LogicRegistry
 import io.getstream.logging.StreamLog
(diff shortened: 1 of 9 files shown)


stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerState.kt
```diff
@@ -59,35 +52,27 @@ internal class ThreadQueryListenerFull(
...
-        val messages = messageRepository?.selectMessagesForThread(messageId, limit)
-        val parentMessage = threadLogic?.getMessage(messageId) ?: messages?.firstOrNull { it.id == messageId }
+        threadLogic.setLoading(true)
+        val messages = messageRepository.selectMessagesForThread(messageId, limit)
+        val parentMessage = threadLogic.getMessage(messageId) ?: messages.firstOrNull { it.id == messageId }
 
-        if (parentMessage != null && messages?.isNotEmpty() == true) {
-            threadLogic?.upsertMessages(messages)
-        } else {
-            val result = getRemoteMessage(messageId)
```
leandroBorgesFerreira, line 71: No need to get remove message from the listener. Listeners should not never make API calls as they would open the possibility for infinite loops when one listener calls the other, which makes everything much more complicated. 

Also, it is not necessary to update this message as it will be updated one the original request is completed successfully.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/ThreadQueryListenerState.kt
```diff
@@ -59,35 +52,27 @@ internal class ThreadQueryListenerFull(
...
-            if (result.isSuccess) {
-                val message = result.data()
-                threadLogic?.upsertMessage(result.data())
-                userRepository?.insertUsers(message.users())
-                messageRepository?.insertMessage(message)
-            }
```
kanat, line 77: Hey @leandroBorgesFerreira ,

I'm wondering where this `getRemoteMessage` block is moved to?

leandroBorgesFerreira: I removed this part of the code. The remote message will be fetched from the backend once the result of the call is completed, then we don't need this call to be made in the pre call method. Also, making calls inside listeners is not good, because it may create a problem of a listener triggering a second listener which may create a loop. 

This changes the original behaviour of this listener, but I believe it is better to erase this call.

kanat (approved):
LGTM

Issue #4104, opened 2022-08-29
Revisit our factories


adasiewiczr:
### Context & proposal

Factories should act as real factories and shouldn't store the instance inside.

PR #4220, opened 2022-09-29
Refactor/factories

JcMinarro:
### 🎯 Goal
Redesign our Factory classes.
They should have only one function: Create instances. Those instances should keep their own lifecycle instead of being attached to the factory
Fix: #4104

#### General

#### Code & documentation

### ☑️Reviewer Checklist


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/Plugin.kt
```diff
@@ -68,6 +70,9 @@ public interface Plugin :
     ChannelMarkReadListener,
     CreateChannelListener {
 
+    public val interceptors: List<Interceptor>
```
leandroBorgesFerreira, line 73: Should plugins have interceptors? Plugins should only trigger side effects after backend calls. Holding interceptors and error handlers looks like adding more responsibilities than the original one. It would be better to keep Plugins with one single responsibility.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/Plugin.kt
```diff
@@ -274,4 +279,8 @@ public interface Plugin :
         channelId: String,
         memberIds: List<String>,
     ): Result<Unit> = Result.success(Unit)
+
+    public fun onUserSet(user: User)
```
leandroBorgesFerreira, line 283: We are changing from `Composition` to `Inheritance` to handle user connection/disconnection. Instead of having a class to handle this matter (InitializationCoordinator), we are now passing the responsibility to the plugin itself, which becomes more complex. 

The way I see it, `StatePlugin` become a class that handles everything related to state, so a single point of access. But the problem is that it also has so many responsibilities related to the logic of state in the SDK. 

I believe a class specialized in the lifecycle of a plugin would be better in this situation. Something like PluginHandler/PluginManager/PluginLifecycle takes care of connection and disconnection, so the plugin itself doesn't have to add those methods on its signature.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/internal/StatePlugin.kt
```diff
@@ -33,66 +36,95 @@ import io.getstream.chat.android.client.plugin.listeners.SendReactionListener
...
-    HideChannelListener by hideChannelListener,
-    MarkAllReadListener by markAllReadListener,
-    DeleteReactionListener by deleteReactionListener,
-    SendReactionListener by sendReactionListener,
-    DeleteMessageListener by deleteMessageListener,
-    SendGiphyListener by sendGiphyListener,
-    ShuffleGiphyListener by shuffleGiphyListener,
-    SendMessageListener by sendMessageListener,
-    TypingEventListener by typingEventListener {
+    QueryChannelsListener by QueryChannelsListenerState(logic),
```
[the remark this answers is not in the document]
JcMinarro: The problem is that we need to handle the lifecycle state of `SyncManager`, it is a dependency of some `ListenerStates`, and if we create them outside of the plugin itself, we can't be sure the `SyncManager` we received is the same one is used into our `ListenerState`.
Almost the same applies to `LogicRegistry`, `StateRegistry` and `ClientState`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/Plugin.kt
JcMinarro: Why plugin should only trigger side effects "after" backend calls?
Interceptors and ErrorHandler are kinds of side effects that we need to apply as well

leandroBorgesFerreira: Sorry, I meant after and before backend calls, not only after it. 

Although plugins can have many functionalities, it would be better to narrow them down to just 1 or fewer functionalities, so it becomes more well-defined. 

The other methods are all pure side effect functions, so none of them have return types, or at least the return type is `Result<Unit>`. Adding the interceptors in the signature of `Plugin`, adds now the possibility of changing the information before it goes to the backend. 

If I had to define the responsibility of a Plugin I would say that it is a class that changes the state of the SDK before and after backend calls. But right now we are adding the ability to `intercept and change information,` `handle errors` and `connect and disconnect with the SDK`. 

I like the idea of accessing all the information from `StatePlugin` and `OfflinePlugin` so it becomes easier to understand how the adding/removing `state` and `offline` impacts the SDK, but I believe we need another interface to be the holder of the information. Like: 

```
public interface AddOn {
    public val plugins: List<Plugin>
    public val interceptors: List<Interceptor>
    public val errorHandler: List<ErrorHandler>
    
    public fun onUserSet(user: User)

    public fun onUserDisconnected()
}
```

It wouldn't change much from the current approach, but it would leave the `Plugin` interface untouched.

leandroBorgesFerreira: But there's a more important observation about `interceptors` and `errorHandlers`. Differently than`Plugin`s,  `ErrorHandler`s, and `Interceptor`s **can't guarantee commutative property** (that's why `ErrorHandler` has the `priority` property). So you can never guarantee that A->B->C == C->B->A. In their case, order matters.

Because of that, interceptors and error handlers should all be in the same pool, so they can be properly sorted if a use case needs that. Otherwise we would have separated pools which won't allow sorting: 
```
[PluginA(ErrorHandler1, ErrorHandler2, ErrorHandler3), PluginA(ErrorHandler4, ErrorHandler5, ErrorHandler6)]
```

In the above case, it would be impossible to put `ErrorHandler4` before `ErrorHandler1` because PluginA comes first.. 

By adding `Interceptor` and `ErrorHandler` inside `Plugin`, we would have to worry about the order in plugins too, which makes it a lot more complex.

Sorry about almost writing a book...

JcMinarro: I agree with you, bu it is out of scope of this PR.
Previously we had "Plugin Factories" that:
1-. Create Plugins
2-. Inject dependencies that were needed in our Plugins into our ChatClient
3-. Sync lifecycle or our SDK and plugins.

Now, our "Plugin Factories" only does one thing, "Create Plugins".
The Plugin Instance is the one that handle the other 2 things.
Maybe we can innclude a new `AddOn` interface, but we will need to create some implementation for that, maybe some factory that helps our customer to use it and change the way we inject it into our `ChatClient`.
Not a trivial change that would need investigation/definition of how we want to go on.

PR #4228, opened 2022-10-03
[4226] Preview failed uploads

### LINKED ISSUE #4226: Issues previewing failed media attachments
### Context & proposal

Media attachments that have failed to send will are not previewable in the message list, furthermore trying to preview video attachments will result in an NPE.

### Acceptance criteria

- [ ] Make the attachments previewable even when they have failed to upload.

MarinTolic:
### 🎯 Goal

closes #4226

Failed video and image uploads have blank previews because they are trying to load from a `null` string instead of trying to upload from the local `Attachment.upload` file.

### 🛠 Implementation details

Implemented checks to see if the file has been uploaded or not and load the media accordingly.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_1664812284 | Screenshot_1664812241 |

### 🧪 Testing

1. Launch the compose app
2. Go to any channel
3. Start uploading a video or an image and turn off the internet mid upload

Expected: You should see a failed upload and it should have a working preview

#### General

#### Code & documentation
- [ ] ~~Changelog is updated with client-facing changes~~ (fixing unreleased stuff)

### ☑️Reviewer Checklist

### 🎉 GIF

gif


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/MediaAttachmentContent.kt
```diff
@@ -337,7 +338,10 @@ internal fun MediaAttachmentContentItem(
 
     val data =
         if (isImage || (isVideo && ChatTheme.videoThumbnailsEnabled)) {
-            attachment.imagePreviewUrl
+            when (message.syncStatus) {
+                SyncStatus.COMPLETED -> attachment.imagePreviewUrl
```
MarinTolic, line 342: Using `Attachment.uploadState` here seemsto be slightly more error prone because it sometimes gets cleared if we kill the app, plus this is what `MessageBubble` uses so this way they are in sync when it comes to the general state.

Also using inner `when` here instead of an if-else because I think it has a lower cognitive load than nesting an if-else inside an if. Just trying to lower repetition.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/MediaAttachmentContent.kt
```diff
@@ -372,13 +376,15 @@ internal fun MediaAttachmentContentItem(
                 interactionSource = MutableInteractionSource(),
                 indication = rememberRipple(),
                 onClick = {
```
filbabic, line 378: having a click listener, which only works if the sync status is correct makes it so that we can click, have the click ripple, but not do anything. 

I would either go for: we disable the click entirely if the attachment is in the wrong state (e.g. failed upload or uploading) or we'd open the menu (as if we long clicked) that lets the user retry the upload.

bychkovdmitry (approved):
Works as described. Maybe worth to mention the improvement in the change log.

### ☑️Reviewer Checklist

PR #4499, opened 2022-11-23
[ISSUE-4331] Unread count volume2

leandroBorgesFerreira:
### 🎯 Goal

Fix unread count errors when the SDK is syncing channel data. **You may notice that the total unread count is also wrong, but this problem will be dealt with with a second PR, this one only fixes the count badge of the channels.**

### 🛠 Implementation details

- The count is buffered while the SDK is requesting the backend for data in of the channel. This avoids mixing the local count with the unread count that came with the backend call. This fix the problem of counting messages twice.
- Now there a set to count with messages were counted. This is necessary because the count happens a little bit latter the insertion of the message, so the `visibleMessages` can't be used in this case. 
- When the SDK queries channels, the `createdAt` of the last messages of each channel becomes the `lastMessageSeenDate` of `ChannelUserRead` this is important for checking if the messages shouldn't be used for counting porpuses because it was already included in the query.

**Important:** I little delay may happen in the count during requests for channels, sometimes is can be seen, and sometimes it is irrelevant. This happens because we don't count the unread messages when the SDK is syncing data of the channels with the backend. 

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/204328626-0a108750-d894-4714-894f-e5846506946c.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/204328673-258a0c1e-14f4-42de-9979-2a2d3d27f52c.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
Open the app right after sending a storm of new messages. Make sure that the count is always correct.

Two cases are important: 
- Many messages come when the SDK is syncing, but not after. 
- Many messages come when the SDK is syncing and they continue coming after it too. 

Sometimes you may notice that the numbers of the total unread count and the channel unread count are not aligned. In this case, send a new message to confirm if the problem is in the total unread count or the channel unread count. The new value of the total unread count will be the truth because it comes from the backend.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

giphy (2)


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -366,7 +365,7 @@ internal class ChannelLogic(
             message.ownReactions = it.ownReactions
         }
 
-        channelStateLogic.upsertMessage(message)
+        channelStateLogic.upsertMessage(message, false)
```
adasiewiczr, line 368:
```suggestion
        channelStateLogic.upsertMessage(message, updateCount = false)
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelMessagesUpdateLogic.kt
```diff
@@ -23,9 +23,13 @@ import io.getstream.chat.android.models.Message
 @InternalStreamChatApi
 public interface ChannelMessagesUpdateLogic {
```
adasiewiczr, line 24: Can we make this interface `internal`? 🤔

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/global/GlobalState.kt
```diff
@@ -60,6 +60,8 @@ public interface GlobalState {
      */
     public val banned: StateFlow<Boolean>
 
+    public val queryingChannelsFree: StateFlow<Boolean>
```
adasiewiczr, line 63: Why do we need to expose this `StateFlow`? 🤔

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelMessagesUpdateLogic.kt
leandroBorgesFerreira: We can't, it is used between modules =|.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/global/GlobalState.kt
leandroBorgesFerreira: This is what triggers all the StartStopBuffers in all `UnreadCountLogic`. They all need to have the same trigger

adasiewiczr (approved):

PR #4599, opened 2022-12-14
[4598] Enable Downloading Attachments Without Valid Name or Title Properties

### LINKED ISSUE #4598: SDK cannot download attachment if the name and title properties are missing
### Context & proposal

We rely on valid, non-null attachment name or title when we download an attachment. We do so because we use it to build a valid Java file, which means that when it is not provided we simply cannot create a valid Java file.

We should implement fallback for this behavior.

### Acceptance criteria

- [x] Add fallback values for the file name.

MarinTolic:
### 🎯 Goal

closes #4598 

### 🛠 Implementation details

Added an extension function which uses regex to find the file name from the URL in case it's missing.

Inside the channel `Giphy testing` there is an image of a pair of headphones that was sent from an iPhone device. The image does not have proper `Attachment.name` or `Attachment.title` properties.

Testing steps:

1. Open the image mentioned above in the gallery
2. Download it

Expected: Will silently fail on `v5`/ `develop` branches but will succeed downloading it here.

#### General
- [x] PR targets the ~~`develop`~~ `v5` branch

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

gif

### DIFF
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/ChatClient.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/ChatClient.kt
@@ -39,8 +39,8 @@ import io.getstream.chat.android.client.utils.map
 import io.getstream.chat.android.client.utils.toResultError
 import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
 import io.getstream.chat.android.offline.event.handler.chat.factory.ChatEventHandlerFactory
-import io.getstream.chat.android.offline.extensions.internal.getAttachmentFallbackName
 import io.getstream.chat.android.offline.extensions.internal.logic
+import io.getstream.chat.android.offline.extensions.internal.parseAttachmentNameFromUrl
 import io.getstream.chat.android.offline.extensions.internal.requestsAsState
 import io.getstream.chat.android.offline.plugin.state.StateRegistry
 import io.getstream.chat.android.offline.plugin.state.channel.ChannelState
@@ -55,6 +55,9 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.stateIn
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
 
 /**
  * [StateRegistry] instance that contains all state objects exposed in offline plugin.
@@ -199,7 +202,8 @@ public fun ChatClient.downloadAttachment(context: Context, attachment: Attachmen
             val logger = StreamLog.getLogger("Chat:DownloadAttachment")
             val downloadManager = context.getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager
             val url = attachment.assetUrl ?: attachment.imageUrl
-            val subPath = attachment.name ?: attachment.title ?: attachment.getAttachmentFallbackName() ?: "attachment"
+            val subPath = attachment.name ?: attachment.title ?: attachment.parseAttachmentNameFromUrl()
+            ?: createAttachmentFallbackName()
 
             logger.d { "Downloading attachment. Name: $subPath, Url: $url" }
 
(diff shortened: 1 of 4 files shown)


stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/internal/Attachment.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Uses a regex pattern to extract the file name from the attachment URL.
+ *
+ * Useful in situations where no attachment name or title has been provided and
+ * we need a name in order to download the file to storage.
+ */
+internal fun Attachment.getAttachmentFallbackName(): String? {
+    val url = when (this.type) {
+        "image" -> this.imageUrl ?: this.thumbUrl ?: this.assetUrl
```
MarinTolic, line 29: using when simply because I think it's prettier/ more readable than an if-else in this kind of situation, no practical value in it otherwise.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/internal/Attachment.kt
```diff
@@ -0,0 +1,34 @@
...
+        else -> this.assetUrl ?: this.imageUrl ?: this.thumbUrl
+    }
+
+    return url?.let { Regex("""[^\/]*(?=\?)""").find(it)?.value }
```
MarinTolic, line 33: If there are any suggestions on making this regex pattern better, I'd appreciate it.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/internal/Attachment.kt
```diff
@@ -0,0 +1,34 @@
...
+ * Useful in situations where no attachment name or title has been provided and
+ * we need a name in order to download the file to storage.
+ */
+internal fun Attachment.getAttachmentFallbackName(): String? {
+    val url = when (this.type) {
+        "image" -> this.imageUrl ?: this.thumbUrl ?: this.assetUrl
+        else -> this.assetUrl ?: this.imageUrl ?: this.thumbUrl
+    }
+
+    return url?.let { Regex("""[^/]+(?=\?)""").find(it)?.value }
```
MarinTolic, line 33: If there are any suggestions on making this regex pattern better, I'd appreciate it.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/internal/Attachment.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Uses substrings to extract the file name from the attachment URL.
+ *
+ * Useful in situations where no attachment name or title has been provided and
+ * we need a name in order to download the file to storage.
+ */
+internal fun Attachment.getAttachmentFallbackName(): String? {
+    val url = when (this.type) {
+        "image" -> this.imageUrl ?: this.assetUrl ?: this.thumbUrl
```
MarinTolic, line 29: using when simply because I think it's prettier/ more readable than an if-else in this kind of situation, no practical value in it otherwise.

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/ChatClient.kt
```diff
@@ -198,7 +199,7 @@ public fun ChatClient.downloadAttachment(context: Context, attachment: Attachmen
             val logger = StreamLog.getLogger("Chat:DownloadAttachment")
             val downloadManager = context.getSystemService(Context.DOWNLOAD_SERVICE) as DownloadManager
             val url = attachment.assetUrl ?: attachment.imageUrl
-            val subPath = attachment.name ?: attachment.title
+            val subPath = attachment.name ?: attachment.title ?: attachment.getAttachmentFallbackName() ?: "attachment"
```
JcMinarro, line 202: Can we use a `Date()` or `Timestamp` as the last default one?
It will prevent we override the attachments if they don't have a proper file name.

JcMinarro (approved):

stream-chat-android-state/src/main/java/io/getstream/chat/android/offline/extensions/ChatClient.kt
MarinTolic: Switched to using a dated string

PR #4600, opened 2022-12-16
[4596]  Implement Stream CDN image resizing capabilities

### LINKED ISSUE #4596: Introduce image resizing
### Context & proposal

The new CDN allows us to resize images when uploading them. We should take advantage of this and implement a clean, simple API allowing users to resize images.

### Acceptance criteria

- [ ] Create the described API

MarinTolic:
### 🎯 Goal

Closes #4596 

### 🛠 Implementation details

Introduced new classes into the `core` and `UI-Common` modules which help set resizing options.
These are to be used in conjunction with the function `applyStreamCdnImageResizingIfEnabled` if you are using UI Components, or `createResizedStreamCdnImageUrl` if you are only using core components.

Both `ChatUI` and `ChatTheme` have had the property `streamCdnImageResizing: StreamCdnImageResizing` added to them. Once image resizing has been enabled, these properties will resize all Stream CDN hosted images containing original height and width parameters hosted inside various message previews such as those  inside the message list, the overview inside the gallery and the avatars.

### 🎨 UI Changes

<ins>**The images here have been resized and cropped to half the  original size so that the changes are more noticeable.**</ins>

#### UI Components (XML)

| Before | After |
| --- | --- |
| xml_before_message_list | xml_after_message_list |

| Before | After |
| --- | --- |
| xml_before_gallery | xml_after_gallery |

#### Compose

| Before | After |
| --- | --- |
| compose_before_message_list | compose_after_message_list |

| Before | After |
| --- | --- |
| compose_before_gallery | compose_after_gallery |

### 🧪 Testing

<details>

<summary>Patch providing resizing parameters</summary>

```
Index: stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt	(revision 3e9ba64fd4d4f4ff6fc626c4782483cd7e1aea9b)
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/App.kt	(date 1671708295911)
@@ -19,6 +19,10 @@
 import android.app.Application
 import io.getstream.chat.android.client.utils.internal.toggle.ToggleService
+import io.getstream.chat.android.models.streamcdn.image.StreamCdnCropImageMode
+import io.getstream.chat.android.models.streamcdn.image.StreamCdnResizeImageMode
+import io.getstream.chat.android.ui.ChatUI
+import io.getstream.chat.android.ui.common.images.resizing.StreamCdnImageResizing
 import io.getstream.chat.ui.sample.data.user.SampleUser
 import io.getstream.chat.ui.sample.data.user.UserRepository
 
@@ -35,6 +39,14 @@
         instance = this
         DebugMetricsHelper.init()
         ApplicationConfigurator.configureApp(this)
+
+        ChatUI.streamCdnImageResizing = StreamCdnImageResizing(
+            imageResizingEnabled = true,
+            resizedWidthPercentage = 0.5f,
+            resizedHeightPercentage = 0.5f,
+            resizeMode = StreamCdnResizeImageMode.CROP,
+            cropMode = StreamCdnCropImageMode.LEFT
+        )
     }
 
     private fun getApiKey(): String {
IDEA additional info:
<+>UTF-8
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt	(revision 3e9ba64fd4d4f4ff6fc626c4782483cd7e1aea9b)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt	(date 1671708671621)
@@ -66,6 +66,9 @@
 import io.getstream.chat.android.compose.viewmodel.messages.MessageComposerViewModel
 import io.getstream.chat.android.compose.viewmodel.messages.MessageListViewModel
 import io.getstream.chat.android.compose.viewmodel.messages.MessagesViewModelFactory
+import io.getstream.chat.android.models.streamcdn.image.StreamCdnCropImageMode
+import io.getstream.chat.android.models.streamcdn.image.StreamCdnResizeImageMode
+import io.getstream.chat.android.ui.common.images.resizing.StreamCdnImageResizing
 import io.getstream.chat.android.ui.common.state.messages.MessageMode
 import io.getstream.chat.android.ui.common.state.messages.Reply
 import io.getstream.chat.android.ui.common.state.messages.list.DeletedMessageVisibility
@@ -92,7 +95,15 @@
 
         setContent {
-            ChatTheme(dateFormatter = ChatApp.dateFormatter) {
+            ChatTheme(dateFormatter = ChatApp.dateFormatter,
+                streamCdnImageResizing = StreamCdnImageResizing(
+                    imageResizingEnabled = true,
+                    resizedWidthPercentage = 0.5f,
+                    resizedHeightPercentage = 0.5f,
+                    resizeMode = StreamCdnResizeImageMode.CROP,
+                    cropMode = StreamCdnCropImageMode.LEFT
+                )
+            ) {
                 MessagesScreen(
                     viewModelFactory = factory,
                     onBackPressed = {
```

</details>

**The following test should be applied to both the XML and the Compose sample apps**

1. Open the app and check take a look at images in the message list and gallery overview
2. Apply the patch
3. Repeat step 1 and notice the differences
4. Try different sizes, resize and crop modes

#### General

#### Code & documentation
- [x] Affected documentation updated (KDocs, ~~docusaurus~~, tutorial) **(Docusaurus guide will be added in a separate PR)**

### ☑️Reviewer Checklist

### 🎉 GIF

gif


stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/extensions/StringExtensions.kt
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * A convenience method which evaluates if [value] is null or not and appends
+ * it with the accompanying parameter name in the form of a query parameter.
+ *
+ * @param value Query parameter value.
+ * @param name Query parameter value.
+ */
+private fun String.appendValueAsQueryParameterIfNotNull(value: Any?, name: String): String {
```
MarinTolic, line 95: Core is not an Android model so I couldn't use Android's `Uri`  here. Java's `URI` seemed more limited so I opted to using a custom string extension.

Open to suggestions here.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
```diff
@@ -162,6 +172,7 @@ public fun ChatTheme(
     ).onEach { if (it is DownloadPermissionHandler) it.ObservePermissionChanges() },
     attachmentsPickerTabFactories: List<AttachmentsPickerTabFactory> = AttachmentsPickerTabFactories.defaultFactories(),
     videoThumbnailsEnabled: Boolean = true,
+    streamCdnImageResizing: StreamCdnImageResizing = StreamCdnImageResizing(),
```
MarinTolic, line 175: Having this kind of centralized approach was discussed earlier, but there's a drawback to it.

Take the following screenshot into consideration:

Screenshot_1671549237

The resized image works great withing the quoted comment, however it's quality is too low for the media preview.

I'm wondering if we should add the ability to resize per component as well, such as adding `StreamCdnImageResizing` to individual factories and UI Components as well.

This would incur a greater development cost, but allow for greater customization. 
However, since this approach would incur a greater BE cost due to potentially resizing a single attachment multiple times, it should not be the default approach, but rather an additional capability. Essentially, the default argument would be the one inside the `ChatTheme`.

filbabic: while it doesn't look great, I think it's okay since the user can always open the attachment to preview something of higher quality. It's common to see lower res images as previews and then you can show them as higher res when previewing details.

Worst case, we can try to make the change if customers request this. We also are not 100% sure how many customers will use the resize, how many use their own CDN etc.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/extensions/StringExtensions.kt
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Returns [StreamCdnOriginalImageDimensions] if the image is hosted by Stream's CDN and is resizable,
+ * otherwise returns null.
+ *
+ * @return Class containing the original width and height dimensions of the image or null.
+ */
+public fun String.getStreamCdnHostedImageDimensions(): StreamCdnOriginalImageDimensions? {
+    return try {
+        val width = this.substringAfter("ow=", "").takeIf { it.isNotBlank() }
+            ?.substringBefore("&")?.toInt()
```
filbabic, line 34: Maybe use `toIntOrNull()` to safeguard against bad formatting? Then we might be able to avoid the try/catch block?

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/extensions/StringExtensions.kt
```diff
@@ -0,0 +1,105 @@
...
+ */
+public fun String?.createResizedStreamCdnImageUrl(
+    @FloatRange(from = 0.0, to = 1.0, fromInclusive = false) resizedWidthPercentage: Float,
+    @FloatRange(from = 0.0, to = 1.0, fromInclusive = false) resizedHeightPercentage: Float,
+    resizeMode: StreamCdnResizeImageMode? = null,
+    cropMode: StreamCdnCropImageMode? = null
+): String? {
+    val streamCdnImageDimensions = this?.getStreamCdnHostedImageDimensions()
+
+    return if (this != null && streamCdnImageDimensions != null) {
```
filbabic, line 71: If we have a this != null clause, why not just have this be an extension to `String` and then call

`string?.createResizedStreamCdnImageUrl()`

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/extensions/StringExtensions.kt
filbabic: Would be good to write some tests for this down the line, can be a future improvement

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/resizing/StreamCdnImageResizing.kt
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @param imageResizingEnabled Enables or disables image resizing.
+ * @param resizedWidthPercentage The percentage of the original image width the resized image width will be.
+ * @param resizedHeightPercentage The percentage of the original image height the resized image height will be.
+ * @param resizeMode Sets the image resizing mode. The default mode is [StreamCdnResizeImageMode.CLIP].
+ * @param cropMode Sets the image crop mode. The default mode is [StreamCdnCropImageMode.CENTER].
+ */
+public data class StreamCdnImageResizing(
+    val imageResizingEnabled: Boolean = false,
+    @FloatRange(from = 0.0, to = 1.0, fromInclusive = false) val resizedWidthPercentage: Float = 0.5f,
```
filbabic, line 37: How about an enum case that allows only aspect ratio followed resize?

In this case, users can do 0.01f for width and 1.0 for height and you have an extremely broken image.

We discussed having enums like:

```
enum class ImageSizeFactor(val percentage: Float) {
 QUARTER(0.25f),
 HALF (0.5f),
 THREE_QUARTERS(0.75f),
 ORIGINAL(1f)
}
```

It's less customizable, but also makes it less easy for users to break the API or request something that won't look good. This way, it also always keeps the aspect ratio.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
```diff
@@ -152,4 +153,13 @@ public object ChatUI {
...
     public var videoThumbnailsEnabled: Boolean = true
+
+    /**
+     * Sets the strategy for resizing images hosted on Stream's CDN. Disabled by default,
+     * set [StreamCdnImageResizing.imageResizingEnabled] to true if you wish to enable resizing images. Note that
+     * resizing applies only to images hosted on Stream's CDN which contain the original width (ow) and height (oh)
+     * query parameters.
+     */
+    @JvmStatic
+    public var StreamCdnImageResizing: StreamCdnImageResizing = StreamCdnImageResizing()
```
filbabic, line 164: Also - rather than just creating a value of `StreamCdnImageResizing()` and it having default values, how about having that class w/o default arguments and having an object or factory function `DefaultStreamCdnImageResizing()` that has these values.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/extensions/StringExtensions.kt
MarinTolic: The try catch block is here on purpose in order to not obscure any parsing errors.
If we used `toIntOrNull()` we wouldn't be able to log any errors and would just fail silently without any way of knowing what happened.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
MarinTolic: Alright, implementing a global way of setting the parameter for now.
Giving fine grained control should be a non-breaking implementation should customers ask for it anyways.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
MarinTolic: We can use that approach.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/extensions/StringExtensions.kt
MarinTolic: That's a good suggestion, will refactor.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
```diff
@@ -1559,6 +1598,36 @@ public class MediaGalleryPreviewActivity : AppCompatActivity() {
          */
         private const val KeyVideoThumbnailsEnabled: String = "videoThumbnailsEnabled"
 
+        /**
+         * Represents the key for the [Boolean] value found inside [StreamCdnImageResizing]
+         * turning image resizing on or off.
+         */
+        private const val KeyImageResizingEnabled: String = "imageResizingEnabled"
```
MarinTolic, line 1605: The core module isn't an android module and doesn't use `kotlin-parcelize`. 

Instead of moving the models to an android module where they might fight less well, I opted out for a more primitive but equally functional serialization approach.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -57,3 +64,135 @@ public fun String.cidToTypeAndId(): Pair<String, String> {
...
+}
+
+/**
+ * A convenience method which evaluates if [value] is null or not and appends
+ * it with the accompanying parameter key name in the form of a query parameter.
+ *
+ * @param key Query parameter key.
+ * @param value Query parameter value.
+ */
+private fun Uri.Builder.appendValueAsQueryParameterIfNotNull(key: String, value: Int?): Uri.Builder {
```
MarinTolic, line 192: We could avoid overloading by using `?.toString()` instead of `.toString()`.

However the linter does not understand that this actually has a use and we would have to suppress the warning. Not an issue in itself, but I'm worried someone who doesn't know the reason behind this might see it in the future, think that it was a mistake and remove it. 

If that happens, when the value is null the `Uri.Buidlder` will append a query param as `querParam=null` instead of omitting it.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -57,3 +65,150 @@ public fun String.cidToTypeAndId(): Pair<String, String> {
...
+/**
+ * Generates a string URL with Stream CDN image resizing query parameters added to it. Once this URL is called, Stream's
+ * CDN will generate a resized image which is accessible using the link returned by this function.
+ *
+ * @param resizedWidthPercentage The percentage of the original image width the resized image width will be.
+ * @param resizedHeightPercentage The percentage of the original image height the resized image height will be.
+ * @param resizeMode Sets the image resizing mode. The default mode is [StreamCdnResizeImageMode.CLIP].
+ * @param cropMode Sets the image crop mode. The default mode is [StreamCdnCropImageMode.CENTER].
+ */
+public fun String.createResizedStreamCdnImageUrl(
```
MarinTolic, line 112: Would be great to be able to test this properly, but unforunatelly `Uri` is an Android class and can't easily be covered by unit tests.

I don't think mocking it is the answer since it's responsible for core behavior which we wouldn't want to get wrong in a mock.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -57,3 +65,150 @@ public fun String.cidToTypeAndId(): Pair<String, String> {
...
+    }
+}
+
+/**
+ * Checks if a string contains resizing related query parameters.
+ *
+ * @return true if the URL contains resizing parameters, false otherwise.
+ */
+private fun Uri.wasImagePreviouslyResized(): Boolean =
+    queryParameterNames.any { queryParam ->
```
filbabic, line 179: Maybe we can do something like:

`queryParameterNames.intersect(listOf(ourParameters)).isNotEmpty()`

filbabic: It scales better if we add more params as we won't have to check every single one

stream-chat-android-compose/detekt-baseline.xml
```diff
@@ -29,6 +28,7 @@
     <ID>MagicNumber:MediaAttachmentFactory.kt$0.25f</ID>
     <ID>MagicNumber:MediaAttachmentQuotedContent.kt$0.8f</ID>
     <ID>MagicNumber:MediaGalleryPreviewActivity.kt$MediaGalleryPreviewActivity$0.2f</ID>
+    <ID>MagicNumber:MediaGalleryPreviewActivity.kt$MediaGalleryPreviewActivity$0.5f</ID>
```
filbabic, line 31: ?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
```diff
@@ -1528,6 +1537,36 @@ public class MediaGalleryPreviewActivity : AppCompatActivity() {
...
+     * Fetches individual image resizing options from the bundle and
+     * packs them into [StreamCdnImageResizing].
+     *
+     * @return An instance of [StreamCdnImageResizing] created from individual options fetched from
+     * the bundle packed inside the given intent.
+     */
+    private fun Intent.createStreamCdnImageResizing(): StreamCdnImageResizing {
+        val imageResizingEnabled = getBooleanExtra(KeyImageResizingEnabled, false)
+
+        val resizedWidthPercentage = getFloatExtra(KeyStreamCdnResizeImagedWidthPercentage, 0.5f)
```
filbabic, line 1550: why are the defaults to halve the size?

filbabic (approved):
Looks good and nicely documented. Left some comments/questions.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -57,3 +65,150 @@ public fun String.cidToTypeAndId(): Pair<String, String> {
...
     return checkNotNull(split(":").takeIf { it.size >= 2 }?.let { it.first() to it.last() })
 }
+
+/**
+ * Returns [StreamCdnOriginalImageDimensions] if the image is hosted by Stream's CDN and is resizable,
+ * otherwise returns null.
+ *
+ * @return Class containing the original width and height dimensions of the image or null.
+ */
+public fun String.getStreamCdnHostedImageDimensions(): StreamCdnOriginalImageDimensions? {
```
JcMinarro, line 75: Should it only works with Stream's CDN?
Are `ow` and `oh` "Standard" param names?
Should we use them if they come from a different host? (eg: `https://www.jc-cdn.com/picture.jpg?ow=15&oh=30`)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -16,7 +16,15 @@
 
 package io.getstream.chat.android.client.extensions
 
+import android.net.Uri
```
JcMinarro, line 19: New methods on this flie are simple "input/output method", could we cover them with UnitTest?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
JcMinarro: It can be unit-tested with robolectric: https://stackoverflow.com/a/46317695

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/StreamCdnResizeImageQueryParameterKeys.kt
```diff
@@ -0,0 +1,46 @@
...
+
+package io.getstream.chat.android.models.streamcdn.image
+
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+
+/**
+ * Contains query parameter keys used to resize Stream CDN hosted images.
+ */
+@InternalStreamChatApi
+public object StreamCdnResizeImageQueryParameterKeys {
```
JcMinarro, line 25: Do we need this class outside of `stream-chat-android-client`?? :thinking:

JcMinarro (changes requested):

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/StreamCdnResizeImageQueryParameterKeys.kt
MarinTolic: We don't really, I just wanted to keep structure consistency.

If you want it transferred to the client module and made internal I can do that. :smile:

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
MarinTolic: Added tests in the following commit: https://github.com/GetStream/stream-chat-android/pull/4600/commits/873f11fe673e0694b7551cfc59c23fffbcb13432

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
MarinTolic: This was discussed internally, but to keep everyone on the same page I'll write points here summarizing why it's believed this logic should be domain agnostic:

- Possibility of introducing custom domains in the future
- If a 3rd party CDN does serve `ow` and `oh` parameters in their links and we append resize parameters, these should most likely simply be ignored by the 3rd party CDN

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
MarinTolic: Suggestion was implemented in the following commit: https://github.com/GetStream/stream-chat-android/commit/873f11fe673e0694b7551cfc59c23fffbcb13432

stream-chat-android-compose/detekt-baseline.xml
MarinTolic: Detekt considers passing the default value parameter as a magic number.

We're using 0.5f as a fallback for fetching an extra float. This case should not occur, I'm not sure it's worth declaring a constant over it, however if you think it is I'll introduce one.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
MarinTolic: I have to declare _a_ number here.

The actual value missing is a case that should happen if anyone uses the code correctly using our "new instance pattern" (`MediaGalleryPreviewActivity.getIntent()`).

It could happen if someone decides to not use it, but I think in that case it could be considered a user error.

So in reality this value of `0.5f` is just something we had to declare, but it should not hold any meaning if the activity was created using the proper method.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
MarinTolic: Implemented the suggestion in the following commit: https://github.com/GetStream/stream-chat-android/pull/4600/commits/5c833a3550d73556dfb47913afe7a479230b4859

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/streamcdn/image/extensions/StringExtensions.kt
MarinTolic: Implemented as suggested.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
MarinTolic: Suggestion implemented.

stream-chat-android-compose/detekt-baseline.xml
filbabic: 🤦 get rekt

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
filbabic: Yeah but any reason why it's not 1f to just not resize and use full size?

JcMinarro (approved):
LGTM.
There are some suggestions from Filip that would be nice to achieve

filbabic (approved):

PR #4613, opened 2023-01-10
Minor optimization/fix

filbabic:
### 🎯 Goal

Just added checks to not trigger loading if there are no messages to load.

### 🧪 Testing

Test A)
1. Run Compose app on v5.
2. Open a new thread. (preferably with an image message as parent)
3. Attempt to scroll top/bottom. 

You should see a lot of requests in Logcat as well as the loading spinner freaking out.

Test B)
1. Run Compose on this branch.
2. Open a new thread. (preferably with an image message as parent)
3. Attempt to scroll top/bottom.

The scroll shouldn't trigger any unnecessary loads.

If you test it in a thread that has _many_ messages, it should work fine and load all the messages.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageList.kt
@@ -258,7 +258,7 @@ public fun MessageList(
 
     when {
         isLoading -> loadingContent()
-        !isLoading && messages.isNotEmpty() -> Messages(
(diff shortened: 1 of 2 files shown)


MarinTolic (commented):
Couldn't confirm any behavioral changes.

Tested on both emulator and physical phone but the same behavior seems to remain.

MarinTolic (approved):
The original issue could be reproduced only in V6 and not in V5. However, this change seems to have no adverse effects while making minor code improvements such as removing redundant checks so approval is given.

PR #4713, opened 2023-03-07
[Do not merge]  `main` to `develop`

MarinTolic:
No really, don't merge.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Channel.kt
```diff
@@ -208,7 +207,7 @@ public fun Channel.incrementUnreadCount(currentUserId: String, lastMessageSeenDa
 
 @InternalStreamChatApi
 public fun Collection<Channel>.applyPagination(pagination: AnyChannelPaginationRequest): List<Channel> {
-    val logger by taggedLogger("ChannelSort")
+    val logger = StreamLog.getLogger("Chat:ChannelSort")
```
MarinTolic, line 210: Actually I think we should keep the old one here.

JcMinarro: We should use 
```suggestion
    val logger by taggedLogger("Chat:ChannelSort")
```

PR #4784, opened 2023-04-14
Read States are only returned on the case memberLimit!=0

JcMinarro:
### 🎯 Goal
Backend doesn't return Read States on the case `memberList==0`, so it could be confusing to customers that use the default constructor of `QueryChannelsRequest`.
To avoid that, the default value on the constructor is being changed to `1`


workspace:
There was a problem with the unread count and member role flickering in my application. After changing the member count to 1 based on this pull request, the issue was resolved.

PR #4790, opened 2023-04-20
Upgrade dependencies

JcMinarro:
### 🎯 Goal
Upgrade Dependencies


buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
```diff
@@ -6,34 +6,34 @@ object Versions {
...
-    internal const val ANDROIDX_ACTIVITY_COMPOSE = "1.6.1"
-    internal const val ANDROIDX_ACTIVITY_KTX = "1.6.1"
+    internal const val ANDROIDX_ACTIVITY_COMPOSE = "1.7.1"
+    internal const val ANDROIDX_ACTIVITY_KTX = "1.7.1"
     internal const val ANDROIDX_ANNOTATIONS = "1.5.0"
     internal const val ANDROIDX_APPCOMPAT = "1.5.1"
-    internal const val ANDROIDX_COMPOSE = "1.3.1"
-    public const val ANDROIDX_COMPOSE_COMPILER = "1.4.3"
+    internal const val ANDROIDX_COMPOSE = "1.4.1"
+    public const val ANDROIDX_COMPOSE_COMPILER = "1.4.4"
```
JcMinarro, line 14: It can't be upgraded to `1.4.5` yet because with this compiler version, gradle needs to use Java 17 and it is incompatible with our basecode.
They are proposing to downgrade it to Java 11 on version `1.4.6`

JcMinarro: Screenshot from 2023-04-20 14-33-23
https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.4.5
 
Because of this compiler, we can't upgrade Kotlin version yet :face_exhaling:

stream-chat-android-benchmark/build.gradle.kts
```diff
@@ -11,12 +11,12 @@ android {
...
     compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
+        sourceCompatibility = JavaVersion.VERSION_11
+        targetCompatibility = JavaVersion.VERSION_11
     }
 
     kotlinOptions {
-        jvmTarget = "1.8"
+        jvmTarget = JavaVersion.VERSION_11.toString()
```
JcMinarro, line 19: Compose Compiler needs to work with Java 11

buildSrc/src/main/kotlin/io/getstream/chat/android/Dependencies.kt
JcMinarro: New version 1.4.6 has been already released!!!! :partying_face: :tada: :tada:

PR #4806, opened 2023-05-08
Thread messages start feature

JcMinarro:
### 🎯 Goal
Be able to set position where the thread messages will start from.
Complete GetStream/android-internal-board#62  

### 🎨 UI Changes

_Add relevant screenshots_

| Bottom | Top |
| --- | --- |
| Screenshot_20230509_112730 |Screenshot_20230509_113943|


kanat (approved):
LGTM.

Are we gonna support this in XML as well?

JcMinarro:
> LGTM.
> 
> Are we gonna support this in XML as well?

It was already implemented on XML: https://getstream.io/chat/docs/sdk/android/ui/message-components/message-list/#messages-start-position

ChrisTitos:
After updating the SDK to the latest version, suddenly the messages in channels (not threads) are always aligned at the top.
I tracked it down to this pull request, because I think this changed the old behavior (which was align bottom)

You see on [this line](https://github.com/GetStream/stream-chat-android/pull/4806/files#diff-fe28e6c1dd72adf314217310803dfff8a8a0118f51bbad4890e12c75bcd91325R184) if there is no parent message, alignment is Top, but it used to be Bottom before this change. 
There is also no option provided to change this behavior.

Should I make a new bug ticket for this?

PR #4808, opened 2023-05-10
Update connection state

JcMinarro:
### 🎯 Goal
Add the user that started the WebSocket connection into the `ConnectionState` class.
Closes: https://github.com/GetStream/android-internal-board/issues/63

### 🛠 Implementation details
Previously we had an enum representing the connection state. Now, instead of having an enum, it has been changed to a Sealed Class that allows us to have different implementations. I am keeping the previous 3 states, but the `Connected` state contains the user that opened the WebSocket Connection.


kanat (approved):
LGTM

PR #4827, opened 2023-05-22
Improve connect user logic

JcMinarro:
### 🎯 Goal

Improve how our connectUser logic works to allow offline connections without waiting the connection result

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -306,7 +306,7 @@ internal constructor(
         logger.i { "Initialised: ${buildSdkTrackingHeaders()}" }
     }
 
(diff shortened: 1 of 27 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -994,18 +996,19 @@ internal constructor(
     //endregion
 
     public fun disconnectSocket() {
-        chatSocket.disconnect()
+        launch { chatSocket.disconnect() }
     }
 
     public fun reconnectSocket() {
```
kanat, line 1002: Not related to this ticket.
What do you think about aligning `disconnectSocket` and `reconnectSocket` signature with other functions in `ChatClient`?
```kotlin
public fun disconnectSocket(): Call<Unit> {
    //...
}

public fun reconnectSocket(): Call<Unit> {
    //...
}
```

kanat (approved):
LGTM

PR #4829, opened 2023-05-24
Keep consistent API in our ChatClient class

JcMinarro:
### 🎯 Goal

Keep consistent API in our ChatClient class 
Close: https://github.com/GetStream/android-internal-board/issues/71


kanat (approved):
LGTM

PR #4836, opened 2023-05-31
Handle ban/unban events v6

JcMinarro:
### 🎯 Goal
Handle properly ban/unban events on ChannelState 

### 🛠 Implementation details
The `ChannelUserBannedEvent` contains a new property to let us know whether it is a shadow or a standard ban.
It has been appropriately handled on DB and State instances


kanat (approved):
LGTM

Issue #4747, opened 2023-03-23
Support showing "user x is typing" is channel list


leandroBorgesFerreira:
### Context & proposal

Currently, the android SDK only supports showing typing information inside the channel. We need to add to the channel list too. 

### Acceptance criteria

- [x] Show who started typing.
- [x] Erase the status after a predefined time
- [x] Make erase time customizable.

PR #4868, opened 2023-07-10
Feature/channel typing indicator v6

JcMinarro:
### 🎯 Goal
Add typing indicator on the channel list view.
Fix: https://github.com/GetStream/stream-chat-android/issues/4747

### 🛠 Implementation details
Added typing users to `ChannelItem` entity and adding a new `bind()` method  that take as argument the `ChannelItem` instance instead of the `Channel`


kanat (approved):
LGTM

PR #4880, opened 2023-07-17
Create a `destroy()` method for our States

JcMinarro:
### 🎯 Goal
Our "States" shouldn't be updated anymore after the current user logout. To ensure it a new `destroy()` method has been created.


kanat (approved):
LGTM

PR #4887, opened 2023-07-24
Feature/update pn

JcMinarro:
### 🎯 Goal
Use the last [stream-android-push](https://github.com/getstream/stream-android-push) version that make `providerName` a mandatory field.
Added the migration guide and docs for it.

Fix: https://github.com/GetStream/android-internal-board/issues/98


kanat (approved):
LGTM

PR #4888, opened 2023-07-24
Fix Partial User Update

JcMinarro:
### 🎯 Goal
Fix Partial User Update by Keeping immutable user after merging it


skydoves (approved):
Looks good to me!

PR #4896, opened 2023-07-26
Dispose lifecycleHandler after an unrecoverable Error

JcMinarro:
### 🎯 Goal
Dispose lifecycleHandler after an unrecoverable Error


kanat (approved):
LGTM

PR #4897, opened 2023-07-26
Keep v6 as "unreleased" version

JcMinarro:
### 🎯 Goal
Keep v6 as "unreleased" version
The current version will be our current v5 version and v6 will be under the `/next` path


kanat (approved):
LGTM

PR #4905, opened 2023-07-27
Fix/typing indicaator on channel list v5

JcMinarro:
### 🎯 Goal
Typing Indicator wasn't being appropriately hidden after a message was received because the TypingStopped Event wasn't fired
Fix: https://github.com/GetStream/android-internal-board/issues/102


kanat (approved):
LGTM

PR #4917, opened 2023-08-09
Align untitled channel name with iOS

JcMinarro:
### 🎯 Goal
A channel without a name should compute the name depending on the members that it contains.
By default, only the first two members' names are shown, and if there is any extra member on the group a +X more is added to the channel name.

Fix: https://github.com/GetStream/android-internal-board/issues/106


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ChannelNameFormatter.kt
kanat, line 1: Should we mention this change in migration guide? 🙂

PR #4947, opened 2023-09-07
Fix build issues with JDK 17 in offline and client module

DanielNovak:
The sample project `stream-chat-android-ui-components-sample` will not compile on a default Android Studio installation with Java 17 because it uses `stream-chat-android-offline` and similarly with `stream-chat-android-client` and these one doesn't set the Kotlin jvmTarget to Java 11 to match with other modules that use Java 11. This will cause a build error:

> 'compileDebugJavaWithJavac' task (current target is 11) and 'kspDebugKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.


kanat (approved):
LGTM

PR #4971, opened 2023-09-26
Fix Thread Loading

JcMinarro:
### 🎯 Goal
Fix Thread Loading.
The state wasn't properly updated because the code after `collect{ }` wasn't executed.

Fix https://github.com/GetStream/android-internal-board/issues/134


kanat (approved):
LGTM

PR #4990, opened 2023-10-04
Mark core models with Stable and Immutable to improve Compose performance

skydoves:
Mark core models with Stable and Immutable.

Compose compiler metrics: https://github.com/androidx/androidx/blob/androidx-main/compose/compiler/design/compiler-metrics.md

Even if the core models are immutable, it's actually not marked as stable between different modules. So we should mark them with stable markers to improve Compose performance and prevent useless recomposition.

Diff: https://github.com/GetStream/stream-chat-android/commit/4cf9bcc09e51a12c212b0c5138b7e15a871ab19b


DanielNovak (approved):

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/GuestUser.kt
```diff
@@ -16,4 +16,7 @@
 
 package io.getstream.chat.android.models
 
+import androidx.compose.runtime.Stable
+
+@Stable
```
kanat, line 21: I'm wondering if this can be `@Immutable`?

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/Member.kt
```diff
@@ -16,12 +16,14 @@
...
 package io.getstream.chat.android.models
 
+import androidx.compose.runtime.Stable
 import io.getstream.chat.android.models.querysort.ComparableFieldProvider
 import java.util.Date
 
 /**
  * Represents a channel member.
  */
+@Stable
```
kanat, line 26: Same here and other places

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/GuestUser.kt
skydoves: If this model doesn't need to be observed or tracked as a state by Compose runtime, we can mark this with an `Immutable` annotation.

kanat (approved):
LGTM

PR #4992, opened 2023-10-06
Update benchmark module and generate baseline profiles

skydoves:
Update benchmark module and generate baseline profile to improve SDK startup performance.

In Video SDK, the startup performance was increased by 35~40%, I believe Chat SDK will get more improvements in the startup time.


kanat (approved):
LGTM

PR #4994, opened 2023-10-07
Improve channel repository cache

JcMinarro:
### 🎯 Goal

Improve channelRepository and include coroutineScope into our DB Wrappers Implementation to perform the `insert` actions asynchronously.


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
```diff
@@ -16,52 +16,64 @@
...
+        cacheChannel(updatedChannels)
+        scope.launch {
+            logger.v {
+                "[insertChannels] inserting ${channelToInsert.size} entities on DB, " +
+                    "updated ${updatedChannels.size} on cache"
+            }
+            channelToInsert
+                .takeUnless { it.isEmpty() }
+                ?.let { channelDao.insertMany(it) }
+        }
```
kanat, line 72: Hey @JcMinarro,

Wdyt about making this scope as a single-thread scope and put all write operations behind that scope?

It seems like we may have race-condition when we insert 1000 channels and delete 5 channels.
Let's save we had
```kotlin
val cidToDelete = "some_cid"
insertChannels(channelList)
// channelList contains cidToDelete
deleteChannel(cidToDelete)
```
In this case `deleteChannel ` will faster than `insertChannels`, so `insertChannels` will re-insert the deleted channel

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
kanat, line 1: Shouldn't we update cache when calling one of these function:
```
override suspend fun setChannelDeletedAt(cid: String, deletedAt: Date)
override suspend fun setHiddenForChannel(cid: String, hidden: Boolean, hideMessagesBefore: Date)
override suspend fun setHiddenForChannel(cid: String, hidden: Boolean)
```
 🤔

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
JcMinarro: Let me check it

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
JcMinarro: Yep, good catch!!!!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
JcMinarro: Looks like single-thread scope won't handle it as we expect: https://stackoverflow.com/questions/61126844/coroutines-on-a-single-thread-and-guaranteed-sequentiality

kanat: Maybe we could use `mutex` then to keep the order?

DanielNovak: Mutex I think is a safe bet, we can just something like this - since we only need to synchronise this one code block and we can synchronise it on the instance itself.

```
        synchronized(DatabaseChannelRepository@this) {
            ....
        }
````

File: stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/user/internal/DatabaseUserRepository.kt [hunk 2]
```diff
@@ -32,11 +38,14 @@ internal class DatabaseUserRepository(
     // the user cache is simple, just keeps the last 100 users in memory
     private val userCache = LruCache<String, User>(cacheSize)
     private val latestUsersFlow: MutableStateFlow<Map<String, User>> = MutableStateFlow(emptyMap())
+    private val dbMutext = Mutex()
```
DanielNovak, line 41: Small typo, should be `dbMutex`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/user/internal/DatabaseUserRepository.kt
```diff
@@ -32,11 +38,14 @@ internal class DatabaseUserRepository(
...
 
     override fun observeLatestUsers(): StateFlow<Map<String, User>> = latestUsersFlow
 
     override suspend fun clear() {
-        userDao.deleteAll()
+        scope.awaitWithMutex(dbMutext) {
```
DanielNovak, line 46: This isn't a real suspending function anymore now 🤔 . It now launches a separate coroutine and thus you can't rely on the clear() to have finished if you call it. Maybe in this case we should use:

    override suspend fun clear() {
        dbMutex.withLock {
            userDao.deleteAll()
        }
    }

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
```diff
@@ -174,8 +182,10 @@ internal class DatabaseMessageRepository(
     override suspend fun clear() {
         messageCache.evictAll()
         replyMessageCache.evictAll()
-        messageDao.deleteAll()
-        replyMessageDao.deleteAll()
+        scope.awaitWithMutex(dbMutex) {
```
DanielNovak, line 185: Same here, I think it's safer and more clearer to just use this here...

    override suspend fun clear() {
        dbMutex.withLock {
            messageDao.deleteAll()
            replyMessageDao.deleteAll()
        }
    }

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
```diff
@@ -16,52 +16,67 @@
...
         if (channels.isEmpty()) return
-        val entities = channels.map { channel -> channel.convertToEntity() }
-        logger.v { "[upsertChannels] entities.size: ${entities.size}" }
-        channelDao.insertMany(entities)
+        val updatedChannels = channels
+            .map { channelCache[it.cid]?.let { cachedChannel -> it.combine(cachedChannel) } ?: it }
+        val channelToInsert = updatedChannels
+            .filter { channelCache[it.cid] != it }
+            .map { it.toEntity() }
+        cacheChannel(updatedChannels)
```
DanielNovak, line 66: I think there could be a case where the number of channels is bigger than 1000 - in that case not all of them will be cached. And this function (like some others in the PR) will return immediately but the cache will not be a complete representation of the state. Can this cause some consistency issues? 🤔  If so - can we detect a case where it doesn't fit the cache and we should fire the DB calls synchronously?...

JcMinarro: In that case, a "double insertion" of the same channels (Channel number 1001 and the following ones).
Not a real case for our customers, it shouldn't be so big and as far as we only use it as "offline support", it will only have some data of the channel that is not up-to-date for a moment but updated whenever a new event arrives on that channel

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/user/internal/DatabaseUserRepository.kt
JcMinarro: `scope.awwaitWithMutex(dbMutex) { }` is not launching a new coroutine, but blocking the current one

DanielNovak: Aha, nice :-).

JcMinarro: TBH, this extension function is not needed anymore, initially, it was using the coroutine scope, but after some changes, I was doing, it is not used and we don't really need it.
Let me update it.

DanielNovak (approved):

Issue #4995, opened 2023-10-09
App crashes on Android 14 when receiving chat push message


damianpetla:
**Describe the bug**
Receiving push notification on Android 14 and letting getstream handle it crashes the app with the following exception:
```kotlin
android.app.InvalidForegroundServiceTypeException: Starting FGS with type none callerApp=ProcessRecord{9c477c0 959:xxxxxxxxxxxxxxx/u0a650} targetSDK=34 has been prohibited
 at android.app.InvalidForegroundServiceTypeException$1.createFromParcel(InvalidForegroundServiceTypeException.java:53)
 at android.app.InvalidForegroundServiceTypeException$1.createFromParcel(InvalidForegroundServiceTypeException.java:49)
 at android.os.Parcel.readParcelableInternal(Parcel.java:4870)
 at android.os.Parcel.readParcelable(Parcel.java:4852)
 at android.os.Parcel.createExceptionOrNull(Parcel.java:3052)
 at android.os.Parcel.createException(Parcel.java:3041)
 at android.os.Parcel.readException(Parcel.java:3024)
 at android.os.Parcel.readException(Parcel.java:2966)
 at android.app.IActivityManager$Stub$Proxy.setServiceForeground(IActivityManager.java:6761)
 at android.app.Service.startForeground(Service.java:862)
 at androidx.work.impl.foreground.SystemForegroundService$1.run(SystemForegroundService.java:124)
 at android.os.Handler.handleCallback(Handler.java:958)
 at android.os.Handler.dispatchMessage(Handler.java:99)
 at android.os.Looper.loopOnce(Looper.java:205)
 at android.os.Looper.loop(Looper.java:294)
 at android.app.ActivityThread.main(ActivityThread.java:8177)
 at java.lang.reflect.Method.invoke(Native Method)
 at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
 at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)
```

It crashes on `LoadNotificationDataWorker` when you set it to be foreground but without type. Check out Google docs where they explain that https://developer.android.com/guide/background/persistent/how-to/long-running#foreground-service-type

**SDK version**
- 5.17.13

**To Reproduce**
1. Enable push notifications and wait for a push message from another device.

**Expected behavior**
Expect getstream handle the push message and display it. Not crash.

**Device:**
 - Vendor and model: Google Pixel 5
 - Android version: 14

[...]

PR #4997, opened 2023-10-10
Declare a Data Sync foreground service type for `LoadNotificationDataWorker`

JcMinarro:
### 🎯 Goal
Starting on Android 14, whenever a Background Service is started, it needs to declare which type it is.
In our case, we use it to sync data in our database and the type to be used needs to be `FOREGROUND_SERVICE_TYPE_DATA_SYNC`

Fix: #4995

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## stream-chat-android-client
 ### 🐞 Fixed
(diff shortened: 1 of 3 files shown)


DanielNovak (approved):

kanat (approved):
LGTM

PR #5010, opened 2023-10-20
Disable minify on our UI-Component sample app

JcMinarro:
### 🎯 Goal
Disable minify in our UI-Component Sample App until we provide rules to not fail on compilation process.
Related PR: https://github.com/GetStream/stream-chat-android/pull/5007


skydoves (approved):
Looks good to me!

PR #5015, opened 2023-10-23
Bump Nexus Plugin to 1.3.0

skydoves:
Bump Nexus Plugin to 1.3.0.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,6 +3,7 @@ name: Publish
 on:
   workflow_run:
     workflows: ["ReleaseStart"]
+    workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -3,6 +3,7 @@ name: Publish
 on:
   workflow_run:
     workflows: ["ReleaseStart"]
+    workflow_dispatch:
```
skydoves, line 6: I added `workflow_dispatch` to run the publication scripts manually, just in case the publication script failed we can try re-running easily instead of creating a new release on GitHub.

JcMinarro (approved):

PR #5036, opened 2023-11-06
Upgrade StreamPuhs Dependency

JcMinarro:
### 🎯 Goal
Upgrade Stream Push artifact to the [last version](https://github.com/GetStream/stream-android-push/releases/tag/v1.1.6) that fixes the race condition due to initializing the SDK after the first activity was created


kanat (approved):
LGTM

PR #5041, opened 2023-11-07
Use shortService instead of dataSync for our workmanager job

JcMinarro:
### 🎯 Goal
Starting on Android 14, whenever a Background Service is started, it needs to declare which type it is.
On #4997 a DataSync service was declared to be able to sync our DB after a PN is received. This permission requires updating Google Play Console App Content data with a video showing functionality using this permission. All our customers would need to update the Google Play Console App Content to complaint with Google Play Policy. 
To avoid it, we can use [ShortService](https://developer.android.com/about/versions/14/changes/fgs-types-required#short-service) foreground service type instead that doesn't require any new permission.


kanat (approved):
LGTM

PR #5062, opened 2023-11-13
Update baseline profiles for each module and README guides

skydoves:
### 🎯 Goal

Update baseline profiles for each module and README guides.

This PR also resolves the plugin issue: https://issuetracker.google.com/issues/309614153


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/login/UserLoginActivity.kt
```diff
@@ -78,7 +76,7 @@ class UserLoginActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
 
         setContent {
-            ChatTheme {
+            ChatTheme(allowUIAutomationTest = true) {
```
kanat, line 79: Just wondering if this is also required by baseline profiles changes :)

skydoves: We should add `.semantics { testTagsAsResourceId = true }` at the root Compose hierarchy to generate proper baseline profiles via the UI Automator. We can add on the demo app one by one for each Activity or new Compose hierarchy, but it will also be messy sample codes and Compose UI components. So, I ended up adding the parameter to the highest level (ChatTheme) component.

kanat (approved):
LGTM

PR #5068, opened 2023-11-14
Fix Gzip compression for requests

DanielNovak:
GZIP configuration was wrong. We were adding `application/gzip` instead of `gzip` into `Accept-Encoding`. But the best solution is to not specify this at all - because if you do, then you have to handle it on your own. OkHttp handles this transparently - we don’t need to specify it.
This is best explained by Jake Wharton here: https://github.com/square/okhttp/issues/2132 

I verified this by placing breakpoints into `GzipSource.read`  in OkHttp and also in `BridgeInterceptor` . The read function is not getting called on the `GzipSource` in develop - only if I delete our accept-encoding header it works correctly.


kanat (approved):
LGTM

PR #5070, opened 2023-11-19
Faster ISO date formatter / parser

DanielNovak:
Our API is currently quite heavy on the usage ISO date which are notoriously slow to parse. There is a plan to replace this with Unix timestamps but in the short term we can use a small ISO parser and formatter - https://github.com/ethlo/itu/ . According to the docs and also to my tests the speed is significantly faster - up 10x, in our API response parsing scenario it seems to be "only" 4x faster than the Java Instant but that is still a massive saving. 

I have also added the usage of Instant into our util DateTimeFormatter. This one is used from our Compose components and while the absolute time saving is minimal (5ms vs ~3ms) on each item drawn - it still can make a difference because this is done on the UI thread. 

In the long run we will make the JSON structure more efficient and replace the ISO date with Unix timestamps, but until that this is worthy upgrade.


kanat (approved):
LGTM

Nice catch 👍

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/DateFormatter.kt
```diff
@@ -92,8 +131,18 @@ internal class DefaultDateFormatter(
...
+                    dateFormatterDayOfWeekNew.format(date.toInstant())
+                } else {
+                    dateFormatterDayOfWeek.format(date)
+                }
+            }
+            else -> if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                dateFormatterFullDateNew.format(date.toInstant())
+            } else {
+                dateFormatterFullDate.format(date)
+            }
```
JcMinarro, line 145: What about creating a couple of `DefaultDateFormater` depending on the SDK Version and only using the appropriate `DateFormater` removing unnecessary if/else blocks into the instance implementation?
At the end, the `DefaultDateFormater` entity is instantiated only within this file, the SDK Version check can be moved there.

PR #5071, opened 2023-11-19
Faster ISO date parser / formatter

DanielNovak:
Our API is currently quite heavy on the usage ISO date which are notoriously slow to parse. There is a plan to replace this with Unix timestamps but in the short term we can use a small ISO parser and formatter - https://github.com/ethlo/itu/ . According to the docs and also to my tests the speed is significantly faster - up 10x, in our API response parsing scenario it seems to be "only" 4x faster than the Java Instant but that is still a massive saving.

In the long run we will make the JSON structure more efficient and replace the ISO date with Unix timestamps, but until that this is worthy upgrade.


kanat (approved):
LGTM

Nice catch 👍

DanielNovak:
leaving this to v6 only

PR #5075, opened 2023-11-19
Preload KClass members for our DTO classes

DanielNovak:
Reflection in Kotlin is very slow - the KClass.members can take even hundreds of miliseconds when it's invoked the first time (subsequent calls are fast). We use this reflection when parsing the extra fields in some of the responses. And this is causing an uncessary delay when the first channels or messages are loaded. We can just eagerly warm up the reflection for each relevant class and prevent this first invocation delay. This seems to save about ~500ms when loading the first channels page.


DanielNovak:
leaving this to v6 only

PR #5077, opened 2023-11-19
Asynchronously create ChannelMutableState()

DanielNovak:
The `ChannelMutableState()` is rather slow to create (many `StateFlows` with `SharingStarted.Eagerly`) - we need to investigate this closer but for now we can prepare the channels in parallel. In my measurements this approach is always faster - it seems to save about 500ms on 30 channels (in the sample app).


DanielNovak:
leaving this to v6 only

PR #5078, opened 2023-11-20
Delay reflection in NotificationHandlerFactory

DanielNovak:
Delaying the reflection saves ~190ms on the main thread while the SDK is being initialised (usually in the Application.onCreate() - thus delaying the whole application cold start).

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactory.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactory.kt
@@ -133,11 +133,14 @@ public object NotificationHandlerFactory {
     private fun provideDefaultUserIconBuilder(context: Context): UserIconBuilder {
         // We search for the StreamCoilUserIconBuilder by reflection and this is slow - we need
         // to postpone to not block the SDK initialisation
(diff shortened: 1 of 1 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactory.kt
```diff
@@ -130,22 +131,54 @@ public object NotificationHandlerFactory {
...
     private fun provideDefaultNotificationPermissionHandler(context: Context): NotificationPermissionHandler {
-        val appContext = context.applicationContext
-        return runCatching {
-            Class.forName(
-                "io.getstream.android.push.permissions.snackbar.SnackbarNotificationPermissionHandler",
-            ).kotlin.primaryConstructor?.call(appContext) as NotificationPermissionHandler
-        }.getOrDefault(
-            DefaultNotificationPermissionHandler.createDefaultNotificationPermissionHandler(appContext as Application),
-        )
+        // We search for the StreamCoilUserIconBuilder by reflection and this is slow - we need
```
JcMinarro, line 153:
```suggestion
        // We search for the SnackbarNotificationPermissionHandler by reflection and this is slow - we need
```

PR #5079, opened 2023-11-20
Fix skip slow mode capability not handled

DanielNovak:
Same implementation as https://github.com/GetStream/stream-chat-swiftui/pull/402/files.
We need to handle the `skip-slow-mode` flag which completely disables slow mode on the client.


JcMinarro (commented):
Looks good.
Could we update the CHANGELOG.md?

aleksandar-apostolov (approved):

PR #5084, opened 2023-11-22
Fix user mutes and unread count incorrect

DanielNovak:
Our unread count, mute list and banned list are incorrect because of a subtle bug. We catch the `ConnectedEvent` in the `ChatClient` and we set the user. And after that we handle the `ConnectedEvent` inside `updateGlobalState` but here we now ignore the user completely because there is no change to the user that was just set. So we never emit the initial unread count, mutes... 
Upon further inspection I believe that we don't need the `MutableValue` at all - this is an unncessary optimisation (which caused the issue). The `StateFlow`s will not emit if the value hasn't changed, so there is no performance hit if we execute it several times.


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -295,62 +277,53 @@ internal class EventHandlerSequential(
             // connection events are never send on the recovery endpoint, so handle them 1 by 1
             when (event) {
                 is ConnectedEvent -> if (batchEvent.isFromSocketConnection && event.me.id == currentUserId) {
-                    me.set(event.me)
-                    totalUnreadCount.set(event.me.totalUnreadCount)
-                    channelUnreadCount.set(event.me.unreadChannels)
+                    updateUserGlobalState(event.me)
```
JcMinarro, line 280: Here we are not updating `totalUnreadCount` neither `channelUnreadCount`

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -295,62 +277,53 @@ internal class EventHandlerSequential(
...
                 }
                 is NotificationMutesUpdatedEvent -> if (event.me.id == currentUserId) {
-                    me.set(event.me)
-                    totalUnreadCount.set(event.me.totalUnreadCount)
-                    channelUnreadCount.set(event.me.unreadChannels)
+                    updateUserGlobalState(event.me)
```
JcMinarro, line 283: Same here

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -295,62 +277,53 @@ internal class EventHandlerSequential(
...
                 }
                 is NotificationChannelMutesUpdatedEvent -> if (event.me.id == currentUserId) {
-                    me.set(event.me)
-                    totalUnreadCount.set(event.me.totalUnreadCount)
-                    channelUnreadCount.set(event.me.unreadChannels)
+                    updateUserGlobalState(event.me)
```
JcMinarro, line 286: Same here

JcMinarro (changes requested):

JcMinarro (approved):

Issue #5034, opened 2023-11-02
 Issue with Reply Message Functionality in Push Notifications


Testtaccount:
**Describe the bug**
The reply message functionality, when offline plugin is connected, is not working as expected. When application is in a killed state and a push notification is displayed, users can reply to a message within the notification. However, after replying to a message, typing directly in notifications input area and sending a message, the chat with the current user stops being displayed, conversations are lost. Instead, it shows "No Messages" until the app's cache is cleared. This behavior is unexpected and disrupts the user experience.

**SDK version**
- 6.0.5

**To Reproduce**
1. Ensure the application has offline chat support and it is in a killed state.
2. Receive a push notification.
3. Click on the notification's  "reply" action and write message in input area and send it
4. After replying, observe that the chat with the current user , chat is no longer displayed
5. The conversation shows "No Messages."

**Expected behavior**
We expect that when a user replies to a message in a push notification, the chat with the current message should continue to be displayed in the user's conversations. The "No Messages" issue should not occur, and users should not need to clear the app's cache to resolve this.

**Device:**
 - Vendor and model: Pixel 7
 - Android version: 14

**Device:**
 - Vendor and model: Redmi note 6 pro
 - Android version: 9

DanielNovak:
Thank you for reporting the issue. We have been able to reproduce the problem with your instructions and we are looking into it now.

DanielNovak:
Fixed in https://github.com/GetStream/stream-chat-android/releases/tag/v6.0.9

PR #5086, opened 2023-11-23
Fix empty message list

JcMinarro:
### 🎯 Goal
Messages sent ofline weren't initializing the date when the message was created and if we didn't update the message from the one that is returned from backend, we could arrive to condition where it is not updated never causing a crash when the date of all messages is used to sort a conversation.

Apart of that, if we jump to a message in the past, it could cause a gap on the conversation due the pagination was not implemented properly.

I have created a method to be sure any gap on the conversation list is filled when the State module is used.

Fix: #5034

### 🧪 Testing
To test it, you will need to jump to a message in the pass within a conversation.
I have create a channel where there are an indexed list of messages to verify the list of messages is loaded properly.
Open the UI-Component sample app, login with "Jc Miñarro" user and click on the conversation with "Daren".
You will need to apply the following patch to ensure you are jumping to the message `50`

<details>

<summary>Patch code to automatically jump to message `50` when go into conversation with "Daren"</summary>

```
Subject: [PATCH] Patch code to automatically jump to message `50` when go into conversation with "Daren"
---
Index: stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt	(revision 83508a78b3d60740438f4f0589b4c3455ec90bf6)
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/channel/list/ChannelListFragment.kt	(date 1700752105892)
@@ -104,7 +104,7 @@
 
             setChannelItemClickListener {
                 requireActivity().findNavController(R.id.hostFragmentContainer)
-                    .navigateSafely(HomeFragmentDirections.actionOpenChat(it.cid))
+                    .navigateSafely(HomeFragmentDirections.actionOpenChat(it.cid, messageId = "jc-a804c3cb-3e9b-4af1-a56a-54419d372b80"))
             }
 
             setChannelDeleteClickListener { channel ->

```

</details>

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -385,9 +385,6 @@ public class MessageListController(
     }
 
     private fun observeChannelState(): StateFlow<ChannelState?> {
-        // Using 0 as limit will skip insertion of new messages if we are loading a specific message when
-        // opening the MessageList screen. Prevents race condition where wrong messages might get loaded.
-        val messageLimit = if (messageId != null) 0 else messageLimit
         logger.d { "[observeChannelState] cid: $cid, messageId: $messageId, messageLimit: $messageLimit" }
         return chatClient.watchChannelAsState(
             cid = cid,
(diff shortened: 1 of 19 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -384,9 +385,6 @@ public class MessageListController(
     }
 
     private fun observeChannelState(): StateFlow<ChannelState?> {
-        // Using 0 as limit will skip insertion of new messages if we are loading a specific message when
-        // opening the MessageList screen. Prevents race condition where wrong messages might get loaded.
-        val messageLimit = if (messageId != null) 0 else messageLimit
```
JcMinarro, line 389: This workarround is not needed anynmore as far as we are ensuring any gap in our state is filled in

DanielNovak (approved):

PR #5105, opened 2023-12-06
[i222] allow connect user with user_id only

kanat:
### 🎯 Goal

Closes:
- https://github.com/GetStream/android-internal-board/issues/222

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/User.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/User.kt
@@ -47,9 +47,9 @@ public data class User(
     var role: String = "",
     var name: String = "",
     var image: String = "",
-    var invisible: Boolean = false,
+    var invisible: Boolean? = null,
     var language: String = "",
-    var banned: Boolean = false,
+    var banned: Boolean? = null,
     var devices: List<Device> = mutableListOf(),
     var online: Boolean = false,
     var createdAt: Date? = null,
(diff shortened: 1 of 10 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/User.kt
```diff
@@ -47,9 +47,9 @@ public data class User(
     var role: String = "",
     var name: String = "",
     var image: String = "",
-    var invisible: Boolean = false,
```
JcMinarro, line 50: I would keep the default value here to avoid breaking changes on the current behavior of our customers. If anyone wants to not use this argument, they can set It up as null

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/User.kt
```diff
@@ -47,9 +47,9 @@ public data class User(
...
+    var invisible: Boolean? = null,
     var language: String = "",
-    var banned: Boolean = false,
+    var banned: Boolean? = null,
```
JcMinarro, line 52: Same here

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/models/User.kt
kanat: I assume the overall idea is that SDK is not supposed to do any magic like sending default values, unless those values are explicitly set by customers :)

nuno-vieira: Yeah I think we are safe here @JcMinarro, puting the default value as nullable, should not break anything 👍 And it is more aligned to the desired behaviour

PR #5109, opened 2023-12-07
Fix Regex not escaped in getOccurrenceRanges

DanielNovak:
This function can lead to a crash if the items contain invalid Regex strings (e.g. "Ryan :)")


kanat (approved):
LGTM

PR #5111, opened 2023-12-08
Revert "Replace stateIn with DerivedStateFlow in ChannelMutableState

DanielNovak:
The original https://github.com/GetStream/stream-chat-android/pull/5083 PR is causing mysterious issues - the chat doesn't scroll to the latest messages. It's related to the `messagesTransformation()` function - the old one works, but new one doesn't. This needs more investigation but until then we can revert the change.


kanat:
Let's see if [this](https://github.com/GetStream/stream-chat-android/pull/5112) fix helps first 🙂

PR #5103, opened 2023-12-05
Add menu option to mark a conversation as unread from a message

JcMinarro:
### 🎯 Goal
This PR implement all the low level code to mark a conversation unread from a message and handle any new events is received regarding this feature.
An UI option to invoke this feature is added to the UI-Components but it is disabled by default until we allow to show an "unread label" within the conversation list.

### 🎨 UI Changes
Screenshot_20231205_102756

### 🧪 Testing
To test it you will need to apply the following patch

<details>

<summary>Patch to enable "Mark as Unread" by default</summary>

```
Subject: [PATCH] Patch code to automatically jump to message `50` when go into conversation with "Daren"
---
Index: stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListViewStyle.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListViewStyle.kt b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListViewStyle.kt
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListViewStyle.kt	(revision 373e7e221faa102b398e9be2de10bb19ff437d99)
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListViewStyle.kt	(date 1702476467699)
@@ -335,7 +335,7 @@
                     attributes.getBoolean(R.styleable.MessageListView_streamUiCopyMessageActionEnabled, true)
 
                 val markAsUnreadEnabled =
-                    attributes.getBoolean(R.styleable.MessageListView_streamUiMarkAsUnreadEnabled, false)
+                    attributes.getBoolean(R.styleable.MessageListView_streamUiMarkAsUnreadEnabled, true)
 
                 val retryMessageEnabled =
                     attributes.getBoolean(R.styleable.MessageListView_streamUiRetryMessageEnabled, true)
```

</details>

PR #5117, opened 2023-12-13
Refactor channel reads

JcMinarro:
### 🎯 Goal

`Channel.read` property keeps a list of  `ChannelUserRead` for every member of the channel. This entity contains the number of unread messages for the current user, and should be the one used to update/obtain the unread messages count of the channel for the current user.

On this PR I updated how it is calculated and created a new extension property on our `Channel` class that helps us to check the number of unread messages on the channel.

This PR depend on https://github.com/GetStream/stream-chat-android/pull/5103


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/ChannelEntity.kt
```diff
@@ -41,6 +41,7 @@ import java.util.Date
  * @param memberCount Number of members in the channel.
  * @param watcherIds The list of watchers` ids.
  * @param watcherCount Number of watchers in the channel.
+ * @param unreadCount Number of unread messages for the current user.
```
kanat, line 44: I can't find `unreadCount` property inside `ChannelEntity` class 🙂

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/database/internal/ChatDatabase.kt
```diff
@@ -67,7 +67,7 @@ import io.getstream.chat.android.offline.repository.domain.user.internal.UserEnt
         CommandInnerEntity::class,
         SyncStateEntity::class,
     ],
-    version = 69,
+    version = 70,
```
kanat, line 70: If I got it right we've increased the version here due `Date` nullability changes inside `ChannelUserReadEntity`, is that right? 🙂

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
```diff
@@ -18,8 +18,42 @@ package io.getstream.chat.android.state.event.handler.internal.utils
...
 internal val ChatEvent.realType get() = when (this) {
     is ConnectedEvent -> "connection.connected"
     else -> type
 }
+
+internal fun MessageReadEvent.toChannelUserRead() = ChannelUserRead(
+    user = user,
+    lastReceivedEventDate = createdAt,
+    lastRead = createdAt,
+// TODO: remove this once the backend is fixed and is sending us the number of unread messages
```
kanat, line 36: If we don't need to wait for this to be done on BE, then PR looks good to me 🙂

kanat (approved):

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
JcMinarro: We don't need it.
We are using an optimistic approach until backend sends us the needed data.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/database/internal/ChatDatabase.kt
JcMinarro: Yep, It is needed because DB Structure is different

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/ChannelEntity.kt
JcMinarro: You are right, at the end this calculated value is not needed on DB and we don't store it.
Let me remove this line

PR #5120, opened 2023-12-18
[i227] fix crash in audio player after relogin

kanat:
### 🎯 Goal

Closes: 
- https://github.com/GetStream/android-internal-board/issues/227

### 🛠 Implementation details

In addition to the crash fix, other minor bugs were fixed as well:
- reply view was showing incorrect duration value (always showing zero)
- seekTo was using incorrect dimension (seconds instead of milliseconds) 

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/audio/NativeMediaPlayer.kt
```diff
@@ -239,45 +293,98 @@ internal class NativeMediaPlayerImpl(
...
+    override fun stop() {
+        if (DEBUG) logger.d { "[stop] no args" }
+        mediaPlayer.stop()
+        state = NativeMediaPlayerState.STOPPED
+    }
 
-    override fun reset() = mediaPlayer.reset()
+    override fun reset() {
+        if (DEBUG) logger.d { "[reset] no args" }
+        mediaPlayer.reset()
```
aleksandar-apostolov, line 345: What do you think about wrapping this into defensive `try-catch` block and setting `state = NativeMediaPlayerState.ERROR` in case it fails.

Maybe also wrap `if (DEBUG) { // Log inside... }` (old habit) :)

Same for the other functions.

kanat: Actually, I'd love to have it crashing, cause those crashes are due to the nature of `MediaPlayer` having a state machine in a wrong state. So for customers it would look like nothing crashes, but nothing works properly and for us this would be harder to debug 🙂 

`if (DEBUG) { // Log inside... }` I intentionally made it a one-liner to have it compact, cause `spotless` will transform it to
```kotlin
if (DEBUG) { 
    // Log inside... 
}
```

DanielNovak (approved):

aleksandar-apostolov (approved):

PR #5153, opened 2024-01-18
Open `ChatClient.getMessageUsingCache()` extension method

JcMinarro:
### 🎯 Goal
Open `ChatClient.getMessageUsingCache()` extension method
fix: https://github.com/GetStream/android-internal-board/issues/238


kanat (approved):
LGTM

Issue #5162, opened 2024-01-30
Sending link with capitalized Https or http, breaks the link.


zeroarst:
**Describe the bug**
Send a Https://www.google.com message to another user.

**SDK version**
- 6.0.8

**To Reproduce**
1. Send a Https://www.google.com message to another user.
2. Click on the link
3. A browser open with broken "http//www.google.com" link.

**Expected behavior**
Support capitalized Https, Http

**Device:**
 - Vendor and model: Samsung S23 Ultra
 - Android version: 14

**Screenshots**
The problem is found in TextUtils.kt
![image](https://github.com/GetStream/stream-chat-android/assets/1834837/33522c1b-b8bb-48b1-ab35-2cc9aedbf4d7)

PR #5163, opened 2024-01-30
Ensure the link builder always uses lowercase

JcMinarro:
### 🎯 Goal
Link should be always used in lowercase.
Fix: #5162


kanat (approved):
LGTM

PR #5175, opened 2024-02-08
Fix/unread counter

JcMinarro:
### 🎯 Goal
Fix unreadCount logic when new messages are received.
Unread count needs to be processed on client-side because backend doesn't provide this data in all events we receive.
Unread Counter is increased whenever a new message from other user arrives.


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
 ## stream-chat-android-state
 ### 🐞 Fixed
 - Clear stale cache inside `StateRegistry`. [#5164](https://github.com/GetStream/stream-chat-android/pull/5164)
+- Unread messates coun is now updated properly. [#5175](https://github.com/GetStream/stream-chat-android/pull/5175)
```
kanat, line 44:
```suggestion
- Unread messages count is now updated properly. [#5175](https://github.com/GetStream/stream-chat-android/pull/5175)
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Channel.kt
```diff
@@ -48,28 +48,37 @@ public val Channel.lastMessage: Message?
...
-    val newLastMessage = lastMessageAt == null || messageCreatedAt.after(lastMessageAt)
     val newReads = read.map { read ->
         read.takeUnless { it.user.id == currentUserId }
             ?: read.copy(
-                lastReceivedEventDate = messageCreatedAt,
-                unreadMessages = read.unreadMessages + 1,
+                lastReceivedEventDate = receivedEventDate,
+                unreadMessages = read.let {
+                    val hasNewUnreadMessage = receivedEventDate.after(it.lastReceivedEventDate) &&
+                        newMessages.size > messages.size &&
```
kanat, line 71: I assume this condition is always true, since `newMessages` equals to `messages` + `message`?

kanat (approved):
LGTM

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Channel.kt
JcMinarro: It should, but during the development, I had some cases where it wasn't the case and the counter counts twice

PR #5185, opened 2024-02-26
Ensure new token is fetch on case reconnecting WS

JcMinarro:
### 🎯 Goal
Our Socket Factory wasn't fetching a new token in case the reconnection state was achieved. 
Whenever a TokenProvider is used to provide a JWT token with an expiration date, when the token expires and the WS is closed, the token used to establish a new WS Connection isn't refreshed, causing an endless retry process unless the user interacts with our Rest API.


kanat (approved):
LGTM

Issue #5186, opened 2024-02-27
ArrayIndexOutOfBoundsException of ApiRequestsDumper


donlan:
**Describe the bug**
I catch a crash with the below console prints:
```
java.lang.ArrayIndexOutOfBoundsException: length=4; index=4
 at java.util.ArrayList.add(ArrayList.java:468)
 at io.getstream.chat.android.client.plugins.requests.ApiRequestsDumper.registerRequest(ApiRequestsDumper.kt:47)
 at io.getstream.chat.android.client.api.interceptor.ApiRequestAnalyserInterceptor.intercept(ApiRequestAnalyserInterceptor.kt:42)
 at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
 at io.getstream.chat.android.client.api.interceptor.HeadersInterceptor.intercept(HeadersInterceptor.kt:33)
 at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
 at io.getstream.chat.android.client.api.interceptor.ApiKeyInterceptor.intercept(ApiKeyInterceptor.kt:31)
 at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
 at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
 at okhttp3.OkHttpOkHttp3Extension.okhttp3RealCall_getResponseWithInterceptorChain$okhttp(OkHttpSourceFile:57)
 at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
 at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
 at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
 at java.lang.Thread.run(Thread.java:1012)
```

**SDK version**
- 6.0.14

**To Reproduce**
I have no idea,occasionally.

**Expected behavior**

**Device:**
 - Vendor and model: Xiaomi M2102J2SC
 - Android version: 13

**Screenshots**
None.

kanat:
Fixed in [v6.1.0](https://github.com/GetStream/stream-chat-android/releases/tag/v6.1.0)

PR #5187, opened 2024-02-27
Use immutable list to ensure thread-safe access

JcMinarro:
### 🎯 Goal
Use immutable list to ensure thread-safe access
Fix: #5186

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugins/requests/ApiRequestsDumper.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugins/requests/ApiRequestsDumper.kt
@@ -31,23 +31,17 @@ internal class ApiRequestsDumper(
     private val dateFormat: DateFormat = SimpleDateFormat("HH:mm:ss.SSS", Locale.getDefault()),
 ) : ApiRequestsAnalyser {
 
-    private val requestsDataMap: MutableMap<String, MutableList<RequestData>> = mutableMapOf()
+    private val requestsDataMap: MutableMap<String, List<RequestData>> = mutableMapOf()
 
     /**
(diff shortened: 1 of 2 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugins/requests/ApiRequestsDumper.kt
```diff
@@ -31,23 +31,17 @@ internal class ApiRequestsDumper(
...
     override fun registerRequest(requestName: String, data: Map<String, String>) {
         val requestData = RequestData(requestName, Date(), data)
-        val requestDataList = requestsDataMap[requestName]
-
-        if (requestDataList != null) {
-            requestDataList.add(requestData)
-        } else {
-            requestsDataMap[requestName] = mutableListOf(requestData)
-        }
+        requestsDataMap[requestName] = (requestsDataMap[requestName] ?: emptyList()) + requestData
```
kanat, line 44: Just off the top of my head, I do remember we had this `ArrayOutOfBoundException` with this approach cause compiler was still iterating behind the scene :)

May we could `synchronize` this peace of code?

kanat: Ah, I mixed it up with ConcurrentModification :)

kanat: Maybe we should wrap it by synchronized block, since `LinkedHashMap` is not thread safe, there might a possibility when 2 threads read and write simultaneously, meaning we may end up with wrong data :)

PR #5188, opened 2024-02-28
Filter messages that has been already processed to increase unread counter

JcMinarro:
### 🎯 Goal
Filter messages that has been already processed to increase unread counter


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelStateLogic.kt
```diff
@@ -646,6 +648,7 @@ internal class ChannelStateLogic(
                     ),
                 )
             }
+        processedMessageIds.add(message.id)
```
kanat, line 651: Should we clear this cache somehow? :)

JcMinarro: Shouldn't be very big cache, but we can use an LRUCache there. Let me check

kanat (approved):
LGTM

PR #5189, opened 2024-02-28
Clean cached channel

JcMinarro:
### 🎯 Goal
Ensure  channel and its messages are removed on case to receive a RemoveChannel Events


kanat (approved):
LGTM

PR #5204, opened 2024-03-13
Fix `String.containsLinks()` extension function

JcMinarro:
### 🎯 Goal
Fix `String.containsLinks()` extension function


kanat (approved):
LGTM

kanat:
Can we add some link to the ticker or place where this was discussed? :)

PR #5210, opened 2024-03-19
Restore docs pages URLs

liviu-timar:
### 🎯 Goal

Restore URLs for files that have broken cached Google search results links.

### 🛠 Implementation details

- Moved back files that changed position.
- Refactored sidebar and switched to manual list of pages for most of the sections.
- Removed numbering for most of the docs directories, as they are not accurate anymore (and might be confusing). Kept numbering for directories that still have autogenerated sections in the sidebar.
- Fixed broken links.


kanat (approved):
LGTM

PR #5229, opened 2024-04-04
Not increase unread counter when an shadowed message arrives

JcMinarro:
### 🎯 Goal
Not increase unread counter when an shadowed message arrives
Fix: https://github.com/GetStream/android-internal-board/issues/277


kanat (approved):
LGTM

PR #5236, opened 2024-04-15
Jump to first unread message

JcMinarro:
### 🎯 Goal
When a user goes into a channel with unread messages, the message list is automatically scrolled to the first unread message position.

Close: https://github.com/GetStream/android-internal-board/issues/199

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
@@ -1942,6 +1942,15 @@ public class MessageListView : ConstraintLayout {
     }
 
     /**
+     * Sets the handler used when the unread label is reached.
+     *
+     * @param listener The listener to use.
+     */
+    public fun setOnUnreadLabelReachedListener(listener: OnUnreadLabelReachedListener) {
+        listenerContainer.unreadLabelReachedListener = listener
+    }
+
+    /**
      * Used to display the moderated message dialog when you long click on a message that has failed the moderation
      * check.
      *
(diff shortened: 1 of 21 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
```diff
@@ -625,6 +629,19 @@ public class MessageListView : ConstraintLayout {
...
+            unreadLabelButton = binding.unreadLabelButton
+            unreadLabelButton?.apply {
+                setTextStyle(unreadLabelButtonStyle.unreadLabelButtonTextStyle)
+                backgroundTintList = ColorStateList.valueOf(unreadLabelButtonStyle.unreadLabelButtonColor)
+                rippleColor = ColorStateList.valueOf(unreadLabelButtonStyle.unreadLabelButtonRippleColor)
+            }
+        } else {
+            binding.unreadLabelButton.isVisible = false
+        }
+    }
```
kanat, line 643:
```suggestion
    private fun initUnreadLabelButton(unreadLabelButtonStyle: UnreadLabelButtonStyle) {
        binding.unreadLabelButton.isVisible = unreadLabelButtonStyle.unreadLabelButtonEnabled
        if (unreadLabelButtonStyle.unreadLabelButtonEnabled) {
            unreadLabelButton = binding.unreadLabelButton
            unreadLabelButton?.apply {
                setTextStyle(unreadLabelButtonStyle.unreadLabelButtonTextStyle)
                backgroundTintList = ColorStateList.valueOf(unreadLabelButtonStyle.unreadLabelButtonColor)
                rippleColor = ColorStateList.valueOf(unreadLabelButtonStyle.unreadLabelButtonRippleColor)
            }
        }
    }
```

kanat (approved):
LGTM

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
JcMinarro: Button should only be visible when the first unread message is not shwon.
This property is only to allow to use the button or disable it.

PR #5247, opened 2024-04-29
Reaction Group

JcMinarro:
### 🎯 Goal
Add Reaction Group field to the Message Entity.
The UI Changes will be achieve in a different PR

PR #5248, opened 2024-04-30
Add ReactionSorting

JcMinarro:
### 🎯 Goal
Creating a new class to provide a sorting behavior for Reactions.
It can be configured on XML Artifact by XML attrs and/or Styles Transformation, and in Comopse Artifact by injecting the ReactionSorting to be used.

Depend on: #5247


kanat (approved):
LGTM

PR #5256, opened 2024-05-08
Feature get newer replies

JcMinarro:
### 🎯 Goal
Create a Low-Level Logic to fetch reply messages from older to newer.
The "State" and UI implementation will be covered on a different PR.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1549,6 +1550,28 @@ internal constructor(
             .share(userScope) { GetRepliesIdentifier(messageId, limit) }
     }
 
+    /**
+     * Fetch replies to the specified message with id [parentId] that are newer than the message with [lastId].
+     * If [lastId] is null, the oldest replies are returned.
+     *
+     * @param parentId The id of the parent message.
+     * @param limit The number of replies to fetch.
+     * @param lastId The id of the last message to fetch.
```
kanat, line 1559:
```suggestion
     * @param lastId The id of the last message to fetch from exclusively.
```

kanat (approved):
LGTM

Issue #5252, opened 2024-05-03
The API are sending correct response but Android SDK is failing to give in succes callback.


Yatish1910:
**Describe the bug**
We are trying to get list of members based upon some filters, in app inspection API is giving 200 Status Code and response is alse there, but from android side failure callbnack is only getting called.
**SDK version**
- 6.3.2

**To Reproduce**
Please see the below code
```
private fun getUsers() {
        Log.d("totalId", args.idList.toList().toString())
        val userGetStreamId = client.getCurrentUser()?.id
        if(userGetStreamId != null){
            val filters = Filters.and(
                Filters.`in`("id", args.idList.toMutableList()),
                Filters.ne("id", userGetStreamId)
            )
            val request = QueryUsersRequest(
                filter = filters,
                offset = 0,
                limit = 0,
                QuerySortByField.ascByName("name")
            )
            client.queryUsers(request).enqueue {
                when(it){
                    is Result.Failure -> {
                        Log.d("createUserListStatus", it.value.message)
                    }
                    is Result.Success -> {
                        val users: List<User> = it.value
                        Log.d("userSuccess", users.toString())
                        createChannelsAdapter.submitList(users)
                    }
                }

            }
        }
```

**Expected behavior**
It should trigger Success callback and list should have data.

**Device:**
 - Moto G52
 - Android version: [13]

GrayaMax:
I have the same issue, if it can help, there is an issue with the parser for User item.
It's an issue with data received.

We get this error : 
`NetworkError(message=Response is failed. See cause, serverErrorCode=1000, statusCode=-1, cause=com.squareup.moshi.JsonDataException: com.squareup.moshi.JsonDataException: Expected BOOLEAN but was null at path $.invisible at $.users[1])`

The model used to parse the response seems to have `invisible` value not nullable and we receive `null` value.

Yatish1910:
Hi Stream , 
Any Update on this bug/issue?

lachlan-eagling:
Hey Stream Team,

My team is also experiencing this issue, we hadn't made any change to the implementation in our app, or change SDK version implying something in API response has changed.

cc @allanswer & @zeroarst

JcMinarro:
Hello there.
Thanks for reporting it.
We are fixing it and will be ready soon.

lachlan-eagling:
Thanks @JcMinarro, appreciate the quick turnaround on this one.

[...]

PR #5257, opened 2024-05-09
Fix crash when parsing users from query user endpoint

JcMinarro:
### 🎯 Goal

Fix crash when parsing users from query user endpoint.
Close: https://github.com/GetStream/stream-chat-android/issues/5252


kanat (approved):
LGTM

PR #5258, opened 2024-05-13
Feature/load thread from older to newer

JcMinarro:
### 🎯 Goal
Implement UI logic to load thread's messages from older to newer.
The low level implementation was implemented on https://github.com/GetStream/stream-chat-android/pull/5256

### 🎉 GIF


kanat (approved):
LGTM

PR #5265, opened 2024-05-17
Headers providers Interfaces to fetch image/video URLs

JcMinarro:
### 🎯 Goal
The SDK already provided the `ImageHeadersProvider` interface to inject custom headers whenever an image URL wanted to be loaded. This interface has been updated to receive as a parameter the URL that will be fetched to be able to inject different headers depending on the URL.

In the same way, the `VideoHeadersProvider` interface has been created to use custom headers when video URLs are fetched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,8 +51,10 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Created `VideoHeadersProvider` interface to provide headers for video requests. [#5265](https://github.com/GetStream/stream-chat-android/pull/5265)
 
 ### ⚠️ Changed
+- The `ImageHeadersProvider.getImageRequestHeaders()` method use the url as a parameter. [#5265](https://github.com/GetStream/stream-chat-android/pull/5265)
 
 ### ❌ Removed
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -51,8 +51,10 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Created `VideoHeadersProvider` interface to provide headers for video requests. [#5265](https://github.com/GetStream/stream-chat-android/pull/5265)
 
 ### ⚠️ Changed
+- The `ImageHeadersProvider.getImageRequestHeaders()` method use the url as a parameter. [#5265](https://github.com/GetStream/stream-chat-android/pull/5265)
```
kanat, line 57: Is this a breaking change?

kanat (approved):
LGTM

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ImageHeadersProvider.kt
```diff
@@ -20,9 +20,16 @@ package io.getstream.chat.android.ui.common.helper
...
+     *
+     * @param url The URL of the image to load.
+     * @return A map of headers to be used for the image loading request.
+     */
+    public fun getImageRequestHeaders(url: String): Map<String, String>
 }
 
 internal object DefaultImageHeadersProvider : ImageHeadersProvider {
-    override fun getImageRequestHeaders(): Map<String, String> = emptyMap()
+    override fun getImageRequestHeaders(url: String): Map<String, String> = emptyMap<String, String>()
```
kanat, line 34: Wondering if it was a request :)

I assume this was added to determine which headers to include based on `url`, right?

JcMinarro: It wasn't requested, but if we had it, the customer would be able to decide which headers to add to every case.
Apart from that, it was a security bug, because this provider is used to load ALL pictures on the message list, the headers could be obtained easily by a final user.

PR #5266, opened 2024-05-21
[PBE-3738] Ensure message has id before starting the send process

JcMinarro:
### 🎯 Goal
The LLC SDK has a preventive mechanism to not send the same message twice.
The mechanism is based on the message id, but the message id wasn't initialized in time which caused some messages to be dropped in the case multiple messages are sent at the same time.
 
Fix https://stream-io.atlassian.net/browse/PBE-3738


kanat (approved):
LGTM

PR #5272, opened 2024-05-31
Add proguard / R8 integration for the default rule configuration

skydoves:
Add proguard / R8 integration for the default rule configuration.

references:

https://github.com/google/gson/pull/2397


JcMinarro (changes requested):
Looks good, but we need to use different file names per artifact to be sure there aren't any conflicts when our customer apps are compiled

JcMinarro (approved):

PR #5273, opened 2024-06-03
[PBE-3762] Polls API

JcMinarro:
### 🎯 Goal
Implement the Low-Level API to create and interact with Polls.
The `Poll` is a new class within `Message`  and contains all info about the poll.
There are new Events related with polls, providing poll info and the updated message as well.


stream-chat-android-core/src/main/java/io/getstream/chat/android/models/Poll.kt
```diff
@@ -0,0 +1,123 @@
...
+ * If set to [VotingVisibility.ANONYMOUS], the votes will be anonymous.
+ * @property enforceUniqueVote If set to true, a user can only vote once.
+ * @property maxVotesAllowed The maximum number of votes a user can cast.
+ * @property allowUserSuggestedOptions If set to true, users can suggest new options.
+ * @property allowAnswers If set to true, users can vote.
+ * @property voteCountsByOption The number of votes for each option.
+ * @property votes The list of votes.
+ * @property ownVotes The list of votes cast by the current user.
+ */
+public data class Poll(
```
skydoves, line 38: We can use the `@Immutable` annotation for all these models to make them stable across multi-module architecture.

skydoves (commented):
Overall looks good to me! I left a trivial comment for Compose.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/Poll.kt
JcMinarro: Good point!!!

kanat (approved):
We still miss some endpoints and params for Polls, but overall it looks good 👍

skydoves (approved):

PR #5277, opened 2024-06-10
[PBE-3847] Add the API to block/unblock users and expose it via chat client

aleksandar-apostolov:
### 🎯 Goal

Add API to block/unblock users.

### 🛠 Implementation details

Added API to the `UserApi`

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDtos.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDtos.kt
@@ -89,5 +89,5 @@ internal data class PartialUpdateUserDto(
 internal data class DownstreamUserBlockDto(
     val blocked_by_user_id: String,
     val blocked_user_id: String,
-    val created_at: String,
+    val created_at: Date,
 )
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDtos.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDtos.kt
@@ -84,3 +84,10 @@ internal data class PartialUpdateUserDto(
     val set: Map<String, Any>,
     val unset: List<String>,
 )
+
+@JsonClass(generateAdapter = true)
+internal data class DownstreamUserBlockDto(
+    val blocked_by_user_id: String,
+    val blocked_user_id: String,
+    val created_at: String,
+)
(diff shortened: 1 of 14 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDtos.kt
```diff
@@ -84,3 +84,10 @@ internal data class PartialUpdateUserDto(
     val set: Map<String, Any>,
     val unset: List<String>,
 )
+
+@JsonClass(generateAdapter = true)
+internal data class DownstreamUserBlockDto(
+    val blocked_by_user_id: String,
+    val blocked_user_id: String,
+    val created_at: String,
```
JcMinarro, line 92: On Dtos we normally use `ExactDate` for dates

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/UserBlock.kt
```diff
@@ -0,0 +1,26 @@
...
+package io.getstream.chat.android.models
+
+/**
+ * Represents a user block.
+ *
+ * @param blockedBy who is the initiator of the block (usually will contain the ID of the current user)
+ * @param userId the ID of the blocked user.
+ * @param blockedAt when the action happened
+ */
+public data class UserBlock(val blockedBy: String, val userId: String, val blockedAt: String)
```
JcMinarro, line 26: Can we use `Date` as type for `blockedAt`?
On the other hand, having every param on a new line would be good.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDtos.kt
aleksandar-apostolov: Used `Date` to be consistent in the file.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UserDtos.kt
```diff
@@ -84,3 +84,10 @@ internal data class PartialUpdateUserDto(
     val set: Map<String, Any>,
     val unset: List<String>,
 )
+
+@JsonClass(generateAdapter = true)
+internal data class DownstreamUserBlockDto(
+    val blocked_by_user_id: String,
+    val blocked_user_id: String,
+    val created_at: Date,
```
kanat, line 92: Can we make these fields `nullable` just in case something gets broken on backend side? :)

aleksandar-apostolov: That way we are hiding a potential error and moving the point of failure deeper into the SDK.

kanat (approved):

PR #5283, opened 2024-06-13
[PBE-4089] Store Poll in our local db

JcMinarro:
### 🎯 Goal
Store Poll in our local db.
Depend on: #5282 
Complete https://stream-io.atlassian.net/browse/PBE-4089?atlOrigin=eyJpIjoiYzdhZGU4ZTgyNjFlNDM4M2FiMGUwYzJlMzM5ODY4NmYiLCJwIjoiaiJ9

PR #5285, opened 2024-06-14
[PBE-3844] Feature/polls attachment view

JcMinarro:
### 🎯 Goal
Create a new view to render a poll attachment.
Complete: https://stream-io.atlassian.net/browse/PBE-3844?atlOrigin=eyJpIjoiODNlM2IxOGYwZGQ3NGZlMjhkMjA5NmFmODU0YzUyYWEiLCJwIjoiaiJ9
Depends on: #5283

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/adapter/view/internal/PollView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/adapter/view/internal/PollView.kt
@@ -1,0 +1,273 @@
+/*
+ * Copyright (c) 2014-2024 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.ui.feature.messages.list.adapter.view.internal
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.ViewGroup
+import androidx.core.view.isVisible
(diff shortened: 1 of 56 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/adapter/view/internal/PollView.kt
```diff
@@ -0,0 +1,273 @@
...
+                onClosePollClick = { onClosePollClick(poll) },
+                onViewPollResultsClick = { onViewPollResultsClick(poll) },
+            )
+            adapter = pollAdapter
+        }
+
+        val pollItems = mutableListOf<PollItem>()
+        pollItems.add(PollItem.Header(poll.name, poll.description))
+        println("JcLog [setPoll]: poll.voteCountsByOption -> ${poll.voteCountsByOption}")
+        println("JcLog [setPoll]: poll.votes.count -> ${poll.votes.size}")
```
skydoves, line 86: Maybe we need to use the logger here or remove it?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/adapter/view/internal/PollView.kt
```diff
@@ -0,0 +1,273 @@
...
+                        ?: emptyList(),
+                    voteCount = poll.voteCountsByOption[option.id] ?: 0,
+                    isVotedByUser = poll.ownVotes.any { it.optionId == option.id },
+                    totalVotes = poll.voteCountsByOption.values.sum(),
+                    closed = poll.closed,
+                    isWinner = poll.voteCountsByOption[option.id] == poll.voteCountsByOption.values.maxOrNull(),
+                )
+            },
+        )
+        PollItem.ViewResults.takeUnless { poll.ownVotes.isEmpty() }
```
skydoves, line 102: It seems to miss something here to do after getting the filtered list.

skydoves (approved):
Overall looks good to me. I left some trivial comments :)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/adapter/view/internal/PollView.kt
JcMinarro: Good catch.
They were only debug logs during development that are not needed at all

PR #5303, opened 2024-06-21
Migrate to Kotlin 2.0.0 and Compose Compiler plugin and optimize Gradle maven repositories

skydoves:
### 🎯 Goal

Migrate to Kotlin 2.0.0 and Compose Compiler plugin and optimize Gradle maven repositories.

### 🛠 Implementation details

- Optimized maven repositories on Gradle
- Migrated to Kotlin 2.0.0 and Compose Compiler plugin
- Enabled Strong Skipping Mode to improve Compose performance

### 🎉 GIF

gif


build.gradle
```diff
@@ -20,17 +20,48 @@ apply plugin: MinorBumpPlugin
 apply plugin: VersionPrintPlugin
 apply plugin: ChangelogAddModelSectionPlugin
 
-
 buildscript {
     repositories {
+        // fetch plugins from google maven (https://maven.google.com)
+        google() {
+            content {
+                includeGroupByRegex("androidx\\..*")
```
aleksandar-apostolov, line 28: Why do we need to be so specific here?

skydoves: We can prevent inefficiencies and security risks of Gradle's default repository order. Gradle searches repositories linearly, slowing down builds and exposing the project to malicious artifacts. To optimize build times and enhance security, we can use repository-specific content filtering, including **GroupByRegex**, to limit searches to relevant repositories. This approach helps avoid unnecessary checks and ensures that only the required dependencies are fetched from the correct sources.

This article describes well about this approach: https://jasonatwood.io/archives/2033

skydoves: Due to improper ordering, Gradle couldn't download the JetBrains Compose compiler plugin from the Maven repository for some reasons, as shown in this attempt: https://github.com/GetStream/stream-chat-android/actions/runs/9589327132/job/26442900934

aleksandar-apostolov (approved):

PR #5310, opened 2024-07-01
Update quoted messages composable to use the quoted text style instead.

aleksandar-apostolov:
Small PR to fix an issue where the quoted message doesn't have the correct style.


kanat (approved):
LGTM

PR #5316, opened 2024-07-08
Update quoted messages composable to use the quoted text style instead

JcMinarro:
### 🎯 Goal
Small PR to fix an issue where the quoted message doesn't have the correct style.

This PR was created initially on #5310 but was merged into `main` branch incorrectly, it should be merged into `develop` branch instead.


kanat (approved):
LGTM

PR #5322, opened 2024-07-15
[PBE-5213] After a message is properly sent from the message composer, it is marked as read

JcMinarro:
### 🎯 Goal
After a message is properly sent from the message composer, it is marked as read

Fix: https://stream-io.atlassian.net/browse/PBE-5213

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
@@ -51,7 +51,9 @@ import io.getstream.log.StreamLog
 import io.getstream.log.TaggedLogger
 import io.getstream.result.Result
 import io.getstream.result.call.Call
+import io.getstream.result.call.doOnResult
 import io.getstream.result.call.map
+import io.getstream.result.onSuccessSuspend
 import io.getstream.sdk.chat.audio.recording.StreamMediaRecorder
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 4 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -601,7 +603,15 @@ public class MessageComposerController(
...
-            )
+            ).doOnResult(scope) { result ->
+                result.onSuccessSuspend { resultMessage ->
+                    chatClient.markMessageRead(
+                        channelType = channelType,
+                        channelId = channelId,
+                        messageId = resultMessage.id,
+                    ).await()
+                }
+            }
```
kanat, line 614: @JcMinarro 👋 

I'm wondering why we do need to fire `markMessageRead` after every `sendMessage`.
Looks like something which needs to be done on backend side 🤔

kanat (approved):
LGTM

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
JcMinarro: It is only fired on the case it was properly sent while the user is within the conversation, if, by some reason, the sendMessage process fails and the message is retried to be sent in background when the user is not within the conversation screen, it is not marked as read to avoid sync issues, like it was described here:
https://getstream.slack.com/archives/CRB9P6L03/p1721204370914149?thread_ts=1721150258.252929&cid=CRB9P6L03

PR #5325, opened 2024-07-22
[PBE-5250] Remove `url` property from `Attachment` class

JcMinarro:
### 🎯 Goal
The property `url` from `Attachment` class is not an expected property on backend, what cause it is stored as "Custom Data". The size of the "Custom Data" that an attachment can contains is limited to 5kb on backend side, and because our CDN provides a long URL to any file uploaded to it, the limit is achieved as far as 6~7 attachments are uploaded.
This `url` property is a legacy property that has been replaced by `assetUrl` on the case of normal files and `imageUrl` on the case of pictures.

Fix: https://stream-io.atlassian.net/browse/PBE-5250


kanat (approved):
LGTM

PR #5338, opened 2024-07-30
Enable/Disable polls attachments tab

JcMinarro:
### 🎯 Goal
Enable/Disable polls attachments tab depending on the own capabilities


skydoves (approved):
Looks good to me! 👍

PR #5363, opened 2024-08-20
[PBE-5618] Fix max allowed votes on a poll regarding to the number of available options

JcMinarro:
### 🎯 Goal
The number of max allowed votes wasn't updated properly according to the number of available options.


skydoves (approved):
Looks good to me! 👍

PR #5364, opened 2024-08-20
[PBE-5622] Use poll description instead of show question twice

JcMinarro:
### 🎯 Goal
Use poll description instead of show question twice


skydoves (approved):
Looks good to me!

PR #5366, opened 2024-08-21
[PBE-5626] Show proper votes count for anonymous polls

JcMinarro:
### 🎯 Goal
The anonymous polls don't provide the list of votes but only the total number of votes per options.


skydoves (approved):
Looks good to me! 👍

PR #5367, opened 2024-08-22
[PBE-4155] Ensure last preview message is not a deleted one

JcMinarro:
### 🎯 Goal
When the last message is a deleted one, there isn't any preview available for the channels list.
On the DB we are filtering deleted messages when the channel is stored to ensure the last message is a non-delated one.
On the state plugin we are merging offline messages with the new ones we are receiving querying channels.


kanat (approved):
LGTM

PR #5370, opened 2024-08-24
[PBE-5678] Expose some components needed to build custom MessagesScreen

JcMinarro:
### 🎯 Goal
We were building some internal components used in the MessagesScreen that would be needed by customers who want to build their customer MessagesScreen.
We can expose those components to be used by our customers.


skydoves (approved):
Looks good to me! 👍

PR #5379, opened 2024-08-28
[PBE-5781] Reorder how ThreadState Properties are updated when new messages arrieve

JcMinarro:
### 🎯 Goal
Fix race-condition when the endOfOlderMessage/endOfNewerMessage flag was updated after the messageList is updated


kanat (approved):
LGTM

PR #5382, opened 2024-08-28
[PBE-5799] Keep members updated on QueryChannelsState

JcMinarro:
### 🎯 Goal
Keep members updated on QueryChannelsState


kanat (approved):
LGTM

PR #5391, opened 2024-09-03
[PBE-5854] Provide default theme for the different activities our SDK provides

JcMinarro:
### 🎯 Goal
Provide default theme for the different activities our SDK provides


skydoves (approved):
Looks good to me :)

PR #5392, opened 2024-09-03
[PBE-5853] Deleted messages shouldn't be updated

JcMinarro:
### 🎯 Goal
After any message is deleted, it shouldn't process any update

### DIFF
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
@@ -22,6 +22,7 @@ import io.getstream.chat.android.client.extensions.getCreatedAtOrDefault
 import io.getstream.chat.android.client.extensions.internal.updateUsers
 import io.getstream.chat.android.client.extensions.internal.wasCreatedAfter
 import io.getstream.chat.android.client.extensions.syncUnreadCountWithReads
+import io.getstream.chat.android.client.utils.message.isDeleted
 import io.getstream.chat.android.client.utils.message.isPinned
 import io.getstream.chat.android.models.Channel
 import io.getstream.chat.android.models.ChannelData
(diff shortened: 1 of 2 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
```diff
@@ -170,6 +174,7 @@ internal class DatabaseMessageRepository(
      * @param message [Message]
      */
     override suspend fun deleteChannelMessage(message: Message) {
+        deletedMessageIds.add(message.id)
         messageCache.remove(message.id)
         scope.launchWithMutex(dbMutex) { messageDao.deleteMessage(message.cid, message.id) }
```
skydoves, line 179: Just out of curiosity, shouldn’t we verify that the DAO has successfully deleted the message before invalidating cache and adding the deleted If to the list? Given that the scope could be canceled or other issues might prevent the database from updating, it seems like this could lead to unsynced data.

skydoves (commented):
Looks good to me, and I left a trivial comment :)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
kanat: Hey @skydoves ,

It should be fine. We delete from cache first to prevent `false` reads from cache.
Event if cache is empty we will hit the DB.

kanat:
Hey @JcMinarro,

I'm curious; what was the issue here? :)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
JcMinarro: Deleted messages should not be requested by our SDKs, the only issue was on the insert process, because a deleted message was updated with a wrong states, but as far we included the message id of  the deleted message into the set of deleted messages, it is not updated anymore

JcMinarro:
> Hey @JcMinarro,
> 
> I'm curious; what/where was the issue here? :)

When an action is done over a deleted message (edit, add/remove reaction, ...) Backend returns an error because the message couldn't be found to perform the desired action. Then, on our SDK we handle this as an error over the message and mark it as pending to sync. The problem is that if a Message is deleted, (And its deleted status is into our DB and State), but the message used to perform the action wasn't marked as deleted, we update our "local message" with a failure and the delete flag is overridden, so the message becomes "no deleted"

aleksandar-apostolov:
Lets merge this.

PR #5393, opened 2024-09-03
[PBE-5890] Deprecate `NotInFilterObject`

JcMinarro:
### 🎯 Goal
Deprecate `NotInFilterObject` because it is no longer supported on the backend side.
This class will be removed soon on a future release.


skydoves (approved):
Looks good to me!

kanat (approved):
LGTM

PR #5394, opened 2024-09-03
[PBE-5891] Remove `NotInFilterObject`

JcMinarro:
### 🎯 Goal
Depend on #5393 
This PR removes the `NotInFilterObject` class that is not supported backend-side anymore.
We needs to wait one release before it can be merged


aleksandar-apostolov:
Why do we need to wait for the backend to remove this? I think it will stay there for backward compatibility. Is it an issue if we remove it client side? I think this is the desired behavior.

JcMinarro:
@aleksandar-apostolov We are not waiting to the backend implementation but our deprecation process.
It is a breaking change that we should provide some adaptative time for it

aleksandar-apostolov (approved):

PR #5395, opened 2024-09-04
[PBE-5896] Fix/async voice compose render

JcMinarro:
### 🎯 Goal
Async-Voice attachments weren't rendered properly on Compose SDK on some situations.
Some internal properties caused it, which we use in our attachments to track the upload process.
Apart of that, the attachments factories has been reordered to ensure Async-Voice attachments use the proper factory

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### 🐞 Fixed
 
 ### ⬆️ Improved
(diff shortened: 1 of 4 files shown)


kanat:
Hey @JcMinarro,

Would you mind waiting when I finish the Voice Recording Impl of Compose? :)

kanat (approved):

PR #5397, opened 2024-09-05
Refactor attachment factories

JcMinarro:
### 🎯 Goal
Keep class extensions as classes instead of functions returning an anonymous class simplify our development, helping to the IDE where this clas is being extended and/or used.


kanat (approved):
LGTM

PR #5405, opened 2024-09-13
Implement Poll Result view

JcMinarro:
### 🎯 Goal
Implement a view to show the poll results

### 🎨 UI Changes
Screenshot_20240913_130752

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
@@ -106,8 +106,8 @@ import io.getstream.chat.android.ui.feature.messages.list.background.MessageBack
 import io.getstream.chat.android.ui.feature.messages.list.background.MessageBackgroundFactoryImpl
 import io.getstream.chat.android.ui.feature.messages.list.internal.HiddenMessageListItemPredicate
 import io.getstream.chat.android.ui.feature.messages.list.internal.MessageListScrollHelper
-import io.getstream.chat.android.ui.feature.messages.list.internal.PollResultsDialogFragment
 import io.getstream.chat.android.ui.feature.messages.list.internal.poll.AllPollOptionsDialogFragment
+import io.getstream.chat.android.ui.feature.messages.list.internal.poll.PollResultsDialogFragment
 import io.getstream.chat.android.ui.feature.messages.list.options.message.MessageOptionItem
 import io.getstream.chat.android.ui.feature.messages.list.options.message.MessageOptionItemsFactory
 import io.getstream.chat.android.ui.feature.messages.list.options.message.MessageOptionsDialogFragment
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
@@ -106,6 +106,7 @@ import io.getstream.chat.android.ui.feature.messages.list.background.MessageBack
 import io.getstream.chat.android.ui.feature.messages.list.background.MessageBackgroundFactoryImpl
 import io.getstream.chat.android.ui.feature.messages.list.internal.HiddenMessageListItemPredicate
 import io.getstream.chat.android.ui.feature.messages.list.internal.MessageListScrollHelper
+import io.getstream.chat.android.ui.feature.messages.list.internal.PollResultsDialogFragment
 import io.getstream.chat.android.ui.feature.messages.list.internal.poll.AllPollOptionsDialogFragment
 import io.getstream.chat.android.ui.feature.messages.list.options.message.MessageOptionItem
 import io.getstream.chat.android.ui.feature.messages.list.options.message.MessageOptionItemsFactory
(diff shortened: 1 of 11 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/PollResultsDialogFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.ui.feature.messages.list.internal
```
aleksandar-apostolov, line 17: Question: ...`internal.poll` wdyt?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/MessageListView.kt
```diff
@@ -592,7 +593,14 @@ public class MessageListView : ConstraintLayout {
     }
     private val defaultOnPollOptionClickListener = OnPollOptionClickListener { _, _, _ -> false }
     private val defaultOnPollCloseClickListener = OnPollCloseClickListener { false }
-    private val defaultOnViewPollResultClickListener = OnViewPollResultClickListener { false }
+    private val defaultOnViewPollResultClickListener = OnViewPollResultClickListener { poll ->
+        val fm = context.getFragmentManager() ?: return@OnViewPollResultClickListener false
```
aleksandar-apostolov, line 597: We should avoid a `return` in the elvis operator as its unreadable. Also why not wrap this in `try-catch` as a defensive measure since we are accessing the `context` here.

Something in the lines of
```kotlin
private val defaultOnViewPollResultClickListener = OnViewPollResultClickListener { poll ->
    try {
        val fragmentManager = context.getFragmentManager()
        val viewShown = fragmentManager?.let { nonNullFragmentManager ->
            PollResultsDialogFragment.newInstance(poll)
                .show(nonNullFragmentManager, PollResultsDialogFragment.TAG)
            true
        } ?: false
        viewShown
    } catch (e: Exception) {
        // Log the exception if needed
        false
    }
}
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/PollResultsDialogFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+         * Creates a new instance of [PollResultsDialogFragment].
+         *
+         * @return A new instance of [PollResultsDialogFragment].
+         */
+        public fun newInstance(poll: Poll): PollResultsDialogFragment = PollResultsDialogFragment().apply {
+            this.poll = poll
+        }
+    }
+
+    private class ResultsAdapter(
```
aleksandar-apostolov, line 97: Question: Is there a reason why the `adapter` classes are internal to the `Fragment` and the `holder` classes are in turn private in the `adapter`? We have packages for view-holder classes.

aleksandar-apostolov (changes requested):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/PollResultsDialogFragment.kt
aleksandar-apostolov: Lets keep them in for now.

aleksandar-apostolov (approved):

kanat (approved):

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollResultsDialogFragment.kt
```diff
@@ -0,0 +1,167 @@
...
+import io.getstream.chat.android.ui.utils.extensions.streamThemeInflater
+
+/**
+ * Represent the bottom sheet dialog that allows users to pick attachments.
+ */
+public class PollResultsDialogFragment : AppCompatDialogFragment() {
+
+    private var _binding: StreamUiFragmentPollResultsBinding? = null
+    private val binding get() = _binding!!
+    private lateinit var poll: Poll
```
skydoves, line 44: What are your thoughts on moving this property inside the primary constructor? Alternatively, we could make the primary constructor private, ensuring that instances can only be created via the `newInstance` method.

```kotlin
public class PollResultsDialogFragment(
    private val poll: Poll
) : AppCompatDialogFragment() {
...
```

If not people can create `PollResultsDialogFragment` without passing the `poll` instance and get crashes.

```kotlin
val fragment = PollResultsDialogFragment()
```

skydoves (changes requested):
Overall looks good to me! I left one comment :)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollResultsDialogFragment.kt
aleksandar-apostolov: I second this, we should keep the default constructor.

skydoves (approved):
Looks good to me!

PR #5409, opened 2024-09-19
[PBE-6026] Expose text formater and actions provider on PN Factory

JcMinarro:
### 🎯 Goal
Expose a couple of lambda functions to customize the text and actions shown on our internal PN Implementation.


kanat (approved):
LGTM

PR #5414, opened 2024-09-20
Expose a property on ChatThem to use a custom ImageHadersProvider with Coil

JcMinarro:
### 🎯 Goal
Expose a property on ChatThem to use a custom ImageHadersProvider with Coil
Close: #5271


skydoves (approved):
Looks good to me!

PR #5416, opened 2024-09-25
[PBE-6115] File transformer

JcMinarro:
### 🎯 Goal
Provide a way to transform files before they are uploaded to the CDN.
 It can be used by our customers to compress video/pictures, add watermarks, or whatever local process they would like to apply to the files before they are uploaded to the CDN (Even if they are using our CDN or their own CDN)

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileTransformer.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileTransformer.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2014-2024 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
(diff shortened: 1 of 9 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileTransformer.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Transforms the [file] before uploading it.
+     * This can be used to compress images or videos before uploading them.
+     * It is safe to block
+     *
+     * @return The transformed file.
+     */
+    @WorkerThread
+    public fun transform(file: File): File
```
kanat, line 38: Would it be possible to make as a `suspend` function?

JcMinarro: Making it a `suspend` function will "complicate" Java Devs to implement it :disappointed:

kanat: Indeed, maybe
```kotlin
public fun transform(file: File): Call<File>
```

kanat (approved):

PR #5431, opened 2024-10-04
[PBE-6183] Fix multianswer option on poll

JcMinarro:
### 🎯 Goal
The option to allow multiple votes on a poll wasn't handle properly when the poll is created.


kanat (approved):
LGTM

PR #5415, opened 2024-09-24
Add User-Agent with the App Data

JcMinarro:
### 🎯 Goal

Include a custom User-Agent to any HTTP Call our SDK does. The User-Agent will looks like:
```
"User-Agent:Chat Sample Android / 6.5.0-DEBUG(1); StandAloneInstall; (Google; sdk_gphone64_x86_64; SDK 32; Android 12)"
```

PR #5440, opened 2024-10-15
Sanetize User Agen Header

JcMinarro:
### 🎯 Goal

`OkHTTP` doesn't allow non-ASCII characters on the headers.
On the User-Agent header, we are using the customer app name, which could be whatever character, so it needs to be sanitized to pass internal OkHTTP checks when Headers are added to the query.

eg: `áéíóů` will be transformed to `aeiou`

We introduced User-Agent Header on this PR: #5415


aleksandar-apostolov (approved):
Looks good. We need to fix Klint before merging.

PR #5455, opened 2024-10-23
[PBE-3749] ThreadList improvements

VelikovPetar:
### 🎯 Goal
- Implement missing event handling related to Threads
- Expose `ThreadList` composable component as public API
- Update the compose sample app with a bottom bar for switching between `Chats` and `Threads`
- Add `ThreadList` docs (to be added in the new docs repo)

**Note: Offline support is done separately in a different PR**

### 🛠 Implementation details
- Add/improve handling for the following web socket events: `NotificationChannelDeletedEvent`, `NotificationThreadMessageNewEvent`, `NotificationMarkUnreadEvent`, `MessageReadEvent`, `MessageDeletedEvent`
- Add optimistic state updates of the `QueryThreadsState` in the state plugin listeners: `DeleteMessageListener`, `DeleteReactionListener`, `EditMessageListener`, `SendAttachmentListener`, `SendGiphyListener`, `SendMessageListener`, `SendReactionListener`, `ShuffleGiphyListener`
- Make `ThreadList`/`ThreadItem` configurable via slot-composable
- Adjust `markRead/markUnread` logic in `MessageListController` to support marking threads as read/unread

### 🧪 Testing
1. Install compose sample app
2. Open `Threads` tab
3. Play around with different scenarios: 
- open thread
- mark thread message as unread
- add a new reply to a thread with a different user
- delete a message from a thread with a different user
- delete a thread parent with a different user


docusaurus/docs/Android/compose/thread-components/thread-list.mdx
```diff
@@ -0,0 +1,241 @@
+# Thread List
```
JcMinarro, line 1: Docs have been moved to a new repository, Could you move this doc there?

VelikovPetar: Ah yes, I wrote this before the new docs were published. I will delete these files from here and update the docs repo 👍

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
```diff
@@ -442,7 +448,8 @@ public data class NotificationMarkUnreadEvent(
     val unreadMessages: Int,
     val firstUnreadMessageId: String,
     val lastReadMessageAt: Date,
-    val lastReadMessageId: String,
+    val lastReadMessageId: String?,
```
JcMinarro, line 451: Why is it optional now?
It doesn't make sense for this event.
Could you check with backend?

JcMinarro (approved):

PR #5459, opened 2024-10-30
Feature/clickable mentions

JcMinarro:
### 🎯 Goal
When an user in mentioned on a message, we should highligh it and make it clickable.
Customers can set a new listener that will be called when a mention is clicked:
```
messageListView.setOnMentionClickListener { user ->
  println("A mention to the user $user was clicked")
  true
}
```

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/TextViewLinkHandler.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/TextViewLinkHandler.kt
@@ -21,6 +21,7 @@ import android.text.method.LinkMovementMethod
 import android.text.style.URLSpan
 import android.view.MotionEvent
 import android.widget.TextView
+import io.getstream.chat.android.models.User
 
 internal abstract class TextViewLinkHandler : LinkMovementMethod() {
     override fun onTouchEvent(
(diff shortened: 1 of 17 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/TextViewLinkHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+package io.getstream.chat.android.ui.utils
+
+import android.text.Spannable
+import android.text.method.LinkMovementMethod
+import android.text.style.URLSpan
+import android.view.MotionEvent
+import android.widget.TextView
+import io.getstream.chat.android.models.User
+
+internal abstract class TextViewLinkHandler : LinkMovementMethod() {
```
VelikovPetar, line 26: This class was previously a `public static` class in the `Utils` class, and now it changed to `internal`. From its usages it seems that such change is OK, but I still wanted to check if is OK to directly change the visibility of a public api member.

VelikovPetar (commented):
I just left one clarification point, otherwise looks good!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/TextViewLinkHandler.kt
JcMinarro: Yes.
This class was only used internally in our XML module. It was public on the ui-common module to be accessible from XML module. After moving it to the proper module where it is used and should live, it doesn't make sense to expose it publicly, as it is not a class our customers will need on their end.

VelikovPetar: Perfect, thank you for the explanation!

VelikovPetar (approved):

PR #5463, opened 2024-11-04
Refactor default date separator handle to add date when messages are from different days

### LINKED ISSUE #5454: Duplicate Date Separator
**Describe the bug**
Show the same date separator twice
* Friday, Oct 18
* Monday, Oct 21

**SDK version**
- v6.5.1

**To Reproduce**
Steps to reproduce the behavior: (App:Chat Sample Compose)
1. Go to Chat Room 'Petar Velikov'
2. Scroll up
4. See duplicate date separator

**Expected behavior**
Display the date separator only when messages cross to a new date.

**Device:**
 - Vendor and model: Pixel 6
 - Android version: Android 14

**Screenshots**
image

JcMinarro:
### 🎯 Goal
Fix #5454

Provide a default DateSeparatorHandler adding date separator whenever two messages are from different days

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/DateSeparatorHandler.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/DateSeparatorHandler.kt
@@ -20,6 +20,8 @@ import io.getstream.chat.android.client.extensions.getCreatedAtOrDefault
 import io.getstream.chat.android.client.extensions.getCreatedAtOrNull
 import io.getstream.chat.android.client.extensions.internal.NEVER
 import io.getstream.chat.android.models.Message
+import java.util.Calendar
+import java.util.Date
 
 /**
  * A SAM designed to evaluate if a date separator should be added between messages.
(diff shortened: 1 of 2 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/DateSeparatorHandler.kt
```diff
@@ -38,60 +40,37 @@ public fun interface DateSeparatorHandler {
...
-         * @param separatorTimeMillis Time difference between two message after which we add the date separator.
+         * Creates a [DateSeparatorHandler] returning true if the messages are not in the same day.
          *
          * @return The default thread date separator handler.
          */
-        public fun getDefaultThreadDateSeparatorHandler(
-            separatorTimeMillis: Long = DateSeparatorDefaultHourThreshold,
-        ): DateSeparatorHandler = DateSeparatorHandler { previousMessage, message ->
-            if (previousMessage == null) {
-                false
```
VelikovPetar, line 65: Quick question: Previously we didn't show the date separator before the first message in a thread view, but now the separator will always be shown - just checking if that change was intentional?

JcMinarro: Good Catch!!!
Fixed

VelikovPetar (approved):

PR #5471, opened 2024-11-13
Fix build-and-test.yml workflow

JcMinarro:
### 🎯 Goal
The apk file name was updated by #5464 but it wasn't updated on the workflow that used it.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -95,6 +95,6 @@ jobs:
       - name: Upload artifact to Emerge
         uses: EmergeTools/emerge-upload-action@v1.0.2
         with:
-          artifact_path: ./stream-chat-android-compose-sample/build/outputs/apk/debug/stream-chat-android-compose-sample-debug.apk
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -95,6 +95,6 @@ jobs:
       - name: Upload artifact to Emerge
         uses: EmergeTools/emerge-upload-action@v1.0.2
         with:
-          artifact_path: ./stream-chat-android-compose-sample/build/outputs/apk/debug/stream-chat-android-compose-sample-debug.apk
+          artifact_path: ./stream-chat-android-compose-sample/build/outputs/apk/debug/stream-chat-android-compose-sample-demo-debug.apk
```
VelikovPetar, line 98: Perhaps I am wrong here, but I think that we should also include `/demo/` in the path, because now we have the new `demo` product flavour, which should also create a middle directory in the `build` folder:

> ./stream-chat-android-compose-sample/build/outputs/apk/**demo**/debug/stream-chat-android-compose-sample-demo-debug.apk

JcMinarro: Ups, you are right, it is not only the apk name

JcMinarro: Fixed ;)

PR #5475, opened 2024-11-14
[AND-8] Update compose to v1.7.5. Migrate RippleTheme to RippleConfiguration.

VelikovPetar:
### 🎯 Goal
This PR updates the `compose` and `compose-material` libraries to their latest stable version: `1.7.5` and `1.3.1` respectively, and migrates the usage of `RippleTheme` to `RippleConfiguration`.

There is currently a problem in following scenario:
1. Use any version of the `compose` chat SDK (which has `compose` version `1.6.8`)
2. Integrating app has `compose` version `1.7.+`
3. Integrating app overrides `ChatTheme#rippleTheme`

This issue happens because starting from `compose v1.7.0`, the `RippleTheme` is deprecated with `Level.ERROR`, and it is not possible to use - therefore for an app using `compose` version `1.7.+`, it is not possible to provide a custom `RippleTheme` for `ChatTheme.rippleTheme`.
There is now a replacement for `RippleTheme` called `RippleConfiguration`.

More details: 
- https://developer.android.com/develop/ui/compose/touch-input/user-interactions/migrate-indication-ripple

Because of this, we need to provide a way for users to customise the ripples, even if they are using `compose 1.7.+`. 
(and we decided to update the compose versions as well, because this change will eventually be inevitable)

### 🛠 Implementation details
🚨 **BREAKING CHANGE** 🚨
- Change the `ChatTheme` `rippleTheme: RippleTheme` argument to a compose version-independent `StreamRippleConfiguration`
- The `StreamRippleConfiguration` is now internally mapped to `RippleConfiguration`, keeping the implementation details of the ripple provider internal (and hopefully more future-proof)
- Replace the usage of `LocalRippleTheme` with the new `LocalRippleConfiguration`
- Replace all usages of `rememberRipple()` with the new `ripple()`, which properly queries `LocalRippleConfiguration`


skydoves (approved):
Looks good to me! 🙌

JcMinarro (approved):
Look Good to me.

BTW, Some of our customers are using Chat and Video SDK together, Could we verify same compose versions are being used on Video SDK and if they are not the same, update to to the last one and verify it works fine? 
cc/ @VelikovPetar @skydoves

PR #5488, opened 2024-11-25
[AND-14] Refactor `Channel.lastMessageAt` logic

JcMinarro:
### 🎯 Goal
Previously the SDK was calculating the date of the last message within the channel, but it could cause errors due to deleted messages, system messages or some configuration  that shouldn't be counted for this calculation.
The backend is calculating this value and sending it us on events that could update the content of the channel itself.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/EventAdapter.kt
```diff
@@ -161,6 +161,8 @@ internal class EventDtoAdapter(
 
         val map: Map<String, Any?> = mapAdapter.fromJson(reader)!!.filterValues { it != null }
 
+        println("JcLog: Event type: ${map["type"]} - ${map["channel_last_message_at"]}")
```
VelikovPetar, line 164: Some leftover logs, I believe these can be deleted.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/ChannelMapping.kt
```diff
@@ -21,16 +21,20 @@ import io.getstream.chat.android.client.extensions.syncUnreadCountWithReads
 import io.getstream.chat.android.models.Channel
 import io.getstream.chat.android.models.User
 import io.getstream.chat.android.models.UserId
+import java.util.Date
 
-internal fun DownstreamChannelDto.toDomain(currentUserId: UserId?): Channel =
+internal fun DownstreamChannelDto.toDomain(
+    currentUserId: UserId?,
+    eventChatLastMessageAt: Date?,
```
VelikovPetar, line 28: Quick question for my better understanding: Is it expected that the `DownstreamChannelDto.last_message_at` field can be outdated, and we should prioritise the `eventChatLastMessageAt`? Looks a bit strange that we can get 2 different timestamps with in the same event (one in the channel model, and one in the event)

VelikovPetar (commented):
Left a small remark and a question, but otherwise looks good to me!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/ChannelMapping.kt
JcMinarro: Ideally, it should be the same, but just to ensure we are using the more updated one (The event was created right now, but the channel could be cached somehow) and to avoid any verification in our business logic, I decided to add it here, on the part where the API is converted to our business domain.
It is only overridden when the channel is parsed within a `ChatEvent`

VelikovPetar: Thank you for the explanation!

VelikovPetar (approved):

PR #5494, opened 2024-11-29
[AND-151] Improve attachment factories

JcMinarro:
### 🎯 Goal
Provide a list of non-deseired type of Attachment Factories to easily filter out from our default factories and allow our AudioRecord Attachment to be played from the local recording on the case the url is not ready yet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- The `StreamAttachmentFactories.defaultFactories()` method now accepts a `skipTypes` parameter to skip specific factory types. [#5494](https://github.com/GetStream/stream-chat-android/pull/5494)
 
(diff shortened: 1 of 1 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
```diff
@@ -108,6 +109,7 @@ public object StreamAttachmentFactories {
             previewHandlers: List<AttachmentPreviewHandler>,
             attachment: Attachment,
         ) -> Unit = ::onFileAttachmentContentItemClick,
+        skipTypes: List<AttachmentFactory.Type> = emptyList(),
```
VelikovPetar, line 112: Should we maybe add this to the CHANGELOG?

PR #5500, opened 2024-12-03
[AND-13] Add channel info when searching for messages

JcMinarro:
### 🎯 Goal
When searching for messages, our customers couldn't know the channel data in which this message was posted. It was a bit problematic for 1:1 channel where the data rendered on the screen could be a bit confusing for the user searching for messages.

The list of messages has been pre-processed, recovering the channel data from our internal DB before it is provided to the view.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/SearchResultItem.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/SearchResultItem.kt
@@ -145,7 +145,7 @@ internal fun RowScope.DefaultSearchResultItemCenterContent(
         verticalArrangement = Arrangement.Center,
     ) {
         Text(
-            text = ChatTheme.searchResultNameFormatter.formatMessageTitle(searchResultItemState.message, currentUser),
+            text = ChatTheme.searchResultNameFormatter.formatMessageTitle(searchResultItemState, currentUser),
             style = ChatTheme.typography.bodyBold,
             fontSize = 16.sp,
             maxLines = 1,
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/SearchResultItem.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/SearchResultItem.kt
@@ -67,6 +67,7 @@ public fun SearchResultItem(
     leadingContent: @Composable RowScope.(ItemState.SearchResultItemState) -> Unit = {
         DefaultSearchResultItemLeadingContent(
             searchResultItemState = it,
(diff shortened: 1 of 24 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/SearchResultItem.kt
```diff
@@ -107,22 +108,33 @@ public fun SearchResultItem(
...
-        modifier = Modifier
-            .padding(
-                start = ChatTheme.dimens.channelItemHorizontalPadding,
-                end = 4.dp,
-                top = ChatTheme.dimens.channelItemVerticalPadding,
-                bottom = ChatTheme.dimens.channelItemVerticalPadding,
+    (
+        searchResultItemState
+            .channel
+            ?.takeIf { it.members.size == 2 }
```
VelikovPetar, line 121: Quick question about this logic: If we have a named group with exactly 2 members (`me` and `other`), search results of messages from `me` would look like:

(avatar of `other`) (`me` in `Group name`)

I was wondering if this is intentional, because the shown avatar would be different from the user who sent the message? (this I guess makes sense for direct messages, but I wanted to check if it is intentional also for 1-to-1 named groups)
Ex: 
search_in_group

JcMinarro: I think so.
At the end, even if the channel has name, it is a 1:1 conversation

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/SearchResultNameFormatter.kt
```diff
@@ -52,19 +58,31 @@ public fun interface SearchResultNameFormatter {
...
+                withStyle(SpanStyle(fontWeight = FontWeight.Bold)) {
+                    append(searchResultItem.message.user.name)
                 }
+                (
+                    searchResultItem.channel
+                        ?.let { ChatTheme.channelNameFormatter.formatChannelName(it, currentUser) }
+                        ?: searchResultItem.message.channelInfo?.name
+                    )
+                    ?.let { channelName ->
+                        append(" in ")
```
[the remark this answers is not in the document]
JcMinarro: I was thinking to extract it while I was refactoring this part but at the end I forgot it :man_facepalming:

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/SearchResultItem.kt
VelikovPetar: Got it, thank you!

PR #5510, opened 2024-12-09
[AND-159] Ensure ChatClient is initialized before internal SDK activities are started

JcMinarro:
### 🎯 Goal
Ensure ChatClient is initialized before internal SDK activities are started


VelikovPetar (commented):
I believe there is another activity that should be covered with such check: `AttachmentDocumentActivity` from the `stream-chat-android-ui-common` module.

VelikovPetar (approved):

PR #5512, opened 2024-12-10
[AND-169] Add an option to show/hide block user action on the MessageOptions

JcMinarro:
### 🎯 Goal

Add an option to show/hide block user action on the MessageOptions


VelikovPetar (approved):
Looks good to me!

PR #5539, opened 2024-12-31
AND-116: SDK size reporting (APK comparison)

andremion:
#### Show SDK size badges in the README file, e.g.:

stream-chat-android-client

#### Compare sizes before and after a PR and show them in a comparison table as a PR comment, e.g.:

| SDK | Before | After | Difference | Status |
|-|-|-|-|-|
| stream-chat-android-client | 2.68 MB | 2.68 MB | 0.00 MB | 🟢 |

Max tolerance is `500 KB`
Fine tolerance is `250 KB`

if the difference is less than zero, then the status is 🚀
if the difference is greater or equal to the max tolerance, then the status is 🔴
if the difference is greater or equal to the fine tolerance, then the status is 🟡
otherwise, the status is 🟢

#### Sizes are calculated for the following SDKs:

- stream-chat-android-client
- stream-chat-android-offline
- stream-chat-android-ui-components
- stream-chat-android-compose

### 🛠 Implementation details

- There are two metrics Android application projects for each SDK:
  - One is used as a baseline, and the other includes the SDK

- The size calculation is achieved by comparing a baseline APK with an APK that includes the SDK

- The comparison task is triggered on every push request

- The metrics update is triggered on every merge to develop

### 🎉 GIF

Size matters


build.gradle
```diff
@@ -127,6 +127,14 @@ apiValidation {
...
             'stream-chat-android-compose-sample',
             'stream-chat-android-ui-guides',
+            'stream-chat-android-client-metrics-baseline',
+            'stream-chat-android-client-metrics-stream',
+            'stream-chat-android-offline-metrics-baseline',
+            'stream-chat-android-offline-metrics-stream',
+            'stream-chat-android-ui-components-metrics-baseline',
+            'stream-chat-android-ui-components-metrics-stream',
+            'stream-chat-android-compose-metrics-baseline',
+            'stream-chat-android-compose-metrics-stream',
```
JcMinarro, line 137: Can't we use only one gradle project with multiple flavors??

README.md
```diff
@@ -12,6 +12,15 @@
...
 </p>
 
+<div align="center">
+
+![stream-chat-android-client](https://img.shields.io/badge/stream--chat--android--client-3.01%20MB-lightgreen)
+![stream-chat-android-offline](https://img.shields.io/badge/stream--chat--android--offline-3.22%20MB-lightgreen)
+![stream-chat-android-ui-components](https://img.shields.io/badge/stream--chat--android--ui--components-7.88%20MB-lightgreen)
+![stream-chat-android-compose](https://img.shields.io/badge/stream--chat--android--compose-9.70%20MB-lightgreen)
+
+</div>
```
JcMinarro, line 22: Should we include some "note" about this SDK size notifying it is the max size our SDK will include to their app on the hypothetical case their app is not using any of the dependencies our SDK uses (Almost impossible, because lot of them would need to be used)?

.github/workflows/sdk-size-updates.yml
```diff
@@ -0,0 +1,97 @@
...
+
+          # Add and commit updated metrics file
+          git add $METRICS_FILE
+          git commit -m "Update SDK size metrics" || echo "No metrics changes to commit"
+          
+          # Add and commit updated README file
+          git add README.md
+          git commit -m "Update SDK size badges" || echo "No README changes to commit"
+
+          git push origin HEAD:$BRANCH_NAME
```
JcMinarro, line 97: `develop` branch is protected, and any commit needs to be pushed using a PR. The only allowed "user" is a bot.
On this [GH Action](https://github.com/GetStream/stream-chat-android/blob/develop/.github/workflows/release-post.yaml#L30-L56) you can see how we use is

.github/workflows/sdk-size-updates.yml
```diff
@@ -0,0 +1,97 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"         
+
+          # Add and commit updated metrics file
+          git add $METRICS_FILE
+          git commit -m "Update SDK size metrics" || echo "No metrics changes to commit"
+          
+          # Add and commit updated README file
+          git add README.md
+          git commit -m "Update SDK size badges" || echo "No README changes to commit"
```
JcMinarro, line 95: We should only add one commit with both changes

metrics/stream-chat-android-client-metrics-baseline/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+            signingConfig = signingConfigs.findByName("debug")
+        }
+    }
+}
+
+dependencies {
+    detektPlugins(Dependencies.detektFormatting)
+
+    implementation(Dependencies.coroutinesCore)
+    implementation(Dependencies.androidxAppCompat)
```
JcMinarro, line 23: We have already migrated to Version Catalog :upside_down_face:

metrics/stream-chat-android-compose-metrics-stream/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+    implementation(Dependencies.composeUiTooling)
+    implementation(Dependencies.composeFoundation)
+    implementation(Dependencies.composeMaterial)
+    implementation(Dependencies.composeMaterialIcons)
+
+    implementation(Dependencies.composeActivity)
+    implementation(Dependencies.composeAndroidLifecycle)
+    implementation(Dependencies.composeViewModel)
+    implementation(Dependencies.composeAccompanistPermissions)
+    implementation(Dependencies.composeAccompanistPager)
```
JcMinarro, line 34: Aren't they transitive dependency? :thinking: 
If not, they should, our Customers shouldn't need external dependencies to their manifest to be able to use our SDK

JcMinarro (changes requested):

build.gradle
andremion: Good point, I will check that, thanks 🙌🏻

.github/workflows/sdk-size-updates.yml
andremion: The script is already using the GitHub action bot 🤔

JcMinarro: No, it isen.
It is only configuring git properties, but it is not using the bot credentials

andremion: How come? 🤔 
Screenshot 2025-01-02 at 12 00 28

andremion: Screenshot 2025-01-02 at 12 06 53

JcMinarro: It is because it shows the git properties for the commit itself, but they are not the credentials.
You could configure them within your local machine as well

JcMinarro (approved):

PR #5543, opened 2025-01-02
Add label to the commit to avoid rerun unnecessary workflows

JcMinarro:
### 🎯 Goal

Add label to the commit to avoid rerun unnecessary workflows

### DIFF
--- a/.github/workflows/sdk-size-updates.yml
+++ b/.github/workflows/sdk-size-updates.yml
@@ -86,7 +86,7 @@ jobs:
           git config user.email "41898282+github-actions[bot]@users.noreply.github.com"         
 
           git add $METRICS_FILE README.md
-          git commit -m "Update SDK sizes" || echo "No changes to commit"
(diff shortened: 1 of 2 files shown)


.github/workflows/sdk-size-updates.yml
```diff
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - develop
-
+    paths-ignore:
+      - 'README.md'
```
andremion, line 8: In that case, we don't need that anymore

JcMinarro: If, by some reason, only the Readme has been updated, we don't really need to rerun it either, right?

andremion: By using [skip ci], no CI is triggered for any commit, no mater what was changed

andremion: Many other files should not trigger this workflow but I think isn't worth it to include all of them there

andremion (approved):
Thank you 🙌🏻

PR #5550, opened 2025-01-09
[AND-215] Refactor how Global State related events are handled

JcMinarro:
### 🎯 Goal
Refactor how events are processed while updating globalState to ensure any event marked as `HasUnread` is properly updating the unread counters.


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -290,76 +291,61 @@ internal class EventHandlerSequential(
...
+        val modifyValuesFromEvent: suspend (HasUnreadCounts) -> Unit = {
+            when (it) {
+                is MarkAllReadEvent -> it
+                is NewMessageEvent -> it.takeIf { hasReadEventsCapability(it.cid) }
+                is NotificationAddedToChannelEvent -> it.takeUnless { it.unreadChannels == 0 }
+                is NotificationChannelDeletedEvent -> it.takeUnless {
+                    it.unreadChannels == 0 && hasReadEventsCapability(it.cid)
+                }
+                is NotificationChannelTruncatedEvent -> it.takeIf { hasReadEventsCapability(it.cid) }
+                is NotificationMarkReadEvent -> it
```
VelikovPetar, line 308: Question: We previously checked whether the user `hasReadEventsCapability` when handling `NotificationMarkReadEvent`, `NotificationMarkUnreadEvent ` and `NotificationMessageNewEvent`, is the removal of the check now intentional?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -290,76 +291,61 @@ internal class EventHandlerSequential(
...
+                }
         }
 
-        val hasReadEventsCapability = parameterizedLazy<String, Boolean> { cid ->
-            // can we somehow get rid of repos usage here?
-            repos.hasReadEventsCapability(cid)
+        val modifyUnreadThreadsCount: suspend (HasUnreadThreadCounts) -> Unit = {
+            when (it) {
+                is NotificationThreadMessageNewEvent -> it.takeIf { hasReadEventsCapability(it.cid) }
+                is NotificationMarkReadEvent -> it
```
VelikovPetar, line 331: Same here: Was the removal of the `hasReadEventsCapability` check intentional?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
JcMinarro: `NotificationXXXEvent`s should only be received when the user is not watching this channel. If they are received, is because backend already checked we have those read capabilities to send us this info, in another case no read events should be received.

VelikovPetar: Makes sense, thanks for the explanation!

VelikovPetar (approved):

PR #5553, opened 2025-01-13
[AND-0] Add local-release script for Chat.

VelikovPetar:
### 🎯 Goal
Adds a `local-release.sh` script which simplifies the process of making a local maven release of each artefact of the Chat SDK.
Usage:
```sh
sh scripts/local-release.sh
```
After invoking the script, you can enter the snapshot version manually, or if not, it will automatically increment the patch version, and append `-local-TIMESTAMP` suffix. Ex:
```
6.9.1-local-20250113093459
```

You can then use it in ex. the video `:demo-app`, where you need to:
- Add `mavenLocal()` as a repository in `settings.gradle`:
```kotlin
dependencyResolutionManagement {
    repositories {
        // ...
        mavenLocal()
    }
}
```
- Update the version of the ChatSdk in `libs.versions.toml`:
```
streamChat = "6.9.1-local-20250113093459"
```
- OR add the dependency directly:
```kotlin
implementation("io.getstream:stream-chat-android-compose:6.9.1-local-20250113093459")
```
### 🛠 Implementation details
Implementation adapted from: https://github.com/GetStream/stream-video-android/blob/develop/scripts/release/local-release


JcMinarro:
It is not working on my side:
```
$ sh scripts/local-release.sh
scripts/local-release.sh: 40: let: not found
Enter version (Press Enter to use 6.9.0 or format X.Y.Z): 
scripts/local-release.sh: 72: Syntax error: redirection unexpected
```
Maybe because wrong Shell Interpreter?? (It is run with `sh` but the definition is with `#!/bin/bash`

I did the script executable and I am having another error:
```
./scripts/local-release.sh
Enter version (Press Enter to use 6.9.3 or format X.Y.Z): 
Version updated to: \033[33m6.9.3-local-20250114094816\033[0m
sed: can't read /signing {/,/}/d: No such file or directory
```

VelikovPetar:
Hey @JcMinarro ,

I will double check it, works quite well on my side 🤔

PR #5558, opened 2025-01-14
[AND-248] Use support library for vector drawables in compose SDK.

VelikovPetar:
### 🎯 Goal
There is currently a 'problem' with the drawable resources that we have, which use the `fillType` property: If we have `vectorDrawable.useSupportLibrary = FALSE`, the resources which use the `fillType` functionality will be packaged in the directories: `drawable-anydpi-v24` (as xml) and in `drawable-ldpi-v4`(and other densities) (as png). Resources which don't use the `fillType` function seem to be OK, as they are packaged just in the `drawable` directory (as xml), and they can be overridden as usual. This is because for API 24 and below, the `fillType` is not supported, so the Vector Drawables will be packaged as raster `.png`s for each screen density in the `.aar`.
By setting the `vectorDrawables.useSupportLibrary = TRUE`, all the resources are packaged in the drawable directory in the aar, and they can be overridden easily.

Keeping `vectorDrawables.useSupportLibrary = FALSE` means that customers need to override a given icon in both the `drawable` and `drawable-anydpi-v24`(and in each `drawable-[x]dpi` directory, if they support API lvl <24) to ensure that the icon will be overridden properly, because the icons from the compose SDK will be packaged in either of them (based on the usage of <24 unsupported functions)

More details here: https://developer.android.com/develop/ui/views/graphics/vector-drawable-resources#vector-drawables-backward-solution

### 🛠 Implementation details
- Set `vectorDrawables.useSupportLibrary = TRUE` in `stream-chat-android-compose`

### 🧪 Testing
Testing before change:
1. Override `stream_compose_ic_new_chat.xml` and `stream_compose_ic_thread.xml` in the `stream-chat-android-compose-sample` `res/drawable` folder
2. Only `stream_compose_ic_thread.xml` will be replaced successfully
3. _`stream_compose_ic_new_chat.xml` needs to be overridden in `res/drawable-anydpi-v24`_

Testing after change
1. Override `stream_compose_ic_new_chat.xml` and `stream_compose_ic_thread.xml` in the `stream-chat-android-compose-sample` `res/drawable` folder
2. Both icons should be replaced with new ones


skydoves (approved):
This makes sense. LGTM

andremion:
> ## SDK Size Comparison 📏
> SDK	Before	After	Difference	Status
> stream-chat-android-client	3.04 MB	3.04 MB	0.00 MB	🟢
> stream-chat-android-offline	3.25 MB	3.25 MB	0.00 MB	🟢
> stream-chat-android-ui-components	7.91 MB	7.91 MB	0.00 MB	🟢
> stream-chat-android-compose	9.14 MB	8.73 MB	-0.41 MB	🚀

Interesting that this change also decreased the SDK size 👏🏻

PR #5568, opened 2025-01-20
[AND-257] Introduce API Model Transformer

JcMinarro:
### 🎯 Goal
Introduce API Model Transformer to allow our customers transformations of API models.
It could be useful on the case they want to to implement E2E encription over the data they put within our entities.
With those transformers, entities can be transformed before they are sent to the backend and after they are parsed from the API responses.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChannelExtension.kt
```diff
@@ -92,15 +93,11 @@ public fun Channel.countUnreadMentionsForUser(user: User): Int {
...
-/**
- * Returns the number of unread messages in the channel for the current user.
- *
- * @return The number of unread messages in the channel for the current user.
- */
-public val Channel.currentUserUnreadCount: Int
-    get() = ChatClient.instance().getCurrentUser()?.let { currentUser ->
-        read.firstOrNull { it.user.id == currentUser.id }?.unreadMessages
-    } ?: 0
+public fun Channel.currentUserUnreadCount(
```
VelikovPetar, line 96: I think that we should keep the KDoc here, as this is still a public method.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ApiModelTransformers.kt
```diff
@@ -0,0 +1,41 @@
...
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.utils
```
VelikovPetar, line 17: Suggestion about the package: Would placing this class into something like `io.getstream.chat.android.client.transformer`(or something similar) maybe be more appropriate than the `utils` package? 
Which would be more specific package for this case, instead of using the general utility package.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/ApiModelTransformers.kt
```diff
@@ -0,0 +1,41 @@
...
+ * A class that holds the transformers used to transform the API models.
+ *
+ * @param sendMessageTransformer The transformer used to transform the message before sending it to the API.
+ * @param receiveMessageTransformer The transformer used to transform the message received from the API.
+ * @param receiveChannelTransformer The transformer used to transform the channel received from the API.
+ * @param sendUserTransformers The transformer used to transform the user before sending it to the API.
+ * @param receiveUserTransformer The transformer used to transform the user received from the API.
+ */
+public class ApiModelTransformers(
+    public val sendMessageTransformer: MessageTransformer = NoOpMessageTransformer,
```
VelikovPetar, line 36: A suggestion: Maybe we can use `incoming/outgoing` prefix instead of the `send/receive` prefix?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/EventMapping.kt
```diff
@@ -138,887 +137,1068 @@ import io.getstream.chat.android.client.events.VoteCastedEvent
...
-        channelType = channel_type,
-        channelId = channel_id,
-        user = user.toDomain(currentUserId),
-        message = message?.toDomain(currentUserId),
-        channel = channel.toDomain(currentUserId, channel_last_message_at),
-        channelLastMessageAt = channel_last_message_at,
-    )
-}
+    /**
+     * Transforms [ChannelUserBannedEventDto] to [ChannelUserBannedEvent].
```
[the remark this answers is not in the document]
JcMinarro: Ouch!!!
Good catch!!!

VelikovPetar (approved):

PR #5570, opened 2025-01-20
Create snapshot tests for channels stateless components

andremion:
### 🎯 Goal

Provide test coverage to the stateless Composable functions before changing them so we can ensure no side effects are introduced.

### 🛠 Implementation details

- Added snapshot tests for `SearchInput`, `ChannelListHeader`, and `ChannelList`.

#### Small improvements

- Fix the IDE scope of library modules, so that we can search in the library code, ignoring docs and test code.
- Refactoring on `BaseComposeTest`
  - Split in:
    - `SnapshotTest` interface: The setup for `Paparazzi`
    - `ComposeTest` interface: Binds a mocked `ChatClient` on every test run and exposes it for use by any `ComposeTest` subclass.

- Use distinct channel IDs of `PreviewChannelData` to allow using all the preview channels in a single `ChannelList`

- Fix previews of `ChannelList` and `ChannelItem`.

- Use default `typingUsers` as an empty list for convenience.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelList.kt
```diff
@@ -372,6 +373,8 @@ public fun DefaultChannelItemDivider() {
 @Preview(showBackground = true, name = "ChannelList Preview (Content state)")
 @Composable
 private fun ChannelListForContentStatePreview() {
+    ChatClient.Builder(apiKey = "apiKey", appContext = LocalContext.current).build()
```
VelikovPetar, line 376: There is already a `ChatPreviewTheme` which does a similar setup - perhaps we can use that one? (same goes for the `ChannelItemPreview`)

skydoves (approved):
Overall looks great to me! 👍

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelList.kt
skydoves: +1

andremion: Good spot, @VelikovPetar 🙌🏻 
I will take a look 😉

PR #5577, opened 2025-01-23
Use Remote Gradle Build Cache on E2E testing workflows

JcMinarro:
### 🎯 Goal

Use Remote Gradle Build Cache on E2E testing workflows


andremion (approved):
👌🏻

PR #5578, opened 2025-01-23
[AND-267] Snapshot tests for messages stateless components

andremion:
### 🎯 Goal

Create snapshot tests to cover stateless components used in the message list.

### 🛠 Implementation details

Besides creating tests for messages stateless components, here are some side tasks that were done:

- Upgrade the Paparazzi library from 1.3.4 to [1.3.5](https://github.com/cashapp/paparazzi/releases/tag/1.3.5) to fix an issue where snapshots were not waiting for recompositions. Hence, we couldn't use, for example, `LaunchedEffect` to scroll before taking the snapshot. The new version creates snapshot files without spaces. That's why we see so many changes in the snapshot files, but some are just automatic renaming.
- Make channel tests to take snapshots in dark mode alongside the light mode.

### 🎉 GIF

gif


stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/channels/ChannelListHeaderTest.kt
```diff
@@ -28,11 +28,11 @@ import org.junit.Test
 internal class ChannelListHeaderTest : SnapshotTest {
 
     @get:Rule
-    override val paparazzi = Paparazzi(deviceConfig = DeviceConfig.PIXEL_4A)
+    override val paparazzi = Paparazzi(deviceConfig = DeviceConfig.PIXEL_2)
```
VelikovPetar, line 31: I was wondering why did the `DeviceConfig` changed here?

andremion: Good point.
The reason was to decrease the size of the snapshot image hence, the comparison will be a bit faster.
However, there are some test cases where the UI won't fit entirely in a `PIXEL_2` screen. I'm using `PIXEL_4A` otherwise.

VelikovPetar (approved):

PR #5580, opened 2025-01-23
Upgrade agp

JcMinarro:
### 🎯 Goal
Upgrade Android Gradle Plugin and depend dependencies


stream-chat-android-core/src/main/java/io/getstream/chat/android/models/FilterObject.kt
```diff
@@ -21,31 +21,61 @@ package io.getstream.chat.android.models
  */
 public sealed class FilterObject
 
+@ConsistentCopyVisibility
```
andremion, line 24: What about using the `-Xconsistent-data-class-copy-visibility` compiler flag instead? So that we could change a single line when this feature becomes stable.

gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 allureKotlin = "2.4.0"
-androidGradlePlugin = "8.4.2"
+androidGradlePlugin = "8.8.0"
```
andremion, line 3: ℹ️ `stream-android-push` is using `8.6.1` with Kotlin `2.0.20`

andremion (commented):
🚀 Thanks @JcMinarro 
If there's no side effect of these updates, I'm OK with that 👏🏻

gradle/libs.versions.toml
JcMinarro: I think it should be fine, at the end it is used for compilation process, shouldn't influe how other SDKs were compiled

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/FilterObject.kt
andremion: For reference: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-consistent-copy-visibility/

VelikovPetar (approved):

Issue #5584, opened 2025-01-24
The DateSeparator not follow DateFormatter


quangchien99:
In the **ChannelList screen (Image 1)** -> the date for last message is **display ok based on the DateFormatter**.
However, in the **MessageList screen (Image 2)** -> the Date Separator is **_display differently ("4 days ago" instead of "Monday")_**

<img width="848" alt="Image" src="https://github.com/user-attachments/assets/aa2d9d05-1c71-4fff-9fae-1a279915b355" />

I checked the code in the SDK, the DefaultMessageDateSeparatorContent currently using DateUtils.getRelativeTimeSpanString to format the content.

<img width="703" alt="Image" src="https://github.com/user-attachments/assets/f245d4b1-6001-411a-9d60-89a4dd21afb9" />

Can the SDK make Date Separator **follow DateFormatter** or **expose another Date Separator to custom**?

PR #5587, opened 2025-01-27
[AND-275] Expose a new method to return a representation of relative date

JcMinarro:
### 🎯 Goal
Expose a new method on `DateFormatter` to return a representation of relative data.
It is now used within our "Date Separator Views" on both SDKs, XML and Compose.
The default implementation continue using `DateUtils` from Android SDK, but it is customizable by our customers providing a custom implementation of `DateFormatter`

Fix: #5584

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/DateFormatter.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/DateFormatter.kt
@@ -39,6 +39,8 @@ public interface DateFormatter {
 
     /**
      * Formats the given date as a String.
+     * A normal implementation of this method would return a string like "Yesterday", "Saturday", "9:25 AM"/"9:25", ...
+     * It is used to format the date of the last message in the channel list and the when an user voted in a poll.
      *
      * @param date The [Date] to format as a String.
      * @return The formatted date-time string.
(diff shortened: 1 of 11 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/DateFormatter.kt
```diff
@@ -278,9 +301,14 @@ internal class DefaultDateFormatter(
         fun dateTimePattern(): String
 
         /**
-         * Returns the context.
+         * Returns the relative time string for the given date.
          */
-        fun context(): Context
+        fun relativeTime(date: Date): String
```
VelikovPetar, line 306: Looks good to me, I just have a small concern about whether it is easy enough for the integrators to know which of the `DateFormatter` methods are used in which places - and to know which method needs to be overridden for which use-case.  

I don't think we should change anything now, but maybe we should consider in the future making these methods a bit 'context-aware', so that it more obvious which methods is used for which purpose. 
Maybe we could also add some more detailed KDoc, to be easier to understand the purpose of each method?

JcMinarro: Yes, it is not easy to know which method is used on every part of our SDK.
Let me see how I can add some details to the KDoc

PR #5590, opened 2025-01-27
[AND-276] Enable channel pinning and provide example of ChatComponentFactory customization for pinned channels

VelikovPetar:
### 🎯 Goal
- Provide example usage of the `ChatComponentFactory` for customising the 'channel items' when using the high-level `ChannelsScreen`.
- Enable channel pinning in the compose sample app

### 🛠 Implementation details
- Override `ChatComponentFactory.ChannelList` to highlight the item background if the channel is pinned
- Pass `CustomChatComponentFactory` to `ChatTheme` in `ChannelsActivity`
- Unrelated change: Add missing KDoc to `Channel.isAnonymousChannel()` extension.
- Unrelated change: Hide the 'Members list' from `MessagesActivity`.

### 🎨 UI Changes
<table>
<thead>
<tr>
<th>Demo</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/8c25e8b4-5196-4f0e-ac83-73303161d57a" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Open compose sample
2. Long press on Channel
3. Click "Pin channel" in the menu
4. The channel should now be highlighted (and vice-versa for unpin)


andremion (approved):
Looks good! 👌🏻

aleksandar-apostolov:
Lets sort it so pinned channels are on top.

VelikovPetar:
With ordering applied (pinned channels on top):


https://github.com/user-attachments/assets/f1393e55-f8a5-419e-b49f-a33f0b6920f2

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt
```diff
@@ -91,7 +91,9 @@ class ChannelsActivity : BaseConnectedActivity() {
         val currentUserId = chatClient.getCurrentUser()?.id ?: ""
         ChannelViewModelFactory(
             chatClient = chatClient,
-            querySort = QuerySortByField.descByName("last_updated"),
+            querySort = QuerySortByField
+                .descByName<Channel>("pinned_at") // pinned channels first
+                .desc("last_updated"), // then by last updated
```
JcMinarro, line 96: Should we add it as default query sort within our factory? :thinking:

VelikovPetar: I was thinking that as well, but in the end we decided to keep the custom sort order only in the sample app, similar to how the other UI customisations (related to channel pinning) are only part of the sample and not the SDK. 
This is also currently aligned with iOS, who also don't apply ordering by `pinned_at` at SDK level.

aleksandar-apostolov (approved):

PR #5593, opened 2025-01-28
[AND-253] Message stateless component factories

andremion:
### 🎯 Goal

- Create message stateless component factories.
- Refactor to have a single component factory object.

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
@@ -16,7 +16,7 @@
 
 package io.getstream.chat.android.compose.sample.ui.component
 
-import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.RowScope
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyItemScope
@@ -31,11 +31,20 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import io.getstream.chat.android.compose.state.channels.list.ItemState
+import io.getstream.chat.android.compose.state.mediagallerypreview.MediaGalleryPreviewResult
+import io.getstream.chat.android.compose.ui.components.messages.factory.MessageContentFactory
 import io.getstream.chat.android.compose.ui.theme.ChatComponentFactory
 import io.getstream.chat.android.compose.ui.theme.ChatTheme
-import io.getstream.chat.android.models.Channel
 import io.getstream.chat.android.models.ConnectionState
+import io.getstream.chat.android.models.Message
+import io.getstream.chat.android.models.Option
+import io.getstream.chat.android.models.Poll
+import io.getstream.chat.android.models.ReactionSorting
 import io.getstream.chat.android.models.User
+import io.getstream.chat.android.models.Vote
+import io.getstream.chat.android.ui.common.state.messages.list.GiphyAction
+import io.getstream.chat.android.ui.common.state.messages.list.MessageItemState
+import io.getstream.chat.android.ui.common.state.messages.poll.PollSelectionType
 
 class CustomChatComponentFactory : ChatComponentFactory(
     channelListHeader = object : ChannelListHeader() {
@@ -93,18 +102,6 @@ class CustomChatComponentFactory : ChatComponentFactory(
         }
     },
     channelList = object : ChannelList() {
-        @Composable
-        override fun LazyItemScope.ChannelItemContent(
-            channelItem: ItemState.ChannelItemState,
-            currentUser: User?,
-            onChannelClick: (Channel) -> Unit,
-            onChannelLongClick: (Channel) -> Unit,
-        ) {
-            Text(
-                text = channelItem.channel.name,
-                color = ChatTheme.colors.textHighEmphasis,
-            )
-        }
 
         @Composable
         override fun LoadingIndicator(modifier: Modifier) {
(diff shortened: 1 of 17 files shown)


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
```diff
@@ -31,95 +31,113 @@ import androidx.compose.ui.Modifier
...
-                text = "Helper content",
-                color = ChatTheme.colors.textLowEmphasis,
-            )
-        }
-    },
-)
+    @Composable
+    override fun ChannelListLoadingIndicator(modifier: Modifier) {
+        Text(
+            text = "Loading...",
```
aleksandar-apostolov, line 104: Is it possible to provide a `context` to the factory and load the string from resources?

andremion: That is a `CustomChatComponentFactory` as an example.
Customers can implement `ChatComponentFactory`  and pass any parameter as needed

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
```diff
@@ -84,10 +85,10 @@ import io.getstream.chat.android.ui.common.state.messages.poll.PollSelectionType
  * or [MessageListController.showDateSeparatorInEmptyThread].
  */
 @Composable
-public fun MessageContainer(
+public fun LazyItemScope.MessageContainer(
```
aleksandar-apostolov, line 88: Adding the `LazyItemScope` is a breaking change here, no?
Isn't it better to mark this as deprecated and have a new `MessageContainer` with `LazyItemScope` as receiver.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
```diff
@@ -62,16 +63,31 @@ public fun MessageContent(
...
     onLinkClick: ((Message, String) -> Unit)? = null,
-    messageContentFactory: MessageContentFactory = ChatTheme.messageContentFactory,
+    messageContentFactory: MessageContentFactory = MessageContentFactory.Deprecated,
     onMediaGalleryPreviewResult: (MediaGalleryPreviewResult?) -> Unit = {},
     giphyEphemeralContent: @Composable () -> Unit = {
-        messageContentFactory.MessageGiphyContent(
-            message = message,
-            onGiphyActionClick = onGiphyActionClick,
-        )
+        if (messageContentFactory == MessageContentFactory.Deprecated) {
```
VelikovPetar, line 69: I could be missing something, but doesn't this break the backwards compatibility with the `MessageContentFactory`?
I can see that currently the `ChatTheme.messageContentFactory` is no longer used, so anyone already overriding it will be forced to change the implementation. 

We should be pushing the usage of `ChatComponentFactory` instead of `MessageContentFactory`, but I don't think that we should fully break the backwards compatibility. I think we should either have the `ChatComponentFactory` fallback to the `MessageComponentFactory` and go through the deprecation process (WARNING -> ERROR -> DELETE).
Or if you think it is OK to introduce this breaking change, to maybe directly mark the `MessageComponentFactory` as `Deprecated(ERROR)`, so that it is obvious that it will not work anymore (I think `Deprecate(WARNING)` shouldn't break the current behaviour)

aleksandar-apostolov: The default implementation of the `ChatComponentFactory` should call `MessageComponentFactory`, we cannot simply remove it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
andremion: This is the only breaking change if a customer uses it out of any LazyList component.
That scope was introduced because no list item content was exposing it, and this is one step toward having out-of-the-box animations.
Having a new `MessageContainer` with `LazyItemScope` won't work because the caller expects that scope, as well as there are some parameters that use that scope.
If you prefer, I can not introduce that scope for now, but soon there will be a bigger breaking change affecting many more parameters when we expose that scope.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
andremion: Good catch!
I made a mistake.
I will fix it in a minute.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
aleksandar-apostolov: We can never know how the components are used. However in this case I think the possibility is low that anyone uses the `MessageContainer` directly and not in a `LazyXYZ` component.

aleksandar-apostolov (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
andremion: We cannot have a `MessageContentFactory` reference inside `ChatComponentFactory`, or get from `ChatTheme.messageContentFactory` inside each `ChatComponentFactory` function because customers could call a composable function and pass a custom `MessageContentFactory` directly, e.g.:
```kotlin
MessageItem(messageContentFactory = MySpecialFactory()...)
```

andremion: That's why we are checking whenever `messageContentFactory` is used:
```kotlin
if (messageContentFactory == MessageContentFactory.Deprecated) {
  // use the new component factory
} else {
  // use the legacy message content factory
}
```

PR #5597, opened 2025-01-29
[AND-252] Add factory methods for MessageComposer.

VelikovPetar:
### 🎯 Goal
Introduces methods in the `ChatComponentFactory` for customizing the `MessageComposer`.

### 🛠 Implementation details
Introduce factory methods for all externally configurable components of the `MessageComposer`:
- `headerContent`
- `footerContent`
- `mentionPopupContent`
- `commandPopupContent`
- `integrations`
- `label`
- `input`
- `audioRecordingContent`
- `trailingContent`

Additionally, introduce factories for some components used internally by the previously listed components, so that a more granular customisation is possible:
- `MessageInputOptions` <- used in `headerContent`
- `ComposerLinkPreview` <- used in `headerContent`
- `MentionSuggestionItem` and its contents (leading/center/trailing) <- used in `mentionsPopupContent`
- `CommandSuggestionItem` and its contents (leading/center) <- used in `commandsPopupContent`
- `ComposerAttachmentsButton` and `ComposerCommandsButton` <- used in `integrations`
- `ComposerQuotedMessage` <- used in `input`
- `ComposerCoolDownIndicator`, `ComposerSendButton` and `ComposerAudioRecordButton` <- used in `trailingContent`

### 🧪 Testing
1. Open compose sample
2. Override a new method in the `CustomChatComponentFactory`
3. The overriden component should be visible


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/composer/MessageInput.kt
```diff
@@ -64,7 +62,9 @@ public fun MessageInput(
     maxLines: Int = DefaultMessageInputMaxLines,
     keyboardOptions: KeyboardOptions = KeyboardOptions(capitalization = KeyboardCapitalization.Sentences),
     label: @Composable (MessageComposerState) -> Unit = {
-        DefaultComposerLabel(ownCapabilities = messageComposerState.ownCapabilities)
+        with(ChatTheme.componentFactory) {
+            MessageComposerLabel(it)
```
andremion, line 66: Using `with` is only required if you call a factory function with a receiver.
For example, if that function declaration was `fun ColumnScope.MessageComposerLabel` then you would have to use `with`.
Otherwise, you can just call `ChatTheme.componentFactory.MessageComposerLabel`

VelikovPetar: Thanks for the suggestion! I initially did this to keep the invocation consistent everywhere, but you are right, let's not complicate if not needed.
I updated the relevant invocations here: 0f417b178cb31963ec2d2154ca88fba6a2e4dc0f

andremion (approved):
Looks fine 👍🏻

PR #5598, opened 2025-01-29
[AND-282] Message footer status indicator factories

andremion:
### 🎯 Goal

Create factories for the message footer status indicator


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -877,4 +878,22 @@ public interface ChatComponentFactory {
     ) {
         MessageFooter(messageItem = messageItem)
     }
+
+    /**
+     * The default status indicator of the message footer, weather the message is sent, pending or read.
+     */
+    @Composable
+    public fun MessageFooterStatusIndicator(
```
VelikovPetar, line 886: Overriding this method would result also in overriding the whole content of the `MessageReadStatusIcon` - including the content shown when `syncStatus == SyncStatus.SYNC_NEEDED || syncStatus == SyncStatus.AWAITING_ATTACHMENTS` and the different content when `syncStatus == SyncStatus.COMPLETED` (part of the `MessageReadStatusIcon` implementation). Should we only provide a way to customise the `isMessageRead == true` case from the `MessageReadStatus`? (or provide a way to customise both parts - the parent `MessageReadStatusIcon` as it is right now, plus option to override only the 'read count' component)?

andremion: IMO, we shouldn't be too granular initially and introduce too many public factories that might be harder to maintain or to change in the future.
It's better to introduce some missing public factory later than introduce something beforehand and have to change it later.
Right now, the customer needs to have a clickable UI so they can implement that `MessageFooterStatusIndicator` function and call `MessageReadStatusIcon` wrapped in a `IconButton` or use the `modifier`.
WDYT?

andremion: What do you think about this?

```kotlin
@Composable
public fun MessageReadStatusIcon(
    message: Message,
    isMessageRead: Boolean,
    readCount: Int = 0,
    readCountContent: @Composable () -> Unit = { ReadCount(readCount = readCount) },
    isReadIcon: @Composable () -> Unit = { IsReadIcon() },
    isPendingIcon: @Composable () -> Unit = { IsPendingIcon(modifier = modifier) },
    isSentIcon: @Composable () -> Unit = { IsSentIcon(modifier = modifier) },
)
```

Keeping a single factory but allowing them to customize any single part

VelikovPetar: I am just worried that this will eventually be reported as a bug, if the customer overrides the whole `MessageReadStatusIcon`. 
I would be OK with such approach - maybe we should just keep the `readCountContent` + `isReadIcon` as a single group (as it is currently).

VelikovPetar (approved):

PR #5602, opened 2025-01-30
[AND-293] Add ChatComponentFactory methods for ThreadList components.

VelikovPetar:
### 🎯 Goal
Introduce factory methods in `ChatComponentFactory` for `ThreadList`.

### 🧪 Testing
1. Override any of the new factory methods in the `CustomChatComponentFactory`
2. Open compose sample app
3. Open the "Threads" tab
4. The changes made in the factory should be visible


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -896,4 +906,126 @@ public interface ChatComponentFactory {
...
+     *
+     * Used in the [ThreadListItem] to display the latest reply in the thread.
+     *
+     * @param thread The thread to display.
+     */
+    @Composable
+    public fun ThreadListItemLatestReplyContent(thread: Thread) {
+        thread.latestReplies.lastOrNull()?.let { reply ->
+            ThreadItemLatestReplyContent(reply)
+        }
```
andremion, line 1001: I would suggest keeping the `ChatComponetFactory` free from logic, as it will grow a lot.
Maybe that logic could be moved to the caller or to a `DefaultX` function if that makes sense.

VelikovPetar: Good point, I will update it!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/threads/ThreadItem.kt
```diff
@@ -194,10 +194,11 @@ internal fun RowScope.ThreadItemUnreadCountContent(unreadCount: Int) {
 /**
  * Default representation of the latest reply content in a thread.
  *
- * @param reply The latest reply [Message] in the thread.
+ * @param thread The thread to display.
  */
 @Composable
-internal fun ThreadItemLatestReplyContent(reply: Message) {
+internal fun ThreadItemLatestReplyContent(thread: Thread) {
+    val reply = thread.latestReplies.lastOrNull() ?: return
```
andremion, line 201: I remember I faced an issue years ago when using early returns in composable functions.

If `thread.latestReplies` changes after recomposition is triggered, Compose might not re-execute the UI code because the previous recomposition path was short-circuited with return.

What do you think about using a simple if (not null) check instead?

VelikovPetar: Updated: [c2a6eee](https://github.com/GetStream/stream-chat-android/pull/5602/commits/c2a6eeeed389f45e2fc61a29d992fdaa2e8c4ac6)

andremion (approved):
Looks good for me 🎉

PR #5605, opened 2025-02-03
[AND-271] Selected menu component factories

andremion:
### 🎯 Goal

Component factories for selected menus of channels, messages, and reactions


aleksandar-apostolov:
Copied the code to another [PR](https://github.com/GetStream/stream-chat-android/pull/5608)

PR #5606, opened 2025-02-03
[AND-268] Convert Shot tests to Paparazzi tests

andremion:
### 🎯 Goal

Run snapshot tests faster

### 🛠 Implementation details

- Remove **Shot** tests in favor of **Paparazzi** tests
- Convert message list header tests
- Convert message composer tests
- Convert selected channel menu tests
- Convert selected message menu tests
- Convert selected reactions menu tests

### 🎉 GIF

gif


stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/channels/SelectedChannelMenuTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    @get:Rule
+    override val paparazzi = Paparazzi(deviceConfig = DeviceConfig.PIXEL_2)
+
+    @Test
+    fun `selected channel`() {
+        snapshot {
+            Box(modifier = Modifier.fillMaxSize()) {
+                SelectedChannelMenu(
+                    modifier = Modifier.align(Alignment.BottomCenter),
+                    selectedChannel = PreviewChannelData.channelWithManyMembers.copy(
```
VelikovPetar, line 45: Looks good to me, I just noticed that the data source is now changed from the `TestData` object to the `Preview*Data` objects: I was wondering if this was intentional, because if someone changes the Preview data (for the purpose of updating a preview), they could potentially(unintentionally) break the tests?

andremion: The test was entirely moved and converted from the `uitest` module to the test source of the compose library to use `Paparazzi` instead of `Shot`.
`TestData` is the one used in the old UI tests.
Preview data has been used in Paparazzi tests since the Poll snapshot tests were introduced.
I don't think it's a problem using them.
Most of the snapshot tests would be very similar to what we have in previews.
If any change breaks the tests, we have the opportunity to fix it as soon as it happens at the PR level.

VelikovPetar: Perfect, thanks for the explanation!
Then I am perfectly OK with keeping it as it is!

VelikovPetar (approved):

PR #5607, opened 2025-02-03
[AND-287] Add component factory for AttachmentsPicker send button.

VelikovPetar:
### 🎯 Goal
Introduce a factory method in `ChatComponentFactory` for the `Send` button in the `AttachmentsPicker`.

_Note: We don't expose other factory methods for the `AttachmentsPicker`, as it can already be customised via the `AttachmentsPickerTabFactory`._


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1316,4 +1317,22 @@ public interface ChatComponentFactory {
...
+
+    /**
+     * The default 'Send' button in the attachments picker.
+     * Shown as ">" icon in the attachments picker header, enabled when there is at least one selected attachment.
+     *
+     * @param hasPickedAttachments If there are any attachments picked.
+     * @param onClick The click handler for the button.
+     */
+    @Composable
+    public fun AttachmentsPickerSendButton(
```
andremion, line 1329: nit: What about naming it as `MessageAttachmentsPickerSendButton`?

andremion (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentsPicker.kt
```diff
@@ -227,26 +227,41 @@ private fun AttachmentPickerOptions(
...
 }
+
+/**
+ * The default "Send" button in the attachments picker heading.
+ *
+ * @param hasPickedAttachments Indicator if there are selected attachments.
+ * @param onClick The action to be taken when the button is clicked.
+ */
+@Composable
+internal fun DefaultAttachmentsPickerSendButton(
```
andremion, line 244: I'm afraid we don't have snapshot tests for the attachment picker 🤔

VelikovPetar: At the moment it's a bit difficult to snapshot test the whole `AttachmentsPicker` composable, as it is quite tightly coupled to the `AttachmentsPickerViewModel`- we will need to introduce a stateless component, but in this case it will be a bit more difficult as there is quite a lot of logic to re-work. 

I could instead add a snapshot test for the `AttachmentPickerOptions` which is the place where the `DefaultAttachmentsPickerSendButton` is actually used in order to finalise this PR today, and afterwards re-work and add tests for the whole `AttachmentsPicker`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
VelikovPetar: Are we gaining anything by adding the `Message-` prefix? I think that `AttachmentsPicker-` is pretty specific to what it represents, and I believe that by adding the `Message-` prefix we will clutter the naming a bit.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentsPicker.kt
andremion: Fine for me to work on that later
Thanks 🙌🏻

VelikovPetar: Added tests here: https://github.com/GetStream/stream-chat-android/pull/5617

PR #5608, opened 2025-02-03
[AND-302] menu item stateless components

aleksandar-apostolov:
### 🎯 Goal

Add components into the component factory related to the menu items and channel menu, messages menu and reactions menu and reaction picker.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1316,4 +1356,596 @@ public interface ChatComponentFactory {
...
+        messageOptions: List<MessageOptionItemState>,
+        onMessageAction: (MessageAction) -> Unit,
+        ownCapabilities: Set<String>,
+    ) {
+        MessageMenuOptions(
+            modifier = modifier,
+            message = message,
+            options = messageOptions,
+            onMessageOptionSelected = { onMessageAction(it.action) },
+        )
```
andremion, line 1543: Without exposing the `ColumnScope`, how can the customer do this when overriding that factory? 🤔 
```kotlin
MessageMenuOptions(
    modifier = modifier.weight(1f),
    message = message,
    options = messageOptions,
    onMessageOptionSelected = { onMessageAction(it.action) },
)
```

aleksandar-apostolov: `modifier.fillMaxWidth()` ?

By exposing a `ColumnScope` here we are exposing and depending on the internal implementation of the calling composable. So if we in the future change to a different implementation then its a breaking change to change the scope.

aleksandar-apostolov: Imagine this scenario:

```
@Composable
fun MyComponent(....) {
   
   Column {
     // this is ColumnScope
     factory.MessageMenuOptions(....)
   }

}
```

Now we decide that we want to do the following:

```
@Composable
fun MyComponent(...) {

     LazyColumn {
        items {
             // I cannot do this without changing the implementation of the factory to have a `LazyItemScope` receiver
             // thus introducing a breaking change.
             factory.MessageMenuOptions(...)
        }
     }
   
}
```

Do you not see this as a concern?

andremion: I do see that concern. But in my opinion, we should not limit customers because of that.
For example, `ChannelListHeader` has `leadingContent` as `@Composable RowScope.() -> Unit` for years.
How many times did we have to change it? 🤷🏻

aleksandar-apostolov: Its a good point, but `ChatComponentsFactory` is new code. I think we can all agree that adding the `ColumnScope` is a limitation from the SDK perspective.

Do we by providing the `ColumnScope` increase the extensibility of the component or not? If yes, we can add scopes, if no we should revisit the idea to have the scopes regardless if the older implementation has them.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messageoptions/MessageOptionItem.kt
```diff
@@ -50,28 +48,27 @@ public fun MessageOptionItem(
     modifier: Modifier = Modifier,
     verticalAlignment: Alignment.Vertical = Alignment.CenterVertically,
     horizontalArrangement: Arrangement.Horizontal = Arrangement.Start,
+    onMessageOptionItemSelected: (MessageOptionItemState) -> Unit = {},
```
VelikovPetar, line 51: Should we maybe deprecate the `MessageOptionItem` as it is no longer used? (also to promote usage of the `GenericMenuOptionItem`)?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: In the case of `ColumnScope`, it contains four functions that it won't be possible to use without exposing the scope: 
```kotlin
fun Modifier.weight(@FloatRange(from = 0.0, fromInclusive = false) weight: Float, fill: Boolean = true): Modifier
fun Modifier.align(alignment: Alignment.Horizontal): Modifier
fun Modifier.alignBy(alignmentLine: VerticalAlignmentLine): Modifier
fun Modifier.alignBy(alignmentLineBlock: (Measured) -> Int): Modifier
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/selectedmessage/SelectedMessageMenu.kt
```diff
@@ -75,23 +69,24 @@ public fun SelectedMessageMenu(
...
-
-        if (ChatTheme.reactionOptionsTheme.areReactionOptionsVisible && canLeaveReaction) {
-            DefaultSelectedMessageReactionOptions(
-                message = message,
-                reactionTypes = reactionTypes,
-                showMoreReactionsDrawableRes = showMoreReactionsIcon,
-                onMessageAction = onMessageAction,
-                showMoreReactionsIcon = onShowMoreReactionsSelected,
-            )
+        with(ChatTheme.componentFactory) {
```
VelikovPetar, line 72: In the `SelectedMessageMenu` the `reactionTypes` and `showMoreReationsIcon` args are no longer used: Should we maybe add that to the Docs of the class? (or maybe deprecate the current `SelectedMessageMenu` to be replaced with a different implementation in the future)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messageoptions/MessageOptionItem.kt
aleksandar-apostolov: Lets deprecate this.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionoptions/ExtendedReactionsOptions.kt
```diff
@@ -60,10 +56,13 @@ public fun ExtendedReactionsOptions(
     cells: GridCells = GridCells.Fixed(DefaultNumberOfColumns),
     reactionTypes: Map<String, ReactionIcon> = ChatTheme.reactionIconFactory.createReactionIcons(),
     itemContent: @Composable LazyGridScope.(ReactionOptionItemState) -> Unit = { option ->
-        DefaultExtendedReactionsItemContent(
-            option = option,
-            onReactionOptionSelected = onReactionOptionSelected,
-        )
+        with(ChatTheme.componentFactory) {
+            ExtendedReactionMenuOptionItem(
+                modifier = Modifier.padding(vertical = 8.dp),
```
andremion, line 61: I can see the padding being applied here and in the default factory implementation. Can we move the padding from the factory to here?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionoptions/ReactionOptions.kt
```diff
@@ -91,55 +96,19 @@ public fun ReactionOptions(
...
 }
 
-/**
- * The default reaction option item.
- *
- * @param option The represented option.
- * @param onReactionOptionSelected The handler when the option is selected.
- */
-@Composable
-internal fun DefaultReactionOptionItem(
```
andremion, line 115: I preferred when we had default internal functions that set up everything instead of spreading the setup between the consumer and the factory.
Now, if we have to adjust some UI, we have to update one file and switch to another file to see the preview. Besides increasing even more, the factory file 🫤

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionoptions/ReactionOptions.kt
```diff
@@ -91,55 +96,19 @@ public fun ReactionOptions(
...
-                modifier = Modifier.clickable(
-                    interactionSource = remember { MutableInteractionSource() },
-                    indication = ripple(bounded = false),
-                    onClick = { onShowMoreReactionsSelected() },
-                ),
-                painter = painterResource(id = showMoreReactionsIcon),
-                contentDescription = LocalContext.current.getString(R.string.stream_compose_show_more_reactions),
-                tint = ChatTheme.colors.textLowEmphasis,
-            )
+            showMore(onShowMoreReactionsSelected, showMoreReactionsIcon)
```
andremion, line 99: Nice 👌🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1441,4 +1485,607 @@ public interface ChatComponentFactory {
...
+     * Factory method for creating a generic menu option item.
+     *
+     * @param modifier The modifier for the menu item.
+     * @param title The title of the menu item.
+     * @param titleColor The color of the title.
+     * @param leadingIcon The leading icon of the menu item.
+     * @param onClick Callback for when the menu item is clicked.
+     */
+    @Composable
+    public fun MenuOptionItem(
```
andremion, line 2068: I think this factory is not needed as it's only called from two other factories here and they can call `GenericMenuOptionItem` directly.

aleksandar-apostolov: Yes, but this allows for overriding both menu items at the same time.

aleksandar-apostolov: Or rather all menu items should default to this implementation and not `GenericMenuOptionItem`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/common/GenericMenuOptionItem.kt
```diff
@@ -0,0 +1,84 @@
...
+ * @param leadingIcon The icon to show on the left side of the item.
+ * @param title The title of the item.
+ * @param titleColor The color of the title.
+ * @param style The style of the title.
+ * @param itemHeight The height of the item.
+ * @param verticalAlignment Used to apply vertical alignment.
+ * @param horizontalArrangement Used to apply horizontal arrangement.
+ */
+@Composable
+public fun GenericMenuOptionItem(
```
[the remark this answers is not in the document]
aleksandar-apostolov: But then it conflicts with the `MenuOptionItem` in the factory.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1441,4 +1485,607 @@ public interface ChatComponentFactory {
...
+     *
+     * @param modifier The modifier for the footer.
+     */
+    @Composable
+    public fun ChannelOptionsItem(
+        modifier: Modifier,
+        option: ChannelOptionState,
+        onClick: () -> Unit,
+    ) {
+        MenuOptionItem(
```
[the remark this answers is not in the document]
aleksandar-apostolov: No, because then if I want to e.g. exclude all the icons from my menus I have to override more than one factory method.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/info/SelectedChannelMenu.kt
```diff
@@ -79,16 +78,22 @@ public fun SelectedChannelMenu(
     shape: Shape = ChatTheme.shapes.bottomSheet,
     overlayColor: Color = ChatTheme.colors.overlay,
     headerContent: @Composable ColumnScope.() -> Unit = {
-        DefaultSelectedChannelMenuHeaderContent(
-            selectedChannel = selectedChannel,
-            currentUser = currentUser,
-        )
+        with(ChatTheme.componentFactory) {
+            ChannelMenuHeaderContent(
+                modifier = Modifier,
```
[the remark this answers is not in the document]
aleksandar-apostolov: we have to supply it here as we cannot add add a default value in the factory.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/info/SelectedChannelMenu.kt
```diff
@@ -79,16 +78,22 @@ public fun SelectedChannelMenu(
...
+        }
     },
     centerContent: @Composable ColumnScope.() -> Unit = {
-        DefaultSelectedChannelMenuCenterContent(
-            onChannelOptionClick = onChannelOptionClick,
-            channelOptions = channelOptions,
-        )
+        with(ChatTheme.componentFactory) {
+            ChannelMenuCenterContent(
+                modifier = Modifier,
```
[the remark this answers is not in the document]
aleksandar-apostolov: see above

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/channels/ChannelOptions.kt
```diff
@@ -81,21 +78,13 @@ public fun ChannelOptions(
...
-                            .size(56.dp)
-                            .padding(16.dp),
-                        painter = option.iconPainter,
-                        tint = option.iconColor,
-                        contentDescription = null,
-                    )
-                },
-                onClick = { onChannelOptionClick(option.action) },
-            )
+            with(ChatTheme.componentFactory) {
```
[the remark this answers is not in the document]
aleksandar-apostolov: I think having the same code everywhere like in the `XYZContent` parameters is a lot easier even thou not really necessary.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/info/SelectedChannelMenu.kt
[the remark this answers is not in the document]
andremion: I mean, we don't need to have that modifier parameter in the factory at all
```kotlin
   @Composable
    public fun ChannelMenuHeaderContent(
        modifier: Modifier, // <!-- This is not needed
        selectedChannel: Channel,
        currentUser: User?,
    ) {
        DefaultSelectedChannelMenuHeaderContent(
            selectedChannel = selectedChannel,
            currentUser = currentUser,
        )
    }
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionoptions/ReactionOptions.kt
aleksandar-apostolov: Most of them are internal and at some point we are going to get asked to either expose them or remove them. I think we should remove all `internal` default implementation and have the implementation in the factory instead.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/common/GenericMenuOptionItem.kt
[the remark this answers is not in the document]
andremion: We already have other examples where we had to use the full package name to avoid conflicts

Screenshot 2025-02-04 at 15 50 48

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionoptions/ExtendedReactionsOptions.kt
aleksandar-apostolov: Nice catch :)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionoptions/ReactionOptions.kt
andremion: Oh my! I cannot see that factory scaling well 😅

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/common/GenericMenuOptionItem.kt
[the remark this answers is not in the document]
aleksandar-apostolov: Having to use your own package name on your own API is a sign of bad design :).
But since we've already done it so be it.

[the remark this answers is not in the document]
andremion: I don't like it either.
It was needed after we had removed some internal `DefaultX` functions and started to call public composables directly.
I preferred to keep using the internal `DefaultX` functions

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/info/SelectedChannelMenu.kt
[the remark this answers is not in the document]
aleksandar-apostolov: Only because we haven't deleted (yet) the default method which is internal.

aleksandar-apostolov:
Merge only after snapshot tests are done for the `ExtendedReactionPickerOptions`

PR #5611, opened 2025-02-04
[AND-305] Add snapshot tests for PinnedMessageList and related components.

VelikovPetar:
🎯 Goal

Add UI(snapshot) tests for `PinnedMessageList` and the related components.

🛠 Implementation details
- Introduce a stateless (VM-independent) `PinnedMessageList` to be tested
- Add snapshot tests for `PinnedMessageList` and `PinnedMessageItem`
- Add `PreviewPinnedMessageData` to be used for previews/tests


stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/pinned/PinnedMessageItemKtTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import app.cash.paparazzi.DeviceConfig
+import app.cash.paparazzi.Paparazzi
+import io.getstream.chat.android.compose.ui.SnapshotTest
+import io.getstream.chat.android.previewdata.PreviewPinnedMessageData
+import io.getstream.chat.android.previewdata.PreviewUserData
+import org.junit.Rule
+import org.junit.Test
+
+internal class PinnedMessageItemKtTest : SnapshotTest {
```
andremion, line 27: Is there any special case why we have that `Kt` in the name of the class?
Same for `PinnedMessageListKtTest` 🤔

andremion (commented):
Overall looks good! 👏🏻 
There's just one finding about the `Kt` in the name of some classes/files

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/pinned/PinnedMessageItemKtTest.kt
VelikovPetar: Ah it was autogenerated and I forgot to remove it - will change it right away.

aleksandar-apostolov (approved):

PR #5616, opened 2025-02-05
[AND-307] Expose `UserId` alias to avoid InternalStreamChatApi issues

JcMinarro:
### 🎯 Goal
The public `Channel.currentUserUnreadCount()` extension function is using `UserId` alias as argument, what make this function unavailable to anyone that is not opt-in for `@InternalStreamChatApi`.
Exposing `UserId` alias avoid @InternalStreamChatApi issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@
 
 ### ❌ Removed
 
+## stream-chat-android-core
+### ✅ Added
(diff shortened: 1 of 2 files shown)


aleksandar-apostolov (approved):

irsyad-dev (commented):
CHANGELOG.md

PR #5625, opened 2025-02-10
[AND-313] Fix the Avatar loading

andremion:
### 🎯 Goal

Increase the stability of fading/loading image of the user avatars.

### 🛠 Implementation details

It seems the `CrossfadePlugin` is causing this instability. After spending some time debugging, the best option might be to stop using `StreamImage`, handle the painter states, and use `ImageAvatar` or `InitialsAvatar` according to the states. 

A positive side-effect of that refactoring is that now we don't need to use a preview avatar anymore. Hence, we can update the snapshots to have the user's initials instead of the preview avatar.

Another positive side-effect is that we could identify a bug in the avatars of Poll components, where the user's initials were not fully displayed due to the font size.

Also, another fix was on the click indicator of user avatars, as now we can see a proper ripple effect.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/b647b7c1-1eb0-4fe6-a2c0-12469414ec2a" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/794fd3da-7c69-4fed-93c1-6c9bd99202a1" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

#### `Avatar` click ripple fix

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/05bb7eae-55b1-4313-a61a-0e326a28c0bd" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/945ff396-8c33-448d-a96b-85e45269bfca" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### Poll `InitialsAvatar` fix

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/a6549b60-fbfb-46e7-a841-d8a298e77490" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/b4ac6c21-0f2b-4b62-ac4e-a1b62048a473" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/Avatar.kt
```diff
@@ -64,60 +68,49 @@ public fun Avatar(
...
-            onClick = onClick,
-            avatarOffset = initialsAvatarOffset,
+    var imageSize by remember { mutableStateOf(Size.ORIGINAL) }
+    Box(
+        modifier = modifier
+            .testTag("Stream_QuotedMessageAuthorAvatar") // This should not be here. Tied to the quoted component.
+            .onSizeChanged { size -> imageSize = Size(size.width, size.height) },
+    ) {
+        val cdnImageResizing = ChatTheme.streamCdnImageResizing
+        val data = remember { imageUrl.applyStreamCdnImageResizingIfEnabled(cdnImageResizing) }
```
JcMinarro, line 78: What about if the `imageUrl` needs to be sign by our customers implementation?
This URL won't be signed and no picture will be shown at all

skydoves (approved):
Looks good to me! Out of my curiosity, what if we wrap the `StreamImage` using `Crossfade` without using `CrossfadePlugin`? If this makes any problems, this issue should be fixed on the library site.

andremion:
> Looks good to me! Out of my curiosity, what if we wrap the `StreamImage` using `Crossfade` without using `CrossfadePlugin`? If this makes any problems, this issue should be fixed on the library site.

I believe `Crossfade` is designed to switch (based on states) between layouts with a crossfade animation. It shouldn't work to wrap a single child. But I gave a try anyway.

Screenshot 2025-02-14 at 08 25 10

[Screen_recording_20250214_082546.webm](https://github.com/user-attachments/assets/e1163a7d-5434-4f4c-ad53-b2d7e907265a)

JcMinarro (approved):

PR #5638, opened 2025-02-18
[CI] Add Authentication E2E Tests

testableapple:
Resolve https://linear.app/stream/issue/AND-220

### 🎯 Goal

Copy-paste all the `Authentication` related E2E tests from iOS to Android

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

gif


stream-chat-android-compose-sample/src/e2e/java/io/getstream/chat/android/compose/sample/ui/InitTestActivity.kt
```diff
@@ -0,0 +1,40 @@
...
+    data object UserLogin : InitTestActivity() {
+        private fun readResolve(): Any = UserLogin
+        override fun createIntent(context: Context): Intent =
+            UserLoginActivity.createIntent(context)
+    }
+
+    data class Jwt(val baseUrl: String) : InitTestActivity() {
+        override fun createIntent(context: Context): Intent {
+            val intent = JwtTestActivity.createIntent(context)
+            intent.putExtra("BASE_URL", baseUrl)
```
JcMinarro, line 36: As `BASE_URL` is a mandatory extra to this activity be working, the `createIntent` should require this argument.
Then, you can move the `putExtra(...)` call within the `createIntent()` method.

.github/workflows/e2e-test.yml
```diff
@@ -61,6 +61,8 @@ jobs:
           arch : x86_64
           emulator-options: ${{ vars.EMULATOR_OPTIONS }}
           script: bundle exec fastlane run_e2e_test batch:${{ matrix.batch }} batch_count:${{ strategy.job-total }}
+        env:
+          STREAM_DEMO_APP_SECRET: ${{ secrets.STREAM_DEMO_APP_SECRET }}
```
JcMinarro, line 65: What is this secret used for?
Do we have it stored somewhere else? (eg: 1Password)

stream-chat-android-compose-sample/src/e2e/java/io/getstream/chat/android/compose/sample/ui/InitTestActivity.kt
testableapple: Updated 👍

.github/workflows/e2e-test.yml
testableapple: Added a comment regarding this at the top of the `AuthTests.kt` 👍

stream-chat-android-compose-sample/src/e2e/java/io/getstream/chat/android/compose/sample/ui/JwtTestActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+    private suspend fun fetchJwtToken(baseUrl: String, userId: String): String {
+        return withContext(Dispatchers.IO) {
+            val endpoint = "$baseUrl/jwt/get?api_key=$API_KEY&user_id=$userId"
+            val request = Request.Builder().url(endpoint).build()
+
+            OkHttpClient().newCall(request).execute().use { response ->
+                if (!response.isSuccessful) {
+                    throw Exception("Failed to fetch token: ${response.code}")
+                }
+                response.body?.string() ?: throw Exception("Empty response body")
```
JcMinarro, line 194: Why do we need to do a network call to obtain a token?
Can't we keep them here?
Maybe adding some logic if we need to return different ones (Valid, Expired, for different users...)
We should avoid as much as possible external factor that could break tests

testableapple: Isn't it how the SDK works in general? It does a network call to the backend to get a token. Here we just do a network call to a mock server that will generate the token as the real backend would do.

JcMinarro: The SDK itself doesn't do any call to obtain a token.
It is implemented on our customers app, but it is part of their implementation.
BTW, It is a test scenario, we should have those scenario's input handled on the test itself, don't see why we need to involve the real server on that.

If the test need to "request" a token (Event multiple tokens), they can be provided on the intent that start this activity, then configure the test according with it.

testableapple: Discussed on Slack ✅

JcMinarro (approved):

PR #5644, opened 2025-02-20
[AND-329] Add sending messages to only specific members

JcMinarro:
### 🎯 Goal
Add sending messages to only specific members


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/MessageDtos.kt
```diff
@@ -47,7 +47,7 @@ internal data class UpstreamMessageDto(
     val silent: Boolean,
     val text: String,
     val thread_participants: List<UpstreamUserDto>,
-
+    val restricted_visibility: List<String>,
```
VelikovPetar, line 50: Question for my understanding: Are we not receiving this field in the `DownstreamMessageDto`? Is this completely handled on back-end side?

JcMinarro: Yep, this field is only handled backend-side

VelikovPetar (approved):

Issue #5624, opened 2025-02-09
Please update to Coil3


amal:
Please update Coil from 2 to Coil3.

Coil2 is an outdated version of Coil with no later support but compose artifact still uses it instead of the Coil3.
https://coil-kt.github.io/coil/upgrading_to_coil3/

JcMinarro:
Hello @amal 
Thanks for the purpose.
We are adding this migration to our roadmap

wasif1:
Hi, Can you assign it to me ? i'll update this.

JcMinarro:
Hello @wasif1 
Feel free to take it if you want to contribute 🙃

Jacks0N23:
@JcMinarro thank you for implementing support for coil3, it was my team request ❤ 

But I found out that [landscapist define wrong dependencies](https://github.com/skydoves/landscapist/issues/700) inside coil3 lib and now we're pulling both ktor and okhttp 😢 

Could you update after they'll fix it?

Also I want you to suggest to use [dependencyGuard](https://github.com/dropbox/dependency-guard). I found the problem using it report

skydoves:
Hey @Jacks0N23, thanks for raising this issue! That was actually a legacy setup from when Coil3 didn’t support the OkHttp fetcher. Great catch—this has now been addressed. As of version 2.5.0, the Ktor dependency has been removed.

Happy to see @JcMinarro here again! 😄 I fixed this issue from my end, please let me know if there's further issue.

PR #5645, opened 2025-02-21
[AND-330] Migrate to Coil3

JcMinarro:
### 🎯 Goal
Migrate to Coil3.
If you are using Coil within your app, you can follow this [migration guide](https://coil-kt.github.io/coil/upgrading_to_coil3/) they published

Close: #5624


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ImageRequestTimeoutHandler.kt
```diff
@@ -1,50 +0,0 @@
...
-/**
- * Triggers [onTimeout] when the connection is established and the image load has failed due to a timeout.
- *
- * @param data The data to load.
- * @param connectionState The state of the network connection.
- * @param imageState The state of the async image painter.
- * @param onTimeout The lambda function called when the timeout conditions have been met.
- */
-@Composable
-internal fun ImageRequestTimeoutHandler(
```
JcMinarro, line 34: It was a workaround added to be able to add a kind of retry policy when fetching pictures and there is a timeout error.  [Ref Here](https://github.com/coil-kt/coil/issues/884#issuecomment-975932886)

On Coil3 a [new restart()](https://github.com/coil-kt/coil/pull/2383) method has been added that allow us to retry this query. And we are retry those case within our [fetching image wrapper](https://github.com/GetStream/stream-chat-android/pull/5645/files#diff-62682457fe2443eb1755b9ddd35c79a3d4216919e763fb73d6be8adf9364148eR169)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/StreamCoilImageLoaderFactory.kt
```diff
@@ -78,20 +78,19 @@ internal object DefaultStreamCoilImageLoaderFactory : StreamCoilImageLoaderFacto
...
     /**
      * Uses Android [Context] to build a new [StreamImageLoaderFactory].
      *
-     * @param context - The context to load the factory with.
      * @return [ImageLoaderFactory] that loads all the images in the app.
      */
-    private fun newImageLoaderFactory(context: Context): ImageLoaderFactory {
-        return StreamImageLoaderFactory(context).apply {
+    private fun newImageLoaderFactory(): SingletonImageLoader.Factory {
+        return StreamImageLoaderFactory().apply {
```
JcMinarro, line 93: `Context` is not needed anymore to create an `ImageLoaderFactory`

andremion (commented):
🚀 🚀 🚀

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ImageUtils.kt
```diff
@@ -161,7 +165,11 @@ internal fun StreamAsyncImage(
                 imageLoader = LocalStreamImageLoader.current,
                 contentScale = contentScale,
             )
-            content(asyncImagePainter.state)
+            val state: AsyncImagePainter.State by asyncImagePainter.state.collectAsState()
+            if ((state as? AsyncImagePainter.State.Error)?.result?.throwable is SocketTimeoutException) {
+                asyncImagePainter.restart()
+            }
```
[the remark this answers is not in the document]
JcMinarro: :thinking: 
Good point....
In the previous implementation we didn't have any kind of limit, this error is only when there is a timeout, not on all "network error", but maybe we can add some limit (eg: retry 3 times). Let's me add it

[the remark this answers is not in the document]
andremion: Good catch @VelikovPetar  👍🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ImageUtils.kt
```diff
@@ -153,6 +160,7 @@ internal fun StreamAsyncImage(
             val context = LocalContext.current
             val imageAssetTransformer = ChatTheme.streamImageAssetTransformer
             val imageHeaderProvider = ChatTheme.streamImageHeadersProvider
+            val fetchRetries = remember { mutableIntStateOf(0) }
```
andremion, line 163: nit: we could remove the need to use `intValue`  if using delegation when declaring `fetchRetries`:

```kotlin
var fetchRetries by remember { mutableIntStateOf(0) }
```

That way, you could just do `if (fetchRetries++ < MaxRetries) {`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ImageUtils.kt
```diff
@@ -161,7 +169,15 @@ internal fun StreamAsyncImage(
                 imageLoader = LocalStreamImageLoader.current,
                 contentScale = contentScale,
             )
-            content(asyncImagePainter.state)
+            val state: AsyncImagePainter.State by asyncImagePainter.state.collectAsState()
```
andremion, line 172: nit: declaring the type just seems to create noise.
`val state by asyncImagePainter.state.collectAsState()` should be enough.

andremion (approved):
Great job! 🎉 
I just sent a couple of optional suggestion to reduce the code noise.

PR #5646, opened 2025-02-21
[CI] Resolve some e2e tests issues

testableapple:
### 🎯 Goal

- Make e2e tests more stable on PRs and Nightly

### 🛠 Implementation details

- Rename logs archives so they are no rewritten.
- Remove API 29 from nightly checks, it was not stable. Having 28 as a min tier should be fine.
- Add timeout to Allure upload (noticed it can get stuck)
- Grant read/write permissions for older APIs
- Implement a `retryOnStaleObjectException` workaround to wait until the message list is unstale after scrolling
- Implement an option to provide a `delay` to participants actions, so we can test offline mode properly

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -114,11 +114,12 @@ lane :run_e2e_test do |options|
   UI.user_error!('Tests have failed!') if result.include?('Failures')
 end
 
+# This lane can be used for local linting
 lane :lint_e2e_test do |options|
   Dir.chdir('..') do
-    ['spotlessApply', 'apiDump', 'detekt'].each do |task|
+    ['spotlessApply', 'detekt'].each do |task|
       sh("./gradlew :stream-chat-android-e2e-test:#{task}")
-      sh("./gradlew :stream-chat-android-compose-sample:#{task}") unless task == 'apiDump'
+      sh("./gradlew :stream-chat-android-compose-sample:#{task}")
     end
   end
 end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -114,6 +114,15 @@ lane :run_e2e_test do |options|
   UI.user_error!('Tests have failed!') if result.include?('Failures')
 end
 
+lane :lint_e2e_test do |options|
+  Dir.chdir('..') do
+    ['spotlessApply', 'apiDump', 'detekt'].each do |task|
+      sh("./gradlew :stream-chat-android-e2e-test:#{task}")
+      sh("./gradlew :stream-chat-android-compose-sample:#{task}") unless task == 'apiDump'
+    end
+  end
+end
+
 lane :upload_attachments do
   ['png', 'pdf'].each do |ext|
     [1, 2].each { |i| sh("adb push attachments/file.#{ext} /sdcard/Download/file_#{i}.#{ext}") }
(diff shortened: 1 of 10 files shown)


fastlane/Fastfile
```diff
@@ -114,6 +114,15 @@ lane :run_e2e_test do |options|
...
 end
 
+lane :lint_e2e_test do |options|
+  Dir.chdir('..') do
+    ['spotlessApply', 'apiDump', 'detekt'].each do |task|
+      sh("./gradlew :stream-chat-android-e2e-test:#{task}")
+      sh("./gradlew :stream-chat-android-compose-sample:#{task}") unless task == 'apiDump'
+    end
+  end
+end
```
andremion, line 124: I'm wondering what that is used for.

`spotlessApply` and `detekt` are already called as part of the PR workflow.

Regarding `apiDump`, I would say that it should be kept as is, as a pre-push hook only.

stream-chat-android-e2e-test/api/stream-chat-android-e2e-test.api
```diff
@@ -6,6 +6,8 @@ public final class io/getstream/chat/android/compose/uiautomator/ActionsKt {
 	public static final fun goToForeground (Landroidx/test/uiautomator/UiDevice;)V
 	public static final fun longPress (Landroidx/test/uiautomator/UiObject2;I)V
 	public static synthetic fun longPress$default (Landroidx/test/uiautomator/UiObject2;IILjava/lang/Object;)V
+	public static final fun retryOnStaleObjectException (Landroidx/test/uiautomator/UiDevice;ILkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public static synthetic fun retryOnStaleObjectException$default (Landroidx/test/uiautomator/UiDevice;ILkotlin/jvm/functions/Function0;ILjava/lang/Object;)Ljava/lang/Object;
```
andremion, line 10: To be honest, I don't why we have API generation for the e2e-test module 🤔

fastlane/Fastfile
testableapple: It's not used anywhere. Tbh I just added it for my local usage because it was quite painful to get the PR blocked because of lint issues and then trying to recall all these commands 🙂

stream-chat-android-e2e-test/api/stream-chat-android-e2e-test.api
JcMinarro: I think it is there because it is is configured globally on the project build.gradle, but it shouildn't be needed for this module as far as it is only used for testing purpose.
This module can be added on the ignore section

JcMinarro: FYI: https://github.com/GetStream/stream-chat-android/pull/5648

fastlane/Fastfile
andremion: In that case, shouldn't `compose-sample` and `apiDump` be removed from there?

tip: a single call with both `spotlessApply` and `detekt` is more performant than two calls, for example:

`./gradlew :stream-chat-android-e2e-test:spotlessApply :stream-chat-android-e2e-test:detekt`

testableapple: Yeah, as Jc opened a PR regarding `apiDump` I already removed it from this PR.

The tests are part of `compose-sample` so I can't remove it.

`./gradlew :stream-chat-android-e2e-test:spotlessApply :stream-chat-android-e2e-test:detekt` updated 👍

andremion (approved):

PR #5655, opened 2025-02-26
Custom TokenProvider when PN is received

JcMinarro:
### 🎯 Goal
Complete: [AND-341](https://linear.app/stream/issue/AND-341)
Add a way to inject a TokenProvider from PN on the case an expired token needs to be refreshed

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationConfig.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationConfig.kt
@@ -17,6 +17,7 @@
 package io.getstream.chat.android.client.notifications.handler
 
 import io.getstream.android.push.PushDeviceGenerator
+import io.getstream.chat.android.client.token.TokenProvider
 
 /**
  * Push notifications configuration class
@@ -55,4 +56,11 @@ public data class NotificationConfig @JvmOverloads constructor(
      * Whether or not the auto-translation feature is enabled.
      */
     val autoTranslationEnabled: Boolean = false,
+
+    /**
+     * A token provider to be used on case of restoring user credentials and an expired token needs to be refreshed.
+     * If not provided, the SDK will create a [TokenProvider] with the last token used within the client.
+     */
+    val tokenProvider: TokenProvider? = null,
+
 )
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/attachment/worker/UploadAttachmentsWorker.kt
```diff
@@ -77,7 +77,7 @@ public class UploadAttachmentsWorker(
                 return Result.Failure(Error.GenericError("Could not set user"))
             }
 
-            chatClient.setUserWithoutConnectingIfNeeded()
+            chatClient.setUserWithoutConnectingIfNeeded(null)
```
VelikovPetar, line 80: Thinking out loud here: Should we also consider this case, when this worker would be started when the user is disconnected, and the token is expired? 
It is an extreme edge case, but perhaps we should also use the new `tokenProvider`?

```kotlin
val tokenProvider = chatClient.config.notificationConfig.tokenProvider
chatClient.setUserWithoutConnectingIfNeeded(tokenProvider)
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationConfig.kt
```diff
@@ -55,4 +56,11 @@ public data class NotificationConfig @JvmOverloads constructor(
      * Whether or not the auto-translation feature is enabled.
      */
     val autoTranslationEnabled: Boolean = false,
+
+    /**
+     * A token provider to be used on case of restoring user credentials and an expired token needs to be refreshed.
+     * If not provided, the SDK will create a [TokenProvider] with the last token used within the client.
+     */
+    val tokenProvider: TokenProvider? = null,
```
VelikovPetar, line 64: Not really a remark, but more of a discussion point: 
To refer the previous comment on the `UploadAttachmentsWorker`, if we need to cover that case as well, I am not sure if we should tie the new `TokenProvider` to the `NotificationConfig`, as it could be used in a context outside of the notifications.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/attachment/worker/UploadAttachmentsWorker.kt
JcMinarro: On the case of "disconnect/reconnect", the `ChatClient` instance already have a valid `tokenProvider`, because the `setUser()` method has been called.
BTW. It is giving me an idea, we don't really need to pass the `tokenProvider` from outside of the `ChatClient`, but use it internally, because it is already on the `ChatClient` itself

PR #5656, opened 2025-02-28
Avoid multiple calls to `TokenProvider.loadToken()` 

JcMinarro:
### 🎯 Goal
With the current implementation of our SDK the CacheableTokenProvider could be called multiple times to loadToken() while the token is being generated by our customers' implementation, what would cause unnecesary refresh token process and it could derive to infinite loop in the case of previous tokens are invalidated.

With the new implementation we avoid calling the `loadToken()` method multiple times while the previous calls are not being completed
Complete: [AND-346](https://linear.app/stream/issue/AND-346)

### 🎉 GIF
giphy (1)


VelikovPetar (approved):
Looks good to me!

PR #5659, opened 2025-03-03
[AND-340] Unify `MessageComposer` padding and set `targetSdk=35` in sample apps

VelikovPetar:
### 🎯 Goal
The default padding (`safeDrawingPadding`) applied to the stateful `MessageComposer(MessageComposerViewModel)` could add undesired padding when used in app which targets SDK 35 and is running on a device on Android 15 or above. This is because the `safeDrawingPadding` applies the `systemBars` padding (top and bottom) + `ime` padding (bottom). If a parent component already applies this padding and the insets are not consumed, this padding will be applied twice. 
Additionally, this padding is not applied to the stateless `MessageComposer(MessageComposerState)` components, so the behaviour between the stateful and the stateless types of the `MessageComposer` is different. 
With this PR, the `safeDrawingPadding` is removed to be unified with the stateless `MessageComposer`, and to prevent double appliance of the padding. (`MessagesScreen` will not be affected, as the root component there already applies the `safeDrawingPadding` in the parent component).

### 🛠 Implementation details
- Remove `safeDrawingPadding` from stateful `MessageComposer(MessageComposerViewModel)` component
- Remove unnecessary elevation in `AttachmentsPicker`
- Update `Configuration.sampleTargetSdk` to `35`
- Update sample apps to support `EdgeToEdge` for demo purposes

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Compose</th>
<th>XML</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/c8ff5b09-bbc8-47b7-bdfd-f0143c843b55" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/f6269ca3-f78e-42c8-a865-29281f3787cc" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Run compose/xml sample apps on device running Android 15
2. Enter a channel
3. Type a message
4. None of the screens/components should be cut of by the system insets (status bar, nav bar, ime...)


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/add/AddChannelScreen.kt
```diff
@@ -89,8 +89,7 @@ fun AddChannelScreen(
     Scaffold(
         modifier = Modifier
             .fillMaxSize()
-            .background(ChatTheme.colors.appBackground)
```
JcMinarro, line 92: Do we need to remove the background?

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/ChannelInfoActivity.kt
```diff
@@ -68,8 +68,7 @@ class ChannelInfoActivity : BaseConnectedActivity() {
          * @param channelId The ID of the channel for which the pinned messages are shown.
          */
         fun createIntent(context: Context, channelId: String) =
-            Intent(context, ChannelInfoActivity::class.java)
-                .putExtra(KEY_CHANNEL_ID, channelId)
+            Intent(context, ChannelInfoActivity::class.java).putExtra(KEY_CHANNEL_ID, channelId)
```
JcMinarro, line 71: Can we revert this change?
Having `extra` in a new line helps us with PR Conflicts in case new extras need to be added

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/add/AddChannelScreen.kt
VelikovPetar: It is actually not needed, because the components apply their own background, and it obstructs the root background. 
Example on Android 14/15:
| Android 14 | Android 15 |
| ----------- | ----------- |
| and14 | and15 |

JcMinarro (approved):

PR #5660, opened 2025-03-03
(Part 1) `Stream-chat-android-client` unit tests - `api` package

VelikovPetar:
### 🎯 Goal
Increase the unit test coverage of the project.
Cover the `api` package from the `stream-chat-android-client` package

### 🛠 Implementation details
Add tests (where applicable) for the `io.getstream.chat.android.client.api` package

### DIFF
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/interceptor/ApiRequestAnalyserInterceptorTest.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/interceptor/ApiRequestAnalyserInterceptorTest.kt
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2014-2025 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.api.interceptor
+
+import io.getstream.chat.android.client.api.FakeChain
+import io.getstream.chat.android.client.api.FakeResponse
+import io.getstream.chat.android.client.plugins.requests.ApiRequestsAnalyser
+import org.junit.Test
+import org.mockito.kotlin.any
+import org.mockito.kotlin.doNothing
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
+import org.mockito.kotlin.whenever
+
+internal class ApiRequestAnalyserInterceptorTest {
+
+    @Test
+    fun testApiRequestAnalyserInterceptorRegistersTheRequestInTheAnalyser() {
+        // given
+        val analyser = mock<ApiRequestsAnalyser>()
+        doNothing().whenever(analyser).registerRequest(any(), any())
+        val interceptor = ApiRequestAnalyserInterceptor(analyser)
+        // when
+        val chain = FakeChain(FakeResponse(200))
+        interceptor.intercept(chain)
+        // then
+        verify(analyser, times(1)).registerRequest(
+            requestName = chain.request().url.toString(),
+            data = mapOf("body" to "no_body"),
+        )
+    }
+}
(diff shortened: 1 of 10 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/interceptor/HeadersInterceptorTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        val interceptor = HeadersInterceptor(isAnonymous, headersUtil)
+        // when
+        val response = interceptor.intercept(FakeChain(FakeResponse(200)))
+        // then
+        response.request.header("User-Agent") `should be equal to` "userAgent"
+        response.request.header("Content-Type") `should be equal to` "application/json"
+        response.request.header("stream-auth-type") `should be equal to` "jwt"
+        response.request.header("X-Stream-Client") `should be equal to` "sdkTrackingHeaders"
+        response.request.header("Cache-Control") `should be equal to` "no-cache"
+    }
```
JcMinarro, line 64: Instead of writting the same test multiple times only because a single parameter is different, we can use `@ParameterizedTest` providing an input source with the different input/expected values.

On the other hand, consider to use `randomString()` when the value "is not important" and our code accepts whatever value, it could raise up hidden bubs in our code (hardcoded values, default behavior given a default value, ...)

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/interceptor/ApiRequestAnalyserInterceptorTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        val analyser = mock<ApiRequestsAnalyser>()
+        doNothing().whenever(analyser).registerRequest(any(), any())
+        val interceptor = ApiRequestAnalyserInterceptor(analyser)
+        // when
+        val chain = FakeChain(FakeResponse(200))
+        interceptor.intercept(chain)
+        // then
+        verify(analyser, times(1)).registerRequest(
+            requestName = chain.request().url.toString(),
+            data = mapOf("body" to "no_body"),
```
JcMinarro, line 44: Are we only testing when the response is empty?
Should we test other cases?
As commented on another comment, we can use `@ParameterizedTest` on that case.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/interceptor/ProgressInterceptorTest.kt
```diff
@@ -0,0 +1,51 @@
...
+        val progressCallback = object : ProgressCallback {
+            override fun onSuccess(url: String?) { /* No-Op */ }
+            override fun onError(error: Error) { /* No-Op */ }
+            override fun onProgress(bytesUploaded: Long, totalBytes: Long) { /* No-Op */ }
+        }
+        val request = Request.Builder()
+            .url("https://hello.url")
+            .post("body".toRequestBody())
+            .tag(ProgressCallback::class.java, progressCallback)
+            .build()
```
JcMinarro, line 44: What about creating a `Mother` method to easily create `Request` with custom value?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/internal/DistinctChatApiTest.kt
```diff
@@ -0,0 +1,585 @@
...
+import kotlinx.coroutines.test.TestScope
+import kotlinx.coroutines.test.runTest
+import org.junit.Assert
+import org.junit.Test
+import org.mockito.kotlin.any
+import org.mockito.kotlin.anyOrNull
+import org.mockito.kotlin.mock
+import org.mockito.kotlin.whenever
+
+internal class DistinctChatApiTest {
```
JcMinarro, line 36: Looks like we are only testing the case where arguments are the same

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/internal/DistinctChatApiTest.kt
```diff
@@ -0,0 +1,585 @@
...
+
+    @Test
+    fun `When calling queryChannel with same arguments, Then same instance of Call is returned`() {
+        // given
+        val distinctChatApi = DistinctChatApi(TestScope(), mock())
+        // when
+        val call1 = distinctChatApi.queryChannel("messaging", "1", QueryChannelRequest())
+        val call2 = distinctChatApi.queryChannel("messaging", "1", QueryChannelRequest())
```
JcMinarro, line 44: Instead of hardcoded values, we should use random ones.
On the other hand, `QueryChannelRequest` constructor is generating an "equal" instance, but we should verify it by initializing it with custom data, not only the default values.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/interceptor/HeadersInterceptorTest.kt
VelikovPetar: I didn't want to write a `@ParameterizedTest` for just 2 tests because I thought it would reduce the readability a bit. But I can change it anyway, since I will change some of the values with `randomString()`

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/internal/DistinctChatApiTest.kt
VelikovPetar: Each method is actually covered with 3 tests:
- same arguments, at the same time
- same arguments, called sequentially
- different arguments, at the same time

JcMinarro (approved):

PR #5661, opened 2025-03-04
[AND-359] Use a more descriptive token provider name

JcMinarro:
### 🎯 Goal
Use a more descriptive token provider name.

### 🎉 GIF

ezgif-379c8bc57d6a85


stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt
```diff
@@ -57,18 +56,13 @@ class ChatInitializer(
                 pushDeviceGenerators = listOf(
                     FirebasePushDeviceGenerator(
                         context = context,
-                        providerName = "Firebase",
-                    ),
-                    HuaweiPushDeviceGenerator(
```
VelikovPetar, line 62: Just for my understanding: Why are we removing the `huawei` push support from the sample?

VelikovPetar (approved):

PR #5668, opened 2025-03-07
[AND-381] Fix WS X-Stream-Client headers placement.

VelikovPetar:
### 🎯 Goal
Fix the placement of the `X-Stream-Client` value in the WebSocket request URL

### 🛠 Implementation details
Move the `X-Stream-Client` value from the `json` query parameter, to a dedicated query parameter in the WebSocket request URL.


JcMinarro (commented):
Looks Good.
@kanat could you confirm it is how Backend expects it?

kanat:
> Looks Good. @kanat could you confirm it is how Backend expects it?

Hey @JcMinarro ,

Yep, the backend takes it from either `Http Header` or `Query Param`.

kanat (approved):
LGTM

PR #5669, opened 2025-03-07
[AND-358] Use partialUpdateMessage when editing messages

JcMinarro:
### 🎯 Goal

Use partialUpdateMessage when editing messages


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -912,7 +912,13 @@ public class MessageComposerController(
      * @param message [Message]
      */
     private fun getEditMessageCall(message: Message): Call<Message> {
-        return chatClient.updateMessage(message)
+        return chatClient.partialUpdateMessage(
```
VelikovPetar, line 915: Maybe it is not something that we should consider, but wouldn't we lose some of the data from the message, if we only `set` the `text` and `attachments` fields? For example, if we mention a user when editing the message, the `mentionedUserIds` field will not be propagated to the `partialUpdateMessage` call.

JcMinarro: Good catch, let me add it

VelikovPetar: Looks good to me, I don't think that we are missing anything else here (commands seem not available for editing messages).

VelikovPetar (approved):

PR #5672, opened 2025-03-10
[AND-390] Revert "[PBE-5891] Remove `NotInFilterObject` (#5394)"

JcMinarro:
### 🎯 Goal
Revert changes where `NotInFilterObject` was removed.
We keep it as deprecated and will be removed on a future version.

This reverts commit 7bfe302fc1c1d273df263b3ca53efc7db8501b4c.


VelikovPetar (approved):

shubhamsinghshubham777:
@JcMinarro What's the alternative for this now? For example, how to query channels where a user id is NOT a member of?

PR #5683, opened 2025-03-18
[AND-403] Fix "Thread reply" enabled while in thread.

VelikovPetar:
### 🎯 Goal
The "Thread reply" item was wrongly shown as a message option when opening the menu inside a thread.

### 🛠 Implementation details
- Add the missing check for `isInThread` in the capability check logic.
- Change the capability check from `QUOTE_MESSAGE` to `SEND_REPLY`

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/ab1fdede-3989-4bea-89ed-d58f1e020661" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/87a628cd-9ac3-4225-9e34-09af29cf5629" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Open a channel
2. Open a thread
3. Long press on a message from the thread
4. The "Thread reply" item should not be visible


andremion (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messageoptions/MessageOptions.kt
```diff
@@ -102,6 +103,7 @@ public fun MessageOptions(
 public fun defaultMessageOptionsState(
     selectedMessage: Message,
     currentUser: User?,
+    isInThread: Boolean,
```
andremion, line 106: ⚠️ This might be a breaking change

VelikovPetar: This is actually 'reverted' to how it was pre `6.12.1`. I will try to see if we can somehow use only the data from the `selectedMessage` to deduct this so that we don't need to pass this value as an argument, but I don't think that we can match it 100%.

andremion: What about defining a default value?

VelikovPetar: It is definitely possible, however I don't think that neither `true` nor `false` is really appropriate to be default, as it really depends in which context the method is called.

andremion: That's fine.
Is it worth mentioning it in the CHANGELOG, then?

VelikovPetar: I checked this in the meantime, I don't think that we can avoid passing the `isInThread` flag. The `Message` model has extensions like `isThreadReply`/`isThreadStart`, however they are not enough to determine whether the menu is shown in the context of a Thread. 

The `isThreadReply` is enough for the thread replies (as they will only be shown in the context of a thread), but the `isThreadStart` will always be `true` for the parent message in the thread, no matter whether we make the check in the context of a `Channel` or in the context of a `Thread`. For that case we need the explicit `isInThread` flag.

I will update the `CHANGELOG.md` to highlight this change.

andremion (approved):

PR #5685, opened 2025-03-18
[AND-401] Fix audio recordings not paused

VelikovPetar:
### 🎯 Goal
A playing audio recording attachment in the Message List is not paused/stopped in the following scenarios:
- Minimize app
- Open a screen on top of the Message List
- Delete a message with a playing audio recording

#### `ui-common`
- Standardise the way the `audioHash` is created for both XML and Compose SDKs. On the XML SDK, the `audioHash` was generated as `Attachment::hashCode`, while on the Compose SDK the `audioHash` was generated as `(attachment.assetUrl ?: attachment.upload?.toUri()?.toString()).hashCode()`. This is now unified in the `Attachment::audioHash` extension, and the same is used in both SDK. This is needed for the change in `MessageListController`:
- Before calling `chatClient.deleteMessage()`, in the `MessageListController`, we check if the deleted message has a playing audio recording -> by comparing the `chatClient.audioPlayer.currentPlayingId` with the `message.attachment.audioHash`. If the message currently being deleted has a playing audio, pause it before launching delete request.

#### `ui-components`
- Add a `LifecycleObserver` in the audio recording attachment view holder, which will pause any running audio in `onPause`.
- Additionally, add a similar observer in `MessageListView`, to handle the case when the audio recording viewHolder is recycled (when scrolling away) which causes the item-level observer to be removed.
- In theory, only the `MessageListView` observer is enough to fix the bug, but I added the one on the Item for the case when a customer would use the `MediaAttachmentsViewHolder` independently form the `MessageListView`

#### `compose`
- Add `LifecycleEventEffect` in `AudioRecordAttachmentContent` and `AudioRecordAttachmentPreviewContent`, which pause any running audio in `onPause`.
- Additionally, add a similar observer in `MessageList`, to handle the case when the audio recording item composable is disposed (when scrolling away) which causes the item-level effect to be removed.
- In theory, only the `MessageListView` observer is enough to fix the bug, but I added the one on the Item for the case when a customer would use the `AudioRecordAttachmentContent` independently form the `MessageList `

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/021c1f7a-56e6-4b8a-9ff4-2f7ee39e5200" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/7bfb9552-7f25-4e73-98cb-5579de3c4ce4" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/7001551b-0b1b-44dc-a829-f84df41b7c2a" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/edbd2650-957a-4261-ae6c-951c88f540d6" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Open Compose/XML SDK
2. Open Channel 
3. Send, or scroll to an audio recording message
4. Play it
5. Open a screen on top of it / minimise the app / delete the active message
6. The audio recording should be paused


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -229,6 +232,12 @@ public fun Messages(
             onLastVisibleMessageChanged(message.message)
         }
     }
+
+    /** Clean up: Pause any playing audio tracks in onPause(). **/
+    LifecycleEventEffect(Lifecycle.Event.ON_PAUSE) {
+        val audioPlayer = ChatClient.instance().audioPlayer
+        audioPlayer.pause()
```
JcMinarro, line 239: Our UI should't use the ChatClient itself.
Could we move this logic to the ViewModel?

JcMinarro (approved):

PR #5686, opened 2025-03-20
And 411 emphasize the selected list item

andremion:
### 🎯 Goal

### 🛠 Implementation details

_Describe the implementation_

### 🎨 UI Changes

[list item emphasis tablet.webm](https://github.com/user-attachments/assets/2b0b2145-89d0-4894-94bd-cc0f570db81a)

### 🧪 Testing

<details>

```
Provide the patch code here
```

</details>

### 🎉 GIF


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/EmphasisBox.kt
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * A box that can be used to emphasize its content by adding a transparent overlay on top of it.
+ */
+@Composable
+internal fun EmphasisBox(
+    modifier: Modifier = Modifier,
+    isEmphasized: Boolean = true,
+    content: @Composable () -> Unit,
+) {
+    val color = ChatTheme.colors.textLowEmphasis
```
VelikovPetar, line 36: Small question: Is the `textLowEmphasis` the most appropriate color for this purpose? Maybe the `highlight`(yellow-ish) color one would be more appropriate? (not that big of a deal as this component is internal and we can change it at any point, just wanted to bring this up)

JcMinarro (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
```diff
@@ -92,7 +93,11 @@ private val LocalShapes = compositionLocalOf<StreamShapes> {
 private val LocalUserPresence = compositionLocalOf<UserPresence> {
     error("No UserPresence provided! Make sure to wrap all usages of Stream components in a ChatTheme.")
 }
-private val LocalComponentFactory = compositionLocalOf<ChatComponentFactory> {
+
+/**
+ * The local composition containing the current [ChatComponentFactory].
+ */
+public val LocalComponentFactory: ProvidableCompositionLocal<ChatComponentFactory> = compositionLocalOf {
```
VelikovPetar, line 100: I think that it would be useful to provide an example of this in the `ChatComponentFactory` docs page, because it can be very useful, but I don't think it would be obvious for a customer how to use this.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/EmphasisBox.kt
andremion: `highlight` is already used for pinned messages. Using it for selected items as well won't provide any differentiation as to whether the item is selected or pinned.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
andremion: Good point. I created a ticket for it. Thanks 🙌🏻

VelikovPetar (approved):

PR #5687, opened 2025-03-20
[AND-406] Draft Messages on UI Components

JcMinarro:
### 🎯 Goal
Depend on: #5682

With this implementation show draft messages within the list of channels and message composer.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>XML</th>
<th>Compose</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/e41116d1-ba2b-4c54-acb3-e12a9158cf6d" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/b0410bc8-a5c9-4077-9a4d-0cef719493bb" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

To test it you will need to enable draft messages feature on Channels List and Message Composer.
The following patch add the needed flags

<details>

```
Subject: [PATCH] Print some logs
---
Index: stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt	(revision b6c511809d418dd52bd19605f381a966c5a35015)
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/ChatUI.kt	(date 1742545910050)
@@ -199,7 +199,7 @@
      * Whether draft messages are enabled.
      */
     @JvmStatic
-    public var draftMessageEnabled: Boolean = false
+    public var draftMessageEnabled: Boolean = true
 
     /**
      * Sets the strategy for resizing images hosted on Stream's CDN. Disabled by default,
IDEA additional info:
<+>UTF-8
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt	(revision b6c511809d418dd52bd19605f381a966c5a35015)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt	(date 1742546563527)
@@ -101,6 +101,7 @@
                 Filters.or(Filters.notExists(CHANNEL_ARG_DRAFT), Filters.eq(CHANNEL_ARG_DRAFT, false)),
             ),
+            isDraftMessageEnabled = true,
         )
     }
     private val threadsViewModelFactory by lazy { ThreadsViewModelFactory() }
IDEA additional info:
<+>UTF-8
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt	(revision b6c511809d418dd52bd19605f381a966c5a35015)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt	(date 1742546563546)
@@ -110,6 +110,7 @@
             deletedMessageVisibility = DeletedMessageVisibility.ALWAYS_VISIBLE,
             messageId = intent.getStringExtra(KEY_MESSAGE_ID),
             parentMessageId = intent.getStringExtra(KEY_PARENT_MESSAGE_ID),
+            isComposerDraftMessageEnabled = true,
         )
     }
 
```

</details>

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -1916,6 +1916,16 @@ internal constructor(
             }
     }
 
+    /**
+     * Create a new draft message.
+     * The call will be retried accordingly to [retryPolicy].
+     *
+     * @param channelType The channel type. ie messaging.
+     * @param channelId The channel id. ie 123.
+     * @param message The draft message to create.
+     *
+     * @return Executable async [Call] responsible for creating a draft message.
+     */
     @CheckResult
     public fun createDraftMessage(
         channelType: String,
@@ -1935,6 +1945,16 @@ internal constructor(
         }
     }
 
+    /**
+     * Delete a draft message.
+     * The call will be retried accordingly to [retryPolicy].
+     *
+     * @param channelType The channel type. ie messaging.
+     * @param channelId The channel id. ie 123.
+     * @param message The draft message to delete.
+     *
+     * @return Executable async [Call] responsible for deleting a draft message.
+     */
     @CheckResult
     public fun deleteDraftMessages(
         channelType: String,
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelViewModelFactory.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelViewModelFactory.kt
@@ -46,6 +46,7 @@ public class ChannelViewModelFactory(
     private val memberLimit: Int = ChannelListViewModel.DEFAULT_MEMBER_LIMIT,
     private val messageLimit: Int = ChannelListViewModel.DEFAULT_MESSAGE_LIMIT,
     private val chatEventHandlerFactory: ChatEventHandlerFactory = ChatEventHandlerFactory(chatClient.clientState),
+    private val isDraftMessageEnabled: Boolean = true,
 ) : ViewModelProvider.Factory {
 
     private val factories: Map<Class<*>, () -> ViewModel> = mapOf(
@@ -58,6 +59,7 @@ public class ChannelViewModelFactory(
                 messageLimit = messageLimit,
                 memberLimit = memberLimit,
                 chatEventHandlerFactory = chatEventHandlerFactory,
+                isDraftMessageEnabled = isDraftMessageEnabled,
             )
         },
     )
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelViewModelFactory.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelViewModelFactory.kt
@@ -46,7 +46,7 @@ public class ChannelViewModelFactory(
     private val memberLimit: Int = ChannelListViewModel.DEFAULT_MEMBER_LIMIT,
     private val messageLimit: Int = ChannelListViewModel.DEFAULT_MESSAGE_LIMIT,
     private val chatEventHandlerFactory: ChatEventHandlerFactory = ChatEventHandlerFactory(chatClient.clientState),
-    private val isDraftMessageEnabled: Boolean = true,
+    private val isDraftMessageEnabled: Boolean = false,
 ) : ViewModelProvider.Factory {
 
     private val factories: Map<Class<*>, () -> ViewModel> = mapOf(
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
@@ -488,7 +488,7 @@ public class MessageComposerController(
     private suspend fun saveDraftMessage(messageMode: MessageMode) {
         if (!config.isDraftMessageEnabled) return
         when (val messageText = messageInput.value.text) {
-            "" -> clearDraftMessage()
+            "" -> clearDraftMessage(messageMode)
             else -> {
                 globalState.getDraftMessageOrEmpty(messageMode).let {
                     chatClient.createDraftMessage(
@@ -633,16 +633,16 @@ public class MessageComposerController(
      */
     public fun clearData() {
         logger.i { "[clearData]" }
-        scope.launch { clearDraftMessage() }
+        scope.launch { clearDraftMessage(messageMode.value) }
         messageInput.value = MessageInput()
         selectedAttachments.value = emptyList()
         validationErrors.value = emptyList()
         alsoSendToChannel.value = false
     }
 
-    private suspend fun clearDraftMessage() {
+    private suspend fun clearDraftMessage(messageMode: MessageMode) {
         if (!config.isDraftMessageEnabled) return
-        globalState.getDraftMessage(messageMode.value)?.let { draftMessage ->
+        globalState.getDraftMessage(messageMode)?.let { draftMessage ->
             chatClient.deleteDraftMessages(
                 channelType = channelType,
                 channelId = channelId,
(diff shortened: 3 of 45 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -399,6 +406,7 @@ public class MessageComposerController(
         }.debounce(TEXT_INPUT_DEBOUNCE_TIME).onEach {
             scope.launch { handleMentionSuggestions() }
             scope.launch { handleLinkPreviews() }
+            scope.launch { updateDraftMessageText(it.text) }
```
VelikovPetar, line 409: Wouldn't this cause frequent requests to the `createDraft` endpoint? Not sure if is this is the desired behaviour, but wouldn't it be better if we create the draft message when the user leaves the channels screen (dispose the composer)? Also isn't it redundant to create a draft during typing, if the user afterwards send the message?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -814,6 +841,19 @@ public class MessageComposerController(
      */
     public fun seekRecordingTo(progress: Float): Unit = audioRecordingController.seekRecordingTo(progress)
 
+    private suspend fun updateDraftMessageText(text: String) {
+        if (!config.isDraftMessageEnabled) return
+        currentDraftMessage = currentDraftMessage.copy(text = text)
```
VelikovPetar, line 846: While testing, I noticed that attachments are not handled during creating/fetching of a draft message, even though the `DraftMessage` model already supports attachments. Is this the expected behaviour, or should we also 'persist' the pending attachments?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1915,6 +1916,67 @@ internal constructor(
             }
     }
 
+    @CheckResult
+    public fun createDraftMessage(
```
VelikovPetar, line 1920: Can we add KDoc on the new `ChatClient` operations?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelViewModelFactory.kt
```diff
@@ -46,6 +46,7 @@ public class ChannelViewModelFactory(
     private val memberLimit: Int = ChannelListViewModel.DEFAULT_MEMBER_LIMIT,
     private val messageLimit: Int = ChannelListViewModel.DEFAULT_MESSAGE_LIMIT,
     private val chatEventHandlerFactory: ChatEventHandlerFactory = ChatEventHandlerFactory(chatClient.clientState),
+    private val isDraftMessageEnabled: Boolean = false,
```
VelikovPetar, line 49: Maybe naming this param in plural would be more clear (`isDraftMessagesEnabled`)? (also we are missing KDoc)

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -469,6 +479,14 @@ public class MessageComposerController(
         this.messageInput.value = MessageInput(value, source)
     }
 
+    private fun fetchDraftMessage(messageMode: MessageMode) {
+        if (!config.isDraftMessageEnabled) return
+        currentDraftMessage = globalState.getDraftMessage(messageMode)
```
VelikovPetar, line 484: This is a bit of an edge-case, but this would have a problem if a user opens the messages screen (or just the composer), without going through the channel list, like for example via notification. 
In that case, the `globalState.draftMessages` will not be populated yet, and calling `globalState.getDraftMessage(messageMode)` will return an empty message, even if a draft actually exists.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/viewmodel/channels/ChannelListViewModelFactory.kt
```diff
@@ -43,6 +44,7 @@ public class ChannelListViewModelFactory @JvmOverloads constructor(
     private val limit: Int = ChannelListViewModel.DEFAULT_CHANNEL_LIMIT,
     private val messageLimit: Int = ChannelListViewModel.DEFAULT_MESSAGE_LIMIT,
     private val memberLimit: Int = ChannelListViewModel.DEFAULT_MEMBER_LIMIT,
+    private val isDraftMessageEnabled: Boolean = ChatUI.draftMessageEnabled,
```
VelikovPetar, line 47: Maybe this param can be named in plural (`isDraftMessagesEnabled`)? (also it is missing KDoc)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/viewmodel/messages/MessageListViewModelFactory.kt
```diff
@@ -94,6 +95,7 @@ public class MessageListViewModelFactory @JvmOverloads constructor(
     private val showDateSeparatorInEmptyThread: Boolean = false,
     private val showThreadSeparatorInEmptyThread: Boolean = false,
     private val threadLoadOlderToNewer: Boolean = false,
+    private val isComposerDraftMessageEnabled: Boolean = ChatUI.draftMessageEnabled,
```
VelikovPetar, line 98: Can we add KDoc for the new public param?

VelikovPetar (commented):
I left couple of remarks/questions, but looks very good! GJ!

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
JcMinarro: It is expected, image/video/file attachments are out of scope for now in all platforms.
The SDK uploads attachments after the message is sent, so it would require to change the way our SDK pushes files.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
JcMinarro: It has a debouncer, but maybe we can improve it.
It can't only be called when it is disposed, because it needs to save thread messages as well.
Let me think about it

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
JcMinarro: :thinking: 
It is sync whenever the user connect to the WS, it doesn't depend on the channel to be load

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
VelikovPetar: Got it, thank you for the confirmation!

testableapple:
LGTM! ✅

The only comment is that the Offline Support is out of scope for this PR. Let's implement it in a separate one.

PR #5688, opened 2025-03-21
[AND-413] Customizable message bubble and error icon

VelikovPetar:
### 🎯 Goal
Provides options for customising the following:
**Via `ChatTheme`:**
- Bubble color for error messages
- Text color for error messages
- `TextStyle` based on the current `Message`

**Via `ChatComponentFactory`:**
- `MessageBubble`
- `MessageFailedIcon`

### 🛠 Implementation details
- Add `MessageBubble` and `MessageFailedIcon` as components in the `ChatComponentFactory`
- Replace the usages in `RegularMessageContent`, `EmojiMessageContent`, `PollMessageContent`, `QuotedMessageContent` and `AiRegularMessageContent` of `MessageBubble` and the failed message `Icon` to the new ones from the `ChatComponentFactory`
- Move the  `getMessageBubbleColor()` from `MessageItem` to the `MessageBubble` file and make it `public (@InternalStreamChatApi)`, so that it can be reused by `AiRegularMessageContent`
- Add `errorTextStyle` to `MessageTheme` to change the text color for error messages - It is now used when the corresponding message passes the check: `Message::isErrorOrFailed`. For backwards compatibility, the `errorTextStyle` is set to be equal to the already existing `textStyle`.
- ⚠️ The `MessageTextFormatter.defaultFormatter` argument `textStyle` now also provides a `Message` object, so that the `TextStyle` can be changed per message: `textStyle: (isMine: Boolean, message: Message) -> TextStyle` (instead of `textStyle: (isMine: Boolean) -> TextStyle`). This can be used for a more granular customisation of the textStyle, based on the message itself, not just the message ownership.
- Add `errorBackgroundColor` to `MessageTheme` to change the background (bubble) color for error messages - It is now used when the corresponding message passes the check: `Message::isErrorOrFailed`. For backwards compatibility, the `errorBackgroundColor ` is set to be equal to the already existing `backgroundColor`.
- Add snapshot tests for `RegularMessageContent`, `EmojiMessageContent` and `PollMessageContent`.

Example usage for `MessageTheme`:
```kotlin
val defaultOwnTheme = MessageTheme.defaultOwnTheme(isInDarkMode, typography, shapes, colors)
val ownTheme = defaultOwnTheme.copy(
        errorBackgroundColor = Color.Red,
        errorTextStyle = defaultOwnTheme.errorTextStyle.copy(color = Color.White),
)
ChatTheme(ownMessageTheme = ownTheme) {
        MessageList(listViewModel)
}
```

Example usage for the `MessageTextFormatter`:
```kotlin
val messageTextFormatter = MessageTextFormatter.defaultFormatter(
      autoTranslationEnabled = false,
      textStyle = { isMine, message ->
            if (message.isError()) {
                ownMessageTheme.textStyle.copy(color = Color.Red)
            } else {
                ownMessageTheme.textStyle
            }
       }
)
ChatTheme(messageTextFormatter = messageTextFormatter) {
        MessageList(listViewModel)
}
```

Example usage for `ChatComponentFactory`:
```kotlin
class CustomChatComponentFactory : ChatComponentFactory {

    @Composable
    override fun MessageBubble(
        modifier: Modifier,
        message: Message,
        color: Color,
        shape: Shape,
        border: BorderStroke?,
        contentPadding: PaddingValues,
        content: @Composable () -> Unit
    ) {
        val newColor = if (message.isError()) Color.Red else color
        super.MessageBubble(modifier, message, newColor, shape, border, contentPadding, content)
    }

    @Composable
    override fun MessageFailedIcon(modifier: Modifier, message: Message) {
        // Don't show error icon
    }
}
```

### 🎨 UI Changes
- _No visible UI changes_

### 🧪 Testing
1. Open Compose sample project
2. Apply any of the new customisation options:
  - Set new values for `MessageTheme.errorTextStyle` / `MessageTheme.errorBackgroundColor`
  - And/Or create a new `MessageTextTransformer` and override `textStyle`
  - And/Or override the `MessageBubble`/`MessageFailedIcon` methods of the `ChatComponentFactory`
3. Build app
4. Send an offensive message that would be moderated, or some message that would fail to send (BE error, not connectivity error)
5. The message bubble/text should be styled according the previously applied customisations


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -895,6 +897,55 @@ public interface ChatComponentFactory {
...
+        border: BorderStroke?,
+        contentPadding: PaddingValues,
+        content: @Composable () -> Unit,
+    ) {
+        io.getstream.chat.android.compose.ui.components.messages.MessageBubble(
+            modifier = modifier,
+            color = color,
+            shape = shape,
+            contentPadding = contentPadding,
+            content = content,
```
andremion, line 926: The `border` seems to be missed

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -895,6 +897,55 @@ public interface ChatComponentFactory {
...
+     * @param color The color of the message bubble.
+     * @param shape The shape of the message bubble.
+     * @param border The border of the message bubble.
+     * @param contentPadding The padding of the message bubble.
+     * @param content The content shown inside the message bubble.
+     */
+    @Composable
+    public fun MessageBubble(
+        modifier: Modifier,
+        message: Message,
```
andremion, line 914: I believe `message` is not needed here because it is already bound in the `content`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -895,6 +897,55 @@ public interface ChatComponentFactory {
...
+    /**
+     * Icon shown at the bottom-end of a [Message] when it failed to send.
+     *
+     * @param modifier Prepared [Modifier] for styling.
+     * @param message The [Message] that failed to send.
+     */
+    @Composable
+    public fun MessageFailedIcon(
+        modifier: Modifier,
+        message: Message,
```
andremion, line 939: Is `message` needed here?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
VelikovPetar: This is actually intentional, because I wanted for the customers to be able to style the message bubble based on the message itself. For example, a customer might want to show a bubble with a different color/border based on the message type (red background when `message.type == "error"`). If they cannot access the message from this factory method, they wouldn't; be able to do such customisation.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
VelikovPetar: Similar to the `MessageBubble`, I provided the `Message` object just in case a customer would need some info from it in order to customise the failed icon. Thinking about it, perhaps we wouldn't really need it, because it would be shown only in the specific case where the `MessageItemState.isErrorOrFailed()` condition is satisfied, so there isn't really that much room for customisation here.
But I think it would still be useful for distinguishing different types of errors, for example the case when we have `moderation = bounce`. A user could distinguish whether the message failed due to being bounced by moderation or a different error, based on the presence of the `moderation` object in the `Message` (and show/hide the icon based on it).

andremion (approved):

PR #5689, opened 2025-03-24
Not show deleted poll messages

JcMinarro:
### 🎯 Goal
Render deleted poll messages as a deleted message instead of use the "poll UI"
Fix: https://linear.app/stream/issue/AND-416/inability-to-delete-polls-in-android-app-using-streams-ui-components

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/011fe8c9-6d4c-4f88-9dd9-3a316e82d352" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/c69332d5-6164-4436-a1fa-6cbc13e5f312" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>


VelikovPetar (approved):
LGTM!

Maybe an idea for the future, because currently the rendering of a deleted poll will be done via the `RegularMessageContent` composable, we can introduce a `deletedMessageContent` lambda on the `PollMessageContent` composable, so that the deleted state can be handled by the `PollMessageContent`, without relying on the parent `MessageItem` composable.

PR #5690, opened 2025-03-25
[AND-274] Add unit tests for stream-chat-android extensions package.

VelikovPetar:
🎯 Goal

Adds unit tests covering the `extensions` package of the `stream-chat-android-client`.

🧪 Testing

Run tests in stream-chat-android-client

### DIFF
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/SocketStateTest.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/SocketStateTest.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright (c) 2014-2025 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.clientstate
+
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+
+internal class SocketStateTest {
+
+    @Test
+    fun testConnectionIdOrErrorThrowsErrorForIdle() {
+        val socketState = SocketState.Idle
+        assertThrows(IllegalStateException::class.java) {
+            socketState.connectionIdOrError()
+        }
+    }
+
+    @Test
+    fun testConnectionIdOrErrorThrowsErrorForPending() {
+        val socketState = SocketState.Pending
+        assertThrows(IllegalStateException::class.java) {
+            socketState.connectionIdOrError()
+        }
+    }
+
+    @Test
+    fun testConnectionIdOrErrorReturnsConnectionIdForConnected() {
+        val connectionId = "connectionId"
+        val socketState = SocketState.Connected(connectionId)
+        assert(socketState.connectionIdOrError() == connectionId)
+    }
+
+    @Test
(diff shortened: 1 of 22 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/SocketStateTest.kt
```diff
@@ -0,0 +1,54 @@
...
+
+package io.getstream.chat.android.client.clientstate
+
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Test
+
+internal class SocketStateTest {
+
+    @Test
+    fun testConnectionIdOrErrorThrowsErrorForIdle() {
```
andremion, line 25: These test names could follow the other tests and be separated by spaces

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/SocketStateTest.kt
```diff
@@ -0,0 +1,54 @@
...
+        assertThrows(IllegalStateException::class.java) {
+            socketState.connectionIdOrError()
+        }
+    }
+
+    @Test
+    fun testConnectionIdOrErrorReturnsConnectionIdForConnected() {
+        val connectionId = "connectionId"
+        val socketState = SocketState.Connected(connectionId)
+        assert(socketState.connectionIdOrError() == connectionId)
```
andremion, line 44: Using assertion functions from `org.junit.jupiter.api` provides a better error message, which helps us identify the root cause, for example:

Using jupiter:
```
Expected :true
Actual   :false
<Click to see difference>

org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
```
Using assert:
```
Assertion failed
java.lang.AssertionError: Assertion failed
```

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/extensions/AttachmentExtensionsTests.kt
```diff
@@ -0,0 +1,91 @@
...
+import io.getstream.chat.android.randomAttachment
+import org.amshove.kluent.shouldBeEqualTo
+import org.junit.jupiter.api.Test
+
+internal class AttachmentExtensionsTests {
+
+    @Test
+    fun `isImage should return true for image mime type`() {
+        val attachment = randomAttachment(mimeType = "image/jpeg")
+        attachment.isImage shouldBeEqualTo true
```
andremion, line 28: Although Kluent provides nice infix functions, `org.junit.jupiter.api` provides a better error message. When comparing objects, it provides a <Click to see difference> button which is very handy.

andremion (commented):
Nice coverage 👏🏻

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/SocketStateTest.kt
VelikovPetar: I will update them!

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/clientstate/SocketStateTest.kt
VelikovPetar: I missed this, I also usually try to avoid doing this!

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/extensions/AttachmentExtensionsTests.kt
VelikovPetar: I can see that across the project,  the `kluent` infix functions are the most used ones methods for performing the assertions. Maybe it would be good to align this across the whole project.

VelikovPetar: Edit: Based on the Slack discussion, I will update the assertions to use `jupiter` instead of `kluent`, but I will make the update in a separate PR, so that we don't mix contexts in this PR (if you agree with that)

andremion: Sure thing 👍🏻

PR #5691, opened 2025-03-25
(part 6) `stream-chat-android-client` unit tests - utilities

VelikovPetar:
🎯 Goal

- Adds unit tests covering different utility classes from `stream-chat-android-client`
- Move `ValidationKtTests` from `offline` to `client` (align packages)
- Add KDocs on tested classes

🧪 Testing
- Run tests in stream-chat-android-client


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/logger/ChatLogLevelTest.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+internal class ChatLogLevelTest {
+
+    @Test
+    fun `ALL should be more or equal than all levels`() {
+        assertTrue(ChatLogLevel.ALL.isMoreOrEqualsThan(ChatLogLevel.ALL))
```
andremion, line 27: `isMoreOrEqualsThan` is an internal function used only for tests.
`ChatLogLevel.severity` is a private attribute used only in `isMoreOrEqualsThan`
I'm wondering if all of them could be removed, including their tests 🤔

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/logger/StreamLoggerHandlerTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    fun `When logging with ASSERT priority and throwable not null Should call logE on ChatLoggerHandler`() {
+        val throwable = Throwable()
+        streamLoggerHandler.log(Priority.ASSERT, tag, message, throwable)
+        verify(chatLoggerHandler).logE(tag, message, throwable)
+        verifyNoMoreInteractions(chatLoggerHandler)
+    }
+}
```
andremion, line 90: It's missing this test case:

```kotlin
    @Test
    fun `When logging with null handler Should not call any function`() {
        val streamLoggerHandler = StreamLoggerHandler(null)
        streamLoggerHandler.log(Priority.VERBOSE, tag, message, null)
        verifyNoMoreInteractions(chatLoggerHandler)
    }
```

VelikovPetar: Nice catch, I will add it!

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/logger/ChatLogLevelTest.kt
VelikovPetar: Hmm interesting, I will look into this.

VelikovPetar: Update on this: It seems safe to delete, and it is also not a breaking change. 
@JcMinarro Do you maybe see a reason to still keep the `severity` property and the `isMoreOrEqualsThan` method, even though they are `internal` and not used?

JcMinarro: I think it is fine.
The logic to show or not a log was moved to the `StreamLog` dependency.
It is mainly used as an intermediary instance to be decoupled of Logger

VelikovPetar: Nice, thank you for the confirmation!
I will clean this up

VelikovPetar: Updated here: https://github.com/GetStream/stream-chat-android/pull/5691/commits/69ce1fdfb2d12976e3cb00c44f96e428c905e2fc

andremion (approved):

PR #5692, opened 2025-03-26
[AND-389] Mention list presentation

andremion:
### 🎯 Goal

Create reusable presentation components for the mention list.

### 🛠 Implementation details

- Introduce a `MentionListController` to manage the mention list presentation logic, which is reused in both view models of `ui-components` and `compose` SDKs.
- Make the list loading item to be reused by pinned messages and thread list UI components.
- Enable the pagination in the `MentionListView` and add the missing loading item UI.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/04f5b1c4-3029-4d29-b15a-39628bce31b5" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/e9ebeb4e-9ef8-4044-957c-9b89cd0491d6" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🎉 GIF

gif


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/mentions/MentionListController.kt
```diff
@@ -0,0 +1,166 @@
...
+     * The current state of the mention list.
+     */
+    public val state: StateFlow<MentionListState> = _state.asStateFlow()
+
+    private val _events = MutableSharedFlow<Event<Any>>(extraBufferCapacity = 1)
+
+    /**
+     * One shot events triggered by the controller. Currently only errors are emitted.
+     */
+    public val events: SharedFlow<Event<Any>> = _events.asSharedFlow()
```
VelikovPetar, line 73: Maybe exposing `Any` would be a bit confusing for an integrator? Perhaps we can model the events by using a `sealed class`?

(I see that previously we only emitted `Event<Unit>` representing errors, so if there are no other possible events, we could also rename the flow to `errorEvents`)

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/mentions/MentionListState.kt
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * The state of the mention list on which the UI should be rendered.
+ *
+ * @param isLoading True if the initial loading is in progress.
+ * @param results The list of mentions to be displayed.
+ * @param nextPage The next page token to be loaded *(Internal usage only)*
+ * @param canLoadMore True if there are more mentions to be loaded.
+ * @param isLoadingMore True if the loading of the next page is in progress.
+ */
+public data class MentionListState(
```
VelikovPetar, line 30: I believe that this is a breaking change from the previous `MentionListViewModel.State` - I think it would be good to mention this in the `CHANGELOG.md`.

andremion: I already mentioned it in the Changed block. I will add a breaking change prefix then.

VelikovPetar (approved):
Looks good to me!

PR #5693, opened 2025-03-26
POC - Process death recovery

VelikovPetar:
### 🎯 Goal

### 🛠 Implementation details

_Describe the implementation_

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -120,6 +120,9 @@ import io.getstream.chat.android.client.persistance.repository.noop.NoOpReposito
 import io.getstream.chat.android.client.plugin.DependencyResolver
 import io.getstream.chat.android.client.plugin.Plugin
 import io.getstream.chat.android.client.plugin.factory.PluginFactory
+import io.getstream.chat.android.client.process.ProcessDeathRecoveryLifecycleObserver
+import io.getstream.chat.android.client.process.ProcessDeathRecoveryStorage
+import io.getstream.chat.android.client.process.SharedPreferencesProcessDeathRecoveryStorage
 import io.getstream.chat.android.client.query.AddMembersParams
 import io.getstream.chat.android.client.query.CreateChannelParams
 import io.getstream.chat.android.client.scope.ClientScope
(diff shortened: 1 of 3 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -399,6 +415,30 @@ internal constructor(
...
     }
 
+    /**
+     * Attempts to recover the SDK state after the process death. Relevant for the case where we try to access a
+     * [Plugin] after the app was restored from the background, and [connectUser] was not called yet.
+     * Attempts to call [connectUser] with the stored [User] in the [ProcessDeathRecoveryStorage], and the token stored
+     * in the [UserCredentialStorage]. If one of them are missing, the recovery is aborted.
+     */
+    @InternalStreamChatApi
+    public fun recoverFromProcessDeath() {
```
VelikovPetar, line 425: This method will attempt to call `connectUser` with the `User` from the last session - but only if that user still exists in the local storage. 

The `User` is persisted when:
- The app is put in background (and there is a connected user) -> because the app becomes a candidate for system kill

The `User` is cleared from storage when:
- `connectUser` is called -> We can assume a **valid** user connection in such case
- `disconnect` is called -> We must not preserve the `User` in this case

If no such user exists, this method will do nothing, and the app would still crash due to calling `resolvePluginDependence` before calling `connectUser`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/process/ProcessDeathRecoveryLifecycleObserver.kt
```diff
@@ -0,0 +1,44 @@
...
+import io.getstream.log.taggedLogger
+
+/**
+ * Lifecycle observer that listens to the process lifecycle events and updates the
+ * [ProcessDeathRecoveryStorage] accordingly.
+ */
+internal class ProcessDeathRecoveryLifecycleObserver(
+    private val clientState: ClientState,
+    private val storage: ProcessDeathRecoveryStorage,
+) : DefaultLifecycleObserver {
```
VelikovPetar, line 31: `LifecycleObserver` attached to the process lifecycle. Stores the current `User` in `onStop`, which is called when the app is backgrounded (could potentially be killed by the system)

PR #5694, opened 2025-03-27
Show/Hide poll icon depending on poll is enabled on the channel for the current user

JcMinarro:
### 🎯 Goal

The AttachmentPicker from the compose artifact was not checking the channel's capabilities to show/hide the poll tab, and then, when it wasn't enabled for a channel, the poll icon didn't do anything, giving the impression it was broken.
With the new change, the poll icon is shown/hidden depending on the capabilities of the channel

Fix: AND-425


VelikovPetar (approved):
LGTM!

PR #5695, opened 2025-03-31
[AND-389] Mention list

andremion:
### 🎯 Goal

### 🛠 Implementation details

_Describe the implementation_

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

<details>

```
Provide the patch code here
```

</details>

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/AppBottomBar.kt
```diff
@@ -77,6 +78,12 @@ fun AppBottomBar(
                 }
             },
         )
+        AppBottomBarOptionTile(
+            icon = UiCommonR.drawable.stream_compose_ic_mentions,
```
VelikovPetar, line 82: Not a big deal, but maybe we shouldn't directly depend on the drawable resources from the SDK in the sample, to avoid any unintentional changes if the resources in the SDK change.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/mentions/MentionList.kt
```diff
@@ -0,0 +1,319 @@
...
+            )
+        }
+    },
+    loadingItemContent: @Composable LazyItemScope.() -> Unit = {
+        with(ChatTheme.componentFactory) {
+            MentionListLoadingItem(
+                modifier = Modifier,
+            )
+        }
+    },
```
VelikovPetar, line 181: Shouldn't we also show (and provide a way to customize) the initial fullscreen loading content? For example, if a customer would like to show a fullscreen loading state (instead of empty + pullToRefresh indicator), it will not be easily possible. (also it is a bit different from the XML implementation)

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/AppBottomBar.kt
andremion: We currently have 10 occurrences in the compose sample of reusing drawables from the SDK. We would have to duplicate all of that.
I don't think it is a big deal, as we can always adjust the sample in case of any changes.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/handlers/LoadMoreHandler.kt
```diff
@@ -34,20 +33,36 @@ import androidx.compose.runtime.remember
 @Composable
 public fun LoadMoreHandler(
     listState: LazyListState,
-    loadMoreThreshold: Int = 3,
+    loadMoreThreshold: Int = DefaultLoadMoreThreshold,
```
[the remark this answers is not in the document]
andremion: Good spot!
Indeed it was intentional.
The message list was using a threshold of 3, but the thread list and pinned messages were using 10.
The search user results was not using any threshold at all.
Do you think there is any issue with using it as 10 in all those cases?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/mentions/MentionList.kt
VelikovPetar: Ah nice, I got focused too much on the pull-to-refresh component. Looks good then.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/handlers/LoadMoreHandler.kt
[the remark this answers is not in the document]
VelikovPetar: I am not sure if there is any historical reason for the discrepancy, but since this is public API, perhaps we should keep the default version the same as the old one. There is a small chance that this could be breaking someone's integration. For places that use a different value than the default one, maybe we can pass the `loadMoreThreshold` from outside?

CHANGELOG.md
```diff
@@ -144,10 +144,12 @@
 - Add `MessageBubble` and `MessageFailedIcon` to `ChatComponentFactory` to allow customization of the message bubbles and the failed message icon. [#5688](https://github.com/GetStream/stream-chat-android/pull/5688)
 - Add `errorTextStyle` and `errorBackgroundColor` to `MessageTheme` to allow customization of the error message bubbles and error message text style. [#5688](https://github.com/GetStream/stream-chat-android/pull/5688)
 - ⚠️Add `Message` as an argument of the `textStyle` lambda parameter of the `MessageTextFormatter.defaultFormatter` method to allow customizing the text style based on the `Message`. [#5688](https://github.com/GetStream/stream-chat-android/pull/5688)
+- Introduce `MentionList` to display a list of mentions for the current user. [#5695](https://github.com/GetStream/stream-chat-android/pull/5695)
```
VelikovPetar, line 147: I think that this is in the wrong part of the CHANGELOG.md, it should higher in the "Unreleased" section.

VelikovPetar (approved):

PR #5705, opened 2025-04-01
[AND-431] Support offline draft messages

JcMinarro:
### 🎯 Goal
Support offline draft messages


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DraftMessageEntity.kt
```diff
@@ -0,0 +1,40 @@
...
+    @PrimaryKey
+    val id: String,
+    val cid: String,
+    val text: String,
+    val parentId: String? = null,
+    val mentionedUsersIds: List<String> = listOf(),
+    val silent: Boolean,
+    val showinChannel: Boolean,
+    val replyMessageId: String? = null,
+    val extraData: Map<String, Any> = mapOf(),
```
VelikovPetar, line 35: I assume the attachments are not included on purpose because they are not supported at the moment? 
We should just make sure not to forget to add them to the BD once we introduce support for them.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DraftMessageEntity.kt
```diff
@@ -0,0 +1,40 @@
...
+) {
+    companion object {
+        internal const val DRAFT_MESSAGE_ENTITY_TABLE_NAME = "draft_message_entity"
```
VelikovPetar, line 38: I see that usually the tables in the DB have the `stream_chat_` prefix in their name - should we maybe add it here as well?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DraftMessageEntity.kt
```diff
@@ -0,0 +1,40 @@
...
+)
+internal data class DraftMessageEntity(
+    @PrimaryKey
+    val id: String,
+    val cid: String,
+    val text: String,
+    val parentId: String? = null,
+    val mentionedUsersIds: List<String> = listOf(),
+    val silent: Boolean,
+    val showinChannel: Boolean,
```
VelikovPetar, line 33: Small typo here: `showinChannel` vs `showInChannel`

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DraftMessageEntity.kt
andremion, line 38: nit: Given that the constant is already part of `DraftMessageEntity`, the `DRAFT_MESSAGE_ENTITY_` prefix seems to be redundant.

`DraftMessageEntity.TABLE_NAME` would be clear enough.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/DraftMessageListenerDatabase.kt
```diff
@@ -0,0 +1,113 @@
...
+     * @param channelId The id of the channel
+     * @param message The draft message to be deleted
+     */
+    override suspend fun onDeleteDraftMessagesResult(
+        result: Result<Unit>,
+        channelType: String,
+        channelId: String,
+        message: DraftMessage,
+    ) {
+        messageRepository.deleteDraftMessage(message)
```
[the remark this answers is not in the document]
JcMinarro: It is done to ensure it is processed on the local db event if there isn't network connection, but you give me an idea to change it

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DraftMessageEntity.kt
JcMinarro: The name is a bit redundant, but as we are using it on the DAO query, it avoid us to use `{}` within the query string

JcMinarro: We can add it :+1:

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DraftMessageEntity.kt
JcMinarro: Attachments are out of scope because they can't be supported without sending the message before.
This offline data will only be available when the user is offline, on the rest of the cases the value from globalState is obtained. Keeping track of attachments within our DB would require following the sync process we currently have for attachments and a embedded relationship in our DB that won't be used

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/DraftMessageListenerDatabase.kt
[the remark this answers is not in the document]
JcMinarro: I just refactored this behavior [here](https://github.com/GetStream/stream-chat-android/pull/5705/commits/7f6aa910896e09458740b50817c83cd8e1a1cad4) :upside_down_face:

andremion (approved):

VelikovPetar (approved):

PR #5712, opened 2025-04-04
[AND-440] Fix the typing list item focus

andremion:
### 🎯 Goal

Focus on the typing item when it appears, and the scroll position is near the bottom.

### 🛠 Implementation details

Return a `NewMessageState` for the typing indicator item, so that the UI can behave properly.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/58da32e6-8aec-4460-9da9-9ccb43dffebc" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/7652a09c-66a0-4148-b975-4d5f0090f39d" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

- Create a custom typing indicator item (the default is empty)

```kotlin
public interface ChatComponentFactory {
    @Composable
    public fun LazyItemScope.MessageListTypingIndicatorItemContent(typingItem: TypingItemState) {
        Text(modifier = Modifier.padding(10.dp), text = "Typing...")
    }
}
```
- Enter a channel and start typing from the other channel member login.

### 🎉 GIF

gif

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/NewMessageState.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/NewMessageState.kt
@@ -30,3 +30,8 @@ public data class MyOwn(val ts: Long?) : NewMessageState()
  * If the message is someone else's (we didn't send it), we show a "New message" bubble.
  */
 public data class Other(val ts: Long?) : NewMessageState()
+
+/**
+ * If it is a typing message, we scroll to the bottom of the list.
+ */
+public data object Typing : NewMessageState()
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/NewMessageState.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/NewMessageState.kt
@@ -16,28 +16,22 @@
 
 package io.getstream.chat.android.ui.common.state.messages.list
 
-import io.getstream.chat.android.core.internal.InternalStreamChatApi
-
 /**
  * Represents the state when a new message arrives to the channel.
  */
-@InternalStreamChatApi
 public sealed class NewMessageState
 
 /**
  * If the message is our own (we sent it), we scroll to the bottom of the list.
  */
-@InternalStreamChatApi
 public data class MyOwn(val ts: Long?) : NewMessageState()
 
 /**
  * If the message is someone else's (we didn't send it), we show a "New message" bubble.
  */
-@InternalStreamChatApi
 public data class Other(val ts: Long?) : NewMessageState()
 
 /**
  * If it is a typing message, we scroll to the bottom of the list.
  */
-@InternalStreamChatApi
 public data object Typing : NewMessageState()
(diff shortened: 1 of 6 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/NewMessageState.kt
```diff
@@ -16,17 +16,28 @@
 
 package io.getstream.chat.android.ui.common.state.messages.list
 
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
+
 /**
  * Represents the state when a new message arrives to the channel.
  */
+@InternalStreamChatApi
```
VelikovPetar, line 24: This would be a breaking change, I don't think that we should hide these classes from the public API.

VelikovPetar (commented):
Wouldn't this change cause an automatic scroll to the bottom of the message list, when the typing indicator shows? I think that such jumps might be confusing for the user if he has scrolled upwards (even if just a bit)

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/NewMessageState.kt
andremion: This doesn't break anything.
It only marks as internal, which in fact, it should be.

andremion:
> Wouldn't this change cause an automatic scroll to the bottom of the message list, when the typing indicator shows? I think that such jumps might be confusing for the user if he has scrolled upwards (even if just a bit)

There's a logic already that automatically scrolls to the bottom on a new message if the user is not far from the bottom.
This PR doesn't change that. It only fixes the scrolling when the typing indicator appears.

VelikovPetar:
> > Wouldn't this change cause an automatic scroll to the bottom of the message list, when the typing indicator shows? I think that such jumps might be confusing for the user if he has scrolled upwards (even if just a bit)
> 
> @VelikovPetar , There's a logic already that automatically scrolls to the bottom on a new message if the user is not far from the bottom. This PR doesn't change that. It only fixes the scrolling when the typing indicator appears.

@andremion I see, that makes sense. But I think that we would see automatic scrolls to the bottom, even when the user doesn't provide a `Composable` for the `typingIndicator`, and it would look like random jumps to the bottom of the chat. But at the moment, I cannot think of way to avoid the scroll, if there is no custom typingIndicator provided.

andremion:
> > > Wouldn't this change cause an automatic scroll to the bottom of the message list, when the typing indicator shows? I think that such jumps might be confusing for the user if he has scrolled upwards (even if just a bit)
> > 
> > 
> > @VelikovPetar , There's a logic already that automatically scrolls to the bottom on a new message if the user is not far from the bottom. This PR doesn't change that. It only fixes the scrolling when the typing indicator appears.
> 
> @andremion I see, that makes sense. But I think that we would see automatic scrolls to the bottom, even when the user doesn't provide a `Composable` for the `typingIndicator`, and it would look like random jumps to the bottom of the chat. But at the moment, I cannot think of way to avoid the scroll, if there is no custom typingIndicator provided.

@VelikovPetar , good point.
I will test it later or Monday to see how it works in that case.
Thanks

andremion:
Hey @VelikovPetar , this is the behavior when typing and the scroll is near the bottom.
As we can see, there's an existing logic that only scrolls to the bottom if there are only 3 hidden items.

[Screen_recording_20250414_155155.webm](https://github.com/user-attachments/assets/c60c291c-f0e9-49f6-882f-7a87bada2dc9)

VelikovPetar (approved):

PR #5721, opened 2025-04-09
Fix crash when opening the ChatsScreen on tablet

andremion:
### 🛠 Implementation details

It's a very strange crash that only happens when using the `ChannelsScreen` in `ChatsScreen`. 

```
FATAL EXCEPTION: main
Process: io.getstream.chat.android.compose.sample.debug, PID: 3813
java.lang.AbstractMethodError: abstract method "void io.getstream.chat.android.compose.ui.theme.ChatComponentFactory.ChannelItemUnreadCountIndicator(int, androidx.compose.ui.Modifier, androidx.compose.runtime.Composer, int)"
at io.getstream.chat.android.compose.ui.channels.list.ChannelItemKt.DefaultChannelItemTrailingContent(ChannelItem.kt:310)
at io.getstream.chat.android.compose.ui.theme.ChatComponentFactory$DefaultImpls.ChannelItemTrailingContent(ChatComponentFactory.kt:440)
```

Maybe it's because we are using compound factories for the channel item, but it does not justify it as a cause.

The fix consists of changing the internal `DefaultChannelItemTrailingContent` to have parameters in which we can define the content of the indicator using the current factories.

#### Unrelated changes

Remembering some searches that are done in the `DefaultChannelItemTrailingContent` to improve UI performance

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/1303653a-21e2-4f76-8409-690379b04798" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/742b82a7-f36e-43f6-89bb-8e7788a4f1f8" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

- Enable Adaptive layout in the Advance settings
- Log in with any user

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelItem.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelItem.kt
@@ -274,7 +274,7 @@ internal fun RowScope.DefaultChannelItemTrailingContent(
     unreadCountIndicator: @Composable (unreadCount: Int) -> Unit,
     readStatusIndicator: @Composable (message: Message) -> Unit,
 ) {
-    val lastMessage = channel.getLastMessage(currentUser)
+    val lastMessage = remember(channel) { channel.getLastMessage(currentUser) }
 
     if (lastMessage != null) {
         Column(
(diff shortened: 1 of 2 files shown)


VelikovPetar (approved):

andremion:
I'm closing this as the issue seems not to happen anymore.

PR #5722, opened 2025-04-09
[AND-448] Fix MessageItem not respecting messageItemMaxWidth.

VelikovPetar:
### 🎯 Goal
The `ChatTheme.dimens.messageItemMaxWidth` wasn't being respected for values larger than `300dp` due to a fixed `widthIn(max = 300.dp)` modifier on the parent component. With this PR, the limit is removed, and we ensure that the `ChatTheme.dimens.messageItemMaxWidth` is respected (if it would fit in the screen)

### 🛠 Implementation details
- Remove fixed `widthIn(max = 300.dp)` modifier
- Ensure the message bubble takes up as much possible space as needed, but not more than the instructed value

### 🎨 UI Changes

| messageItemMaxWidth | Before | After |
| --- | --- | --- |
| 250dp (default)| before-250 | after-250 |
| 300dp | before-300 | after-300 |
| 400dp | before-400 | after-400 |
| 500dp | before-500 | after-500 |

### 🧪 Testing
1. Set `ChatTheme.dimens.messageItemMaxWidth` to a specific value
2. Run Compose sample
3. The message bubble show respect the new maxWidth (if possible)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -291,12 +292,15 @@ public fun MessageItem(
...
                 modifier
-                    .widthIn(max = 300.dp)
+                    .wrapContentWidth()
                     .then(clickModifier)
                     .testTag("Stream_MessageCell"),
             ) {
                 leadingContent(messageItem)
-                Column(horizontalAlignment = messageAlignment.contentAlignment) {
+                Column(
+                    modifier = Modifier.weight(1f, fill = false),
```
andremion, line 301: Is this modifier really necessary there?
I believe it works without it 🤔

VelikovPetar: I actually think it is necessary, because without it, it obstructs the `trailingContent` after it. (the trailing spacer on own messages is gone when the maxWidth is set to a big value):
no-modifier

andremion (approved):

PR #5723, opened 2025-04-10
Properly Store/Obtain reactions from offline DB

JcMinarro:
### 🎯 Goal
After migrating reactions to the new structure, they weren't properly stored on the local DB.
It caused the offline data of a conversation to be considered different than the data we receive from the API even if there wasn't any update on it. As far as our MessageList components fetch data in parallel from offline DB and Server API,  it caused multiple UI updates while messages were rendered on the screen causing a flickering effect if there is multiple messages with reactions.

Fix: AND-423

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/152c5a8b-94b2-420b-b0f6-3d402d56a21b" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/d0d83ff9-394d-485c-99a0-385ffc26eb9e" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>


VelikovPetar (approved):

andremion (approved):
🎉

PR #5733, opened 2025-04-16
[AND-453] Fix crash when multiple answers field is not a valid number when creating a poll

andremion:
Blocked by #5735 

### 🛠 Implementation details

Entering an invalid integer in the multiple answers field during the Poll creation makes the app crash.
That's because the multiple answer field is configured with `KeyboardType.Decimal`, but the code parses the value from `String` to `Int` in both cases for `KeyboardType.Number` and `KeyboardType.Decimal`.

This PR introduces a testable function that returns `PollOptionError` if the range validation fails and supports integers and floats, according to the `KeyboardType`.

The snapshot test was updated to `PIXEL_4A` instead of `PIXEL_2` to avoid clipping the UI.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/2ae5dec8-c95e-4441-98ea-306bdf1daea3" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/7c57fa6e-051e-40d9-80b6-aea4139b0de8" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>


VelikovPetar (approved):
Looks good!

PR #5735, opened 2025-04-17
Review poll snapshot tests

andremion:
- Split `PollUITest` into a test class per component.
- Reuse preview composables in snapshot tests.
- Removing unnecessary null check of `selectedPoll`.
- Stop using `darkMode: Boolean` from internal composables in the snapshot tests. Use `ChatTheme` in each `Preview` function.

Blocks #5733


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/poll/PollCreationDiscardDialog.kt
```diff
@@ -123,9 +124,15 @@ public fun PollCreationDiscardDialog(
...
-        PollCreationDiscardDialog(
-            onCancelClicked = {},
-            onDiscardClicked = {},
-        )
+        PollCreationDiscardDialog()
     }
 }
+
+@Composable
+internal fun PollCreationDiscardDialog() {
```
VelikovPetar, line 132: Not something that we need to address now (or ever), just a small concern that it might be a bit confusing in some scenarios which composable we need to use, if both the actual composable and its preview/test target composables have the same name.

stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.messages.attachments.poll_PollOptionListTest_blank_items.png
VelikovPetar, line 1: Quick question about the `PollOptionList` composable and its previews: I could be missing something, but shouldn't the previews also display the `Options` title? 

Screenshot 2025-04-22 at 09 31 13

andremion: I don't know if I got your question.
We have this function that shows the title too:

```kotlin
@Composable
internal fun PollOptionListDuplicatedError() {
    @Suppress("MagicNumber")
    PollOptionList(
        optionItems = List(3) { PollOptionItem("This is a poll item $it") } +
            PollOptionItem(
                title = "This is a poll item with error",
                pollOptionError = PollOptionDuplicated("duplicated!"),
            ),
        onQuestionsChanged = {},
    )
}
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/poll/PollCreationDiscardDialog.kt
andremion: Functions reused in the snapshot tests should be wrapped in a ChatTheme because the snapshot tests will wrap it with the proper configuration.
So we have the internal function reused by the previews and snapshot tests, and a private suffixed `Preview` function only used in previews.
That was a way I found to reuse them, prevent duplication, and increase code coverage.
We can always improve it later if you have any other ideas. 🙏🏻

stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.messages.attachments.poll_PollOptionListTest_blank_items.png
VelikovPetar: In the actual screenshots, the `Title` is not visible (ex: the stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.messages.attachments.poll_PollOptionListTest_blank_items.png), although it should always be shown. But I think it is because the `PollOptionList` does not wrap its children in a column, so they are overlapping. But since the `PollOptionList` is used in a column, this is not visible in the actual implementation, but just in the previews.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/poll/PollCreationDiscardDialog.kt
VelikovPetar: I think we can keep it like this, let's see if it causes some confusion/issues in the future!

stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.messages.attachments.poll_PollOptionListTest_blank_items.png
andremion: Ohh, now I got you.
Good point.
Let me check it.

andremion: ✅ 
Thank you @VelikovPetar

VelikovPetar (approved):

PR #5739, opened 2025-04-21
[AND-467] Add `ChatClient` `castPollVote`/`removePollVote` methods which use option/vote IDs.

VelikovPetar:
### 🎯 Goal
The `ChatClient.castPollVote(String, String, Option)` requires the user to send a full `Option` object, even though the operation only requires the `Option.id`. This makes the method a bit difficult to use, if a customer is only using the LLC. Therefore this PR introduces a new `ChatClient.castPollVote(String, String, String)` method which accepts only the option ID, instead of the full object. Additionally, it deprecates the 'old' method, so that it can be cleaned up in the future.
_The same thing is done for the `ChatClient.removePollVote(String, String, Vote) method.`_

### 🛠 Implementation details
- Add `ChatClient.castPollVote(messageId: String, pollId: String, optionId: String)` method.
- Deprecate `ChatClient.castPollVote(messageId: String, pollId: String, option: Option)` method.
- Add `ChatClient.removePollVote(messageId: String, pollId: String, voteId: String)` method.
- Deprecate `ChatClient.removePollVote(messageId: String, pollId: String, vote: Vote)` method.
- Replace usages of the deprecated methods with the new ones.
- Fix wrong error event emitted from the `MessageListController.removeVote` (`PollCastingVoteError` instead of `PollRemovingVoteError`)

### 🎨 UI Changes
_NA_

### 🧪 Testing
Ensure all poll operations are working correctly:
1. Open a sample app
2. Open a channel
3. Create a poll
4. Vote on the poll, Change the poll vote, Remove the poll vote


VelikovPetar:
Closing the PR as we agreed to not make this change. Will open a separate PR which fixes the error emission bug.

PR #5740, opened 2025-04-22
Deprecate third-party calls related APIs (Agora and 100ms).

JcMinarro:
### 🎯 Goal
The integration of video calls using third-party libraries is no longer natively supported. 
In this PR, all related API calls to our server have been marked as deprecated and will be removed in upcoming releases.


andremion (approved):
Sweep them out 🧹

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/VideoCallToken.kt
```diff
@@ -25,6 +25,10 @@ import androidx.compose.runtime.Immutable
  * @property agoraUid The Uid of Agora.
  * @property agoraAppId The App Id of Agora.
  */
+@Deprecated(
+    "This third-party library integration is deprecated. Contact to the support team for more information.",
+    level = DeprecationLevel.WARNING,
+)
```
VelikovPetar, line 31: There are also couple more related models: `VideoCallInfo` and `AgoraChannel` - should we deprecate them as well?

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/VideoCallToken.kt
```diff
@@ -25,6 +25,10 @@ import androidx.compose.runtime.Immutable
  * @property agoraUid The Uid of Agora.
  * @property agoraAppId The App Id of Agora.
  */
+@Deprecated(
+    "This third-party library integration is deprecated. Contact to the support team for more information.",
```
VelikovPetar, line 29: Maybe we could say: "Contact the support team for more information." instead of "Contact to the support team for more information"?

VelikovPetar (approved):

PR #5746, opened 2025-04-24
[AND-469] Remove the divider of the last item in lists

andremion:
### 🎯 Goal

### 🛠 Implementation details

_Describe the implementation_

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

<details>

```
Provide the patch code here
```

</details>

### 🎉 GIF


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/Channels.kt
```diff
@@ -92,17 +92,20 @@ public fun Channels(
...
+            itemsIndexed(
                 items = channelItems,
-                key = ItemState::key,
-            ) { item ->
+                key = { _, item -> item.key },
+            ) { index, item ->
                 itemContent(item)
 
-                divider()
+                if (index < channelItems.lastIndex) {
```
VelikovPetar, line 101: Changes look good, I am just worried that this changes the default appearance of the `ChannelList` - can we really treat this change as a bug fix?

andremion: I believe so. The item divider should never be renderer below the last item. Some implementations do that just for simplicity, although this is not consistent between all Stream SDKs.

PR #5749, opened 2025-04-25
[AND-462] (compose): Replace VideoView with ExoPlayer + PlayerView.

VelikovPetar:
### 🎯 Goal
Introduces `media3.exoplayer` to the `compose` SDK to improve playing of video (and audio) attachments. 

### 🛠 Implementation details
- Adds `androidx.media3:media3-exoplayer` and `androidx.media3:media3-ui` dependencies. (There is also a compose artefact (`media3-ui-compose`), however it is in very early development, and doesn't provide all the options like the `media3-ui` does (fast forward/backward, speed controls, progress indicator...))
- Rework the `MediaGalleryVideoPage` to use `PlayerView` + `ExoPlayer` (instead of `VideoView`)
- Rework the `MediaAttachmentActivity` to user `PlayerView` + ExoPlayer (instead of `VideoView`)

### 🎨 UI Changes

_Add relevant screenshots_

<table>
<thead>
<tr>
<th>Scenario</th>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Video (MediaGalleryVideoPage)
</td>
<td>
<video src="https://github.com/user-attachments/assets/7b5d38c2-6086-49c3-af32-8a55e8dfdc44" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/8852011e-ebdf-4b7c-b33c-7c071b71c553" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
Audio (MediaPreviewActivity)
</td>
<td>
<video src="https://github.com/user-attachments/assets/4701eaf6-8c4f-455b-92d8-73def24a8150" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/6567e4be-5ae7-4734-9699-03294e310ec2" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Open compose sample app
2. Open channel with Video attachments
3. Play the video attachment (the performance should be much better than previously)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
```diff
@@ -277,180 +274,127 @@ private fun ErrorIcon(modifier: Modifier) {
...
+        }
     }
-
-    val contentView = remember {
-        val frameLayout = FrameLayout(context).apply {
-            layoutParams = FrameLayout.LayoutParams(
-                ViewGroup.LayoutParams.MATCH_PARENT,
-                ViewGroup.LayoutParams.MATCH_PARENT,
+    // Draw player
+    player?.let { preparedPlayer ->
```
andremion, line 328: Can we also show `LoadingIndicator` while player is null?
We could use the `matchParentSize` modifier inside that box so that the loading indicator respects the constraints of the `MediaGalleryVideoPage`'s modifier.
WDYT?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
```diff
@@ -277,180 +274,127 @@ private fun ErrorIcon(modifier: Modifier) {
...
+        Box(
+            modifier = modifier,
+            contentAlignment = Alignment.Center,
+        ) {
+            // Video player
+            AndroidView(
+                modifier = Modifier
+                    .fillMaxSize()
```
andremion, line 336: What about using `matchParentSize` instead?
So that the `AndroidView` respects the constraints of the `MediaGalleryVideoPage`'s modifier instead of forcing max size.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
```diff
@@ -277,180 +274,127 @@ private fun ErrorIcon(modifier: Modifier) {
...
+    showPlayButton: Boolean,
+    onPlayClick: () -> Unit,
+) {
+    val data = if (ChatTheme.videoThumbnailsEnabled) {
+        thumbnailUrl
+    } else {
+        null
+    }
+    Box(
+        modifier = Modifier.fillMaxSize(),
```
andremion, line 374: What about exposing the modifier and use `matchParentSize` in `MediaGalleryVideoPage`'s `Box`?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
```diff
@@ -178,92 +182,57 @@ public class MediaPreviewActivity : AppCompatActivity() {
      * Represents a video player with media controls.
      *
      * @param url The URL of the stream for playback.
-     * @param onBackPressed Handler for back press action.
      * @param onPlaybackError Handler for playback errors.
      */
+    @androidx.annotation.OptIn(UnstableApi::class)
```
andremion, line 187: nit: import `OptIn`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
```diff
@@ -178,92 +182,57 @@ public class MediaPreviewActivity : AppCompatActivity() {
...
-                layoutParams = FrameLayout.LayoutParams(
-                    ViewGroup.LayoutParams.MATCH_PARENT,
-                    ViewGroup.LayoutParams.MATCH_PARENT,
-                ).apply {
-                    gravity = Gravity.CENTER
-                }
+        var showBuffering by remember { mutableStateOf(false) }
+        val onBuffering: (Boolean) -> Unit = { showBuffering = it }
+        // Create player
+        var player by remember { mutableStateOf<Player?>(null) }
```
andremion, line 197: Since `MediaGalleryVideoPage` doesn't have page context anymore. What about renaming it and reuse it in `MediaPreviewContent`? So we avoid duplication.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
VelikovPetar: I think that this should work, I will test it and update it if everything's OK.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
VelikovPetar: Sounds good, let's do it that way

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
VelikovPetar: It is definitely possible, but in which case would such `LoadingIndicator` be visible? It will be between `onCreate` and `onResume` which would be a very short period of time.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
VelikovPetar: This actually is not directly possible, because `@androidx.annotation.OptIn` clashes with `@kotlin.OptIn` which is also used in the same class (line: 148). And for some reason, the both `OptIn` annotations seem to incompatible with their other target. Ex:
- `@kotlin.OptIn` doesn't work with `UnstableApi::class`
- `@androidx.annotation.OptIn` doesn't work with `ExperimentalMaterial3Api::class`

We can do it though with some import alias if you think it would make sense.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
VelikovPetar: I was also thinking the same thing, but the `MediaGalleryVideoPage` has some extra stuff (like the `VideoThumbnail`) and some differences in the behaviour (the `MediaGalleryVideoPage` doesn't use `playWhenReady`, and doesn't use `playerView.showController()`). I could try to extract the common view preparation logic, with some configuration passed from outside (`playWhenReady: Boolean`, `showControllerInitially: Boolean` etc), and then reuse the common composable in both places. WDYT?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
andremion: I'm not sure how long it takes to create that player in low-end devices.
The idea was to prevent any blinking.
I trust you to decide whether it is worth it or not.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
andremion: Yes. Something like a customizable internal `StreamMediaPlayer` should do the trick.

VelikovPetar: Let me then write some common implementation for both places, and I will update the PR!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
VelikovPetar: I will test it out on a low-end device and get back to you.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
VelikovPetar: I reworked the implementation a bit, please have a look: https://github.com/GetStream/stream-chat-android/pull/5749/commits/d5cc8fd08e438a864675b04e084683b3a0cfa258

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamVideoPlayerContent.kt
```diff
@@ -0,0 +1,255 @@
...
+@OptIn(UnstableApi::class)
+@Composable
+internal fun StreamVideoPlayerContent(
+    assetUrl: String?,
+    playWhenReady: Boolean,
+    showControllerInitially: Boolean,
+    onPlaybackError: () -> Unit,
+    modifier: Modifier = Modifier,
+    thumbnailUrl: String? = null,
+    thumbnailEnabled: Boolean = false,
```
andremion, line 83: A nullable `thumbnailUrl` isn't it enough to know whether the thumbnail is enabled?
Do we need the `thumbnailEnabled` parameter?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamVideoPlayerContent.kt
```diff
@@ -0,0 +1,255 @@
...
+ * @param playWhenReady Whether playback should start automatically after preparation.
+ * @param showControllerInitially Whether to show the player controls when first displayed.
+ * @param onPlaybackError Callback invoked when video playback encounters an error.
+ * @param modifier Modifier to be applied to the player container.
+ * @param thumbnailUrl Optional URL of the thumbnail image to display before playback starts.
+ * @param thumbnailEnabled Whether to show the thumbnail before playback starts.
+ */
+@OptIn(UnstableApi::class)
+@Composable
+internal fun StreamVideoPlayerContent(
```
andremion, line 76: This component isn't also used to play audio?
Maybe its name could be `StreamMediaPlayerContent`, WDYT?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamVideoPlayerContent.kt
```diff
@@ -0,0 +1,255 @@
...
+ * @param modifier Modifier to be applied to the thumbnail container.
+ */
+@Composable
+private fun VideoThumbnail(
+    thumbnailUrl: String?,
+    showPlayButton: Boolean,
+    onPlayClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    val data = if (ChatTheme.videoThumbnailsEnabled) {
```
andremion, line 172: Maybe this verification could be done by the composable that calls `VideoThumbnail` so this entire composable wouldn't be called if `ChatTheme.videoThumbnailsEnabled` is false.
WDYT?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
VelikovPetar: Here's a short demo done on Samsung A05:
<table>
<thead>
<tr>
<th>Audio</th>
<th>Video</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/d9df12c7-0f2e-4e98-9236-68915cab7061" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/0b21c951-feb4-4627-8bda-47eff8df8ac2" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>
I don't really think it is necessary, (for video it will never be visible, because we also show a thumbnail over it initially), but if you think it improves the user experience, we could add it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamVideoPlayerContent.kt
VelikovPetar: It is a bit different. For example, for video, even if the `thumbnailUrl` is null (not delivered from BE for example), we would still show a black background with the play button as a cover for the video before it is played.

andremion: Maybe the confusion here is the play button, which should not be part of the `VideoThumbnail`.
The black background is already out of the `VideoThumbnail` and the play button should also be out.
So if thumbnailUrl is null, we would still have play button over a black background.
Does it make sense?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
andremion: That's fine.
It's an internal API, so we can always improve it later if we see it's necessary.
Thank you 🙌🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamVideoPlayerContent.kt
VelikovPetar: But in the audio case, we don't show neither a thumbnail, nor a play button, and we need to have a way to distinguish it from the video case, where we show the play button (+ thumbnail or black surface).

andremion (approved):

PR #5757, opened 2025-04-30
[AND-498] Fix new token not sent with the request after refreshing it.

VelikovPetar:
### 🎯 Goal
The `AuthTokenInterceptor` detects when the BE returns a token expired error, and automatically calls the `TokenProvider` to fetch a new token. Afterwards, it re-performs the call, but still uses the old token, and NOT the refreshed one. This causes the re-performed call to always fail, even though we have a new valid token.

With this PR, we ensure that the newly fetched token is used for the repeated invocation of the failed call.

### 🛠 Implementation details
- Adds the newly fetched token from the `TokenProvider` as a header to the `Request` before re-performing it.

### 🧪 Testing
1. Use the app until the Stream token expires
2. Do some action that would trigger a BE call
3. You should observe only one failed call in the logs (the first 'Invalid token' response) instead of 2

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/interceptor/TokenAuthInterceptor.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/interceptor/TokenAuthInterceptor.kt
@@ -49,16 +49,18 @@ internal class TokenAuthInterceptor internal constructor(
 
             tokenManager.ensureTokenLoaded()
 
-            val request: Request = addTokenHeader(chain.request())
+            val request: Request = addTokenHeader(chain.request(), token = tokenManager.getToken())
             var response: Response = chain.proceed(request)
 
             if (!response.isSuccessful) {
                 val err = parser.toError(response)
                 if (err.serverErrorCode == ChatErrorCode.TOKEN_EXPIRED.code) {
                     tokenManager.expireToken()
-                    tokenManager.loadSync()
+                    val freshToken = tokenManager.loadSync()
                     response.close()
-                    response = chain.proceed(request)
+                    // Rebuild the request with the new token and retry the request.
+                    val requestWithFreshToken = addTokenHeader(chain.request(), token = freshToken)
+                    response = chain.proceed(requestWithFreshToken)
                 } else {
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/interceptor/TokenAuthInterceptor.kt
```diff
@@ -67,17 +69,15 @@ internal class TokenAuthInterceptor internal constructor(
...
-            request.newBuilder().header(AUTH_HEADER, token).build()
-        } catch (e: IllegalArgumentException) {
-            throw ChatRequestError(
-                "${ChatErrorCode.INVALID_TOKEN.description}: '$token'",
-                ChatErrorCode.INVALID_TOKEN.code,
-                -1,
-                e,
-            )
-        }
+    private fun addTokenHeader(request: Request, token: String): Request = try {
```
JcMinarro, line 72: Why do we need to pass the token as parameter?
Shouldn't it be already updated within the `tokenManager` and be accesible when `tokenManager.getToken()` is called? :thinking:

VelikovPetar: It wasn't needed, I reworked it a bit so it is more clear what is happening - when we were reading the token just by calling `tokenManager.getToken()` - I feel it wasn't really understandable that we actually expect the new token to be fetched and applied to the request. But ofc, we can revert it as it was before, and just call `addTokenHeader(Request)`

VelikovPetar: Ok, I will revert the change 👍

VelikovPetar: Updated here: https://github.com/GetStream/stream-chat-android/pull/5757/commits/06a359254d7a4cdd73119480d9db3e8a7cc10aa3

(I also had to update the `FakeTokenManager` to ensure the test works)

PR #5761, opened 2025-05-06
[AND-515] Fix the channel members count of message list header

andremion:
### 🎯 Goal

Fix the channel members count of message list header

### 🛠 Implementation details

Do not filter the current user out when building the message list header title

### 🎨 UI Changes

SDK | Before | After |
| --- | --- | --- |
| XML on Pixel4 API 30 | xml before | xml after |
| Compose on Pixel8 API 35 | compose before | compose after


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/MessageListHeader.kt
```diff
@@ -184,7 +184,7 @@ public fun DefaultMessageListHeaderCenterContent(
     onHeaderTitleClick: (Channel) -> Unit = {},
 ) {
     val title = when (messageMode) {
-        MessageMode.Normal -> ChatTheme.channelNameFormatter.formatChannelName(channel, currentUser)
+        MessageMode.Normal -> ChatTheme.channelNameFormatter.formatChannelName(channel = channel, currentUser = null)
```
JcMinarro, line 187: It is a breaking change.
Customer implement `ChatThem.channelNameFormatter.formatChannelName()` method, and they would be using the current user value. If we pass `null` by default, we are breaking this logic.

JcMinarro (changes requested):
I don't think it is a bug on the Message List Header.
The title should show "the name of the others", not your one.
Apart of that, this change introduce a new bug, when the DM is 1:1, you won't see the name of the other user, but the other user name plus your own name

JcMinarro:
[Screen_recording_20250506_192938.webm](https://github.com/user-attachments/assets/4a85a260-a97c-4abf-a6d8-5035033efaec)

andremion:
As per the discussion, no action needs to be taken here as it works as expected

PR #5764, opened 2025-05-07
[AND-505] Introduce ChannelInfoController

andremion:
### 🎯 Goal

### 🛠 Implementation details

_Describe the implementation_

### 🎉 GIF

### DIFF
--- a/stream-chat-android-core/src/main/java/io/getstream/chat/android/extensions/ChannelDataExtensions.kt
+++ b/stream-chat-android-core/src/main/java/io/getstream/chat/android/extensions/ChannelDataExtensions.kt
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2014-2025 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.extensions
+
+import io.getstream.chat.android.models.ChannelData
+
+/**
+ * Group channels are channels with more than 2 members or channels that are not distinct.
+ */
+public val ChannelData.isGroupChannel: Boolean
+    get() = memberCount > 2 || !isDistinct
+
+/**
+ * A distinct channel is a channel created for a particular set of users, usually for one-to-one conversations.
+ */
+public val ChannelData.isDistinct: Boolean
+    get() = id.startsWith("!members")
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
@@ -76,7 +76,7 @@ public class ChannelInfoController(
             .flatMapLatest { channelState ->
                 logger.d { "[onChannelState]" }
                 combine(
-                    channelState.channelData.onEach { logger.d { "[onChannelData] cid: ${it.cid}, name: ${it.name}" } },
+                    channelState.channelData.onEach { logger.d { "[onChannelData] name: ${it.name}" } },
                     channelState.members.onEach { logger.d { "[onMembers] size: ${it.size}" } },
                     channelState.muted.onEach { logger.d { "[onMuted] $it" } },
                     channelState.hidden.onEach { logger.d { "[onHidden] $it" } },
@@ -142,6 +142,7 @@ public class ChannelInfoController(
     }
 
     public fun expandMembers() {
+        logger.d { "[expandMembers]" }
         _state.updateContent { content ->
             content.copy(
                 members = content.members.copy(
@@ -152,6 +153,7 @@ public class ChannelInfoController(
     }
 
     public fun collapseMembers() {
+        logger.d { "[collapseMembers]" }
         _state.updateContent { content ->
             content.copy(
                 members = content.members.copy(
@@ -162,9 +164,11 @@ public class ChannelInfoController(
     }
 
     public fun renameChannel(name: String) {
+        logger.d { "[renameChannel] name: $name" }
         scope.launch {
             channelClient.updatePartial(set = mapOf("name" to name)).await()
                 .onError { error ->
+                    logger.e { "[renameChannel] error: ${error.message}" }
                     _events.tryEmit(
                         ChannelInfoEvent.RenameChannelError(message = error.message),
                     )
@@ -173,9 +177,11 @@ public class ChannelInfoController(
     }
 
     public fun muteChannel() {
+        logger.d { "[muteChannel]" }
         scope.launch {
             channelClient.mute().await()
                 .onError { error ->
+                    logger.e { "[muteChannel] error: ${error.message}" }
                     _events.tryEmit(
                         ChannelInfoEvent.MuteChannelError(message = error.message),
                     )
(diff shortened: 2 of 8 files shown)


stream-chat-android-core/src/main/java/io/getstream/chat/android/extensions/ChannelDataExtensions.kt
```diff
@@ -0,0 +1,31 @@
...
+ * Group channels are channels with more than 2 members or channels that are not distinct.
+ */
+public val ChannelData.isGroupChannel: Boolean
+    get() = memberCount > 2 || !isDistinct
+
+/**
+ * A distinct channel is a channel created for a particular set of users, usually for one-to-one conversations.
+ */
+public val ChannelData.isDistinct: Boolean
+    get() = id.startsWith("!members")
```
JcMinarro, line 31: Do we need those methods as public?
Can't we keep them internal?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
```diff
@@ -0,0 +1,420 @@
...
+    init {
+        @Suppress("OPT_IN_USAGE")
+        channelState
+            .flatMapLatest { channel ->
+                logger.d { "[onChannelState]" }
+                combine(
+                    channel.channelData.onEach { logger.d { "[onChannelData] name: ${it.name}" } },
+                    channel.members.onEach { logger.d { "[onMembers] size: ${it.size}" } },
+                    channel.muted.onEach { logger.d { "[onMuted] $it" } },
+                    channel.hidden.onEach { logger.d { "[onHidden] $it" } },
```
JcMinarro, line 103: Do we need this log on every update?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
```diff
@@ -0,0 +1,420 @@
...
+            )
+        }
+
+        requireCapability(
+            permission = { canLeaveChannel },
+            onError = onError,
+        ) {
+            scope.launch {
+                runCatching {
+                    requireNotNull(chatClient.getCurrentOrStoredUserId()) { "User not connected" }
```
JcMinarro, line 322: A user shouldn't be able to "leave a channel" if the user is not set (That is different that it is not connected)
I don't think this case needs to be handled, we can use the current user from the UserState and not adding extra code that won't be executer never

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
```diff
@@ -0,0 +1,420 @@
...
+            scope.launch {
+                channelClient.delete().await()
+                    .onSuccess { _events.tryEmit(ChannelInfoEvent.DeleteChannelSuccess) }
+                    .onError(onError)
+            }
+        }
+    }
+
+    private fun List<Member>.filterNotCurrentUser() =
+        filter { member -> member.user.id != chatClient.getCurrentOrStoredUserId() }
```
JcMinarro, line 363: `getCurrentOrStoredUserId()` is mainly used to obtain an userId even if the ChatClient is in a "not-ready" state.
We shouldn't abuse to use this method. I would prefer to not use this method at all

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
```diff
@@ -0,0 +1,420 @@
...
+
+    private fun requireCapability(
+        permission: ChannelInfoState.Content.Capability.() -> Boolean,
+        onError: (error: Error) -> Unit,
+        onSuccess: () -> Unit,
+    ) {
+        (_state.value as? ChannelInfoState.Content)?.capability?.let { capability ->
+            runCatching { require(permission(capability)) }
+                .onFailure { onError(Error.GenericError("User doesn't have permission")) }
```
JcMinarro, line 372: Can't we forward the previous error?
It is hiding the root cause of the failure

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/channel/info/ChannelInfoEvent.kt
```diff
@@ -0,0 +1,87 @@
...
+ * including success and error events for specific actions.
+ */
+@ExperimentalStreamChatApi
+public sealed interface ChannelInfoEvent {
+    /**
+     * Indicates an error occurred while renaming a channel.
+     *
+     * @param message The error message.
+     */
+    public data class RenameChannelError(val message: String) : ChannelInfoEvent
```
JcMinarro, line 34: Instead of a "error message", can't we provide more useful data here?
I assume this error is because some error on the API/Network/Whatever, and we should be able to provide a `Result.Failure` or something similar that help to understand why it failed

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
andremion: It was pretty handy when debugging. That's why I used `logger.d`

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
andremion: All the operations can be called by the consumer at any time, so it's not guarantee that there will be always a connected user. That's why `getCurrentOrStoredUserId` is nullable.
I also used `getCurrentOrStoredUserId`  because it was simple to mock it in tests.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
andremion: The goal was to make it unit-testable. The current `ChatClient` dependencies are really hard to unit-test because they are too coupled and requires a lot of integration.
Unit testing a class which depends on `ChatClient` should be easier.
I would love to hear your suggestions...

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
andremion: `permission` is a `ChannelInfoState.Content.Capability.() -> Boolean`
`require` is a Kotlin function which throw if the parameter is not true.
The thrown message is too generic: `Failed requirement`
The actual cause is due to the user not having permission enabled

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
andremion: Now, I understand what you mean. We should avoid using `getCurrentOrStoredUserId` because it might get the user from the cache even when it is not connected.
I will use `getCurrentUser` instead.
Thanks

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
andremion: I will use `getCurrentUser` instead 👍🏻

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoController.kt
andremion: I can review them after have it fully integrated with the view models and the UI 🙏🏻

JcMinarro (approved):

PR #5766, opened 2025-05-07
Feature/fallback avatar

JcMinarro:
### 🎯 Goal
The error state wasn't easy to customize on the `Avatar` and as a fallback showed "Initial letters" without options to change this behavior.

On this PR we introduce a new property to Avatar allowing a custom Painter to be used when there is an error fetching the picture and a new method for overriding the default `FallbackAvatar`

Fix: AND-524


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1756,12 +1808,36 @@ public interface ChatComponentFactory {
...
         )
     }
 
+    /**
+     * The default fallback avatar, which renders initials in a circle.
+     * It is used when the image URL is empty or there is an error loading the image.
+     */
+    @Composable
+    public fun FallbackAvatar(
+        initials: String,
```
VelikovPetar, line 1824: Thinking a bit ahead of time on this probably, but do you think it would make sense to also pass the original `imageUrl` to the `FallbackAvatar` (which failed to load)? 
A customer could potentially want to access it perhaps for retry-purposes, and introducing it later would be a breaking change. 
_(I don't think we need to introduce it immediately, just bringing it up as a discussion point)_

VelikovPetar (commented):
Left only a minor discussion point, otherwise looks good to me!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
JcMinarro: Good point.
I was thinking to pass it as well, but at the end, because it wasn't used I didn't.
Let me see if we can send it easily

VelikovPetar (approved):

PR #5772, opened 2025-05-13
[AND-543] POC for create local message.

VelikovPetar:
### 🎯 Goal
A customer wants to be able to optimistically send messages only in the local state/database, without sending the message to the Stream BE.

### 🛠 Implementation details
The PR introduces two new `ChatClient` operations:
- `ChatClient.createLocalMessage` - creates a message and stores it in the State/Offline plugin storage without calling `API.sendMessage`. (however it also takes care of submitting any attachments to the CDN)
- `ChatClient.updateLocalMessage` - updates a message from the State/Offline without calling `API.updateMessage`
- Disable the `MessageMenu` if the message is only local (doesn't make sense to perform operations over a message which is not 'sent' yet)

**These method are marked as experimental, and could be subject to change.**

### Example usage
1. Override send message logic
```kotlin
// Override MessageComposer sendMessage
MessageComposer(
  viewModel = messageComposerViewModel,
  onSendMessage = { message ->
    // override default send message logic
    // submit message only locally
    messageComposerViewModel.createLocalMessage(message) { result ->
      when (result) {
        is Result.Success -> {
          // Your BE call here
          val localMessage = result.value
          val sendMessageResult = customSendMessage(localMessage)
          if (sendMessageResult is Result.Failure) {
            // Update the local message to mark it as failed (if needed)
            messageComposerViewModel.updateLocalMessage(localMessage, withError = true)
          }
        }
        is Result.Error -> {
          // Creation of the message locally has failed
        }
      } 
    }
  }
)
```
2. Override resend logic from message menu
```kotlin
class CustomChatComponentFactory : ChatComponentFactory {

    @Composable
    override fun MessageMenu(
        modifier: Modifier,
        message: Message,
        messageOptions: List<MessageOptionItemState>,
        ownCapabilities: Set<String>,
        onMessageAction: (MessageAction) -> Unit,
        onShowMore: () -> Unit,
        onDismiss: () -> Unit,
    ) {
        val customOnMessageAction = { action: MessageAction ->
            // Custom logic for handling message actions
            when (action) {
                is Resend -> {
                    // Your custom resend logic here
                }
                else -> onMessageAction(action)
            }
        }
        super.MessageMenu(
            modifier = modifier,
            message = message,
            messageOptions = messageOptions,
            ownCapabilities = ownCapabilities,
            onMessageAction = customOnMessageAction,
            onShowMore = onShowMore,
            onDismiss = onDismiss
        )
    }
}
```


VelikovPetar:
We are currently not exposing this. We will come back to it if it is needed by a customer.

PR #5773, opened 2025-05-13
[AND-506] Introduce experimental `DirectChannelInfoScreen` and `GroupChannelInfoScreen`

andremion:
### 🎯 Goal

Introduce experimental `DirectChannelInfoScreen` and `GroupChannelInfoScreen` for displaying channel information and the available channel actions.

### 🛠 Implementation details

- Introduce `MessageListHeaderViewModel`, `MessageListHeaderViewState`, and `MessageListHeaderViewController` to decouple the states of `MessageListHeader` from the channel info screens.
- Update `ChannelInfoViewModelFactory` to create instances of `MessageListHeaderViewModel` in addition to `ChannelInfoViewModel`.

- Introduce `ChannelInfoViewAction` that represents user interactions within the channel information view to prevent passing many callbacks back and forth. `onViewAction` is now the entry point of view actions to `ChannelInfoViewController`

- Create a sealed interface `Error` to group all error events within `ChannelInfoViewEvent`. That helps when handling events in case of a separate flow for errors.

- Create a sealed interface `Navigation` to group all navigation events within `ChannelInfoViewEvent`. That helps when handling events in case of a separate flow for navigation.

- Create a sealed interface `Modal` to group all modal events within `ChannelInfoViewEvent` supporting confirmation modals for hide, leave, and delete channel actions.. XML expects modal dialog navigations as one-shot events. Compose can handle them as both. Modeling Modal events allows us to support both SDKs.

- Add ability to copy the user id to clipboard in the direct channel info.

- Model state to contain the available options based on capabilities instead of exposing the capabilities themselves.

- Add snapshot tests for direct and group channel info contents.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Phone</th>
<th>Tablet (with Adaptive layout enabled)</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/8de461ac-689e-4bd2-9cb5-17b8fc86b757" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/414d3120-04bd-4585-8375-bd20828d827c" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🎉 GIF

gif

### DIFF
--- a/stream-chat-android-ui-guides/detekt-baseline.xml
+++ b/stream-chat-android-ui-guides/detekt-baseline.xml
@@ -19,8 +19,6 @@
     <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoibGVhbmRybyJ9.CjlYUr79r4GopAhXIbqLBighl3meLsT4dQKzdKX7L3g"</ID>
     <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoibWFydG9uIn0.22wjzwYCNdaG5FLVeTD49NqVA11UJpEwrNRjZxZrcK8"</ID>
     <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoic2FtdWVsIn0.SusttZNc2Y0sc-JPEOPCmTa5FuKDHRcWGO_7kYrC1C0"</ID>
-    <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicWF0ZXN0MSJ9.H1nlYibjgp1HfaOd0sA_T4038tjsN61mJWxvUjmRQI0"</ID>
-    <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicWF0ZXN0MiJ9.GYp9ikLtU2eG9Mq7tmHThzbV7C8W82j18sExuO7-ogc"</ID>
     <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicmFmYWwifQ.7Y4QCvc42Km8ETLdCQT5ynjiKVbZZbuN0XTiGxJNU6k"</ID>
     <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoidG9tbWFzbyJ9.lNaWC2Opyq6gmV50a2BGxK-5gm5mwCpefnUA30_k9YA"</ID>
     <ID>MaxLineLength:LoginUsers.kt$LoginUsers$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoidGhpZXJyeSJ9.hZi4pBPt2v2HSoS-7Yn7Ll2a1twhs763MlRGFAday2c"</ID>
--- a/stream-chat-android-ui-uitests/detekt-baseline.xml
+++ b/stream-chat-android-ui-uitests/detekt-baseline.xml
@@ -17,11 +17,7 @@
     <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoiamFld29vbmcifQ.SoSQ1zkdpD-xzrdTnNioFOZQ5pKKcLJj4Jw8wAzOhno"</ID>
     <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoiamMifQ.WtWn2rgZyJNOpg48xUgYoMnG3BvMD5524RND7e7Mmhk"</ID>
     <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoibGVhbmRybyJ9.HGQfqwt1Jj2W25QQLO8hbnNeOrUl04q4a_AjHy0sFWM"</ID>
-    <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoibWFyaW4ifQ.jSLjDAcGcz7hHrDyIU8OaSgRDg63aw2cW4C1nqIsB5c"</ID>
-    <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicWF0ZXN0MSJ9.T-LLXoVmdX1Huxg-gjUnBJ7TEnlyGSfa9zFw-N50zLY"</ID>
-    <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicWF0ZXN0MiJ9.EVNBLb7-yR823jv2jtDcjYBQCxBAQnhLDNhpMjcJ1uI"</ID>
     <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicmFmYWwifQ.Fx_BMBwlwbfZzSonziklhGt0KWTKVnLRtXRI5RIuJKU"</ID>
-    <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoidG9taXNsYXYifQ.-n5L1OmG4KpL0O16h8SeqKJID0DzUmy_nilqTgi8EZQ"</ID>
     <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoidG9tbWFzbyJ9.1H7yz3CZ_wtSxj0B36tOkfdFtKzULSPwRyesufrTGpM"</ID>
     <ID>MaxLineLength:UserCredentials.kt$token = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoidGhpZXJyeSJ9.pfG5thPeT7ytMSExrg3yoXQ02AqrLj2gIxuBZkTlr_c"</ID>
     <ID>VariableNaming:ComponentBrowserSearchViewFragment.kt$ComponentBrowserSearchViewFragment$private val TAG = ComponentBrowserSearchViewFragment::class.simpleName</ID>


stream-chat-android-ui-common/src/main/res/values/strings_channel_info.xml
```diff
@@ -0,0 +1,50 @@
...
+    <string name="stream_ui_channel_info_expand_button">%d more</string>
+    <string name="stream_ui_channel_info_copy_user_handle">Copy user handle</string>
+    <string name="stream_ui_channel_info_name_field_label">NAME</string>
+    <string name="stream_ui_channel_info_name_field_placeholder">Add a group name</string>
+    <string name="stream_ui_channel_info_rename_group_error">Failed to rename group</string>
+    <string name="stream_ui_channel_info_option_mute_conversation">Mute User</string>
+    <string name="stream_ui_channel_info_option_mute_conversation_error">Failed to switch the conversation mute state</string>
+    <string name="stream_ui_channel_info_option_mute_group">Mute Group</string>
+    <string name="stream_ui_channel_info_option_mute_group_error">Failed to switch the group mute state</string>
+    <string name="stream_ui_channel_info_option_hide_conversation">Hide User</string>
```
VelikovPetar, line 29: The value seems wrong, I assume it should be `Hide conversation` instead of `Hide User`?

stream-chat-android-ui-common/src/main/res/values/strings_channel_info.xml
```diff
@@ -0,0 +1,50 @@
...
+-->
+<resources>
+    <string name="stream_ui_channel_info_member_owner">Owner</string>
+    <string name="stream_ui_channel_info_member_moderator">Moderator</string>
+    <string name="stream_ui_channel_info_expand_button">%d more</string>
+    <string name="stream_ui_channel_info_copy_user_handle">Copy user handle</string>
+    <string name="stream_ui_channel_info_name_field_label">NAME</string>
+    <string name="stream_ui_channel_info_name_field_placeholder">Add a group name</string>
+    <string name="stream_ui_channel_info_rename_group_error">Failed to rename group</string>
+    <string name="stream_ui_channel_info_option_mute_conversation">Mute User</string>
```
VelikovPetar, line 25: The value seems wrong, I assume it should be `Mute conversation` instead of `Mute User`?

stream-chat-android-ui-common/src/main/res/values/strings_channel_info.xml
```diff
@@ -0,0 +1,50 @@
...
+    <string name="stream_ui_channel_info_option_mute_conversation_error">Failed to switch the conversation mute state</string>
```
VelikovPetar, line 26: In some places like the `SelectedChannelMenu`, we are using the string "Channel" inside the options items (ex: "Mute channel"). In this new feature, we are using either "conversation" or "group". Not a big deal, but perhaps we should be consistent in the naming here (maybe use "Channel" everywhere)

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListHeaderViewController.kt
```diff
@@ -0,0 +1,75 @@
...
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.stateIn
+
+// The XML SDk has its own ViewModel for the MessageListHeaderView.
+// This class can be later shared after a refactor of the XML SDK.
+/**
+ * Controller responsible for managing the state of the message list header.
+ */
+@InternalStreamChatApi
+public class MessageListHeaderViewController(
```
VelikovPetar, line 42: Since we are introducing this, we should eventually rework the actual message list to also use this controller/viewModel. (something in the future). I am also thinking if maybe `ChannelHeaderController` would be a better name? (it fits better when used in the `MessageList` and the `ChannelInfoScreen`)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfoOption.kt
```diff
@@ -0,0 +1,186 @@
...
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import io.getstream.chat.android.compose.R
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.compose.ui.util.clickable
+
+@Composable
+internal fun ChannelInfoOption(
```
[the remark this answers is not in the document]
andremion: All the component factories will be done by another task to avoid an even more bigger PR. 🙏🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfoScreenModal.kt
```diff
@@ -0,0 +1,147 @@
...
+import io.getstream.chat.android.compose.R
+import io.getstream.chat.android.compose.ui.components.SimpleDialog
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.compose.ui.util.clickable
+import io.getstream.chat.android.ui.common.feature.channel.info.ChannelInfoViewAction
+import io.getstream.chat.android.ui.common.feature.channel.info.ChannelInfoViewEvent
+
+@Composable
+internal fun ChannelInfoScreenModal(
+    modal: ChannelInfoViewEvent.Modal?,
```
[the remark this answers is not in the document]
andremion: I think you already answered. 😅 
For example, in the next PR, I will create another modal which uses a `ModalBottomSheet`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/DirectChannelInfoScreen.kt
```diff
@@ -0,0 +1,242 @@
...
+ * @param viewModelFactory The factory used to create the [ChannelInfoViewModel].
+ * @param modifier The [Modifier] to be applied to this screen.
+ * @param viewModelKey Key to differentiate between instances of [ChannelInfoViewModel].
+ * @param onNavigationIconClick Callback invoked when the navigation icon is clicked.
+ * @param onPinnedMessagesClick Callback invoked when the pinned messages option is clicked.
+ * @param topBar The top bar to be displayed on the screen.
+ */
+@ExperimentalStreamChatApi
+@Composable
+public fun DirectChannelInfoScreen(
```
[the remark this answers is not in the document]
andremion: That's a valid point.
I looked at the iOS SDK and it handles separately.
I followed what we currently have and tried to not change that.
But it's worth to review and see if we can have it as single.
I wouldn't do it in this PR since it would change the current structure the sample.
We can brainstorm together maybe next week.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/GroupChannelInfoScreen.kt
```diff
@@ -0,0 +1,396 @@
...
+            Text(
+                text = user.getLastSeenText(LocalContext.current),
+                style = ChatTheme.typography.footnote,
+                color = ChatTheme.colors.textLowEmphasis,
+                maxLines = 1,
+                overflow = TextOverflow.Ellipsis,
+            )
+        }
+        val role = when (val role = member.role) {
+            is ChannelInfoViewState.Content.Role.Owner ->
```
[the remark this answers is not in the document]
andremion: That's why I had the `value` exposed.
But that is already refactored and now, the `Member` modal is exposed instead

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoViewController.kt
```diff
@@ -67,13 +71,15 @@ import kotlinx.coroutines.launch
...
 public class ChannelInfoViewController(
+    private val context: Context,
     cid: String,
     private val scope: CoroutineScope,
     private val chatClient: ChatClient = ChatClient.instance(),
     channelState: Flow<ChannelState> = chatClient
         .watchChannelAsState(cid = cid, messageLimit = 0, coroutineScope = scope)
         .filterNotNull(),
     private val channelClient: ChannelClient = chatClient.channel(cid),
+    private val clipboardManager: ClipboardManager = requireNotNull(context.getSystemService<ClipboardManager>()),
```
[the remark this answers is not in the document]
andremion: I would like to hear more about your concern.
What are the blockers for that?

stream-chat-android-ui-common/src/main/res/values/strings_channel_info.xml
andremion: iOS is using `User` and `Group`... 🤔 
Customers can override those strings in any case... 🤷🏻

andremion: I would avoid using channel everywhere tho.
Using different string keys make it easier for customers that just want to use different wording depending on the case...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/DirectChannelInfoScreen.kt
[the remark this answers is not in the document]
VelikovPetar: Got it, thanks for the explanation!
If that is the case then I would keep it as it as, let's be aligned with iOS as customers would expect a similar behaviour 👍

stream-chat-android-ui-common/src/main/res/values/strings_channel_info.xml
VelikovPetar: I find a bit confusing because I associate "Mute User" with muting the user - and not muting the conversation with the user. And we have different endpoints for mute channel vs mute user. We can of course keep the initial implementation as it can be easily customised, but I think the default strings might be a bit confusing.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListHeaderViewController.kt
andremion: ✅ https://github.com/GetStream/stream-chat-android/pull/5773/commits/3283a7e7ee5fe98b19a02055213fd2cb45fe8a77

stream-chat-android-ui-common/src/main/res/values/strings_channel_info.xml
andremion: But yes, between `User` and `Conversation` I also prefer the latter.

andremion: https://github.com/GetStream/stream-chat-android/pull/5773/commits/f4a905910b18172f71e48c2f5c512b43c9b9100a

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoViewController.kt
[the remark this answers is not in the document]
VelikovPetar: We would ideally have the classes with business logic as pure kotlin, perhaps it will be easier to migrate to KMP :) 
On a more realistic note, it would make testing easier, as now we would have to mock `context.getSystemService`. Also, I am generally against passing the context via a constructor to the ViewModels, because if not careful, we could introduce memory leaks (not the case here as far as I can see, but still it would be good to prevent it).

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListHeaderViewController.kt
VelikovPetar: Thank you!

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoViewController.kt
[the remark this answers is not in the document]
andremion: Thanks for the elaboration ❤️ 

Let's review some context about that class.

- The controller doesn't manage domain nor data related state.
- It is responsible for managing state of a view.
- It lives in the `ui-common` module.

Given that, we can consider that it is not a business logic class.

In the unit tests, we [ignore](https://github.com/GetStream/stream-chat-android/pull/5773/files#diff-03f1202d2b08f68863482443028e0c3687723177925f42fc62db5cf139549bb1R1075) the mocked `context` parameter and use the `clipboardManager` do setup what is expected from the mock.

Regarding the memory leak... I can push another layer of safety by making sure we only only send the application context 👍🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfoContentOption.kt
```diff
@@ -0,0 +1,147 @@
...
+    isGroupChannel: Boolean,
+    onViewAction: (action: ChannelInfoViewAction) -> Unit,
+    onPinnedMessagesClick: () -> Unit,
+) {
+    when (option) {
+        is ChannelInfoViewState.Content.Option.Separator -> {
+            StreamHorizontalDivider(thickness = 8.dp)
+        }
+
+        is ChannelInfoViewState.Content.Option.AddMember -> {
```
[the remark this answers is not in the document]
VelikovPetar: Thanks for the explanation, that makes sense. But considering that, can we really consider the `Separator` as a `ChannelInfoViewState.Content.Option`? All the others a specific actions that can be taken by the user, but the separator is only a visual component.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoViewController.kt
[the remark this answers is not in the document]
VelikovPetar: Still, wouldn't it be better if we at least pass the `clipboardManager` from outside, so that the controller/viewModel doesn't directly depend on the `Context`? In every other similar case we use the `context` to create the dependency derived from it, and inject that dependency in the VM/Controller via a constructor. (ClipboardHandler, StreamMediaRecorder...)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfoContentOption.kt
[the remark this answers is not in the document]
andremion: The idea was to make the UI dumb and not care about the position of the separator.
If we remove the separator from the view state, how easy for the UI would be to include that visual element?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoViewController.kt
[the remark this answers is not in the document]
andremion: https://github.com/GetStream/stream-chat-android/pull/5773/commits/b278d23e3c3ddd39a92825ff37bc3207eb8efbe7
Let me know what you think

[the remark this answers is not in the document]
VelikovPetar: Looks good, thank you for updating it!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/ChannelInfoContentOption.kt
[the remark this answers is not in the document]
VelikovPetar: That's a good point as well. I am thinking if it makes sense to have the concept of options groups, where we could group the different options (something like destructive vs non-destructive), but that is a more complicated approach. We can keep it as it is, and maybe see if there are better options for doing this when the customisation options are implemented.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/handlers/ClipboardHandler.kt
```diff
@@ -38,6 +43,11 @@ public fun interface ClipboardHandler {
  * @param clipboardManager System service that allows for clipboard operations, such as putting
  * new data on the clipboard.
  */
+@Deprecated(
+    message = "Use `ClipboardHandlerImpl` from `ui.common.helper` instead.",
+    replaceWith = ReplaceWith("ClipboardHandlerImpl", "io.getstream.chat.android.ui.common.helper"),
+    level = DeprecationLevel.WARNING,
+)
```
JcMinarro, line 50: Shouldn't we use the implementation of `io.getstream.chat.android.ui.common.helper.ClipboardHandlerImpl` here?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/ChannelHeaderViewState.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * The connection state of the chat client.
+     */
+    val connectionState: ConnectionState = ConnectionState.Offline,
+
+    /**
+     * The channel associated with the message list.
+     */
+    val channel: Channel = Channel(),
```
JcMinarro, line 42: Do we need this default value?
Does it make sense to create `ChannelHeaderViewState` with this default value? :thinking:

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/handlers/ClipboardHandler.kt
andremion: That interface and class are not used anywhere in the project and are duplicates of those in `io.getstream.chat.android.ui.common.helper`
That's why I didn't remove it, just marked it as deprecated.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/ChannelHeaderViewState.kt
andremion: That was to prevent nullables, as the legacy components expect non-nulls.

Screenshot 2025-05-22 at 08 28 28

The goal of the channel header view state/controller was not to refactor the legacy components, but to decouple the state from the other view controllers.

JcMinarro: What I mean is, Do we need to init it with a default value?
We can keep `channel` as non-nullable property, but asking to be initialized outside, not by this default argument

PR #5775, opened 2025-05-14
POC for overriding `sendMessage` operation

VelikovPetar:
### 🎯 Goal
The PR aims to provide a way for customers to override the default `sendMessage` operation. 
Example use-case: replace the calls to send messages to the Stream server, and send them to your own server instead (where they can later be synced with the Stream server)

Resolves: https://linear.app/stream/issue/AND-543/subito-allow-sending-local-message

### 🛠 Implementation details
- [public] Introduce `SendMessageInterceptor` which can be implemented to provide custom logic for uploading a message to the server.
- [public] Introduce `ChatClient.Builder` method: `sendMessageInterceptor` to register custom `SendMessageInterceptor`
- [internal] Introduce `ProxyChatApi` serving as a coordinator for delegating/overriding the Stream API operations.

### 🧪 Testing
Example usage:
```kotlin
// Define custom SendMessageInterceptor
@OptIn(ExperimentalStreamChatApi::class)
class CustomSendMessageInterceptor : SendMessageInterceptor {

    override fun sendMessage(channelType: String, channelId: String, message: Message): Result<Message> {
        return runBlocking {
            sendMessageToOwnServer(channelType, channelId, message)
        }
    }

    private suspend fun sendMessageToOwnServer(
        channelType: String,
        channelId: String,
        message: Message,
    ): Result<Message> {
        // Send message to your own server
        // ....
        // Wrap the result in a Result object
        return Result.Success(message)
    }
}

// Register with ChatClient.Builder
val client = ChatClient.Builder(apiKey, context)
            .withPlugins(offlinePlugin, statePluginFactory)
            .sendMessageInterceptor(CustomSendMessageInterceptor())
            .build()
```


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/interceptor/SendMessageInterceptor.kt
```diff
@@ -0,0 +1,47 @@
...
+     * or a failure if the upload had failed.
+     *
+     * @see [Result.Success]
+     * @see [Result.Failure]
+     */
+    public fun sendMessage(
+        channelType: String,
+        channelId: String,
+        message: Message,
+    ): Result<Message>
```
nuno-vieira, line 46: What about skip push etc? Should we add an "options" paremeter, so that we include extra stuff. Since if we need to pass more info in the future, it will break the interface, and so be a breaking change

VelikovPetar: The `skipPush` (and `skipEnrichUrl`) are currently handled inside the `Message` model itself. The `Message` model has both flags as properties, and we extract them before sending them to the BE. Here's how this is currently done in the 'default' implementation:
```kotlin
    // MoshiChatApi
    override fun sendMessage(
        channelType: String,
        channelId: String,
        message: Message,
    ): Call<Message> = messageApi.sendMessage(
        channelType = channelType,
        channelId = channelId,
        message = SendMessageRequest(
            message = with(dtoMapping) { message.toDto() },
            skip_push = message.skipPushNotification,
            skip_enrich_url = message.skipEnrichUrl,
        ),
    ).mapDomain { response ->
        response.message.toDomain()
    }
```

The public api for sending messages (`ChatClient.sendMessage`) also has the same contract, accepting only `channelType`, `channelId` and `Message` object (which holds the `skipPushNotification` and the `skipEnrichUrl` flags)

nuno-vieira: Ah I see, in that case it should be fine I guess 👍

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/ProxyChatApiTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        // Given
+        val channelType = randomString()
+        val channelId = randomString()
+        val message = randomMessage()
+        val originalApi = mock<ChatApi>()
+        whenever(originalApi.sendMessage(any(), any(), any()))
+            .doReturn(RetroSuccess(message).toRetrofitCall())
+        val proxyApi = ProxyChatApi(originalApi, UserScope(ClientScope()), null)
+        // When
+        proxyApi.sendMessage(channelType, channelId, message).await()
```
JcMinarro, line 49: SHould we verify the returned result is the one is expected depending on the collaborator is used? :thinking:

JcMinarro (approved):
Looks Good

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api/ProxyChatApiTest.kt
VelikovPetar: Added the additional assertions: https://github.com/GetStream/stream-chat-android/pull/5775/commits/f7cf8497c22081eabf4eaa43f2168ac0579e7c4f

PR #5784, opened 2025-05-21
[AND-550] Add pending messages support in LLC.

VelikovPetar:
### 🎯 Goal
Introduces a way to access the pending messages in the channel for the current user. Additionally, exposes an operation to fetch a pending message (and its metadata) by its ID.
Docs: https://getstream.io/chat/docs/react/pending_messages/

### 🛠 Implementation details
- Add `pendingMessages` field to `Channel`: Holds the pending messages for the current user in the given channel.
- Introduce `ChatClient.getPendingMessage` method to fetch a pending message and its metadata by its IDs. **Note: This operation uses the same endpoint as the getMessage operation. I introduced a new method, because it is not possible to return the pending message metadata without a breaking change on the getMessage method. I also wanted to avoid introducing a new filed in the Message object, as the metadata doesn't really belong to it. This would potentially be a discrepancy between different SDKs, but I wasn't able to find a different way to do it without a breaking change.**
- Add `markMessagesPending` flag to the channel `Config` field - indicator if pending messages are enabled for the channel. 

**Note: These additions are not used internally, they are just exposed for client integration.**

### 🎨 UI Changes
_NA_

### 🧪 Testing
1. Enable pending messages in the App dashboard (or contact me to provide a test user/apikey)
2. Send a message
3. Call queryChannels
4. Observe the response field: `result.channels[x].pendingMessage` -> It should contain all pending messages for the user in the channel
5. Observe the response field: `result.channels[x].config.markMessagesPending` -> should be true otherwise false based on the dashboard config


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
```diff
@@ -145,6 +148,18 @@ internal class DomainMappingTest {
...
+    @Test
+    fun `DownstreamPendingMessageDto is correctly mapped to PendingMessage`() {
+        val downstreamPendingMessageDto = randomDownstreamPendingMessageDto()
+        val sut = Fixture().get()
+        val expected = PendingMessage(
+            message = with(sut) { downstreamPendingMessageDto.message.toDomain() },
+            metadata = downstreamPendingMessageDto.metadata.orEmpty(),
+        )
+        val result = with(sut) { downstreamPendingMessageDto.toDomain() }
+        Assertions.assertEquals(expected, result)
```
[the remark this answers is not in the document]
VelikovPetar: Intentional change, we are migrating away from `kluent` to `Jupiter`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -342,6 +343,15 @@ constructor(
         }
     }
 
+    override fun getPendingMessage(messageId: String): Call<PendingMessage> {
+        return messageApi.getMessage(messageId).mapDomain {
+            PendingMessage(
+                message = it.message.toDomain(),
+                metadata = it.pending_message_metadata.orEmpty(),
```
andremion: I see you mentioned not introducing a new field on the message object.
I'm just trying to understand why not.
Given that the customer would send a pending message ID to get an object that contains metadata, why not have a metadata field in the message result 🤔

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/DistinctChatApiEnabler.kt
```diff
@@ -62,6 +63,10 @@ internal class DistinctChatApiEnabler(
         return getApi().getMessage(messageId)
     }
 
+    override fun getPendingMessage(messageId: String): Call<PendingMessage> {
```
andremion, line 66: Other platforms support retrieving multiple messages by their IDs.
Don't we have plans to also support it?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/ChatClientMessageApiTests.kt
```diff
@@ -93,6 +95,34 @@ internal class ChatClientMessageApiTests : BaseChatClientTest() {
         verify(plugin).onGetMessageResult(messageId, result)
     }
 
+    @Test
+    fun getPendingMessageSuccess() = runTest {
+        // given
+        val messageId = randomString()
+        val pendingMessage = randomPendingMessage()
+        val sut = Fixture()
+            .givenGetPendingMessageResult(RetroSuccess(pendingMessage).toRetrofitCall())
```
andremion: Small suggestion to use a helper extension function.

```suggestion
            .givenGetPendingMessageResult(pendingMessage.asCall())
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/DistinctChatApiEnabler.kt
VelikovPetar: I wouldn't say it is directly part of the "Pending messages" feature - but I can implement it separately to align it with the other SDKs.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
VelikovPetar: I didn't want to add it as part of the `Message` object, because it is a field which will never be present, unless the user explicitly tries to retrieve a pending message. So I didn't want to introduce a confusing (and not commonly used) field in an already complex model.

andremion: If I got it right, according to the other platforms, there's no specific `getPendingMessages` function.
I assume returning the pending metadata is solely based on the fact that a message might contain or not contain the metadata.
It's based on the message content (which is retrieved by an ID), not by any other explicit intention/API.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -342,6 +343,15 @@ constructor(
...
+            )
```
JcMinarro: Could we create a mapper function for it?
Then it can be reused on different methods

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
JcMinarro, line 178: I am a bit concern how we can improve those tests....
Because shouldn't test our implementation with the same implementation...
I mean, we are testing `sut` instance, but at the same time it is being used to obtain the expected value...
I know that it is out of the scope of this PR, and it is how we already did on other parts of our tests, but we should find an alternative for it. Coding it manually could be a tedious job, but maybe a more sense test (Even if it could skip some parsing steps)

I say that because if we introduce a bug in our mapper functions, as far as we are using the same mapper function to build the "expected result" we are not going to be covered by this test...

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/PendingMessage.kt
```diff
@@ -0,0 +1,86 @@
...
+     * @return A Builder instance for creating a modified copy of this PendingMessage.
+     */
+    @SinceKotlin("99999.9")
+    @Suppress("NEWER_VERSION_IN_SINCE_KOTLIN")
+    public fun newBuilder(): Builder = Builder(this)
+
+    /**
+     * Builder class for constructing a [PendingMessage] instance.
+     */
+    public class Builder() {
```
JcMinarro, line 44: Does it make sense to have an empty builder constructor? :thinking: 
This builder was mainly though for Java User that doesn't have an easy way to call `sometthing.copy(...)`
They shouldn't build a instance from scratch.
Even if they want to build an instance from scratch, they can use the "insance constructor" with its defaults arguments instead of using the Builder class

VelikovPetar: There is already an existing `constructor(pendingMessage: PendingMessage)`, but we also provide similar empty constructors for other models (see `Message`, `Channel`, `User`). Wouldn't it be better to be aligned with the rest of the models?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
VelikovPetar: Yes this is a good point, we should probably re-think this test class. 
But then again, each `DomainMapper.toDomain()` should already have a separate test for itself, so if we break something in the mapping logic, it should be detected by the test specific to that method.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
VelikovPetar: I will first circle back to this comment: https://github.com/GetStream/stream-chat-android/pull/5784#discussion_r2168675638

Do you think we should keep the `PendingMessage` model separate from the `Message`, or to completely remove the `PendingMessage` model, and use `Message` with a new field: `pendingMessageMetadata`?

VelikovPetar:
Moved the PR to **Draft** - there are some newer changes/requirement that need to be done.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
JcMinarro: I think keeping it as a new model is better, adding a new field that will be almost always empty/null could be a bit confusing

VelikovPetar:
> Should we keep "pending messages" within our local db as well??

I was considering this, but given the fact that currently the `Channel.pendingMessages` field is not used in the state/offline layers, and is just exposed on the `Channel` model from the LLC - I am not sure if it makes too much sense to also store them in the DB, as they will never be accessed. 
I think it will just introduce overhead if we store the pending messages but never read them (of course, this will need to change if we eventually add specific state/ui handling of pending messages)

CC @JcMinarro

JcMinarro (approved):

PR #5788, opened 2025-05-22
Improve Localazy integration

JcMinarro:
### 🎯 Goal
On this PR we are unifying all strings on a sincle `string.xml` file per module.
The Localazy config has been update to achieve this new behavior and the Github Actions related with Localazy has been update 

Complete: AND-565


.github/workflows/localazy-download.yml
```diff
@@ -5,19 +5,23 @@ on:
...
+
 jobs:
   localazy-download:
     name: Download strings from Localazy
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3.1.0
+      - uses: actions/checkout@v4.2.2
+        with:
+          fetch-depth: 0
```
andremion, line 17: Do we need to fetch all history for all branches and tags?

.github/workflows/localazy-upload.yml
```diff
@@ -4,18 +4,20 @@ on:
...
 jobs:
   localazy-upload:
     name: Upload strings to Localazy
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v1
-      - uses: GetStream/android-ci-actions/actions/setup-java@main
+      - uses: actions/checkout@v4.2.2
+        with:
+          fetch-depth: 0
```
andremion, line 20: Do we need to fetch all history for all branches and tags?

PR #5791, opened 2025-05-26
Add File attachment composable methods to ChatComponentFactory

VelikovPetar:
### 🎯 Goal
Expose file attachments related composables as overridable methods in the `ChatComponentFactory`.
Resolves: https://linear.app/stream/issue/AND-574/request-to-add-fileattachmentcontent-and-fileattachmentitem-to

### 🛠 Implementation details
- Expose `FileAttachmentPreviewContent`, `FileAttachmentContent` and `FileAttachmentItem` as overridable methods in the `ChatComponentFactory`
- Expose `FileUploadContent` and `FileUploadItem` as overridable methods in the `ChatComponentFactory`
- Add snapshot tests for `FileAttachmentFactory`
- Add snapshot tests for `UploadAttachmentFactory`

### 🎨 UI Changes
_NA_

### 🧪 Testing
1. Override `FileAttachmentPreviewContent`/`FileAttachmentContent`/`FileAttachmentItem`/`FileUploadContent` /`FileUploadItem`
2. Observe the changes when uploading a file attachment


stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/factory/FileAttachmentFactoryTest.kt
```diff
@@ -0,0 +1,86 @@
...
+import io.getstream.chat.android.compose.state.messages.attachments.AttachmentState
+import io.getstream.chat.android.compose.ui.SnapshotTest
+import io.getstream.chat.android.compose.ui.attachments.AttachmentFactory
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.models.Attachment
+import io.getstream.chat.android.previewdata.PreviewMessageData
+import org.junit.Rule
+import org.junit.Test
+
+internal class FileAttachmentFactoryTest : SnapshotTest {
```
andremion: We already have `AttachmentsContentTest`
Maybe you could enrich the existing tests instead of creating `FileAttachmentFactoryTest` and `UploadAttachmentFactoryTest`.
WDYT?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -2833,4 +2835,111 @@ public interface ChatComponentFactory {
...
+            attachments = attachments,
+            onAttachmentRemoved = onAttachmentRemoved,
+        )
+    }
+
+    /**
+     * Composable function for rendering the content of a file attachment.
+     *
+     * This function delegates the rendering to the default implementation of
+     * [FileAttachmentContent] provided by the Stream Chat Compose library.
```
andremion: I'm not sure if we should mention `This function delegates the rendering to...`
No other kdoc here mentions that.
Maybe just `Factory method for creating the content of file attachments` would be enough.
WDYT?

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/factory/FileAttachmentFactoryTest.kt
VelikovPetar: Oh I missed that. Yes, I will rework the tests.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
VelikovPetar: Will update it!

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/factory/FileAttachmentFactoryTest.kt
VelikovPetar: Just a follow up point: The new tests are covering the `AttachmentFactory.content()` methods, so they would also ensure that we don't make accidental changes in the `Factory.content()` methods, not just in the composables they are using. Perhaps we can keep these tests as well, but also enrich the existing tests in the `AttachmentsContentTest ` with the missing content composables?

andremion: I think those factories don't need to be tested as snapshot tests, since they just call the component factory
```kotlin
    previewContent = @Composable { modifier, attachments, onAttachmentRemoved ->
        ChatTheme.componentFactory.FileAttachmentPreviewContent(
            modifier = modifier,
            attachments = attachments,
            onAttachmentRemoved = onAttachmentRemoved,
        )
    },
```
It makes sense to have snapshot tests for the `public fun FileAttachmentPreviewContent` only.
I can also see that `FileAttachmentPreviewContent` doesn't have a preview that shares the content with the snapshot tests, as we have in `FileAttachmentContent`. Maybe it would be good to do the same.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/content/AttachmentsContentTest.kt
```diff
@@ -32,6 +32,13 @@ internal class AttachmentsContentTest : SnapshotTest {
         renderingMode = SessionParams.RenderingMode.SHRINK,
     )
 
+    @Test
+    fun `file attachment preview content`() {
+        snapshotWithDarkMode {
```
andremion: Can you use `snapshotWithDarkModeRow` instead?
So that the snapshot is gonna be smaller
Screenshot 2025-06-12 at 12 59 27

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/content/AttachmentsContentTest.kt
```diff
@@ -76,4 +83,11 @@ internal class AttachmentsContentTest : SnapshotTest {
             MediaAttachmentQuotedContent()
         }
     }
+
+    @Test
+    fun `file upload content`() {
+        snapshotWithDarkMode {
```
andremion: Can you use `snapshotWithDarkModeRow` instead?
So the snapshot is gonna be smaller.
Screenshot 2025-06-12 at 12 59 23

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/content/AttachmentsContentTest.kt
VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-android/pull/5791/commits/e2b1d1eb5ee9a4d907fa9d3d2f091618fd3efd00

andremion (approved):

PR #5797, opened 2025-05-28
Message members from the member bottom sheet of channel info screen

andremion:
### 🎯 Goal

Support messaging a member from the group channel info screen

### 🛠 Implementation details

Messaging a member from the group channel info screen can enter two possible funnels:
- When there is an existing direct channel between the current user and the clicked member, the app navigates directly to the messages screen.
- When there is no direct channel for those users, the app opens a draft channel screen, and after sending the first message, the channel is converted to a non-draft channel and navigates to the messages screen.

The draft view controller and its state, and events are part of the `ui-common`, so it can be reused by the `ui-components` SDK. The draft screens and view model are not.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Starting a new chat</th>
<th>Messaging an existing chat</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/f97c74eb-c295-40b1-ad56-2c2d86f3875e" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/958494c7-1d1b-4c6d-b70e-d795ffd0957d" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

- Open a group chat
- Click on the top bar title 
- Click on a member
- Click on the `Message` option in the member bottom sheet

### 🎉 GIF

gif

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/draft/DraftChannelActivity.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/draft/DraftChannelActivity.kt
@@ -1,0 +1,92 @@
+/*
+ * Copyright (c) 2014-2025 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.compose.sample.feature.channel.draft
+
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import android.widget.Toast
+import androidx.activity.compose.setContent
+import androidx.activity.viewModels
+import androidx.compose.foundation.layout.systemBarsPadding
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.Modifier
+import io.getstream.chat.android.compose.sample.R
+import io.getstream.chat.android.compose.sample.ui.BaseConnectedActivity
+import io.getstream.chat.android.compose.sample.ui.MessagesActivity
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.ui.common.feature.channel.draft.DraftChannelViewEvent
+import kotlinx.coroutines.flow.collectLatest
+
+class DraftChannelActivity : BaseConnectedActivity() {
+
+    companion object {
+        private const val KEY_MEMBER_IDS = "memberIds"
+
+        /**
+         * Creates an [Intent] for starting the [DraftChannelActivity].
+         *
+         * @param context The calling [Context], used for building the [Intent].
+         * @param memberIds The list of member IDs to be used for creating the channel.
+         */
+        fun createIntent(context: Context, memberIds: List<String>) =
+            Intent(context, DraftChannelActivity::class.java)
+                .putExtra(KEY_MEMBER_IDS, memberIds.toTypedArray())
+    }
+
+    private val viewModelFactory by lazy {
+        DraftChannelViewModelFactory(
+            memberIds = requireNotNull(intent.getStringArrayExtra(KEY_MEMBER_IDS)).toList(),
+        )
+    }
+    private val viewModel by viewModels<DraftChannelViewModel> { viewModelFactory }
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContent {
+            ChatTheme {
+                DraftChannelScreen(
+                    modifier = Modifier.systemBarsPadding(),
+                    viewModel = viewModel,
+                    onNavigationIconClick = ::finish,
+                )
+                LaunchedEffect(viewModel) {
+                    viewModel.events.collectLatest { event ->
+                        when (event) {
+                            is DraftChannelViewEvent.NavigateToChannel -> {
+                                startActivity(
+                                    MessagesActivity.createIntent(
+                                        context = applicationContext,
+                                        channelId = event.cid,
+                                    ),
+                                )
+                                finish()
+                            }
+
+                            is DraftChannelViewEvent.DraftChannelError ->
+                                Toast.makeText(
+                                    applicationContext,
+                                    R.string.draft_channel_error,
+                                    Toast.LENGTH_SHORT,
+                                ).show()
+                        }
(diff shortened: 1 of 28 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/draft/DraftChannelViewController.kt
```diff
@@ -0,0 +1,128 @@
...
+    private suspend fun createDraftChannel(onSuccess: suspend (channel: Channel) -> Unit = {}) {
+        logger.d { "[createDraftChannel] memberIds: $memberIds" }
+
+        val onError: (Error) -> Unit = { error ->
+            logger.e { "[createDraftChannel] error: $error" }
+            _events.tryEmit(DraftChannelViewEvent.DraftChannelError)
+        }
+
+        runCatching {
+            requireNotNull(chatClient.getCurrentUser()?.id) { "User not connected" }
```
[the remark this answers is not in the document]
andremion: `runCatching` is used for that purpose. In case of error, an event is emitted.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/draft/DraftChannelViewController.kt
```diff
@@ -0,0 +1,128 @@
...
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+/**
+ * Controller responsible for managing the state and events related to the draft channel.
+ */
+@InternalStreamChatApi
+public class DraftChannelViewController(
```
VelikovPetar, line 43: I am a bit confused about the exposure of this class. It is currently marked as `InternalStreamChatApi` so it is not public, but it is only used in `DraftChannelViewModel`, which is not part of the SDK, but rather of the sample app. 

So customers should not be able to use this class on their own, but if we don't provide any UI/VMs, they wouldn't be able to use it all. 

I am wondering if we should even expose this controller / logic from the SDK? Perhaps the logic for creating a draft channel is better to be handled on app side only? Also considering the fact that the default `createChannel` logic from this controller might not be configurable enough for customers.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/draft/DraftChannelActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.Modifier
+import io.getstream.chat.android.compose.sample.R
+import io.getstream.chat.android.compose.sample.ui.BaseConnectedActivity
+import io.getstream.chat.android.compose.sample.ui.MessagesActivity
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.ui.common.feature.channel.draft.DraftChannelViewEvent
+import kotlinx.coroutines.flow.collectLatest
+
+class DraftChannelActivity : BaseConnectedActivity() {
```
VelikovPetar, line 35: Some small info: in the `AddChannelViewModel` (from the sample apps), we also have the concept of creating a `draft` channel. It is not really the same case as here, but maybe it is something that can be aligned.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/draft/DraftChannelViewController.kt
andremion: The intention behind the `DraftChannelViewController` was to follow the approach of a reusable view controller for XML and Compose SDKs, similar to `MessageComposerController`, `PinnedMessageListController`, etc.
In fact, it's only used by the sample for now. However, it's made for reuse in both SDKs.
The logic for creating a draft channel can still be done by the app.
The `ChannelInfoViewModel` triggers a `ChannelInfoViewEvent.NavigateToDraftChannel` event, which can be handled by the app.
We could improve the customization options for creating a draft channel when we move its UI to the public SDK.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/draft/DraftChannelActivity.kt
andremion: Thanks for the info.
I believe the `AddChannelViewModel` components have a wider scope, as they also handle searching and selecting members.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/draft/DraftChannelActivity.kt
```diff
@@ -0,0 +1,92 @@
...
+         * @param memberIds The list of member IDs to be used for creating the channel.
+         */
+        fun createIntent(context: Context, memberIds: List<String>) =
+            Intent(context, DraftChannelActivity::class.java)
+                .putExtra(KEY_MEMBER_IDS, memberIds.toTypedArray())
+    }
+
+    private val viewModelFactory by lazy {
+        DraftChannelViewModelFactory(
+            memberIds = requireNotNull(intent.getStringArrayExtra(KEY_MEMBER_IDS)).toList(),
```
harrytmthy, line 53: Hi @andremion, this line will throw an `IllegalArgumentException` if memberIds are missing:

```kotlin
requireNotNull(intent.getStringArrayExtra(KEY_MEMBER_IDS)).toList()
```

This can happen in edge cases such as:
- OS-triggered process death due to low memory, or
- Activity being recreated without preserved intent extras.

#### Suggestion 1: Immediate error handling
You can make this more robust by adding an early return or fallback:

```kotlin
val memberIds = intent.getStringArrayExtra(KEY_MEMBER_IDS)?.toList()
if (memberIds.isNullOrEmpty()) {
    finish() // Or show a Toast / log a warning
    return
}
```

#### Suggestion 2: Persist across process death
For more resilience, especially on process recreation, consider persisting the data using `onSaveInstanceState()` or `SavedStateHandle`:

```kotlin
// Save to outState
override fun onSaveInstanceState(outState: Bundle) {
    super.onSaveInstanceState(outState)
    outState.putStringArray(KEY_MEMBER_IDS, memberIds.toTypedArray())
}

// Then retrieve like this:
val memberIds = savedInstanceState?.getStringArray(KEY_MEMBER_IDS)
    ?: intent.getStringArrayExtra(KEY_MEMBER_IDS)
```

Hope this helps!

andremion: Thanks for the suggestion @harrytmthy-dev
In case of the `KEY_MEMBER_IDS` `Intent` argument, which is a primitive data type, it is automatically restored in both of the scenarios you described.
That is enough for a sample app.

PR #5801, opened 2025-06-02
Refactor the XML sample view models and UI components to use ChannelInfoViewController

andremion:
### 🎯 Goal

Refactor the channel info XML sample view models and UI components to use `ChannelInfoViewController`

### 🛠 Implementation details

Remove the legacy channel info view models in favor of using the new `ChannelInfoViewController` and `ChannelInfoMemberViewController` via view model wrappers.

#### Direct channel info
https://github.com/user-attachments/assets/5da96e6b-a904-44f9-9238-8371b7e44c51

#### Group channel info
https://github.com/user-attachments/assets/2f00e42d-2843-4be5-a162-1aa2a06faacd

### 🎉 GIF

gif


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/SingleLiveEvent.kt
```diff
@@ -0,0 +1,60 @@
...
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+/**
+ * A lifecycle-aware observable that sends only new updates after subscription.
+ * Used for events like navigation.
+ *
+ * This avoids emitting values again on configuration changes.
+ */
+internal class SingleLiveEvent<T> : MutableLiveData<T>() {
```
VelikovPetar, line 35: Short info - we already have the concept of `Event` + `EventObservable` which is quite similar to this. While I prefer this new approach, perhaps it will be more clear if we stay consistent.

stream-chat-android-ui-components-sample/src/main/res/navigation/main_nav_graph.xml
```diff
@@ -234,15 +234,14 @@
...
-            app:nullable="true"
             />
 
-        <argument
-            android:name="userData"
-            android:defaultValue="@null"
-            app:argType="io.getstream.chat.ui.sample.feature.chat.info.UserData"
-            app:nullable="true"
-            />
+<!--        <argument-->
```
VelikovPetar: Can we delete this commented out code? (and on line 406)

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@
 ### ✅ Added
```
VelikovPetar, line 64: We should also probably list the new public components here (the new ViewModelFactories)

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatViewModelFactory.kt
```diff
@@ -14,17 +14,25 @@
...
-import io.getstream.chat.ui.sample.feature.chat.info.group.GroupChatInfoViewModel
+import io.getstream.chat.android.core.ExperimentalStreamChatApi
 import io.getstream.chat.ui.sample.feature.chat.info.group.users.GroupChatInfoAddUsersViewModel
 
-class ChatViewModelFactory(private val cid: String) : ViewModelProvider.Factory {
+class ChatViewModelFactory(private val context: Context, private val cid: String) : ViewModelProvider.Factory {
     private val factories: Map<Class<*>, () -> ViewModel> = mapOf(
         ChatViewModel::class.java to { ChatViewModel(cid) },
-        GroupChatInfoViewModel::class.java to { GroupChatInfoViewModel(cid) },
+        // ChannelInfoViewModel::class.java to {
```
VelikovPetar: Can we remove this commented code?

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/ChatViewModelFactory.kt
```diff
@@ -14,17 +14,25 @@
...
 
+import android.content.Context
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
-import io.getstream.chat.ui.sample.feature.chat.info.group.GroupChatInfoViewModel
+import io.getstream.chat.android.core.ExperimentalStreamChatApi
 import io.getstream.chat.ui.sample.feature.chat.info.group.users.GroupChatInfoAddUsersViewModel
 
-class ChatViewModelFactory(private val cid: String) : ViewModelProvider.Factory {
+class ChatViewModelFactory(private val context: Context, private val cid: String) : ViewModelProvider.Factory {
```
VelikovPetar: Also the `context` arg seems unused.

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/ChatInfoFragment.kt
```diff
@@ -68,109 +73,114 @@ class ChatInfoFragment : Fragment() {
...
-            if (state.loading) {
-                binding.optionsRecyclerView.isVisible = false
-                binding.progressBar.isVisible = true
-                return@observe
+        viewModel.events.observe(viewLifecycleOwner) { event ->
+            when (event) {
+                is ChannelInfoViewEvent.Error -> showError(event, isGroupChannel = false)
+                is ChannelInfoViewEvent.Navigation -> onNavigationEvent(event)
+                is ChannelInfoViewEvent.Modal -> showModal(event, viewModel, isGroupChannel = false) { modal ->
+                    if (modal is ChannelInfoViewEvent.HideChannelModal) {
```
VelikovPetar: This is a bit strange, should we maybe update the toggle only after the user confirms the modal?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/SingleLiveEvent.kt
andremion: For customers, it is transparent since they are internal.
However, I think `SingleLiveEvent` and `asSingleLiveEvent` have better semantics, which are very close to `asLiveData` and don't require any additional class as `Event`.

VelikovPetar: Yes of course, I was more concerned about staying consistent in the `internal` code, and the recommended way of using it 😄. But if you think it is OK to have both approaches, let's keep it like this.

stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/feature/chat/info/ChatInfoFragment.kt
andremion: You're right.
I reverted most of the logic to keep the same behavior as before.

PR #5790, opened 2025-05-24
fix(README): remove duplicate word 'by' in Chat Tutorial link description

Kargatharaakash:
This PR fixes a small typo in the README file where the phrase

"Learn the basics of the SDK by by building a simple messaging app"

contains a duplicated word **"by"**. This change removes the duplicate to improve readability.

### 🎯 Goal

Improve documentation clarity by fixing a minor typo.

### 🛠 Implementation details

Removed the duplicated word **"by"** in the README file.

### 🧪 Testing

No testing required, as this change only affects documentation.

### ☑️ Contributor Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (if required)
- [x] PR is linked to the GitHub issue it resolves (if any)
- [x] Affected documentation updated

PR #5808, opened 2025-06-04
Revert "fix(README): remove duplicate word 'by' in Chat Tutorial link description"

andremion:
Reverts GetStream/stream-chat-android#5790


andremion:
It will be reverted directly in the main without PR

PR #5820, opened 2025-06-09
Remove duplicate Indices on PrimaryKeys.

Sagar0-0:
### 🎯 Goal

Remove unnecessary @ColumnInfo(index = true) from a @PrimaryKey column, as Room and SQLite automatically index primary keys.

### 🛠 Implementation details

Removed @ColumnInfo(index = true) annotation from the primary key column in these two entity..
This cleanup helps avoid redundant annotations and keeps the entity definition minimal and accurate.

### 🧪 Testing

No runtime or behavioral impact expected, as primary keys are indexed by default.
Verified that Room compiles the schema correctly and no migrations are required for this change.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/attachment/internal/AttachmentEntity.kt
```diff
@@ -39,7 +38,6 @@ import io.getstream.chat.android.offline.repository.domain.message.internal.Mess
     indices = [Index("messageId")],
 )
 internal data class AttachmentEntity(
-    @ColumnInfo(index = true)
```
VelikovPetar, line 42: We would actually need to increment the DB version in `ChatDatabase`. Otherwise we would end up in an invalid state. You can reproduce the following by:
1. Build the app from `develop`
2. Login
3. Re-build again from this branch
4. You can see an error in the `ChannelList` with the exception: `java.lang.IllegalStateException: Room cannot verify the data integrity. Looks like you've changed schema but forgot to update the version number. You can simply fix this by increasing the version number`

VelikovPetar (changes requested):

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/attachment/internal/AttachmentEntity.kt
Sagar0-0: I see, but I couldn't understand why this issue is there. IMO, it should not cause this crash as the Index already exists on the column. Adding an index in the annotation also creates the same Index. Do you have any idea what is happening on the DB level.

JcMinarro: I am not sure why. Maybe it is some inconsistence with the data already stored on the DB and some internal check that Room.
It shouldn't be a problem because the DB is only used as offline option, but repopulated whenever new info come from the API. Old DB data will be removed when a new update of the app with the new SDK version is instaled, but it shouild be fine.
So, to be able to merge it, we will need to increase the DB version as @VelikovPetar said before

PR #5824, opened 2025-06-11
Keep updated `backAction` lambda

JcMinarro:
### 🎯 Goal
Keep updated `backAction` lambda
Complete: AND-599

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/MessagesScreen.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/MessagesScreen.kt
@@ -48,6 +48,7 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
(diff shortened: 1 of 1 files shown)


VelikovPetar (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/MessagesScreen.kt
```diff
@@ -179,7 +179,12 @@ public fun MessagesScreen(
     }
 
     val isImeVisible = WindowInsets.ime.getBottom(LocalDensity.current) > 0
-    val backAction: BackAction = remember(listViewModel, composerViewModel, attachmentsPickerViewModel) {
+    val backAction: BackAction = remember(
+        listViewModel,
+        composerViewModel,
+        attachmentsPickerViewModel,
+        isImeVisible,
+    ) {
```
andremion: I'm wondering... why do we need to remember that callback?
It is triggered when the user presses the system back button (or uses the gesture).
Am I missing something? 🤔

andremion: What about this?

```kotlin
val isImeVisible by rememberUpdatedState(
    WindowInsets.ime.getBottom(LocalDensity.current) > 0
)

BackHandler {
    val isStartedForThread = listViewModel.isStartedForThread
    val isInThread = listViewModel.isInThread
    val isShowingOverlay = listViewModel.isShowingOverlay

    when {
        isImeVisible -> Unit
        attachmentsPickerViewModel.isShowingAttachments ->
            attachmentsPickerViewModel.changeAttachmentState(false)
        isShowingOverlay -> listViewModel.selectMessage(null)
        isStartedForThread -> onBackPressed()
        isInThread -> {
            listViewModel.leaveThread()
            composerViewModel.leaveThread()
        }
        else -> onBackPressed()
    }
}
```

andremion: `rememberUpdatedState` ensures that `isImeVisible` inside the lambda always has the latest value.

PR #5779, opened 2025-05-16
[AND-421] Message Reminders

VelikovPetar:
### 🎯 Goal
Introduces the "Message Reminders" feature to the low-level client SDK. 
Allows creating/updating/deleting reminders for messages for which the user will be notified in the future, or saving a message as a bookmark (reminder without due date). It also allows querying of existing reminders, for the purpose of building lists where the users could have an overview of all saved messages.

Additionally, it introduces an example in the `Compose` sample app, serving as a guideline on how a reminders screen could be built using the LLC operations.

Specs: https://www.notion.so/stream-wiki/CHA-530-Snooze-message-remind-me-1886a5d7f9f6803abff5c271ad651d22

#### SDK
- Add `ChatClient.createReminder` - operation for creating a new message reminder / saving a message for later
- Add `ChatClient.updateReminder` - operation for updating the due date (`remindAt` time) for an existing reminder
- Add `ChatClient.deleteReminder` - operation for deleting an existing reminder
- Add `ChatClient.queryReminders` - operation for querying existing reminders

- Add `ReminderCreateEvent` - event delivered when a new reminder is created
- Add `ReminderUpdatedEvent` - event delivered when a reminder is updated
- Add `ReminderDeletedEvent` - event delivered when a reminder is deleted
- Add `NotificationReminderDueEvent` - event delivered when a reminder is due

- Add `Message.reminder` field - holds information about the reminder set for the message (stored in DB as well)

**Note: No real offline support is introduced for reminders, since we don't provide any state/UI for the reminders (other than the enhancement of the Message object), and the main functionality of the reminder is a push notification delivered when the reminder is due. We could introduce something like optimistic creations/updates/deletion of the reminders, but I don't think it makes too much sense, as the reminder wouldn't make sense if it is created on the BE after its due date.**

#### Sample App
- Add new side navigation drawer to the `ChannelsActivity` - it is now aligned with the XML sample + provides a new link to the Reminders screen
- Add new "Reminders" screen where the user can filter between `overdue`, `upcoming`, `saved for later`...
- They can delete or change the due date of the reminder by long-pressing the reminder items
- They can create/update reminders directly from the `MessageList` - Ui created by `MessageRemindersComponentFactory`

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Query Reminders</th>
<th>Create/Edit/Delete</th>
</tr>
</thead>
<tbody>
<tr>
<td>
[...]

PR #5833, opened 2025-06-17
Add support for different types of notifications (and handling for `notification.reminder_due`)

VelikovPetar:
### 🎯 Goal
**Note: This PR targets the feature/message-reminders branch, as a separate change on which the "Message Reminders" depends**
- Extends the SDK logic for handling push notifications of type other than `message.new`.
- Adds specific handling for the `notification.reminder_due` push notification

### 🛠 Implementation details
- Extend the `PayloadValidator` logic to also allow handling of `notification.reminder_due`
- Introduce a new `NotificationType` object defining the supported notification types
- Add new `NotificationHandler.showNotification(type: String, channel: Channel, message: Message)` method accepting the push notification `type`.
- Deprecate the `NotificationHandler.showNotification(channel: Channel, message: Message)`. For backwards compatibility, the new method calls the old deprecated method by default without passing the `type`.
- Introduce a new `MessagingStyleNotificationFactory` which builds the notification objects/ids for the `MessagingStyleNotificationHandler`
- Add default handling for the `notification.reminder_due` PN, showing the title: "You asked me to remind you about this message", and the body: message text. (Follows Slack example)
- Deprecates `io.getstream.chat.android.client.notifications.handler.PushDeviceGenerator` as it is not used (clients should use `PushDeviceGenerator` from the push artifact)

### 🎨 UI Changes

| Reminder notification |
| --- |
|Screenshot_20250617_151357 |

### 🧪 Testing
1. Existing PNs should not be affected and behave the same way 
2. `notification.reminder_due` testing will be covered in the message reminders [PR](https://github.com/GetStream/stream-chat-android/pull/5779)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/LoadNotificationDataWorker.kt
```diff
@@ -45,6 +45,7 @@ internal class LoadNotificationDataWorker(
     private val logger by taggedLogger(TAG)
 
     override suspend fun doWork(): Result {
+        val type = inputData.getString(DATA_TYPE)!!
```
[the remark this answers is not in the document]
VelikovPetar: This is intentional. The `LoadNotificationsDataWorker` is always started via the `start`, which ensures that the `DATA_TYPE` is always present. With proper usage, this will never crash.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -4468,7 +4469,10 @@ internal constructor(
         @JvmOverloads
         public fun handlePushMessage(pushMessage: PushMessage) {
             ensureClientInitialized().run {
-                if (!config.notificationConfig.ignorePushMessagesWhenUserOnline || !isSocketConnected()) {
+                val shouldHandle = !config.notificationConfig.ignorePushMessagesWhenUserOnline ||
+                    !isSocketConnected() ||
+                    pushMessage.type == NotificationType.NOTIFICATION_REMINDER_DUE
```
JcMinarro: Why this new type is ignored?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/ChatNotifications.kt
```diff
@@ -38,7 +39,7 @@ internal interface ChatNotifications {
     fun onPushMessage(message: PushMessage, pushNotificationReceivedListener: PushNotificationReceivedListener)
     fun onNewMessageEvent(newMessageEvent: NewMessageEvent)
     suspend fun onLogout(flushPersistence: Boolean)
-    fun displayNotification(channel: Channel, message: Message)
+    fun displayNotification(type: String, channel: Channel, message: Message)
```
JcMinarro: Instead of an string defining the type of PN, Why don't we use a sealed class??
It will allow us in the future to provide different data for every PN type without a breaking change

JcMinarro (changes requested):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: I intentionally added this, because I think that we should show this notification, even if the user is online. Mostly because the notification for a due reminder is time sensitive, so I think it makes sense to ensure that it is always shown. 
We could in theory though add a handler for the `notification.reminder_due` web socket event instead, and handle it in `ChatClient`, similar to how we have a handler for the `message.new` web socket event.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/ChatNotifications.kt
VelikovPetar: To make sure we are on the same page here: You mean to define something like:

```kotlin
sealed class Notification {
  
  data class NewMessage(val channel: Channel, val message: Message): Notification
  
  data class ReminderDue(val channel: Channel, val message: Message): Notification
  
  // ...Other potential types...
}
```
And then in `ChatNotifications` and in the `NotificationHandler` to instead call:

```kotlin
class ChatNotificationsImpl: ChatNotifications {

  override fun displayNotification(notification: Notification) {
    handler.showNotification(notification)
  }
} 
```

Also, additions to a sealed class could also be considered a breaking change, if we introduce a new child, while a customer is doing:
```kotlin
when (notification) {
  is Notification.MessageNew -> {}
  is Notification.ReminderDue -> {}
}
```

But maybe we could go with a data class, which will hold all the relevant data, such as:
```kotlin
data class Notification(
  val type: String,
  val message: Message,
  val channel: Channel,
)
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
JcMinarro: The class `ChatNotifications` already is invoked when a new message is received from the WS to allow our customers to show a notification or not (Maybe depending if they are on the current channel or not), we should follow the same approach.
I know this PN Type is time-sensitive, and just becouse that if the user is connected to the WS we should rely on it instead on the PN that could fail.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/ChatNotifications.kt
JcMinarro: The problem by keeping a "single data class" is that it force us to have always the same properties.
Just think on a PN type "Reaction", would be good to have the message where the reaction was added, the channel where the message is present, but we should know Who and Which reaction was added... The information could be inside of the Message itself, but we need to do extra logic to try to identify which reaction was the one we are receiving this PN from.
Instead of that, if this info is received from the PN Payload and mapped to a new `Notification.Reaction` class it would be easier to handle.
```
data class Reaction(val reaction: Reaction, val from: User, val message: Message, val channel: Channel) : Notification
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/ChatNotification.kt
```diff
@@ -0,0 +1,63 @@
...
+
+package io.getstream.chat.android.client.notifications.handler
+
+import io.getstream.chat.android.models.Channel
+import io.getstream.chat.android.models.Message
+
+/**
+ * Represents a chat notification with a specific type.
+ *
+ * @property type The type of the notification.
```
JcMinarro: Do we really need this property?

VelikovPetar: Currently no, but I thought it could be useful to have a reference to the raw type name. But I can remove it if you think it won't be necessary to maintain it.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -223,7 +223,8 @@ internal class DomainMapping(
                 restrictedVisibility = emptyList(),
                 reminder = reminder?.toDomain(),
                 extraData = extraData.toMutableMap(),
-        ).let(messageTransformer::transform)}
+            ).let(messageTransformer::transform)
```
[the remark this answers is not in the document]
VelikovPetar: The change is intentional.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/ChatNotification.kt
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5833/commits/d8a9f2530edb876a8d783ba849e3388ff99b0f89

PR #5834, opened 2025-06-17
Rearrange MessageList parameters.

Sagar0-0:
### 🎯 Goal

Improve API clarity and usability of the MessageList composable by consistently ordering parameters:
- Mandatory params first (for easier discoverability and reduced errors).
- modifier as the first optional param (to align with Compose best practices and user expectations).

This makes it easier for developers to use the composable and provides a more consistent experience across the API.


andremion:
Thanks for the contribution!
While we appreciate the intention to improve API clarity, we try to avoid introducing breaking changes—especially ones that don’t bring tangible value or functional improvement to the library. Even small changes in parameter order can cause unnecessary friction for users upgrading their code.
For now, we’ll close this PR to maintain stability and avoid disrupting existing usage.
Thanks again for your effort and understanding!

PR #4122, opened 2022-08-31
[ISSUE-4060] Fixing unread count when syncing at the same time

leandroBorgesFerreira:
### 🎯 Goal

Fix unread count when SDK is syncing at the same time

### 🛠 Implementation details

1 - Buffering events when requesting channels
This is necessary because this is the moment when we will know the information about the SDK, for example, we will know the unread count of a channel. 

Counting the unread messages before syncing with the backend makes no sense as we would wrongly assume that it is zero. Trusting the database is not a good idea also, because we have no guarantee that we have all messages loaded in the database, we could have 100+ unread messages, but have loaded only 30 messages of the channel. 

2 - Increasing the unread count only when the message is not already in the state. If the message is already in the SDK, it means that the count was already done (by the sync). Counting it again would create a wrong offset in it. 

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://user-images.githubusercontent.com/10619102/187703885-95e9b1da-559f-458b-bd30-9a7e20a345d1.mov" controls="controls" muted="muted" />
</td>
<td>
<video src="https://user-images.githubusercontent.com/10619102/187703956-35d8b773-0cec-4112-8a8f-d6f340972b9a.mov" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

As the video shows, right after start the sample app, send a storm of messages and check that the unread count of the summed channels is equal to the unread count of the backend (that little icon in the bottom left of the sample app asserts that)

#### General
- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] Assigned a person / code owner group (required)
- [ ] Thread with the PR link started in a respective Slack channel (#android-chat-core or #android-chat-ui) (required)
- [x] PR targets the `develop` branch
- [x] PR is linked to the GitHub issue it resolves

#### Code & documentation
- [ ] Changelog is updated with client-facing changes
- [x] New code is covered by unit tests
- [x] Comparison screenshots added for visual changes
- [x] Affected documentation updated (KDocs, docusaurus, tutorial)

### ☑️Reviewer Checklist
- [ ] UI Components sample runs & works
- [ ] Compose sample runs & works
- [ ] UI Changes correct (before & after images)
[...]

PR #5836, opened 2025-06-18
Unread message count should not be reset on entering a thread

### LINKED ISSUE #5753: Unread message count disappears on returning to ChannelListFragment
### Description
We're using the Stream Chat SDK with `ChannelListFragment`. The unread message count shows correctly when the fragment is first loaded, and it updates in real-time as expected. However, when the user navigates away from the fragment (e.g., back press or navigating to another screen) and later returns to it, all unread counts are reset to zero — even though there are still unread messages - as I havn't open the channel having unread messages

We are not modifying or customizing `ChannelListViewModel` directly and are relying on the default behavior of `ChannelListFragment`.

This seems to suggest that the ViewModel or internal channel query state is not being retained or re-initialized correctly when the fragment is recreated.

- SDK version
- v6.12.0

### Acceptance criteria
- Unread message counts persist after navigating away and returning to `ChannelListFragment`.
- ViewModel retains or re-queries correct channel state upon re-attachment.

### Screenshots
Image
Image

andremion:
### 🎯 Goal

Fix #5753 

### 🛠 Implementation details

The `MessageReadEvent` and `NotificationMarkReadEvent` were being processed for both messages and threads. This change ensures that they are only processed if it's not from a thread (`event.thread == null`). Thread message read events are handled by `QueryThreadsLogic`.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/0402bbdb-2e3c-4c35-93a2-fd745170444d" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/aabe3096-003a-477a-b986-1d5906987470" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🎉 GIF

gif


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -201,9 +201,9 @@ internal class ChannelMutableState(
     override val read: StateFlow<ChannelUserRead?> =
         combineStates(rawReads, userFlow) { readsMap, user -> user?.id?.let { readsMap[it] } }
 
-    val lastMarkReadEvent: StateFlow<Date?> = read.mapState { it?.lastRead }
-
-    override val unreadCount: StateFlow<Int> = read.mapState { it?.unreadMessages ?: 0 }
+    override val unreadCount: StateFlow<Int> = read.mapState {
+        it?.run { unreadThreads ?: unreadMessages } ?: 0
```
VelikovPetar: Hmm can we do this? What if the user has an `m` number of unread threads, but also has `n` number of unread messages in the channel -> Wouldn't this flow emit a wrong value? (always `m` instead of `n`)

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
```diff
@@ -42,17 +42,17 @@ internal fun NotificationMarkReadEvent.toChannelUserRead() = ChannelUserRead(
     user = user,
     lastReceivedEventDate = createdAt,
     lastRead = createdAt,
-// TODO: remove this once the backend is fixed and is sending us the number of unread messages
-    unreadMessages = 0,
-// TODO: Backend should send us the last read message id
-    lastReadMessageId = null,
+    unreadMessages = unreadChannels,
```
VelikovPetar: Can the `unreadChannels` be safely mapped to `unreadMessages` - Doesn't the `NotificationMarkReadEvent.unreadChannels` field represent the total number of unread channels for the user?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
```diff
@@ -42,17 +42,17 @@ internal fun NotificationMarkReadEvent.toChannelUserRead() = ChannelUserRead(
...
+    unreadThreads = unreadThreads,
```
VelikovPetar: Similar question here - Isn't the `NotificationMarkReadEvent.unreadThreads` the total number of unread threads for the user (not scoped for the channel)?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
andremion: Maybe it's time to separate those, but I'm afraid that's gonna be a breaking change.
I can see this `TODO` in the `MessageListController`
Screenshot 2025-06-27 at 10 28 50
However, that [PR](https://github.com/GetStream/stream-chat-android/pull/4122) got closed without merging.
@JcMinarro , do you have any context of that?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
andremion: That's what I see in the kdoc 🤷🏻 
Screenshot 2025-06-27 at 10 31 39

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
andremion: According to the kdoc, no.
I will test to see if it is triggered even for other channels.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
andremion: @VelikovPetar , I just checked that the event is triggered scoped to specific channels.
For example, when a user enters a channel A, the `NotificationMarkReadEvent` is triggered only for that channel.
The `ChannelLogic` which handles those events are scoped to a specific channel
I think the confusion might be because of the name `unreadChannels`, which should be `unreadMessages`.
Am I missing something?

andremion: That's still confusing because https://getstream.io/chat/docs/java/event_object/ says different

andremion: Maybe @JcMinarro has more context about it..

VelikovPetar: Yes the `NotificationMarkReadEvent` would be triggered for a specific channel being marked as read - But does the `unreadChannels` field reflect the number of unread messages in that specific channel (it should always be 0 because the channel was marked as read), or does it reflect the total number of unread channels for the user (not directly related to this specific channel)?

andremion: You're right. `unreadChannels` reflects the number of channels with unread messages.

JcMinarro: Yes, as Petar said, unreadChannels and unreadMessages are different.

"MarkRead" are triggered for an specific channel, but it return more info, not only the one related with the current channel, but the current state after the channel is read.
UnreadMessages we hardcoded as 0 because this event is marking the channel as read, so on that case you shouldn't have any pending message as unread, but added a "TODO comment" to revisit it if any time in the future backend is sending us this info, that wasn't sent when it was implemented but would be good to avoid race-conditions or any other corner-cases that was not covered at the moment of this feature was implemented

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -588,8 +588,12 @@ internal class ChannelLogic(
                     is NotificationChannelTruncatedEvent -> removeMessagesBefore(event.createdAt)
                     is TypingStopEvent -> channelStateLogic.setTyping(event.user.id, null)
                     is TypingStartEvent -> channelStateLogic.setTyping(event.user.id, event)
-                    is MessageReadEvent -> channelStateLogic.updateRead(event.toChannelUserRead())
-                    is NotificationMarkReadEvent -> channelStateLogic.updateRead(event.toChannelUserRead())
+                    is MessageReadEvent -> if (event.thread == null) {
+                        channelStateLogic.updateRead(event.toChannelUserRead())
+                    }
```
JcMinarro, line 593: What about thread?
aren't we updating it?
Or it is already updated somewhere else?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
```diff
@@ -44,8 +44,7 @@ internal fun NotificationMarkReadEvent.toChannelUserRead() = ChannelUserRead(
     lastRead = createdAt,
 // TODO: remove this once the backend is fixed and is sending us the number of unread messages
     unreadMessages = 0,
-// TODO: Backend should send us the last read message id
-    lastReadMessageId = null,
+    lastReadMessageId = lastReadMessageId,
```
JcMinarro: Is `lastReadMessageId` the right one?
I think in the past it wasn't a valid one and we had some issues with `markUnread` feature.
This `lastReadMessageId` is used within our MessageListControler to find the position, if any, where the "Unread Messages" label needs to be shown.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
andremion: I believe the thread is handled in `QueryThreadsLogic` 🤔

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
andremion: I thought it would be the right one because it comes from the event.
I can revert it too. The name seems an obvious mapping, but it seems not 😅

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
JcMinarro: Threads is something wasn't implemented as a separated feature. 
Initially it was only a list of messages with a common parentMessageId (Even now we keep this implementation on the MessageList yet).
But recently we introduced a "real concept" of Thread (It was implemente by @VelikovPetar some time ago).

We will need to revamp our curretnt implementation to be using the new Thread concept instead of the old implementation, but it is not ready yet

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
andremion: It was reverted 🙏🏻

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
VelikovPetar: Added handling here: https://github.com/GetStream/stream-chat-android/pull/5836/commits/d2110a3e7219307578c106be0044dc948e919ebd

PR #5843, opened 2025-06-23
Show original translation (XML)

VelikovPetar:
### 🎯 Goal

Adds a new (optional) functionality in the Message List, allowing users to quickly show the original text of an auto-translated message.
Resolves: https://linear.app/stream/issue/AND-450/show-original-translation

### 🛠 Implementation details
- Add new `ChatUI` flag which enables the feature: `showOriginalTranslationEnabled` (default: `false`)
- When this flag is set to true, the `Translated to ...` is extended with the text: ` · Show original`. Clicking on this label, will result in showing the original text of the message. When the original text is shown, a new label would be shown: Show translation. Tapping on it, will again show the translated text.
- Add new field to the `MessageItemState`- `showOriginalText` - indicating that the message was manually 'reverted' to show its original, non-translated text.
- Expose `MessageListController.toggleOriginalText(messageId:String)` method for toggling between translated and original text for the given message.
- Update the default `ChatUi.messageTextTransformer` to take the `MessageItemState.showOriginalText` into consideration before transforming the text.

**Additional changes/fixes:** 
- The `Translated to ...` label is now always visible underneath the message bubble, and is not hidden when the message item is not showing the message footer. (Aligned to the Compose SDK).

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Apply the given patch (Sets the ChatUi. showOriginalTranslationEnabled to true, and removes the custom message formatter)
2. Write a message from a different user (with different lang from the current user)
3. The auto-translated messages in the MessageList show also have the  · Show original label

<details>

<summary>Sets the ChatUi. showOriginalTranslationEnabled to true, and removes the custom message formatter</summary>

```
Index: stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt	(revision 2e8c5212961eca11b2cf71cda7faa162bbfc63f2)
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt	(date 1750695425669)
@@ -25,7 +25,6 @@
 import io.getstream.chat.android.client.logger.ChatLogLevel
 import io.getstream.chat.android.client.notifications.handler.NotificationConfig
 import io.getstream.chat.android.client.notifications.handler.NotificationHandlerFactory
-import io.getstream.chat.android.markdown.MarkdownTextTransformer
 import io.getstream.chat.android.models.ReactionSortingByLastReactionAt
@@ -110,9 +109,8 @@
             .build()
 
         // Using markdown as text transformer
-        val messageTranslator = MessageTranslator(client::getCurrentUser, autoTranslationEnabled)
         ChatUI.autoTranslationEnabled = autoTranslationEnabled
-        ChatUI.messageTextTransformer = MarkdownTextTransformer(context, messageTranslator)
+        ChatUI.showOriginalTranslationEnabled = true
         ChatUI.draftMessagesEnabled = true
 
         TransformStyle.viewReactionsStyleTransformer = StyleTransformer { defaultStyle ->

```

</details>

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -141,6 +141,7 @@ import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import java.util.Date
 import io.getstream.chat.android.ui.common.state.messages.Flag as FlagMessage
@@ -223,6 +224,7 @@ public class MessageListController(
     public val unreadLabelState: MutableStateFlow<UnreadLabel?> = MutableStateFlow(null)
     private val showUnreadButtonState = MutableSharedFlow<Boolean>(extraBufferCapacity = 1)
     private val updateUnreadLabelState = MutableStateFlow(true)
+    private val messagesInOriginalLanguage = MutableStateFlow<Set<String>>(emptySet())
 
     /**
      * Holds information about the abilities the current user is able to exercise in the given channel.
@@ -469,6 +471,7 @@ public class MessageListController(
                 unreadLabelState,
                 channelState.members,
                 channelState.endOfOlderMessages,
+                messagesInOriginalLanguage,
             ) { data ->
                 val state = data[0] as MessagesState
                 val reads = data[1] as List<ChannelUserRead>
@@ -483,6 +486,7 @@ public class MessageListController(
                 val unreadLabel = data[10] as UnreadLabel?
                 val members = data[11] as List<Member>
                 val endOfOlderMessages = data[12] as Boolean
+                val messagesInOriginalLanguage = data[13] as Set<String>
 
                 when (state) {
                     is MessagesState.Loading,
@@ -517,6 +521,7 @@ public class MessageListController(
                                 read = reads,
                             ),
                             ownCapabilities = channel.ownCapabilities,
+                            messagesInOriginalLanguage = messagesInOriginalLanguage,
                         ),
                         endOfNewMessagesReached = endOfNewerMessages,
                     )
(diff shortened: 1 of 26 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -223,6 +224,7 @@ public class MessageListController(
     public val unreadLabelState: MutableStateFlow<UnreadLabel?> = MutableStateFlow(null)
     private val showUnreadButtonState = MutableSharedFlow<Boolean>(extraBufferCapacity = 1)
     private val updateUnreadLabelState = MutableStateFlow(true)
+    private val messagesInOriginalLanguage = MutableStateFlow<Set<String>>(emptySet())
```
JcMinarro:
```suggestion
    private val messageIdsInOriginalLanguage = MutableStateFlow<Set<String>>(emptySet())
```
It is a bit confusing if we don't use `id'' because a message is normally a text (String) as well

JcMinarro (approved):

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
VelikovPetar: Ah nice catch, I will update this.

VelikovPetar: Renamed it to `messagesInOriginalLanguageIds`: https://github.com/GetStream/stream-chat-android/pull/5843/commits/291ade92d44d7a81b009bb2233c709d14ec10b8d

PR #5844, opened 2025-06-23
Polish MessageInputOptions.

Sagar0-0:
### 🎯 Goal

- Use vectorResource instead of painterResource to avoid unnecessary recompositions.
- Add contentDescription
- Add modifier param name.

Ig no changelog is required.


andremion:
We appreciate your collaboration.
However, we are closing this for now, as there is no proof that those changes make any difference in terms of performance.
Thanks,

PR #5846, opened 2025-06-24
Extend list of non permanent errors with `ConnectException`

VelikovPetar:
### 🎯 Goal

Resolves: https://linear.app/stream/issue/AND-578/androidchat-sdk-does-not-send-reaction-request-to-mock-server-after

Detected as part of an E2E test, where an operation which failed due to no network connectivity, was marked as a non-recoverable error. In these tests, the call failed with the `ConnectException` which was not considered a recoverable error. 
In this PR we extend the list of recoverable exceptions with the `ConnectException` (in addition to `UnknownHostException`). 

### 🛠 Implementation details
- Extend the list of recoverable exceptions with `ConnectException`.

### 🎨 UI Changes
_NA_

### 🧪 Testing
`E2E test_userAddsReactionWhileOffline` should now pass on CI


testableapple (approved):
Works great for me 👍

PR #5847, opened 2025-06-25
Refactor how Draft Messages are queried to avoid potential issues on the sync process

JcMinarro:
### 🎯 Goal
Refactor how Draft Messages are queried to avoid potential issues on the sync process.
Instead of fetching them when the SDK sync, they are obtained from the queryChannels.

Complete: AND-627

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
@@ -149,6 +149,11 @@ public interface MessageRepository {
     public suspend fun selectDraftMessagesByCid(cid: String): DraftMessage?
 
     /**
+     * Selects a draft message by its parent message ID.
+     */
+    public suspend fun selectDraftMessageByParentId(parentId: String): DraftMessage?
+
+    /**
      * Delete a draft message.
      */
     public suspend fun deleteDraftMessage(message: DraftMessage)
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
@@ -144,6 +144,11 @@ public interface MessageRepository {
     public suspend fun selectDraftMessages(): List<DraftMessage>
 
     /**
+     * Selects a draft message by its channel ID.
+     */
+    public suspend fun selectDraftMessagesByCid(cid: String): DraftMessage?
+
+    /**
      * Delete a draft message.
      */
     public suspend fun deleteDraftMessage(message: DraftMessage)
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/noop/NoOpRepositoryFactory.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/noop/NoOpRepositoryFactory.kt
@@ -26,7 +26,6 @@ import io.getstream.chat.android.client.persistance.repository.ThreadsRepository
 import io.getstream.chat.android.client.persistance.repository.UserRepository
 import io.getstream.chat.android.client.persistance.repository.factory.RepositoryFactory
 import io.getstream.chat.android.models.Channel
-import io.getstream.chat.android.models.DraftMessage
 import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.models.User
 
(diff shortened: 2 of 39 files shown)


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
```diff
@@ -234,29 +230,6 @@ internal class SyncManager(
...
     }
 
-    private suspend fun syncOfflineDraftMessages() {
-        repos.selectDraftMessages().forEach { draftMessage ->
-            mutableGlobalState.updateDraftMessage(draftMessage)
-        }
-    }
-
-    private fun syncDraftMessages(
-        limit: Int = QUERY_DRAFT_MESSAGES_LIMIT,
```
VelikovPetar, line 244: Can we also delete the `QUERY_DRAFT_MESSAGES_LIMIT` constant as it is no longer used.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/Channel.kt
```diff
@@ -215,6 +217,7 @@ public data class Channel(
             membership = channel.membership
             cachedLatestMessages = channel.cachedLatestMessages
             isInsideSearch = channel.isInsideSearch
+            draftMessage = channel.draftMessage
```
VelikovPetar: I think we should also add the `withDraft(DraftMessage)` method to the `Builder`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/noop/NoOpRepositoryFactory.kt
```diff
@@ -26,6 +26,7 @@ import io.getstream.chat.android.client.persistance.repository.ThreadsRepository
 import io.getstream.chat.android.client.persistance.repository.UserRepository
 import io.getstream.chat.android.client.persistance.repository.factory.RepositoryFactory
 import io.getstream.chat.android.models.Channel
+import io.getstream.chat.android.models.DraftMessage
```
VelikovPetar: This seems unused, can it be removed?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
```diff
@@ -143,6 +143,11 @@ public interface MessageRepository {
      */
     public suspend fun selectDraftMessages(): List<DraftMessage>
 
+    /**
+     * Selects a draft message by its channel ID.
+     */
+    public suspend fun selectDraftMessagesByCid(cid: String): DraftMessage?
```
VelikovPetar, line 149: We should rename this to `selectDraftMessageByCid`, as it only retrieves a single object.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/Channel.kt
```diff
@@ -94,6 +94,7 @@ public data class Channel(
     val membership: Member? = null,
     val cachedLatestMessages: List<Message> = emptyList(),
     val isInsideSearch: Boolean = false,
+    val draftMessage: DraftMessage? = null,
```
VelikovPetar, line 97: Not really a big deal, but maybe we should keep this named just `draft` - to be in line with the BE model?  (Same for `ChannelData.draftMessage` and `Thread.draftMessage`)

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/MessageDao.kt
```diff
@@ -88,6 +88,12 @@ internal interface MessageDao {
     @Query("SELECT * FROM $DRAFT_MESSAGE_ENTITY_TABLE_NAME")
     suspend fun selectDraftMessages(): List<DraftMessageEntity>
 
+    @Query("SELECT * FROM $DRAFT_MESSAGE_ENTITY_TABLE_NAME WHERE cid = :cid")
```
VelikovPetar: Small concern here: Because we query the drafts by cid, wouldn't it be possible for this query to return a Draft which corresponds for a Thread? (those drafts could still have a cid + parentMessageId). I think it would be safer if here we query by cid + parentMessageId = null, to ensure that we only get the channel related draft.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/global/internal/MutableGlobalState.kt
```diff
@@ -102,6 +102,7 @@ internal class MutableGlobalState : GlobalState {
     }
 
     fun updateDraftMessage(draftMessage: DraftMessage) {
+        println("JcLog: updateDraftMessage: $draftMessage")
```
VelikovPetar, line 105: Some leftover logs, it would be great if we can remove them.

VelikovPetar (approved):
Looks good!

PR #5832, opened 2025-06-17
Add ability to cancel link previews

andremion:
This commit introduces the ability to cancel link previews in the message composer.

Key changes:
- Added a `cancelLinkPreview` method to `MessageComposerController` and `MessageComposerViewModel`.
- Updated `ComposerLinkPreview` to accept an `onCancel` callback.
- When sending a message, if `linkPreviews` is empty, `skipEnrichUrl` will be set to `true`.
- Connected the cancel action in the UI to the new methods.
- Updated `ChatComponentFactory` and `MessageComposer` to pass the new `onCancelLinkPreviewClick` callback.

### 🎯 Goal

_Describe why we are making this change_

### 🛠 Implementation details

_Describe the implementation_

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="" controls="controls" muted="muted" />
</td>
<td>
<video src="" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

_Explain how this change can be tested (or why it can't be tested)_

_Provide a patch below if it is necessary for testing_

<details>

<summary>Provide the patch summary here</summary>

```
Provide the patch code here
```

</details>

### 🎉 GIF

_Please provide a suitable gif that describes your work on this pull request_

PR #5848, opened 2025-06-25
Deprecate `AttachmentType.LINK` and fix link styling not applied in `MessageReplyView`. 

VelikovPetar:
### 🎯 Goal
The attachment type `link` is not an officially supported (see: https://getstream.io/chat/docs/android/message_format/#attachment-format and https://getstream.io/chat/docs/android/message_format/#open-graph-scraper). Exposing the `AttachmentType.LINK` as a public type can be confusing for customers, because it gives a false impression that links are a separate type of attachment. Because we cannot delete this directly, it is just deprecated for the moment, and to be removed in an upcoming release. 

Additionally, this PR fixes a bug where the link styling (`MessageReplyStyle.linkBackgroundColorMine`/`MessageReplyStyle.linkBackgroundColorTheirs`/`MessageReplyStyle.linkStyleMine`/`MessageReplyStyle.linkStyleTheirs`) is not applied (because the current checks for isLink() were always returning false).

### 🛠 Implementation details
 - Deprecate `AttachmentType.LINK`
 - Deprecate `Attachment.isLink()`
 - Rework `isLink()` check in `MessageReplyView` to correctly check if the message has a link (currently it always returns `false`)

### 🎨 UI Changes

| Before | After |
| --- | --- |
| quoted_links_before | quoted_links_after |

### 🧪 Testing
1. Customise the MessageReplyStyle properties:
  - linkBackgroundColorMine
  - linkBackgroundColorTheirs
  - linkStyleMine
  - linkStyleTheirs
2. Run the XML sample app
3. Quote a message containing a link
4. The quoted message should have the customisations (background color, text color) applied.


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/adapter/view/MessageReplyView.kt
```diff
@@ -223,9 +224,7 @@ public class MessageReplyView : FrameLayout {
                 displayedText
             }
         } else {
-            if (attachment.isLink()) {
```
VelikovPetar, line 230: This code was unreachable:
- The `attachment.isLink()` would always return `false`
- We can never enter this `else` branch, because the `displayedText` will never be blank when a link is sent

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/adapter/view/MessageReplyView.kt
```diff
@@ -183,7 +183,8 @@ public class MessageReplyView : FrameLayout {
     }
 
     private fun isLink(message: Message) = message.attachments.run {
-        size == 1 && last().isLink()
+        // Link messages have an image attachment with title_link or og_scrape_url set.
+        size == 1 && last().isImage() && (last().titleLink != null || last().ogUrl != null)
```
andremion: What happens when a message has a link but the user skips the preview? Will it have an attached image anyway? Should that logic consider [canceled preview links](https://github.com/GetStream/stream-chat-android/pull/5832)?

VelikovPetar: I will need to test this scenario, I am not sure what happens when a link is sent without enrichment.

VelikovPetar: Nice catch, this will not work with canceled preview links, in such cases the message does not contain an attachment: 
```json
{
  "message": {
    "id": "pvelikov-e0a88364-1222-4e25-8c8b-86931fada5d1",
    "text": "www.instagram.com",
    "html": "\u003cp\u003ewww.instagram.com\u003c/p\u003e\n",
    "type": "regular",
    "user": {
      "id": "pvelikov",
      "name": "Petar Velikov",
      "image": "https://ca.slack-edge.com/T02RM6X6B-U07LDJZRUTG-a4129fed05b6-512",
      "language": "en",
      "role": "user",
      "teams": [],
      "created_at": "2024-09-26T09:47:56.738237Z",
      "updated_at": "2025-04-04T13:26:14.547567Z",
      "banned": false,
      "online": true,
      "last_active": "2025-07-03T11:46:54.506114572Z",
      "blocked_user_ids": []
    },
    "attachments": [],
    "latest_reactions": [],
    "own_reactions": [],
    "reaction_counts": {},
    "reaction_scores": {},
    "reply_count": 0,
    "deleted_reply_count": 0,
    "cid": "messaging:c0309fc8-ec2e-4227-97a1-afc4ead418aa",
    "created_at": "2025-07-03T11:47:31.621476Z",
    "updated_at": "2025-07-03T11:47:31.621476Z",
    "shadowed": false,
    "mentioned_users": [],
    "silent": false,
    "pinned": false,
    "pinned_at": null,
    "pinned_by": null,
    "pin_expires": null,
    "restricted_visibility": []
  },
  "duration": "24.19ms"
}
```
This also 'breaks' the styling of the regular message bubble in the message list. I will need to see how to properly detect the presence of a link, when the message is sent without enrichment of the url.

VelikovPetar: Getting back to this (finally) - In the case where `skip_enrich_url` is set to `true` we don't get an `attachment` in the message, so the current check for link presence based on the `message.attachments` will always fail. So the only thing that we can do is to check the `message.text` for the presence of a URL. I am thinking to do something like: 
```kotlin
private fun containsUrl(message: String): Boolean =
        Patterns.WEB_URL.matcher(message).find()
```
(still need to test how this behaves and whether it is reliable).

What's your opinion on this? (I think we also need to consider whether we want to consider a message with an un-enriched link as a `link` message in general)

andremion: I can't find any other way.

PS. Consider using `PatternsCompat.AUTOLINK_WEB_URL` instead, as it is the one used in other parts of the SDK.

VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-android/pull/5848/commits/b4d2278553474f14678c7375f46076eb24f30fb8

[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5848/commits/b4d2278553474f14678c7375f46076eb24f30fb8

PR #5850, opened 2025-06-25
Support adding members to group in the channel info screen

andremion:
### 🎯 Goal

Showcase how to support adding members to a group in the channel info screen.

### 🛠 Implementation details

The `DraftChannelViewController` implementation is also moved to the sample source.

It was previously added (alongside its unit tests) to the SDK, foreseeing a shareable component between SDKs. Since there are no plans to introduce that feature to the SDK, as each customer can have different logic on creating channels, we are moving its implementation to the sample only.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Phone</th>
<th>Tablet</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/69ec040f-47a3-4349-91d4-da3db5fd7390" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/e9aca27d-aea0-4417-87fe-97eb67f0d33f" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### DIFF
--- a/metrics/size.json
+++ b/metrics/size.json
@@ -3,7 +3,7 @@
     "stream-chat-android-client": 15296,
     "stream-chat-android-offline": 15612,
     "stream-chat-android-ui-components": 24136,
-    "stream-chat-android-compose": 26984
+    "stream-chat-android-compose": 26988
   },
   "release": {
     "stream-chat-android-client": 3236,


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/AddMembersViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+                        .addMembers(params)
+                        .await()
+                        .onSuccess { _events.tryEmit(AddMembersViewEvent.MembersAdded) }
+                }
+            }
+        }
+    }
+
+    private fun String.toRequest(currentMembers: List<Member>): QueryUsersRequest {
+        val currentMembersFilter = Filters.nin("id", currentMembers.map(Member::getUserId))
```
VelikovPetar: Just a small heads-up: The `nin` filter is deprecated and we would most likely remove it in the next major release - so we would need to re-work this logic. (perhaps with a filter after we get the results)

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/AddMembersViewModel.kt
```diff
@@ -0,0 +1,186 @@
...
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.launch
+
+class AddMembersViewModel(
+    private val cid: String,
+    private val resultLimit: Int,
+    private val chatClient: ChatClient = ChatClient.instance(),
+) : SearchUsersViewModel(chatClient) {
```
VelikovPetar: I am not sure whether extending the `SearchUsersViewModel` is necessary - I don't think any of its fields/methods are used in the `AddMembersViewModel`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/info/GroupChannelInfoScreen.kt
```diff
@@ -90,6 +90,7 @@ public fun GroupChannelInfoScreen(
     modifier: Modifier = Modifier,
     currentUser: User? = ChatClient.instance().getCurrentUser(),
     onNavigationIconClick: () -> Unit = {},
+    onAddMembersClick: () -> Unit = {},
```
VelikovPetar, line 94: Can we add KDoc for the new method?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoMemberViewController.kt
```diff
@@ -112,14 +112,14 @@ public class ChannelInfoMemberViewController(
     private lateinit var member: Member
     private var distinctCid: String? = null
 
-    private fun queryDistinctChannel(): Flow<Channel?> =
+    private fun queryDistinctChannel(type: String): Flow<Channel?> =
```
VelikovPetar: One small concern -> This assumes that the distinct channel will have the same type as the original channel. I am not sure if we can safely make this assumption, we should perhaps make this check somehow configurable.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/AddMembersViewModel.kt
andremion: I see that `nin` is deprecated, but the deprecation doesn't inform us which other filter we should use.
Filtering locally doesn't seem performant, though.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/AddMembersViewModel.kt
andremion: Great catch!
That was an initial approach.
Let me fix it.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/AddMembersViewModel.kt
JcMinarro: Backend is not supporting an alternative filter for it.
The only option would be to remove it and do some extra filtering process locally.
Not ideal, but as far as it is only used on the Sample App, should be the way to go

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/ChannelInfoMemberViewController.kt
andremion: I would suggest keeping it as is and making any changes on demand.
I believe that it's a safe assumption. However, since that feature is experimental, we can improve it in the future.

JcMinarro: In our sample app we only used `massaging` type, and any "Distinct channel" was created with this type (What is the more common)
But I think on this case we are exposing it as a feature, we shouldn't filter by type, but only by membersId, that will return us any distinct channel between those members, doesn't mind which channelType

PR #5851, opened 2025-06-25
Support ImageVector and contentDescription for BackButton.

Sagar0-0:
### 🎯 Goal

Improve the efficiency and accessibility of the BackButton composable by avoiding unnecessary recompositions and ensuring proper content descriptions for accessibility tools.

🛠 Implementation Details
- Added a new BackButton overload that accepts an ImageVector instead of a Painter for the icon.
Why: Painter can cause additional recompositions when used with vector assets, which is unnecessary when ImageVector is available.

- Updated both the old and new BackButton composables to accept a contentDescription parameter.
Why: To ensure accessibility compliance and support screen readers or other assistive technologies.

✅ Benefits
- Avoids extra recompositions by using ImageVector for static vector assets.
- Improves accessibility with customizable content descriptions.
- Maintains backward compatibility by keeping the original Painter-based overload.

I also recommend moving the onBackPressed parameter to the end of the function signature. This aligns with Kotlin's convention for better readability and enables the use of trailing lambda syntax when invoking the composable.


andremion:
Thank you for your time in collaborating with this, @Sagar0-0 
We appreciate it.

Regarding using `vectorResource` instead of `painterResource`.. We need to evaluate on a case-by-case basis.
It's fine to use `painterResource` when it's a small resource or it's not used in recomposition-heavy contexts. For example, the functions you have changed are not recomposed often, and the resource is very small.

It would be nice if you could prove that using the current approach causes lags on a screen, so that we will know that using a different approach fixes it.

Thanks,

Sagar0-0:
@andremion
Using painterResource or imageVector in a non-composition-heavy context doesn’t make much of a practical difference.
However, based on my observations, painterResource can trigger unnecessary recompositions, especially when used with static vector assets.
Avoiding that may help optimize recomposition behavior, though depending on the use case, it might be a meaningful improvement or just premature optimization.

andremion:
We are closing this for now, as there is no proof that those changes make any difference in terms of performance.
Thanks,

PR #5852, opened 2025-06-26
Location sharing

JcMinarro:
### 🎯 Goal

Introduce [Location Sharing](https://www.notion.so/stream-wiki/Location-Sharing-1a86a5d7f9f680458c14c1b60afab06f#1ad6a5d7f9f6805d91faedc11ae5ae94) to the LLC

#### Added new functions to the `ChatClient` API:
- `queryActiveLocations`: Queries the active locations (non-expired) shared by the current user.
- `sendStaticLocation`: Sends a static location message to the given channel.
- `startLiveLocationSharing`: Starts live location sharing for the given channel.
- `updateLiveLocation`: Updates the live location info of a message.
- `stopLiveLocationSharing`: Stops the live location sharing for a message.

#### Introduced the `LiveLocationListener` interface to handle live location-related request events:
- `onQueryActiveLocationsResult`: Called when the result of `ChatClient.queryActiveLocations` is received.
- `onStartLiveLocationSharingResult`: Called when the result of `ChatClient.startLiveLocationSharing` is received.
- `onUpdateLiveLocationPrecondition`: Precondition for the request of updating live location. It can be used, for example, to throttle requests.
- `onUpdateLiveLocationResult`:  Called when the result of `ChatClient.updateLiveLocation` is received.
- `onStopLiveLocationSharingResult`: Called when the result of `ChatClient.stopLiveLocationSharing` is received.

#### Dependency Updates:
- Added `play-services-location` library to the demo project dependencies to use the `FusedLocationProviderClient`
- Included `androidx.compose.material:material-icons-extended` for extended Compose Material icons. 

#### Service
- Added `SharedLocationService`, a simple class to handle shared location updates for simplicity. It listens for the user's active live locations and requests location updates from the device if any active live locations are detected. Whenever a location result is returned from the device, it requests to update the live location in the chat client.

#### Live location sharing UI
- Added a location picker to the attachments tab, which contains actions to send a static location or a live location with an option to choose the sharing duration.
- Added a `CustomChatComponentFactory.MessageItemCenterContent` factory, which renders a location item if the message has a shared location.

> [!NOTE]
>  To simplify the demo and prevent using paid Map APIs that require API keys, we are using a `WebView` with [LeafletJs](https://leafletjs.com/).

### 🎨 UI Changes

<table>
<tbody>
<tr>
<th colspan=2>Live Location sharing expiring after 1 minute</th>
</tr>
<tr>
<td colspan=2>
<video src="https://github.com/user-attachments/assets/4ed23b64-720f-4045-b07c-fd6a2d03c1e2" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<th>Stopping a Live Location sharing</th>
<th>Sending a Static Location</th>
</tr>
<tr>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/e4fd8966-28c8-4978-9374-9caf600d0c47" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/dec016a4-dabd-4ff1-bc2c-7df510c708db" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
```diff
@@ -147,6 +149,12 @@ internal class SyncManager(
         logger.v { "[awaitSyncing] completed" }
     }
 
+    private suspend fun scheduleLiveLocationExpiration() {
+        mutableGlobalState.removeExpiredLiveLocations()
+        delay(LIVE_LOCATION_EXPIRATION_WAIT_TIME)
+        scheduleLiveLocationExpiration()
```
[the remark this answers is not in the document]
andremion: It's already inside a coroutine

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
```diff
@@ -147,29 +145,40 @@ internal class SyncManager(
...
```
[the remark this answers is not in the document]
andremion: `delay` already takes into consideration whether the coroutine is active or not

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1548,9 +1550,153 @@ internal constructor(
...
+     * Sends a static location message to the given channel.
+     *
+     * @param cid The full channel id, i.e. "messaging:123" to which the location will be sent.
+     * @param latitude The latitude of the location.
+     * @param longitude The longitude of the location.
+     * @param deviceId The device ID from which the location is sent.
+     */
+    @CheckResult
+    public fun sendStaticLocation(
+        cid: String,
```
VelikovPetar, line 1576: One small remark: Almost all of the `ChatClient` methods accept `channelType` and `channelId` instead of `cid` -> perhaps we should keep that aligned?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
```diff
@@ -137,10 +139,21 @@ internal class DtoMapping(
                     text = text,
                     thread_participants = threadParticipants.map { it.toDto() },
                     restricted_visibility = restrictedVisibility,
+                    shared_location = sharedLocation?.toDto(),
                     extraData = extraData,
                 )
             }
 
+    internal fun Location.toDto(): UpstreamLocationDto = UpstreamLocationDto(
+        // channel_cid = cid,
```
VelikovPetar: Should these lines be commented?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/ThrottlingPlugin.kt
```diff
@@ -16,25 +16,57 @@
...
 import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.models.User
-import io.getstream.log.StreamLog
+import io.getstream.log.taggedLogger
 import io.getstream.result.Error
 import io.getstream.result.Result
 import kotlin.reflect.KClass
 
-internal class ThrottlingPlugin : Plugin {
+internal class ThrottlingPlugin(
```
VelikovPetar, line 27: Is this used? As far as I can see the `ThrottlingPluginFactory` is never applied in the project. Also, if we need to apply it, we need to check if the `onChannelMarkReadPrecondition` logic is still valid - otherwise we could introduce a bug.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
```diff
@@ -75,8 +78,10 @@ internal class QueryChannelsMutableState(
 
     private var _endOfChannels: MutableStateFlow<Boolean>? = MutableStateFlow(false)
     private val sortedChannels: StateFlow<List<Channel>?> =
-        mapChannels.combine(latestUsers) { channelMap, userMap ->
-            channelMap?.values?.updateUsers(userMap)
+        combine(mapChannels, latestUsers, activeLiveLocations) { channelMap, userMap, activeLocations ->
```
VelikovPetar, line 81: I am not sure I follow this part: Why do we need the `activeLocations` in the `QueryChannels` state?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
```diff
@@ -147,29 +148,78 @@ internal class SyncManager(
         logger.v { "[awaitSyncing] completed" }
     }
 
+    private fun scheduleLiveLocationMessageUpdates() {
+        logger.d { "[scheduleLiveLocationMessageUpdates] no args" }
+        // Fetch initial active live locations
+        chatClient.queryActiveLocations().enqueue { result ->
```
VelikovPetar: I am a bit worried about adding this on every connection, as it could end up wasting resources by doing unnecessary work. Also, this would end up performing a call to the BE even if the integration app is not using live locations at all. 

Can we somehow schedule the updates only if the integrator app called the `queryActiveLocations()` (meaning they explicitly want to do this), or if the queryChannels has returned a channel with an active location? (I see that the `ChannelResponse` is now enriched with liveLocations)

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
```diff
@@ -147,29 +148,78 @@ internal class SyncManager(
...
     internal fun onEvent(event: ChatEvent) {
         when (event) {
             is ConnectingEvent -> syncScope.launch {
                 logger.i { "[onEvent] ConnectingEvent received" }
             }
+
             is ConnectedEvent -> syncScope.launch {
                 logger.i { "[onEvent] ConnectedEvent received" }
                 onConnectionEstablished(currentUserId)
+                syncScope.launch { syncOfflineDraftMessages() }
```
VelikovPetar, line 198: Was the `syncOfflineDraftMessages` intentionally moved here?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/StateRegistry.kt
```diff
@@ -99,7 +102,7 @@ public class StateRegistry constructor(
      */
     internal fun mutableChannel(channelType: String, channelId: String): ChannelMutableState {
         return channels.getOrPut(channelType to channelId) {
-            ChannelMutableState(channelType, channelId, userStateFlow, latestUsers, now)
+            ChannelMutableState(channelType, channelId, userStateFlow, latestUsers, activeLiveLocations, now)
```
VelikovPetar, line 105: Do we need to pass this data from here? Wouldn't we initially get the channels liveLocations from the queryChannels call? (and then process potential new location messages from events)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: I see...
Don't know if that is a legacy pattern or not.
Not sure if using `channelType` and `channelId` improves the customer experience of the API.
I can change that, no problem.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
andremion: Oops, leftovers... 😅 
I will fix that 🙌🏻

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/ThrottlingPlugin.kt
andremion: @JcMinarro introduced it, but I'm in doubt if that was just to showcase how to throttle requests.
We can move it to the sample, then decide later.
WDYT?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
andremion: `Channel` model has `activeLiveLocations` attribute, which contains the active live locations of that given channel.
However, since you mentioned, I believe we don't need to update it there since the channel should already be updated at that point.
WDYT?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
andremion: `queryActiveLocations` returns the active locations (non-expired) shared by the current user.
If the app is not using location, nothing else is done, so the second suggestion is already covered in a different way.
About the first suggestion, it doesn't help the integrator's life. Which kind of resource wasting do you mean? I believe that's minimal.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
andremion: Yes. Otherwise, it could end up being canceled on the disconnect event before it is completed, as we have seen with the live location scheduling.
I know that, in case of `syncOfflineDraftMessages`, it should be faster than with the live location scheduling, but who knows with different devices/internet connection?! 🤷🏻‍♂️ 
Do you see any issue with doing it in the connected event?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/StateRegistry.kt
andremion: Initially, there were `currentActiveLiveLocations` (holds the current user's active live locations) and `activeLiveLocations` (holds all users' live locations).
I realized that the global state doesn't need to hold all users' live locations, as it is wasteful.
However, I see that the previous code relies on all users' live locations.
I have to review that...

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: I'm still wondering whether we should move to `type:id` instead of `cid`.. 
Look at this function:
```kotlin
public fun startLiveLocationSharing(
    cid: String,
    latitude: Double,
    longitude: Double,
    deviceId: String,
    endAt: Date,
): Call<Location> = Location(
```
We already have 5 parameters... 😫

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/ThrottlingPlugin.kt
andremion: Let's keep it as is for now.
We can review it with @JcMinarro when he is back

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
andremion: I will move the call to the demo for now

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/StateRegistry.kt
andremion: I will revert the initial logic and reintroduce the `currentUserActiveLiveLocations` as the entire logic relies on that

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: The let's keep it like this - we can always introduce an override on this method using `id` + `type`.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/StateRegistry.kt
VelikovPetar: Is this still needed? Is it not possible to use the `live_locations` from the `Channel` object?

VelikovPetar (approved):

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/StateRegistry.kt
VelikovPetar: I will resolve this remark - it is not a blocker and is just an improvement suggestion.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
VelikovPetar: I will resolve this remark - it is not a blocker and is just an improvement suggestion.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/StateRegistry.kt
andremion: I would ask @JcMinarro to understand the point of it

stream-chat-android-state/src/test/java/io/getstream/chat/android/state/internal/SyncManagerTest.kt
```diff
@@ -207,24 +213,27 @@ internal class SyncManagerTest {
...
         val message2 = randomMessage(deletedAt = localDate())
         val message3 = randomMessage(deletedAt = null)
         val message4 = localRandomMessage()
-        whenever(repositoryFacade.selectMessageIdsBySyncState(SYNC_NEEDED)) doReturn listOf(
+        whenever(repositoryFacade.selectMessageIdsBySyncState(SyncStatus.SYNC_NEEDED)) doReturn listOf(
             message1.id, message2.id, message3.id,
         )
-        whenever(repositoryFacade.selectMessageIdsBySyncState(AWAITING_ATTACHMENTS)) doReturn listOf(message4.id)
+        whenever(repositoryFacade.selectMessageIdsBySyncState(SyncStatus.AWAITING_ATTACHMENTS)) doReturn
+            listOf(message4.id)
```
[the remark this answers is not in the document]
andremion: It can't because of the max line constraints

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/SharedLocationItem.kt
```diff
@@ -0,0 +1,286 @@
...
+        MapBox(
+            modifier = Modifier.aspectRatio(1f),
+            latitude = location.latitude,
+            longitude = location.longitude,
+            onClick = onMapClick,
+        ) {
+            val animatedPadding by rememberInfiniteTransition().animateFloat(
+                initialValue = 0f,
+                targetValue = if (isLiveLocationEnded) 0f else 10f,
+                animationSpec = infiniteRepeatable(animation = tween(AnimationDurationMillis)),
```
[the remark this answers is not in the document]
andremion: It's because it's a private constant

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/MessageMapper.kt
```diff
@@ -131,6 +132,8 @@ internal fun Message.toEntity(): MessageEntity = MessageEntity(
         messageTextUpdatedAt = messageTextUpdatedAt,
         pollId = poll?.id,
         reminder = reminder?.toEntity(),
+        restrictedVisibility = restrictedVisibility,
+        sharedLocation = sharedLocation?.toEntity(),
```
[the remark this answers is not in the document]
andremion: Wrong review

PR #5853, opened 2025-06-26
Polish MessageList.

Sagar0-0:
- Rearrange parameters in their expected order.


andremion:
Thanks for the contribution!
While we appreciate the intention to improve API clarity, we try to avoid introducing breaking changes—especially ones that don’t bring tangible value or functional improvement to the library. Even small changes in parameter order can cause unnecessary friction for users upgrading their code.
For now, we’ll close this PR to maintain stability and avoid disrupting existing usage.
Thanks again for your effort and understanding!

PR #5854, opened 2025-06-27
Prevent focusing message if the listState flow was cancelled before a message item was loaded.

VelikovPetar:
### 🎯 Goal

Resolves: https://linear.app/stream/issue/AND-634/upgrade-to-latest-stream-sdk-causing-illegalstateexception-crash-in

Prevents calling `focusMessage` in the `init` block of the `MessageListController` if the `listState` flow was cancelled before a message was loaded. The following setup in the `initialFocusMessage()` method:
```kotlin
private fun initialFocusMessage() {
   messageId ?: return // No initial focus if no message id is provided
   scope.launch {
       listState
         .onCompletion {
            // ...
         }
         .first { it.messageItems.isNotEmpty() }
   }
}
```
behaves in the following way: The `onCompletion` block will be invoked in two scenarios:
- When the flow is completed (by `.first { it.messageItems.isNotEmpty() }`)
- When the flow is cancelled (by cancelling the `scope`)

We should prevent calling the `focusChannelMessage` from the `onCompletion` block in the case when the flow cancelled, because the UI might not be present anymore OR no messages are loaded yet OR the user could already be disconnected. 

**Note: The crash stack trace is: ChatClient::connectUser() must be called before resolving any dependency - but the crash is triggered much earlier in the flow in the `focusMessage` method, where we attempt to focus a message when we cancel the MessageList scope.**

(See attached videos and testing for reproduction steps)

### 🛠 Implementation details
- Prevent calling `focusChannelMessage` from the `onCompleted` block in the `MessageListController.initialFocusMessage()` message.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/a65a8aff-de4a-4f23-aacd-281cba74e523" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/54680b6d-61bd-46c0-80ed-edab2dd960fc" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

<details>

```
Subject: [PATCH] crash investigation
---
IDEA additional info:
<+>UTF-8
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt	(revision abeb6061ec27faab00e208f4ad4c4691b074b13f)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt	(date 1751013421511)
@@ -77,6 +77,7 @@
 import io.getstream.chat.android.compose.ui.theme.ChannelOptionsTheme
 import io.getstream.chat.android.compose.ui.theme.ChatTheme
 import io.getstream.chat.android.compose.ui.threads.ThreadList
+import io.getstream.chat.android.compose.ui.util.getLastMessage
 import io.getstream.chat.android.compose.viewmodel.channels.ChannelListViewModel
 import io.getstream.chat.android.compose.viewmodel.channels.ChannelViewModelFactory
 import io.getstream.chat.android.compose.viewmodel.mentions.MentionListViewModel
@@ -130,9 +131,9 @@
          */
         setContent {
             var selectedTab by rememberSaveable { mutableStateOf(AppBottomBarOption.CHATS) }
-            val globalState = ChatClient.instance().globalState
-            val unreadChannelsCount by globalState.channelUnreadCount.collectAsState()
-            val unreadThreadsCount by globalState.unreadThreadsCount.collectAsState()
+            // val globalState = ChatClient.instance().globalState
+            // val unreadChannelsCount by globalState.channelUnreadCount.collectAsState()
+            // val unreadThreadsCount by globalState.unreadThreadsCount.collectAsState()
             ChatTheme(
                 dateFormatter = ChatApp.dateFormatter,
@@ -149,8 +150,8 @@
                     modifier = Modifier.systemBarsPadding(),
                     bottomBar = {
                         AppBottomBar(
-                            unreadChannelsCount = unreadChannelsCount,
-                            unreadThreadsCount = unreadThreadsCount,
+                            unreadChannelsCount = 0,
+                            unreadThreadsCount = 0,
                             selectedOption = selectedTab,
                             onOptionSelected = { selectedTab = it },
                         )
@@ -346,7 +347,7 @@
             MessagesActivity.createIntent(
                 context = this,
                 channelId = channel.cid,
-                messageId = null,
+                messageId = channel.getLastMessage(ChatClient.instance().getCurrentUser())?.id,
                 parentMessageId = null,
             ),
         )
Index: stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/BaseConnectedActivity.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/BaseConnectedActivity.kt b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/BaseConnectedActivity.kt
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/BaseConnectedActivity.kt	(revision abeb6061ec27faab00e208f4ad4c4691b074b13f)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/BaseConnectedActivity.kt	(date 1751010131065)
@@ -28,9 +28,9 @@
         // The application was killed in the background. Starting from the launcher activity.
-        if (savedInstanceState != null && lastNonConfigurationInstance == null) {
-            startActivity(packageManager.getLaunchIntentForPackage(packageName))
-            finishAffinity()
-        }
+        // if (savedInstanceState != null && lastNonConfigurationInstance == null) {
+        //     startActivity(packageManager.getLaunchIntentForPackage(packageName))
+        //     finishAffinity()
+        // }
     }
 }
```

</details>

Currently only reproduce-able after a process death
1. Set background limit process to "No background processes"
2. Apply the provided patch
3. Open compose app
4. Open a channel (note that the channel auto focuses the last message - change from patch)
5. Put app in background
6. Open a different app
7. Return to compose sample
8. The app should be restored on the `MessageList` in loading state (because we don't have reconnect logic after process death)
9. Tap back (this cancels the `scope`)
10. The app should not crash


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -637,6 +637,8 @@ public class MessageListController(
         scope.launch {
             listState
                 .onCompletion {
+                    // Prevent focusing if the list is empty (if the listState flow is cancelled)
```
andremion: I know it's not related to this issue.
However, can we have `initialFocusMessage` as a suspend function and remove launching another coroutine here?

andremion:
@VelikovPetar , if no objections, I would also remove `BaseConnectedActivity` in this PR

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
VelikovPetar: I believe so, let me just re-test it first.

Edit: Should be OK, everything seems to work fine.

PR #5863, opened 2025-07-10
Fix media gallery overriding initially passed attachments

VelikovPetar:
### 🎯 Goal
Ensures the `MediaGalleryActivity` uses the `Attachments` that were passed via the `MediaGalleryPreviewContract.Input`.
Resolves: https://linear.app/stream/issue/AND-643/mediagallerypreviewviewmodel-overrides-initial-data-from-getmessage

### 🛠 Implementation details
In the `MediaGalleryPreviewViewModel` we are currently overriding the initially passed message with a fresh one from back-end. This can be problematic, if the passed message has its `Attachments` modified or overridden. In this PR, we ensure that the passed attachments always have priority over those in the `Message` fetched from the server.

### 🧪 Testing
1. Apply the provided patch
2. Build compose sample
3. Open a channel
4. Send a video
5. Open the video
6. Observe the logs with tag: `StreamMediaPlayer`: The updated URL should not be overridden by the original one

<details>

<summary>Override the attachment URLs</summary>

```
Index: stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamMediaPlayerContent.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamMediaPlayerContent.kt b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamMediaPlayerContent.kt
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamMediaPlayerContent.kt	(revision 5cb5a5991c8464ea05df28a2997fc335cfa63295)
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/StreamMediaPlayerContent.kt	(date 1752166730582)
@@ -17,6 +17,7 @@
 package io.getstream.chat.android.compose.ui.attachments.preview.internal
 
 import android.content.Context
+import android.util.Log
 import android.view.LayoutInflater
 import androidx.annotation.OptIn
 import androidx.compose.foundation.background
@@ -100,6 +101,7 @@
     LaunchedEffect(player, assetUrl) {
         if (player != null && assetUrl != null) {
             player?.setMediaItem(MediaItem.fromUri(assetUrl))
+            Log.d("StreamMediaPlayer", "Preparing player with asset URL: $assetUrl")
             player?.prepare()
             player?.playWhenReady = playWhenReady
         }
IDEA additional info:
<+>UTF-8
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt	(revision 5cb5a5991c8464ea05df28a2997fc335cfa63295)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt	(date 1752166730576)
@@ -20,6 +20,7 @@
 import android.content.Intent
 import android.os.Bundle
 import android.util.Log
+import androidx.activity.compose.ManagedActivityResultLauncher
 import androidx.activity.compose.setContent
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
@@ -63,7 +64,10 @@
 import io.getstream.chat.android.compose.sample.ui.channel.DirectChannelInfoActivity
 import io.getstream.chat.android.compose.sample.ui.channel.GroupChannelInfoActivity
 import io.getstream.chat.android.compose.sample.ui.component.CustomChatComponentFactory
+import io.getstream.chat.android.compose.state.mediagallerypreview.MediaGalleryPreviewResult
 import io.getstream.chat.android.compose.state.mediagallerypreview.MediaGalleryPreviewResultType
+import io.getstream.chat.android.compose.ui.attachments.StreamAttachmentFactories
+import io.getstream.chat.android.compose.ui.attachments.preview.MediaGalleryPreviewContract
 import io.getstream.chat.android.compose.ui.components.composer.MessageInput
 import io.getstream.chat.android.compose.ui.components.messageoptions.MessageOptionItemVisibility
 import io.getstream.chat.android.compose.ui.components.messageoptions.defaultMessageOptionsState
@@ -90,8 +94,12 @@
 import io.getstream.chat.android.compose.viewmodel.messages.MessageListViewModel
 import io.getstream.chat.android.compose.viewmodel.messages.MessagesViewModelFactory
+import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.models.ReactionSortingByFirstReactionAt
 import io.getstream.chat.android.models.ReactionSortingByLastReactionAt
+import io.getstream.chat.android.ui.common.helper.DownloadAttachmentUriGenerator
+import io.getstream.chat.android.ui.common.helper.DownloadRequestInterceptor
+import io.getstream.chat.android.ui.common.images.resizing.StreamCdnImageResizing
 import io.getstream.chat.android.ui.common.state.messages.MessageMode
 import io.getstream.chat.android.ui.common.state.messages.Reply
 import io.getstream.chat.android.ui.common.state.messages.list.DeletedMessageVisibility
@@ -140,6 +148,38 @@
         val shapes = StreamShapes.defaultShapes()
         val messageComposerTheme = MessageComposerTheme.defaultTheme(isInDarkMode, typography, shapes, colors)
         val ownMessageTheme = MessageTheme.defaultOwnTheme(isInDarkMode, typography, shapes, colors)
+        val attachmentFactories = StreamAttachmentFactories.defaultFactories(
+            onMediaContentItemClick = {
+                    mediaGalleryPreviewLauncher: ManagedActivityResultLauncher<MediaGalleryPreviewContract.Input, MediaGalleryPreviewResult?>,
+                    message: Message,
+                    attachmentPosition: Int,
+                    videoThumbnailsEnabled: Boolean,
+                    downloadAttachmentUriGenerator: DownloadAttachmentUriGenerator,
+                    downloadRequestInterceptor: DownloadRequestInterceptor,
+                    streamCdnImageResizing: StreamCdnImageResizing,
+                    skipEnrichUrl: Boolean,
+                    ->
+                val updatedMessage = message.copy(
+                    attachments = message.attachments.map { attachment ->
+                        attachment.copy(
+                            assetUrl = attachment.assetUrl?.plus("&some_query_arg=true"),
+                            thumbUrl = attachment.thumbUrl?.plus("&some_query_arg=true"),
+                        )
+                    }
+                )
+                mediaGalleryPreviewLauncher.launch(
+                    MediaGalleryPreviewContract.Input(
+                        message = updatedMessage,
+                        initialPosition = attachmentPosition,
+                        videoThumbnailsEnabled = videoThumbnailsEnabled,
+                        downloadAttachmentUriGenerator = downloadAttachmentUriGenerator,
+                        downloadRequestInterceptor = downloadRequestInterceptor,
+                        streamCdnImageResizing = streamCdnImageResizing,
+                        skipEnrichUrl = skipEnrichUrl,
+                    ),
+                )
+            }
+        )
         ChatTheme(
             isInDarkMode = isInDarkMode,
             colors = colors,
@@ -150,6 +190,7 @@
             autoTranslationEnabled = ChatApp.autoTranslationEnabled,
             isComposerLinkPreviewEnabled = ChatApp.isComposerLinkPreviewEnabled,
             allowUIAutomationTest = true,
+            attachmentFactories = attachmentFactories,
             messageComposerTheme = messageComposerTheme.let {
                 it.copy(
                     attachmentCancelIcon = it.attachmentCancelIcon.copy(

```

</details>


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
```diff
@@ -133,14 +152,15 @@ public class MediaGalleryPreviewViewModel(
...
+
                 is ConnectionState.Connecting -> this.connectionState = connectionState
                 is ConnectionState.Offline -> this.connectionState = connectionState
             }
         }
     }
 
     private suspend fun onConnected() {
-        if (message.id.isEmpty() || !hasCompleteMessage) {
+        if (freshMessage.value?.id.isNullOrEmpty() || !hasCompleteMessage) {
```
andremion: Is there any possibility `hasCompleteMessage` is `true` and `freshMessage` is not fetched yet?
Checking `hasCompleteMessage` only isn't enough?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewScreen.kt
```diff
@@ -727,7 +727,10 @@ internal fun MediaGalleryPreviewTitle(
             )
         }
 
-        Timestamp(date = message.updatedAt ?: message.createdAt ?: Date())
+        val timestamp = message.updatedAt ?: message.createdAt
+        if (timestamp != null) {
+            Timestamp(date = timestamp)
+        }
```
andremion, line 730: Is this another bug fix? Should it be listed in the CHANGELOG?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
```diff
@@ -178,22 +198,85 @@ public class MediaGalleryPreviewViewModel(
...
+                    chatClient.updateMessage(message = it).enqueue()
                 }
             }
         }
     }
+
+    private fun deleteMessage() {
+        chatClient.deleteMessage(message.id).enqueue { result ->
+            if (result is Result.Success) {
+                freshMessage.value = result.value
```
andremion, line 295: I see that the `freshMessage` update triggers an update on the `message` state, which the UI checks for deletion to finish the Activity.
However, the new message emission will contain wrong attachments, as it gets from the `initialMessage` (`fresh.copy(attachments = initial.attachments)`).
Given this is a public class (some customers might use it without the Stream Activity), should we make sure the emitted message contains the correct attachments (which in that case must be none)?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
```diff
@@ -178,22 +198,85 @@ public class MediaGalleryPreviewViewModel(
...
+     */
+    internal fun setInitialMessage(message: Message) {
+        initialMessage.value = message
+    }
+
+    private fun deleteAttachmentFromMessage(attachment: Attachment, skipEnrichUrl: Boolean) {
+        val initialMessage = initialMessage.value
+        val freshMessage = freshMessage.value
+
+        if (initialMessage != null && freshMessage != null) {
```
andremion, line 219: I'm not sure if all the condition branches are needed 🤔 
In which case do we have:
- `initialMessage == null, freshMessage != null`
- `initialMessage != null, freshMessage == null`

?

I couldn't delete an attachment in offline mode 🤔

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewScreen.kt
VelikovPetar: Nice catch, I will add it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
VelikovPetar: Hm I believe that deleting the result of the `deleteMessage` operation would return the 'original' message, but with the `deleted_at` property set to timestamp of the deletion. I will re-test this, but I think that the attachments in the message will not be affected by the deletion.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
VelikovPetar: Yes, this could be redundant, but I wanted to make sure that every potential scenario is covered. 
- `initialMessage == null, freshMessage != null` -> this would be a valid scenario when a customer uses the `MediaGalleryPreviewViewModel` directly, without using our Activity which sets the passed `Message`. In such case, the `initialMessage` would be null, and the VM would only know the `messageId`, and use it to fetch the `freshMessage`.
- `initialMessage != null, freshMessage == null` -> This would be a valid scenario if the delete operation is called before we fetch the `freshMessage`, but as you said, this is not a possible scenario when using the SDK `MediaGalleryActivity`. But again, in theory, this could happen if a customer is using the VM but not the Activity.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
VelikovPetar: I believe the `if (!hasCompleteMessage)` check should be enough.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
andremion: I see.
In that case, what do you think about refactoring that function to prevent duplication and reuse some code?
Something like this:
```kotlin
    val initial = initialMessage.value
    val fresh = freshMessage.value

    if (initial != null && fresh != null) {
        removeAttachmentAndUpdate(initial, { this.initialMessage.value = it })
        removeAttachmentAndUpdate(fresh, { this.freshMessage.value = it }, updateServer = true)
    } else if (initial != null) {
        removeAttachmentAndUpdate(initial, { this.initialMessage.value = it })
    } else if (fresh != null) {
        removeAttachmentAndUpdate(fresh, { this.freshMessage.value = it }, updateServer = true)
    }
```

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/mediapreview/MediaGalleryPreviewViewModel.kt
andremion: That's correct. `deleteMessage` only deletes the message, not the last attachment.

PR #5866, opened 2025-07-14
Remove `@ExperimentalStreamChatApi` annotation usage

andremion:


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/chats/ChatsScreen.kt
```diff
@@ -121,7 +120,6 @@ import kotlin.math.abs
  * @param infoContent The content to display in the optional info pane given the provided arguments.
  */
 @OptIn(ExperimentalMaterial3Api::class)
-@ExperimentalStreamChatApi
 @Suppress("LongMethod")
 @Composable
 public fun ChatsScreen(
```
VelikovPetar, line 125: Just to verify: We are intentionally making the `ChatsScreen` public right?

andremion: Yes 👍🏻

PR #5871, opened 2025-07-21
Apply `ThrottlingPluginFactory` by default

andremion:


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/Plugin.kt
```diff
@@ -411,9 +411,13 @@ public interface Plugin :
...
 
-    public fun onUserSet(user: User)
+    public fun onUserSet(user: User) {
+        /* No-Op */
+    }
 
-    public fun onUserDisconnected()
+    public fun onUserDisconnected() {
+        /* No-Op */
+    }
```
JcMinarro, line 420: Why do we need to implement it now on the interface?
It shouldn't be implemented here because it is normally a methdo needed to be implemented on the plugin subclass (They are used to init/destroy the plubing itself).
If the new plugin doesn't need it because there isn't an initialization needed, it should impelement it as "empty method", but keeping the interface without default implementation

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/ThrottlingPlugin.kt
```diff
@@ -31,16 +28,6 @@ internal class ThrottlingPlugin(
     private val lastMarkReadMap: MutableMap<String, Long> = mutableMapOf()
```
JcMinarro: This map shouldn't be needed anymore

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/Plugin.kt
andremion: Why should we force clients to implement it?
Apart from the `ThrottlePlugin`, there are 3 concrete classes where only one implements it. Everything else is empty. 

I'm thinking more about this:

> Interface Segregation Principle (ISP): Clients should not be forced to depend on methods they do not use.... If a class implements an interface with many methods, and the client only needs a few of them, the client should not be forced to depend on the entire interface

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/ThrottlingPlugin.kt
andremion: Good catch ;)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/Plugin.kt
JcMinarro: As I said, They are important methods to initialize/destroy elements within the plugin.
On the case of `ThrottlePlugin` those method should be used to init/clean internal maps used within the plugin

andremion: I'm not saying they're not important. They should be implemented optionally.
For example, the `onCreate` function of an `Activity` is one of the most important and widely used functions in Android.
Despite this, its implementation is optional.

JcMinarro (approved):

PR #5877, opened 2025-07-24
Feature/limit number of messages within state plugin

JcMinarro:
### 🎯 Goal
Limit the number of messages that are kept in memory whil an user is within a LiveStream Channel.
It is typical that tons of messages are sent in a short period of time while a live event is alive, what increase a lot the memory used by the SDK to show those messages. With this PR we introduce a new configuration that our customers can use to filter out old messages. By default this implementation is only used for Livestream channel types, but customers can configure it adding/removing new channel types.

Apart of that, an extra config has been added to the Offline Plugin to avoid the insert process when a message belong to a livestream channel type. Again, it is a config that our customers can override.

Complete: AND-675


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
```diff
@@ -113,6 +115,9 @@ internal class DatabaseMessageRepository(
         if (messages.isEmpty()) return
         val validMessages = messages
             .filter { message -> message.cid.isNotEmpty() }
+            .filterNot { message ->
+                message.type.ifEmpty { message.cid.substringBefore(":") } in ignoredChannelTypes
```
VelikovPetar: Just for my understanding: Why do we first check the `Message.type`, and only if it is empty we fallback to the `Message.cid`? Shouldn't we only check the `Message.cid`?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/config/StatePluginConfig.kt
```diff
@@ -25,10 +25,23 @@ import io.getstream.chat.android.models.TimeDuration
...
+)
+
+/**
+ * Configuration for message limits in channels.
+ *
+ * @param channelTypes The set of channel types for which the limit applies.
+ * @param limit The maximum number of messages to keep in memory for the channel.
+ */
+public data class MessageLimitConfig(
+    public val channelTypes: Set<String> = setOf("livestream"),
```
VelikovPetar: Would it make sense to make the limit configurable per channel type? Something along the lines of:
```kotlin
public data class MessageLimitConfig(
    public val limits: Map<String, Int> = mapOf("livestream" to 1000),
)
```
It would allow for customers to configure different limits for different channel types, for example "livestream" => 1000, and (hypothetical)"live_chat" => 300, which is something that would not be possible with the current implementation.

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
JcMinarro: `Message.type` should contain the channel type, but by default it is `""` 
Because on the previous `filter` we are filtering out messages with empty cid, I fallback on this property, but only in the case `type` is empty, because if we use `cid` we need to process the substring on all messages.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/config/StatePluginConfig.kt
JcMinarro: We can, but I think it over-complicate config our customer needs to add.
If we want to go on that direcction, instead of using a `Map<String, Int>` we should use a more concrete class defining it, something like:
```
data class ChannelMessageLimit(
  val channelType: String,
  val limit: Int,
)
```
And then accept a `Set<ChannelMessageLimit>`
WDYT?

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
VelikovPetar: Wait I am a bit confused, wouldn't the `Message.type` be one of the `regular`, `system` etc... Would it really contain the channel type?

JcMinarro: Oh wait, good catch. I wanted to avoid pre-processing of every `cid` and I mismatch the properties for both entities :man_facepalming: 
Let me change it. We will need to process the cid of all messages :disappointed:

PR #5887, opened 2025-08-05
Feature/simple livestream channel optimization poc

JcMinarro:
### 🎯 Goal
Resolves: https://linear.app/stream/issue/AND-691/livestream-appropriate-channel-with-message-limit

### 🛠 Implementation details
- Adds support for limiting the number of messages in the `ChannelMutableState`
- Dynamic limit adjustment when the user tries to load older messages

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

<details>

```
Provide the patch code here
```

</details>


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -538,6 +545,7 @@ internal class ChannelLogic(
                 when (event) {
                     is NewMessageEvent -> {
                         upsertEventMessage(event.message)
+                        applyMessageLimitIfNeeded(event.message)
```
JcMinarro: Couldn't we perform a call to this method whenever the "Messages List" is updated with newer messages?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -389,11 +400,7 @@ internal class ChannelLogic(
     internal fun upsertMessage(message: Message) {
         logger.d { "[upsertMessage] message.id: ${message.id}, message.text: ${message.text}" }
         channelStateLogic.upsertMessage(message)
-    }
-
-    internal fun upsertMessages(messages: List<Message>) {
-        logger.d { "[upsertMessages] messages.size: ${messages.size}" }
-        channelStateLogic.upsertMessages(messages)
```
JcMinarro: Why did we remove "upsertMessages" method? :thinking: 
Are we always inserting messages one by one? :thinking:

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -688,9 +697,32 @@ internal class ChannelLogic(
         }
     }
 
-    fun toChannel(): Channel = mutableState.toChannel()
+    private fun applyMessageLimitIfNeeded(message: Message) {
+        // If no message limit is set, restriction is not applied
+        if (messageLimit == null) return
+        // Re-apply restriction if current user sent the message (most likely the message list is scrolled to bottom)
+        resetLoadOlderMessagesCounter(message)
```
JcMinarro: The name of this method is a bit confusing, it looks like it will be resetting it "always", but internally it is checking if the message is from the current user

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -688,9 +697,32 @@ internal class ChannelLogic(
...
+        trimOldMessagesIfNeeded()
+    }
+
+    private fun trimOldMessagesIfNeeded() {
+        if (messageLimit == null) return
+        if (!trimOldMessages) {
+            logger.d { "[trimOldMessagesIfNeeded] trimming disabled, count: $loadOlderMessagesCount" }
+            return
+        }
+        channelStateLogic.trimOldMessagesIfNeeded(keep = messageLimit)
```
JcMinarro: If the list of channels lives within `ChannelStateLogic`, why don't we move this logic to this class?

stream-chat-android-state/src/test/java/io/getstream/chat/android/state/channel/controller/WhenHandleEvent.kt
```diff
@@ -128,25 +127,6 @@ internal class WhenHandleEvent : SynchronizedCoroutineTest {
...
-            showInChannel = true,
-        )
-        channelLogic.upsertMessages(listOf(message))
-
-        val messageUpdateEvent = randomMessageUpdateEvent(message = message)
-
-        channelLogic.handleEvent(messageUpdateEvent)
-
-        verify(channelStateLogic).upsertMessages(listOf(messageUpdateEvent.message))
-    }
```
JcMinarro: Isn't this test useful anymore?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
VelikovPetar: I kept it isolated to the cases where an actual new message would be added to the state, to avoid calling it in case where a message is only updated (events such as `message.updated`). Are you thinking about calling it whenever we call `channelStateLogic.upsertMessage(message)?`

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
VelikovPetar: Ah my mistake - this method was only used in a test (which I also removed in they PR), but now I see that the test doesn't really test this method - but rather uses it to setup the test, and tests the `handleEvent` method.. I will revert this

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
VelikovPetar: Good point, I will rename it

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
VelikovPetar: Are you thinking about moving the whole state management to the `ChannelStateLogic` (including the `loadOlderMessagesCount` logic)?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
VelikovPetar: We can also remove this method, and use the `upsertMessage()` to setup the test. WDYT?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
JcMinarro: Well, I was thinking to have a centralized place where it is called, but not be choosing the different places were we think new messages should be added.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
JcMinarro: As we were discussing, `loadOlderMessagesCount` should not be needed.
Within the `MutableChannelState` we already have some flags to know if a `loadOlderMessages` query is executed or not, I think the window-limit can be applied within this class and keep the rest of them only handling "the real logic"

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -274,8 +276,36 @@ internal class ChannelMutableState(
...
+                    messageLimit = null
+                }
+            } else {
+                // we are still within the limit, so we can keep it
+                logger.d {
+                    "[setLoadingOlderMessages] Keeping message limit, " +
+                        "current size: $currentMessageCount, limit: $messageLimit, buffer: $TRIM_BUFFER"
+                }
+            }
+        }
```
JcMinarro: Can't we simplify it?
- `pageSize` should't be needed. I know it can be modified by our customers, but normally, the default value (30) is used. Apart from that, this method will be called before the query is processed, just at the moment it is called, maybe the limit is enough, but before the query is completed, multiple messages can arrive by WS and overpass the limit before the query is completed.
- If the limit is exceeded the first time and we increase it by 1.5, why do we disable it the second time instead of increasing it again?

I think we can simplify this logic with a more heuristic approach, something like:
```
val currentMessageCount = _messages?.value?.size ?: return
if (isLoading) {
    messageLimit = messageLimit?.let {
        val multiplier = LIMIT_MULTIPLIER.takeUnless {
            currentMessageCount + TRIM_BUFFER >= it
        } ?: NEUTRAL_MULTIPLIER
        (it * multiplier).toInt()
    }
}
```

This way we increase the limit if messageLimit is almost overpassed, keep increasing it while the user is loading more messages, but don't disable the limit

VelikovPetar: Yeah I think that this approach would work as well - I will simplify it and we can discuss it tomorrow!

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt
```diff
@@ -114,7 +114,7 @@ class ChannelsActivity : ComponentActivity() {
                 .descByName<Channel>("pinned_at") // pinned channels first
                 .desc("last_updated"), // then by last updated
             filters = Filters.and(
-                Filters.eq("type", "messaging"),
+                Filters.`in`("type", "livestream", "messaging"),
```
JcMinarro: It needs to be reverted before we merge it

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -206,6 +215,7 @@ internal class ChannelLogic(
     internal suspend fun loadOlderMessages(messageLimit: Int, baseMessageId: String? = null): Result<Channel> {
         logger.i { "[loadOlderMessages] messageLimit: $messageLimit, baseMessageId: $baseMessageId" }
         channelStateLogic.loadingOlderMessages()
+        // Increment the counter of older messages requests
```
JcMinarro: This comment is not needed anymore

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelStateLogic.kt
```diff
@@ -741,6 +741,13 @@ internal class ChannelStateLogic(
         }
     }
 
+    fun loadingNewestMessages() {
+        // Reset messages if there is a limit set
+        if (mutableState.baseMessageLimit != null) {
+            mutableState.setMessages(emptyList())
+        }
+    }
```
JcMinarro: I think it is from the previous implementation where the logic was living outside of the mutableState. Do we continue needing it?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -51,7 +51,7 @@ internal class ChannelMutableState(
     private val userFlow: StateFlow<User?>,
     latestUsers: StateFlow<Map<String, User>>,
     activeLiveLocations: StateFlow<List<Location>>,
-    private val messagesLimitFilter: (Collection<Message>) -> Collection<Message>,
+    val baseMessageLimit: Int?,
```
JcMinarro, line 70: Maybe add kdocs here about this limit and how it is increased by the multiplier?
Just to ensure in the future we know why it was implemented on that way

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelStateLogic.kt
VelikovPetar: I will double check it

VelikovPetar: Ok, the idea was that when we are loading the latest messages, we can assume that we are re-entering the channel (or opening it for the first time) - so we are using that moment to also clear the already loaded messages, and reset the `messageLimit` => so that if we have previously overshot the limit, to re-initialize it. 

_Note: Entering, exiting and re-entering the channel usually wouldn't result in clearing the previously loaded messages, that's why I introduced this logic._

Edit: We could potentially just reset the limit in this case, which should result in trimming the already loaded messages to fit the initial limit.

JcMinarro: Sound good approach, but, do we really need to drop already fetched messages? :thinking: 
Can't we only reset the limit?
The UX for an user that go away the conversation and go into the vonersation again will be very bad

JcMinarro: Yes, maybe a `resetLimit()` method would be better, handled internally, we don't really need to check if it is null from outside of the mutableState, but within the MutableState we can decide to update the messageLimit to the baseMessageLimit configured on the constructor

VelikovPetar: I don't think we need to drop everything, resetting the limit should be enough. Let me just re-test this manually, and I will make the update.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -637,7 +674,42 @@ internal class ChannelMutableState(
...
+        // If no message limit is set or we are loading older messages, restriction is not applied
+        if (messageLimit == null || loadingOlderMessages.value) {
+            logger.d {
+                "[applyMessageLimitIfNeeded] no need to trim, " +
+                    "messageLimit: $messageLimit, loadingOlderMessages: ${loadingOlderMessages.value}"
+            }
+            return messages
+        }
+        // Add buffer to avoid trimming too often
+        return if (messages.size > messageLimit!! + TRIM_BUFFER) {
```
harrytmthy, line 679: NIT: `messageLimit` is a `var`, yet currently there are multiple functions that could update it, including `setLoadingOlderMessages()` and `resetMessageLimit()`. Might be worth clarifying thread-safety assumptions around `messageLimit!!` 🤔

VelikovPetar: This cannot really fail. 
If the `messageLimit` is initially set, it can never be set to `null`. (the other way around is also true, if it is initially `null`, it can not be set to a concrete value)

PR #5889, opened 2025-08-06
Improve uploaded attachment loading UI

andremion:
### 🎯 Goal

Prevent unnecessary loading UI after uploading attachments.

### 🛠 Implementation details

Introduces a new `imagePreviewData` extension property for `Attachment` to centralize the logic for obtaining image preview data.

This refactoring simplifies the code in various attachment content composables.

- Added `Attachment.imagePreviewData` extension property to handle image preview data retrieval.
- Updated `FileAttachmentContent`, `MediaGalleryPhotosMenu`, `MediaAttachmentQuotedContent`, `MediaAttachmentContent`, and `FileAttachmentQuotedContent` to use the new `imagePreviewData` property.
- Adjusted snapshot test for `MediaAttachmentQuotedContent` due to changes in preview data.

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

<details>

```
Provide the patch code here
```

</details>

### 🎉 GIF


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileAttachmentQuotedContent.kt
```diff
@@ -43,17 +43,8 @@ public fun FileAttachmentQuotedContent(
...
-    val data = if (isImage) {
-        val dataToLoad =
-            attachment.imagePreviewUrl?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
-                ?: attachment.upload
-
-        dataToLoad
-    } else {
-        MimeTypeIconProvider.getIconRes(attachment.mimeType)
-    }
+    val isImage = attachment.isImage()
```
VelikovPetar, line 46: Small question here: previously we only check if the `attachment.type == "image"`, but now the `attachment.imagePreviewData` will also check if the `attachment` is of type `video`. is this change not breaking anything?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/MediaAttachmentContent.kt
```diff
@@ -409,17 +408,7 @@ internal fun MediaAttachmentContentItem(
...
-            when (message.syncStatus) {
-                SyncStatus.COMPLETED ->
-                    attachment.imagePreviewUrl?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
-
-                else -> attachment.upload
-            }
-        } else {
-            null
-        }
+    val data = attachment.imagePreviewData
```
VelikovPetar, line 411: Similar question here: The new `attachment.imagePreviewData` no longer considers the `message.syncStatus` - is that OK?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileAttachmentQuotedContent.kt
andremion: Actually, it's an improvement.
The previous logic ignored video thumbnails.
Now it's considered.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/MediaAttachmentContent.kt
andremion: That's the actual fix.
Even before the sync is completed, we already have enough data to render the image.
No need to wait until it's completed.
That's the reason for the issue.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileAttachmentQuotedContent.kt
VelikovPetar: Aha I see, nice!

VelikovPetar (approved):

PR #5895, opened 2025-08-12
Add `user.messages.deleted` event handling

JcMinarro:
### 🎯 Goal
Adds handling for the new `user.messages.deleted` event. This event is (optionally) fired after banning a user, if the "Delete messages for user" is selected. The handling involves deleting all messages related to the User in the SDK state/database.
Resolves: https://linear.app/stream/issue/AND-735/add-handling-for-usermessagesdeleted-event

### 🛠 Implementation details
1. Introduce the `UserMessagesDeletedEvent` (and its DTO and parsing logic)
2. Introduce handling of the event in state layer: Delete user-related messages in `ChannelState`
3. Introduce handling of the event in offline layer: Delete user-related messages in the DB

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Scenario</th>
<th>Banned User</th>
<th>Other User</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Channel Ban
</td>
<td>
<video src="https://github.com/user-attachments/assets/ad3cca0b-1d17-49bb-a625-9f0e94fb384e" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/54553323-ae88-486f-8547-f3b2655aa31c" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
Global Ban
</td>
<td>
<video src="https://github.com/user-attachments/assets/546447b8-fb46-42b9-93a4-42fbdafc5dcb" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/380b5fbd-e3bd-4e2c-99bf-065e992a9e91" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
Ban a user from the dashboard, following the steps:
1. Open Dashboard
2. Open `Moderation`
3. Select `Users` or `Text`
4. Find the message/user that you want to ban (if you cannot find it, you could `Flag` the message from the sample app)
5. Hover over the user/message
6. Select either `Channel ban` or `Global ban`
7. In the prompt, select `Soft delete all the messages of the user`
8. Click `Ban`
9. The event should now be fired to all relevant users
10. Observe the messages getting soft deleted

Patch:  [user_messages_deleted_testing.patch](https://github.com/user-attachments/files/22407453/user_messages_deleted_testing.patch)

**Note: The hardDelete option is not currently available on the dashboard, you might need to do some hardCoding to verify the case**

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 ![stream-chat-android-client](https://img.shields.io/badge/stream--chat--android--client-3.21%20MB-lightgreen)
 ![stream-chat-android-offline](https://img.shields.io/badge/stream--chat--android--offline-3.43%20MB-lightgreen)
 ![stream-chat-android-ui-components](https://img.shields.io/badge/stream--chat--android--ui--components-10.51%20MB-lightgreen)
-![stream-chat-android-compose](https://img.shields.io/badge/stream--chat--android--compose-12.71%20MB-lightgreen)
+![stream-chat-android-compose](https://img.shields.io/badge/stream--chat--android--compose-12.72%20MB-lightgreen)
 
 </div>
 
--- a/metrics/size.json
+++ b/metrics/size.json
@@ -2,13 +2,13 @@
   "debug": {
     "stream-chat-android-client": 4376,
     "stream-chat-android-offline": 4708,
-    "stream-chat-android-ui-components": 13252,
-    "stream-chat-android-compose": 16460
+    "stream-chat-android-ui-components": 13256,
+    "stream-chat-android-compose": 16480
   },
   "release": {
     "stream-chat-android-client": 3292,
     "stream-chat-android-offline": 3520,
     "stream-chat-android-ui-components": 10768,
-    "stream-chat-android-compose": 13016
+    "stream-chat-android-compose": 13032
   }
 }


stream-chat-android-state/src/test/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableStateTests.kt
```diff
@@ -377,6 +376,188 @@ internal class ChannelMutableStateTests {
...
+    @Test
+    fun `getMessagesFromUser should return messages in original order`() = runTest {
+        // given
+        val user = User(id = "user1", name = "User 1")
+        val now = currentTime()
+
+        val messages = listOf(
+            randomMessage(id = "msg1", cid = CID, user = user, createdAt = Date(now + 1000)),
+            randomMessage(id = "msg2", cid = CID, user = user, createdAt = Date(now + 2000)),
+            randomMessage(id = "msg3", cid = CID, user = user, createdAt = Date(now + 3000)),
```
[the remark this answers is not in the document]
VelikovPetar: not relevant

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -495,13 +495,25 @@ internal class ChannelMutableState(
...
+
+    fun deleteMessages(messages: List<Message>) {
+        val messageIds = messages.map { it.id }.toSet()
+        logger.v { "[deleteMessages] messages.ids: $messageIds" }
+        _messages?.apply { value = value - messageIds }
+        deletePinnedMessages(messages)
     }
 
     fun deletePinnedMessage(message: Message) {
         logger.v { "[deletePinnedMessage] message.id=${message.id}, message.text=${message.text}" }
```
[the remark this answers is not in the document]
VelikovPetar: not needed

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -495,13 +495,25 @@ internal class ChannelMutableState(
 
     fun deleteMessage(message: Message) {
         logger.v { "[deleteMessage] message.id: ${message.id}" }
```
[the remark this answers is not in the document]
VelikovPetar: not needed

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/message/internal/MessageDao.kt
```diff
@@ -230,6 +230,12 @@ internal interface MessageDao {
     @Transaction
     suspend fun selectBySyncStatus(syncStatus: SyncStatus, limit: Int = NO_LIMIT): List<MessageEntity>
 
+    @Query("SELECT * FROM $MESSAGE_ENTITY_TABLE_NAME WHERE userId = :userId")
+    suspend fun selectByUserId(userId: String): List<MessageEntity>
+
+    @Query("SELECT * FROM $MESSAGE_ENTITY_TABLE_NAME WHERE cid = :cid AND userId = :userId")
+    suspend fun selectByCidAndUserId(cid: String, userId: String): List<MessageEntity>
```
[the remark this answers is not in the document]
VelikovPetar: We must fetch all of them unfortunately.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -840,6 +872,24 @@ internal class EventHandlerSequential(
...
+            repos.selectMessagesForUser(userId)
+        }
+        if (hard) {
+            // Remove messages from DB
+            messages.forEach { repos.deleteChannelMessage(it) }
+        } else {
+            // Mark messages as deleted
+            val markedAsDeleted = messages.map { it.copy(deletedAt = deletedAt) }
+            repos.insertMessages(markedAsDeleted)
+        }
```
gpunto, line 892: Just passing by: can't we do a delete query (hard=true) or an update query (hard=false) instead of a select + insert?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
```diff
@@ -51,6 +51,21 @@ public interface MessageRepository {
         limit: Int,
     ): List<Message>
 
+    /**
+     * Selects all messages from a given user.
+     *
+     * @param userId The user ID to filter messages.
+     */
+    public suspend fun selectMessagesForUser(userId: String): List<Message>
```
andremion:
```suggestion
    public suspend fun selectAllUserMessages(userId: String): List<Message>
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
```diff
@@ -51,6 +51,21 @@ public interface MessageRepository {
...
+
+    /**
+     * Selects all messages from a given user in a given channel.
+     *
+     * @param cid The channel ID to filter messages.
+     * @param userId The user ID to filter messages.
+     */
+    public suspend fun selectMessagesInChannelForUser(cid: String, userId: String): List<Message>
```
andremion:
```suggestion
    public suspend fun selectAllChannelUserMessages(cid: String, userId: String): List<Message>
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/EventMapping.kt
```diff
@@ -1144,6 +1147,19 @@ internal class EventMapping(
         )
     }
 
+    private fun UserMessagesDeletedEventDto.toDomain(): UserMessagesDeletedEvent = with(domainMapping) {
+        return UserMessagesDeletedEvent(
+            type = type,
+            createdAt = created_at.date,
+            rawCreatedAt = created_at.rawDate,
```
andremion, line 1154: What do we need `rawCreatedAt` for?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
andremion: Another option would be `selectAllUserMessages` with an optional `cid`.
If it's not null, it means all channel user messages.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelLogic.kt
```diff
@@ -696,7 +697,8 @@ internal class ChannelLogic(
             is NotificationChannelMutesUpdatedEvent -> event.me.channelMutes.any { mute ->
                 mute.channel?.cid == mutableState.cid
             }.let(channelStateLogic::updateMute)
-
+            is UserMessagesDeletedEvent ->
+                channelStateLogic.deleteMessagesFromUser(event.user.id, event.hardDelete, event.createdAt)
```
andremion: Can you use parameter names there?
So that we can see that `createdAt` is used as `deletedAt`

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -230,6 +232,25 @@ internal class EventHandlerSequential(
         scope.coroutineContext.job.cancelChildren()
     }
 
+    private fun handleUserMessagesDeletedEvent(event: UserMessagesDeletedEvent) {
```
andremion, line 235: Given the recent comments and the coverage of this PR, maybe you need to create some unit tests here to increase it and be at least closer to 70%.
WDYT?

andremion (commented):
There are some small name suggestions you can decide if they make sense.
Although I still have to have access to the Moderation dashboard in order to test it.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
VelikovPetar: Yep makes sense

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/MessageRepository.kt
VelikovPetar: I will rename it, although I would keep them as separate methods - in my opinion its more clear what sort of data we fetch

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/EventMapping.kt
VelikovPetar: It is a property in the `ChatEvent` superclass, so we have to provide it. To be honest I am not very familiar with the details, but it is used in the `SyncManager`.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
VelikovPetar: I will add them!

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
VelikovPetar: The thing is that the `repos.deleteChannelMessage` and `repos.insertMessages` are not just accessing the DB, but also updating some caches + accessing tables which are not the `messages` table (ex. channels, for keeping everything in sync). 
I will revisit this, to see what can be done.

gpunto: Ah ok, then don't worry, we can maybe do it as an optimization later if feasible. Not a blocker!

VelikovPetar: Short update here:
I did some analysis, and the updates to the DB itself are rather straight forward queries - we can easily delete or update the `deleteAt` fields in the BD. 

The issues are on the higher level - in the `MessagesRepository/ChannelsRepository/ThreadsRepository`: These repository hold some caches of messages/channels/threads to avoid accessing the DB if possible.
I will only talk about the `MessagesRepository`, but it is quite similar in the other ones as well. This repository holds a cache of messages to avoid accessing the DB if possible + a set of deleted messages ID to optimise some unnecessary operations over the DB. In order to keep this cache updated, we would need to know the affected messages IDs, which room doesn't provide when deleting rows in the table. So we would have to make at least 2 DB queries:
1. Get all IDs that would be deleted (to update the repository cache)
2. Delete all messages by their IDs
With this we could optimise the `hard` delete case.

For the `soft` delete case, we could potentially make a workaround:
1. Update all messages in the cache that fit the criteria (cid + userId)
2. Update the underlying DB

Or we could just clean the cache after such operations, but I would like to avoid doing that...

Let me know what do you think about such approach, or if you have any ideas!

gpunto: If that makes the code complex & harder to understand it might not be worth it. In the end we'd still need to fetch the IDs (which is less data, but we still have to query anyway).

Maybe we could optimize at least this call:
```kotlin
messages.forEach { repos.deleteChannelMessage(it) }
```
to something that deletes all messages at once. I see that each deletion does a fair bit of work, updating the DB and reallocating some lists. But again, it's not a blocker. I'd merge if we're happy with how it's working and optimizations can be done at a later time.

VelikovPetar: Did some optimisations here: https://github.com/GetStream/stream-chat-android/pull/5895/commits/6c9329339f1d291df4baa6f788b53f44e67c1ad6

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
```diff
@@ -109,6 +109,19 @@ internal class DatabaseChannelRepository(
...
+        val messagesPerChannel = messages.groupBy { it.cid }
+        messagesPerChannel.forEach { (cid, messages) ->
+            channelCache[cid]?.let { cachedChannel ->
+                val updatedChannel = cachedChannel.copy(
+                    messages = cachedChannel.messages.filter { msg -> messages.none { it.id == msg.id } },
+                    pinnedMessages = cachedChannel.pinnedMessages.filter { msg -> messages.none { it.id == msg.id } },
+                )
+                cacheChannel(updatedChannel)
+            }
+        }
```
gpunto: There's an optimization we can also apply to reduce filtering from O(n^2) to O(n), i.e. grouping into a map of cid to set of ids, and then use those sets to check.

```suggestion
        val messageIdsPerChannel = messages
            .groupingBy(Message::cid)
            .foldTo(
                destination = mutableMapOf(),
                initialValueSelector = { _, _ -> mutableSetOf<String>() },
                operation = { _, accumulator, element -> accumulator.apply { add(element.id) } }
            )

        messageIdsPerChannel.forEach { (cid, ids) ->
            channelCache[cid]?.let { cachedChannel ->
                val updatedChannel = cachedChannel.copy(
                    messages = cachedChannel.messages.filter { msg -> msg.id !in ids },
                    pinnedMessages = cachedChannel.pinnedMessages.filter { msg -> msg.id !in ids },
                )
                cacheChannel(updatedChannel)
            }
        }
```

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
gpunto: Awesome, thanks!

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/domain/channel/internal/DatabaseChannelRepository.kt
VelikovPetar: Thanks, will update it!

stream-chat-android-client-test/src/main/java/io/getstream/chat/android/client/test/Mother.kt
```diff
@@ -684,6 +685,25 @@ public fun randomChannelUserBannedEvent(
     )
 }
 
+public fun randomUserMessagesDeletedEvent(
+    createdAt: Date = Date(),
+    user: User = randomUser(),
+    cid: String? = randomCID(),
+    hardDelete: Boolean = false,
```
andremion: nit: should use `randomBoolean` for `hardDelete` and `randomDate` for `createdAt` ?

VelikovPetar: oops, nice catch! 
I will update it!

VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-android/pull/5895/commits/5f378223f97f01dd5a8cae137751d3295ec73eb5

VelikovPetar:
Hi @andremion @gpunto ,
I found a bug in the current implementation in a specific scenario where:
1. `hard_delete = true`
2. `OfflinePlugin` is configured to not store messages in the DB for the specific channel type

What was happening was the following:
1. In `EventHandlerSequential.deleteMessagesFromUser` we initially fetch the messages that should be deleted
2. If `hard = true`, we delete them from the DB and the Channels Cache

However, there is an edge case where the DB is disabled for messages (when `StreamOfflinePluginFactory.ignoredChannelTypes = ["messaging"]`): In this case, fetching the messages from the DB will return an empty list. This means that we can no longer delete them from the Channels cache. That's why in [this commit](https://github.com/GetStream/stream-chat-android/pull/5895/commits/73e5f9de2f30d2ec5cd718a5143888a8ed7b40a0), I introduced a different approach for updating the channels cache -> by clearing the messages from the cache that match the `cid`/`userId`. (The removal of the messages from the messages table is still the same). 
It would be great if you can review this logic again!

gpunto (approved):

andremion:
What happens with their messages if a user is unbanned? Should they be "restored"? 🤔

VelikovPetar:
> What happens with their messages if a user is unbanned? Should they be "restored"? 🤔
https://github.com/GetStream/stream-chat-android/pull/5895#issuecomment-3334400194

At the moment, there is no event published in such scenario. If the messages are un-deleted, they would become available on the next reload of the channel

PR #5897, opened 2025-08-12
Support Unread Counts API

andremion:
### 🎯 Goal

Support Unread Counts API

### 🛠 Implementation details

- Introduce `getUnreadCounts` to the `ChatClient` to return the following data structure:

```kotlin
UnreadCounts {
    // The total number of unread messages.
    messagesCount: Int
    // The total number of unread threads.
    threadsCount: Int
    // A map containing the count of unread messages by team ID.
    messagesCountByTeam: Map<String, Int>
    // A list of unread channels with their details.
    channels: List<UnreadChannel>
    // A list of unread threads with their details.
    threads: List<UnreadThread>
    // A list of unread channels grouped by type.
    channelsByType: List<UnreadChannelByType>
}
```

- Introduce the Unread Counts UI to the Compose User Profile screen so that we can showcase unread counts from the API

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>No teams</th>
<th>With teams</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/ec2572a8-9b86-404c-9fea-90c534a479ae" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/a2c96740-7061-45b1-9722-422cb46ec278" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

- Follow the path you see in the videos.

> [!IMPORTANT]
> To test the Counts by Team, you need to use a multi-tenant app with teams.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UnreadDtos.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UnreadDtos.kt
@@ -23,7 +23,7 @@ import java.util.Date
 internal data class UnreadDto(
     val total_unread_count: Int = 0,
     val total_unread_threads_count: Int = 0,
-    val total_unread_count_by_team: Map<String, Int> = emptyMap(),
+    val total_unread_count_by_team: Map<String, Int>? = emptyMap(),
     val channels: List<UnreadChannelDto> = emptyList(),
     val threads: List<UnreadThreadDto> = emptyList(),
     val channel_type: List<UnreadChannelByTypeDto> = emptyList(),
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
@@ -783,7 +783,7 @@ internal class DomainMapping(
     internal fun UnreadDto.toDomain(): UnreadCounts = UnreadCounts(
         messagesCount = total_unread_count,
         threadsCount = total_unread_threads_count,
-        messagesCountByTeam = total_unread_count_by_team,
+        messagesCountByTeam = total_unread_count_by_team.orEmpty(),
         channels = channels.map { it.toDomain() },
         threads = threads.map { it.toDomain() },
         channelsByType = channel_type.map { it.toDomain() },
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTestArguments.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTestArguments.kt
@@ -559,7 +559,7 @@ internal object MoshiChatApiTestArguments {
             val model = UnreadCounts(
                 messagesCount = dto.total_unread_count,
                 threadsCount = dto.total_unread_threads_count,
-                messagesCountByTeam = dto.total_unread_count_by_team,
+                messagesCountByTeam = dto.total_unread_count_by_team!!,
                 channels = dto.channels.map { dto ->
                     UnreadChannel(
                         cid = dto.channel_id,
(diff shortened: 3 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UnreadDtos.kt
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Date
+
+@JsonClass(generateAdapter = true)
+internal data class UnreadDto(
+    val total_unread_count: Int = 0,
+    val total_unread_threads_count: Int = 0,
+    val total_unread_count_by_team: Map<String, Int> = emptyMap(),
+    val channels: List<UnreadChannelDto> = emptyList(),
+    val threads: List<UnreadThreadDto> = emptyList(),
+    val channel_type: List<UnreadChannelByTypeDto> = emptyList(),
```
JcMinarro: Are we sure we are not receiving null values here?
Other DTOs are returning us null values on the collections, which could crash the execution of our SDK.
If that is the case, could we mark it as nullable and properly parse it within our mappers to not allow nullable values in our business model?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
```diff
@@ -2176,6 +2178,25 @@ internal class MoshiChatApiTest {
         assertEquals(location, result.getOrThrow())
     }
 
+    @ParameterizedTest
+    @MethodSource("io.getstream.chat.android.client.api2.MoshiChatApiTestArguments#getUnreadCounts")
```
JcMinarro, line 2182: To avoid IDE lint about "not used method", would be good if you can add a kdocs comment over this test mentioning the method name, something like:
```
/** This method use [MoshiChatApiTestArguments.getUnreadCounts] as a source of arguments. */
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/UnreadDtos.kt
andremion: Actually, there is only one optional attribute, which is `total_unread_count_by_team`:

Image

Everything else is mandatory.

Swift also follows that:

Image

I will update `total_unread_count_by_team`

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
andremion: Sorry, where do you see the `not used method` warning?
I can see `1 Usage` 🤔 
Screenshot 2025-08-19 at 13 18 03
Which Android Studio version are you using?

JcMinarro: :thinking: 
It looks like it has been fixed, and method names within `@MethodSource` annotation are considered as usage of that. It is now even clickable and navigates to the Method.
BTW, The object class itself is marked as not used if we only use a reference to the methods within the annotation, but it is marked as used if we mention it from a kdocs....
Screenshot from 2025-08-19 14-37-50

andremion: I used `@Suppress("UNUSED")` as we have in `EventMappingTestArguments`.
I would say it's fine to suppress UNUSED, instead of adding kdoc just for that case.
However, I can make the change if you think it's important.

JcMinarro: It is fine.
Hopefully they fix it soon and we can remove the `@Suppress("UNUSED")` as well

JcMinarro (approved):

PR #5901, opened 2025-08-18
Consider the attachment title when formatting the quoted message text

andremion:


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/QuotedMessageTextFormatter.kt
```diff
@@ -265,4 +258,14 @@ private class DefaultQuotedMessageTextFormatter(
...
+
+    private fun Attachment.getMessageText(): String? = when {
+        name != null -> name
+        text != null -> text
+        title != null -> title
+        isImage() -> context.getString(R.string.stream_compose_quoted_message_image_tag)
+        isGiphy() -> context.getString(R.string.stream_compose_quoted_message_giphy_tag)
+        isAnyFileType() -> context.getString(R.string.stream_compose_quoted_message_file_tag)
+        else -> null
+    }
```
JcMinarro, line 270:
```suggestion
    private fun Attachment.getMessageText(defaultMessageText: String): String = when {
        name != null -> name
        text != null -> text
        title != null -> title
        isImage() -> context.getString(R.string.stream_compose_quoted_message_image_tag)
        isGiphy() -> context.getString(R.string.stream_compose_quoted_message_giphy_tag)
        isAnyFileType() -> context.getString(R.string.stream_compose_quoted_message_file_tag)
        else -> defaultMessageText
    }
```
Can't we pass the default value as a parameter?

andremion: @JcMinarro , there is an issue there.
Changing the method return to `String` won't work because the compiler will smartcast these branches to `String?
`
```kotlin
name != null -> name
text != null -> text
title != null -> title
```
Screenshot 2025-08-19 at 13 36 38

One option would be to force non-nullables when the conditions are met:
```kotlin
name != null -> name!!
text != null -> text!!
title != null -> title!!
```
To be honest, I would prefer to return `String?` instead of using `!!`
WDYT?

JcMinarro: Ohh, sh*t.... SmartCast is not "enough smart" in that case, even though the attachment properties are immutable :disappointed:

JcMinarro (approved):

PR #5908, opened 2025-08-25
Feature/refactor attachments factories

JcMinarro:
### 🎯 Goal
Old AttachmentFactories only provided a reduced number of data on the onClick lambda that wasn't enought for some of our customers. The new accepted lambda for `Giphy`, `Link` and `Media` attachments provide a holder  object with more data related to the attachment is being clicked.

Fix AND-704


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/EventMapping.kt
```diff
@@ -1140,6 +1146,30 @@ internal class EventMapping(
         )
     }
 
+    private fun ChannelUserMessagesDeletedEventDto.toDomain(): ChannelUserMessagesDeletedEvent = with(domainMapping) {
```
andremion: This is unrelated to the attachments factories, right?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
```diff
@@ -222,3 +284,16 @@ internal fun onGiphyAttachmentContentClick(context: Context, previewUrl: String)
...
+ * Data class that holds information about a giphy attachment click event.
+ *
+ * @param context The context in which the click event occurred.
+ * @param url The URL of the giphy attachment that was clicked.
+ * @param message The message containing the giphy attachment.
+ */
+public data class GiphyAttachmentClickData(
+    val context: Context,
+    val url: String,
+    val message: Message,
```
andremion, line 307: Should we also consider adding the `attachment` object?
What if a customer wants to make a decision based on that?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
```diff
@@ -222,3 +284,16 @@ internal fun onGiphyAttachmentContentClick(context: Context, previewUrl: String)
...
 }
+
+/**
+ * Data class that holds information about a giphy attachment click event.
+ *
+ * @param context The context in which the click event occurred.
+ * @param url The URL of the giphy attachment that was clicked.
+ * @param message The message containing the giphy attachment.
+ */
+public data class GiphyAttachmentClickData(
```
andremion: Should we consider making the construct internal so that we prevent customers from instantiating this object, allowing us to include any properties in the future without having to make them nullable or optional?

andremion (commented):
Good job! 👏🏻 
This PR seems to have changes unrelated to this task.
Can you double-check it?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
JcMinarro: Good point

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
JcMinarro: Let me add it as well

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/EventMapping.kt
JcMinarro: Yep, it is from another PRs.
Looks like I rebased the wrong branch

PR #5909, opened 2025-08-26
Support upload file/image standalone endpoints

andremion:
### 🎯 Goal

Support upload standalone file/image endpoints as per https://github.com/GetStream/stream-chat-js/pull/1564

### 🛠 Implementation details

Introduced `ChatClient.uploadFile`, `ChatClient.deleteFile`, `ChatClient.uploadImage`, and `ChatClient.deleteImage`, to upload/delete files/images that are not related to any channel. 

```kotlin
chatClient.uploadImage(
    file = anImageFile,
).enqueue()
```

Added a functionality to upload the profile picture to the Compose demo app.

> [!NOTE]
> All the updated profile pictures are reset on a new app cold launch or login, so that we keep the original avatars of the demo app.

> [!IMPORTANT]
> It was set `-Xjvm-default=all` to the Kotlin compiler of the client module, so that default implementations of interfaces are available in **Java**, e.g., using the `ProgressCallback` doesn't require implementing all the callback functions.

### 🎨 UI Changes

<video src="https://github.com/user-attachments/assets/b1092121-b778-4130-9712-f8f68639316d" controls="controls" muted="muted" />

### DIFF
--- a/stream-chat-android-client/build.gradle.kts
+++ b/stream-chat-android-client/build.gradle.kts
@@ -69,6 +69,7 @@ tasks.withType<KotlinCompile>().configureEach {
             listOf(
                 "-progressive",
                 "-Xexplicit-api=strict",
+                "-Xjvm-default=all",
                 "-opt-in=kotlin.RequiresOptIn",
                 "-opt-in=io.getstream.chat.android.core.internal.InternalStreamChatApi",
             ),
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -196,6 +196,7 @@ import io.getstream.chat.android.models.UploadAttachmentsNetworkType
 import io.getstream.chat.android.models.UploadedFile
 import io.getstream.chat.android.models.User
 import io.getstream.chat.android.models.UserBlock
+import io.getstream.chat.android.models.UserId
 import io.getstream.chat.android.models.VideoCallInfo
 import io.getstream.chat.android.models.VideoCallToken
 import io.getstream.chat.android.models.Vote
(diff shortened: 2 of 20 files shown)


stream-chat-android-client/build.gradle.kts
```diff
@@ -69,6 +69,7 @@ tasks.withType<KotlinCompile>().configureEach {
             listOf(
                 "-progressive",
                 "-Xexplicit-api=strict",
+                "-Xjvm-default=all",
```
JcMinarro, line 72: Why was this flag needed?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1030,6 +1031,86 @@ internal constructor(
...
+     * @return Executable async [Call] which completes with [Result] containing an instance of [UploadedFile]
+     * if the file was successfully uploaded.
+     *
+     * @see FileUploader
+     */
+    @CheckResult
+    @JvmOverloads
+    public fun uploadFile(
+        file: File,
+        user: User? = null,
```
JcMinarro: Only "the current user" can upload files.
`ChatClient` already contains the current user instance.
We shouldn't add this parameter here to simplify the API

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/FileUploader.kt
```diff
@@ -115,12 +117,81 @@ public interface FileUploader {
...
+     * @param progressCallback The callback to be invoked periodically to report upload progress.
+     * @return The [Result] object containing an instance of [UploadedFile] in the case of a successful upload
+     * or an exception if the upload failed.
+     *
+     * @see [Result.success]
+     * @see [Result.failure]
+     */
+    public fun uploadFile(
+        file: File,
+        user: User?,
```
JcMinarro: The same here, the user is not needed

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/StreamFileUploader.kt
```diff
@@ -38,106 +47,115 @@ internal class StreamFileUploader(
...
+        val filename = filenameSanitizer.sanitize(filename = name)
+        return MultipartBody.Part.createFormData(name = "file", filename = filename, body = body)
+    }
+
+    private fun User.asBodyPart(): MultipartBody.Part = with(dtoMapping) {
+        val adapter = Moshi.Builder().build().adapter(UpstreamUserDto::class.java)
+        val json = adapter.toJson(toDto())
+        val body = json.toRequestBody(contentType = "application/json".toMediaType())
+        MultipartBody.Part.createFormData(name = "user", filename = null, body = body)
     }
```
JcMinarro: Do we need to map the full user entity here? :thinking:

stream-chat-android-client/build.gradle.kts
andremion: The default interface implementation is not enabled for the JVM.
The `ProgressCallback` interface now has no-op default implementations, so that consumers are not forced to implement all of them

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: User is optional. Which means it is not necessary about the current user object.
Take a look at the JS implementation: https://github.com/GetStream/stream-chat-js/pull/1564/files#diff-25d66d74617fe2e23d7946bd6e3ba95640ab1b9bc8947445d604fc271c7c1f12R4599

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/StreamFileUploader.kt
andremion: Just the upstream DTO

https://getstream.github.io/protocol/#/product%3Achat/UploadFileGlobal

https://github.com/GetStream/stream-chat-js/blob/5a1bd6d2a1d8d832bbd83f93f7936d4fe074a529/src/types.ts#L887

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: Let me double-check with the BE team.

JcMinarro: Even if it is optional, it could be confusing on our public api...
Imagine I upload a picture with your user.... Doesn't make sense.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/StreamFileUploader.kt
JcMinarro: I think it is only a "easy to use" implementation, but I don't think backend need the full `UpstreamUserDto`
Screenshot from 2025-09-08 11-15-02

I assume they would only need the userId, and maybe they are using json notation, but something like the following would be enough. 
```
user: { id: 123 }
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: My 2 cents here: The JS SDK is used for both client and server side. When operations from it are used server-side, customers must provide user/userId, but that is not required when using it client side - because there will already be a connected user. 
So it is entirely possible that the `user` here is redundant - but let's maybe wait for confirmation from the BE team.

stream-chat-android-client/build.gradle.kts
andremion: It is in the PR description
Screenshot 2025-09-08 at 16 56 22

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -563,6 +563,58 @@ constructor(
...
+    ): Call<UploadedFile> = CoroutineCall(coroutineScope) {
+        fileTransformer.transform(file)
+            .let { transformedFile ->
+                fileUploader.uploadImage(
+                    file = transformedFile,
+                    userId = userId,
+                    progressCallback = progressCallback,
+                )
+            }.onSuccess { uploadedFile ->
+                progressCallback?.onSuccess(url = uploadedFile.file)
```
VelikovPetar, line 606: Do we need to manually invoke the `progressCallback`? I see that in the existing methods (ex. `sendImage`) we don't invoke the callback manually, but rather we just pass it to the retrofit Api?

andremion: The `ProgressInterceptor` is only responsible for notifying `onProgress`. 
`onSuccess` and `onError` are not called at all

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/uploader/StreamFileUploader.kt
andremion: The user is now entirely removed as it is not needed to be sent from the mobile clients

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: BE confirmed. Mobile clients don't need to send the user when uploading files/images. Thanks!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
JcMinarro: :thinking: 
Should we include this change on previous calls as well?

PR #5911, opened 2025-08-29
Improve accessibility of message list items

andremion:
### 🎯 Goal

Improve accessibility of the message list items

### 🛠 Implementation details

- Merge nodes of message date, unread label, and thread separators
- The `onUserAvatarClick` parameter in `MessagesScreen` and `MessageContainer` is now nullable, as well as the `onChannelAvatarClick` parameter and its internal `MessagesScreenContent` composable, allowing to have better semantics if the click is unhandled.
- Add content descriptions for message read status icons and improve the thread footer text handling.
- Add content descriptions for media attachments and message reaction items to enhance accessibility.
- Clear semantics of the other message author to prevent duplication with the avatar semantics.
- Improve content descriptions of the commands button

#### Out of scope changes

- Fix the order of some `assertEquals` arguments in `UserRobotMessageListAsserts`.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/5bc7b624-c3ff-45f4-be44-cd899e7f664f" controls="controls"  />
</td>
<td>
<video src="https://github.com/user-attachments/assets/8010fe40-a694-4a2c-a8d4-941e6d6b8b88" controls="controls" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Turn **Talkback** ON in **System settings** -> **Accessibility options**


stream-chat-android-ui-common/src/main/res/values/strings.xml
```diff
@@ -58,6 +58,17 @@
 
     <!-- Message List -->
     <string name="stream_ui_message_list_message_deleted">This message was deleted.</string>
+    <string name="stream_ui_message_list_message_status_read">Message read</string>
```
VelikovPetar: Small question/suggestion: Should we maybe distinguish the accessibility texts from the other ones by maybe adding something like `_tooltip` suffix? I don't really see it as necessary, just maybe something that would allow us to separate them more easily.

VelikovPetar (approved):
Looks pretty good to me!
We should try to follow the approach of defining lambdas as nullable when they are optional(like the changes you did, such as `onUserAvatarClick: (() -> Unit)? = null`), so that the talkback behaviour is better (for new features)

stream-chat-android-ui-common/src/main/res/values/strings.xml
andremion: I added `semantics` to the string names

VelikovPetar: Nice, thank you!

PR #5916, opened 2025-09-03
Add `MessageComposerState.sendEnabled` property to be able to locally enable/disable sned action

JcMinarro:
### 🎯 Goal
Add a new `MessageComposerState.sendEnabled` property that allow to disable the Message Compose Input depending on custom logic the customer could have locally on the app.

### 🧪 Testing
You only need to modify the `MessageComposerState.sendEnabled`.
The following snippet toggles every 5 seconds the `sendEnabled` property to reflect how it could be edited locally from the `ChatComponentFactory` implementation

<details>

<summary>Git Path</summary>

```
diff --git a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
index 71740664d3..8aa642af5e 100644
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
@@ -18,9 +18,12 @@ package io.getstream.chat.android.compose.sample.ui.component
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.RowScope
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.lazy.LazyItemScope
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import io.getstream.chat.android.client.extensions.isPinned
 import io.getstream.chat.android.client.utils.message.hasSharedLocation
@@ -29,17 +32,24 @@ import io.getstream.chat.android.compose.sample.feature.reminders.MessageReminde
 import io.getstream.chat.android.compose.state.channels.list.ItemState
 import io.getstream.chat.android.compose.state.mediagallerypreview.MediaGalleryPreviewResult
 import io.getstream.chat.android.compose.ui.channels.list.ChannelItem
+import io.getstream.chat.android.compose.ui.messages.composer.actions.AudioRecordingActions
 import io.getstream.chat.android.compose.ui.theme.ChatComponentFactory
 import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.models.Attachment
+import io.getstream.chat.android.models.Command
+import io.getstream.chat.android.models.LinkPreview
 import io.getstream.chat.android.models.Option
 import io.getstream.chat.android.models.Poll
 import io.getstream.chat.android.models.Vote
+import io.getstream.chat.android.ui.common.state.messages.composer.MessageComposerState
 import io.getstream.chat.android.ui.common.state.messages.list.GiphyAction
 import io.getstream.chat.android.ui.common.state.messages.list.MessageItemState
 import io.getstream.chat.android.ui.common.state.messages.poll.PollSelectionType
+import kotlinx.coroutines.flow.flow
 
 class CustomChatComponentFactory(
     private val delegate: ChatComponentFactory = MessageRemindersComponentFactory(),
@@ -70,6 +80,70 @@ class CustomChatComponentFactory(
         )
     }
 
+    @Composable
+    override fun MessageComposer(
+        messageComposerState: MessageComposerState,
+        onSendMessage: (String, List<Attachment>) -> Unit,
+        modifier: Modifier,
+        onAttachmentsClick: () -> Unit,
+        onCommandsClick: () -> Unit,
+        onValueChange: (String) -> Unit,
+        onAttachmentRemoved: (Attachment) -> Unit,
+        onCancelAction: () -> Unit,
+        onLinkPreviewClick: ((LinkPreview) -> Unit)?,
+        onMentionSelected: (User) -> Unit,
+        onCommandSelected: (Command) -> Unit,
+        onAlsoSendToChannelSelected: (Boolean) -> Unit,
+        recordingActions: AudioRecordingActions,
+        headerContent: @Composable (ColumnScope.(MessageComposerState) -> Unit),
+        footerContent: @Composable (ColumnScope.(MessageComposerState) -> Unit),
+        mentionPopupContent: @Composable ((List<User>) -> Unit),
+        commandPopupContent: @Composable ((List<Command>) -> Unit),
+        integrations: @Composable (RowScope.(MessageComposerState) -> Unit),
+        label: @Composable ((MessageComposerState) -> Unit),
+        input: @Composable (RowScope.(MessageComposerState) -> Unit),
+        audioRecordingContent: @Composable (RowScope.(MessageComposerState) -> Unit),
+        trailingContent: @Composable ((MessageComposerState) -> Unit),
+    ) {
+        val sendMessageState = remember {
+            flow {
+                var enabled = false
+                while (true) {
+                    enabled = !enabled
+                    emit(enabled)
+                    delay(5000)
+                }
+            }
+        }.collectAsState(initial = true)
+        super.MessageComposer(
+            messageComposerState
+                .copy(
+                    sendEnabled = sendMessageState.value
+                ),
+            onSendMessage,
+            modifier,
+            onAttachmentsClick,
+            onCommandsClick,
+            onValueChange,
+            onAttachmentRemoved,
+            onCancelAction,
+            onLinkPreviewClick,
+            onMentionSelected,
+            onCommandSelected,
+            onAlsoSendToChannelSelected,
+            recordingActions,
+            headerContent,
+            footerContent,
+            mentionPopupContent,
+            commandPopupContent,
+            integrations,
+            label,
+            input,
+            audioRecordingContent,
+            trailingContent
+        )
+    }
+
     @Composable
     override fun ColumnScope.MessageItemCenterContent(
         messageItem: MessageItemState,

```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ## stream-chat-android-client
 ### 🐞 Fixed
+- Fix API_KEY being sent with the request done to the CDN when sharing a video from the media gallery. [#5921](https://github.com/GetStream/stream-chat-android/pull/5921)
 
 ### ⬆️ Improved
 
@@ -62,6 +63,8 @@
 - Ensure any callback action called from the CoilStreamImageLoader is run on the caller thread. [#5915](https://github.com/GetStream/stream-chat-android/pull/5915)
 
 ### ⬆️ Improved
+- Avoid marking a pending message as read. [#5926](https://github.com/GetStream/stream-chat-android/pull/5926)
+- Improve the read count calculation on every message based on `channelUserRead.lastReadMessageId`. [#5923](https://github.com/GetStream/stream-chat-android/pull/5923)
 
 ### ✅ Added
 
@@ -89,6 +92,7 @@
 ### ⬆️ Improved
 - Improved accessibility of the message list items. [#5911](https://github.com/GetStream/stream-chat-android/pull/5911)
 - Add MessageComposerState.sendEnabled property to be able to locally enable/disable sned action. [#5916](https://github.com/GetStream/stream-chat-android/pull/5916)
+- Show quoted message over messages with attachments. [#5924](https://github.com/GetStream/stream-chat-android/pull/5924)
 
 ### ✅ Added
 - Added `MessageReactionList` and `MessageReactionItem` component factories. [#5925](https://github.com/GetStream/stream-chat-android/pull/5925)
--- a/metrics/size.json
+++ b/metrics/size.json
@@ -1,6 +1,6 @@
 {
   "debug": {
-    "stream-chat-android-client": 4376,
+    "stream-chat-android-client": 4372,
     "stream-chat-android-offline": 4708,
     "stream-chat-android-ui-components": 13252,
     "stream-chat-android-compose": 16448
--- a/stream-chat-android-client/api/stream-chat-android-client.api
+++ b/stream-chat-android-client/api/stream-chat-android-client.api
@@ -264,6 +264,7 @@ public final class io/getstream/chat/android/client/ChatClient$Builder : io/gets
 	public final fun okHttpClient (Lokhttp3/OkHttpClient;)Lio/getstream/chat/android/client/ChatClient$Builder;
 	public final fun retryPolicy (Lio/getstream/result/call/retry/RetryPolicy;)Lio/getstream/chat/android/client/ChatClient$Builder;
 	public final fun sendMessageInterceptor (Lio/getstream/chat/android/client/interceptor/SendMessageInterceptor;)Lio/getstream/chat/android/client/ChatClient$Builder;
+	public final fun shareFileDownloadRequestInterceptor (Lokhttp3/Interceptor;)Lio/getstream/chat/android/client/ChatClient$Builder;
 	public final fun uploadAttachmentsNetworkType (Lio/getstream/chat/android/models/UploadAttachmentsNetworkType;)Lio/getstream/chat/android/client/ChatClient$Builder;
 	public final fun withApiModelTransformer (Lio/getstream/chat/android/client/transformer/ApiModelTransformers;)Lio/getstream/chat/android/client/ChatClient$Builder;
 	public final fun withPlugins ([Lio/getstream/chat/android/client/plugin/factory/PluginFactory;)Lio/getstream/chat/android/client/ChatClient$Builder;
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -232,6 +232,7 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withTimeout
 import kotlinx.coroutines.withTimeoutOrNull
+import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import okhttp3.ResponseBody
 import java.io.File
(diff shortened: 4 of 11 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/composer/MessageComposerState.kt
```diff
@@ -60,4 +62,5 @@ public data class MessageComposerState @JvmOverloads constructor(
     val currentUser: User? = null,
     val recording: RecordingState = RecordingState.Idle,
     val pollsEnabled: Boolean = false,
+    val sendEnabled: Boolean = true,
```
VelikovPetar, line 65: Small concern here: if an integrator is using the `MessageComposer` composable which uses the `MessageComposerViewModel`, wouldn't it be difficult for them to override this flag?

JcMinarro: If they are using `MessageComposer`, they can add their own `input` lambda or modify the `ChatComopnentFactory`, which shouldn't be difficult for them.
But if you see another way to do it without adding a breaking change, we can do it.

VelikovPetar: TBH I cannot really think of a silver bullet to cover all scenarios, I think this customisation will be difficult for one type of integration, but some other types of customisation could make others more difficult. 
I am thinking if adding a flag `enabled` on the `MessageComposer` composable would simplify the integration, but then again, if a customer is using the `MesagesScreen` component, they would still face the same issue 🤔

andremion: What about adding a lambda to the composer controller/factory so that the integrator could override it, if needed?

Something like:

```kotllin
public class MessageComposerController(
    isSendEnabled: (currentState: MessageComposerState) -> Boolean = { it.ownCapabilities.contains(ChannelCapabilities.SEND_MESSAGE) }
)
```

So that the integrator could do something like this:

```kotlin
private val factory by lazy {
    MessagesViewModelFactory(
        isSendEnabled = { custom logic which takes the current state as input and returns a boolean  }
    )
)
```

Does that make sense?

JcMinarro: How will they update it dynamically? :thinking: 
Do they need to wait until any "Stream event" happens that updates the Composable state?

andremion: Good point 😅

andremion: So your implementation should cover it.

I would just avoid having this in the composables:

```kotlin
val canSendMessage = messageComposerState.sendEnabled &&
        messageComposerState.ownCapabilities.contains(ChannelCapabilities.SEND_MESSAGE)
```

How about using ownCapabilities to define the sendEnabled in the view controller instead?
So that logic would be inside the controller and not in the UI component.

JcMinarro: We should change `ownCapabilities` client-side, as it is an option configurable on backend. So our components should follow it, but the customers wanted a way to "locally disable" it as well, because their app has some busness rules that they need to follow, and those rules need to be applied locally :(

andremion: What I mean is to let the composables just read `messageComposerState.sendEnabled` and assign it in the view controller with the ownCapabilities.
In case the integrator wants, they can override it in the messagecomposer as you already did.

PR #5918, opened 2025-09-05
Expose Channel Messages Count

JcMinarro:
### 🎯 Goal
Expose Channel Messages Count.
This property has been added to multiple entities:
- `NewMessageEvent.channelMessageCount`
-  MessageDeletedEvent.channelMessageCount`
- `ChannelState.messagesCount`
- `Channel.messagesCount`

Complete: AND-721

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
@@ -174,7 +174,7 @@ internal class DomainMapping(
             ownCapabilities = own_capabilities.toSet(),
             membership = membership?.toDomain(),
             activeLiveLocations = active_live_locations.map { it.toDomain() },
-            messagesCount = message_count,
+            messageCount = message_count,
             extraData = extraData.toMutableMap(),
         ).syncUnreadCountWithReads(currentUserIdProvider())
             .let(channelTransformer::transform)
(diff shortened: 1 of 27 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -171,6 +171,7 @@ internal class DomainMapping(
             ownCapabilities = own_capabilities.toSet(),
             membership = membership?.toDomain(),
             activeLiveLocations = active_live_locations.map { it.toDomain() },
+            messagesCount = message_count,
```
VelikovPetar: Small thing: Should we rename the `Channel` property to `messageCount` (from `messagesCount`)? It would be the same as the BE model, and aligned with the name in the events.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/ChannelDtos.kt
```diff
@@ -77,5 +77,6 @@ internal data class DownstreamChannelDto(
     val own_capabilities: List<String> = emptyList(),
     val membership: DownstreamMemberDto?,
     val active_live_locations: List<DownstreamLocationDto> = emptyList(),
+    val message_count: Int? = null,
```
VelikovPetar, line 80: Would it make sense to also persist this info in the DB? (in `ChannelEntity`)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
JcMinarro: DONE

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/ChannelDtos.kt
JcMinarro: I didn't add it because it is normally a data that is constantly updated. But I just added it and upgrade de DB Version

PR #5920, opened 2025-09-07
(WIP) Skip Stream-specific interceptors for `FileDownloadApi`.

VelikovPetar:
### 🎯 Goal
The `FileDownloadApi` is used only internally for downloading a video from an URL to a `File` (called from `StreamFileUtil`). The video URL is always read as `attachment.assetUrl`. This method is only used when the user attempts to `Share` the file from the `MediaGallery`.
The `FileDownloadApi` is currently setup as all other Stream APIs - with interceptors setting the API_KEY, headers and Auth logic. This can cause problems, if the `attachment.assetUrl` is a custom, non-Stream CDN URL -> meaning that the application of the Stream API_KEY / headers, could result in failures to download the desired file from the URL.
This PR introduces a proposal for resolving this issue, by removing the interceptors which provide the API_KEY, the Stream headers and the auth logic from the `FileDownloadApi`.

### 🛠 Implementation details
`BaseChatModule`:
- setup the base `OkHttpClient.Builder` with timeout, api analysis and logging logic
- use the base client for the `FileDownloadApi`
- use the extended base client with Stream-specific interceptors for all other `-Api`s

### 🎨 UI Changes

_Add relevant screenshots_

| Before | After |
| --- | --- |
| img | img |

_Add relevant videos_

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
_This is a bit difficult to verify that it works for a custom CDN, so the testing steps show how to verify that the current behaviour using the default CDN is not broken. However, any custom CDN should not depend on the presence of the Stream-specific interceptors._
1. Open a channel with Video attachment
2. Open the Video attachment (in `MediaGallery`)
3. Tap on the `Share` button
4. The share should be successful (the download done in the background should be successful)

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
@@ -86,6 +86,7 @@ import io.getstream.chat.android.models.UserId
 import io.getstream.log.StreamLog
 import okhttp3.OkHttpClient
 import retrofit2.Retrofit
+import retrofit2.create
 import java.util.concurrent.TimeUnit
 
 @Suppress("TooManyFunctions")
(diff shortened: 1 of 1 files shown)


VelikovPetar:
Full implementation done in: https://github.com/GetStream/stream-chat-android/pull/5921

PR #5921, opened 2025-09-09
Skip Stream-specific interceptors for `FileDownloadApi`. 

VelikovPetar:
### 🎯 Goal
The `FileDownloadApi` is used only internally for downloading a video from an URL to a `File` (called from `StreamFileUtil`). The video URL is always read as `attachment.assetUrl`. This method is only used when the user attempts to `Share` the file from the `MediaGallery`.
The `FileDownloadApi` is currently setup as all other Stream APIs - with interceptors setting the API_KEY, headers and Auth logic. This can cause problems, if the `attachment.assetUrl` is a custom, non-Stream CDN URL -> meaning that the application of the Stream API_KEY / headers, could result in failures to download the desired file from the URL.
This PR introduces a proposal for resolving this issue, by removing the interceptors which provide the API_KEY, the Stream headers and the auth logic from the `FileDownloadApi`.

Additionally this PR introduces a `ShareFileDownloadRequestInterceptor` which can be overridden to customise the request used to download the file before sharing it.

### 🛠 Implementation details
`BaseChatModule`:
- setup the base `OkHttpClient.Builder` with timeout, api analysis and logging logic
- use the base client for the `FileDownloadApi`
- use the extended base client with Stream-specific interceptors for all other `-Api`s
`ChatClient`:
- add `ChatClient.Builder.shareFileDownloadRequestInterceptor` to provide custom `ShareFileDownloadRequestInterceptor` to modify the request of the download operation before the share.

### 🧪 Testing
_This is a bit difficult to verify that it works for a custom CDN, so the testing steps show how to verify that the current behaviour using the default CDN is not broken. However, any custom CDN should not depend on the presence of the Stream-specific interceptors._
1. Open a channel with Video attachment
2. Open the Video attachment (in `MediaGallery`)
3. Tap on the `Share` button
4. The share should be successful (the download done in the background should be successful)

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
@@ -142,7 +142,9 @@ constructor(
     private val defaultApi by lazy { buildApi(config) }
     internal val chatSocket: ChatSocket by lazy { buildChatSocket(config) }
     private val defaultFileUploader by lazy {
-        StreamFileUploader(buildRetrofitCdnApi())
+        StreamFileUploader(
+            retrofitCdnApi = buildRetrofitCdnApi(),
+        )
     }
 
     val lifecycleObserver: StreamLifecycleObserver by lazy { StreamLifecycleObserver(userScope, lifecycle) }
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
@@ -59,6 +59,7 @@ import io.getstream.chat.android.client.api2.mapping.EventMapping
 import io.getstream.chat.android.client.clientstate.UserStateService
 import io.getstream.chat.android.client.debugger.ChatClientDebugger
 import io.getstream.chat.android.client.interceptor.SendMessageInterceptor
+import io.getstream.chat.android.client.interceptor.ShareFileDownloadRequestInterceptor
 import io.getstream.chat.android.client.logger.ChatLogLevel
 import io.getstream.chat.android.client.network.NetworkStateProvider
 import io.getstream.chat.android.client.notifications.ChatNotifications
@@ -86,6 +87,7 @@ import io.getstream.chat.android.models.UserId
 import io.getstream.log.StreamLog
 import okhttp3.OkHttpClient
 import retrofit2.Retrofit
+import retrofit2.create
 import java.util.concurrent.TimeUnit
 
 @Suppress("TooManyFunctions")
@@ -101,13 +103,13 @@ constructor(
     private val fileTransformer: FileTransformer,
     private val fileUploader: FileUploader? = null,
     private val sendMessageInterceptor: SendMessageInterceptor? = null,
+    private val shareFileDownloadRequestInterceptor: ShareFileDownloadRequestInterceptor? = null,
     private val tokenManager: TokenManager = TokenManagerImpl(),
     private val customOkHttpClient: OkHttpClient? = null,
     private val clientDebugger: ChatClientDebugger? = null,
     private val lifecycle: Lifecycle,
     private val appName: String?,
     private val appVersion: String?,
-    private val httpClientConfig: (OkHttpClient.Builder) -> OkHttpClient.Builder = { it },
 ) {
 
     private val headersUtil = HeadersUtil(appContext, appName, appVersion)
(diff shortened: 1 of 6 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -101,13 +103,13 @@ constructor(
     private val fileTransformer: FileTransformer,
     private val fileUploader: FileUploader? = null,
     private val sendMessageInterceptor: SendMessageInterceptor? = null,
+    private val shareFileDownloadRequestInterceptor: ShareFileDownloadRequestInterceptor? = null,
```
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5921/commits/cbe1380dd00ca5b1e1a78553019b903357fa11e2

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -316,6 +346,49 @@ constructor(
...
+    private fun buildFileDownloadApi(): FileDownloadApi {
+        val okHttpClient = baseClientBuilder(BASE_TIMEOUT)
+            .addInterceptor { chain ->
+                // Apply custom modifications to the request
+                val requestBuilder = chain.request().newBuilder()
+                val modifiedRequestBuilder =
+                    shareFileDownloadRequestInterceptor?.intercept(requestBuilder) ?: requestBuilder
+                val newRequest = modifiedRequestBuilder.build()
+                chain.proceed(newRequest)
+            }
```
JcMinarro: Why do we create an inline `interceptor` and within this one we use the `sharedFileDownloadRequestInterceptor`?? :thinking: 
Why don't we add the `sharedFileDownloadRequestInterceptor`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
```diff
@@ -101,13 +129,13 @@ constructor(
     private val fileTransformer: FileTransformer,
     private val fileUploader: FileUploader? = null,
     private val sendMessageInterceptor: SendMessageInterceptor? = null,
+    private val shareFileDownloadRequestInterceptor: ShareFileDownloadRequestInterceptor? = null,
```
JcMinarro: This class is only used by us, and only in one part of our code; we should avoid default params here.
If not, we can keep params like `httpClientConfig` that wasn't used at all

VelikovPetar: I will update it!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/di/BaseChatModule.kt
VelikovPetar: The `ShareFileDownloadRequestInterceptor` doesn't extend from `okhttp3.Interceptor` so we cannot directly attach it using `addInterceptor(shareFileDownloadRequestInterceptor)`. The `ShareFileDownloadRequestInterceptor ` is just an `interface`, and its `intercept` method provides only the `Request.Builder`, not the whole `Interceptor.Chain`. Do you think we should rework it so that it also implements `okhttp3.Interceptor`?

JcMinarro: The interface is already coupled to `okhttp3.Request`, I think, requiring an `okhttp3.Interceptor` should be fine and the integration is easier

VelikovPetar: What's your opinion on exposing an `okhttp3.Interceptor` directly?
```kotlin
// ChatClient.Builder
public fun shareFileDownloadRequestInterceptor(
       interceptor: okhttp3.Interceptor,
): Builder {
       this.shareFileDownloadRequestInterceptor = shareFileDownloadRequestInterceptor
       return this
}
```

Or do you think we should 'wrap' the `okhttp3.Interceptor` in our own specific interface? 
```kotlin
public interface ShareFileDownloadRequestInterceptor: okhttp3.Interceptor

// ChatClient.Builder
public fun shareFileDownloadRequestInterceptor(
       interceptor: ShareFileDownloadRequestInterceptor,
): Builder {
       this.shareFileDownloadRequestInterceptor = interceptor
       return this
}
```

JcMinarro: I think we can expose it directly, we don't really need to wrap it

PR #5925, opened 2025-09-10
Add `MessageReactionList` and `MessageReactionItem` component factories

andremion:
### 🎯 Goal

Expose message reactions in component factories, so that customers can customize them as needed.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1178,6 +1186,47 @@ public interface ChatComponentFactory {
...
 
+    /**
+     * The default list of reactions displayed above the message bubble and is part of [MessageItemHeaderContent].
+     */
+    @Composable
+    public fun MessageReactionList(
+        modifier: Modifier,
+        message: Message,
+        reactions: List<ReactionOptionItemState>,
+        onClick: (message: Message) -> Unit,
```
JcMinarro: Can we change it by a "holder object"?
It will help us to include extra argument in the future if it is needed

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1178,6 +1186,47 @@ public interface ChatComponentFactory {
...
+        )
+    }
+
+    /**
+     * The default individual reaction item shown inside [MessageReactionList].
+     */
+    @Composable
+    public fun RowScope.MessageReactionItem(
+        modifier: Modifier,
+        reaction: ReactionOptionItemState,
```
JcMinarro: Same here

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: Good point!
Done ✅

JcMinarro (approved):

PR #5926, opened 2025-09-11
Avoid calling MarkMessageUnread on PendingMessage

JcMinarro:
### 🎯 Goal
To ensure that the sent messages using the UI Components are marked as read after they are delivered to the backend, we added a call to mark this specific message as read on #5322
Now that we have the Pending Message feature implemented, following the same approach is problematic because the Backend doesn't allow us to mark as read a message that has not been committed yet, returning an error in that case.
To avoid this wrong API call that we know will fail, we can add a precondition to check if the "Pending Messages" feature is enabled and avoid this call to the server.

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
@@ -720,7 +720,7 @@ public class MessageComposerController(
                 message.copy(showInChannel = isInThread && alsoSendToChannel.value),
             ).doOnResult(scope) { result ->
                 result.onSuccessSuspend { resultMessage ->
-                    if (channelState.value?.channelConfig?.value?.markMessagesPending != false) {
+                    if (channelState.value?.channelConfig?.value?.markMessagesPending == false) {
                         chatClient.markMessageRead(
                             channelType = channelType,
                             channelId = channelId,


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -724,7 +741,9 @@ public class MessageComposerController(
                         channelType = channelType,
                         channelId = channelId,
                         messageId = resultMessage.id,
-                    ).await()
+                    )
+                        .withPrecondition(scope) { pendingMessagePrecondition.value }
```
VelikovPetar: Do we need to have the `pendingMessagePrecondition` defined as a `StateFlow`? 
Can't we just directly access its value in the `withPrecondition` block? Ex. 
```kotlin
.withPrecondition(scope) {
      when (channelState.value?.channelConfig?.value?.markMessagesPending) {
               true -> Result.Failure(Error.GenericError("Pending messages are disabled in this channel"))
               else -> Result.Success(Unit)
       }
}
```
Or maybe even a simpler `if-else` check: 
```kotlin
if (!mark_messages_pending) {
    chatClient.markMessageRead(/* ... */)
}
```

JcMinarro: Ummm, yes, at the end, the result is not checked either needed

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -720,11 +720,14 @@ public class MessageComposerController(
                 message.copy(showInChannel = isInThread && alsoSendToChannel.value),
             ).doOnResult(scope) { result ->
                 result.onSuccessSuspend { resultMessage ->
-                    chatClient.markMessageRead(
-                        channelType = channelType,
-                        channelId = channelId,
-                        messageId = resultMessage.id,
-                    ).await()
+                    if (channelState.value?.channelConfig?.value?.markMessagesPending != false) {
```
VelikovPetar: Shouldn't this check be the other way around? 
```kotlin
if (channelState.value?.channelConfig?.value?.markMessagesPending == false) {
    // mark message as read only when `mark_messages_pending` is false
}
```

JcMinarro: Ups, you are right

JcMinarro:
```suggestion
                    if (channelState.value?.channelConfig?.value?.markMessagesPending == false) {
```

PR #5927, opened 2025-09-11
Introduce share action to the ChannelMediaAttachmentsPreviewScreen

andremion:
### 🎯 Goal

Introduce share action to the `ChannelMediaAttachmentsPreviewScreen`

### 🛠 Implementation details

There is no need to distinguish between sharing an image and a video.
Both can be handled the same way, including showing a warning dialog when the file is bigger than 10MB.
Before sharing a file, the app needs to download it to the cache directory, which is managed by the OS and automatically pruned when more storage is needed. With the file in cache, the app can now trigger the native sharing mechanism, where the user can choose the destination for it.
Some logic is shared between the compose gallery preview and the channel attachments preview.
Previously, the gallery preview had different logic for images and video downloading, which is not necessary.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Sharing uncached video</th>
<th>Sharing cached video</th>
<th>Sharing image</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/e1018200-f647-4bb7-8ea3-ff40d2333797" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/b604f6a3-8e0a-4794-8059-02d6bbc7dd03" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/42540280-9b93-41f4-8865-a7a1eb143a49" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### DIFF
--- a/stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivityTest.kt
+++ b/stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivityTest.kt
@@ -37,6 +37,7 @@ import io.getstream.chat.android.ui.common.images.resizing.StreamCdnImageResizin
 import io.getstream.result.Result
 import kotlinx.coroutines.flow.MutableStateFlow
 import org.junit.Before
+import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 import org.junit.jupiter.api.Assertions.assertEquals
@@ -141,6 +142,7 @@ internal class MediaGalleryPreviewActivityTest : ComposeTest {
         }
     }
 
+    @Ignore("https://linear.app/stream/issue/AND-741")
     @Test
     fun `should show large file warning dialog when sharing a large file`() {
         val message = PreviewMessageData.messageWithUserAndAttachment.run {


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -3573,25 +3573,23 @@ public interface ChatComponentFactory {
...
-        ) {
-            Text(
-                text = text,
-                style = ChatTheme.typography.title3Bold,
-                color = ChatTheme.colors.textHighEmphasis,
-                maxLines = 1,
-            )
-        }
+    public fun ChannelMediaAttachmentsPreviewBottomBar(
+        params: ChannelMediaAttachmentsPreviewBottomBarParams,
```
VelikovPetar, line 3612: Wouldn't this be a breaking change?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
```diff
@@ -45,3 +46,16 @@ public data class MessageReactionItemParams(
...
+
+/**
+ * Parameters for the [ChatComponentFactory.ChannelMediaAttachmentsPreviewBottomBar] component.
+ *
+ * @param centerContent Composable lambda for center content in the bottom bar.
+ * @param leadingContent Composable lambda for leading content in the bottom bar.
+ * @param trailingContent Composable lambda for trailing content in the bottom bar.
+ */
+public data class ChannelMediaAttachmentsPreviewBottomBarParams(
+    val centerContent: @Composable () -> Unit,
```
VelikovPetar, line 58: A bit strange setup in my opinion, to pass composables as part of a `-Params` class. Should they maybe be separated entries in the `ComponentFactory` instead?

stream-chat-android-ui-common/src/main/res/values/strings.xml
```diff
@@ -169,7 +169,6 @@
     <string name="stream_ui_channel_attachments_media_empty_title">No media</string>
     <string name="stream_ui_channel_attachments_media_empty_text">Media sent in this chat will appear here.</string>
     <string name="stream_ui_channel_attachments_media_load_error">Failed to load media attachments!</string>
-    <string name="stream_ui_channel_attachments_media_preview_index">%d of %d</string>
```
VelikovPetar, line 172: Is it really OK to just remove this, wouldn't it be a breaking change?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
```diff
@@ -330,36 +337,19 @@ public class MediaGalleryPreviewActivity : AppCompatActivity() {
...
 
-    /**
-     * Fetches an image from Coil's cache and shares it.
-     *
-     * @param attachment The attachment used to prepare the URI.
-     */
-    private suspend fun shareImage(attachment: Attachment) {
-        val attachmentUrl = attachment.imagePreviewUrl
-        if (attachmentUrl != null) {
-            val bitmap = StreamImageLoader.instance().loadAsBitmap(applicationContext, attachmentUrl)
```
VelikovPetar, line 349: One question about this change: Because currently we load the image via the `StreamImageLoader`, it also applies any transformations made in a custom `ImageAssetTransformer`/`ImageHeadersProvider`. By migrating to the `StreamFileUtils` method, we would lose this logic for custom headers/transformation - this could potentially break the integration if the customer relies on these headers/transformations.

VelikovPetar (commented):
Looks like a good addition, I just left several comments about some concerns. 
I am also curious: Was this a parity issue, where we are missing this feature from the SDK (compared to the other SDKs)?

andremion:
> I am also curious: Was this a parity issue, where we are missing this feature from the SDK (compared to the other SDKs)?
Yes. We already have the sharing option in the SwiftUI SDK and Android XML.

https://github.com/user-attachments/assets/d6fbdd7d-7211-4d39-bcb9-c24a0421d3fa

[Screen_recording_20250916_092310.webm](https://github.com/user-attachments/assets/264f6c00-1018-4a4a-920d-d6b7a02f6beb)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: Yes.
However, the impact should be very low because this feature was released 3 weeks ago (6.22.0) and has not been used much to have customers do customizations on that UI.
Do you have any idea to support that action without breaking changes?

stream-chat-android-ui-common/src/main/res/values/strings.xml
andremion: Same as above

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
andremion: The idea was to provide more flexibility when customizing components, as that component has a small scope

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
andremion: I see. I will push a fix to that. 👍🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
VelikovPetar: Could we somehow 'compose' the old bottom bar inside the new one? I can see that the old factory method now kinda corresponds to the `MediaGalleryPreviewPageIndicator` -> can we somehow re-wire the `ChatTheme.componentFactory.ChannelMediaAttachmentsPreviewBottomBar` invocation to call the old bottom bar method? Something like: 

```kotlin
ChatTheme.componentFactory.ChannelMediaAttachmentsPreviewBottomBar(
        params = ChannelMediaAttachmentsPreviewBottomBarParams(
            centerContent = {
                if (viewState.isPreparingToShare) {
                    MediaGalleryPreviewSharingInProgressIndicator()
                } else {
                    // Call old bottom bar method here
                }
            },
            leadingContent = {
                if (config.isShareVisible) {
                    MediaGalleryPreviewShareIcon(
                        connectionState = connectionState,
                        isSharingInProgress = viewState.isPreparingToShare,
                        onClick = { onShareClick(items[pagerState.currentPage].attachment) },
                    )
                }
            },
        ),
    )
```

Even though it is a new feature, it is still a risk to publish such change, as there are customers already using the component, and we cannot guarantee that this will not break anything...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
andremion: I changed to use loadAsBitmap for images.
I also changed to write the temporary image in the cacheDir instead of in the public picture directories, so that temp file can be pruned when needed

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: The problem is that now we can't reuse `MediaGalleryPreviewPageIndicator`
But it's the price to pay. We can reuse it when we remove the deprecated one.

andremion: Done

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channel/ChannelMediaAttachmentsPreviewViewModel.kt
```diff
@@ -192,6 +193,18 @@ internal class AttachmentFileController(
     suspend fun getFileFromCache(attachment: Attachment): Result<Uri> =
         StreamFileUtil.getFileFromCache(context, attachment)
 
+    suspend fun downloadImage(attachment: Attachment): Result<Uri> =
```
VelikovPetar: This seems unused (yet), probably not intentional?

VelikovPetar:
I think that we are doing something heavy on the UI thread when sharing an image from the `MediaGalleryPreviewActivity`:

https://github.com/user-attachments/assets/7fe4d63b-7bcc-4a42-946e-b9c9639f30b9

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/attachments/ChannelMediaAttachmentsPreviewScreen.kt
```diff
@@ -0,0 +1,263 @@
...
+private fun ChannelMediaAttachmentsPreviewBottomBar(
+    items: List<ChannelAttachmentsViewState.Content.Item>,
+    viewState: ChannelMediaAttachmentsPreviewViewState,
+    connectionState: ConnectionState,
+    pagerState: PagerState,
+    onShareClick: (Attachment) -> Unit,
+) {
+    val config = ChatTheme.mediaGalleryConfig
+
+    ChatTheme.componentFactory.ChannelMediaAttachmentsPreviewBottomBar(
```
VelikovPetar, line 192: Just one small thing I noticed on my final review - the bottom bar is a bit misaligned (because of the leading icon). Do you think we should align the text centrally? 
Screenshot 2025-09-19 at 09 35 14

andremion: Definitely, yes.
I will fix that later.
Thank you 🙌🏻

VelikovPetar: I think we also need to update the Snapshot test.

stream-chat-android-ui-common/src/main/res/values/strings.xml
VelikovPetar: Don't we need this string now, when the `ChannelMediaAttachmentsPreviewBottomBar` change is reverted?

scripts/sonar.gradle
```diff
@@ -10,6 +10,7 @@ ext.sonar = [
             'stream-chat-android-e2e-test',
             'stream-chat-android-test',
             'stream-chat-android-ui-guides',
+            'stream-chat-android-ui-components',
```
VelikovPetar, line 13: Do we really want to remove this? Maybe it's good to keep it to make sure it is not forgotten?

scripts/ci-unit-tests.sh
```diff
@@ -1,6 +1,7 @@
 command_file="build/tmp/unit-tests-command.sh"
 
-./gradlew test-debug && \
-chmod +x $command_file && \
-./$command_file && \
-rm $command_file
\ No newline at end of file
+#./gradlew test-debug && \
```
VelikovPetar, line 3: Can we delete this instead of commenting it out?

andremion: I will rather add a TODO to entirely remove UnitTestsPlugin as part of another task.
WDYT?

VelikovPetar: Fine with me!

scripts/sonar.gradle
andremion: My next task is to see how to increase the coverage of XML.
If it can't, I would rather do what iOS did with the UIKit: ignore it.
https://github.com/GetStream/stream-chat-swift/blob/develop/sonar-project.properties#L9

stream-chat-android-ui-common/src/main/res/values/strings.xml
andremion: `stream_ui_channel_attachments_media_preview_index` is not used anywhere.
Screenshot 2025-09-24 at 10 50 33

VelikovPetar: But then again, wouldn't this be a breaking change? If we changed the previously used resource with a new one?

andremion: I reverted it

PR #5931, opened 2025-09-15
Add filtering and sorting options for `QueryThreads`.

VelikovPetar:
### 🎯 Goal
Add support for filtering and sorting the results of the `QueryThreads` operation.
Resolves: https://linear.app/stream/issue/AND-633/query-threads-improvements

#### LLC
1. Add `filter` and `sort` properties to the `QueryThreadsRequest`
2. Add `extraData` properties on `Thread` and `ThreadInfo`
#### State/Offline
1. Add state/logic handling per `QueryThreadsRequest`
#### UI Common
1. Change `ThreadListController` to accept `QueryThreadsRequest` in the constructor instead of separate arguments.
#### Compose / XML
1. Add `ThreadsViewModelFactory(QueryThreadsRequest)` constructor
2. Deprecate `ThreadsViewModelFactory(threadLimit, threadReplyLimit, threadParticipantLimit)` constructor in favor of `ThreadsViewModelFactory(QueryThreadsRequest)` constructor.

### 🎨 UI Changes
No visible UI changes, the changes are additions to the LLC API, and in the instantiation of the `ThreadsViewModelFactory`

### 🧪 Testing
1. Update the `QueryThreadsRequest` passed in the existing `ThreadsViewModelFactory` (in `ChannelsActivity` in the compose sample or `ThreadsFragment` in the XML sample)
2. Observe that the thread list respects the passed filter/sort specs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Add support for `user.messages.deleted` event. [#5895](https://github.com/GetStream/stream-chat-android/pull/5895)
 
 ### ⚠️ Changed
 
--- a/README.md
+++ b/README.md
@@ -19,9 +19,9 @@
 <div align="center">
 
 ![stream-chat-android-client](https://img.shields.io/badge/stream--chat--android--client-3.21%20MB-lightgreen)
-![stream-chat-android-offline](https://img.shields.io/badge/stream--chat--android--offline-3.43%20MB-lightgreen)
+![stream-chat-android-offline](https://img.shields.io/badge/stream--chat--android--offline-3.44%20MB-lightgreen)
 ![stream-chat-android-ui-components](https://img.shields.io/badge/stream--chat--android--ui--components-10.51%20MB-lightgreen)
-![stream-chat-android-compose](https://img.shields.io/badge/stream--chat--android--compose-12.72%20MB-lightgreen)
+![stream-chat-android-compose](https://img.shields.io/badge/stream--chat--android--compose-12.73%20MB-lightgreen)
 
 </div>
 
--- a/metrics/size.json
+++ b/metrics/size.json
@@ -1,14 +1,14 @@
 {
   "debug": {
-    "stream-chat-android-client": 4376,
-    "stream-chat-android-offline": 4708,
-    "stream-chat-android-ui-components": 13256,
-    "stream-chat-android-compose": 16480
+    "stream-chat-android-client": 4380,
+    "stream-chat-android-offline": 4716,
+    "stream-chat-android-ui-components": 13260,
+    "stream-chat-android-compose": 16488
   },
   "release": {
-    "stream-chat-android-client": 3292,
-    "stream-chat-android-offline": 3520,
-    "stream-chat-android-ui-components": 10768,
-    "stream-chat-android-compose": 13032
+    "stream-chat-android-client": 3296,
+    "stream-chat-android-offline": 3528,
+    "stream-chat-android-ui-components": 10772,
+    "stream-chat-android-compose": 13036
   }
 }
--- a/stream-chat-android-client-test/src/main/java/io/getstream/chat/android/client/test/Mother.kt
+++ b/stream-chat-android-client-test/src/main/java/io/getstream/chat/android/client/test/Mother.kt
@@ -46,6 +46,7 @@ import io.getstream.chat.android.client.events.ReminderDeletedEvent
 import io.getstream.chat.android.client.events.ReminderUpdatedEvent
 import io.getstream.chat.android.client.events.TypingStartEvent
 import io.getstream.chat.android.client.events.TypingStopEvent
+import io.getstream.chat.android.client.events.UserMessagesDeletedEvent
 import io.getstream.chat.android.client.events.UserStartWatchingEvent
 import io.getstream.chat.android.client.extensions.cidToTypeAndId
 import io.getstream.chat.android.client.parser2.adapters.internal.StreamDateFormatter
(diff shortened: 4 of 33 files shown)


stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/database/internal/ChatDatabase.kt
```diff
@@ -86,7 +86,7 @@ import io.getstream.chat.android.offline.repository.domain.user.internal.UserEnt
         ThreadOrderEntity::class,
         DraftMessageEntity::class,
     ],
-    version = 91,
+    version = 92,
```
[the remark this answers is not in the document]
VelikovPetar: Not needed, we fallback to destructive migration.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/query/request/ChannelFilterRequest.kt
```diff
@@ -23,14 +23,16 @@ import io.getstream.chat.android.models.Channel
...
 import io.getstream.result.Result
 
-/* Default filter to include FilterObject in a channel by its cid
-*
-* @param filter - the filter to be included.
-* @param offset - the offset to be included with the filter.
-* @param limit - the filter to be included with the filter.
-*/
 @InternalStreamChatApi
 public object ChannelFilterRequest {
```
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5931/commits/5d23105589e33985943a3f33afd02dce7f905a8e

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryThreadsRequest.kt
```diff
@@ -33,6 +68,8 @@ import io.getstream.chat.android.models.User
  * @property userId The user ID for which the threads are queried. Defaults to null.
  */
 public data class QueryThreadsRequest @JvmOverloads constructor(
+    public val filter: FilterObject? = null,
+    public val sort: QuerySorter<Thread> = DefaultSort,
```
andremion, line 74: What if the customer wants no sort?
I can see that the API supports nullable sort.
I'm fine with the current implementation, but I'm not sure if all customers want to use sorting.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryThreadsRequest.kt
```diff
@@ -42,4 +79,20 @@ public data class QueryThreadsRequest @JvmOverloads constructor(
...
+         * The default sort order for threads:
+         * 1. Unread status (unread threads first)
+         * 2. Last message timestamp (newest first)
+         * 3. Parent message ID (descending)
+         */
+        @InternalStreamChatApi
+        public val DefaultSort: QuerySorter<Thread> = QuerySortByField
+            .descByName<Thread>("has_unread")
+            .descByName("last_message_at")
+            .descByName("parent_message_id")
```
andremion, line 100: This might be part of the entire sort mechanism of the SDK, but I'm just curious: What happens if customers use wrong field names?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryThreadsRequest.kt
```diff
@@ -42,4 +79,20 @@ public data class QueryThreadsRequest @JvmOverloads constructor(
     public val replyLimit: Int = 2,
     public val user: User? = null,
     public val userId: String? = null,
```
andremion, line 85: I know that's out of the scope, but isn't it strange to have both user properties there? 🤔

VelikovPetar: This is actually some old code, we shouldn't really expose these, but we cannot remove them due to breaking changes. I can go with deprecation as a warning to not use them?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryThreadsRequest.kt
VelikovPetar: Not sure how BE handles this, but for local sorting, we will try to find the corresponding field in the `extraData` (see `Thread.getComparableField`). I believe that BE should behave the same way, it will try to sort on custom data. 
But I will check this and get back. Side note, with the `feeds` and `core` SDK we have made a [type-safe sorting system](https://github.com/GetStream/stream-core-android/blob/develop/stream-android-core/src/main/java/io/getstream/android/core/api/sort/SortField.kt), so it will be possible to improve this in the future, once we make chat depend on `core`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryThreadsRequest.kt
VelikovPetar: The `DefaultSort` actually matches the default sort on BE - first by `has_unread`, then by `last_message_at` and finally by `parent_message_id`. So even if a user has the option to pass null, it will still apply the same sort order. We could make the field nullable, but in the end, it will not make a difference.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryThreadsRequest.kt
VelikovPetar: Providing an invalid sort arg would result in an error:
> <-- HTTP FAILED: streamCode: 4, statusCode: 400, message: QueryThreads failed with error: "invalid sort parameter: [{unknown:-1},{parent_message_id:-1}], allowed sort combinations: [[has_unread last_message_at parent_message_id] [last_message_at parent_message_id] [created_at] [updated_at] [reply_count] [participant_count] [active_participant_count] [parent_message_id]]"

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryThreadsRequest.kt
VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-android/pull/5931/commits/655aa45f0bf64668711465dc8693535966185c69

Deprecated the public properties and removed their usage internally (they were ignored anyway)

PR #5935, opened 2025-09-20
Add support for Push V3

VelikovPetar:
### 🎯 Goal
- Add support for `message.updated` and `reaction.new` push notifications. 
- Add `skipPush` parameter to `ChatClient.sendReaction` to prevent sending push notifications for the reaction.
- Add `ReactionPushEmojiFactory` for customising the default emojis for reaction psh notifications.
Resolves: https://linear.app/stream/issue/AND-733/add-support-for-push-v3-messageupdated-reactionnew

#### LLC
- Introduce `ChatNotification.MessageUpdated` and `ChatNotifications.ReactionNew` + default handling
- Introduce `skipPush` parameter to `ChatClient.sendReaction` to prevent sending push notifications for the reaction
- Introduce `emojiCode/skipPush` to `Reaction`
#### Offline
- Add support for `emojiCode/skipPush` to ensure the args are respected when syncing offline reactions
#### UI common
-  Add `ReactionPushEmojiFactory` for customising the default emojis for reaction psh notifications.
#### XML
- Add `ChatUi.reactionPushEmojiFactory` for customizing the emoji codes for reaction push notifications
#### Compose
- Add `ChaTheme.reactionPushEmojiFactory` for customizing the emoji codes for reaction push notifications

### 🎨 UI Changes

| With emojiCode | Without emojiCode |
| --- | --- |
| reaction_with_emoji1 | reaction_without_emoji |

### 🧪 Testing
`reaction.new`:
1. Use an app migrated on Push V3
2. React to a message
3. A PN should be delivered to the message sender

`message.updated`
1. Use an app migrated on Push V3
2. Update a message which a different user is 'watching' (ex thread parent)
3. A PN should be delivered to a different user
4. Note: This PN is triggered by `client.updateMessage()`, but NOT by `client.partialUpdateMessage()`, so simple edit will not work.

To make sure `client.updateMessage()` is called, you can do the following:
1. User A - send a message with two images
2. User B - thread reply to the message from user A
3. User A - open the message attachments in the Media Gallery screen
4. User A - delete one attachment (triggers `client.updateMessage()`)
5. User B - should receive a PN of type `message.updated`

### DIFF
--- a/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/database/internal/ChatDatabase.kt
+++ b/stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/repository/database/internal/ChatDatabase.kt
@@ -86,7 +86,7 @@ import io.getstream.chat.android.offline.repository.domain.user.internal.UserEnt
         ThreadOrderEntity::class,
         DraftMessageEntity::class,
     ],
-    version = 92,
+    version = 93,
     exportSchema = false,
 )
 @TypeConverters(


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/ChatApi.kt
```diff
@@ -127,9 +126,6 @@ internal interface ChatApi {
     @CheckResult
     fun getDevices(): Call<List<Device>>
 
-    @CheckResult
-    fun searchMessages(request: SearchMessagesRequest): Call<List<Message>>
-
```
andremion, line 132: Nice clean up 👌🏻

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ReactionDefaults.kt
```diff
@@ -0,0 +1,41 @@
...
+    public const val THUMBS_UP: String = "like"
+
+    @InternalStreamChatApi
+    public const val THUMBS_DOWN: String = "sad"
+
+    @InternalStreamChatApi
+    public const val LOL: String = "haha"
+
+    @InternalStreamChatApi
+    public const val WUT: String = "wow"
```
andremion, line 40: Given that those constants are related to reactions, not emojis, should they follow the string contents? For example, `LIKE` instead of `THUMBS_UP`, `SAD` instead of `THUMBS_DOWN`, and so on.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ReactionPushEmojiFactory.kt
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * A "love" reaction will show the reaction type as text instead of an emoji:
+ * `Reacted :love: to "Your message"`.
+ *
+ * Important: To use this feature, you must migrate you project to use Push V3 in the Stream Dashboard.
+ *
+ * See [Overview](https://getstream.io/chat/docs/android/push_introduction/) and
+ * [Legacy Push System](https://getstream.io/chat/docs/android/legacy_push_system/).
+ */
+public interface ReactionPushEmojiFactory {
```
andremion: Given there's only one function, can we make it a `fun interface`? so that the integrators could use it this way:
```kotlin
ChatTheme(
    reactionPushEmojiFactory = ReactionPushEmojiFactory { type ->
        when (type) {
            "like" -> "👍"
            else -> null
        }
    },
)
```

andremion (commented):
Can you add more tests to increase the coverage a bit more on the new code of the LLC?  `ChatNotificationHandler`, etc

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ReactionDefaults.kt
VelikovPetar: It would be possible, we just need to consider the public api from the `SupportedReactions.DefaultreactionTypes`. The we would need to have something like:
```kotlin
public object DefaultReactionTypes {
        public const val LOVE: String = ReactionDefaults.LOVE
        public const val THUMBS_UP: String = ReactionDefaults.LIKE
        // ....
    }
```
Which could be a bit strange (note we cannot changes the names here, because it will be a breaking change)

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ReactionPushEmojiFactory.kt
VelikovPetar: Should be possible 👍

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ReactionDefaults.kt
andremion: That's fine.
We cannot change the public API; however, `ReactionDefaults` is a new one, and it's internal

VelikovPetar: Ok I will rename the new internal constants 👍

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ReactionPushEmojiFactory.kt
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5935/commits/492348c6a9a6cc64d0638fecc361199083407b4e

VelikovPetar:
> Can you add more tests to increase the coverage a bit more on the new code of the LLC? `ChatNotificationHandler`, etc

Added some extra Android tests here: https://github.com/GetStream/stream-chat-android/pull/5935/commits/9e6d7fd724ad005309a77abd37bc6338c01f5248

andremion (approved):

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/helper/ReactionDefaults.kt
VelikovPetar: After some consideration, we decided to not make this change - mostly because there other related public components which also follow this naming pattern. Renaming these constants would result in having a mismatch between those resources, which might introduce some confusion.

PR #5939, opened 2025-09-26
6.24.0 release testing

andremion:
### 🛠 Implementation details

- XML was not showing media attachments that contain `thumbUrl` only.
- Compose was showing link attachments where it should not.

The fix creates a unified local filter in the view controller used for both SDKs.

### Fix the inconsistency in the channel attachments items between XML and Compose SDKs.

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/dba050d1-6002-4c99-8571-1e5293a8ce35" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/94cccd2f-88fd-4b4c-aa65-89a5e715409e" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/attachments/ChannelMediaAttachmentsActivity.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/attachments/ChannelMediaAttachmentsActivity.kt
@@ -34,6 +34,7 @@ import io.getstream.chat.android.compose.viewmodel.channel.ChannelAttachmentsVie
 import io.getstream.chat.android.compose.viewmodel.channel.ChannelAttachmentsViewModelFactory
 import io.getstream.chat.android.models.AttachmentType
 import io.getstream.chat.android.ui.common.feature.channel.attachments.ChannelAttachmentsViewEvent
+import io.getstream.chat.android.ui.common.utils.extensions.imagePreviewUrl
 import kotlinx.coroutines.flow.collectLatest
 
 class ChannelMediaAttachmentsActivity : ComponentActivity() {
@@ -49,6 +50,7 @@ class ChannelMediaAttachmentsActivity : ComponentActivity() {
         ChannelAttachmentsViewModelFactory(
             cid = requireNotNull(intent.getStringExtra(KEY_CID)),
             attachmentTypes = listOf(AttachmentType.IMAGE, AttachmentType.VIDEO),
+            localFilter = { !it.imagePreviewUrl.isNullOrEmpty() && it.titleLink.isNullOrEmpty() },
         )
     }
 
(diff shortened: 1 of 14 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/attachments/ChannelAttachmentsViewController.kt
```diff
@@ -120,7 +123,7 @@ public class ChannelAttachmentsViewController(
         nextPage = result.next
         val items = result.messages.flatMap { message ->
             message.attachments
-                .filter { it.type in attachmentTypes }
+                .filter(localFilter)
```
VelikovPetar, line 126: Shouldn't we apply both filters here? (`it.type in attachmentTypes` + `localFilter`)

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/attachments/ChannelMediaAttachmentsActivity.kt
```diff
@@ -49,6 +50,7 @@ class ChannelMediaAttachmentsActivity : ComponentActivity() {
         ChannelAttachmentsViewModelFactory(
             cid = requireNotNull(intent.getStringExtra(KEY_CID)),
             attachmentTypes = listOf(AttachmentType.IMAGE, AttachmentType.VIDEO),
+            localFilter = { !it.imagePreviewUrl.isNullOrEmpty() && it.titleLink.isNullOrEmpty() },
```
VelikovPetar, line 53: Is this a fix that should be applied in the samples, or on SDK level?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/channel/attachments/ChannelAttachmentsViewController.kt
andremion: That was my initial implementation, but the attachment types are already filtered on the query:
`val messageFilter = Filters.`in`("attachments.type", attachmentTypes)`

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/attachments/ChannelMediaAttachmentsActivity.kt
andremion: That local filter is introduced to allow the integrator to define them at the same level as the attachment types, and now it's the same for XML and Compose.
The view models of channel attachments and their dependencies are part of the SDK, alongside the Compose UI.
The XML UI is not.

PR #5941, opened 2025-09-29
Improve testing coverage

andremion:
### 🧪 Testing

Some of the changes here modified components to be testable. Those changes should not impact any functionality.
However, it is worth testing the functionality of `DefaultStreamMediaRecorder` to ensure that no regression issues are introduced. Test sending an audio record from the message composer of both XML and Compose SDKs.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -62,22 +62,23 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Check out code
-        uses: actions/checkout@v3.1.0
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # fetch all history for sonar analysis
       - uses: GetStream/android-ci-actions/actions/setup-java@main
       - uses: GetStream/android-ci-actions/actions/gradle-cache@main
         with:
           cache-name: gradle-test
       - name: Unit tests
         run: ./scripts/ci-unit-tests.sh
-      - name: Upload testDebugUnitTest results
+      - name: Upload test results
         uses: actions/upload-artifact@v4.4.3
         if: failure()
         with:
-          name: testDebugUnitTest
-          path: ./**/build/reports/tests/testDebugUnitTest
+          name: testResults
+          path: ./**/build/reports/tests/**
       - uses: GetStream/android-ci-actions/actions/setup-ruby@main
       - name: Sonar
-        if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
         run: bundle exec fastlane run_sonar_analysis
         env:
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
--- a/scripts/coverage.gradle
+++ b/scripts/coverage.gradle
@@ -1,72 +1,111 @@
-if (!rootProject.ext.sonar.ignoreModules.contains(name) && file("src/test").exists()) {
-    apply plugin: "org.jetbrains.kotlinx.kover"
-    apply plugin: "org.sonarqube"
+ext.koverClassExcludes = [
+        '*R',
+        '*R$*',
+        '*BuildConfig',
+        '*Manifest*',
+        '*Composable*'
+]
+ext.koverAnnotationExcludes = [
+        'androidx.compose.ui.tooling.preview.Preview'
+]
 
-    afterEvaluate {
-        def isAndroidModule = plugins.hasPlugin("com.android.library") || plugins.hasPlugin("com.android.application")
-        def hasPaparazziPlugin = plugins.hasPlugin("app.cash.paparazzi")
-
-        if (isAndroidModule) {
-            android {
-                buildTypes {
-                    debug {
-                        testCoverageEnabled = true
-                        enableUnitTestCoverage = true
-                        enableAndroidTestCoverage true
-                    }
+def configureKoverReports(project) {
+    project.kover {
+        reports {
+            verify {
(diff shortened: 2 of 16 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
```diff
@@ -139,6 +139,9 @@ public class MediaGalleryPreviewActivity : AppCompatActivity() {
         val streamCdnImageResizing = intent?.createStreamCdnImageResizing()
             ?: StreamCdnImageResizing.defaultStreamCdnImageResizing()
         val messageId = uiState?.messageId ?: ""
+        if (messageId.isBlank()) {
```
VelikovPetar: Just curious why this changed?

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/threads/ThreadListControllerTest.kt
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        fun get(
+            scope: CoroutineScope,
+            query: QueryThreadsRequest = QueryThreadsRequest(),
+        ) = ThreadListController(
+            query = query,
+            chatClient = mockChatClient,
+            scope = scope,
+            queryThreadsAsState = { queryThreadsStateFlow },
```
VelikovPetar, line 211: Think out loud here: Do we have to pass the `queryThreadsAsState` from the outside? 
Can't we instead mock the `chatClient.queryThreadsAsState()` method on the `ChatClient` instance that we pass to the `Controller`?

scripts/coverage.gradle
```diff
@@ -1,72 +1,109 @@
-if (!rootProject.ext.sonar.ignoreModules.contains(name) && file("src/test").exists()) {
-    apply plugin: "org.jetbrains.kotlinx.kover"
-    apply plugin: "org.sonarqube"
+ext.koverClassExcludes = [
+        '*R',
+        '*R$*',
+        '*BuildConfig',
+        '*Manifest*',
+        '*Composable*'
```
VelikovPetar, line 17: Curious why `*Composable*`is excluded? DOn't we already cover that code via the snapshot tests?

.github/workflows/pr-checks.yml
```diff
@@ -62,22 +62,23 @@ jobs:
...
         uses: actions/upload-artifact@v4.4.3
         if: failure()
         with:
-          name: testDebugUnitTest
-          path: ./**/build/reports/tests/testDebugUnitTest
+          name: testResults
+          path: ./**/build/reports/tests/**
       - uses: GetStream/android-ci-actions/actions/setup-ruby@main
       - name: Sonar
-        if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
```
VelikovPetar, line 80: Are we supposed to remove this?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
andremion: I had added it because I found a required non-empty check in a later stage of the rendering. So I made that check to be done earlier. I will try to find it again and might remove this part if I have confused myself 😅

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/threads/ThreadListControllerTest.kt
andremion: `ChatClient.queryThreadsAsState` is an extension function.

scripts/coverage.gradle
andremion: Those exclusions are for generated classes, not annotations.
Compose generates classes with `Composable` in their names.
Generated classes should not be covered.

.github/workflows/pr-checks.yml
andremion: Yes.
That was probably a copy-and-paste issue.
That condition is invalid, as this workflow is made for pull requests

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/threads/ThreadListControllerTest.kt
VelikovPetar: Ah right, it's part of the `-State` logic, disregard this comments.

scripts/coverage.gradle
VelikovPetar: Got it, thanks for the explanation!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaGalleryPreviewActivity.kt
andremion: I couldn't find the other check, so I removed the new one https://github.com/GetStream/stream-chat-android/pull/5941/commits/3fffd0e84be2f3733611933884090916ed7d145f

PR #5942, opened 2025-09-29
Add support for push preferences

VelikovPetar:
### 🎯 Goal
Add support for push preferences to control the push notifications per channel and globally.
Resolves: https://linear.app/stream/issue/AND-734/add-support-for-push-preferences

#### LLC
Adds the following operations for setting the push level:
- `ChatClient.setUserPushPreference(level: PushPreferenceLevel)` - for setting the user (global) preferences
- `ChatClient.setChannelPushPreference(cid: String, level: PushPreferenceLevel)` - for setting channel specific preferences

Where `PushPreferenceLevel` can be:
- `all` - allow all notificaitons
- `mentions` - allow only notifications for mentions
- `none` - disallow notification

Add the following operation for pausing/snoozing notifications:
- `ChatClient.snoozeUserPushNotifications(until: Date)` - for global snoozing of notifications
- `ChatClient.snoozeChannelPushNotifications(until: Date)` - for channel specific snoozing of notifications

Extends the `User` and `Channel` models with `pushPreference` field, holding the preferences for the current user or the specific channel.

#### Sample App (Compose)
Adds a configuration screen for managing the user(global) push preferences.

### 🎨 UI Changes

**No UI changes in the SDKs. The following demo is only in the Compose Sample App.**

<table>
<thead>
<tr>
<th>Demo (Sample)</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/002ebd99-a977-41f4-a652-55444c816b52" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

#### Global config
1. Open Compose Sample app
2. Open side bar and tap of profile
3. Tap on `Push Preferences`
4. Set any of the push levels and/or snooze the notifications
5. Send notification from a different device (and user) to verify that the set push levels are respected


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/profile/UserProfileViewModel.kt
```diff
@@ -52,24 +47,12 @@ class UserProfileViewModel(
...
-            val request = QueryUsersRequest(filter, offset = 0, limit = 1)
-            chatClient.queryUsers(request)
-                .await()
-                .onSuccessSuspend { users -> emit(users.firstOrNull()) }
-                .onErrorSuspend { error -> _events.emit(UserProfileViewEvent.LoadUserError(error)) }
+        viewModelScope.launch {
+            chatClient.clientState.user
+                .filterNotNull()
+                .collectLatest { user ->
+                    _state.update { it.copy(user = user, progressIndicator = null) }
```
andremion: No need to `progressIndicator = null`, as null is the initial state

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/profile/UserProfileScreen.kt
```diff
@@ -155,6 +160,23 @@ fun UserProfileScreen(
             UnreadCounts(state.unreadCounts)
         }
 
+        ModalSheet.PushPreferences -> ModalBottomSheet(
```
andremion, line 164: This sheet looks better with `skipPartiallyExpanded = true`

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/profile/UserProfilePushPreferencesScreen.kt
```diff
@@ -0,0 +1,483 @@
...
+                ) {
+                    Text(
+                        text = "✓",
+                        style = ChatTheme.typography.body.copy(
+                            color = ChatTheme.colors.primaryAccent,
+                            fontSize = 18.sp,
+                        ),
+                    )
+                }
+            }
```
andremion: I don't think this UI is necessary, as we already have the radio button that highlights the selected option

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/profile/UserProfilePushPreferencesScreen.kt
```diff
@@ -0,0 +1,483 @@
...
+    isSelected: Boolean,
+    isEnabled: Boolean,
+    onSelect: () -> Unit,
+) {
+    Column {
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .minimumInteractiveComponentSize()
+                .clickable(enabled = isEnabled, onClick = onSelect)
```
andremion: Can you use the ripple effect there?

```suggestion
                .clickable(
                    interactionSource = null,
                    indication = ripple(),
                    enabled = isEnabled,
                    onClick = onSelect
                )
```

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/profile/UserProfilePushPreferencesScreen.kt
```diff
@@ -0,0 +1,483 @@
...
+            add(Calendar.HOUR_OF_DAY, 1)
+        }.time
+        mutableStateOf(preferences.disabledUntil ?: fallback)
+    }
+
+    Column(
+        modifier = Modifier
+            .fillMaxWidth()
+            .background(ChatTheme.colors.appBackground)
+            .padding(16.dp),
```
andremion: I believe it's better to let the children define their padding so we can make the clickable semantics expand to the left and right bounds and have this:
Image
instead of:
Image

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/PushPreference.kt
```diff
@@ -0,0 +1,73 @@
...
+    public val level: PushPreferenceLevel?,
+    public val disabledUntil: Date?,
+)
+
+/**
+ * Represents the possible levels for chat push notifications.
+ *
+ * @param value The string representation of the chat level.
+ */
+public data class PushPreferenceLevel(public val value: String) {
```
andremion, line 40: Just out of curiosity: what was the reason to choose a data class instead of an interface or an enum for that model?

andremion (commented):
Good job! 🚀 
The only thing I noticed during the manual testing was this:
When I snooze notifications, they become disabled even after the `disabledUntil` date.
I had to go to the user profile and enable it + mark `All notifications` again to have them working.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/profile/UserProfilePushPreferencesScreen.kt
VelikovPetar: Good point!

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/PushPreference.kt
VelikovPetar: The idea is to be a bit more future proof: Assuming the scenario where the backend introduces a new push preferences level. If we define the `PushPreferenceLevel` as an `enum`, we must do an SDK update to support the new level. Plus, if we try to deserialise such value delivered from BE, we would fail (although this is a breaking change, so it would never happen). When defined as a `PushPreferenceLevel(public val value: String)` a customer could easily provide the newly introduce level, without the need for an update (and we can make the update at any convenient time). 
Basically the idea is to make it easier to integrate new values for the `PushPreferenceLevel` and reduce the possibility of having to introduce a breaking change,

VelikovPetar:
> Good job! 🚀 The only thing I noticed during the manual testing was this: When I snooze notifications, they become disabled even after the `disabledUntil` date. I had to go to the user profile and enable it + mark `All notifications` again to have them working.

Hmm interesting catch. I actually retested this (sending notifications from user `Petar` to `Andre` after `disabledUntil`) and the push were delivered, but the config screen still stayed in the `disabled` state. This is because the BE still delivers the `disabledUntil` even though it is in the past. (I will change this in the sample)

Can you maybe tell me the users with which you faced this issue?

andremion:
> > Good job! 🚀 The only thing I noticed during the manual testing was this: When I snooze notifications, they become disabled even after the `disabledUntil` date. I had to go to the user profile and enable it + mark `All notifications` again to have them working.
> 
> Hmm interesting catch. I actually retested this (sending notifications from user `Petar` to `Andre` after `disabledUntil`) and the push were delivered, but the config screen still stayed in the `disabled` state. This is because the BE still delivers the `disabledUntil` even though it is in the past. (I will change this in the sample)
> 
> Can you maybe tell me the users with which you faced this issue?

I was testing with R2-D2 receiving PNs and C3-PO sending

VelikovPetar:
> > > Good job! 🚀 The only thing I noticed during the manual testing was this: When I snooze notifications, they become disabled even after the `disabledUntil` date. I had to go to the user profile and enable it + mark `All notifications` again to have them working.
> > 
> > 
> > Hmm interesting catch. I actually retested this (sending notifications from user `Petar` to `Andre` after `disabledUntil`) and the push were delivered, but the config screen still stayed in the `disabled` state. This is because the BE still delivers the `disabledUntil` even though it is in the past. (I will change this in the sample)
> > Can you maybe tell me the users with which you faced this issue?
> 
> I was testing with R2-D2 receiving PNs and C3-PO sending

Is it possible that it was a one-time occurrence? At the moment, I cannot reproduce this unfortunately :/ 

https://github.com/user-attachments/assets/01be199d-162a-453c-b2be-858d66b8bc1b

andremion:
> > > > Good job! 🚀 The only thing I noticed during the manual testing was this: When I snooze notifications, they become disabled even after the `disabledUntil` date. I had to go to the user profile and enable it + mark `All notifications` again to have them working.
> > > 
> > > 
> > > Hmm interesting catch. I actually retested this (sending notifications from user `Petar` to `Andre` after `disabledUntil`) and the push were delivered, but the config screen still stayed in the `disabled` state. This is because the BE still delivers the `disabledUntil` even though it is in the past. (I will change this in the sample)
> > > Can you maybe tell me the users with which you faced this issue?
> > 
> > 
> > I was testing with R2-D2 receiving PNs and C3-PO sending
> 
> Is it possible that it was a one-time occurrence? At the moment, I cannot reproduce this unfortunately :/
> 
>  Screen.Recording.2025-10-07.at.13.05.30.mov

Yes. I retested and it works now 🤷🏻

andremion (approved):

PR #5944, opened 2025-10-02
Fix recompositions of MessageListStartOfTheChannelItemContent

andremion:
### 🛠 Implementation details

- Add a unique identifier for `MessageListItemState`.
- Add `distinctUntilChanged` before mapping the message list state to ensure that the reversed message list is only recreated when the actual state changes, not on every emission.

### 🧪 Testing

Apply the provided patch and look for log entries of `alor: MessageListStartOfTheChannelItemContent` when entering a channel, scrolling to the very first sent message.

<details>

<summary>MessageListStartOfTheChannelItemContent recompositions</summary>

```
Subject: [PATCH] MessageListStartOfTheChannelItemContent recompositions
---
Index: stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt	(revision 58eca85b0fd27686866ef68bcf708dc18a57edc6)
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt	(date 1759409581641)
@@ -989,6 +989,8 @@
     public fun LazyItemScope.MessageListStartOfTheChannelItemContent(
         startOfTheChannelItem: StartOfTheChannelItemState,
     ) {
+        println("alor: MessageListStartOfTheChannelItemContent")
+        Text("MessageListStartOfTheChannelItemContent")
     }
 
     /**
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@
 ## stream-chat-android-compose
 ### 🐞 Fixed
 - Fix mute/unmute instant command icons. [#5938](https://github.com/GetStream/stream-chat-android/pull/5938)
+- Fix recompositions of `MessageListStartOfTheChannelItemContent`. [#5944](https://github.com/GetStream/stream-chat-android/pull/5944) 
 
 ### ⬆️ Improved
 
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
@@ -171,9 +171,7 @@ public fun Messages(
 
             itemsIndexed(
                 items = messages,
-                key = { _, item ->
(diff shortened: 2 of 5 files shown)


VelikovPetar (approved):
Approved but. the coverage seems pretty low: should we add tests for the new `id` property?

andremion:
> Approved but. the coverage seems pretty low: should we add tests for the new `id` property?

We should not merge anything until we have https://github.com/GetStream/stream-chat-android/pull/5941 merged. 
so that we will have a better view of the coverage

PR #5946, opened 2025-10-03
Improve SwipeToReply component in scroller containers

andremion:
### 🎯 Goal

Improve `SwipeToReply` component in scroller containers by only handling gesture events if the horizontal scrolling dominates the vertical scrolling

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
</td>
<td>
<video src="https://github.com/user-attachments/assets/c43902e3-9f85-4cc2-b905-933957e83d05" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>


VelikovPetar (approved):
Looks good!
Just for my understanding, is it not possible to disable the `SwipeToReply` action if there is a vertical scroll active? (I am not suggesting it as a better solution, just as an idea).

andremion:
> Looks good! Just for my understanding, is it not possible to disable the `SwipeToReply` action if there is a vertical scroll active? (I am not suggesting it as a better solution, just as an idea).

Sorry, can you elaborate `if there is a vertical scroll active`?

VelikovPetar:
> > Looks good! Just for my understanding, is it not possible to disable the `SwipeToReply` action if there is a vertical scroll active? (I am not suggesting it as a better solution, just as an idea).
> 
> Sorry, can you elaborate `if there is a vertical scroll active`?

I was thinking something like:
```kotlin
val lazyListState = rememberLazyListState()
val swipeEnabled = !lazyListState.isScrollInProgress
```
But on a closer inspection, we cannot really easily access the the list state here unfortunately.

andremion:
> > > Looks good! Just for my understanding, is it not possible to disable the `SwipeToReply` action if there is a vertical scroll active? (I am not suggesting it as a better solution, just as an idea).
> > 
> > 
> > Sorry, can you elaborate `if there is a vertical scroll active`?
> 
> I was thinking something like:
> 
> ```kotlin
> val lazyListState = rememberLazyListState()
> val swipeEnabled = !lazyListState.isScrollInProgress
> ```
> 
> But on a closer inspection, we cannot really easily access the the list state here unfortunately.

Yes.
It would introduce a strong dependency on an upper state of the hierarchy.
The provided approach seems more flexible as it can be useful for any application, not only related to the scrolling of a list.

PR #5949, opened 2025-10-06
PR cleanup workflow

andremion:
### 🎯 Goal

Mark and close state PRs


.github/workflows/pr-cleanup.yaml
```diff
@@ -0,0 +1,17 @@
+name: PR cleanup
+
+on:
+  schedule:
+    - cron: "0 7 * * 1"  # every Monday 09:00 CEST (UTC+2)
```
gpunto: Note that this time changes twice a year (next time at the end of October), so it would be stale for 6 months. In feeds we put:
```
# every Monday 07:00 UTC (08:00 CET / 09:00 CEST)
```

gpunto (approved):

PR #5954, opened 2025-10-08
Expose `StorageHelper` and `AttachmentFilter` to fix `StorageHelperWrapper` not instantiable

VelikovPetar:
### 🎯 Goal
The public `StorageHelperWrapper` cannot be instantiated because it has constructor dependencies on 2 classes marked as `@InternalStreamChatApi`:
- `StorageHelper`
- `AttachmentFilter`

This PR exposes both `StorageHelper` and `AttachmentFilter` as `public`, as it is the only way to allow instantiation of `StorageHelperWrapper`

### 🛠 Implementation details
- Expose `StorageHelper` and `AttachmentFilter` as `public`.
- Add missing KDocs on the new public API

### 🎨 UI Changes
_NA_

### 🧪 Testing
- Instantiate `StorageHelperWrapper` outside of the SDK modules (ex. in `stream-chat-android-compose-sample`)

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/StorageHelperWrapper.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/StorageHelperWrapper.kt
@@ -35,8 +35,8 @@ import io.getstream.chat.android.ui.common.state.messages.composer.AttachmentMet
  */
 public class StorageHelperWrapper(
     private val context: Context,
-    private val storageHelper: StorageHelper,
-    private val attachmentFilter: AttachmentFilter,
+    private val storageHelper: StorageHelper = StorageHelper(),
+    private val attachmentFilter: AttachmentFilter = AttachmentFilter(),
 ) {
 
     /**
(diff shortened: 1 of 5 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/StorageHelperWrapper.kt
```diff
@@ -35,8 +35,8 @@ import io.getstream.chat.android.ui.common.state.messages.composer.AttachmentMet
  */
 public class StorageHelperWrapper(
     private val context: Context,
-    private val storageHelper: StorageHelper,
-    private val attachmentFilter: AttachmentFilter,
+    private val storageHelper: StorageHelper = StorageHelper(),
+    private val attachmentFilter: AttachmentFilter = AttachmentFilter(),
```
andremion, line 39: optional: Since those have default values, can you replace occurrences of:
`StorageHelperWrapper(context, StorageHelper(), AttachmentFilter())`
with 
`StorageHelperWrapper(context)`
?

VelikovPetar: Done: https://github.com/GetStream/stream-chat-android/pull/5954/commits/c9fdb2acb5516e6042657b9aa1655909355952dd

PR #5966, opened 2025-10-15
Add missing polls LLC operations

VelikovPetar:
### 🎯 Goal
Implements poll and poll options management operations on the LLC.
Added:
- `ChatClient.getPoll(pollId: String)`
- `ChatClient.updatePoll(poll: UpdatePollRequest)`
- `ChatClient.partialUpdatePoll(pollId: String, set: Map<String, Any>, unset: List<String>)`
- `ChatClient.createPollOption(pollId: String, option: PollOption)`
- `ChatClient.updatePollOption(pollId: String, option: PollOption)`
- `ChatClient.deletePollOption(pollId: String, optionId: String)`
- `ChatClient.queryPollVotes(pollId: String, filter: FilterObject?, limit: Int?, next: String?, sort: QuerySorter<Vote>?)`
- `ChatClient.queryPolls(filter: FilterObject?, limit: Int?, next: String?, sort: QuerySorter<Poll>?)`
- `ChatClient.castPollVote(messageId: String, pollId: String, optionId: String)`
- `ChatClient.removePollVote(messageId: String, pollId: String, voteId: String)`
- `PollConfig` (used in `ChatClient.sendPoll`) constructor which accepts extra data for the poll and the poll options

Deprecated:
- `PollConfig` constructor without extra data
- `ChatClient.suggestPollOption` in favour of `ChatClient.createPollOption`
- `ChatClient.castPollVote(messageId: String, pollId: String, option: Option)` in favor of `ChatClient.castPollVote(messageId: String, pollId: String, optionId: String)`
- `ChatClient.removePollVote(messageId: String, pollId: String, voteId: String)` in favor of `ChatClient.removePollVote(messageId: String, pollId: String, vote: Vote)`

Resolves: https://linear.app/stream/issue/AND-793/align-llc-polls-api
Docs PR: https://github.com/GetStream/docs-content/pull/690

### 🛠 Implementation details
Implements the missing poll-related operations, alongside the Retrofit APIs, network/domain models, and offline storage updates. Replaces the usage of the deprecated methods with the new ones.

Resolves: 
https://linear.app/stream/issue/AND-793/align-llc-polls-api

### 🎨 UI Changes
_NA_

### 🧪 Testing
Standard poll testing:
1. Create a poll (with different options)
2. Add / Change / Delete votes
3. Add comment
4. Create option
5. Close poll
Ensure everything works fine

For the new endpoints - they are not linked to the UI, so you need to manually call them and verify they return the expected results.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -30,7 +30,6 @@ import io.getstream.chat.android.client.ChatClient.Companion.MAX_COOLDOWN_TIME_S
 import io.getstream.chat.android.client.api.ChatApi
 import io.getstream.chat.android.client.api.ChatClientConfig
 import io.getstream.chat.android.client.api.ErrorCall
-import io.getstream.chat.android.client.api.models.CreatePollOptionRequest
 import io.getstream.chat.android.client.api.models.GetThreadOptions
 import io.getstream.chat.android.client.api.models.PinnedMessagesPagination
 import io.getstream.chat.android.client.api.models.QueryChannelRequest
@@ -38,7 +37,6 @@ import io.getstream.chat.android.client.api.models.QueryChannelsRequest
 import io.getstream.chat.android.client.api.models.QueryThreadsRequest
 import io.getstream.chat.android.client.api.models.QueryUsersRequest
 import io.getstream.chat.android.client.api.models.SendActionRequest
-import io.getstream.chat.android.client.api.models.UpdatePollOptionRequest
 import io.getstream.chat.android.client.api.models.UpdatePollRequest
 import io.getstream.chat.android.client.api.models.identifier.AddDeviceIdentifier
 import io.getstream.chat.android.client.api.models.identifier.ConnectUserIdentifier
@@ -193,6 +191,7 @@ import io.getstream.chat.android.models.Option
 import io.getstream.chat.android.models.PendingMessage
 import io.getstream.chat.android.models.Poll
 import io.getstream.chat.android.models.PollConfig
+import io.getstream.chat.android.models.PollOption
 import io.getstream.chat.android.models.PushMessage
 import io.getstream.chat.android.models.PushPreference
 import io.getstream.chat.android.models.PushPreferenceLevel
@@ -1996,23 +1995,23 @@ internal constructor(
         pollId: String,
         option: String,
     ): Call<Option> {
-        val request = CreatePollOptionRequest(option)
-        return createPollOption(pollId, request)
+        return createPollOption(pollId, option = PollOption(text = option))
+            .map { Option(id = it.id ?: "", text = it.text, extraData = it.extraData) }
     }
 
     /**
      * Create a new option for a poll.
      *
      * @param pollId The poll id.
-     * @param option The option to create.
+     * @param option The option to create. Note: Don't pass [PollOption.id] as it is ignored for creation.
      *
      * @return Executable async [Call] responsible for creating a new option.
      */
     @CheckResult
     public fun createPollOption(
         pollId: String,
-        option: CreatePollOptionRequest,
-    ): Call<Option> {
+        option: PollOption,
+    ): Call<PollOption> {
         return api.createPollOption(pollId, option)
     }
 
@@ -2020,13 +2019,13 @@ internal constructor(
      * Update an existing option in a poll.
      *
      * @param pollId The poll id.
-     * @param option The option to update.
+     * @param option The option to update. Note: [PollOption.id] is mandatory for performing an update.
      */
     @CheckResult
     public fun updatePollOption(
         pollId: String,
-        option: UpdatePollOptionRequest,
-    ): Call<Option> {
+        option: PollOption,
+    ): Call<PollOption> {
         return api.updatePollOption(pollId, option)
     }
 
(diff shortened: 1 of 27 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -511,6 +520,22 @@ internal class DomainMapping(
         else -> throw IllegalArgumentException("Unknown voting visibility: $this")
     }
 
+    /**
+     * Transforms [DownstreamPushPreferenceDto] to [PushPreference].
```
andremion:
```suggestion
     * Transforms [QueryPollVotesResponse] to [QueryPollVotesResult].
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1951,29 +2110,6 @@ internal constructor(
         return api.removePollVote(messageId, pollId, vote.id)
```
andremion: What do you think about creating:

```kotlin
removePollVote(
    messageId: String,
    pollId: String,
    voteId: String,
):
```
and deprecate:
```kotlin
removePollVote(
    messageId: String,
    pollId: String,
    vote: Vote,
):
```
?
So that would be align with the specs https://www.notion.so/stream-wiki/Polls-c53f9695fb4e4d9083e5b6b2c49017fd#79a88ba6da7e4fbe8b7e05ce119ba1bc

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1898,12 +1873,196 @@ internal constructor(
...
+    /**
+     * Partially update a poll.
+     *
+     * @param pollId The poll id.
+     * @param set Map of fields to set.
+     * @param unset List of fields to unset.
+     * @return Executable async [Call] responsible for updating a poll.
+     */
+    @CheckResult
+    public fun partialUpdatePoll(
```
andremion, line 1952: What about the full update poll?

https://www.notion.so/stream-wiki/Polls-c53f9695fb4e4d9083e5b6b2c49017fd?pvs=97&qid=1:147f7b66-a0d0-4513-933b-b266a4aabd8e:1#db2aec6dc74a49bb8a0e221d38991465

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1898,12 +1873,196 @@ internal constructor(
...
+     *
+     * @param pollId The poll id.
+     * @param option The option to create.
+     *
+     * @return Executable async [Call] responsible for creating a new option.
+     */
+    @CheckResult
+    public fun createPollOption(
+        pollId: String,
+        option: CreatePollOptionRequest,
```
andremion: Since this is a public API, what do you think about removing the `Request` from the data class name?
Maybe also consider reusing the `PollOption` model.
The `id` could be introduced to `PollOption` as optional, so it could be reused by create and update, instead of using `CreatePollOptionRequest` and `UpdatePollOptionRequest` models

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: I will add it!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: I was actually considering to use the `PollOption`, but I wasn't; sure about defining it as:
```kotlin
data class PollOption(val id: String? = null)
```
But I was worried about being a bit confusing for usars. Let me try again to see how it would affect the API, and I will get back to you.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: Good point, I can do that!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: The thing with using `PollOption` with optional `id` is that the ID is mandatory for the `update` operation, but not for the `create` one. So it might not be obvious that the users must supply a `PollOption` option with ID, otherwise they would end up in an error. 

I would be OK with this since it will be documented, but I would also hear your opinion on this.

andremion: It makes sense. What about removing the `Request` suffix?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -471,7 +472,7 @@ internal class DomainMapping(
     }
 
     /**
-     * Transforms DownstreamOptionDto to Option
+     * Transforms [DownstreamPollOptionDto] to [Option]
```
andremion, line 475: `Transforms [DownstreamPollOptionDto] to [PollOption]`

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
andremion: My bad.
It's correct

PR #5967, opened 2025-10-15
Allow users to delete messages only for themselves (`Delete message for me` - feature)

andremion:
### 🎯 Goal

Introduce the "Delete message for me" feature to the LLC, allowing users to delete messages only from their own view while keeping them visible to other users.

### 🛠 Implementation details

- Added a new `deletedForMe` field to the `Message` model, corresponding database entity, and `MessageDeletedEvent`, along with introducing `ChatClient.deleteMessageForMe` function.
- Added XML and Compose sample implementations.

### 🎨 UI Changes

> [!NOTE]
> Demo apps only. No UI is included in the SDK for this feature.

<table>
<thead>
<tr>
<th>XML</th>
<th>Compose</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/422bfee6-5f62-4bd1-b6bc-2845219dc9a1" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/1aa1e594-5919-4288-a2a8-530ae2ec6943" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

#### Using both XML and Compose demo apps:

- Open a channel.
- Long-press any current user message.
- Click on the "Delete Message for Me" option.
- It should show "Message deleted" and show the "Deleted only for me" label.
- It should not delete the message for the other chat members.

#### For different visibilities when a message is deleted, you can change `deletedMessageVisibility`:

`DeletedMessageVisibility.ALWAYS_VISIBLE`:

- It should show the "Message deleted" for all the members when a message is deleted.
- It should show the "Message deleted" + "Deleted only for me" label for the current user only.

`DeletedMessageVisibility.ALWAYS_HIDDEN`:

- It should not show the "Message deleted" nor "Deleted only for me" when a message is either deleted or deleted for the current user only.

`DeletedMessageVisibility.VISIBLE_FOR_CURRENT_USER`:

- It should not show the "Message deleted" for all the members when a message is deleted.
- It should show the "Message deleted" + "Deleted only for me" label for the current user only.

#### Compose

Update `MessagesActivity.factory` to change the `deletedMessageVisibility`

#### XML

Update `ChatFragment.initMessageListViewModel` to `setDeletedMessageVisibility`

> [!IMPORTANT]
> Test in both online and offline modes

### DIFF
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/sync/internal/SyncManager.kt
@@ -532,7 +532,7 @@ internal class SyncManager(
     }
 
     @VisibleForTesting
-    private suspend fun retryMessages() {
+    internal suspend fun retryMessages() {
         logger.d { "[retryMessages] no args" }
         retryMessagesWithSyncedAttachments()
         retryMessagesWithPendingAttachments()
@@ -562,7 +562,10 @@ internal class SyncManager(
             val message = repos.selectMessage(id) ?: return@forEach
             val channelClient = chatClient.channel(message.cid)
             val result = when {
-                message.isDeleted() -> retryDeletionOfMessageWithSyncedAttachments(id, message, channelClient)
+                message.isDeleted() && !message.deletedForMe -> {
+                    retryDeletionOfMessageWithSyncedAttachments(id, message, channelClient)
+                }
+                message.deletedForMe -> retryDeleteMessageForMe(id)
                 message.updatedLocallyAt != null && message.createdAt != null -> {
                     retryUpdateOfMessageWithSyncedAttachments(id, message, channelClient)
                 }
@@ -571,6 +574,11 @@ internal class SyncManager(
             }
             logger.v { "[retryMgsWithSyncedAttachments] result(${message.id}).isSuccess: ${result is Result.Success}" }
         }
+    }
+
+    private suspend fun retryDeleteMessageForMe(messageId: String): Result<Message> {
+        logger.d { "[retryDeleteMessageForMe] messageId: $messageId" }
+        return chatClient.deleteMessageForMe(messageId).await()
     }
 
     /**
--- a/stream-chat-android-state/src/test/java/io/getstream/chat/android/state/internal/SyncManagerTest.kt
+++ b/stream-chat-android-state/src/test/java/io/getstream/chat/android/state/internal/SyncManagerTest.kt
@@ -226,8 +226,8 @@ internal class SyncManagerTest {
     fun `test sync max threshold for messages`() = runTest(testDispatcher) {
         /* Given */
         val message1 = localRandomMessage()
-        val message2 = randomMessage(deletedAt = localDate())
-        val message3 = randomMessage(deletedAt = null)
+        val message2 = randomMessage(deletedAt = localDate(), deletedForMe = false)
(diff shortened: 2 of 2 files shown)


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/DeleteMessageForMeComponentFactory.kt
```diff
@@ -0,0 +1,148 @@
...
+
+                else -> onMessageAction(action)
+            }
+        }
+
+        if (showDeleteForMeConfirmationDialog) {
+            SimpleDialog(
+                modifier = Modifier.padding(16.dp),
+                title = "Delete for me",
+                message = "Are you sure you want to delete this message for you?",
```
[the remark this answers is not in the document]
andremion: Those are from the demo app only

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/DeleteMessageForMeComponentFactory.kt
```diff
@@ -0,0 +1,148 @@
...
+    override fun MessageFooterContent(messageItem: MessageItemState) {
+        Row(
+            verticalAlignment = Alignment.CenterVertically,
+            horizontalArrangement = Arrangement.spacedBy(8.dp),
+        ) {
+            super.MessageFooterContent(messageItem)
+            if (messageItem.message.deletedForMe) {
+                Text(
+                    modifier = Modifier.padding(vertical = 4.dp),
+                    text = "Deleted only for me",
```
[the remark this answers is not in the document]
andremion: Those are from the demo app only

stream-chat-android-offline/src/main/java/io/getstream/chat/android/offline/plugin/listener/internal/DeleteMessageForMeListenerDatabase.kt
```diff
@@ -0,0 +1,71 @@
...
+                messageRepository.insertMessage(
+                    result.value.copy(syncStatus = SyncStatus.COMPLETED),
+                )
+            }
+
+            is Result.Failure -> {
+                messageRepository.selectMessage(messageId)?.let { message ->
+                    val updatedMessage = message.copy(
+                        deletedForMe = true,
+                        syncStatus = SyncStatus.SYNC_NEEDED,
```
VelikovPetar, line 63: Should we distinguish here between permanent and non-permanent errors? 
I am thinking if we can end up stuck in a loop retrying the delete, if the operation fails with a permanent error. (`Error.isPermanent()`)

andremion: Do you know how likely is this to happen?
I'm using the same logic as the `DeleteMessageListenerDatabase`

VelikovPetar: Hmm I see... As the new operation uses the same endpoint as the `deleteMessage`, let's then keep it as it is.

PR #5968, opened 2025-10-15
Fixed an issue where a poll would not update if its a thread root

VelikovPetar:
### 🎯 Goal
Fixes the case where polls are not updated live when they are the root message of a thread.
Adds`HasPoll` event handling to the `ThreadLogic`.

### 🛠 Implementation details
1. Add the `messageId` property on the `HasPoll` events - used to identify the thread in which the poll lives
2. Add handling for each `HasPoll` event in the `ThreadLogic` - updating the thread parent message state

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/6ed5850c-4a14-41d0-996e-25e7a84f96eb" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/51ec9e39-092a-48ad-8259-114c004f8933" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Create a poll
2. Thread reply to the poll
3. Open the thread
4. Interact with the poll
5. The poll should update live

### DIFF
--- a/stream-chat-android-compose/detekt-baseline.xml
+++ b/stream-chat-android-compose/detekt-baseline.xml
@@ -47,7 +47,6 @@
     <ID>MagicNumber:WaveformSlider.kt$50</ID>
     <ID>MaxLineLength:AttachmentsPickerTabFactories.kt$AttachmentsPickerTabFactories$*</ID>
     <ID>MaxLineLength:ChatTheme.kt$error("No attachments picker tab factories provided! Make sure to wrap all usages of Stream components in a ChatTheme.")</ID>
-    <ID>MaxLineLength:MessageItem.kt$message.isDeleted() &amp;&amp; messageItem.deletedMessageVisibility == DeletedMessageVisibility.VISIBLE_FOR_CURRENT_USER</ID>
     <ID>MaxLineLength:MessageOptions.kt$title = if (selectedMessage.pinned) R.string.stream_compose_unpin_message else R.string.stream_compose_pin_message</ID>
     <ID>MaxLineLength:MessagesScreen.kt$*</ID>
     <ID>MaxLineLength:PollCreationHeader.kt$*</ID>


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
```diff
@@ -74,6 +74,7 @@ public sealed interface HasOwnUser {
 }
 
 public sealed interface HasPoll {
+    public val messageId: String?
```
andremion, line 77: Shouldn't we keep this as non-nullable?
I can see `message_id` in every poll event.
Which event doesn't have a `message_id`?

VelikovPetar: This actually follows the specs from [BE code](https://github.com/GetStream/chat/blob/master/events/poll.go#L49) (see `omitempty` marker on the `PollEvent`). While this will most likely never be null for the chat events (note: it will be empty for feeds-related event), I am bit more inclined to keep it aligned to how it is specified on BE. I think it is safer to be have more relaxed checks (keep it nullable), because if we are more strict, we will end up falling to parse the WS events.

andremion: > it will be empty for feeds-related event

You mean empty or null?

Should we consider keeping the DTO as nullable but the model as non-null?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/thread/internal/ThreadLogic.kt
```diff
@@ -128,4 +137,25 @@ internal class ThreadLogic(
...
+                    is PollClosedEvent -> event.processPoll { poll }
+                    is PollUpdatedEvent -> event.processPoll { poll }
+                    is VoteRemovedEvent -> event.processPoll { poll }
+                    is VoteCastedEvent -> event.processPoll(currentUserId) { poll }
+                    is VoteChangedEvent -> event.processPoll(currentUserId) { poll }
+                    is PollDeletedEvent -> null // poll is deleted, remove from state
+                }
+                mutableState.updateParentMessagePoll(processedPoll)
+            }
+    }
```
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5968/commits/fcea8ad6d9eed2b68baf5fb4192c5fccb8fa3b71

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/events/ChatEvent.kt
VelikovPetar: I meant null. 
> Should we consider keeping the DTO as nullable but the model as non-null?

If we do this, it will no longer be possible to convert the model to nullable, as it is a public API.

andremion: > If we do this, it will no longer be possible to convert the model to nullable, as it is a public API.
I know.

Let's keep it as nullable since feeds doesn't need that anyway.

PR #5970, opened 2025-10-16
Do not allow sending polls in threads

VelikovPetar:
### 🎯 Goal
The BE disallows sending polls inside a thread. We currently don't respect this, and we still show the Poll attachment picker even when inside a thread. 
Furthermore, when we send the poll, we don't link it to the thread, so it will be posted inside the main channel instead. 
In this PR, we remove the Poll tab, when the attachment picker is opened inside a thread.

#### Compose
- Pass `MessageMode` to the `AttachmentsPicker` composable
- Extend the `AttachmentsPickerTabFactoryFilter` with checks for the current `MessageMode`
#### XML
- Extend the check for `arePollsEnabled` with checks for the current `MessageMode`

**Note: We have slightly different implementations between XML and Compose, because the attachment pickers behave a bit differently. On XML, the `AttachmentPicker` is started through the `MessageComposerView`, so we can directly access the current state (message mode, channel config and capabilities). On compose, the `AttachmentPicker` is independent from the `MessageComposer`, so we have to pass the message mode from outside.**

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/4ab31139-8998-4332-abda-8e775b6f3ff9" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/f291435d-e5f7-4574-a8ca-33f2632c888e" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/74e852cd-1c4a-4d32-9350-1d2ab3d4e0d6" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/8047ec6f-d367-40d2-9328-7f6991d35480" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Open thread
2. Open attachment picker
3. Polls should NOT be an option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 
 ## stream-chat-android-state
 ### 🐞 Fixed
+- Fix poll state getting overridden by `message.new` events. [#5963](https://github.com/GetStream/stream-chat-android/pull/5963)
 
 ### ⬆️ Improved
 
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Thread.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Thread.kt
@@ -44,8 +44,11 @@ public fun Thread.updateParentOrReply(message: Message): Thread {
 public fun Thread.updateParent(parent: Message): Thread {
     // Skip update if [parent] is not related to this Thread
     if (this.parentMessageId != parent.id) return this
+    // Enrich the poll as it might not be present in the event
+    val poll = parent.poll ?: this.parentMessage.poll
+    val updatedParent = parent.copy(poll = poll)
     return this.copy(
-        parentMessage = parent,
+        parentMessage = updatedParent,
         deletedAt = parent.deletedAt,
         updatedAt = parent.updatedAt ?: this.updatedAt,
     )
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/extensions/internal/ThreadExtensionsTests.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/extensions/internal/ThreadExtensionsTests.kt
@@ -18,11 +18,15 @@ package io.getstream.chat.android.client.extensions.internal
 
 import io.getstream.chat.android.models.ChannelUserRead
 import io.getstream.chat.android.models.Message
+import io.getstream.chat.android.models.Option
+import io.getstream.chat.android.models.Poll
 import io.getstream.chat.android.models.Thread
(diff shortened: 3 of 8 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/factory/AttachmentsPickerTabFactoryFilter.kt
```diff
@@ -47,11 +52,23 @@ internal class AttachmentsPickerTabFactoryFilter {
...
         return AttachmentsPickerSystemTabFactory(config)
     }
+
+    private fun AttachmentsPickerTabFactory.isEnabledForMessageMode(mode: MessageMode): Boolean {
+        return when (this) {
+            // The default "Polls" tab is only shown in Normal mode
+            is AttachmentsPickerPollTabFactory -> mode is MessageMode.Normal
+            else -> true
+        }
+    }
```
tafelnl, line 73: Wouldn't it make more sense to include this logic in the `isPickerTabEnabled` method? With this PR it makes the showing logic even more scattered across the code base

VelikovPetar: That is correct, however it is not possible to do at this moment, because changing the signature of the `isPickerTabEnabled(channel: Channel)` would be a breaking change. Unfortunately the `messageMode` is not deductible from the `channel` object, so we have to 'inject' the `messageMode` in a different way. 

This is something that could be improved in the next major release.

aleksandar-apostolov (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/factory/AttachmentsPickerTabFactoryFilter.kt
tafelnl: > would be a breaking change

If you were to provide it with a default value, it doesn't have to be a breaking change. You can change the signature to for example:

```isPickerTabEnabled(channel: Channel, messageMode: MessageMode)```

and then deprecate the other signature

> This is something that could be improved in the next major release.

I don't think you're following the semver in this repository anyways. Because minor and patch versions both have contained (major) breaking changes in the past

PR #5973, opened 2025-10-18
Ensure `MessageComposer` respects the `send-` capabilities

VelikovPetar:
### 🎯 Goal
Currently the message composer only considered the `send-message` capability, ignoring the `send-reply` one. With this PR, we ensure the proper capability is checked, depending on whether the composer is used in a channel or in a thread.

Docs PR: https://github.com/GetStream/docs-content/pull/686

### 🛠 Implementation details
- Check `send-message` capability when in channel
- Check `rend-reply` capability when in thread
- Additional fix: Don't allow sending audio recordings if no rights to send a message (or reply)
- The checks are now placed in a single file: `MessageComposerCapabilities`, which expose the `canSendMessage` and `canUploadFile`. **Note: They are exposed as public, because I think they would be useful to provide to integrators developing custom composers.**

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Capabilities</th>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
`send-message`, `send-reply`
</td>
<td>
<video src="https://github.com/user-attachments/assets/808ef09c-d851-480d-8dde-ca04bfdf13c5" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/e76ba5ca-3cdf-4c77-8b42-fcbf9785dd28" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
`send-message`
</td>
<td>
<video src="https://github.com/user-attachments/assets/6cb1b630-1c08-48df-a79e-c7246aa11309" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/f731ac3d-7276-423e-9360-5b6cb41ddbe3" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
`send-reply`
</td>
<td>
<video src="https://github.com/user-attachments/assets/c617d3a6-6951-4c52-b248-920b9d693b9e" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/122ab3c1-5837-4d2f-97e4-0d359ee0e44a" controls="controls" muted="muted" />
</td>
</tr>
<tr>
<td>
/
</td>
<td>
<video src="https://github.com/user-attachments/assets/4850893e-9448-4fa3-9718-39e9598a24e1" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/6c6f2056-2533-416c-9ef0-53d5536aa167" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. To test without `send-message`, disable the `CreateMessage` permission on dashboard
2. To test without both, disable both `CreateMessage` and `CreateReply` permissions on dashboard
3. To test without `send-reply`, disable `Threads & Replies` on dashboard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 ## stream-chat-android-state
 ### 🐞 Fixed
 - Fix poll state getting overridden by `message.new` events. [#5963](https://github.com/GetStream/stream-chat-android/pull/5963)
+- Fix polls not updated live when they are a thread parent message. [#5968](https://github.com/GetStream/stream-chat-android/pull/5968)
 
 ### ⬆️ Improved
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 
 ![stream-chat-android-client](https://img.shields.io/badge/stream--chat--android--client-3.22%20MB-lightgreen)
 ![stream-chat-android-offline](https://img.shields.io/badge/stream--chat--android--offline-3.45%20MB-lightgreen)
-![stream-chat-android-ui-components](https://img.shields.io/badge/stream--chat--android--ui--components-10.53%20MB-lightgreen)
+![stream-chat-android-ui-components](https://img.shields.io/badge/stream--chat--android--ui--components-10.54%20MB-lightgreen)
 ![stream-chat-android-compose](https://img.shields.io/badge/stream--chat--android--compose-12.76%20MB-lightgreen)
 
 </div>
--- a/metrics/size.json
+++ b/metrics/size.json
@@ -2,13 +2,13 @@
   "debug": {
     "stream-chat-android-client": 4388,
     "stream-chat-android-offline": 4728,
-    "stream-chat-android-ui-components": 13280,
+    "stream-chat-android-ui-components": 13284,
     "stream-chat-android-compose": 16520
   },
   "release": {
-    "stream-chat-android-client": 3300,
+    "stream-chat-android-client": 3304,
     "stream-chat-android-offline": 3536,
-    "stream-chat-android-ui-components": 10792,
-    "stream-chat-android-compose": 13068
+    "stream-chat-android-ui-components": 10796,
+    "stream-chat-android-compose": 13072
   }
 }
--- a/stream-chat-android-client-test/src/main/java/io/getstream/chat/android/client/test/Mother.kt
+++ b/stream-chat-android-client-test/src/main/java/io/getstream/chat/android/client/test/Mother.kt
@@ -16,6 +16,7 @@
 
 package io.getstream.chat.android.client.test
 
+import io.getstream.chat.android.client.events.AnswerCastedEvent
 import io.getstream.chat.android.client.events.ChannelDeletedEvent
 import io.getstream.chat.android.client.events.ChannelHiddenEvent
 import io.getstream.chat.android.client.events.ChannelUpdatedByUserEvent
@@ -39,7 +40,9 @@ import io.getstream.chat.android.client.events.NotificationMessageNewEvent
 import io.getstream.chat.android.client.events.NotificationMutesUpdatedEvent
 import io.getstream.chat.android.client.events.NotificationReminderDueEvent
 import io.getstream.chat.android.client.events.NotificationRemovedFromChannelEvent
+import io.getstream.chat.android.client.events.PollClosedEvent
 import io.getstream.chat.android.client.events.PollDeletedEvent
+import io.getstream.chat.android.client.events.PollUpdatedEvent
 import io.getstream.chat.android.client.events.ReactionNewEvent
 import io.getstream.chat.android.client.events.ReminderCreatedEvent
 import io.getstream.chat.android.client.events.ReminderDeletedEvent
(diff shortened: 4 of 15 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -511,10 +513,7 @@ internal fun DefaultComposerIntegrations(
     val isAttachmentsButtonEnabled = !hasCommandInput && !hasCommandSuggestions && !hasMentionSuggestions
     val isCommandsButtonEnabled = !hasTextInput && !hasAttachments
 
-    val canSendMessage = messageInputState.sendEnabled &&
-        ownCapabilities.contains(ChannelCapabilities.SEND_MESSAGE)
-    val canSendAttachments = messageInputState.sendEnabled &&
-        ownCapabilities.contains(ChannelCapabilities.UPLOAD_FILE)
+    val canSendMessage = canSendMessage(messageInputState)
```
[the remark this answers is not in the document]
VelikovPetar: The `canUploadFile` was accessed only if `canSendMessage == true`, so that logic is the same.(note how `canUploadFile` is called inside if block where `canSendMessage = true`)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -557,7 +557,7 @@ internal fun DefaultComposerIntegrations(
 @Composable
 internal fun DefaultComposerLabel(state: MessageComposerState) {
     val text = with(state) {
```
andremion: I believe there's no need to have that `with` block anymore

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/capabilities/MessageComposerCapabilities.kt
```diff
@@ -0,0 +1,63 @@
...
+ * - In regular mode: requires [ChannelCapabilities.SEND_MESSAGE] capability
+ *
+ * The final decision also considers the [MessageComposerState.sendEnabled] flag, which allows
+ * temporary disabling of sending (e.g., while uploading attachments or validating input).
+ *
+ * @param state The current [MessageComposerState] containing capability and mode information.
+ * @return `true` if the user can send a message or reply based on both capabilities and
+ * the send enabled flag; `false` otherwise.
+ */
+public fun canSendMessage(state: MessageComposerState): Boolean {
```
andremion: Nice to introduce those functions 👍🏻 

What about converting the `state` parameter to receiver, so they could be used, for example, like this: `state.canSendMessage()`

Should those be public from the start?
What about marking them as internal, and then promoting them to public after a while?

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/capabilities/MessageComposerCapabilitiesTest.kt
```diff
@@ -0,0 +1,184 @@
...
+        messageMode: MessageMode,
+        ownCapabilities: Set<String>,
+        expected: Boolean,
+    ) {
+        val state = MessageComposerState(
+            sendEnabled = sendEnabled,
+            messageMode = messageMode,
+            ownCapabilities = ownCapabilities,
+        )
+        Assertions.assertEquals(expected, canSendMessage(state))
```
andremion: nit: import `assertEquals`

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/capabilities/MessageComposerCapabilities.kt
VelikovPetar: Yes, let's make them extensions of `MessageComposerState `, I think it makes sense. 

Concerning the `public` vs `internal`, I initially thought to make them internal, but then I realised that in the [Docs](https://getstream.io/chat/docs/sdk/android/ui/guides/adding-custom-attachments/#sending-date-attachments) for sending custom attachments via the composer , we are also instructing the integrators how to check the send message capabilities. (See: `val canSendMessage = state.ownCapabilities.contains(ChannelCapabilities.SEND_MESSAGE)`, which is now kinda wrong/incomplete). So unless customers have access to these methods, it might be difficult for them to correctly calculate the `canSendMessage` value. So my idea was to expose these methods from the start + [update](https://github.com/GetStream/docs-content/pull/686/files) the docs to use these methods. 

Another option would be to still make these methods internal, but update the docs with notice on how to check the `canSendMessage` state based on channel vs thread. WDYT?

andremion: Fare enough
Let's keep it public then
👍🏻

PR #5979, opened 2025-10-23
Introduce message delivery receipts

andremion:
### 🎯 Goal

- Send message delivery receipts, supporting local data buffering, batched reporting, event handling, and UI indicators.
- Marks messages as delivered on:
  - Single channel query result (first page only)
  - Channel list query result
  - Push notification received
  - WS `message.new` event

https://linear.app/stream/issue/AND-769

#### Client module
- New persistence and repository for delivery receipts (MessageReceiptDao/Entity/Repository, ChatClientDatabase, DateConverter).
- New delivery reporting pipeline (MessageReceiptManager, MessageReceiptReporter) with 1s buffer and max 100 confirmations per call.
- `message.delivered` WS event support.
- Introduced new Channel extensions: `readsOf`, `deliveredReadsOf`.
- Introduced `ChatClient.markMessageAsDelivered` to mark a message as delivered for the current user.
- `PrivacySettings.delivery_receipts` support.
- Channel read state extended with `last_delivered_at` and `last_delivered_message_id`.

#### Delivery receipts are:
- Buffered and sent every 1s, max 100 per request.
- Not sent for own messages, not sent if already read or already delivered.
- Reconciled against reads (reads supersede delivered).

#### Privacy:
- Suppressed if `ownUser.privacySettings.deliveryReceipts.enabled` == false.

#### UI:
- Compose and XML indicators updated to support the new grey double tick (delivered)
- Unified logic to hide the message status indicator when a message is deleted / ephemeral / sync status failed permanently.

### 🎨 UI Changes

**1-to-1 chat**

https://github.com/user-attachments/assets/4eabf3ca-2bc9-40ac-b4f4-8913918ab320

**Group chat**

https://github.com/user-attachments/assets/5634d393-a9e0-40d2-9f1a-6bfbce4d61b0

### 🧪 Testing

- Open the channel list from two devices connected with different users.
- Send a message from user A to user B.
- Long-press on the new message from user A.
- Click on **Message Info**.
- The message should be **delivered**.
- Enter the chat from user B.
- From user A, notice that the message should now be **read**.

### DIFF
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactoryTest.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactoryTest.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) 2014-2025 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.notifications.handler
+
+import android.content.Context
+import androidx.test.core.app.ApplicationProvider
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import org.junit.Test
+import org.junit.jupiter.api.assertInstanceOf
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@RunWith(AndroidJUnit4::class)
+@Config(sdk = [33])
+internal class NotificationHandlerFactoryTest {
+
+    private val context: Context = ApplicationProvider.getApplicationContext()
+    private val notificationConfig = NotificationConfig()
+
+    @Test
+    @Config(sdk = [21])
+    fun `createNotificationHandler should return ChatNotificationHandler`() {
+        val handler = NotificationHandlerFactory.createNotificationHandler(context, notificationConfig)
+
+        assertInstanceOf<ChatNotificationHandler>(handler)
+    }
+
+    @Test
+    fun `createNotificationHandler should return MessagingStyleNotificationHandler since API 23`() {
+        val handler = NotificationHandlerFactory.createNotificationHandler(context, notificationConfig)
+
+        assertInstanceOf<MessagingStyleNotificationHandler>(handler)
+    }
+}


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/MessageDeliveredPlugin.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    override suspend fun onQueryChannelResult(
+        result: Result<Channel>,
+        channelType: String,
+        channelId: String,
+        request: QueryChannelRequest,
+    ) {
+        result.onSuccessSuspend { channel ->
+            if (request.pagination() == null) { // only mark as delivered on initial load
```
xsahil03x: We should also mark as delivered even when the latest message is fetched for the channel in case the query is started from the middle of the channel. (in case of unread or pinned message).

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1388,6 +1400,28 @@ public interface ChatComponentFactory {
...
+    ) {
+        if (params.messageItem.isMessageDelivered) {
+            MessageReadStatusIcon(
+                modifier = params.modifier,
+                message = params.messageItem.message,
+                isMessageRead = params.messageItem.isMessageRead,
+                isMessageDelivered = params.messageItem.isMessageDelivered,
+                readCount = params.messageItem.messageReadBy.size,
+            )
+        } else {
```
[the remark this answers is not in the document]
andremion: This is done to prevent breaking changes in the component factory.
If `isMessageDelivered == true`, that means we can call the new overloaded function; otherwise, we have to call the deprecated one.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChannelExtension.kt
```diff
@@ -117,3 +120,46 @@ public fun Channel.syncUnreadCountWithReads(
...
 ): Channel =
     copy(unreadCount = currentUserUnreadCount(currentUserId))
+
+/**
+ * Returns the user's read state for this channel.
+ *
+ * @param userId The ID of the user whose read state is to be retrieved.
+ * @return The [ChannelUserRead] object representing the user's read state, or null if not found.
+ */
+public fun Channel.userRead(userId: UserId): ChannelUserRead? =
```
VelikovPetar: Just wondering if need to expose these new methods as public - do you think they would be useful for integrators, without the risk of us having to change the implementation/Public API?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChannelExtension.kt
```diff
@@ -117,3 +120,46 @@ public fun Channel.syncUnreadCountWithReads(
...
+ * - The user is not the sender of the message
+ * - The user has read the message
+ *
+ * @param message The [Message] object for which to find read reads.
+ * @return A list of [ChannelUserRead] objects representing users who have read the message
+ */
+public fun Channel.readsOf(message: Message): List<ChannelUserRead> =
+    read.filter { read ->
+        read.user.id != message.user.id &&
+            read.lastRead >= message.getCreatedAtOrThrow()
```
VelikovPetar: I would use the non-throwing method here, like `getCreatedAtOrDefault()`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/MessageDeliveredPlugin.kt
```diff
@@ -0,0 +1,59 @@
...
+import io.getstream.chat.android.client.receipts.MessageReceiptManager
+import io.getstream.chat.android.models.Channel
+import io.getstream.chat.android.models.User
+import io.getstream.result.Result
+import io.getstream.result.onSuccessSuspend
+
+/**
+ * A plugin that marks messages as delivered when channels are queried.
+ */
+internal class MessageDeliveredPlugin(
```
VelikovPetar, line 32: Do you think it would be possible to avoid introducing new `Plugin`s, as it is something that we want to completely remove in the future. I think that this introduces another layer of complexity over the `ChatClient`. Wouldn't it be possible to achieve the same thing by calling the appropriate `MessageReceiptManager` methods from the corresponding `ChatClient` methods (`queryChannels`/`queryChannel`)?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/channels/MessageReadStatusIcon.kt
```diff
@@ -73,26 +76,36 @@ public fun MessageReadStatusIcon(
...
+    isDeliveredIcon: @Composable () -> Unit = { IsDeliveredIcon(modifier = modifier) },
 ) {
     val syncStatus = message.syncStatus
-    when {
-        isMessageRead -> isReadIcon()
 
-        syncStatus == SyncStatus.SYNC_NEEDED ||
-            syncStatus == SyncStatus.AWAITING_ATTACHMENTS -> isPendingIcon()
+    when (syncStatus) {
+        SyncStatus.IN_PROGRESS,
```
VelikovPetar, line 97: This branch now includes the `IN_PROGRESS` status as well - is this intentional?

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/User.kt
```diff
@@ -99,6 +99,11 @@ public data class User(
      */
     val isReadReceiptsEnabled: Boolean get() = privacySettings?.readReceipts?.enabled ?: true
 
+    /**
+     * Determines if the user has delivery receipts enabled.
+     */
+    val isDeliveryReceiptsEnabled: Boolean get() = privacySettings?.deliveryReceipts?.enabled ?: true
```
VelikovPetar, line 105: Should we really fallback to `true` if the `privacySettings/deliveryReceipts` is null?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/utils/ChatEventUtils.kt
```diff
@@ -38,6 +40,18 @@ internal fun MessageReadEvent.toChannelUserRead() = ChannelUserRead(
...
     lastReadMessageId = null,
 )
+
+internal fun MessageDeliveredEvent.toChannelUserRead() = ChannelUserRead(
+    user = user,
+    lastReceivedEventDate = createdAt,
+    lastDeliveredAt = lastDeliveredAt,
+    lastDeliveredMessageId = lastDeliveredMessageId,
+    // The following fields are not applicable for delivered events
+    lastRead = NEVER,
```
VelikovPetar, line 50: Wouldn't this potentially overwrite any relevant `read` status fields? Should we maybe update only the `delivered` fields after receiving the event, but keep the `read` fields intact?

Edit: (I saw the logic in `ChannelMutableState` already covers this, ignore this comment)

CHANGELOG.md
```diff
@@ -49,6 +53,7 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Introduce message delivery receipts. [#5979](https://github.com/GetStream/stream-chat-android/pull/5979)
```
VelikovPetar: Do we need this line in every section? Maybe we can add it in the `## Common changes for all artifacts` section instead?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1388,6 +1400,28 @@ public interface ChatComponentFactory {
         )
     }
 
+    @Composable
+    public fun MessageFooterStatusIndicator(
+        params: MessageFooterStatusIndicatorParams,
+    ) {
+        if (params.messageItem.isMessageDelivered) {
+            MessageReadStatusIcon(
```
VelikovPetar, line 1408: Wouldn't this break any customisations of the old `MessageFooterStatusIndicator`, if a customer enables the `delivery-receipts` feature?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /**
+     * Request to mark the given message as delivered
+     * if delivery receipts are enabled for the current user.
+     *
+     * @see [markChannelsAsDelivered] for the conditions to mark a message as delivered.
+     */
+    suspend fun markMessageAsDelivered(message: Message) {
+        val currentUser = getCurrentUser() ?: run {
```
VelikovPetar: Would this work for the case when push notifications are received, if the app is killed? In that case, we wouldn't have a connected user to check against.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
```diff
@@ -0,0 +1,205 @@
...
+import java.util.Date
+
+/**
+ * Manages message delivery receipts: creating and storing them in the repository
+ * for later reporting to the server.
+ */
+internal class MessageReceiptManager(
+    private val now: () -> Date,
+    private val getCurrentUser: () -> User?,
+    private val repositoryFacade: RepositoryFacade,
```
VelikovPetar: Can we somehow avoid the dependency on the `OfflinePlugin` `RepositoryFacade`?
To be honest I am mostly concerned about the case where the `OfflinePlugin` is not present, and we have to make calls to `getMessage/getChannel`, which if happens often enough, we  could end up in a rate limit error. (Note the case where a customer doesn't have the `OfflinePlugin`, and is in a channel where multiple `message.new` events are received in short amount of time - we could end up firing `getChannel` lots of times)

Can we somehow remove the need for the `retrieveChannel/retrieveMessage` methods in the manager?
I am thinking in the direction of: 
- Instead of depending on the `OfflinePlugin`/`HTTP calls`, why don't we store some channel config in the `client` database after fetching the channel(s). Later, when we call `markMessageAsDelivered`, we can check that stored config, to make sure we should mark the message. 

This whole logic should be easier when the state/db is moved down to the client...

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceipt.kt
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.receipts
+
+import java.util.Date
+
+internal data class MessageReceipt(
+    val messageId: String,
+    val type: String,
```
VelikovPetar: Is the type needed? As far as I can see, if is always set to `delivery`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistence/repository/ChatClientRepository.kt
```diff
@@ -0,0 +1,38 @@
...
+
+package io.getstream.chat.android.client.persistence.repository
+
+import io.getstream.chat.android.client.ChatClient
+import io.getstream.chat.android.client.persistence.db.ChatClientDatabase
+
+/**
+ * Repository that aggregates all internal repositories used by [ChatClient].
+ */
+internal class ChatClientRepository(
```
VelikovPetar, line 25: I am not sure whether we really need this class - wouldn't it just introduce another layer of complexity?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptReporter.kt
```diff
@@ -0,0 +1,83 @@
...
+                    }
+
+                    if (messages.isNotEmpty()) {
+                        logger.d { "Reporting delivery receipts for ${messages.size} messages…" }
+                        api.markDelivered(messages)
+                            .execute()
+                            .onSuccessSuspend {
+                                logger.d { "Successfully reported delivery receipts for ${messages.size} messages" }
+                                val deliveredMessageIds = messages.map(Message::id)
+                                messageReceiptRepository.deleteMessageReceiptsByMessageIds(deliveredMessageIds)
```
VelikovPetar, line 62: I am thinking of a scenario where the `api.markDelivered` would fail multiple times, so the message IDs keep piling up in the DB. Should we maybe introduce some logic that would clear the receipts from the db?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChannelExtension.kt
andremion: I would say so. But I'm fine with making it internal and exposing it on demand.
I will change it then.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/MessageDeliveredPlugin.kt
andremion: I wouldn’t say the long-term goal is to remove plugins entirely, but rather to make them internal and non-optional.
In this case, regardless of the approach, the implementation would likely live outside ChatClient — which is already close to 5K LOC.
Keeping it as an internal plugin for now helps isolate responsibilities and prevents adding more logic directly to ChatClient.
We can revisit and consolidate it later as part of the broader plugin refactoring.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/channels/MessageReadStatusIcon.kt
andremion: Yes, that’s intentional.
Handling all enum branches explicitly ensures that the when statement remains exhaustive.
This way, if new enum values are added in the future, the compiler will prompt updates in all relevant places, preventing silent logic gaps or missing cases.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/User.kt
andremion: Yes. That's the default value for those settings
https://github.com/GetStream/stream-chat-swift/pull/3846/files#diff-084c69e0cb8da2e8ff706fb9f646e50e317e494ca56ab6e74712fed0bfdd6c11R100

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: To enable the delivery-receipts feature, a customer would also need to update the SDK, where the old `MessageFooterStatusIndicator` is now deprecated.
Please let me know if you have any other approach, I'm open to suggestions 😬

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/User.kt
VelikovPetar: Ah got it, I didn't expect it to be different from the general pattern.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChannelExtension.kt
VelikovPetar: I don't really have a very strong opinion here - if you think this would be a stable API, we can leave it exposed.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
VelikovPetar: Not really TBH, but I think that we should document it somewhere, so that integrators know about this. It can be a bit difficult to expect this change, especially if you enable the feature via dashboard flag.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
andremion: I see. In that case, we need to store more than just the channel config. We use message data and channel read, too.
I will include `Message`, `Channel`, `ChannelUserRead`, and `ChannelConfig` in the client database then.
Expects an even bigger PR 😅

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceipt.kt
andremion: The goal is to support other types of message receipt, like `read`, in the future.
The current read implementation is quite confusing and hard to maintain 😓

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistence/repository/ChatClientRepository.kt
andremion: The goal is to avoid having many constructor parameters when a class depends on multiple repositories.
For example, when the manager depends on `MessageRepository`, `ChannelRepository`, and `MessageReceiptRepository`, we would have to pass 3 dependencies instead of 1

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptReporter.kt
andremion: Good point! I will raise this with the feature team. Thanks! 🙌🏻

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/MessageDeliveredPlugin.kt
VelikovPetar: I think that we are reducing not more than 3 lines at best in the `client` by introducing the `Plugin` - two invocations of the `MessageReceiptManager`. I am generally against the introduction of new Plugins, because they are executing crucial business logic as a side-effect, and it's really difficult to locate that logic, unless you explicitly know about the existence of that specific Plugin.

But nevertheless, let's keep it like this, but I definitely agree that should revisit this as part of the planned refactor.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptReporter.kt
andremion: Ticket created to handle this as another task
https://linear.app/stream/issue/AND-895

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
VelikovPetar: Hmm but my question still stands: Isn't hitting rate limit errors a real possibility? 

Side note: We can also access some of the State plugin data from Client via:
```kotlin
logicRegistry
            ?.channelStateLogic("type", "id")
            ?.listenForChannelState()
            ?.toChannel() // ?.getMessageById(id)
```

Perhaps this can be utilised to check for existing channel data (but this still wouldn't cover the case when the client is used without the state plugin)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceipt.kt
VelikovPetar: I am just afraid that this will become legacy code, if we don't actually update this to support the different receipts.

andremion: Let's remove it then

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: Lets consider this on the docs

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChannelExtension.kt
```diff
@@ -117,3 +120,46 @@ public fun Channel.syncUnreadCountWithReads(
...
+ * - The user is not the sender of the message
+ * - The user has received (delivered) the message
+ *
+ * @param message The [Message] object for which to find delivered reads.
+ * @return A list of [ChannelUserRead] objects representing users who have delivered the message.
+ */
+public fun Channel.deliveredReadsOf(message: Message): List<ChannelUserRead> =
+    read.filter { read ->
+        read.user.id != message.user.id &&
+            (read.lastDeliveredAt ?: NEVER) >= message.getCreatedAtOrDefault(NEVER)
```
xsahil03x, line 164: You should probably want to also check for `lastRead` and promote it to `delivered` for supporting old messages where `lastDeliveredAt` is not present.

andremion: The `Channel.deliveredReadsOf` function returns a list of `ChannelUserRead` objects representing which ones have delivered the given message, regardless of the read state. If `lastDeliveredAt` is null, as in old messages, it is not returned in the list.
For reads, we can use `Channel.readsOf` function, which checks specifically read states, regardless of the delivered states.
That's to give customers the flexibility to list who has read messages and who has delivered messages, exclusively.
If they want to list delivered reads that are not read, they can do:
`channel.deliveredReadsOf(message) - channel.readsOf(message)`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
andremion: The channel state is not completely filled when the app is not in memory (for example, the read list is empty). So we can't rely on it.
We can make the delivery receipts feature require the offline plugin, so we can rely on the local storage instead (local storage will be available soon as a built-in SDK feature).

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -642,6 +650,7 @@ internal constructor(
         tokenManager.setTokenProvider(tokenProvider)
         appSettingsManager.loadAppSettings()
         warmUp()
+        messageReceiptReporter.start()
```
VelikovPetar, line 653: Should we also have a corresponding `stop()` method? Or do we rely on cancelling the scope to cancel the running job?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/ChatNotifications.kt
```diff
@@ -38,7 +38,12 @@ internal interface ChatNotifications {
     fun onSetUser(user: User)
     fun setDevice(device: Device)
     suspend fun deleteDevice()
-    fun onPushMessage(message: PushMessage, pushNotificationReceivedListener: PushNotificationReceivedListener)
+    fun onPushMessage(
+        pushMessage: PushMessage,
+        pushNotificationReceivedListener: PushNotificationReceivedListener =
+            PushNotificationReceivedListener { _, _ -> },
```
VelikovPetar, line 44: I believe that the default PushNotificationReceivedListener is no longer needed.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactory.kt
```diff
@@ -75,6 +78,7 @@ public object NotificationHandlerFactory {
             getDefaultActionsProvider(context),
         notificationBuilderTransformer: (NotificationCompat.Builder, ChatNotification) -> NotificationCompat.Builder =
             { builder, _ -> builder },
+        onNewMessage: (pushMessage: PushMessage) -> Boolean = { false },
```
VelikovPetar: I am not sure whether this should also control the default showing of push notifications, because it is already configurable by extending the `NotificationHandler`. Should this maybe just be a side effect? 

Also I believe that this method will be called for any PN, not just for `message.new`, so for the delivery receipt case, the integrators would also need to check the type of the message.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistence/db/ChatClientDatabase.kt
```diff
@@ -0,0 +1,56 @@
...
+    DateConverter::class,
+)
+internal abstract class ChatClientDatabase : RoomDatabase() {
+    abstract fun messageReceiptDao(): MessageReceiptDao
+
+    companion object {
+        fun build(context: Context) = Room.databaseBuilder(
+            context = context.applicationContext,
+            klass = ChatClientDatabase::class.java,
+            name = "stream_chat_client.db",
```
VelikovPetar, line 44: Do we need the `.db` suffix in the name of the DB?

Also another thing that I missed initially: Should we have different DB files per user (similar to the OfflinePlugin DB)?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistence/db/converter/DateConverter.kt
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.persistence.db.converter
+
+import androidx.room.TypeConverter
+import java.util.Date
+
+internal class DateConverter {
```
VelikovPetar, line 22: Small point for the future, no need to change anything now: We could potentially unify the common converters with the ones from the offline plugin.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatHelper.kt
```diff
@@ -69,6 +69,13 @@ object ChatHelper {
                     parentMessageId = message.parentId,
                 )
             },
+            onNewMessage = { pushMessage ->
```
VelikovPetar: As I mentioned in a previous comment, I believe that this will be called for every PN type -> shouldn't we do this just for `message.new`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2870,6 +2883,33 @@ internal constructor(
...
+     * - Not yet marked as read by the current user.
+     * - Not yet marked as delivered by the current user.
+     *
+     * @param messageId The ID of the message to mark as delivered.
+     */
+    @CheckResult
+    public fun markMessageAsDelivered(messageId: String): Call<Unit> =
+        CoroutineCall(userScope) {
+            messageReceiptManager.markMessageAsDelivered(messageId)
+            Result.Success(Unit)
```
VelikovPetar: Should we maybe return `Result.Error` from this method, if the message cannot be marked as delivered (ex. if some of the preconditions are not satisfied)?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -2870,6 +2883,33 @@ internal constructor(
             }
     }
 
+    /**
+     * Request to mark the message with the given id as delivered if:
```
VelikovPetar, line 2886: I think it would be good to document that this method might attempt to internally call `getMessage/getChannel`, just for transparency sake. Or at least it would be good to document this in the Docs page as well. I think that it is an important information, that integrators should know before using this method.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
VelikovPetar: No longer relevant, PNs will be handled by customer integrations.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
VelikovPetar: Already resolved

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: For simplicity, no need for any additional function as the scope cancels its children on user disconnect

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactory.kt
andremion: The idea was to make it simpler for integrators who don't have a custom notification handler and want to mark a new push message as delivered.
What would you suggest?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistence/db/ChatClientDatabase.kt
andremion: I don't see reasons to have an internal DB file per user. Why would it be needed?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistence/db/converter/DateConverter.kt
andremion: Sure. That's just because we have different modules and offline depends on the client, not the other way around.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatHelper.kt
andremion: What are the other types that are triggered as part of `onPushMessage` with a `PushMessage` rather than on new messages?
That logic remains the same as the previous PR pushes 🤔

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: I used the same convention of `public fun clearPersistence(): Call<Unit>`
I will have to change the internals of markMessageAsDelivered in that case...

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
andremion: BTW, we will mention that this feature requires the offline plugin to work properly and efficiently.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistence/db/ChatClientDatabase.kt
VelikovPetar: Ah I see that we always clear the DB upon user logout - then we should be fine. I wrongly assumed that the clearing of this DB is optional (like on the OfflinePlugin), so I was worried that we would end up in a case where the DB could store message IDs from different users. 
Closing the remark.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatHelper.kt
VelikovPetar: We can get the new push v3 types: `notification.reminder_due`, `reaction.new` or `message.updated`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: I was thinking if it would be more transparent - but we could maybe introduce this additionally if it is not initially required.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: As long as it is transparent, I think it would be fine. (I am mostly concerned about customer reports that we call `getMessage/getChannel` when it wasn't expected, so it would be good to be transparent about the side-effects of this method).

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/ChatNotifications.kt
andremion: Actually, we need. Otherwise, we have to pass it even if we don't want it
Screenshot 2025-11-13 at 09 58 59

VelikovPetar: Ah ok, lets keep it then

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatHelper.kt
andremion: Good point!
I will handle it
Thanks!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/notifications/handler/NotificationHandlerFactory.kt
VelikovPetar: Ok yeah, that makes sense, this will not be possible without a fully custom `NotificationHandler`, lets keep it like this then.

VelikovPetar (approved):

PR #5980, opened 2025-10-23
Migrate MediaPlayer to ExoPlayer

VelikovPetar:
### 🎯 Goal
The current implementation of the `AudioPlayer` is based on `android.media.MediaPlayer`. While this was working well in most cases, we had occurrences where a voice recoding could not be played (to be more specific: `mime-type = "audio/mp4;codecs=opus"`). By migrating to ExoPlayer we are able to play such files (+ potentially to support a wider range of codecs)

### 🛠 Implementation details
- Adds `exoplayer` dependency to `stream-chat-android-client`
- Replaces the usages of the `MediaPlayer` with `ExoPlayer` (inside the `NativeMediaPlayerImpl` wrapper)

### 🎨 UI Changes
_NA_

### 🧪 Testing
Test ALL features related to voice recording:
1. Recording a new voice message (XML and Compose)
2. Listen (with skipping, pausing and resuming) the recording before sending it to the channel
3. Listen to already sent voice messages (with skipping, pausing and resuming) - both own and others
4. Switch between playing different voice messages
5. Send voice messages from iOS/React, and verify that they are playable on Android too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,28 @@
 
 ### ❌ Removed
 
+# November 11th, 2025 - 6.27.0
+## stream-chat-android-offline
+### ⬆️ Improved
+- Fallback to empty `User` instead of crashing if the searched user is not yet inserted into the DB. [#5998](https://github.com/GetStream/stream-chat-android/pull/5998)
+
+## stream-chat-android-state
+### 🐞 Fixed
+- Fix jump to message overwriting the initially loaded messages in the channel. [#6000](https://github.com/GetStream/stream-chat-android/pull/6000)
+
+## stream-chat-android-ui-common
+### ✅ Added
+- Add `MessageListController.hideUnreadSeparator()` method to manually hide the unread separator. [#6001](https://github.com/GetStream/stream-chat-android/pull/6001)
+
+## stream-chat-android-ui-components
+### ✅ Added
+- Add `MessageListViewModel.hideUnreadSeparator()` method to manually hide the unread separator. [#6001](https://github.com/GetStream/stream-chat-android/pull/6001)
+
+## stream-chat-android-compose
+### ✅ Added
+- Add `MessageListViewModel.hideUnreadSeparator()` method to manually hide the unread separator. [#6001](https://github.com/GetStream/stream-chat-android/pull/6001)
+- Add `MentionStyleFactory` for customizing mentions in the message composer. [#5984](https://github.com/GetStream/stream-chat-android/pull/5984)
+
 # November 07th, 2025 - 6.26.1
 ## stream-chat-android-state
 ### 🐞 Fixed
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 
 ![stream-chat-android-client](https://img.shields.io/badge/stream--chat--android--client-3.22%20MB-lightgreen)
 ![stream-chat-android-offline](https://img.shields.io/badge/stream--chat--android--offline-3.45%20MB-lightgreen)
-![stream-chat-android-ui-components](https://img.shields.io/badge/stream--chat--android--ui--components-10.53%20MB-lightgreen)
+![stream-chat-android-ui-components](https://img.shields.io/badge/stream--chat--android--ui--components-10.54%20MB-lightgreen)
 ![stream-chat-android-compose](https://img.shields.io/badge/stream--chat--android--compose-12.76%20MB-lightgreen)
 
 </div>
--- a/buildSrc/src/main/kotlin/io/getstream/chat/android/Configuration.kt
+++ b/buildSrc/src/main/kotlin/io/getstream/chat/android/Configuration.kt
@@ -2,8 +2,8 @@ package io.getstream.chat.android
 
 object Configuration {
     const val majorVersion = 6
-    const val minorVersion = 26
-    const val patchVersion = 1
+    const val minorVersion = 27
+    const val patchVersion = 0
     const val versionName = "$majorVersion.$minorVersion.$patchVersion"
     const val snapshotVersionName = "$majorVersion.$minorVersion.${patchVersion + 1}-SNAPSHOT"
     const val artifactGroup = "io.getstream"
--- a/metrics/size.json
+++ b/metrics/size.json
@@ -3,12 +3,12 @@
     "stream-chat-android-client": 4388,
     "stream-chat-android-offline": 4728,
     "stream-chat-android-ui-components": 13284,
-    "stream-chat-android-compose": 16520
+    "stream-chat-android-compose": 16524
   },
   "release": {
     "stream-chat-android-client": 3304,
     "stream-chat-android-offline": 3536,
-    "stream-chat-android-ui-components": 10792,
-    "stream-chat-android-compose": 13068
+    "stream-chat-android-ui-components": 10796,
+    "stream-chat-android-compose": 13076
   }
 }
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/RepositoryFacade.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/persistance/repository/RepositoryFacade.kt
@@ -239,9 +239,9 @@ public class RepositoryFacade private constructor(
         ): RepositoryFacade {
             val userRepository = factory.createUserRepository()
             val getUser: suspend (userId: String) -> User = { userId ->
-                requireNotNull(userRepository.selectUser(userId)) {
-                    "User with the userId: `$userId` has not been found"
-                }
+                // Fallback to an empty user if it cannot be found in the DB (in case we try to fetch the user before
+                // it was inserted).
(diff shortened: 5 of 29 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/audio/NativeMediaPlayer.kt
```diff
@@ -343,141 +373,175 @@ internal class NativeMediaPlayerImpl(
...
+    private val exoPlayer: ExoPlayer
+        get() {
+            return _exoPlayer ?: builder().also {
+                _exoPlayer = it.setupListeners()
+                state = NativeMediaPlayerState.IDLE
+            }
         }
-    }
+
+    @SuppressLint("UnsafeOptInUsageError")
```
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5980/commits/7ea6b8bc03b6c9123b8c5d877009f86f31186420

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/audio/NativeMediaPlayer.kt
```diff
@@ -16,149 +16,186 @@
...
+import androidx.media3.exoplayer.ExoPlayer
+import androidx.media3.exoplayer.source.MediaSource
+import androidx.media3.exoplayer.source.ProgressiveMediaSource
+import androidx.media3.extractor.DefaultExtractorsFactory
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import io.getstream.log.taggedLogger
 import java.io.IOException
 
+/**
+ * Defines the contract for communicating with a native media player implementation (ex. ExoPayer).
```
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/5980/commits/7ea6b8bc03b6c9123b8c5d877009f86f31186420

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/AudioPlayerController.kt
```diff
@@ -205,11 +205,11 @@ public class AudioPlayerController(
         val initialState = curState.newCurrentState(audioHash, recordingUri, waveform, playbackInMs, durationInMs)
         setState(initialState)
 
-        if (seekTo > 0) audioPlayer.seekTo(playbackInMs, audioHash)
         audioPlayer.registerOnAudioStateChange(audioHash, this::onAudioStateChanged)
         audioPlayer.registerOnProgressStateChange(audioHash, this::onAudioPlayingProgress)
         audioPlayer.registerOnSpeedChange(audioHash, this::onAudioPlayingSpeed)
```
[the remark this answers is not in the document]
VelikovPetar: I will revert this, it is not needed

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/audio/NativeMediaPlayer.kt
```diff
@@ -343,64 +373,98 @@ internal class NativeMediaPlayerImpl(
...
+
+    private val mediaSourceFactory: MediaSource.Factory = ProgressiveMediaSource.Factory(
+        DefaultDataSource.Factory(context),
+        DefaultExtractorsFactory().setConstantBitrateSeekingEnabled(true),
+    )
 
     private var onCompletionListener: (() -> Unit)? = null
-    private var onErrorListener: ((what: Int, extra: Int) -> Boolean)? = null
+    private var onErrorListener: ((errorCode: Int) -> Boolean)? = null
     private var onPreparedListener: (() -> Unit)? = null
```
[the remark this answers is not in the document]
VelikovPetar: removed, it was no longer used

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/audio/NativeMediaPlayer.kt
```diff
@@ -308,7 +330,7 @@ public interface NativeMediaPlayer {
      * Returning false, or not having an OnErrorListener at all, will
      * cause the OnCompletionListener to be called.
      */
-    public fun setOnErrorListener(listener: (what: Int, extra: Int) -> Boolean)
+    public fun setOnErrorListener(listener: (errorCode: Int) -> Boolean)
```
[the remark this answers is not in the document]
VelikovPetar: this is not correct

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -4709,17 +4708,18 @@ internal constructor(
...
-                mediaPlayer = NativeMediaPlayerImpl {
-                    MediaPlayer().apply {
-                        AudioAttributes.Builder()
-                            .setContentType(AudioAttributes.CONTENT_TYPE_MUSIC)
-                            .build()
-                            .let(this::setAudioAttributes)
-                    }
+            val audioPlayer: AudioPlayer = StreamAudioPlayer(
+                mediaPlayer = NativeMediaPlayerImpl(appContext) {
+                    ExoPlayer.Builder(appContext)
```
gpunto, line 4723: I'm a bit worried about the size bump of the SDK. Some companies pay attention to that.

Can we do it in a way to check if the dependency exists at runtime and decide whether to use ExoPlayer based on that?

I'm thinking, maybe we can add ExoPlayer as `compileOnly` instead of `implementation` and then check if the class exists like:

```kotlin
private boolean isExoPlayerAvailable() {
    try {
        Class.forName("androidx.media3.exoplayer.ExoPlayer");
        return true;
    } catch (ClassNotFoundException e) {
        return false;
    }
}
```

aleksandar-apostolov: and if it does not? Fallback to `MediaPlayer`?

aleksandar-apostolov: I think it adds unnecessary complexity to already complex chat. I think we need to live with the 1.9MBs. 

@VelikovPetar why has offline increased for 2MB?

VelikovPetar: The `offline` module includes the `client` as a dependency, so its size has increased transitively. 
The `ui` modules are not affected, because they already include the `exoplayer` dependency (for video playback). 

Another (a bit more complex) solution would be to move the `AudioPlayer` up to the `ui` modules, where we already have the `exoplayer` library. Currently the `ChatClient` only accesses the `AudioPlayer` after a logout for cleanup, but with some effort we can ensure proper clean-up in the components where the `AudioPlayer` is used. The `AudioPlayer` API is already internal (with the exception of some now unused constants) so there won't be any breaking changes. 
But such approach of course would introduce bigger complexity in the wiring of the player, because now can just access it as `chatClient.audioPlayer`, if we move it upwards, we would need to handle its instantiation better.

aleksandar-apostolov: The correct way to avoid the 2MB increase is I think to have the contract live in chat and default `MediaPlayer` implementation.
Then as a separate module include the `ExoPlayer` backed implementation of said contract. You can then supply an external `StreamMediaPlayer` implementation of whichever player you choose. 

However this goes against our goal of reducing integration complexity and dependencies. I think including exo player is not an optional dependency when `MediaPlayer` is buggy. We've had several issues with this already. And in the end we are just going to tell people to include the dependency anyway and they will still suffer the 2MB size increase.

PratimMallick: hey @VelikovPetar - Maybe try this 

Instead of includign the entire exoplayer dep `implementation(libs.androidx.media3.exoplayer)` 

We can include its modules that we need as of now 


media3-exoplayer-core -> this one is always required 

Apart from this, we only need the type of streaming we are supporting :

The other modules are : 

media3-exoplayer-dash ->  For DASH (MPEG-DASH) streaming
media3-exoplayer-hls -> For HLS streaming
media3-exoplayer-smoothstreaming -> For SmoothStreaming
media3-exoplayer-rtsp -> For RTSP playback
media3-exoplayer-transformer -> For media editing/transcoding
media3-ui -> For default UI components (optional)

VelikovPetar: We checked this, but unfortunately the legacy `exoplayer-core` has now become the new `media3-exoplayer`(and there is no `media3-exoplayer-core`), so we are already using the minimal required dependencies for the `exoplayer`.

aleksandar-apostolov (approved):

PR #5986, opened 2025-10-30
Apply common Stream convention plugins

gpunto:
### 🎯 Goal

Part of [AND-864](https://linear.app/stream/issue/AND-864/introduce-plugin-to-replace-base-android-modules-setup)

We now have started publishing Stream Gradle plugins, so I'm doing the initial integration here. Atm they don't do a lot (they just configure the Java version), but I'm building upon this first version to move more and more shared configuration to them.

### 🛠 Implementation details

Replace custom `android.gradle` with convention plugins

### 🎨 UI Changes

None

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF


build.gradle.kts
```diff
@@ -52,10 +57,31 @@ apply(from = "${rootDir}/scripts/sonar.gradle")
 apply(from = "${rootDir}/scripts/coverage.gradle")
 
 subprojects {
-    if (name != "stream-chat-android-docs"
-            && buildFile.exists()) {
+    if (name != "stream-chat-android-docs" && buildFile.exists()) {
         apply(from = "${rootDir}/spotless/spotless.gradle")
     }
+
+    // Configure Android projects with common SDK versions as soon as either plugin is applied
```
gpunto, line 64: This was done through `android.gradle`

gpunto: We can consider moving this in the plugins too if we prefer. I can add it as a root plugin configuration and then each module plugins reads from it and applies it when configuring the app/library Android plugin (to double check if it's doable, but it should).

VelikovPetar: I am not sure if we would gain much by moving it to the plugins, if we still need to do some manual configuration in each SDK (for example if we have different `minSdk` for different products). 
I would personally be OK with whatever approach we take.

VelikovPetar (approved):

PR #5994, opened 2025-11-05
Fix unread badge not disappearing upon opening the chat

andremion:
### 🎯 Goal

Fix `markChannelAsRead` not working when `lastReadMessageId` is correct but `unreadCount` > 0

### 🛠 Implementation details

The `markChannelAsRead` function is now allowed to proceed if there are unread messages, even if the `lastReadMessageId` already matches the ID of the latest message. This handles cases where the server state for the unread count might be out of sync with the client.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/3823d8d3-6b01-45a6-ab64-2f10fcde5d0d" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/fa5b0448-3d2b-485a-9fa8-b6fcfdf8122a" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

Open a channel that contains a single message, and the `lastReadMessageId` is the same as `message.id` but `unreadMessages` is > 0

### DIFF
--- a/stream-chat-android-state/src/test/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableStateTests.kt
+++ b/stream-chat-android-state/src/test/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableStateTests.kt
@@ -583,8 +583,8 @@ internal class ChannelMutableStateTests {
         val actual = channelState.markChannelAsRead()
 
         assertEquals(true, actual)
-        assertEquals(lastMessage.createdAt, channelState.read.value?.lastReceivedEventDate)
-        assertEquals(lastMessage.createdAt, channelState.read.value?.lastRead)
+        assertEquals(lastMessage.createdLocallyAt, channelState.read.value?.lastReceivedEventDate)
+        assertEquals(lastMessage.createdLocallyAt, channelState.read.value?.lastRead)
         assertEquals(0, channelState.read.value?.unreadMessages)
     }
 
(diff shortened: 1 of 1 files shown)


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
```diff
@@ -572,9 +572,12 @@ internal class ChannelMutableState(
             when (val currentUserRead = read.value) {
                 null -> true
                 else ->
+                    // Allow marking as read if:
+                    // 1. The last read message ID differs from the last message, OR
+                    // 2. There are unread messages (even if lastReadMessageId matches, server state may differ)
                     currentUserRead
-                        .takeIf { it.lastReadMessageId != lastMessage.id }
-                        ?. let {
+                        .takeIf { it.lastReadMessageId != lastMessage.id || it.unreadMessages > 0 }
```
VelikovPetar, line 580: Just curious how can we end up in such state, where `it.lastReadMessageId == lastMessage.id` and `it.unreadMessages > 0`?

andremion: To be honest, I don't know. I could only reproduce this using the customer app ID.

VelikovPetar (approved):

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/state/channel/internal/ChannelMutableState.kt
VelikovPetar: Let's keep this fix - if this turns out to be a back-end issue, we could revisit if we need to revert it.

PR #6002, opened 2025-11-12
Use unique cache directory for coil image caching

VelikovPetar:
### 🎯 Goal
There is a possibility if a customer is using `coil2` while using the Chat SDK (which uses `coil3`) for a race condition: Both `coil2` and `coil3` use the same cache directory (`image_cache`) for storing image data, however they have slightly different output format. So if one version of coil writes to this cache, but is read by the different one, the parsing will fail and image will fail to load.

### 🛠 Implementation details
Use a Stream specific cache dir for the `coil3` instance to avoid potential clash with the default (`image_cache`) from the customer coil instance: `stream_coil3_image_cache`.

### 🎨 UI Changes
_NA_

### 🧪 Testing
It is a bit difficult to reproduce as it is caused by a race condition. It most often happens when the same image (from same URL resource) is rendered both via `coil2` and `coil3`. The result is that one of the image views will fail to render.

To test the more common scenario, ensure that none of the current features which include images are not broken. Ex. media attachments, media gallery, channel media attachments gallery...

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/StreamImageLoaderFactory.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/StreamImageLoaderFactory.kt
@@ -35,7 +35,7 @@ import okio.Path.Companion.toOkioPath
 
 private const val DEFAULT_MEMORY_PERCENTAGE = 0.25
 private const val DEFAULT_DISK_CACHE_PERCENTAGE = 0.02
-private const val DISK_CACHE_DIRECTORY = "stream_coil3_image_cache"
+private const val DISK_CACHE_DIRECTORY = "stream_image_cache"
 
 public class StreamImageLoaderFactory(
     private val builder: ImageLoader.Builder.() -> Unit = {},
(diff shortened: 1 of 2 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/StreamImageLoaderFactory.kt
```diff
@@ -35,7 +35,7 @@ import okio.Path.Companion.toOkioPath
 
 private const val DEFAULT_MEMORY_PERCENTAGE = 0.25
 private const val DEFAULT_DISK_CACHE_PERCENTAGE = 0.02
-private const val DISK_CACHE_DIRECTORY = "image_cache"
+private const val DISK_CACHE_DIRECTORY = "stream_coil3_image_cache"
```
andremion: What about `stream_image_cache`?
I believe no need to use `coil3`

VelikovPetar: Yeah, I will update it.

andremion (approved):

PR #6016, opened 2025-11-24
Prevent duplicate `GetOrCreateChannel` calls when paginating messages

VelikovPetar:
### 🎯 Goal
Resolves two issues which can result in multiple unnecessary calls to `GetOrCreateChannel` when paginating messages in a channel:
1. Don't call `loadOlderMessages` if `MessageListController` is already loading older messages
2. Don't call `fillTheGap` if `ChannelLogic` is loading older/newer messages. In this case, there will be no gap to be filled, so any calls to it are redundant. 

### 🛠 Implementation details
1. Add guard for `loadingOlderMessages` in `MessageListController.loadOlderMessages`
2. Call `ChannelLogic.fillTheGap` ONLY when loading messages around ID (`isFilteringAroundIdMessages`)

### 🎨 UI Changes
_NA_

### 🧪 Testing
1. Open a channel
2. Scroll backwards (to load older messages)
3. Each page load should trigger ONLY 1 `GetOrCreateChannel` call (`POST https://chat.stream-io-api.com/channels/<channelType>/<channelId>/query`)
4. Observe the logs for no duplicate calls when paginating


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -1210,15 +1210,14 @@ public class MessageListController(
...
-                is MessageMode.Normal -> {
-                    if (channelState.value?.endOfOlderMessages?.value == true) return
-                    chatClient.loadOlderMessages(cid, messageLimit).enqueue()
-                }
-
-                is MessageMode.MessageThread -> threadLoadMore(this)
+        when (val mode = _mode.value) {
+            is MessageMode.Normal -> {
+                val endOfOlderMessages = channelState.value?.endOfOlderMessages?.value == true
+                val loadingOlderMessages = channelState.value?.loadingOlderMessages?.value == true
```
andremion, line 1216: Since we now have that guard, is that possible that the pagination gets stuck if we get an error 
(`onQueryChannelResult` error flow), and `loadingOlderMessages` doesn't get to be reset as we do on `updateDataForChannel`.

Should we also reset those states on error?

VelikovPetar: Good point, let me revisit this scenario.

VelikovPetar: I will introduce the following change in `ChannelStateLogic#propagateQueryError`:
```kotlin
mutableState.setLoadingOlderMessages(false)
mutableState.setLoadingNewerMessages(false)
```
I think this changes makes sense anyway, it is strange that we weren't resetting this flag even before this PR.

WDYT?

andremion: Yes, I think that should fix in the case of error 👍🏻

VelikovPetar: Done: https://github.com/GetStream/stream-chat-android/pull/6016/commits/21a1403dbbc05f8c5115ec2fb963e18bfad975c8

andremion (approved):

PR #6017, opened 2025-11-24
Add `isAutomaticSyncOnReconnectEnabled` for configuring the automatic sync process on connect/reconnect

VelikovPetar:
### 🎯 Goal
Allow customers to opt-out of the automatic `sync` process when connecting/reconnecting. The drawback is that they will need to manually `query/watch` the relevant `Channel`s.
Docs PR: https://github.com/GetStream/docs-content/pull/788

### 🛠 Implementation details
1. Add `StatePluginConfig.isAutomaticSyncOnReconnectEnabled` flag
2. Skip calling `/sync` endpoint and `QueryChannels`/`WatchChannel` from the `SyncManager` if `StatePluginConfig.isAutomaticSyncOnReconnectEnabled = false`

### 🎨 UI Changes
_NA_

### 🧪 Testing
1. Set `StatePluginConfig.isAutomaticSyncOnReconnectEnabled = false`
2. Put app in background
3. Restore app
4. No `/sync` or `/channels` calls should be made upon restoring the app


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatHelper.kt
```diff
@@ -86,7 +86,7 @@ object ChatHelper {
 
         val statePluginFactory = StreamStatePluginFactory(
             config = StatePluginConfig(
-                backgroundSyncEnabled = true,
+                backgroundSyncEnabled = false,
```
aleksandar-apostolov, line 89: We disable sync on the Sample, this is not accidental right? Same on other places in other samples.

VelikovPetar: It was actually intentional (because this flag is now deprecated), and since we recommend disabling it, I decided to also disable it in the samples.

PR #6023, opened 2025-11-27
Prevent potential duplicate calls to `CreateDevice`

VelikovPetar:
### 🎯 Goal
We still have reports of the `CreateDevice` endpoint hitting rate limits. In this PR we introduce another level of guarding against duplicate calls, by introducing a local tracking of the device registered in the current session, plus a locking mechanism.

### 🛠 Implementation details
- Tie the device registration process to the currently logged in user (note: the `CreateDevice`/`DeleteDevice` endpoints will fail if called without a connected user) to support the case where different users a switched in the same `ChatClient`
- Introduce an additional check in the `PushTokenUpdateHandler#addDevice`: check if the same token was registered previously in the current session (in addition to the check agains `user.devices`)
- Add a locking mechanism to prevent multiple `CreateDevice` invocations at the same time 
- Additionally do some small clean-up in the `DeviceApi` and `ChatApi` by removing unnecessary parameters

### 🎨 UI Changes
_NA_

### 🧪 Testing
Pretty difficult to test in a real-world scenario. One possible scenario to test is to manually delete the firebase token:
```kotlin
FirebaseMessaging.getInstance().deleteToken()
```
Then put the app in background, and then back to foreground. There should be two attempts to register the new token:
1. via `onNewToken`
2. when user is reconnected

But only one `CreateDevice` should be fired. 

Another possibility is to simulate repeated calls to `FirebaseMessagingDelegate.registerPushDevice` - again, only one `CreateDevice` should be fired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 
 ## stream-chat-android-state
 ### 🐞 Fixed
+- Prevent duplicate `GetOrCreateChannel` calls when paginating messages. [#6016](https://github.com/GetStream/stream-chat-android/pull/6016)
 
 ### ⬆️ Improved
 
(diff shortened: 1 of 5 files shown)


andremion (commented):
I can't reproduce the issue even on develop

I deleted the token, then background, then foreground, but found one request `POST https://chat.stream-io-api.com/devices`

PR #6029, opened 2025-12-01
Fix `StrictMode` violations in the `AttachmentsPicker`

VelikovPetar:
### 🎯 Goal
Implements fixes for StrictMode violations (Read/Write disk on main thread) originating from the attachments picker.
Resolves: https://linear.app/stream/issue/AND-938/fix-strictmode-violations-in-attachmentspicker

### 🛠 Implementation details
- Introduce new `AttachmentsProcessingViewModel` which defers the `StorageHelperWrapper` operations to a background thread.
- Use the new view model in `AttachmentsPickerFilesTabFactory`, `AttachmentsPickerImagesTabFactory` and in `AttachmentsPickerSystemTabFactory`
- Add similar async operations in the `AttachmentsPickerViewModel`
- Rework the `AttachmentsPicker` composable to use the new asynchronous methods

### 🎨 UI Changes
_NA_

### 🧪 Testing
1. Open attachment picker
2. Select from the `Images` or `Files` tab (or do the same with `ChatTheme(useDefaultSystemMediaPicker = true)`)
3. There should be no violations logged in the LogCat

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/factory/AttachmentsProcessingViewModel.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/factory/AttachmentsProcessingViewModel.kt
@@ -49,51 +49,27 @@ internal class AttachmentsProcessingViewModel(
 
     /**
      * Flow of events emitted when attachments metadata is retrieved from URIs.
-     *
-     * This [SharedFlow] emits [AttachmentsMetadataFromUris] events that contain both the original URIs
-     * and the retrieved [AttachmentMetaData]. The UI can collect from this flow to react to metadata
-     * retrieval and update the attachment state accordingly.
      */
     val attachmentsMetadataFromUris: SharedFlow<AttachmentsMetadataFromUris> =
         _attachmentsMetadataFromUris.asSharedFlow()
 
     /**
      * Flow of events emitted when files metadata is retrieved.
-     *
-     * This [SharedFlow] emits lists of [AttachmentMetaData] representing the files retrieved from storage.
-     * The UI can collect from this flow to react to file metadata retrieval and update the attachment state
-     * accordingly.
      */
     val filesMetadata: SharedFlow<List<AttachmentMetaData>> =
         _filesMetadata.asSharedFlow()
 
     /**
      * Flow of events emitted when media metadata is retrieved.
-     *
-     * This [SharedFlow] emits lists of [AttachmentMetaData] representing the media retrieved from storage.
-     * The UI can collect from this flow to react to media metadata retrieval and update the attachment state
-     * accordingly.
      */
     val mediaMetadata: SharedFlow<List<AttachmentMetaData>> =
         _mediaMetadata.asSharedFlow()
 
     /**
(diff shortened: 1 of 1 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/factory/AttachmentsProcessingViewModel.kt
```diff
@@ -0,0 +1,249 @@
...
+import androidx.lifecycle.viewModelScope
+import io.getstream.chat.android.compose.ui.util.StorageHelperWrapper
+import io.getstream.chat.android.core.internal.coroutines.DispatcherProvider
+import io.getstream.chat.android.ui.common.state.messages.composer.AttachmentMetaData
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.launch
+
+/**
```
andremion, line 31: I would recommend avoiding long KDoc blocks like this, especially for internal components. If the explanation needs this much detail, it’s usually a signal that the code is either hard to read or overly complex. In those cases, it’s preferable to simplify or restructure the code so its behavior is clearer without requiring extensive documentation.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/AttachmentsPickerViewModel.kt
andremion, line 123: What should happen with that one?
Should we deprecate it (as it is not used by the SDK anyway)?

VelikovPetar: Interesting, I didn't realise that this wasn't used. I will deprecate it, let's use this opportunity to eventually clean this up.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/factory/AttachmentsProcessingViewModel.kt
VelikovPetar: I went with a more detailed KDocs here, because as you mentioned it is not very straight-forward to understand what this view model does. But I don't think that there is another way for us to achieve this fix (without any breaking changes). I can of course simplify the docs here, or if you have any suggestions for simplifying the logic contained in this VM.

andremion: Looking at the implementation itself, it’s straightforward to understand. The extensive documentation actually made me question the complexity more than the code did.
This suggests that most of the added docs are unnecessary?
WDYT?

VelikovPetar: Ok I will simplify the docs

PR #6036, opened 2025-12-08
Fix `StrictMode` violations in `AudioRecordingController`

VelikovPetar:
### 🎯 Goal
Fixes several `StrictMode` violations from the `AudioRecordingController` related to accessing the disk on main thread.

### 🛠 Implementation details
- Make `AudioRecordingController` `startRecording`, `stopRecording` and `completeRecording` `suspend`
- `startRecording`: wrap `mediaRecorder.startAudioRecording` with `withContext(DispatcherProvider.IO)` to ensure it is run on IO
- `stopRecording`: wrap `mediaRecorder.stopRecording()` with `withContext(DispatcherProvider.IO)` to ensure it is run on IO
- `completeRecording`: wrap `mediaRecorder.stopRecording()` with `withContext(DispatcherProvider.IO)` to ensure it is run on IO and post state updates on `Main` to ensure they are processed in order
- `completeRecordingSync`: Introduce a new method which returns the `Result` of the recording instead of updating the internal state. This was required to ensure the `MessageComposerController.sendRecording()` works properly, because it previously relied on the sequential execution of  `completeRecording` (which is no longer the case)
- call the changed method in a coroutine from the `MessageComposerController`
**Note: The corresponding methods from `StreamMediaRecorder` were not changed to `suspend` because it is public API.**

### 🎨 UI Changes
Run the following tests, ensuring that nothing is broken + there are no related `StrictMode` violations in the LogCat (both `compose` and `xml`)
1. Tap an hold recoding button -> sends the recording in chat
2. Tap and hold + swipe up to lock the recording. Tap `Stop`, then `Send`  -> sends the recording in chat
3. Tap and hold + swipe up to lock the recording. Tap `Send`  -> sends the recording in chat
4. Set `ChatTheme.messageComposerTheme.audioRecording.sendOnComplete` to `false` - re-run the same tests: For cases 2 and 3, the recording should first be added to the composer, and sent only when tapping the `Send` button.

### DIFF
--- a/stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/AudioRecordingControllerTest.kt
+++ b/stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/AudioRecordingControllerTest.kt
@@ -25,6 +25,7 @@ import io.getstream.result.Result
 import io.getstream.sdk.chat.audio.recording.RecordedMedia
 import io.getstream.sdk.chat.audio.recording.StreamMediaRecorder
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.test.resetMain
 import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.test.setMain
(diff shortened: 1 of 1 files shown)


andremion (approved):

VelikovPetar:
@andremion 

> If possible, try to add unit tests to AudioRecordingController at least

Added tests here: https://github.com/GetStream/stream-chat-android/pull/6036/commits/672f73ae9d99048dfb1e1c09d1efa9013a44e7ca

PR #6039, opened 2025-12-10
Deprecate invalid `getReactions` method

VelikovPetar:
### 🎯 Goal
The `ChannelClient.getReactions(messageId: String, firstReactionId: String, limit: Int)` is invalid, there is no support for fetching reactions after a given `reactionId`. The method internally uses `getRepliesMore`, which is also wrong. The PR deprecates the method because it will never return correct values.

Docs update: https://github.com/GetStream/docs-content/pull/856

### 🛠 Implementation details
- Deprecate `ChannelClient.getReactions(messageId: String, firstReactionId: String, limit: Int)`

### 🎨 UI Changes
_NA_

### 🧪 Testing
_NA_


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/channel/ChannelClient.kt
```diff
@@ -673,6 +673,11 @@ public class ChannelClient internal constructor(
         return client.getReactions(messageId, offset, limit)
     }
 
+    @Deprecated(
+        message = "This operation is not supported. " +
+            "Use ChannelClient.getReactions(messageId: String, offset: Int, limit: Int) instead.",
+        level = DeprecationLevel.ERROR,
```
gpunto: Just for me to understand, how do we usually choose the deprecation level? For example, is `ERROR` in this case because using the function actually yields wrong results?

VelikovPetar: Good question, it should actually be `WARNING`, otherwise it is a breaking change. I instinctively made this `ERROR`, because the usage of this API is completely wrong. 

I will change the deprecation level...

andremion (approved):

PR #6043, opened 2025-12-15
Make Poll Results Dialog Shows All Votes with Pagination Support

andremion:
### 🎯 Goal

Fix poll results dialogs showing only 10 votes when polls have more votes. The `/channels/{type}/{id}/query` endpoint returns a limited number of latest votes in `latest_votes_by_option`, while the total `vote_count` can be higher.

Closes https://linear.app/stream/issue/AND-953

### 🛠 Implementation details

- Added `PollResultsViewController` in `ui-common` to manage poll results state and pagination
- Implemented pagination using `/polls/{poll_id}/votes` endpoint (page size: 25, same as iOS)
- Added `PollResultsViewModel` for both Compose and XML SDKs
- Refactored `PollViewResultDialog` (Compose) and `PollResultsDialogFragment` (XML) to use ViewModel
- Added loading states, error handling, and automatic pagination on scroll
- Moved poll option sorting and winner calculation to the controller
- Added reference counting for poll data to handle configuration changes

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>XML Before</th>
<th>XML After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/93bb9892-2ec6-48f7-97a7-c8378412fd24" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/e0577638-250d-4169-a912-1cfd2bf16476" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

<table>
<thead>
<tr>
<th>Compose Before</th>
<th>Compose After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/4e66ed37-9f2d-4f79-8ff1-e3d72d4c721a" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/38225a15-fa2a-4ff9-91cc-362848af1706" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing

- Unit tests
- Manual testing with polls containing 10+ votes in both SDKs
- Verified configuration changes (screen rotation) don't lose poll data
- Updated snapshot tests for Compose UI

## Summary by CodeRabbit

* **New Features**
  * Redesigned poll results dialog as a bottom-sheet.
  * View all votes per option with paginated vote lists and a "Show All" action.
  * Winner indicator and state-driven poll results view; new view models/controllers expose observable state.

* **Improvements**
  * Improved loading, loading-more and error states with user-facing messages.
  * Layout and accessibility refinements for poll results and vote items.

* **Tests**
  * Added UI snapshots and controller/unit tests for poll results and vote pagination.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@
 ### ⬆️ Improved
 
 ### ✅ Added
+- Add poll results + poll option votes state/controllers to support pagination and error handling.
 
 ### ⚠️ Changed
 
@@ -58,10 +59,13 @@
 
 ## stream-chat-android-ui-components
 ### 🐞 Fixed
+- Fix poll results sheet cutting off the last option items.
+- Fix poll dialogs layout on edge-to-edge devices by applying system bar insets.
 
 ### ⬆️ Improved
 
 ### ✅ Added
+- Add XML poll option votes dialog with pagination support.
 
 ### ⚠️ Changed
 
@@ -69,11 +73,13 @@
 
 ## stream-chat-android-compose
 ### 🐞 Fixed
+- Fix poll results sheet cutting off the last option items.
 
 ### ⬆️ Improved
 - Increase default search debounce period to 300ms. [#6076](https://github.com/GetStream/stream-chat-android/pull/6076)
 
 ### ✅ Added
+- Add Compose poll option votes dialog with pagination support.
 
 ### ⚠️ Changed
 


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ViewModelStore.kt
```diff
@@ -0,0 +1,54 @@
...
+ * Provides a fresh ViewModelStore for the composable content.
+ * The store is cleared when the composable leaves the composition.
+ */
+@Composable
+internal fun ViewModelStore(
+    vararg keys: Any?,
+    content: @Composable () -> Unit,
+) {
+    // Create a fresh ViewModelStore on each new composition
+    val viewModelStore = remember { ViewModelStore() }
```
VelikovPetar, line 37: I might be missing something, but shouldn't we scope the whole `ViewModelStore` composable on the `keys` arg? 
```kotlin
key(keys) {
   val viewModelStore = remember { ViewModelStore() }
   // other setup...
}
```

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/poll/PollResultsViewState.kt
```diff
@@ -0,0 +1,37 @@
...
+ * @param isLoading True if the initial load is in progress. Defaults to true.
+ * @param poll The poll with votes fetched so far.
+ * @param winner The winning option (option with the most votes).
+ * @param canLoadMore True if there are more votes to be loaded. Defaults to true.
+ * @param isLoadingMore True if the loading of the next page is in progress. Defaults to false.
+ */
+public data class PollResultsViewState(
+    val isLoading: Boolean = true,
+    val poll: Poll,
+    val winner: Option?,
```
VelikovPetar: One questions about the winner: If it is a tie between 2 or more option, are we supposed to show the `winner` icon for each winner, or not show it at all?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
```diff
@@ -74,79 +82,100 @@ import io.getstream.chat.android.ui.common.utils.extensions.initials
...
-
-        item { PollViewResultTitle(title = poll.name) }
+    LoadMoreHandler(
+        lazyListState = listState,
+        loadMore = onLoadMoreRequested,
+    )
 
-        item { Spacer(modifier = Modifier.height(16.dp)) }
+    Box(
+        modifier = Modifier.systemBarsPadding(),
```
VelikovPetar: Could adding this padding potentially break integrations which already apply this padding? (by break, I mean to have the padding applied twice)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
```diff
@@ -74,79 +82,100 @@ import io.getstream.chat.android.ui.common.utils.extensions.initials
...
+                )
+            }
 
-        pollViewResultContent(poll = poll)
+            pollResultsContent(state = state)
+        }
 
-        item { Spacer(modifier = Modifier.height(16.dp)) }
+        if (state.isLoading || state.isLoadingMore) {
+            LoadingIndicator(modifier = Modifier.fillMaxSize())
```
VelikovPetar: Wouldn't this be shown over the list of votes? Is that intentional?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollResultsDialogFragment.kt
```diff
@@ -79,41 +99,110 @@ public class PollResultsDialogFragment : AppCompatDialogFragment() {
...
+
     public companion object {
-        public const val TAG: String = "create_poll_dialog_fragment"
+        public const val TAG: String = "PollResultsDialogFragment"
         private const val ARG_POLL: String = "arg_poll"
+
+        // Store polls temporarily to handle configuration changes
+        // Use reference counting to handle multiple fragments or configuration changes
         private val polls = mutableMapOf<String, Poll>()
+        private val pollReferenceCounts = mutableMapOf<String, Int>()
```
VelikovPetar: Bit confused about this part: Why isn't the ViewModel handling configuration changes?

VelikovPetar:
I think the wrong ticket is linked here, should be: AND-953

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ViewModelStore.kt
andremion: No need. The store is clear when the keys are changed.
But let me know if you find any issues with that.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/poll/PollResultsViewState.kt
andremion: Not showing at all. There was an old issue on that scenario, which was aligned with iOS.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
andremion: I believe it's very unlikely, because the old implementation used a Popup dialog (which doesn't provide system insets) and `PollMoreOptionsDialog` never exposed any modifier.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
andremion: Yes. It is shown for the initial load and the pagination loads.
The idea was not to show only the loading UI and block the user, since we already have an initial poll object (which contains 10 max votes).

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollResultsDialogFragment.kt
andremion: We pass the poll object in `newInstance` of the Fragment.
In order to survive config changes, we should use the fragment's argument because `Poll` is a non-parcelable model object. That's why the legacy implementation used that static map.
The problem with that map was that it leaked the poll objects.
With that new implementation, we make sure we remove references on fragment destruction.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ViewModelStore.kt
VelikovPetar: No, if clearing the single store is enough, then no need for other changes. I initially assumed that we always need to create a new store, but if clearing a single store is enough, no need to complicate things

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/poll/PollResultsViewState.kt
VelikovPetar: Got it, thanks for confirming!

VelikovPetar: Got it, thanks for confirming!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
VelikovPetar: Bit strange if you ask me... Maybe this is something to bring up for the upcoming design refresh?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollResultsDialogFragment.kt
VelikovPetar: Ah OK I see. 
Not something to change now, but maybe in the future we can consider 2 different approaches:
1. Pass only ID, and fetch the poll via API
2. Introduce a way to retrieve the poll from the `ChatClient` state, similar to what we do with the channel.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
andremion: Sure.
What UX would you suggest?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollResultsDialogFragment.kt
andremion: Sure.
The problem with the first approach is fetching data when we already have some in memory.
A better UX would be going with the second approach, which requires more code.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
VelikovPetar: I don't really have any good suggestions... Based on how the pagination works in this case, we cannot really know where to show a loading placeholder, so any assumptions could be wrong.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollOptionViewResultDialog.kt
```diff
@@ -0,0 +1,327 @@
...
+ * Displays loading states and error messages when vote data cannot be loaded.
+ *
+ * @param poll The poll containing the option for which results are displayed.
+ * @param option The specific poll option for which vote results are displayed.
+ * @param onDismissRequest Handler invoked when the dialog should be dismissed.
+ * @param onBackPressed Handler invoked when the back button is pressed.
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+internal fun PollOptionViewResultDialog(
```
VelikovPetar: Maybe a more fitting name would be `PollOptionVotesDialog`?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/PollOptionResultsViewModel.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.coroutines.flow.StateFlow
+
+/**
+ * ViewModel for managing poll option results state in Compose UI components.
+ *
+ * @param poll The poll for which the results are displayed.
+ * @param option The option for which the results are displayed.
+ * @param controllerProvider The provider for [PollOptionResultsViewController].
+ */
+public class PollOptionResultsViewModel(
```
VelikovPetar: Similar suggestion here, maybe `PollOptionVotesViewModel` would be more fitting name? (also for the related events, controllers...)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
```diff
@@ -195,70 +207,81 @@ private fun PollViewResultItem(
...
+            ) {
+                Text(text = stringResource(R.string.stream_compose_poll_show_all_votes))
+            }
+        }
     }
 }
 
 @Composable
 private fun PollVoteItem(vote: Vote) {
+    val user = vote.user ?: return
```
VelikovPetar: I believe the `PollVoteItem` is duplicated in the `PollOptionVotesDialog` - would it be possible to unify them?

andremion:
@coderabbitai help

andremion:
@coderabbitai summary

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/poll/PollResultsViewAction.kt
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.ui.common.feature.messages.poll
+
+/**
+ * Represents actions that can be performed from the poll results view.
+ */
+public sealed interface PollResultsViewAction {
```
VelikovPetar: I believe this is no longer used - there is no option to paginate the results on this screen.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/poll/PollResultsViewController.kt
```diff
@@ -0,0 +1,91 @@
...
+            )
+        },
+    )
+
+    /**
+     * The current state of the poll results view.
+     */
+    public val state: StateFlow<PollResultsViewState> = _state.asStateFlow()
+
+    private val _events = MutableSharedFlow<PollResultsViewEvent>(extraBufferCapacity = 1)
```
VelikovPetar: I believe errors are no longer possible in this flow - can we maybe clean this up a bit? (The `PollResultsViewEvent` will not be used anymore too)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollOptionVotesDialogFragment.kt
```diff
@@ -0,0 +1,256 @@
...
+        _binding = StreamUiFragmentPollOptionVotesBinding
+            .inflate(requireContext().streamThemeInflater, container, false)
+        return binding.root
+    }
+
+    override fun getTheme(): Int = R.style.StreamUiBottomSheetDialogTheme
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        setupToolbar(binding.toolbar)
```
VelikovPetar, line 99: It seems this screen doesn't respect edge-to-edge setup: Would it be a big effort to add the appropriate systemBars padding? 
xml1_e2e

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/list/internal/poll/PollResultsDialogFragment.kt
```diff
@@ -57,18 +77,10 @@ public class PollResultsDialogFragment : AppCompatDialogFragment() {
...
-    }
-
-    /**
-     * Initializes the dialog.
-     */
-    private fun setupDialog(poll: Poll) {
         setupToolbar(binding.toolbar)
-        binding.question.text = poll.name
-        binding.optionList.adapter = ResultsAdapter(poll)
+        binding.optionList.adapter = resultsAdapter
```
VelikovPetar, line 84: Similar to the previous comment, it seems we don't handle edge-to-edge here as well:
xml_e2e

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollViewResultDialog.kt
```diff
@@ -278,21 +301,45 @@ private fun PollViewResultTitle(title: String) {
...
+                putString("text", it.text)
+            }
+        }
+    },
+    restore = { bundle ->
+        val id = bundle.getString("id") ?: return@Saver null
+        val text = bundle.getString("text") ?: return@Saver null
+        Option(id = id, text = text)
+    },
+)
```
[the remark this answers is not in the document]
andremion: That's fine. extraData is not used in that case. id and text are the only ones used.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/poll/PollOptionVotesViewController.kt
```diff
@@ -0,0 +1,188 @@
...
+    init {
+        loadRequests.onStart { emit(Unit) } // Triggers the initial load
+            .flatMapLatest { flowOf(fetchPollVotes()) }
+            .onEach { result ->
+                result
+                    .onSuccess(::onSuccessResult)
+                    .onError(::onFailureResult)
+            }
+            .launchIn(scope)
+    }
```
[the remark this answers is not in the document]
andremion: It's unnecessary. It's a very, very remote chance that tryEmit won't emit on that setup.

PR #6051, opened 2025-12-18
Fix showing loading state when channels are loaded offline

VelikovPetar:
### 🎯 Goal
We currently have the following scenario: When opening the ChannelList offline, and there are no matching channels in the DB, we still show a loading screen for up to 5 seconds. This is caused by two seemingly unrelated things:
1. Postponing the `QueryChannels` call for up to 5 seconds, if there is no current `connectionId` (no WS connection)
2. Hiding the loading state ONLY if we have channels in the DB matching the query

```kotlin
// MoshiChatApi.kt

override fun queryChannels(query: QueryChannelsRequest): Call<List<Channel>> {
        val request = io.getstream.chat.android.client.api2.model.requests.QueryChannelsRequest(
            // ...
        )

        val lazyQueryChannelsCall = {
            channelApi.queryChannels(
                // ...
            )
        }

        val isConnectionRequired = query.watch || query.presence
        return if (connectionId.isBlank() && isConnectionRequired) {
            logger.i { "[queryChannels] postponing because an active connection is required" }
            postponeCall(lazyQueryChannelsCall)
        } else {
            lazyQueryChannelsCall()
        }
    }
```

```kotlin
// QueryChannelsLogic.kt

internal suspend fun queryOffline(pagination: AnyChannelPaginationRequest) {
        // ...

        queryChannelsDatabaseLogic.let { dbLogic ->
            fetchChannelsFromCache(pagination, dbLogic)
                .also { channels ->
                    if (channels.isNotEmpty()) {
                        addChannels(channels)
                        loadingPerPage(false, hasOffset)
                    }
                }
        }
    }
```

Because the DB query doesn't return channels we don't reset the loading state. Additionally, since the `QueryChannels` call is postponed for 5 seconds, the loading state is reset only after 5 seconds.

In this PR we now separate the logic for updating the loading state in `QueryChannelsState` covering two separate scenarios:
1. No data in DB for the current query (first time querying) - we keep the current logic. We don't reset the loading state, and rely fully on the online result.
2. Data for the current query exists in the DB (could be zero or more channels) - we optimistically set the data and reset the loading state (because we have already written data for that query in the DB)

**Note: This has the drawback that if the first load of a specific query is done offline, the same 'bug' will appear again. But we currently cannot reliably know from the `QueryChannelsLogic` how the actual query will behave (will it be postponed, will it fail or succeed), so we cannot pre-emptively hide the loading state.**

### 🛠 Implementation details
Implement different handling when we don't have any data in the DB for the current query (first run on the query), and different handling when we have data for the query (zero or more channels)

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After (empty - first load)</th>
<th>After (empty - second load)</th>
<th>After (non-empty)</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/8b4eac95-db87-4150-a751-0b629d47fffa" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/ae88d19f-2dc1-46c2-8511-7e15bcf972d0" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/613b3277-6c56-4160-ad92-55316b617765" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/3961a659-5344-4e73-a957-0bc939c2feb6" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### 🧪 Testing
1. Apply the given patch (adds a query filter which doesn't match any channels)
2. Login with "Petar Velikov"
3. Kill app
4. Turn of internet / Enter airplane mode
5. Open the app
6. The `No channels` screen should be immediately visible without bigger delay

<details>

```
IDEA additional info:
<+>UTF-8
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt	(revision 08e7c73ec4c3a3022f2035aaf5b1cce2fd18e3e1)
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/feature/channel/list/ChannelsActivity.kt	(date 1766059245424)
@@ -60,7 +60,6 @@
 import io.getstream.chat.android.compose.sample.ChatApp
 import io.getstream.chat.android.compose.sample.ChatHelper
 import io.getstream.chat.android.compose.sample.R
-import io.getstream.chat.android.compose.sample.feature.channel.ChannelConstants.CHANNEL_ARG_DRAFT
 import io.getstream.chat.android.compose.sample.feature.channel.add.AddChannelActivity
 import io.getstream.chat.android.compose.sample.feature.channel.add.group.AddGroupChannelActivity
 import io.getstream.chat.android.compose.sample.feature.channel.isGroupChannel
@@ -121,7 +120,7 @@
             filters = Filters.and(
                 Filters.`in`("members", listOf(currentUserId)),
-                Filters.or(Filters.notExists(CHANNEL_ARG_DRAFT), Filters.eq(CHANNEL_ARG_DRAFT, false)),
+                Filters.eq("name", "unknown"), // just a filter that yields no results
             ),
             isDraftMessageEnabled = true,

```

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed loading-state display when querying channels offline — UI no longer shows a persistent loading indicator when no matching channels exist in local storage; offline-to-online transitions behave more predictably.

* **Tests**
  * Added comprehensive tests for offline query caching, loading-state transitions, empty-cache scenarios, and cache-delegation behaviors to improve reliability of offline channel queries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 
 ## stream-chat-android-state
 ### 🐞 Fixed
+- Fix thread replies updating channel unread count. [#6056](https://github.com/GetStream/stream-chat-android/pull/6056)
 - Fix showing channels loading state when querying offline and no matching channels are found in the DB. [#6051](https://github.com/GetStream/stream-chat-android/pull/6051)
 
 ### ⬆️ Improved
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelStateLogic.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelStateLogic.kt
@@ -863,29 +863,54 @@ internal class ChannelStateLogic(
      * @param message The message that was received.
      */
     fun updateCurrentUserRead(eventReceivedDate: Date, message: Message) {
+        // Skip update if the message was already processed
+        val isProcessed = processedMessageIds[message.id] == true
+        if (isProcessed) {
+            return
+        }
+        // Skip update if the channel is muted
         val isMuted = mutableState.muted.value
         if (isMuted) {
-            // Skip update to the unread count if the channel is muted
             processedMessageIds.put(message.id, true)
             return
         }
-        mutableState.read.value
-            ?.takeUnless { it.lastReceivedEventDate.after(eventReceivedDate) }
-            ?.takeUnless { processedMessageIds[message.id] == true }
-            ?.takeUnless {
-                message.user.id == clientState.user.value?.id ||
-                    message.parentId?.takeIf { message.showInChannel } != null
-            }
-            ?.takeUnless { message.shadowed }
-            ?.takeUnless { message.silent }
-            ?.let {
-                updateRead(
-                    it.copy(
-                        lastReceivedEventDate = eventReceivedDate,
-                        unreadMessages = it.unreadMessages.inc(),
-                    ),
-                )
-            }
+        // Skip update for thread replies not shown in channel
+        val isThreadReplyNotInChannel = message.parentId != null && !message.showInChannel
+        if (isThreadReplyNotInChannel) {
+            processedMessageIds.put(message.id, true)
+            return
+        }
+        // Skip update for messages from current user
+        val isFromCurrentUser = message.user.id == clientState.user.value?.id
+        if (isFromCurrentUser) {
+            processedMessageIds.put(message.id, true)
+            return
+        }
+        // Skip update for messages from shadow banned users
+        if (message.shadowed) {
+            processedMessageIds.put(message.id, true)
+            return
+        }
+        // Skip update for silent messages
+        if (message.silent) {
+            processedMessageIds.put(message.id, true)
+            return
+        }
+        // Skip update if the event is outdated
+        val currentRead = mutableState.read.value
+        if (currentRead != null && currentRead.lastReceivedEventDate.after(eventReceivedDate)) {
+            processedMessageIds.put(message.id, true)
+            return
+        }
+        // Update the unread count
+        currentRead?.let {
+            updateRead(
+                it.copy(
+                    lastReceivedEventDate = eventReceivedDate,
+                    unreadMessages = it.unreadMessages.inc(),
+                ),
+            )
+        }
         processedMessageIds.put(message.id, true)
     }
 
--- a/stream-chat-android-state/src/test/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelStateLogicTest.kt
+++ b/stream-chat-android-state/src/test/java/io/getstream/chat/android/state/plugin/logic/channel/internal/ChannelStateLogicTest.kt
@@ -31,7 +31,6 @@ import io.getstream.chat.android.models.Member
 import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.models.User
 import io.getstream.chat.android.models.toChannelData
-import io.getstream.chat.android.positiveRandomInt
 import io.getstream.chat.android.randomCID
 import io.getstream.chat.android.randomChannel
 import io.getstream.chat.android.randomChannelMute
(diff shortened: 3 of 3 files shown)


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
```diff
@@ -57,14 +57,16 @@ internal class QueryChannelsLogic(
...
+        val offlineChannels = fetchChannelsFromCache(pagination, queryChannelsDatabaseLogic)
+        when {
+            offlineChannels == null -> {
+                // No cached spec found, rely on online data. Don't reset loading state here, and await online data.
+            }
+            else -> {
+                // Channels for the spec found (0 or more). Optimistic update and reset loading state.
+                addChannels(offlineChannels)
+                loadingPerPage(false, hasOffset)
+            }
```
gpunto, line 69: What about this case?
1. We start with no channels, i.e. api request succeeds and we cache 0 channels
2. Close the app
3. One or more channels are created in a different instance of the app
4. Reopen app
5. ?

Am I correct if I say that in this case we'll show the empty screen and then, after the api call succeeds, the recently created channels? If so, is this behavior acceptable?

VelikovPetar:
> Forgot to approve: just left a comment to understand a similar scenario (which is not changed by this PR)
@gpunto I think it is a valid concern: Before this change, the user would see a loading screen (not the empty one) before showing the new results. But after the change, they will see empty state -> results state.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
VelikovPetar: Yes this unfortunately is a valid scenario, and the user will first see the empty screen, then the recently created channels. I don't think that we can resolve this issue just by taking the offline data into consideration. 

Another solution that I was considering, is to resolve this by removing the postpone logic, if the device is offline. It would result in instantly firing the call (and updating the UI immediately). While that approach would resolve the original issue, we are kind of breaking the scenario where the device would become online in the postpone timeframe. 

Do you maybe have some other ideas of improving this?

gpunto: No, tbh. Given (what I understand of) the current architecture, it's not easy to handle this. I think it's fine to merge, also because the scenario I described is similar to this other one:
1. Start with **stale** cached channels
2. Open the app
3. Show stale cached channels
4. Show refreshed channels as the api call succeeds

So if we're fine with one I think we should be fine with the other.

gpunto:
> > Forgot to approve: just left a comment to understand a similar scenario (which is not changed by this PR)
> 
> @gpunto I think it is a valid concern: Before this change, the user would see a loading screen (not the empty one) before showing the new results. But after the change, they will see empty state -> results state.

Right! We were checking for emptiness 👍

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
VelikovPetar: Yes unfortunately with the current architecture, this case is pretty difficult to cover completely. The ideal approach Imo would be to completely remove the postponing logic, but at the moment this is not really feasible, as this might have pretty big side effects, and is in theory a breaking change.

gpunto (approved):

PR #6062, opened 2025-12-30
Remove `stream-chat-android-offline` module and move offline support to `stream-chat-android-client`

VelikovPetar:
### Goal
Removes the `stream-chat-android-offline` module and moves the `StreamOfflinePluginFactory/OfflinePlugin` to the LLC(`stream-chat-android-client`). 

**Important:** This is just the first part of the offline support refactor. The next step is to move the `state` support down to the client. Afterwards, we need to revisit the points documented in the following [ticket](https://linear.app/stream/issue/AND-1002/offline-support-enhancements-for-v7). This is **extremely** important before publishing `V7`. 

### Implementation
- Moves ALL classes (and tests) to the `client` module, in package: `io.getstream.chat.android.client.internal.offline`
- The `StreamOfflinePluginFactory` is now `internal` instead of public to disallow instantiation
- The offline support is now configurable by a new `OfflineConfig` class, via the `ChatClient.Builder.offlineConfig(OfflineConfig)` method
- Reworked the logic a bit in `ChatClient` related to the plugin setup - but in the end, the `OfflinePlugin` is still setup internally based on the `OfflineConfig`.

### Testing
The offline support is now enabled by default.
You can test that everything works as expected without any changes in the samples.
- You can opt-out of offline support by setting `ChatClient.Builder().offlineSupport(OfflineConfig(enabled = false))` - and verify that no data is available offline.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -4891,7 +4895,16 @@ internal constructor(
          * @param pluginFactories The factories to be added.
          */
         public fun withPlugins(vararg pluginFactories: PluginFactory): Builder = apply {
-            this.pluginFactories.addAll(pluginFactories)
+            this.pluginFactories = pluginFactories.toList()
```
gpunto: Should we update the KDoc since now we're overwriting the factories instead of adding them?

VelikovPetar: I will update it!
Although after the refactor of offline/state, I would personally remove this method - I don't think we should allow registering custom plugins. (but we can discuss that separately)

PR #6064, opened 2026-01-02
Update the outdated copyright headers (2026)

VelikovPetar:
### 🎯 Goal
Updates outdated copyright headers:
OLD: `Copyright (c) 2014-2022 Stream.io Inc. All rights reserved.`
NEW: `Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.`

**Note**: I will do the same for the other projects, and try to create a GH action to do this every new year.

### DIFF
--- a/metrics/stream-chat-android-metrics/src/main/kotlin/io/getstream/chat/android/metrics/MainActivity.kt
+++ b/metrics/stream-chat-android-metrics/src/main/kotlin/io/getstream/chat/android/metrics/MainActivity.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014-2024 Stream.io Inc. All rights reserved.
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
  *
  * Licensed under the Stream License;
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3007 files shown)


metrics/stream-chat-android-metrics/src/main/kotlin/io/getstream/chat/android/metrics/MainActivity.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014-2024 Stream.io Inc. All rights reserved.
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
```
gpunto, line 2: Do we want to automate this? By default, Spotless doesn't update the target year for already-existing headers, but I think I can change how our plugin works to make it do that.

VelikovPetar: If possible to handle it at plugin level, I think it would be great. It will be much easier to manage, and the copyright year will always be updated in the first PR of the year

gpunto: Then, I'll do that and test it with the other SDKs

VelikovPetar: Thank you!

PR #6069, opened 2026-01-05
Remove `stream-chat-android-state` module and move state support to `stream-chat-android-client`

VelikovPetar:
### Goal
Removes the `stream-chat-android-state` module and moves the `StreamStatePluginFactory/StatePlugin` to the LLC (`stream-chat-android-client`).

### Implementation
- Moves ALL classes to the `client` module: `internal` classes to the package `io.getstream.chat.android.client.internal.state`, and `public` classes to `io.getstream.chat.android.client.api.state`
- The `StreamStatePluginFactory` in now marked as `InternalStreamChatApi` to disallow instantiation. (note: will be made `internal` in an upcoming PR)
- The state is configured via a renamed class `StateConfig`(previously named `StatePluginConfig`), via the `ChatClient.Builder.stateConfig(StateConfig)` method
- The `StatePlugin` is now always instantiated in the `ChatClient`, there is no option to opt-out of it

### Testing
There should be no functional differences - test with either the Compose or XML SDKs and everything should work as before.

### DIFF
--- a/stream-chat-android-compose/api/stream-chat-android-compose.api
+++ b/stream-chat-android-compose/api/stream-chat-android-compose.api
@@ -4731,8 +4731,8 @@ public final class io/getstream/chat/android/compose/viewmodel/channel/ChannelIn
 
 public final class io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel : androidx/lifecycle/ViewModel {
 	public static final field $stable I
-	public fun <init> (Lio/getstream/chat/android/client/ChatClient;Lio/getstream/chat/android/models/querysort/QuerySorter;Lio/getstream/chat/android/models/FilterObject;ILjava/lang/Integer;Ljava/lang/Integer;Lio/getstream/chat/android/state/event/handler/chat/factory/ChatEventHandlerFactory;JZLkotlinx/coroutines/flow/Flow;)V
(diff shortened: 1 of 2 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -4891,7 +4895,16 @@ internal constructor(
          * @param pluginFactories The factories to be added.
          */
         public fun withPlugins(vararg pluginFactories: PluginFactory): Builder = apply {
-            this.pluginFactories.addAll(pluginFactories)
+            this.pluginFactories = pluginFactories.asList()
```
gpunto, line 4938: Do you plan to merge https://github.com/GetStream/stream-chat-android/pull/6062 before/after? I think this and some other files will get conflicts, right?

gpunto (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: I think it doesn't really matter - I will have to update the 2nd PR to merge anyway, it should be the same.

PR #6073, opened 2026-01-08
Introduce StreamButton and StreamIconButton

gpunto:
### 🎯 Goal

[AND-995: Implement new buttons](https://linear.app/stream/issue/AND-995/implement-new-buttons)

Introduce `StreamButton` and `StreamIconButton`

### 🛠 Implementation details

I created base buttons kinda following the familiar Material/Compose API but with Stream-specific styling. They are not meant to be arbitrarily flexible. Rather, they are made to follow closely the Stream design system. And anyway, as we discussed for similar cases, integrators will be able to either customize colors through the theme or replace the component altogether.

Nothing is public for the time being, until we decide if we actually want to expose these base buttons to integrators.

### 🎨 UI Changes

| StreamButton | StreamIconButton |
| --- | --- |
| Screenshot 2026-01-08 at 15 55 47 | Screenshot 2026-01-08 at 15 39 03 |

### 🧪 Testing

You can go to the previews and launch them on a device.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.compose.ui.components.button
+
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.RowScope
+import androidx.compose.foundation.layout.defaultMinSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material3.LocalTextStyle
+import androidx.compose.material3.Surface
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.semantics.role
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.compose.ui.theme.StreamSpacings
+
+@Composable
+internal fun StreamButton(
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    style: StreamButtonStyle = StreamButtonStyleDefaults.primarySolid,
+    size: StreamButtonSize = StreamButtonSize.Medium,
+    content: @Composable RowScope.() -> Unit,
+) {
+    Surface(
+        onClick = onClick,
+        modifier = modifier.semantics { role = Role.Button },
+        enabled = enabled,
+        shape = RoundedCornerShape(percent = 50),
+        color = style.containerColor(enabled),
+        contentColor = style.contentColor(enabled),
+        border = style.border(enabled),
+        interactionSource = remember(::MutableInteractionSource),
+    ) {
+        CompositionLocalProvider(LocalTextStyle provides defaultButtonTextStyle) {
+            Row(
+                Modifier
+                    .defaultMinSize(minHeight = size.minimumSize, minWidth = size.minimumSize)
+                    .padding(horizontal = StreamSpacings.sm),
+                horizontalArrangement = Arrangement.Center,
+                verticalAlignment = Alignment.CenterVertically,
+                content = content,
+            )
+        }
+    }
+}
+
+private val defaultButtonTextStyle = TextStyle(
+    fontWeight = FontWeight.Medium,
+    fontSize = 16.sp,
+    lineHeight = 20.sp,
+)
+
+@Preview(showBackground = true)
+@Composable
+private fun StreamButtonPreview() {
+    ChatTheme {
+        val styles = listOf(
+            StreamButtonStyleDefaults.primarySolid to "Primary Solid",
+            StreamButtonStyleDefaults.primaryGhost to "Primary Ghost",
+            StreamButtonStyleDefaults.secondaryOutline to "Secondary Outline",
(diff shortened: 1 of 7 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
```diff
@@ -0,0 +1,109 @@
...
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.compose.ui.components.button
```
gpunto, line 17: I followed the existing package structure, but if we decide to extract this as part of the core design system, we should change the package to a more generic one

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButtonStyle.kt
```diff
@@ -0,0 +1,128 @@
...
+
+@Stable
+internal fun StreamButtonStyle.containerColor(enabled: Boolean) =
+    if (enabled) containerColor else disabledContainerColor
+
+@Stable
+internal fun StreamButtonStyle.border(enabled: Boolean) =
+    if (enabled) border else disabledBorder
+
+internal object StreamButtonStyleDefaults {
```
gpunto, line 47: For how the tokens are structured at the moment, these styles are not trivial to extract away from this repository. The reason is that, as far as I understand, the colors in Figma all belong to the same set which contains both "design system" generic colors & chat-specific ones.

So with the current state, what we could extract is everything except the actual instantiation of the button styles, i.e. this object.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamSpacings.kt
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package io.getstream.chat.android.compose.ui.theme
+
+import androidx.compose.ui.unit.dp
+
+@Suppress("ObjectPropertyNaming")
+internal object StreamSpacings {
+    val none = 0.dp
+    val _2xs = 4.dp
```
xsahil03x, line 24: maybe we should follow the naming here as 

`xxs` instead of `_2xs`
`xxl` or `xxxl` instead of `_2xl` or `_3xl`

gpunto: I thought about that, but my main concern is that it makes it easier to mistake them 🤷‍♂️. But in the end, when these are autogenerated, we'll inherit what comes from Figma, so maybe we can check with Jurgen

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
```diff
@@ -0,0 +1,122 @@
...
+import androidx.compose.ui.unit.sp
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.compose.ui.theme.StreamSpacings
+
+@Composable
+internal fun StreamButton(
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    style: StreamButtonStyle = StreamButtonStyleDefaults.primarySolid,
```
renefloor, line 51: In Figma `Primary` is the type and `solid` is the style. I think it's better if we keep them as separate variables here as well. I have a comment in figma that not all types implement all styles.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButtonStyle.kt
```diff
@@ -0,0 +1,128 @@
...
+    val primarySolid: StreamButtonStyle
+        @Composable
+        get() {
+            val colors = ChatTheme.colors
```
renefloor, line 51: I think we want to extract colors from `ChatTheme` into a more general `StreamTheme` so the colors can also be used for other products later.

What I'm thinking is to create a primary app color separate from the primary button color, so you probably only have to set your app color and not all the colors on all the components per se.

So it becomes: `colors.buttonTypePrimaryBg ?? colors.primaryColor`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamIconButton.kt
```diff
@@ -0,0 +1,105 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.semantics.Role
+import androidx.compose.ui.tooling.preview.Preview
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.compose.ui.theme.StreamSpacings
+
+@Composable
+internal fun StreamIconButton(
```
renefloor, line 47: Is `IconButton` a different component, or is `IconButton` just a regular `StreamButton` with no label?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
gpunto: At the moment, `StreamButtonStyle` contains a complete styling description for the button and for rendering we just take the values as they are.

If we split it into, let's say, `ButtonType` and `ButtonStyle`, what would you put in each? I guess we could decide colors based on the `type` and then conditionally apply them to the container/border/content based on the `style`, but wouldn't that add some complexity at rendering time that we can avoid? 🤔

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButtonStyle.kt
gpunto: > I think we want to extract colors from ChatTheme into a more general StreamTheme so the colors can also be used for other products later.

I agree, but I have a concern (basically what I mention [here](https://github.com/GetStream/stream-chat-android/pull/6073/changes/BASE..8724d33e764fec27479f3107a0d3dadd5d647b37#r2672663775)). Atm, in Figma there's no distinction between design system and chat-specific tokens (right?), so all the colors live in the same place.

> What I'm thinking is to create a primary app color separate from the primary button color, so you probably only have to set your app color and not all the colors on all the components per se.

I'm a bit worried if we deviate from Figma that it will make it more likely to have inconsistencies. I think that if we want this, we should do it in Figma, as we can easily have tokens referencing other tokens, and then trickle it down to the code.

> So it becomes: colors.buttonTypePrimaryBg ?? colors.primaryColor

If we go for this, I'd centralize this check in the colors instantiation rather than in component-specific code. Meaning the button (and any other component) doesn't need to know each default and doesn't perform any check at rendering time. For example a primary button would always uses `buttonTypePrimaryBg`, which has its default already resolved and only once, when Chat colors were instantiated.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamIconButton.kt
gpunto: I made them different components to enforce padding and shapes, but I can check what it would take to merge them into one

gpunto: Yeah, I think the only issue is padding. As far as I can see in Figma, the button has padding specified when it's `IconOnly=false` but no padding in `IconOnly=true`

Screenshot 2026-01-09 at 13 04 43
Screenshot 2026-01-09 at 13 04 49

gpunto:
Closing this for now, to avoid leaving it dangling here. Will open when needed.

PR #6089, opened 2026-01-19
Introduce new Avatar designs

gpunto:
### 🎯 Goal

Redesigns the avatar components

### 🛠 Implementation details

Note: this PR only introduces the components. I'll replace the old ones in a follow up PR.

Introduce new:
- `Avatar`
- `UserAvatar` + placeholder
- `ChannelAvatar` + placeholde
- `AvatarStack` & `UserAvatarStack`

### 🎨 UI Changes

[Figma](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y/Chat-SDK-Design-System?node-id=164-2813&p=f&m=dev)

Screenshot 2026-01-19 at 14 01 30

### 🧪 Testing

Not visible in the sample yet, since they're not used, but you can run the compose preview.

#### General

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

## Summary by CodeRabbit

* **New Features**
  * Added avatar UI components with online/offline status indicators, multiple sizing options, placeholder support with dynamically assigned colors based on user data, and multi-user stacking capabilities
  * Expanded application color palette with semantic colors for button styling (primary, secondary, destructive, outline, ghost), component states, presence indicators, and avatar background variations
  * Added new team and user vector icons

* **Style**
  * Introduced typography emphasis variants for body, caption, and metadata text to improve visual hierarchy


.idea/codeStyles/Project.xml
```diff
@@ -17,6 +17,40 @@
         </value>
       </option>
     </AndroidXmlCodeStyleSettings>
+    <JavaCodeStyleSettings>
```
gpunto, line 20: Updating this since Android Studio wants to set this style now.

I'm still not convinced we should have this file committed. I understand is to have a shared code style, but we have Spotless anyway for enforcing that.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/StreamAvatar.kt
```diff
@@ -0,0 +1,259 @@
...
+import io.getstream.chat.android.compose.ui.util.applyIf
+import io.getstream.chat.android.models.Channel
+import io.getstream.chat.android.models.User
+import io.getstream.chat.android.previewdata.PreviewChannelData
+import io.getstream.chat.android.previewdata.PreviewUserData
+import io.getstream.chat.android.ui.common.images.resizing.applyStreamCdnImageResizingIfEnabled
+import io.getstream.chat.android.ui.common.utils.extensions.initials
+
+@Composable
+internal fun Avatar(
```
gpunto, line 59: At the moment the various avatar implementations are all in this file to avoid conflicting with the existing ones. I'll move them in the follow up PR when replace old with new implementation.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/utils/extensions/String.kt
```diff
@@ -19,10 +19,24 @@ package io.getstream.chat.android.ui.common.utils.extensions
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
 
 internal fun String.initials(): String {
```
gpunto, line 22: Optimization to do it in O(N) time and avoid the various allocations the previous impl was doing, especially given that this is called at rendering time.

gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPlaceholder.kt
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package io.getstream.chat.android.compose.ui.components.avatar
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
```
VelikovPetar, line 24: Would it make sense to rename this file to `AvatarPlaceholder`, or move the utility methods to `StreamAvatar`? I am not a big fan of `-Utils` classes, as they tend to grow out of control.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/StreamAvatar.kt
```diff
@@ -0,0 +1,315 @@
...
+import io.getstream.chat.android.compose.ui.util.StreamAsyncImage
+import io.getstream.chat.android.compose.ui.util.applyIf
+import io.getstream.chat.android.models.Channel
+import io.getstream.chat.android.models.User
+import io.getstream.chat.android.previewdata.PreviewChannelData
+import io.getstream.chat.android.previewdata.PreviewUserData
+import io.getstream.chat.android.ui.common.images.resizing.applyStreamCdnImageResizingIfEnabled
+import io.getstream.chat.android.ui.common.utils.extensions.initials
+
+@Composable
```
VelikovPetar, line 58: Do we need the `Stream-` prefix on the file name? I think it is redundant, as we anyway don't have any `Stream-` prefixed component in the file.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarStack.kt
```diff
@@ -0,0 +1,69 @@
...
+internal inline fun AvatarStack(
+    overlap: Dp,
+    modifier: Modifier = Modifier,
+    content: @Composable RowScope.() -> Unit,
+) {
+    Row(modifier, horizontalArrangement = Arrangement.spacedBy(-overlap), content = content)
+}
+
+@Composable
+internal fun UserAvatarStack(
```
VelikovPetar: quick question: Are we not exposing these components initially as `public`?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/utils/extensions/String.kt
andremion: I believe the original function is already O(n).
To avoid repeated regex compilation if this is called frequently, maybe we could store "\s+".toRegex() as a private top-level property.
Personally, I’d prefer keeping the original one as it is simpler.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamColors.kt
```diff
@@ -104,9 +154,45 @@ public data class StreamColors(
     public val showMoreCountText: Color,
 
     // Design System semantic colors
-    public val borderCoreOnDark: Color,
+    public val accentError: Color,
```
andremion, line 157: I thought the idea was to create another object with the new design colors.
maybe I missed something...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPlaceholder.kt
```diff
@@ -0,0 +1,60 @@
...
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+
+/** Selects avatar placeholder colors from the avatar color palette based on the [identifier]. */
+@Suppress("MagicNumber")
+@Composable
+internal fun rememberAvatarPlaceholderColors(identifier: Any): Pair<Color, Color> {
+    val colors = ChatTheme.colors
+
+    return remember(identifier, colors) {
+        when (identifier.hashCode() % 5 + 1) {
```
VelikovPetar, line 34: Just to verify: Should this modulo logic be aligned across platforms? I am thinking that the `identifier.hashCode() % 5 + 1` logic might not produce the same remainder for other platforms, so we might end up with a different colours for same user/ avatar. 

(I am not sure if this is a valid concern, but I wanted to point it out)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/StreamAvatar.kt
```diff
@@ -0,0 +1,315 @@
...
+                        painter = painter,
+                        contentDescription = null,
+                    )
+                }
+            }
+        },
+    )
+}
+
+public object AvatarSize {
```
VelikovPetar, line 93: We are missing KDoc here

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/utils/extensions/String.kt
gpunto: Yes, it's O(N) but iterates multiple times, allocating a few intermediate objects: 1 string for `trim`, 1 array of strings + X strings for `split`, a list for `take(2)`, 2-4 strings for `take(1).uppercase()`, the final result and a few iterators along the way. It feels a lot for finding 2 character in a string, especially if we're calling it at rendering time.

But it's true that the original was simpler code-wise. I'm going to revert it 👍

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamColors.kt
gpunto: My plan was to add the new colors here and then remove the old ones while we're redesigning as they won't be referenced anymore. We could add a new object, but `StreamColor` sounds like the correct name for this

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/StreamAvatar.kt
gpunto: No, we definitely don't. I'm going to remove this file in my [next PR](https://github.com/GetStream/stream-chat-android/compare/redesign/avatar...redesign/avatar-replace?expand=1) when I replace the old avatars with the new ones. In there I'm removing the current files (e.g. we already have Avatar.kt) and splitting the implementations into their own files.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPlaceholder.kt
gpunto: Reading [in Figma](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y/Chat-SDK-Design-System?node-id=1780-17638&m=dev) I got the impression that we didn't have to be aligned across platforms. That would be a bit tricky because it'd mean we'd need a consistent hashing algorithm across platforms, but I'll ask in a comment in Figma!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/StreamAvatar.kt
gpunto: Yeah, I noticed while working on the rest of this task and is fixed in my [next PR](https://github.com/GetStream/stream-chat-android/compare/redesign/avatar...redesign/avatar-replace?expand=1). Would that be fine with you? It's just to avoid some conflicts since StreamAvatar.kt is deleted in that branch

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPlaceholder.kt
gpunto: Me neither 😅 I ended up with this because later I'm splitting stuff into `Avatar.kt`, `ChannelAvatar.kt` and `UserAvatar.kt` and these are common to both placeholders. But I can definitely rename it to `AvatarPlaceholder`!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/StreamAvatar.kt
VelikovPetar: Sounds good thank you!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarStack.kt
gpunto: Good catch, let me address it

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPlaceholder.kt
VelikovPetar: Yeah I assumed that this is probably not really feasible to align, but I wanted to double check.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/StreamAvatar.kt
VelikovPetar: Yes of course, just wanted to make sure we don't miss it!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPlaceholder.kt
VelikovPetar: Resolving the remark as I don't think it is a blocker, we can address it additionally if needed.

PR #6090, opened 2026-01-20
Implement operation to clear local data created by the SDK

VelikovPetar:
### 🎯 Goal

Add a new public API method to `ChatClient` that allows developers to clear all cache and temporary files created by the Stream Chat SDK.
Additionally, it re-works the current cache and file management logic to a more streamlined one, routing all operations via a new `StreamFileManager`.

### 🛠 Implementation details
1. Adds a new public `ChatClient.clearCacheAndTemporaryFiles(context: Context): Call<Unit>` for clearing ALL cache and files stored in the external storage.
2. Remove the `StreamFileUtil` and replace it with:
    - `StreamFileManager` - low-level cache and external storage manager
    - `StreamShareFileManager` - high-level manager which handles the logic for sharing files from the SDK
    - `ShareableFileProvider` - utility class for generating sharable URIs via a `FileProvider`
3. Relpace all usages of the `StreamFileUtil` with the new managers

### 🎨 UI Changes

_No UI changes - this is an API addition_

### 🧪 Testing
There should be no changes in the behaviour of the SDK. 
Only calling `ChatClient.clearCacheAndTemporaryFiles` manually should call clear the default cache, image cache and external storage.

### 🎉 GIF

cache-cleanup

## Summary by CodeRabbit

* **New Features**
  * Added a public action to clear the SDK cache and temporary media files using the app Context.

* **Refactor**
  * Reworked file, cache and sharing flows to use new injectable file managers and a shareable-URI provider; replaced legacy static utility with these abstractions and updated media capture, sharing and cleanup paths.

* **Tests**
  * Added extensive unit and instrumented tests covering caching, external media, sharing and cleanup behaviors.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotChannelListAsserts.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotChannelListAsserts.kt
@@ -29,10 +29,9 @@ fun UserRobot.assertChannelAvatar(): UserRobot {
     return this
 }
 
-fun UserRobot.assertMessageInChannelPreview(text: String, fromCurrentUser: Boolean): UserRobot {
-    val expectedPreview = if (fromCurrentUser) "You: $text" else text
+fun UserRobot.assertMessageInChannelPreview(text: String, fromCurrentUser: Boolean? = null): UserRobot {
+    val expectedPreview = if (fromCurrentUser == true) "You: $text" else text
     assertEquals(expectedPreview, Channel.messagePreview.waitToAppear().text.trimEnd())
-    assertTrue(Channel.timestamp.isDisplayed())
     return this
 }
 
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
@@ -242,6 +242,12 @@ fun UserRobot.assertMentionWasApplied(): UserRobot {
     return this
 }
 
+fun UserRobot.assertComposerText(expectedText: String): UserRobot {
+    val actualText = Composer.inputField.waitToAppear().text
+    assertEquals(expectedText, actualText)
+    return this
+}
+
 fun UserRobot.assertScrollToBottomButton(isDisplayed: Boolean): UserRobot {
     if (isDisplayed) {
         assertTrue(MessageListPage.MessageList.scrollToBottomButton.waitToAppear().isDisplayed())
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/DraftMessagesTests.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/DraftMessagesTests.kt
@@ -1,0 +1,140 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.compose.tests
+
+import io.getstream.chat.android.compose.robots.assertComposerText
+import io.getstream.chat.android.compose.robots.assertMessageInChannelPreview
+import io.getstream.chat.android.compose.sample.ui.InitTestActivity
+import io.getstream.chat.android.compose.uiautomator.device
+import io.getstream.chat.android.compose.uiautomator.disableInternetConnection
+import io.getstream.chat.android.compose.uiautomator.enableInternetConnection
+import io.qameta.allure.kotlin.Allure.step
+import io.qameta.allure.kotlin.AllureId
+import org.junit.Test
+
+class DraftMessagesTests : StreamTestCase() {
+
+    override fun initTestActivity() = InitTestActivity.UserLogin
+
+    private val firstMessage = "message"
+    private val draftMessage = "alright"
+
+    @AllureId("10106")
+    @Test
+    fun test_updateChannelDraftMessage() {
+        step("GIVEN user opens the channel") {
+            userRobot.login().openChannel()
+        }
+        step("AND user sends a message") {
+            userRobot.sendMessage(firstMessage)
+        }
+        step("WHEN user inputs some text in the composer") {
+            userRobot.typeText(draftMessage)
+        }
+        step("AND user leaves the Channel") {
+            userRobot.tapOnBackButton()
+        }
+        step("THEN the draft message is in preview") {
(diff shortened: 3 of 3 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/internal/file/StreamShareFileManager.kt
```diff
@@ -0,0 +1,147 @@
...
+
+    private suspend fun getCachedFileForAttachment(context: Context, attachment: Attachment): Result<File> {
+        val fileName = getCacheFileName(attachment)
+        return fileManager.getFileFromCache(context, fileName).flatMap { file ->
+            // Ensure attachment was really cached
+            if (file.exists() && file.length() == attachment.fileSize.toLong()) {
+                Result.Success(file)
+            } else {
+                Result.Failure(Error.GenericError("Cached file is invalid or incomplete."))
+            }
```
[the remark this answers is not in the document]
VelikovPetar: This is OK, we can always assume correct `attachment.fileSize`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
```diff
@@ -0,0 +1,360 @@
...
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    private fun getOrCreateCacheDir(context: Context): Result<File> {
+        return try {
+            val cacheDir = getStreamCacheDir(context)
+            cacheDir.mkdirs()
+            Result.Success(cacheDir)
+        } catch (e: Exception) {
+            Result.Failure(Error.ThrowableError("Could not get or create cache directory.", e))
```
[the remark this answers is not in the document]
VelikovPetar: This follows the old implementation from `StreamFileUtil`/`StorageHelper`. `mkdirs` also returns `false` if the directory already exist, so that doesn't fit our use-case.
```kotlin
public boolean mkdirs() {
        if (exists()) {
            return false;
        }
        if (mkdir()) {
            return true;
        }
        File canonFile = null;
        try {
            canonFile = getCanonicalFile();
        } catch (IOException e) {
            return false;
        }

        File parent = canonFile.getParentFile();
        return (parent != null && (parent.mkdirs() || parent.exists()) &&
                canonFile.mkdir());
    }
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
```diff
@@ -0,0 +1,360 @@
...
+            Result.Failure(Error.ThrowableError("Failed to write file to cache.", e))
+        }
+    }
+
+    /**
+     * Writes data from an InputStream to a file in a timestamped cache folder.
+     *
+     * This method creates a unique timestamped folder (format: `STREAM_HHmmssSSS`) and writes
+     * the file there. This is useful for caching files from URIs where unique folder isolation
+     * is needed to prevent naming conflicts.
```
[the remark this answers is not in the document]
VelikovPetar: This is intentional, it follow the old approach. We might re-visit this in the future.

stream-chat-android-ui-common/api/stream-chat-android-ui-common.api
```diff
@@ -3098,8 +3098,8 @@ public abstract interface class io/getstream/chat/android/ui/common/utils/typing
 
 public final class io/getstream/sdk/chat/audio/recording/DefaultStreamMediaRecorder : io/getstream/sdk/chat/audio/recording/StreamMediaRecorder {
 	public static final field $stable I
-	public fun <init> (Landroid/content/Context;IIIIII)V
-	public synthetic fun <init> (Landroid/content/Context;IIIIIIILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Landroid/content/Context;IIIIIILio/getstream/chat/android/client/internal/file/StreamFileManager;)V
+	public synthetic fun <init> (Landroid/content/Context;IIIIIILio/getstream/chat/android/client/internal/file/StreamFileManager;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public fun deleteRecording (Ljava/io/File;)Lio/getstream/result/Result;
```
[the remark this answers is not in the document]
VelikovPetar: It's oK, we don't put JvmOverloads

VelikovPetar:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
```diff
@@ -0,0 +1,381 @@
...
+                source.use { input ->
+                    file.outputStream().use { output ->
+                        input.copyTo(output)
+                    }
+                }
+                Result.Success(file)
+            }
+        } catch (e: Exception) {
+            Result.Failure(Error.ThrowableError("Failed to write file to timestamped cache.", e))
+        }
```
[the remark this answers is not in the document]
VelikovPetar: This is intentional, it is not possible to change this without making a breaking change. The usage of this method in wrapped in  a proper coroutines higher in the hierarchy

VelikovPetar:
@coderabbitai review

VelikovPetar:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
```diff
@@ -0,0 +1,391 @@
...
+     * - Videos: `STREAM_VID_{timestamp}.mp4`
+     *
+     * @param context Android context for accessing external storage directories
+     * @return [Result.Success] if cleanup succeeded, or [Result.Failure] with an error
+     */
+    @Suppress("TooGenericExceptionCaught")
+    public fun clearExternalStorage(context: Context): Result<Unit> {
+        return try {
+            val picturesDir = context.getExternalFilesDir(Environment.DIRECTORY_PICTURES)
+            val moviesDir = context.getExternalFilesDir(Environment.DIRECTORY_MOVIES)
```
gpunto, line 276: Unrelated to this PR: why do we use external directory movies & pictures at all? Aren't those meant for user-visible content? Maybe we should evaluate ditching this strategy for v7 and just use the cache.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
```diff
@@ -0,0 +1,391 @@
...
+            }
+
+            // Clear Stream videos from Movies directory
+            moviesDir?.listFiles { file ->
+                file.isFile && file.name.startsWith(VIDEO_PREFIX)
+            }?.forEach { file ->
+                if (!file.delete()) {
+                    allDeleted = false
+                }
+            }
```
gpunto, line 296: Very unlikely, but here we're risking deleting user content if they somehow have that prefix (not a blocker here, but would be solved if we move away from external folders in v7)

VelikovPetar: We can maybe strengthen the check here, to fully match the pattern used for creating the file name `"${prefix}_${dateFormat.format(Date().time)}.$extension"`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
VelikovPetar: To be honest I am not sure about the historic reason for storing the captured photos/videos in the external files dir - it is how the `CaptureMediaContract` handled it previously. 
But I think that we should change this for V7 (it is kind of a breaking change to change it now)

VelikovPetar: Created a [ticket](https://linear.app/stream/issue/AND-1027/verify-usage-of-getexternalfilesdir-in) for V7

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
gpunto: Sounds good to me!

PR #6093, opened 2026-01-22
Fix `markDelivered` not working when switching users

VelikovPetar:
### 🎯 Goal
The `MessageReceiptManager` keeps a `lazy` reference to the `RepositoryFacade`. The facade is switched with a new instance if the connected `User` changes, but the `MessageReceiptManager` is not. 
So if the connected `User` changes, the `MessageReceiptManager` by keeping a `lazy` reference to the (old) `RepositoryFacade` attempts to read outdated data (from the previously connected `User`).

### 🛠 Implementation details
- Remove the `lazy` persistence of the `RepositoryFacade` from the `MessageReceiptManager `

### 🎨 UI Changes
_NA_

### 🧪 Testing
1. Log in, log out, and log in with a different user
2. Check the logs for `Chat:MessageReceiptManager`
3. There should be no WARNING logs with text: `[markChannelsAsDelivered] Current user not found`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the markDelivered operation would not function correctly when switching between users.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
@@ -36,16 +36,19 @@ import java.util.Date
  * for later reporting to the server.
  *
  * @param now Function to provide the current date.
- * @param repositoryFacade Function to provide the [RepositoryFacade] tied to the currently logged in user.
+ * @param getRepositoryFacade Function to provide the [RepositoryFacade] tied to the currently logged in user.
  * @param messageReceiptRepository The [MessageReceiptRepository] to store the created receipts.
  * @param api The [ChatApi] to fetch data if needed.
  */
 internal class MessageReceiptManager(
     private val now: () -> Date,
-    private val repositoryFacade: () -> RepositoryFacade,
+    private val getRepositoryFacade: () -> RepositoryFacade,
     private val messageReceiptRepository: MessageReceiptRepository,
     private val api: ChatApi,
 ) {
+
+    private val repositoryFacade: RepositoryFacade
+        get() = getRepositoryFacade()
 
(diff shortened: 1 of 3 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
```diff
@@ -34,14 +34,18 @@ import java.util.Date
...
+ *
+ * @param now Function to provide the current date.
+ * @param repositoryFacade Function to provide the [RepositoryFacade] tied to the currently logged in user.
+ * @param messageReceiptRepository The [MessageReceiptRepository] to store the created receipts.
+ * @param api The [ChatApi] to fetch data if needed.
  */
 internal class MessageReceiptManager(
     private val now: () -> Date,
-    private val getRepositoryFacade: () -> RepositoryFacade,
+    private val repositoryFacade: () -> RepositoryFacade,
```
andremion: `getRepositoryFacade` seems to have better semantics for a lambda that gets the repository

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/receipts/MessageReceiptManager.kt
```diff
@@ -34,14 +34,18 @@ import java.util.Date
...
     private val messageReceiptRepository: MessageReceiptRepository,
     private val api: ChatApi,
 ) {
-    private val repositoryFacade by lazy { getRepositoryFacade() }
```
andremion, line 44: What about this? `private val repositoryFacade get() = getRepositoryFacade()`
So we keep the semantics of `getRepositoryFacade`, which is actually a function to get something

VelikovPetar: I will update it!

PR #6094, opened 2026-01-22
[CI] Add E2E tests for draft messages

testableapple:
### 🔗 Issue Links

Resolve https://linear.app/stream/issue/IOS-930 and https://linear.app/stream/issue/AND-618

### 🎯 Goal

Add E2E tests for draft messages

### 🧪 Manual Testing Notes

Run new tests locally from Android Studio or CLI

## Release Notes

* **Tests**
  * Expanded test coverage for draft message functionality, including channel drafts, thread drafts, and offline scenarios to ensure reliability.
  * Added assertion helpers for validating composer text and message preview states during automated testing, improving test suite robustness and flexibility.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/StreamTestCase.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/StreamTestCase.kt
@@ -50,7 +50,7 @@ abstract class StreamTestCase {
     var testName: TestName = TestName()
 
     @get:Rule
-    val retryRule = RetryRule(1)
+    val retryRule = RetryRule(3)
 
     @Before
     fun setUp() {


stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotChannelListAsserts.kt
```diff
@@ -29,10 +29,9 @@ fun UserRobot.assertChannelAvatar(): UserRobot {
     return this
 }
 
-fun UserRobot.assertMessageInChannelPreview(text: String, fromCurrentUser: Boolean): UserRobot {
-    val expectedPreview = if (fromCurrentUser) "You: $text" else text
+fun UserRobot.assertMessageInChannelPreview(text: String, fromCurrentUser: Boolean? = null): UserRobot {
```
VelikovPetar, line 32: Just a small suggestion: Maybe the `fromCurrentUser` can be defined as non-nullable: `fromCurrentUser: Boolean = false`? The check later would be simpler:
```kotlin
val expectedPreview = if (fromCurrentUser) "You: $text" else text
```

testableapple: i was thinking about it, but then it might be a bit misleading with draft messages, as the draft message is always `fromCurrentUser` but if you pass `true` the assertion will fail. WDYT?

testableapple: we discussed this internally, decided to move forward with the current implementation. Later, if this will be required a refactoring due to some other issue, we can create a new func `assertDraftMessageInChannelPreview`

PR #6103, opened 2026-01-26
Simplify MessagePositionHandler logic and API

gpunto:
### 🎯 Goal

While working on the message redesign, I noticed we could simplify the logic of `MessagePositionHandler` & related code

### 🛠 Implementation details

- Changed output from `List<MessagePosition>` to `MessagePosition` since a message can't be simultaneously in multiple positions and we were always using one position from the list anyway
- Refactor the logic of the default handler to be easier to understand

### 🎨 UI Changes

None

### 🧪 Testing

The message position affects the styling, so you should see no styling differences in the sample as this PR doesn't change behavior

## Summary by CodeRabbit

* **Refactor**
  * Streamlined message position handling to use single values instead of lists for improved code clarity.
  * Updated message styling to use explicit theme parameters for consistent bubble rendering.
  * Consolidated message rendering logic across components for better maintainability.

### DIFF
--- a/stream-chat-android-ui-components/api/stream-chat-android-ui-components.api
+++ b/stream-chat-android-ui-components/api/stream-chat-android-ui-components.api
@@ -2882,10 +2882,10 @@ public final class io/getstream/chat/android/ui/feature/messages/list/adapter/Me
 }
 
 public final class io/getstream/chat/android/ui/feature/messages/list/adapter/MessageListItem$MessageItem : io/getstream/chat/android/ui/feature/messages/list/adapter/MessageListItem {
-	public fun <init> (Lio/getstream/chat/android/models/Message;Ljava/util/List;ZLjava/util/List;ZZZZZ)V
-	public synthetic fun <init> (Lio/getstream/chat/android/models/Message;Ljava/util/List;ZLjava/util/List;ZZZZZILkotlin/jvm/internal/DefaultConstructorMarker;)V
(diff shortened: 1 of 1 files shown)


gpunto:
@coderabbitai review

gpunto:
@coderabbitai review

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/utils/extensions/MessageListItem.kt
```diff
@@ -43,7 +43,7 @@ public fun MessageListItemCommon.toUiMessageListItem(): MessageListItem {
         is TypingItemState -> MessageListItem.TypingItem(users = typingUsers)
         is MessageItemState -> MessageListItem.MessageItem(
             message = message,
-            positions = groupPosition,
+            positions = listOf(groupPosition),
```
VelikovPetar: We can probably update the `MessageListItem.MessageItem#positions` as well - it is still defined as `val positions: List<MessagePosition>`

gpunto: I opted to avoid that to limit the scope since we're not changing the xml SDK, but I can propagate the changes

VelikovPetar: Ah sorry, didn't want to introduce more work. But I see it is already done, thanks a bunch!

gpunto: No problem at all :)

PR #6106, opened 2026-01-26
Update app background color

gpunto:
### 🎯 Goal

Updating the background color in its own PR because it causes most snapshots to change.

### 🛠 Implementation details

Update `appBackground` and the snapshots

### 🎨 UI Changes

The color is now white as (barely) visible in the new snapshots


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamColors.kt
```diff
@@ -179,6 +179,7 @@ public data class StreamColors(
     public val textSecondary: Color,
     public val stateBgDisabled: Color,
     public val stateTextDisabled: Color,
+    public val appBackground: Color = backgroundElevationElevation0,
```
andremion, line 182: weird name: `backgroundElevationElevation0` 😅

gpunto: Yeah, it's because we're merging the token path & name 😅 

Screenshot 2026-01-27 at 15 03 08

PR #6113, opened 2026-01-30
Update link and media attachment design in messages

gpunto:
### 🎯 Goal

Continuing with the design updates for the message component. This time updating link & media attachment design

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_20260130_101447 | Screenshot_20260130_102436 |
| Screenshot_20260130_101407 | Screenshot_20260130_102428 |

### 🧪 Testing

They can be checked in the sample

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -8,7 +8,7 @@ concurrency:
 
 jobs:
   base-android-ci:
-    uses: GetStream/stream-build-conventions-android/.github/workflows/android-ci.yml@v0.6.1
+    uses: GetStream/stream-build-conventions-android/.github/workflows/android-ci.yml@v0.7.1
     secrets: inherit
 
   detekt:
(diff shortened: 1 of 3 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/LinkAttachmentContent.kt
```diff
@@ -154,140 +149,116 @@ public fun LinkAttachmentContent(
...
-        if (imagePreviewUrl != null) {
-            LinkAttachmentImagePreview(attachment, isMine)
+        attachment.imagePreviewUrl?.let {
+            LinkAttachmentImagePreview(it)
         }
 
-        val title = attachment.title
-        if (title != null) {
-            LinkAttachmentTitle(title)
+        Spacer(Modifier.height(12.dp))
```
andremion: Should you use `StreamTokens.spacingSm`?

gpunto: Yes, good catch!

PR #6119, opened 2026-02-02
Use component factory when rendering file attachments in messages

gpunto:
### 🎯 Goal

This is the first PR restructuring the message content. We're going to remove the bespoke `AttachmentFactory` implementations in favor of relying on `ChatComponentFactory`.

This PR is intentionally small to show the approach in one single case and get feedback.

### 🛠 Implementation details

Stop using the `AttachmentFactory` for file attachment and directly call the component factory instead.

### 🎨 UI Changes

None

### 🧪 Testing

It's testable in the sample by checking that file attachments are rendered correctly

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
@@ -37,6 +37,7 @@ import io.getstream.chat.android.models.AttachmentType
 import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.models.User
 import io.getstream.chat.android.ui.common.state.messages.list.GiphyAction
+import io.getstream.chat.android.ui.common.utils.extensions.isUploading
 
 /**
  * Represents the default message content within the bubble that can show different UI based on the message state.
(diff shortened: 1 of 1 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
```diff
@@ -115,11 +112,6 @@ public object StreamAttachmentFactories {
             streamCdnImageResizing: StreamCdnImageResizing,
             skipEnrichUrl: Boolean,
         ) -> Unit = ::onMediaAttachmentContentItemClick,
-        showFileSize: (Attachment) -> Boolean = { true },
-        onFileContentItemClick: (
```
VelikovPetar, line 119: Just for my understanding - what is the plan for handling (and intercepting/overriding) user clicks on the attachments? 

Currently some are handled via the `ChatThem.attachmentPreviewHandlers` (like this one), but some of them are handled directly (ex. `onMediaContentItemClick`)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
```diff
@@ -166,6 +172,24 @@ internal fun DefaultMessageContent(
...
+        }
+        val attachmentState = AttachmentState(
+            message = message,
+            isMine = message.user.id == currentUser?.id,
+            onLongItemClick = onLongItemClick,
+            onMediaGalleryPreviewResult = onMediaGalleryPreviewResult,
+        )
+
+        if (AttachmentType.FILE in attachmentTypes || AttachmentType.AUDIO in attachmentTypes) {
+            componentFactory.FileAttachmentContent(
```
VelikovPetar: I believe that currently we have a small UI glitch here, because during file upload, we can see 2 attachments in the `MessageList` (one from the `componentFactory.FileAttachmentContent`, and one from the `MessageAttachmentsContent`):

https://github.com/user-attachments/assets/4ec3e995-8018-443c-993a-a69fb9814a85

gpunto: Good point, it's addressed it in my last branch in this series (`message-restructure-all`), but there's no point in not having it correct from the beginning. Let me address it

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/StreamAttachmentFactories.kt
gpunto: No plan yet. Do you mean that we should change the behavior we have to unify it across attachments?

VelikovPetar: No not really, I was just curious because I saw this parameter being removed here. Probably would be good if we can unify it, but I guess it is better to handle this after all the attachment factories are re-worked. Resolving the comment.

VelikovPetar (approved):

PR #6121, opened 2026-02-02
Introduce new buttons

gpunto:
### 🎯 Goal

Here we go again: here are new buttons with Stream-specific styling. They are not meant to be arbitrarily flexible. Rather, they are made to follow closely the Stream design system. Nothing is public for the time being. If we decide we actually want to expose these base buttons to integrators, we can do so later.

### 🛠 Implementation details

- Create base `StreamButton` that accept a style and size and can contain arbitrary content
- Create `StreamTextButton` that is a shortcut for the common icon-text-icon layout
- Added/updated colors

### 🎨 UI Changes

Screenshot 2026-02-02 at 14 04 22

### 🧪 Testing

They're not in use atm, but you can check the preview

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
@@ -1,0 +1,152 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/button/StreamButton.kt
```diff
@@ -0,0 +1,152 @@
...
+        ) {
+            leadingIcon?.let { Icon(painter = it, contentDescription = null) }
+            Text(text)
+            trailingIcon?.let { Icon(painter = it, contentDescription = null) }
+        }
+    }
+}
+
+@Composable
+internal fun StreamButton(
```
gpunto, line 82: I used the `Stream` prefix to avoid colliding with the Material buttons, but let me know if you have a better suggestion!

andremion (approved):

PR #6122, opened 2026-02-02
Fix globalState race condition crash during disconnect

VelikovPetar:
### 🎯 Goal

Fix a race condition that causes `IllegalStateException: Plugin 'StatePlugin' was not found` crash when `globalStateFlow` is being collected during user disconnect.

**The Problem:**

The `globalStateFlow` extension filters for `InitializationState.COMPLETE` before accessing `globalState`:

```kotlin
public val ChatClient.globalStateFlow: Flow<GlobalState>
    get() = clientState.initializationState
        .filter { it == InitializationState.COMPLETE }
        .map { globalState }  // Calls resolveDependency<StatePlugin>()
```

In `disconnectUserSuspend()`, the ordering is:
```kotlin
// ChatClient.kt

plugins.forEach { it.onUserDisconnected() }
plugins = emptyList()                    // plugins cleared FIRST
// ... other cleanup ...
mutableClientState.clearState()          // Sets InitializationState to NOT_INTIALIZED
```

This created a race window where:
- `initializationState == COMPLETE` (not yet cleared)
- `plugins == emptyList()` (already cleared)

Any `globalStateFlow` collection could potentially pass the `initializationState == COMPLETE` check, but try to access an empty `plugins` field.

Resolves: https://linear.app/stream/issue/AND-1038

### 🛠 Implementation details

Set `initializationState` to `NOT_INITIALIZED` **before** clearing plugins:

```kotlin
notifications.onLogout()
// Set initializationState to NOT_INITIALIZED BEFORE clearing plugins to prevent race condition.
// This ensures globalStateFlow's filter { it == COMPLETE } won't pass while plugins is being cleared.
mutableClientState.setInitializationState(InitializationState.NOT_INITIALIZED)

plugins.forEach { it.onUserDisconnected() }
plugins = emptyList()
```

This ensures the flow's `.filter { it == COMPLETE }` stops emitting before `plugins` becomes empty.

### 🎨 UI Changes

No UI changes.

### 🧪 Testing

This issue was reported by a customer and is not reproducible with the sample apps under normal conditions due to timing dependencies.

**To simulate the race condition:**

1. Add a delay in `resolvePluginDependency()` after the initialization state check:
   ```kotlin
   internal inline fun <reified P : DependencyResolver, reified T : Any> resolvePluginDependency(): T {
       val initState = awaitInitializationState(RESOLVE_DEPENDENCY_TIMEOUT)
       if (initState != InitializationState.COMPLETE) {
           throw IllegalStateException("...")
       }
       Thread.sleep(500)  // Add this to widen the race window
       val resolver = plugins.find { it is P } ?: throw IllegalStateException("...")
       // ...
   }
   ```

2. Trigger a disconnect shortly after connect completes while `globalStateFlow` is being collected:
   ```kotlin
   MainScope().launch {
       delay(350)
       ChatHelper.disconnectUser()
   }
   ```

3. Without the fix: crash with `IllegalStateException: Plugin 'StatePlugin' was not found`
4. With the fix: no crash, flow stops emitting gracefully

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition crash that could occur when disconnecting the chat client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## stream-chat-android-client
 ### 🐞 Fixed
+- Fix race condition crash when accessing `globalState` during disconnect. [#6122](https://github.com/GetStream/stream-chat-android/pull/6122)
 
 ### ⬆️ Improved
 


aleksandar-apostolov:
LGTM 👍         
  
One minor question: are there any `onUserDisconnected()` plugin callbacks that depend on `initializationState == COMPLETE`? If so, they now see `NOT_INITIALIZED`. Probably fine since they're shutting down anyway, but wanted to flag it.

VelikovPetar:
> LGTM 👍
> 
> One minor question: are there any `onUserDisconnected()` plugin callbacks that depend on `initializationState == COMPLETE`? If so, they now see `NOT_INITIALIZED`. Probably fine since they're shutting down anyway, but wanted to flag it.

No, the `onUserDisconnected()` implementations don't depend on the `initializationState` at all. They are all empty actually, only the internal `ThrottlingPlugin` has logic in it, but not one that depends on `initializationState

PR #6129, opened 2026-02-04
Configure publishing through convention plugins

gpunto:
### Goal

Configure publishing through convention plugins

### Implementation

- Replace the bespoke publishing logic with the common one from the build conventions repo
- Also removed the custom changelog-related gradle plugins and the `buildSrc` module with them
- Remove separate docs release workflow, as now it's a step when publishing the release

### 🎨 UI Changes

None

### Testing

We'll test it in the next release

## Release Notes

* **Chores**
  * Updated version to 6.32.0
  * Migrated release workflow to use external build conventions
  * Updated copyright year to 2026 across source files
  * Simplified build configuration by consolidating publishing and release processes
  * Updated build dependencies to latest versions

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -244,5 +244,4 @@ stream-project = { id = "io.getstream.project", version.ref = "streamConventions
 stream-android-library = { id = "io.getstream.android.library", version.ref = "streamConventions" }
 stream-android-application = { id = "io.getstream.android.application", version.ref = "streamConventions" }
 stream-java-library = { id = "io.getstream.java.library", version.ref = "streamConventions" }
-stream-publish = { id = "io.getstream.publish", version.ref = "streamConventions" }
 paparazzi = { id = "app.cash.paparazzi", version.ref = "paparazzi"}


gpunto:
@coderabbitai review

stream-chat-android-ai-assistant/build.gradle.kts
```diff
@@ -1,13 +1,9 @@
-import com.vanniktech.maven.publish.AndroidSingleVariantLibrary
-import io.getstream.chat.android.Configuration
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 plugins {
     alias(libs.plugins.stream.android.library)
     alias(libs.plugins.kotlin.android)
     alias(libs.plugins.kotlin.compose)
-    alias(libs.plugins.maven.publish)
-    alias(libs.plugins.dokka)
```
VelikovPetar, line 10: Just to double-check: removing `dokka` won't break the publishing of the GH pages right? (it should already by applied by the shared plugin)

scripts/local-release.sh
```diff
@@ -1,193 +0,0 @@
-#!/bin/bash
```
VelikovPetar, line 1: I am thinking whether this would be useful to keep (`mavenLocal` publish). Sometimes it is useful to be able to create a local release (ex. to test some compatibility with the other product SDKs). I see that we would need to update this file, but maybe it can be useful to keep it.

Edit: After a bit more consideration, I think it is not needed - let's delete it.

stream-chat-android-ai-assistant/build.gradle.kts
gpunto: Yes, exactly!

scripts/local-release.sh
gpunto: To publish locally we can now use this command
```
ORG_GRADLE_PROJECT_version=local-test ./gradlew publishToMavenLocal
```

I usually pass `local-test` as done above, but we can pass whatever we want as a version there and it will be picked up. Should I update the script accordingly to keep the logic to compute the version?

gpunto: Ah, I hadn't see the edit appear when responding 😅

PR #6130, opened 2026-02-04
Attachment picker components API update

andremion:
### 🎯 Goal

Simplify and modernize the Attachment Picker API.

The existing attachment picker used a factory pattern with 7 factory classes (`AttachmentsPickerTabFactory` + 6 implementations), scattered configuration, and a sealed class hierarchy that limited extensibility. This PR introduces a cleaner, more intuitive API that:

- Reduces complexity by ~70% (~1,400 lines removed vs ~600 added)
- Centralizes configuration in a single `AttachmentPickerConfig` data class
- Replaces the factory pattern with direct `ChatComponentFactory` composable overrides
- Uses typed mode classes with meaningful configuration properties
- Removes deprecated/unused code and improves KDoc documentation

### 🛠 Implementation details

**New API Structure:**

1. **`AttachmentPickerMode` interface** - Replaces the `AttachmentsPickerMode` sealed class
   - `GalleryPickerMode(allowMultipleSelection, mediaType)` - Pick images/videos from gallery
   - `FilePickerMode(allowMultipleSelection)` - Pick files from storage
   - `CameraPickerMode(captureMode)` - Capture photos/videos
   - `PollPickerMode(autoShowCreateDialog)` - Create polls
   - `CommandPickerMode` - Select commands

2. **`AttachmentPickerConfig`** - Centralized picker configuration
   - `useSystemPicker: Boolean` - Toggle between in-app and system picker
   - `modes: List<AttachmentPickerMode>` - Configure available picker modes

3. **`ChatComponentFactory` extensions** - New factory methods for customization:
   - `AttachmentPickerMenu()`, `AttachmentPicker()`, `AttachmentTypePicker()`
   - `AttachmentPickerContent()`, `AttachmentMediaPicker()`, `AttachmentFilePicker()`
   - `AttachmentCameraPicker()`, `AttachmentPollPicker()`, `AttachmentCommandPicker()`
   - `AttachmentSystemPicker()`, `AttachmentTypeSystemPicker()`

**Removed:**
- `AttachmentsPickerTabFactory` interface and all implementations
- `AttachmentsPickerTabFactories` static factory methods
- `AttachmentsProcessingViewModel` from factory package (moved to viewmodel package)
- `CaptureMediaLauncher.kt`
- Legacy `AttachmentsPickerMode` sealed class
- `ChatTheme.useDefaultSystemMediaPicker` parameter and property
- `ChatComponentFactory.AttachmentsPickerSendButton()`
- `DefaultAttachmentsPickerSendButton()`
- `SystemAttachmentsPickerConfig`

**Refactored:**
- Permission-related components moved to dedicated `permission` sub-package
- `AttachmentsPickerViewModel` updated to use nullable `pickerMode` for proper initialization
- Improved permission state observation to trigger callbacks only on actual state changes
- Comprehensive KDoc updates for all public attachment picker components

**Permissions:**
- Declared permissions in the common module are now declared in the samples only, so that customer apps won't contain those permissions by default, making the system picker the default option.

### 🎨 UI Changes

No visual changes - this is an API refactoring. The attachment picker UI remains identical.

### 🧪 Testing

**Manual Testing:**
1. Open the Compose sample app
2. Navigate to a channel and tap the attachment button
3. Verify all picker tabs work correctly (Gallery, Files, Camera, Poll)
4. Test system picker mode by setting `useSystemPicker = true` in `AttachmentPickerConfig`
5. Verify location sharing works in the sample (updated for new API)

**Automated Testing:**
- Unit tests updated for `RequiredPermission` logic
- Removed obsolete factory tests (`AttachmentsPickerTabFactoriesTest`, `AttachmentsPickerPollTabFactoryTest`, `AttachmentsProcessingViewModelTest`)
- Paparazzi snapshots updated

### 🎉 GIF

KISS

## Summary by CodeRabbit

* **New Features**
  * Location sharing integrated into the composer.
  * Redesigned, mode-based attachment picker with unified menu (gallery, files, camera, polls, commands) and improved single/multi-select behavior.
  * Composer exposes an attachments button to toggle the picker; picker coordinates keyboard/auto-dismiss flows.

* **Breaking Changes**
  * Simplified attachment picker configuration and UI; legacy tab-factory customization removed.
  * Permission prompts updated for modern media access.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
```diff
@@ -85,33 +104,36 @@ public fun AttachmentPickerMenu(
...
-                attachmentsPickerViewModel.changeAttachmentState(showAttachments = false)
-                composerViewModel.addSelectedAttachments(attachments)
-            },
-            onAttachmentPickerAction = { action ->
-                when (action) {
-                    is AttachmentPickerPollCreation -> {
-                        attachmentsPickerViewModel.changeAttachmentState(showAttachments = false)
-                        composerViewModel.createPoll(action.pollConfig)
+        Box(modifier = Modifier.height(menuHeight)) {
+            ChatTheme.componentFactory.AttachmentPicker(
```
gpunto: Could we pass the modifier to `AttachmentPicker` instead of wrapping it into a box?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentSystemPicker.kt
```diff
@@ -85,12 +84,16 @@ internal fun AttachmentSystemPicker(
     onAttachmentsSubmitted: (List<AttachmentMetaData>) -> Unit = {},
 ) {
     val context = LocalContext.current
+    val pickerModes = ChatTheme.attachmentPickerConfig.modes
 
-    val processingViewModel = viewModel<AttachmentsProcessingViewModel>(
-        factory = AttachmentsProcessingViewModelFactory(StorageHelperWrapper(context.applicationContext)),
+    val processingViewModel = viewModel<AttachmentProcessingViewModel>(
+        factory = AttachmentProcessingViewModelFactory(StorageHelperWrapper(context.applicationContext)),
```
gpunto: Should we `remember(context.applicationContext) {}` around the factory?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerContent.kt
```diff
@@ -18,58 +18,61 @@ package io.getstream.chat.android.compose.ui.messages.attachments
...
+        is CommandPickerMode -> ChatTheme.componentFactory.AttachmentCommandPicker(
+            pickerMode = pickerMode,
             commands = commands,
             onAttachmentPickerAction = onAttachmentPickerAction,
         )
 
-        is System -> TODO()
-        is CustomPickerMode -> TODO()
+        // Custom modes are handled by overriding AttachmentPickerContent in ChatComponentFactory
+        else -> Unit
```
gpunto, line 76: Is it feasible to have a default `AttachmentCustomPicker` where we pass modes that we don't handle? So that customers that just want to add custom attachments don't need to care about the default `when` cases.
```suggestion
        else -> ChatTheme.componentFactory.AttachmentCustomPicker(...)
```

Originally [it was proposed](https://www.notion.so/stream-wiki/Message-Composer-2ea6a5d7f9f680eda0a4cb6261a2840e) with a question mark, not sure what other SDKs ended up doing.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentTypePicker.kt
```diff
@@ -75,37 +73,35 @@ internal fun AttachmentTypePicker(
...
+            AttachmentPickerModeInfos[mode::class]?.let { typeInfo ->
                 AttachmentPickerToggleButton(
                     pickerTypeInfo = typeInfo,
                     isSelected = isSelected,
-                    onClick = { onPickerTypeClick(index, attachmentsPickerMode) },
+                    onClick = { onModeSelected(mode) },
                 )
             }
         }
+        trailingContent()
```
gpunto, line 88: With this, customers are losing the ability to display custom tabs before/among default ones by just passing a different list of modes, right? Well, TBF they can just override `AttachmentTypePicker` if they want and it's very little code. Is that the idea?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
andremion: I did that beforehand. Then I refactored to remove the modifier from the component factory declaration.
I've been thinking whether we should have modifiers in the factories or not...
In that case, the behavior of changing the height is part of the menu, not of the component itself...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentSystemPicker.kt
andremion: Good catch 👍🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerContent.kt
andremion: That was part of the legacy approach, where extra types were limited due to the sealed class (instead of an interface) and the absence of component factories.
In the sample, there's an example of how to add extra options (the shared location case) without caring about the default ones.
Let me know what you think

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentTypePicker.kt
andremion: Yes. Supporting all possible variants is hard.
Currently, we also support changing the order (or removing some) of the default ones without touching the UI (by using the `AttachmentPickerConfig`).
The idea is to provide minimal support for the most common cases. If they want to customize further, they can override the `AttachmentTypePicker `

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerContent.kt
gpunto: Right, they can just call super 👍

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
gpunto: I'd say that given that the modifier is created by the menu, the behavior of deciding the height is still responsibility of the menu. The `AttachmentPicker` component doesn't know anything about that, it just accepts a modifier and uses it, as composables normally do.

What would change is that a customer overriding the `AttachmentPicker` can choose not to use the modifier even if we pass it, but then that's the customer's choice.

andremion:
@coderabbit full review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
andremion: I updated that, thanks! 🙌🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/attachments/AttachmentPickerItemState.kt
```diff
@@ -19,37 +19,44 @@ package io.getstream.chat.android.compose.state.messages.attachments
...
+     * Represents the selection state of an attachment picker item.
      */
     public sealed interface Selection {
 
         /**
-         * The selected state of the attachment.
+         * Indicates that the attachment is selected.
          *
-         * @param position The position of the selected attachment in the list.
+         * @param position The 1-based position of this attachment in the selection order.
```
VelikovPetar, line 44: Is this really 1-based?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/attachments/AttachmentPickerMode.kt
```diff
@@ -0,0 +1,124 @@
...
+ * Mode for picking files from device storage.
+ *
+ * When selected, displays a list of files from the device's storage including documents,
+ * audio files, and other file types. Requires storage permissions when using the in-app picker.
+ *
+ * @param allowMultipleSelection When `true`, users can select multiple files. When `false`,
+ * only single selection is allowed. Defaults to `true`.
+ */
+public data class FilePickerMode(
+    val allowMultipleSelection: Boolean = true,
```
VelikovPetar, line 60: Question about this: Is this applied to both the 'system' and the 'inline' picker? I am experiencing something strange when using the `system` picker: After selecting a single attachment, the picker is closed and only a single attachment is selected (probably not related to this PR, but I wanted to mention it):

https://github.com/user-attachments/assets/3442a08a-e111-4c06-bc12-30d5c64ca82f

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/attachments/AttachmentPickerConfig.kt
```diff
@@ -0,0 +1,77 @@
...
+ * @param useSystemPicker When `true` (default), uses the system's native file/media picker which does not require
+ * storage permissions. When `false`, uses the in-app picker which shows a grid of media files but requires
+ * storage permissions to be granted and declared in your manifest.
+ * @param modes The list of [AttachmentPickerMode] instances that define which attachment types are available.
+ * The order of modes determines the order of tabs in the picker. Defaults to gallery, files, camera, poll,
+ * and commands.
+ */
+public data class AttachmentPickerConfig(
+    val useSystemPicker: Boolean = true,
+    val modes: List<AttachmentPickerMode> = DefaultAttachmentPickerModes,
```
VelikovPetar, line 68: I have a small concern about how easy would be customise only specific parts of the configuration. For example, if a customer want's to only update the `FilePickerMode.allowMultipleSelection`, they would have to do:

```kotlin
val config = AttachmentPickerConfig(
    modes = listOf(
        GalleryPickerMode(),
        FilePickerMode(allowMultipleSelection = false),
        CameraPickerMode(),
        PollPickerMode(),
        CommandPickerMode,
    )
)
```

It is probably OK if we document this properly, but I wanted to raise this concern, and that we might need to expose some convenience for customising those single properties (maybe something like a DSL or similar).

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
```diff
@@ -85,11 +106,17 @@ public fun AttachmentPickerMenu(
...
     ) {
-        AttachmentPicker(
-            attachmentsPickerViewModel = attachmentsPickerViewModel,
+        ChatTheme.componentFactory.AttachmentPicker(
             modifier = Modifier.height(menuHeight),
+            attachmentsPickerViewModel = attachmentsPickerViewModel,
+            messageMode = messageMode,
             onAttachmentItemSelected = { attachmentItem ->
-                attachmentsPickerViewModel.changeSelectedAttachments(attachmentItem)
+                val allowMultipleSelection = when (val mode = attachmentsPickerViewModel.pickerMode) {
```
VelikovPetar: I see this code for allowMultipleSelection duplicated several times - would it make sense to consolidate it in a read-only property or a separate method?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentSystemPicker.kt
```diff
@@ -190,9 +199,13 @@ internal fun AttachmentSystemPicker(
         }
     }
 
-    if (showCommandsPickerDialog) {
+    val commandPickerMode = remember(pickerModes) {
```
VelikovPetar, line 203: One thing to verify: If the channel doesn't support commands, will this action button be shown?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentTypePicker.kt
```diff
@@ -174,50 +169,55 @@ private fun AttachmentPickerButton(
...
+private data class AttachmentPickerModeInfo(
+    @get:DrawableRes val icon: Int,
     @get:StringRes val contentDescription: Int,
     val testTag: String,
 )
 
-private val AttachmentPickerTypeInfos = mapOf(
-    Images to AttachmentPickerTypeInfo(
+private val AttachmentPickerModeInfos = mapOf(
+    GalleryPickerMode::class to AttachmentPickerModeInfo(
```
VelikovPetar, line 196: I assume it is intentional that we don't allow customisations of these icons?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -3573,4 +3565,301 @@ public interface ChatComponentFactory {
...
+     *
+     * Override this to customize the picker container behavior, including animations,
+     * keyboard coordination, and composer integration.
+     *
+     * @param attachmentsPickerViewModel Controls picker visibility and manages attachment state.
+     * @param composerViewModel Receives selected attachments and handles poll/command actions.
+     */
+    @Composable
+    public fun AttachmentPickerMenu(
+        attachmentsPickerViewModel: AttachmentsPickerViewModel,
```
VelikovPetar, line 3580: Just one thing to check: Until now, these components were stateless. Is it OK to start introducing stateful components to the Factory?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/attachments/AttachmentPickerItemState.kt
andremion: Yes. That's the number we see when we start selecting...
Screenshot_20260206_125837

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/attachments/AttachmentPickerMode.kt
andremion: That's the normal behavior of the system file picker. You have to click and hold in order to select more items.
when allowMultipleSelection is false, you cannot do multiple selections at once.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/attachments/AttachmentPickerConfig.kt
andremion: Good point. You're right that the current API requires recreating the entire list just to modify a single property.
However, the current list-based approach has a clear advantage: the order of modes in the list directly defines the tab order in the UI. This is explicit and intuitive and would just require a few lines of code. A DSL would make ordering less obvious (relying on declaration order or requiring an explicit order property).

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
andremion: There are only two places, but I can extract it in a function...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentSystemPicker.kt
andremion: Good point! Let me verify...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/messages/attachments/AttachmentPickerMode.kt
VelikovPetar: Ohhh I never realised this. Thanks for clarifying!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
VelikovPetar: I leave it up to you. If you think it is more clear this way, I am perfectly fine with it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentSystemPicker.kt
andremion: I will add a verification by channel.config.command.isNotEmpty

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentTypePicker.kt
andremion: Nah. With this API review, we want to prevent exposing and maintaining so many different configurations and encourage customers to use the factories for UI customizations.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: Good point!
We should avoid, yes.
In that particular case, it was due to reusing the legacy view models to avoid an even bigger breaking change.
Do you see any issue with that? What would you suggest?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentTypePicker.kt
VelikovPetar: Excellent!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
andremion: I moved to a shared internal function 👍🏻

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
VelikovPetar: I am not sure to be honest, in this case it is also probably useful for integrators to have direct access to the VM, so they can use the public methods. I am worried that if we expose one such component, there will be requests for exposing other stateful components. 

@gpunto What's your opinion on this?

gpunto: Tbh I didn't realize we were avoiding that. Actually I was going to add `AudioRecordAttachmentContent` [to the factory](https://github.com/GetStream/stream-chat-android/blob/redesign/message-restructure-all/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt#L2897) (the PR isn't open yet), and it internally creates a VM, but it's not part of the function's signature.

Is the problem that we want to completely avoid VMs in components we expose in the factory, or that they shouldn't accept VM as parameters?

In my example above, `AudioRecordAttachmentContent` is transitively used by a message, so even if today the message item component seems "stateless" from its signature, it's actually stateful at the very minimum because it has the audio player VM somewhere inside.

VelikovPetar: That is also a good point. If you guys think this won't cause any issues, then let's do it like this, I wouldn't block this PR.

andremion: We can revisit this later, before releasing v7 👍🏻

PR #6133, opened 2026-02-06
Use component factory when rendering message attachments

gpunto:
### Goal

This PR finishes the restructuring of the message content, where instead of using attachment factories, we just delegate to the component factory.

### Implementation

- Remove `MessageAttachmentsContent` (which was using the attachment factories) and directly call the component factory for the remaining attachment types: links, recordings, unknown/custom
- Removed `content` parameter from `FileAttachmentFactory` and updated examples
- Fix image links appearing among media attachments

### 🎨 UI Changes

There should be no UI changes here. The redesign of audio & file attachments is coming in the next PR.

### Testing

Can be tested in the sample by checking messages with different attachments

## Summary by CodeRabbit

* **New Features**
  * Added new attachment content customization APIs in ChatComponentFactory for audio recordings, files, links, and custom attachments.
  * Enhanced link attachment handling with structured click data callback instead of context/URL pair.

* **Breaking Changes**
  * Removed MessageAttachmentsContent composable; use component factory methods instead.
  * Removed content property from AttachmentFactory; customize via new ChatComponentFactory overrides.
  * Media attachment previews now exclude link-type attachments.

* **Improvements**
  * Simplified attachment factory APIs by removing explicit content parameters.
  * Consolidated link attachment overloads for a cleaner public interface.

### DIFF
--- a/stream-chat-android-ui-guides/src/main/java/io/getstream/chat/android/guides/catalog/compose/customattachments/CustomChatComponentFactory.kt
+++ b/stream-chat-android-ui-guides/src/main/java/io/getstream/chat/android/guides/catalog/compose/customattachments/CustomChatComponentFactory.kt
@@ -43,7 +43,7 @@ import io.getstream.chat.android.guides.R
 import io.getstream.chat.android.models.Attachment
 
 /**
- * A custom [AttachmentFactory] that adds support for date attachments.
+ * A custom [ChatComponentFactory] that adds support for date attachments.
  */
 object CustomChatComponentFactory : ChatComponentFactory {
     @Composable


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
```diff
@@ -174,11 +174,9 @@ internal fun DefaultMessageContent(
         val info = message.rememberMessageInfo()
 
         if (info.hasUploads) {
-            MessageAttachmentsContent(
-                message = message,
-                currentUser = currentUser,
-                onLongItemClick = onLongItemClick,
-                onMediaGalleryPreviewResult = onMediaGalleryPreviewResult,
+            FileUploadContent(
```
VelikovPetar, line 177: Just checking: This will eventually be delegated to the `ChatComponentFactory` as well?

gpunto: This is actually temporary and I'll remove it later, since we recently got new designs for the upload state. It will be handled by the specific attachments themselves like below instead of being a separate component

Screenshot 2026-02-09 at 10 46 22

VelikovPetar: Oh nice, I wasn't aware of this. Thanks for clarifying!

gpunto:
> I believe there's an issue with the play button when we click to preview a video. Not a blocker for this PR. It can be fixed later.
> 
>  Screen_recording_20260209_092339.webm


The video doesn't load for me, but I checked and I imagine you mean the play button is completely white, right? Good catch, I'll fix it in my next PR 👍 

Screenshot 2026-02-09 at 10 55 10

PR #6134, opened 2026-02-06
Fix the wrong voice recording label in the channel list item

### LINKED ISSUE #6045: Voice Message List Name incorrect
When we share a voice note in group from android the  name of message in the list is this

Image

and when we send the voice from IOS side the name of Voice note is some message ID i think, in IOS side the name is Voice Message but in android its some id when shared from IOS. 

Image

We are using getStream Version 6.24.0 but this issue is also in 6.30.0 we have updated the sdk version but the issue still persists. 

The issue is there for images as well.

andremion:
### Goal

Fix the wrong voice recording label in the channel list item. 
Fix #6045

### Implementation

Show `Voice Message` instead of the filename in the channel list item when the message contains an audio recording.

> [!NOTE]
> This is a quick-win. In the v7 release we might align the entire message preview formatter with compose and other platform SDKs 

### 🎨 UI Changes

| Before | After |
| --- | --- |
| img | img |

### Testing

- Open the channel list where some channels' last messages are an audio recording.

## Summary by CodeRabbit

* **New Features**
  * Audio messages now display with dedicated formatting that clearly identifies audio content. Audio-reply information appears in italics within message previews and thread discussions instead of standard text and attachment previews.


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/preview/internal/MessagePreviewView.kt
```diff
@@ -124,6 +126,9 @@ internal class MessagePreviewView : FrameLayout {
     }
 
     private fun formatMessagePreview(message: Message, currentUserMention: String?): CharSequence {
+        if (message.hasAudioRecording()) {
```
VelikovPetar, line 129: What do you think about placing this check in the `Message.getAttachmentsText()` method instead? Otherwise we are overriding all other attachments texts here.

andremion: I believe that's not what the compose SDK does
https://github.com/GetStream/stream-chat-android/blob/6e0a8e89ecff5bfbf48f1efddaf4a37b9b33f1e9/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MessagePreviewFormatter.kt#L187
Do you have a test scenario where this might be an issue?

VelikovPetar: Oh I see - then let's keep it aligned between the SDKs.

PR #6135, opened 2026-02-09
Set `QueryChannels` `message_limit` and `member_limit` defaults to `null` (server-driven)

VelikovPetar:
### Goal
We are setting the default `message_limit` and `member_limit` to `null` so that the request follows (a potentially dynamic) server-side default. 
Note: This only affects the defaults - if a customer overrides them, this change has no effect. 
Note: See related discussion linked in the [ticket](https://linear.app/stream/issue/AND-1056/make-the-messages-and-members-limit-in-channel-list-query-optional).

### Implementation
Change the `QueryChannels` `message_limit` and `member_limit` defaults to `null`.

### 🎨 UI Changes
_NA_

### Testing
1. Open a channel list
2. Observe the logs
3. The `QueryChannels` request should not pass `message_limit`/`member_limit` in its body.

## Summary by CodeRabbit

* **Refactor**
  * Updated QueryChannelsRequest and related ViewModels to support nullable message and member limit parameters. When limits are omitted, server-side defaults are applied instead of client-side fixed defaults.
  * Adjusted ChannelListViewModelFactory and corresponding builders to align with the new optional parameter approach for improved configuration flexibility.

### DIFF
--- a/stream-chat-android-compose-sample/src/e2e/java/io/getstream/chat/android/compose/sample/ui/JwtTestActivity.kt
+++ b/stream-chat-android-compose-sample/src/e2e/java/io/getstream/chat/android/compose/sample/ui/JwtTestActivity.kt
@@ -188,9 +188,9 @@ class JwtTestActivity : AppCompatActivity() {
 
             OkHttpClient().newCall(request).execute().use { response ->
                 if (!response.isSuccessful) {
-                    throw Exception("Failed to fetch token: ${response.code}")
+                    return@withContext ""
                 }
-                response.body?.string() ?: throw Exception("Empty response body")
+                response.body?.string() ?: ""
             }
         }
     }


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelsRequest.kt
```diff
@@ -28,16 +28,16 @@ import io.getstream.chat.android.models.querysort.QuerySorter
...
  */
 public data class QueryChannelsRequest(
     public val filter: FilterObject,
     public var offset: Int = 0,
     public var limit: Int,
     public val querySort: QuerySorter<Channel> = QuerySortByField(),
-    public var messageLimit: Int = 0,
-    public var memberLimit: Int = 1,
+    public var messageLimit: Int? = null,
+    public var memberLimit: Int? = null,
```
gpunto, line 40: From reading the thread linked in the ticket, I understand we accept to ship this breaking change, but shouldn't we also mark the PR with the corresponding label so it shows up in the changelog?

VelikovPetar: Let me double check and align this with iOS

VelikovPetar: Update: Had a sync with iOS, we should release this in a `minor` version, and declare it as a `Change` - we won't consider this breaking. But since we don't support `pr:change` label, I think `pr:improvement` is the closest alternative.

gpunto: Ok, that's fine by me (but keep in mind compilation will fail for consumers if they were accessing these properties)

gpunto (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelsRequest.kt
VelikovPetar: I think that this is a lower risk - even though it is possible, I think that that probability is low, and the drawbacks are small. A bit higher risk would be if they depend on the defaults in their integration, but this is something that Is already anticipated.

gpunto: > I think that this is a lower risk - even though it is possible, I think that that probability is low, and the drawbacks are small

Definitely!

PR #6136, opened 2026-02-09
Fix token generation failure test flakiness

gpunto:
### Goal

`test_tokenGenerationFails` intermittently fails in CI. That's because the token generator used in the test app throws when the backend response is unsuccessful, but that's incorrect. `TokenProvider.loadToken` is supposed to return an empty string if the token can't be generated.

### Implementation

Return empty string instead of throwing

### 🎨 UI Changes

None

### Testing

Running `test_tokenGenerationFails` should not fail

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling in token authentication processes. The system now gracefully manages failed requests and empty responses by returning safe defaults instead of throwing exceptions, enhancing application stability and reliability.


gpunto:
@coderabbitai review

VelikovPetar (approved):

PR #6137, opened 2026-02-09
Update audio recording and file attachments design

gpunto:
### Goal

Update audio recording and file attachments design. For audio recording, this also entailed changing the playback speed logic to allow for 1) setting different speed per attachment and 2) changing speed while the audio is not playing

### Implementation

Updated designs and audio speed controlling logic. Now it's handled per audio hash instead of being tied to a single aucio.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_20260209_111155 | Screenshot_20260209_145551 |

### Testing

Can be tested in the sample by checking the designs and that the audio recording 

## Summary by CodeRabbit

* **New Features**
  * Per-track audio speed control: each audio file now maintains its own playback speed setting independently.

* **Bug Fixes & Improvements**
  * Enhanced audio attachment UI with improved layout and styling.
  * Improved file attachment rendering with better full-size support.
  * Refined playback timer display with updated visual styling.

* **Theme Updates**
  * Simplified messaging theme system for better consistency.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
@@ -417,7 +417,6 @@ fun UserRobot.assertFile(isDisplayed: Boolean, count: Int = 1): UserRobot {
     if (isDisplayed) {
         assertEquals(count, Message.fileName.waitForCount(count).size)
         assertEquals(count, Message.fileSize.findObjects().size)
-        assertEquals(count, Message.fileDownloadButton.findObjects().size)
         assertEquals(count, Message.fileImage.waitForCount(count).size)
         if (count > 1) {
             assertTrue(Message.columnWithMultipleFileAttachments.isDisplayed())
@@ -426,7 +425,6 @@ fun UserRobot.assertFile(isDisplayed: Boolean, count: Int = 1): UserRobot {
         assertFalse(Message.fileName.waitToDisappear().isDisplayed())
         assertFalse(Message.fileSize.isDisplayed())
         assertFalse(Message.fileImage.isDisplayed())
-        assertFalse(Message.fileDownloadButton.isDisplayed())
     }
     return this
 }


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/StreamColors.kt
```diff
@@ -476,6 +478,7 @@ internal object StreamPrimitiveColors {
     val slate100 = Color(0xFFF2F4F6)
     val slate150 = Color(0xFFD5DBE1)
     val slate200 = Color(0xFFE2E6EA)
+    val slate300 = Color(0xFFA3ACBA)
     val slate400 = Color(0xFFB8BEC4)
```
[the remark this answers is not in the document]
gpunto: The colors changed in the designs between the addition of slate300 and slate400. I'll do a pass to refresh them in a separate PR at a later point when they're reasonably stable, to avoid having to re-record all screenshots

gpunto:
@coderabbitai review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/AudioPlayerState.kt
```diff
@@ -27,13 +27,15 @@ import io.getstream.chat.android.models.Attachment
...
  * @property seekTo The seekTo state of the audio player.
+ * @property speeds The playback speeds state of the audio player (keyed by audio hash).
  * @property getRecordingUri A function that returns the URI of the recording.
  */
 @Immutable
 @InternalStreamChatApi
 public data class AudioPlayerState(
     val current: CurrentAudioState = CurrentAudioState(),
     val seekTo: IntFloatMap = intFloatMapOf(),
+    val speeds: IntFloatMap = intFloatMapOf(),
```
VelikovPetar, line 38: This is interesting, do we always get the same `AudioPlayerState` instance for each `audio recording` attachment in the message list? (nothing to change, just to double check)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileAttachmentContent.kt
```diff
@@ -137,39 +135,24 @@ public fun FileAttachmentItem(
...
-            FileAttachmentImage(
-                attachment = attachment,
-                isMine = isMine,
-            )
-            FileAttachmentDescription(
-                attachment = attachment,
-                isMine = isMine,
-                showFileSize = showFileSize,
-            )
-            FileAttachmentDownloadIcon(
```
VelikovPetar, line 165: Is the download button no longer supported, or will it be handled differently? (if no longer supported, I think we would be able to clean up the downloading logic as well)

VelikovPetar (commented):
Left couple of questions but looks pretty good!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileAttachmentContent.kt
gpunto: Actually that's a great question. I just left a comment on Figma to clarify it, let's see.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/AudioPlayerState.kt
gpunto: Yes, exactly. We have a single `AudioPlayerViewModel` that exposes one `StateFlow<AudioPlayerState>`

VelikovPetar: Got it, thanks for confirming!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/FileAttachmentContent.kt
gpunto: So [the behavior will be](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y?node-id=4030-142283&m=dev#1626115163):
- Preview files we can preview -> and show a share button that delegates to system sharing
- If we can't preview it -> download + ask the system to open it

If it's fine with you, I'll do this in a separate task

VelikovPetar: yes of course, it sounds like a completely separate task!
Thanks for checking!

PR #6139, opened 2026-02-09
Decouple message composer buttons from BasicTextField to prevent unintended focus

andremion:
### Goal

Fix message composer buttons (send, record) inadvertently triggering the `BasicTextField` focus and showing the keyboard when tapped. This happened because the trailing content lived inside `BasicTextField`'s `decorationBox`, so any pointer event in that area was claimed by the text field. Additionally, introduce a `MessageComposerInputCenterContent` factory to give consumers a dedicated customization point for the text input area.

> [!NOTE]
> This is a preliminary task of the audio recording design update

### Implementation

- Decoupled the visual container (border, background, shadow) from `BasicTextField`. `MessageInput` is now a pure layout shell: an outer `Column` (owning the visual styling) with a `Row` containing three sibling slots — `leadingContent`, `centerContent`, `trailingContent`. `BasicTextField` is confined to the center slot, so pointer events on the send/record buttons never reach the text field's focus handler.
- Added `ChatComponentFactory.MessageComposerInputCenterContent(state, onValueChange, modifier)` — an overridable factory for the text input area. The default implementation (`MessageComposerInputCenterContent.kt`) manages the `String`↔`TextFieldValue` bridge, visual transformation (links, mentions), placeholder text, and accessibility semantics.
- Extracted `MessageComposerInputTrailingContent` from `MessageComposerDefaults.kt` into its own file for consistency with the other content slot files.
- Removed the `label` parameter from `MessageInput`, `DefaultMessageComposerInput`, `ChatComponentFactory.MessageComposerInput`, and both `MessageComposer` overloads. The placeholder is now self-contained inside the center content factory, reducing parameter threading across the call chain.
- Removed `ChatComponentFactory.MessageComposerLabel()` and `DefaultComposerLabel()` — the placeholder is now part of the center content implementation.

### 🎨 UI Changes

No visual changes. The composer looks and behaves identically; only the internal layout hierarchy changed.

### Testing

The composer should behave as it is now. This is a preliminary task of the audio recording design update

## Release Notes

* **Refactor**
  * Restructured message input and composer components for improved flexibility and customization
  * Replaced message label with a center content composable parameter, enabling better control over input area layout
  * Optimized message composer internal composition and parameter structure
  * Updated component factory interfaces to align with new content-based composition approach

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt
@@ -37,6 +37,8 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
+import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.OutlinedTextFieldDefaults
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
@@ -44,6 +46,7 @@ import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.unit.dp
 import androidx.core.net.toUri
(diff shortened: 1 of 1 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/composer/MessageInput.kt
```diff
@@ -85,111 +56,84 @@ import io.getstream.chat.android.ui.common.state.messages.composer.MessageCompos
...
+        )
+    },
     trailingContent: @Composable RowScope.() -> Unit = {
         ChatTheme.componentFactory.MessageComposerInputTrailingContent(
             state = messageComposerState,
-            onSendClick = onSendClick,
             recordingActions = recordingActions,
+            onSendClick = onSendClick,
         )
     },
```
gpunto, line 100: Not necessarily tied to this PR and definitely not blocking, more as a general note:
Shouldn't we actually remove the slots from components signatures when they have a corresponding factory function? They can be overridden through the factory, and I believe the idea is that that should be the way to customize components

andremion: Definitely agree.
I will do it in another PR.
Thanks!

PR #6141, opened 2026-02-10
Introduce compact link attachment design

gpunto:
### Goal

Introduce compact link attachment to be displayed when a message has other attachments in addition to the link.

### Implementation

Implement new component and display it if the link is **not** the only attachment.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_20260210_133236 | after |

### Testing

Can be checked in the sample

## Summary by CodeRabbit

* **New Features**
  * Introduced adaptive layout for link attachments: single attachments display in full-size format with expanded preview, while multiple attachments render in a compact, space-efficient layout.
  * Enhanced link attachment previews with cropped image rendering for improved visual presentation.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/LinkAttachmentContent.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/LinkAttachmentContent.kt
@@ -30,6 +30,7 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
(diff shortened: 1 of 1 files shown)


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/LinkAttachmentContent.kt
```diff
@@ -106,52 +103,137 @@ public fun LinkAttachmentContent(
...
-private fun LinkAttachmentImagePreview(imageUrl: String) {
-    val contentScale = ContentScale.FillWidth
+private fun CompactLinkAttachmentContent(
+    modifier: Modifier,
+    attachment: Attachment,
+    previewUrl: String,
+    textColor: Color,
+) {
+    Row(
+        modifier = modifier.padding(
```
VelikovPetar, line 203: I think that here (or maybe in the `baseModifier` higher in this class) we should add a `.fillMaxWidth()` modifier - otherwise short links will not fill the whole message width (unless it is intentional):

Image

gpunto: Good catch, done!

PR #6143, opened 2026-02-11
Update naming and structure of the main message factory functions

gpunto:
### Goal

We're restructuring the main message factory functions according to [the new naming](https://www.notion.so/stream-wiki/Message-View-2ea6a5d7f9f6808d826dff503e9f7d1d) that was aligned across platforms.

### Implementation

Most notable changes:
- What we used to call "message container" is now "message item" and vice versa
  - I've renamed the files this way: `MessageItem -> DefaultMessageContainer` and `MessageContainer -> DefaultMessageItem`. I did this for the sole purpose of avoiding a misleading git diff where it would seem the implementations changed. They can be renamed in a separate PR
- We're dropping header/footer naming in favor of top/bottom
- Composable slots in function signatures were inlined in the body: customization should happen through ChatComponentFactory instead
- Remove `DefaultMessageContainer` composable, which was just calling `MessageContainer`

### 🎨 UI Changes

No changes.

### Testing

There should be no functional changes, so the message should work as before.

## Release Notes

* **Refactor**
  * Reorganized message composition API with renamed components for improved clarity: MessageListItemContainer → MessageItem, MessageListItemContent → MessageContainer, MessageItemHeaderContent → MessageTop, MessageItemFooterContent → MessageBottom, MessageItemLeadingContent → MessageAuthor, MessageItemCenterContent → MessageContent
  * Simplified message component implementation through consolidated component factory approach
  * Updated reaction component naming from MessageReactionList to MessageReactions with enhanced onClick parameter support


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
```diff
@@ -271,30 +209,61 @@ public fun MessageItem(
                     .then(clickModifier)
                     .testTag("Stream_MessageCell"),
             ) {
-                leadingContent(messageItem)
+                with(ChatTheme.componentFactory) {
+                    MessageAuthor(
```
andremion, line 213: For consistency with the trailing, wouldn't it be better keep `MessageItemLeadingContent`?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
```diff
@@ -127,16 +127,11 @@ import kotlin.math.roundToInt
...
- * @param headerContent The content shown at the top of a message list item.
- * @param centerContent The content shown at the center of a message list item.
- * @param footerContent The content shown at the bottom of a message list item.
- * @param trailingContent The content shown at the end of a message list item.
  */
 @Suppress("LongMethod")
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
-public fun MessageItem(
+public fun MessageContainer(
```
andremion, line 134: Maybe the file can also be renamed to `MessageContainer` ?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
```diff
@@ -75,17 +77,9 @@ import io.getstream.chat.android.ui.common.state.messages.poll.PollSelectionType
...
- * @param systemMessageContent Composable that represents system messages.
- * @param messageItemContent Composable that represents regular messages.
- * @param typingIndicatorContent Composable that represents a typing indicator.
- * @param emptyThreadPlaceholderItemContent Composable that represents placeholders inside of an empty thread.
- * This content is disabled by default and can be enabled via [MessagesViewModelFactory.showDateSeparatorInEmptyThread]
- * or [MessageListController.showDateSeparatorInEmptyThread].
  */
 @Composable
-public fun LazyItemScope.MessageContainer(
+public fun LazyItemScope.MessageItem(
```
andremion, line 82: I'm confused.
The legacy `MessageItem` is now `MessageContainer`.
This legacy `MessageContainer` is now `MessageItem` ?

andremion: I believe that's the intention. I just read the PR description again 😅

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
gpunto: For now I renamed it to `DefaultMessageContainer` to avoid a misleading git diff because MessageContainer already exists and it would seem as if we're changing its contents, while it's just a rename. Same for `MessageItem`. I could do it in a separate PR so we have a separate commit with the rename

WDYT?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageItem.kt
gpunto: Yeah, unfortunately the new naming is swapped vs what we had.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
gpunto: I renamed it accordingly to the [new naming](https://www.notion.so/stream-wiki/Message-View-2ea6a5d7f9f6808d826dff503e9f7d1d). Actually I should've renamed `MessageItemTrailingContent` too, to `MessageSpacer`

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
andremion: I would do it in this PR, since the renames of the main functions are done here. But I will let you decide.

gpunto: No problem! I changed it now, especially given that you already had a chance to review the changes before the renaming

PR #6144, opened 2026-02-11
Introduce `ReactionIcon` and `ReactionToggle` components

gpunto:
### Goal

Introduce `ReactionIcon` and `ReactionToggle` components. They will be used later for displaying reactions in the SDK instead of the current implementations.
 
### Implementation

They are simple components rendering an emoji (`ReactionIcon`) + a checked state (`ReactionToggle`). We only support emojis as that's the direction we're taking going forward as emojis is the standard in chat apps. If customers want image-based reactions, they will be able to replace `ReactionIcon`.

### 🎨 UI Changes

[Figma](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y/Chat-SDK-Design-System?node-id=132-2034&m=dev)

Screenshot 2026-02-11 at 12 14 00

[Figma](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y/Chat-SDK-Design-System?node-id=2201-25686&m=dev)

Screenshot 2026-02-11 at 12 11 39

### Testing

They aren't in use yet, so they can only be checked in previews.

## Summary by CodeRabbit

* **New Features**
  * Added customizable reaction icon component for displaying emoji reactions across five size options (Small, Medium, Large, ExtraLarge, ExtraExtraLarge)
  * Added interactive reaction toggle component with three size options and visual feedback for checked/unchecked states
  * Extended theme factory to support new reaction components, ensuring design consistency

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactions/ReactionIcon.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactions/ReactionIcon.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.compose.ui.components.reactions
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import io.getstream.chat.android.compose.ui.theme.ChatTheme
+
+/**
+ * Component for rendering reaction emojis.
+ *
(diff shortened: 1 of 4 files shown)


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactions/ReactionIcon.kt
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * Component for rendering reaction emojis.
+ *
+ * @param size The size of the reaction icon.
+ * @param modifier Modifier for styling.
+ */
+@Composable
+internal fun ReactionIcon(
+    type: String,
+    emoji: String?,
```
VelikovPetar, line 41: Bit curious about the `type` / `emoji` params: I can see that `type` is not used, and `emoji` is optional - does it really make sense to be optional? Is there a case where we would need to call this composable with `emoji = null`?

gpunto: So my thinking was this:

In v7, we're dropping out of the box support for non-emoji reactions, so if we can't resolve an emoji, we could assume we're not going to display anything and not call these composables (`ReactionIcon`/`ReactionToggle`) at all. However, this makes it a bit more difficult for integrators to support non-emoji reactions if they want. They would need to replace every component that can render reactions instead of replacing just `ReactionIcon`. But maybe that's ok?

VelikovPetar: Hmm this is an interesting issue, not sure if we should also be considering backwards compatibility (assuming the case where users are both on versions which have integrated V6 and V7). But will every emoji reaction also have a type?

gpunto: Yes, `type` is the unique identifier for the reaction so it will always be there. Then, each type can map to an emoji: we're going to have a default mapping and integrators could provide their own (basically what we do today with `ReactionPushEmojiFactory`)

In theory, users on different majors should be able to see reactions properly, as long as either:
- all existing types an integrator uses have an emoji mapping in v7
- or integrators provide a custom ReactionIcon that can render things other than emojis

VelikovPetar: Ah OK, now I understand this better, thanks for the explanation! 
In that case I think my comment no longer makes sense, I will resolve this comment.

PR #6146, opened 2026-02-12
Refactor attachment picker to support cross-tab selection

andremion:
### Goal

Support cross-tab selection in the Compose attachment picker, where selecting a media attachment, switching to the files tab, and selecting a file would clear the previously selected media attachment. Users need to be able to select attachments from both tabs and have all selections preserved when sending a message.

### Implementation

**Separate state per tab:** Replaced the single `attachments` list with two private lists — `mediaAttachments` and `fileAttachments` — so selections survive tab switches. The existing `attachments` property is now a computed get/set that routes to the active tab based on `pickerMode`.

**Cross-tab selection sync:** Files that appear in both tabs (e.g. an image visible in the gallery and the files list) share selection state. Selecting in one tab marks the matching item in the other, matched by content URI. This is handled by:
- `applyCrossTabSelections` — when a tab loads, marks items that are already selected in the other tab.
- `syncChangesToOtherTab` — propagates real-time selection/deselection changes to the other tab.

**URI-based matching for removal:** `removeSelectedAttachment` now matches by content URI instead of the fragile `attachment.name`. The original MediaStore content URI is preserved in `Attachment.extraData` under `StorageHelperWrapper.EXTRA_SOURCE_URI` during the `AttachmentMetaData` → `Attachment` conversion.

**Deduplication:** `getSelectedAttachments` combines both tabs and deduplicates by URI to prevent the same file from being submitted twice.

### 🎨 UI Changes

No visual changes. The picker UI remains the same — selections are now correctly preserved across tab switches.

### Testing

1. Open a channel and tap the attachment (clip) button to open the picker
2. In the gallery tab, select one or more media items (images/videos)
3. Switch to the files tab — verify the media selections are still visible in the composer
4. Select one or more files
5. Switch back to the gallery tab — verify both media and file selections are preserved
6. If the same file appears in both tabs, verify it shows as selected in both and is only sent once
7. Remove an attachment from the composer — verify it is deselected in both tabs
8. Send the message — verify all selected attachments are included

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/AttachmentsPickerViewModel.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/AttachmentsPickerViewModel.kt
@@ -198,7 +198,7 @@ public class AttachmentsPickerViewModel(
     public fun getSelectedAttachments(): List<Attachment> {
         val allSelected = (mediaAttachments + fileAttachments)
             .filter(AttachmentPickerItemState::isSelected)
-            .deduplicateByUri()
+            .distinctBy { it.attachmentMetaData.uri }
             .map(AttachmentPickerItemState::attachmentMetaData)
         return storageHelper.getAttachmentsForUpload(allSelected)
     }
@@ -416,18 +416,6 @@ public class AttachmentsPickerViewModel(
                     }
                 },
             )
-        }
-
-        /**
-         * Removes duplicate items that share the same [URI][AttachmentMetaData.uri],
-         * keeping the first occurrence.
-         */
-        private fun List<AttachmentPickerItemState>.deduplicateByUri(): List<AttachmentPickerItemState> {
-            val seen = mutableSetOf<Uri>()
-            return filter { item ->
-                val uri = item.attachmentMetaData.uri
-                uri == null || seen.add(uri)
-            }
         }
     }
 }


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/AttachmentsPickerViewModel.kt
```diff
@@ -264,8 +244,190 @@ public class AttachmentsPickerViewModel(
...
+        /**
+         * Removes duplicate items that share the same [URI][AttachmentMetaData.uri],
+         * keeping the first occurrence.
+         */
+        private fun List<AttachmentPickerItemState>.deduplicateByUri(): List<AttachmentPickerItemState> {
+            val seen = mutableSetOf<Uri>()
+            return filter { item ->
+                val uri = item.attachmentMetaData.uri
+                uri == null || seen.add(uri)
+            }
```
gpunto: Could it be implemented with `distinctBy { it.attachmentMetaData.uri }`?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/AttachmentsPickerViewModel.kt
```diff
@@ -75,36 +77,67 @@ public class AttachmentsPickerViewModel(
...
+     * Items from the device gallery (images and videos).
      */
-    public var attachments: List<AttachmentPickerItemState> by mutableStateOf(emptyList())
+    private var mediaAttachments: List<AttachmentPickerItemState> by mutableStateOf(emptyList())
 
     /**
-     * Gives us information if we're showing the attachments picker or not.
+     * Items from the device file system.
+     */
+    private var fileAttachments: List<AttachmentPickerItemState> by mutableStateOf(emptyList())
```
gpunto, line 93: Instead of having two lists, could we have a map from identifiers (the Uri?) of the selected files/media to any metadata we need (like the selection position)? Then `attachments` are mapped accordingly on selection changes.

It could be something like like `val selectionState: MutableMap<Uri, Selection>`.

andremion: That's a good point.
The trade-off is that, with a separate Map<Uri, Selection>, we'd need to derive the attachments list on every read by merging raw items with the map, adding an extra mapping layer between the state and the UI.

andremion: The legacy code contained two list and I had removed them to a single attachment for simplicity.
With that cross-tab support, I had to rethink and realized that we would need those two.

andremion: Let me check again to see how that would look like with a single one again...

andremion: Let's merge it as is. We can rework on that later if that makes sense.

PR #6147, opened 2026-02-12
Update reactions to be emoji-only by default

gpunto:
### Goal

We're dropping out-of-the-box support for picture reactions in favor of using emojis by default, as it's common for many apps.

### Implementation

- Introduce `ReactionResolver`, to be used in the compose module instead of `ReactionPushEmojiFactory` and `ReactionIconFactory`. The XML module isn't changed.
- Use `ReactionIcon` & `ReactionToggle` to render the reactions as emojis instead of images

### 🎨 UI Changes

This PR contains minimal UI changes, i.e. those needed to move to displaying emojis. Actually updating the designs will come later.

| Before | After |
| --- | --- |
| Screenshot_20260212_152623 | Screenshot_20260212_152507 |
| Screenshot_20260212_152638 | Screenshot_20260212_152518 |
| Screenshot_20260212_152652 | Screenshot_20260212_152532 |

### Testing

You can check how reactions are displayed in the sample now.

## Summary by CodeRabbit

* **New Features**
  * Added a ReactionResolver with default implementation for emoji-based reactions.

* **Bug Fixes**
  * Improved reaction selection and ownership tracking via explicit isSelected/isMine flags.

* **Refactor**
  * Migrated from painter/drawable-based reaction icons to resolver/emoji-string rendering.
  * Simplified theme and component APIs by removing map-based reaction parameters and consolidating reaction config.


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/previewdata/PreviewReactionOptionData.kt
```diff
@@ -55,30 +26,26 @@ internal object PreviewReactionOptionData {
 
     @Composable
     fun reactionOption1() = ReactionOptionItemState(
-        painter = painterResource(R.drawable.stream_compose_ic_reaction_thumbs_up),
         type = "like",
-        emojiCode = null,
+        emojiCode = "👍",
```
andremion, line 30: I'm wondering whether we still need the `type` 🤔 
Shouldn't `emoji = "👍"` be enough?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ReactionResolver.kt
```diff
@@ -0,0 +1,86 @@
...
+    override fun emojiCode(type: String): String? = emojiMapping[type]
+
+    private companion object {
+        // Explicitly use linkedMapOf to make it clear that map iteration follows insertion order
+        private val defaultEmojiMapping: Map<String, String> = linkedMapOf(
+            ReactionDefaults.THUMBS_UP to "👍",
+            ReactionDefaults.LOVE to "❤️",
+            ReactionDefaults.LOL to "😂",
+            ReactionDefaults.WUT to "😮",
+            ReactionDefaults.THUMBS_DOWN to "👎",
```
andremion, line 83: Same here. Given that an emoji is just a string, do we still need type?
Just exercising here... Do we need this resolver at all? 🤔

gpunto: So the logic is that reactions are uniquely identified by the type, which is what we receive from the backend. Then, by default, we only render them as emojis, but we leave customers the option to display something else by overriding `ReactionIcon`. In that case, there could be types that don't map to an emoji. It also makes it easier for integrators who will have customers on both v6 and v7, as they can configure in v7 the same types they have in v6 and their customers will see reactions properly in both versions.

In the end, this is kinda similar to how Slack does reactions, i.e. with `:type:`, with the exception that we have a default bias for emojis.

We actually had a similar discussion [in here](https://getstream.slack.com/archives/C0A3JU6C6G3/p1770810659043649).

stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.components.reactionpicker_ReactionsPickerTest_Reaction_picker_with_more_reactions.png
VelikovPetar, line 1: I am not sure if this test will still be relevant (extended reactions menu), but we should probably either update it eventually, or maybe even delete it. (as it seems to be duplicated by `stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.components.reactionpicker_ReactionsPickerTest_Default_reaction_picker`)

gpunto: Good point. We'll still have the extended reactions menu and it will later show all the reactions we support, so I think we can remove one test

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/userreactions/UserReactionItemState.kt
```diff
@@ -16,18 +16,19 @@
...
  * @param type The string representation of the reaction.
+ * @param isMine Whether the reaction belongs to the current user.
+ * @param emojiCode The optional emoji code to be shown for the reaction.
  */
 public data class UserReactionItemState(
     public val user: User,
-    public val painter: Painter,
     public val type: String,
+    public val isMine: Boolean = false,
+    public val emojiCode: String? = null,
```
andremion, line 33: Given that the idea is to support mapping reactions to anything customers need, what about naming it as just `code`?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/ReactionResolver.kt
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public val supportedReactions: Set<String>
+
+    /**
+     * Returns the emoji code for the given reaction [type], or null if the type is not supported.
+     *
+     * @param type The reaction type (e.g. "like", "love", "haha").
+     * @return The emoji code for the reaction, or null if unsupported.
+     */
+    public fun emojiCode(type: String): String?
```
andremion, line 41: Perhaps we could simply name it `code` instead of `emijiCode` (or another non-emoji-biased name), as it could be mapped to anything the customer needs.

gpunto: I'm a bit unsure about that 🤔 As I imagined it, the main idea is not to map to anything generic, but that what's returned here is displayed as-is as a reaction in components and notifications (like before with `ReactionPushEmojiFactory` for notifications). If customers return another string, they'll see that text rendered in the UI where they'd expect a reaction.

Of course they should replace the `ReactionIcon` component for further customization, but at that point they don't really need to map `type` to a `code`, because type is already a unique identifier.

PR #6150, opened 2026-02-12
Consolidate audio recording components into the new message composer structure

andremion:
### Goal

Prepare the Compose audio recording components for the upcoming design update by consolidating them into a self-contained structure aligned with the new message composer.

### Implementation

- Unified all recording UI into a single internal `AudioRecordingButton` that manages the full recording lifecycle, removing the `audioRecordingContent` slot from `MessageComposer` and `MessageComposerAudioRecordingContent` from `ChatComponentFactory`
- Split the monolithic `DefaultMessageComposerRecordingContent` into focused files: `AudioRecordingButton`, `AudioRecordingContent`, `AudioRecordingGesture`, and `AudioRecordingPermission`
- Inlined `MessageInput` directly into `DefaultMessageComposerContent`
- Extracted `SnackbarPopup` as a reusable component
- Replaced hardcoded gesture thresholds with platform `viewConfiguration` values
- Added haptic feedback and spring animation to the recording interaction

### 🎨 UI Changes

_No visual changes intended — this is a structural refactor of the existing recording UI._

### Testing

- Long-press to record, drag up to lock, slide left to cancel

## Release Notes

* **Improvements**
  * Redesigned audio recording UI with enhanced gesture controls and streamlined permission handling.
  * Updated guidance text for recording clarity ("Hold to record. Release to send").
  * Added new theme color property for improved visual consistency.
  * Optimized notification display with improved snackbar implementation.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingPermission.kt
```diff
@@ -0,0 +1,186 @@
...
+                    Text(
+                        style = theme.buttonTextStyle,
+                        text = stringResource(id = R.string.stream_ui_message_composer_permissions_setting_button)
+                            .uppercase(),
+                    )
+                }
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
andremion: It will be updated in a later PR

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingGesture.kt
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/**
+ * Waits for the platform long-press timeout. Returns `true` if the user held long enough,
+ * `false` if they released early (tap).
+ */
+private suspend fun AwaitPointerEventScope.awaitHoldThreshold(
+    down: PointerInputChange,
+): Boolean {
+    val releasedEarly = withTimeoutOrNull(viewConfiguration.longPressTimeoutMillis) {
```
gpunto, line 96: Is this enough time for the use case? Do you know what threshold other SDKs use?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingGesture.kt
```diff
@@ -0,0 +1,161 @@
...
+        onShowHint()
+        return
+    }
+
+    recordingActions.onStartRecording(Offset.Zero)
+
+    val result = awaitDragResult(down, config, currentState, recordingActions.onHoldRecording)
+    when (result) {
+        DragResult.Released -> {
+            if (currentState() !is RecordingState.Locked) recordingActions.onSendRecording()
```
gpunto: Here we should respect the config from the theme (`AudioRecordingTheme.sendOnComplete`). Provided, [it's already a bug today](https://getstream.slack.com/archives/C08NJKNDEC9/p1769789495251049) so it can be addressed later if you want (and also align with iOS)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingPermission.kt
```diff
@@ -0,0 +1,186 @@
...
+import kotlinx.coroutines.delay
+
+/**
+ * Wrapper around Accompanist's [rememberPermissionState] that returns `null` in
+ * preview / Paparazzi environments (where there is no Activity context).
+ */
+@OptIn(ExperimentalPermissionsApi::class)
+@Composable
+internal fun rememberAudioRecordingPermission(): AudioRecordingPermission? {
+    if (LocalInspectionMode.current) return null
```
gpunto: Would it be incorrect to return the granted permission state in tests? Like `AudioRecordingPermission(PermissionStatus.Granted, {}, {})`

We avoid propagating the nullability, which also atm makes `gateRecording` a bit unclear. It correctly treats null like granted as it means inspection mode, but it's not clear from `gateRecording` that that's the reason.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingGesture.kt
andremion: That would be the recommended threshold because it's a long-press action.
The UX of the legacy code was not good because it triggered the hold action right after the finger-down event. If the user just clicked, we used to see a blink from the hold state, then the hint toast appeared.
Now, we just wait for a long-press to start the hold state. If the user just clicks, they will only see the hint toast.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingGesture.kt
andremion: well. not sure if we will keep that with the new designs.
also, we want to get rid of the AudioRecordingTheme (intermediate theming).
if we decide to have that configuration, we have to think of a central place for behavioral customizations.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingPermission.kt
andremion: I wrote in the kdoc

```
* Wrapper around Accompanist's [rememberPermissionState] that returns `null` in
* preview / Paparazzi environments (where there is no Activity context).
```

That null guard is for the preview / paparazzi, which breaks if we follow the permission flow.

gpunto: But if we return `AudioRecordingPermission(PermissionStatus.Granted, {}, {})` instead of null in tests why would we need the activity context? 🤔

andremion: Oh, now I get what you mean. Let me check it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingGesture.kt
andremion: I reviewed the AudioRecordingActions. I believe that bug was caused by the misleading contract.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingPermission.kt
andremion: It works. Thanks 🙌🏻

PR #6152, opened 2026-02-13
Introduce new  `ChannelLogic` implementation with time and memory optimizations

VelikovPetar:
### Goal

Introduce a new, optimized `ChannelLogic` and `ChannelState` implementation that improves time and memory performance for channel state management. Supports bi-directional pagination, and removes the `fillTheGap` logic. The new implementation is gated behind a feature flag (`StateConfig.useLegacyChannelLogic`, defaulting to `true`) to allow a safe, incremental rollout while preserving the existing behavior.

Resolves: https://linear.app/stream/issue/AND-887/channellogic-improvements

### Implementation

**Dual implementation strategy with feature flag:**
- Added `useLegacyChannelLogic` parameter to `StateConfig` (defaults to `true`), propagated through `StateRegistry` and `LogicRegistry` to conditionally select between legacy and new implementations.

**New channel state management (`ChannelStateImpl`):**
- Introduced `ChannelStateImpl` — a new channel state holder that directly manages messages, members, watchers, reads, typing events, polls, pinned messages, and quoted message references with optimized data structures.
- Introduced `ChannelLogicImpl` — a rewritten channel logic that operates directly on `ChannelStateImpl` without the `ChannelStateLogic` intermediary layer, simplifying the call chain and reducing overhead.
- Introduced `ChannelEventHandlerImpl` — a new event handler that updates `ChannelStateImpl` directly, with improvements such as: proper `insideSearch` handling for new message events (caching instead of dropping), `TypingEventPruner` integration, explicit quoted message reference tracking on message updates/deletes, and smarter pinned message management.
- Introduced `ChannelMessagesUpdateLogicImpl` — a bridge between the low-level client and the new channel state for message and reply operations.
- Added `List.kt` utility file with optimized list operations used by the new state implementation.

**Preserved legacy path:**
- Extracted the existing `ChannelLogicImpl` into `ChannelLogicLegacyImpl` (unchanged behavior).
- Extracted the existing `ChannelEventHandler` into `ChannelEventHandlerLegacyImpl` (unchanged behavior).
- Renamed `ChannelMutableState` to `ChannelStateLegacyImpl` for clarity.
- `ChannelEventHandler` is now an interface implemented by both legacy and new handlers.

**Interface and listener refactoring:**
- Expanded the `ChannelLogic` interface with methods previously accessed through `ChannelStateLogic` (e.g., `upsertMembers`, `setHidden`, `markRead`, `typingEventsEnabled`, `setPushPreference`, `setRepliedMessage`, `updateLastMessageAt`).
- Refactored state plugin listeners (`ChannelMarkReadListenerState`, `MarkAllReadListenerState`, `TypingEventListenerState`, `EditMessageListenerState`, `HideChannelListenerState`, `PushPreferencesListenerState`, `SendGiphyListenerState`, `SendAttachmentListenerState`, `QueryMembersListenerState`) to operate through the `ChannelLogic` interface instead of directly accessing `StateRegistry` or `ChannelStateLogic`.
- Simplified `ChannelMessagesUpdateLogic` interface (internal API) — removed methods that are no longer needed and renamed `listenForChannelState()`/`replyMessage()` to `channelState()`/`setRepliedMessage()`.

**Public API compatibility:**
- `StateRegistry` constructor: new parameters (`mutedUsers`, `useLegacyChannelState`) are placed last with default values and `@JvmOverloads`, preserving the original 7-parameter constructor signature.
- `StateConfig`: new `useLegacyChannelLogic` parameter has a default value (`true`), so existing usage is unaffected.

**Compose UI fix:**
- In `Messages.kt`, used `rememberUpdatedState` for `endOfOldMessages` and added a `LaunchedEffect` to handle the edge case where the message list doesn't fill the screen but older messages are available to load (e.g., after a re-watch clears and replaces messages).

### Testing

- **`ChannelLogicImplTest`** (~1,365 lines): Comprehensive tests for the new `ChannelLogicImpl` covering database loading, pagination direction, query results (success/failure), watch, loadBefore/loadAfter/loadAround, message upsert/delete, member management, hidden state, push preferences, replied messages, mark-read, typing events, channel data updates, and event handling delegation.
- **`ChannelEventHandlerImplTest`** (~1,360 lines): Tests for the new event handler covering all event types — message events (new, update, delete, thread), reaction events, member events, watcher events, channel update/hide/visible/delete/truncate events, typing events, read/delivery receipt events, invitation events, ban events, poll events, reminder events, and global events (user presence, mark-all-read, mute updates, user message deletion).
- **`LogicRegistryTest`** (~786 lines): Tests verifying correct instantiation of legacy vs. new channel logic based on the feature flag, channel caching, active channel tracking, and channel-from-message resolution.
- **`StateRegistryTest`** (expanded by ~578 lines): Tests for both legacy and new `StateRegistry` paths covering channel state creation, caching, `isActiveChannel`, `getActiveChannelStates`, `clear`, and `handleBatchEvent`.
- **`ChannelStateImpl*Test`** (11 new test files, ~4,000+ lines total): Granular tests for `ChannelStateImpl` organized by domain — channel data, members, messages, pinned messages, polls, quoted messages, read receipts, typing, users, watchers, and non-channel states (loading, recovery, search, config).
- Updated existing tests for legacy implementations (`ChannelEventHandlerLegacyImplTest`, `ChannelLogicLegacyImplTest`, `ChannelStateLogicTest`) and listener tests (`EditMessageListenerStateTest`, `HideChannelListenerStateTest`, `PushPreferencesListenerStateTest`, `QueryMembersListenerStateTest`, `SendGiphyListenerStateTest`, `TypingEventsTest`, etc.) to align with the refactored interfaces.

## Summary by CodeRabbit

* **New Features**
  * Configurable channel-state mode (legacy vs. new) and a new option to enable legacy channel behavior.
  * New channel APIs: set replied message, push preferences, hide/remove messages, mark read, and typing/keystroke controls.

* **Improvements**
  * Reworked channel event handling for more reliable message, reaction, poll and reminder updates.
  * Better message caching, trimming and offline synchronization.
  * Improved typing indicators and read-receipt processing.

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatHelper.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ChatHelper.kt
@@ -83,6 +83,7 @@ object ChatHelper {
         val stateConfig = StateConfig(
             backgroundSyncEnabled = false,
             userPresence = true,
+            useLegacyChannelLogic = false,
         )
 
         val logLevel = if (BuildConfig.DEBUG) ChatLogLevel.ALL else ChatLogLevel.NOTHING
--- a/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt
+++ b/stream-chat-android-ui-components-sample/src/main/kotlin/io/getstream/chat/ui/sample/application/ChatInitializer.kt
@@ -97,6 +97,7 @@ class ChatInitializer(
         val stateConfig = StateConfig(
             backgroundSyncEnabled = false,
             userPresence = true,
+            useLegacyChannelLogic = false,
         )
 
         val client = ChatClient.Builder(apiKey, context)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/utils/internal/List.kt
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * This implementation uses lazy allocation - if no elements match the filter, the original list
+ * reference is returned without any allocation. This is important for StateFlow usage where
+ * returning the same reference prevents unnecessary emissions.
+ *
+ * @param filter A predicate function that determines whether an element should be updated.
+ * @param update A function that takes an element and returns its updated version.
+ * @return The original list if no elements match the filter, or a new list with updated elements.
+ */
+internal inline fun <T> List<T>.updateIf(filter: (T) -> Boolean, update: (T) -> T): List<T> {
```
VelikovPetar, line 31: These are mostly taken from `feeds` - The plan is to move them to `core`, and then add `core` to `chat.`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1499 @@
...
+    fun addMember(member: Member) {
+        _members.update { current ->
+            if (current.any { it.getUserId() == member.getUserId() }) {
+                current
+            } else {
+                current + member
+            }
+        }
+        _memberCount.update { it + 1 }
+    }
```
aleksandar-apostolov, line 675: Should `_memberCount` only be incremented when a member is actually inserted?

Currently this increments unconditionally, but the list update above skips duplicates. If `addMember` is called twice with the same member (e.g., duplicate events), the count would inflate while the list stays correct.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1499 @@
...
+     *
+     * @param id The ID of the member to delete.
+     */
+    fun deleteMember(id: String) {
+        _members.update { current ->
+            current.filter { it.getUserId() != id }
+        }
+        _memberCount.update { max(0, it - 1) }
+    }
+
```
aleksandar-apostolov, line 724: Same as `addMember` above — should `_memberCount` only be decremented when a member is actually removed?

Currently this decrements unconditionally (with `max(0, ...)` to hide negatives). If `deleteMember` is called for a non-existent member, the count would decrease while the list stays unchanged.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1499 @@
...
+    private var lastStartTypingEvent: Date? = null
+    private var keystrokeParentMessageId: String? = null
+
+    /**
+     * Keeps track of messages who are linked with polls.
+     * Key: Poll ID
+     * Value: Set of Message IDs linked to the poll
+     */
+    private val messagesWithPolls = mutableMapOf<String, Set<String>>()
+    private val polls = mutableMapOf<String, Poll>()
```
aleksandar-apostolov: Are these maps accessed from multiple coroutine dispatchers?

Unlike the `StateFlow` properties, plain `mutableMapOf` isn't thread-safe. If WebSocket events and API callbacks can mutate these concurrently, consider `ConcurrentHashMap` or wrapping in synchronized access.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/utils/internal/List.kt
```diff
@@ -0,0 +1,239 @@
...
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
```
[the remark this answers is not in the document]
aleksandar-apostolov: +1

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImplMembersTest.kt
```diff
@@ -0,0 +1,421 @@
...
+            // given
+            val member = createMember(1)
+            channelState.addMember(member)
+            val initialCount = channelState.membersCount.value
+            // when
+            channelState.addMember(member) // Add same member again
+            // then - count is incremented even though member wasn't added
+            // This is the current behavior - may need to be reconsidered
+            assertEquals(initialCount + 1, channelState.membersCount.value)
+        }
```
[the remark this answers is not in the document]
aleksandar-apostolov: +1

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
```diff
@@ -20,135 +20,251 @@ import io.getstream.chat.android.client.ChatClient
...
+        val request = QueryChannelPaginationRequest()
+            .apply {
+                messageFilterValue = messageId
+                messageFilterDirection = Pagination.AROUND_ID
+            }
+            .toWatchChannelRequest(userPresence)
+            .apply {
+                // Don't refresh the whole state when loading messages around a specific message, because `fillTheGap`
+                // will load the missing messages between the already loaded and the requested messages.
+                shouldRefresh = true
```
aleksandar-apostolov, line 204: This comment references `fillTheGap` which was removed in this PR, and says "don't refresh" while `shouldRefresh = true`. Should update the comment to reflect the new behavior.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
```diff
@@ -20,135 +20,251 @@ import io.getstream.chat.android.client.ChatClient
...
+                val isPermanent = result.value.isPermanent()
+                state.setRecoveryNeeded(recoveryNeeded = !isPermanent)
+                // Reset loading states
+                state.setLoadingOlderMessages(false)
+                state.setLoadingNewerMessages(false)
+            }
         }
     }
 
     override suspend fun watch(limit: Int, userPresence: Boolean): Result<Channel> {
```
aleksandar-apostolov, line 161: The legacy implementation had duplicate request protection here — checking `loading.value` before making the request to prevent DDOS-ing the API. Was this intentionally removed, or should it be preserved?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/utils/internal/List.kt
```diff
@@ -0,0 +1,239 @@
...
+ *   sorted position as-is.
+ */
+internal fun <T, ID> List<T>.upsertSorted(
+    element: T,
+    idSelector: (T) -> ID,
+    comparator: Comparator<in T>,
+    update: (old: T) -> T = { element },
+): List<T> {
+    val elementId = idSelector(element)
+    val existingIndex = this.indexOfFirst { idSelector(it) == elementId }
```
aleksandar-apostolov, line 108: For large message lists, `indexOfFirst` is O(n) per upsert. Since this is called on every message event, it could add up in busy channels. Have you considered maintaining a secondary `Map<ID, Int>` index for O(1) lookups, or is the typical message count small enough that this isn't a concern?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1499 @@
...
+            current + (quotedMessageId to newValue)
+        }
+    }
+
+    /**
+     * Updates each message that quotes the given message.
+     *
+     * @param quotedMessage The message whose quoting messages should be updated.
+     */
+    fun updateQuotedMessageReferences(quotedMessage: Message) {
```
aleksandar-apostolov, line 504: These 3 sequential StateFlow updates could trigger 3 separate UI recompositions. Would batching them (or using a combined state object) reduce UI churn?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/state/StateRegistry.kt
```diff
@@ -48,23 +51,28 @@ import java.util.concurrent.ConcurrentHashMap
...
     private val userStateFlow: StateFlow<User?>,
     private var latestUsers: StateFlow<Map<String, User>>,
     private val activeLiveLocations: StateFlow<List<Location>>,
     private val job: Job,
     private val now: () -> Long,
     private val scope: CoroutineScope,
     private val messageLimitConfig: MessageLimitConfig,
+    private val mutedUsers: StateFlow<List<Mute>> = MutableStateFlow(emptyList()),
+    private val useLegacyChannelState: Boolean = true,
 ) {
```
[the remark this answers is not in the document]
VelikovPetar: Its intentional

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
```diff
@@ -20,135 +20,251 @@ import io.getstream.chat.android.client.ChatClient
...
+                messageFilterDirection = Pagination.AROUND_ID
+            }
+            .toWatchChannelRequest(userPresence)
+            .apply {
+                // Don't refresh the whole state when loading messages around a specific message, because `fillTheGap`
+                // will load the missing messages between the already loaded and the requested messages.
+                shouldRefresh = true
+            }
+        return queryChannel(request)
     }
```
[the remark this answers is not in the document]
VelikovPetar: I will update

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/legacy/ChannelLogicLegacyImpl.kt
```diff
@@ -0,0 +1,448 @@
...
+                                messageLimit,
+                                requestedMessageIds.first(),
+                            ),
+                        )
+
+                    else -> null
+                }?.onSuccess { fillTheGap(messageLimit, loadedMessages, it.messages) }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
VelikovPetar: This code is not new, just moved. We are aware of it, and we want to decommission this class (and replace it with `ChannelLogicImpl`)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1499 @@
...
+     * Deletes a member from the current state.
+     *
+     * @param id The ID of the member to delete.
+     */
+    fun deleteMember(id: String) {
+        _members.update { current ->
+            current.filter { it.getUserId() != id }
+        }
+        _memberCount.update { max(0, it - 1) }
+    }
```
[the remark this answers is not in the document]
VelikovPetar: I will revisit this logic

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/utils/internal/List.kt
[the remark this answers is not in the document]
VelikovPetar: Will fix

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: I will re-visit this logic. I think we need to take into account the different sources form where a member can be added (events `member.added` / `notification.added_to_channel`).

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: Yes I will revisit the add/remove member logic.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: I think so, I will double check check, but I think we should use ConcurrentHashMap

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
VelikovPetar: Will fix

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
VelikovPetar: Good point - however in the legacy implementation, if though we had the `loading` check, the same flag was never set - so the check was redundant. But I will re-think this, it is probably the best if we incorporate the guard, and make sure the flag is updated properly.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/utils/internal/List.kt
VelikovPetar: We cannot know the message count, but we can assume a potentially large number. Do you think it is OK (memory wise) to keep an additional lookup map?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: In my opinion, introducing a combined state object would pollute the API a bit. Note that `_latestCachedMessages` is not observed on the UI (at least not in the message list), so from UI perspective, only `_messages`  and maybe `_pinnedMessages` would be observed.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/c30e8215d5c49b097a6d28b7507d347ff796f325

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/c30e8215d5c49b097a6d28b7507d347ff796f325

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/a427404a94799de01641ddf90591c7f05333fa71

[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/a427404a94799de01641ddf90591c7f05333fa71

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/utils/internal/List.kt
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/dca3f58d0fe993b2c347369bcbb8bd5521de6f4a

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: Update: I will fix this - we shouldn't increment the count if the member was already tracked .

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: Here I think we must keep the decrement, because we don't have any guarantees that we will have the member who left the channel already loaded. (we initially get a limited number of members when loading the channel, but not all of them - the user must then paginate to load every member of the channel)

VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/106beda9b74a0ea9a97f1547b5a2143c8f08f873

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/106beda9b74a0ea9a97f1547b5a2143c8f08f873

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImplMembersTest.kt
```diff
@@ -0,0 +1,421 @@
...
+            // given
+            val members = createMembers(3)
+            channelState.setMembers(members)
+            channelState.setMemberCount(3)
+            // when
+            channelState.deleteMember("non_existing_id")
+            // then
+            assertEquals(3, channelState.members.value.size)
+            assertEquals(2, channelState.membersCount.value) // Count is still decremented
+        }
```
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/106beda9b74a0ea9a97f1547b5a2143c8f08f873

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/106beda9b74a0ea9a97f1547b5a2143c8f08f873

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6152/commits/106beda9b74a0ea9a97f1547b5a2143c8f08f873

VelikovPetar:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1513 @@
...
+        // Update the unread count
+        currentRead?.let {
+            updateRead(
+                it.copy(
+                    lastReceivedEventDate = eventReceivedDate,
+                    unreadMessages = it.unreadMessages.inc(),
+                ),
+            )
+        }
+        processedMessageIds.put(message.id, true)
```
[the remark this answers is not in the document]
VelikovPetar: That's expected

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelEventHandlerImpl.kt
```diff
@@ -0,0 +1,401 @@
...
+            }
+
+            is ChannelVisibleEvent -> state.setHidden(false)
+            is ChannelDeletedEvent -> {
+                state.removeMessagesBefore(event.createdAt)
+                state.deleteChannel(event.createdAt)
+            }
+
+            is ChannelTruncatedEvent -> state.removeMessagesBefore(event.createdAt, event.message)
+            is NotificationChannelTruncatedEvent -> state.removeMessagesBefore(event.createdAt)
```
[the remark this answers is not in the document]
VelikovPetar: done

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1513 @@
...
+     *
+     * **Important:** It wouldn't increment the [membersCount], only adds the newly acquired members of the channel.
+     *
+     * @param members The list of members to upsert.
+     */
+    fun upsertMembers(members: List<Member>) {
+        val currentMembers = _members.value.associateBy(Member::getUserId).toMutableMap()
+        currentMembers += members.associateBy(Member::getUserId)
+        _members.value = currentMembers.values.toList()
+    }
```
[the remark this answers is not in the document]
VelikovPetar: done

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -0,0 +1,1513 @@
...
+     * @param watcherCount The total count of watchers.
+     */
+    fun upsertWatchers(watchers: List<User>, watcherCount: Int) {
+        val currentWatchers = _watchers.value.associateBy { it.id }.toMutableMap()
+        for (watcher in watchers) {
+            currentWatchers[watcher.id] = watcher
+        }
+        _watchers.value = currentWatchers.values.sortedBy(User::createdAt)
+        _watcherCount.value = max(0, watcherCount)
+    }
```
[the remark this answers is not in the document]
VelikovPetar: done

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelEventHandlerImplTest.kt
```diff
@@ -0,0 +1,1361 @@
...
+            channelId = randomString(),
+            user = randomUser(),
+            message = systemMessage,
+            channel = randomChannel(),
+        )
+
+        handler.handle(event)
+
+        verify(state).removeMessagesBefore(event.createdAt, systemMessage)
+    }
```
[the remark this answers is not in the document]
VelikovPetar: done

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelEventHandlerImplTest.kt
```diff
@@ -0,0 +1,1361 @@
...
+
+    @Test
+    fun `When NotificationChannelTruncatedEvent is handled, Then messages are removed`() {
+        val event = randomNotificationChannelTruncatedEvent(cid = cid)
+
+        handler.handle(event)
+
+        verify(state).removeMessagesBefore(event.createdAt)
+    }
```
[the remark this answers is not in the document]
VelikovPetar: done

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelEventHandlerImpl.kt
```diff
@@ -0,0 +1,401 @@
...
+                state.updateMessage(enrichedMessage)
+                // Update Quoted Messages references
+                state.updateQuotedMessageReferences(enrichedMessage)
+                // Handle Pinned status
+                if (event.message.isPinned(now)) {
+                    state.addPinnedMessage(enrichedMessage)
+                } else {
+                    state.deletePinnedMessage(event.message.id)
+                }
+            }
```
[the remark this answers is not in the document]
VelikovPetar: This is OK, we can rely on the server data from the event.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/utils/internal/List.kt
aleksandar-apostolov: Depends on how big our memory footprint is. This comes down to sacrificing CPU or Memory.

aleksandar-apostolov: I think for now we can sacrifice CPU in favour of mem.

aleksandar-apostolov (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/internal/ChannelLogicImpl.kt
VelikovPetar: We will handle this with a `SingleFlight` API implementation once we migrate `V7` to core

PR #6153, opened 2026-02-13
Disallow plugin instantiation and registration

VelikovPetar:
### Goal
Make the plugin system internal-only so external developers can no longer register custom plugins via `ChatClient.Builder.withPlugins()`.

### Implementation
- Removed `ChatClient.Builder.withPlugins(vararg pluginFactories: PluginFactory)`
- Removed the `pluginFactories` field from the `Builder`
- Simplified `setupPluginFactories()` so it only uses built-in plugins (`ThrottlingPluginFactory`, `MessageDeliveredPluginFactory`, `StreamStatePluginFactory`, `StreamOfflinePluginFactory`)
- Annotated `Plugin` and `PluginFactory` with `@InternalStreamChatApi`
- Annotated all 24 listener interfaces (`SendMessageListener`, `QueryChannelListener`, etc.) with `@InternalStreamChatApi`
- Marked `StreamOfflinePluginFactory` as internal
- Removed plugin setup examples from the Kotlin and Java Getting Started guides

### Testing
Everything should work the same, only `withPlugins` is removed.

## Release Notes

* **Breaking Changes**
  * Removed public plugin extension system. Applications can no longer register custom plugins to extend ChatClient functionality.

* **Documentation**
  * Removed examples and guides for plugin configuration.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/listeners/QueryThreadsListener.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/listeners/QueryThreadsListener.kt
@@ -23,7 +23,7 @@ import io.getstream.chat.android.models.QueryThreadsResult
 import io.getstream.result.Result
 
 /**
- * Listener for [ChatClient.queryChannels] requests.
+ * Listener for [ChatClient.queryThreads] requests.
  */
 @InternalStreamChatApi
 public interface QueryThreadsListener {


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/plugin/listeners/QueryThreadsListener.kt
```diff
@@ -18,12 +18,14 @@ package io.getstream.chat.android.client.plugin.listeners
 
 import io.getstream.chat.android.client.ChatClient
 import io.getstream.chat.android.client.api.models.QueryThreadsRequest
+import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import io.getstream.chat.android.models.QueryThreadsResult
 import io.getstream.result.Result
 
 /**
  * Listener for [ChatClient.queryChannels] requests.
```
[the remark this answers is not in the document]
VelikovPetar: https://github.com/GetStream/stream-chat-android/pull/6153/commits/5d8d5a9798fad44c66204ba026cc51129a269435

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -4919,16 +4918,6 @@ internal constructor(
             repositoryFactoryProvider = provider
         }
 
-        /**
-         * Adds plugins factories to be used by the client.
-         * @see [PluginFactory]
-         *
-         * @param pluginFactories The factories to be added.
-         */
-        public fun withPlugins(vararg pluginFactories: PluginFactory): Builder = apply {
```
gpunto, line 4928: Do you know if/how iOS exposes the functionalities that we were exposing as plugins, like all the listeners we have as plugins?

gpunto (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: Not really, with the exception of some specific cases which we also support independently from the plugins setup (ex. `SendMessageInterceptor`).
I would personally remove any public plugin-related functionality, and if needed, expose different ways for customising the same thing. Otherwise, the plugin setup is very limiting and extremely difficult to maintain (IMO)

gpunto: Fully agree, thanks!

PR #6158, opened 2026-02-16
Introduce new stacked group avatars

gpunto:
### Goal

The group avatars have a new design, where we stack user avatars depending on how many there are.

### Implementation

The implementation uses the channel image if there's one, otherwise it checks the number of users in the channel and switches to the appropriate layout where we display 2, 3, 4 avatars or 2 avatars + a count badge.

### 🎨 UI Changes

Preview

Screenshot 2026-02-16 at 16 26 17

| Before | After |
| --- | --- |
| Screenshot_20260216_162812 | Screenshot_20260216_162630 |

### Testing

Can be checked in the sample

## Release Notes

* **New Features**
  * Added stacked avatar support for group channels with member count overflow badges
  * Added online status indicators for avatars
  * Enhanced color theming with new avatar placeholder colors, badge variants, and selection state colors
  * Added numeric large typography style for text styling

* **Bug Fixes**
  * Improved color selections for image picker selection indicators and video overlay elements

* **Refactor**
  * Consolidated component logic and removed preview-only code

### DIFF
--- a/stream-chat-android-compose/api/stream-chat-android-compose.api
+++ b/stream-chat-android-compose/api/stream-chat-android-compose.api
@@ -2175,10 +2175,6 @@ public final class io/getstream/chat/android/compose/ui/components/userreactions
 	public static final fun UserReactions (Ljava/util/List;Landroidx/compose/ui/Modifier;Lkotlin/jvm/functions/Function3;Landroidx/compose/runtime/Composer;II)V
 }
 
-public final class io/getstream/chat/android/compose/ui/composables/AudioWaveSeekbarKt {
-	public static final fun AudioWaveVSeekbar-t6yy7ic (Ljava/util/List;Landroidx/compose/ui/Modifier;JJLkotlin/jvm/functions/Function3;Landroidx/compose/runtime/Composer;II)V
-}
-
 public final class io/getstream/chat/android/compose/ui/mentions/ComposableSingletons$MentionListKt {
 	public static final field INSTANCE Lio/getstream/chat/android/compose/ui/mentions/ComposableSingletons$MentionListKt;
 	public static field lambda-1 Lkotlin/jvm/functions/Function3;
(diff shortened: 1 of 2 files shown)


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/attachments/images/ImagesPicker.kt
```diff
@@ -187,26 +187,26 @@ private fun SelectedIndicator(
             .size(SelectionIndicatorSize)
             .background(
                 shape = CircleShape,
-                color = ChatTheme.colors.borderCoreOnDark,
+                color = ChatTheme.colors.borderCoreOnAccent,
```
gpunto, line 190: Probably some colors changed since the initial implementation of the picker. When I started using `badgeText`, I saw it was pointing to a different color vs Figma, which led me to update it, check its usages and realized some colors had to be updated

gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/ChannelAvatar.kt
```diff
@@ -76,22 +91,31 @@ public fun ChannelAvatar(
...
     modifier: Modifier,
 ) {
-    BoxWithConstraints(modifier) {
+    WithChannelIndicator(
+        channel = channel,
+        currentUser = currentUser,
+        showIndicator = showIndicator,
+        modifier = modifier,
+    ) {
         Avatar(
```
andremion, line 119: If I got it right, according to [that thread](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y?node-id=2867-55764&m=dev#1613407466), for a single user, we always show the placeholder for the second one.

Image

gpunto: Ah good catch, I think it got changed since my initial old implementation 👀 

Fixed!

gpunto:
> Good job! Can you also check why the avatar and the channel text look inconsistent? e.g. the avatar shows 2 members + 11, whereas the channel texts shows 2 members + 10

Good point, it's because we're excluding the current user in the channel title count. I've just asked [in Figma](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y?node-id=5915-268392&m=dev#1636184375) because, as far as I understand, we're not excluding it for the avatar and we don't have examples for this case of the channel title yet.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/ChannelAvatar.kt
andremion: Screenshot 2026-02-17 at 11 50 26
Sorry, I got confused. I still see a single avatar in 1:1 chat. 🤔

gpunto: That's correct, assuming those are direct message chats. We decide this way: if the chat is a direct message (as defined by the `Channel.isOneToOne` extension function), then we show the user avatar. Otherwise, it's a group chat -> this is where the stacked avatars come in.

You can see both cases here:
Screenshot_20260217_125448

andremion: That was my understanding, too, then I thought it got changed when I read the Figma comments

gpunto: I think it's because the thread is on the group avatar section where you can have groups with 1-2 members. Checking the [channel list designs](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y/Chat-SDK-Design-System?node-id=5915-267452&m=dev) we find also the single avatar case

gpunto:
> > Good job! Can you also check why the avatar and the channel text look inconsistent? e.g. the avatar shows 2 members + 11, whereas the channel texts shows 2 members + 10
> 
> Good point, it's because we're excluding the current user in the channel title count. I've just asked [in Figma](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y?node-id=5915-268392&m=dev#1636184375) because, as far as I understand, we're not excluding it for the avatar and we don't have examples for this case of the channel title yet.

So, Jurgen update channel list design examples so that the title fallback logic matches the avatar logic, so we're good to go here! We'll address the inconsistency when updating the channel list design.

PR #6160, opened 2026-02-16
Update design of message reactions

gpunto:
### Goal

Update message reactions design, including two styles: clustered and segmented

### Implementation

- Introduce three composables: clustered + segmented + adaptive that selects one of the two based on available width
- Introduce `MessageReactionItemState` because we need to pass the reactions count
 
### 🎨 UI Changes

| Before | After (segmented) | (clustered) |
| --- | --- | --- |
| Screenshot_20260216_181509 | Screenshot_20260217_112358 | Screenshot_20260216_181624 |

### Testing

By default we're using segmented. If you want to see the clustered style in the sample you can:
- Call it directly from `ChatComponentFactory.MessageReactions` (easiest option)
- Manage to add enough reactions to a message to overflow the screen width

## Summary by CodeRabbit

* **New Features**
  * Added adaptive message reactions with segmented and clustered display modes that automatically adjust based on available space
  * Reactions now position above message content with improved visual layout

* **Improvements**
  * Simplified reaction filtering and display logic
  * Enhanced reaction rendering with updated visual design and spacing

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
@@ -125,6 +125,7 @@ import io.getstream.chat.android.compose.ui.components.composer.CoolDownIndicato
 import io.getstream.chat.android.compose.ui.components.composer.MessageInput
 import io.getstream.chat.android.compose.ui.components.composer.MessageInputOptions
 import io.getstream.chat.android.compose.ui.components.messageoptions.MessageOptions
+import io.getstream.chat.android.compose.ui.components.messages.ClusteredMessageReactions
 import io.getstream.chat.android.compose.ui.components.messages.DefaultMessageContent
 import io.getstream.chat.android.compose.ui.components.messages.DefaultMessageDeletedContent
 import io.getstream.chat.android.compose.ui.components.messages.GiphyMessageContent
@@ -1186,7 +1187,9 @@ public interface ChatComponentFactory {
     }
 
     /**
-     * The default reactions displayed overlaying the message bubble border.
+     * The component displaying the reactions on a message. Defaults to [SegmentedMessageReactions],
+     * but an equivalent implementation with a different visual style is available through
+     * [ClusteredMessageReactions].
      */
     @Composable
     public fun MessageReactions(


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageReactions.kt
```diff
@@ -17,85 +17,202 @@
...
- * Preview of the [MessageReactions] with many reactions.
- */
 @Preview
 @Composable
-private fun ManyMessageReactionsPreview() {
+private fun MultipleSegmentedMessageReactionsPreview() {
     ChatTheme {
-        MessageReactions(options = PreviewReactionOptionData.manyReactions())
+        SegmentedMessageReactions(reactions = PreviewReactionData.manyReactions())
     }
```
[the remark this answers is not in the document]
gpunto: StreamPreview doesn't exist

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageReactions.kt
```diff
@@ -17,85 +17,201 @@
...
- * @param options The list of reactions to display.
+ * @param reactions The list of reactions to display.
  * @param modifier Modifier for styling.
- * @param itemContent Composable that represents each reaction item in the row of message reactions.
+ * @param onClick Handler when the reaction list is clicked.
  */
 @Composable
-public fun MessageReactions(
-    options: List<ReactionOptionItemState>,
+public fun ClusteredMessageReactions(
```
andremion, line 59: `ClusteredMessageReactions` seems not to be used. What's the plan with those public components? To provide clustered and segmented reactions, but only using the latter by default?

gpunto: Yeah, that's my current understanding.

Before the latest two commits, the clustered style was also used as a fallback when the segmented didn't fit, but after the design changes that were presented today I think that's not the case anymo. I have [a comment in Figma](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y?node-id=1381-80711&m=dev#1636665177) double checking that.

gpunto: Confirmed by Jurgen, the fallback doesn't apply anymore.

andremion: So we should remove ClusteredMessageReactions then?

andremion: If that's the case, maybe we can keep the name as `MessageReactions`. WDYT?

gpunto: I've asked [in slack](https://getstream.slack.com/archives/C0A3JU6C6G3/p1771420559890229), because the designs show both styles as options for customization. Let's see.

But I agree: if we don't want to offer the clustered style out of the box we should remove it & rename the composable

PR #6163, opened 2026-02-18
Bump Compose version to 1.9

gpunto:
### Goal

After some hours of fighting, I discovered the compose version we use has a bug with full screen dialogs, which I plan to use for the overlay reactions menu, so I'm bumping to get the fix.

### Implementation

- Bump compose
- Decouple Material icons version as they're not tied anymore

### 🎨 UI Changes

None

### Testing

There should be no changes

## Summary by CodeRabbit

* **Chores**
  - Updated Android Compose framework to the latest stable version, ensuring the app uses the most current UI framework capabilities
  - Refreshed Material Icons library to version 1.7.8 with updated icon resources
  - Streamlined dependency version management across UI component libraries

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -11,7 +11,8 @@ androidxActivityCompose = "1.9.3"
 androidxActivityKtx = "1.9.3"
 androidxAnnotations = "1.9.1"
 androidxAppcompat = "1.7.0"
-androidxCompose = "1.7.5"
+androidxCompose = "1.9.0"
+androidxComposeMaterialIcons = "1.7.8"
 androidxComposeConstraintLayout = "1.1.0"
 androidxComposeMaterial3 = "1.3.1"
 androidxComposeMaterial3Adaptive = "1.0.0"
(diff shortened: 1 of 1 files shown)


gradle/libs.versions.toml
```diff
@@ -11,7 +11,8 @@ androidxActivityCompose = "1.9.3"
 androidxActivityKtx = "1.9.3"
 androidxAnnotations = "1.9.1"
 androidxAppcompat = "1.7.0"
-androidxCompose = "1.7.5"
+androidxCompose = "1.9.0"
```
gpunto, line 14: Btw, is there any reason we're not using the BoM? Maybe it's time to move to that?

gpunto:
> 1 step closer to Compose 2.x 🥳

😄

I've been conservative with this bump to avoid jumping too far, but we should probably discuss which versions we want to end up with in v7

andremion:
> I've been conservative with this bump to avoid jumping too far, but we should probably discuss which versions we want to end up with in v7

But it's a good move. Baby-steps, so we can incrementally test it. 😉

PR #6164, opened 2026-02-18
Message composer focus improvements, keyboard fixes and code cleanup

andremion:
### Goal

- Fix the message composer so the keyboard opens automatically after selecting a command (e.g. `/giphy`)
- Ensure that keyboard and attachment picker is not visible at the same time (issues on OEM devices)
- Clean up redundant RTL mirroring logic and unused avatar gradient utilities.

### Implementation

- **Input focus after command selection:** Added `inputFocusEvents` (`SharedFlow<Unit>`) to `MessageComposerController`, emitted from `selectCommand()`. The `MessageComposerViewModel` exposes it, and `MessageComposer` collects it via a `LaunchedEffect` to call `FocusRequester.requestFocus()` on the input field.
- **Keyboard dismissal via focus clearing:** Replaced `SoftwareKeyboardController.hide()` with `FocusManager.clearFocus()` in `AttachmentPickerMenu`. Some devices ignore the keyboard controller; clearing focus is more reliable.
- **Keyboard visibility detection:** Replaced `WindowInsets.ime.getBottom(LocalDensity.current) > 0` with a `ViewTreeObserver.OnGlobalLayoutListener` + `getWindowVisibleDisplayFrame` approach. This works regardless of the host activity's `windowSoftInputMode` (`adjustResize`, `adjustPan`) and across all OEM ROMs, because it measures actual visible screen area rather than relying on inset dispatch.
- **Remove `mirrorRtl` modifier:** The affected drawables (`ic_arrow_back`, `ic_mic`, send icon) already declare `android:autoMirrored="true"`, making the manual `scaleX` flip redundant. Removed the `mirrorRtl` extension and all call sites.
- **Remove unused `initialsGradient()` function from `ImageUtils.kt` and its associated avatar gradient color resources from `colors.xml`.

### 🎨 UI Changes

No visual changes expected — behaviour should be identical.

### Testing

1. **Command focus**: Open the attachment picker, select a command (e.g. `/giphy`). The picker closes, the input is populated with `/giphy `, and the keyboard appears with the input focused — ready for the user to type the argument.

2. **Keyboard/picker overlap**: On an OEM device, open a channel, open the attachment picker, then tap the message input to open the keyboard. The picker should dismiss. Vice versa: with keyboard open, tap the attachment button — the keyboard should dismiss before the picker appears.

## Summary by CodeRabbit

* **New Features**
  * Added input focus event handling to automatically focus the message composer input when commands are selected.
  * Introduced keyboard visibility detection for improved attachment picker behavior.

* **Bug Fixes**
  * Improved keyboard handling in the attachment picker using focus management instead of direct keyboard controller.

* **Refactor**
  * Removed RTL (right-to-left) mirroring from back buttons and icons across UI components for simplified layout handling.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
```diff
@@ -77,17 +75,17 @@ public fun AttachmentPickerMenu(
 
     var isShowingDialog by rememberSaveable { mutableStateOf(false) }
 
-    // Ensure keyboard is closed when the attachments picker is shown (if instructed by ChatTheme)
-    val keyboardController = LocalSoftwareKeyboardController.current
+    // Dismiss the keyboard when the attachments picker is shown (if instructed by ChatTheme).
+    val focusManager = LocalFocusManager.current
     val shouldCloseKeyboard = ChatTheme.keyboardBehaviour.closeKeyboardOnAttachmentPickerOpen
```
VelikovPetar, line 80: Not related to this PR: We should consider removing the `ChatTheme.keyboardBehaviour.closeKeyboardOnAttachmentPickerOpen` - I am not sure it makes sense to have it anymore with the new picker.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -239,6 +242,13 @@ public class MessageComposerController(
         .flatMapLatest { it.threadDraftMessages }
         .stateIn(scope, SharingStarted.Eagerly, emptyMap())
 
+    private val _inputFocusEvents = MutableSharedFlow<Unit>(extraBufferCapacity = 1)
+
+    /**
+     * Emits each time the message input field should request focus (e.g. after a command is selected).
+     */
+    public val inputFocusEvents: SharedFlow<Unit> = _inputFocusEvents.asSharedFlow()
```
VelikovPetar, line 250: I am maybe overthinking this, but since this is public API, should we maybe change the flow type to `Boolean`? Mostly if we have a case in the future where we would need to 'manually' request removal of focus - so we can emit `false` in such cases.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/AttachmentPickerMenu.kt
andremion: Yes. I'm removing it as part of another task 🙌🏻

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
andremion: Removing focus is simpler, and we don't need any reference. We already do this in the `AttachmentPickerMenu`:
```kotlin
val focusManager = LocalFocusManager.current
...
focusManager.clearFocus()
````

VelikovPetar: yeah I was thinking if there is a case where we cannot deduct that we need to clear the focus from Compose only. But if you think it is redundant, no need to add it.

PR #6318, opened 2026-03-31
Fix video duration always showing 0 in the composer

gpunto:
### Goal

Video duration in the composer always shows 0

### Implementation

- Fix `StorageHelper.getAttachmentFromCursor()` where the duration null-check was incorrectly using `fileSizeIndex` instead of `durationIndex`
- Include `videoLength` as `EXTRA_DURATION` in `Attachment.extraData` when converting from `AttachmentMetaData` in `AttachmentStorageHelper.toAttachments()`
- Query the `DURATION` column in `StorageHelper.getAttachmentsFromUriList()` so that videos picked via the system file picker also have their duration set

### 🎨 UI Changes

| Before | After |
| --- | --- |
| Screenshot_20260331_120143 | Screenshot_20260331_120458 |

### Testing

- Pick a video from the media picker and verify the duration badge in the composer shows the correct value
- Pick a video via the system file picker and verify the duration is also shown correctly

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect column validation when querying video duration metadata during attachment retrieval from device storage.

* **New Features**
  * Enhanced attachment handling to properly capture video duration from device storage and include this information in video attachment metadata for improved playback information.


components-alignment-report.md
VelikovPetar, line 1: Probably committed by mistake?

gpunto: Definitely!

VelikovPetar (approved):

PR #6325, opened 2026-04-01
Rename channel screen components to use Channel naming

gpunto:
### Goal

Rename Compose SDK components to use "Channel" naming instead of "Messages" for the screen that displays a channel's content, since it represents the channel -> this follows with what's defined in the [alignments doc](https://www.notion.so/stream-wiki/Components-Alignment-2e76a5d7f9f680a8afaecf0c716338b2).

Docs PR: https://github.com/GetStream/docs-content/pull/1140

### Implementation

- `ChannelViewModelFactory` (channel list) → `ChannelListViewModelFactory` — renamed first to free up the name
- `MessagesViewModelFactory` → `ChannelViewModelFactory`
- `MessagesScreen` → `ChannelScreen`
- `MessageListHeader` → `ChannelHeader` (including all related classes like `ChannelHeaderParams`, `ChannelHeaderCenterContentParams`, etc.)
- Updated `ChatComponentFactory` methods to match the new names

### Testing

Verify that the channel screen, header, and channel list all work as before.

## Summary by CodeRabbit

* **Refactor**
  * Renamed message UI components and headers (MessagesScreen → ChannelScreen, MessageListHeader → ChannelHeader) and reorganized ViewModel factories to separate channel vs. list responsibilities; minor internal renames for message rendering and header defaults.

* **Documentation**
  * Updated samples, guides, and snippets to use the new component and factory names.

* **Tests**
  * Updated UI snapshots and tests to target the renamed components.

* **Chore**
  * Adjusted benchmark synchronization to wait for the channel screen.


gpunto:
@coderabbitai review

VelikovPetar (approved):
Nice one!
Just to double check, we keep the XML naming the same right?

gpunto:
> Nice one! Just to double check, we keep the XML naming the same right?

Yeah, I left it as is. Do you think we should change it? Maybe it's still better to align?

VelikovPetar:
> > Nice one! Just to double check, we keep the XML naming the same right?
> 
> Yeah, I left it as is. Do you think we should change it? Maybe it's still better to align?

I would say if they are now misaligned with both the Compose SDK, and the UIKit SDK - it would be good to rename them. In the end, it would be easier even for us to navigate the project if we have consistent naming

gpunto:
> I would say if they are now misaligned with both the Compose SDK, and the UIKit SDK - it would be good to rename them. In the end, it would be easier even for us to navigate the project if we have consistent naming

Ok! I applied the renaming for ui-components in https://github.com/GetStream/stream-chat-android/pull/6325/commits/ca347bb22818cb9e395c3337fcb756c30283d95c. It ended up being more changes than expected, mostly due to all the references in resources. Let me know if that's fine or if you think it's too many breaking changes.

VelikovPetar:
> > I would say if they are now misaligned with both the Compose SDK, and the UIKit SDK - it would be good to rename them. In the end, it would be easier even for us to navigate the project if we have consistent naming
> 
> Ok! I applied the renaming for ui-components in [ca347bb](https://github.com/GetStream/stream-chat-android/commit/ca347bb22818cb9e395c3337fcb756c30283d95c). It ended up being more changes than expected, mostly due to all the references in resources. Let me know if that's fine or if you think it's too many breaking changes.

Ugh I didn't realise that resources would be affected as well. But IMO, this is good change and we should do it. As long as we explain the changes in the migration guide, I think we will be good. And as I mentioned, we already had discrepancies between the XML and Compose namings, I think that unifying them is the right thing to do! (additionally, there aren't too many breaking changes in the XML SDK, so even with these ones, a migration should be rather simple)

gpunto:
@coderabbitai review

PR #6327, opened 2026-04-01
Update actions for channels in the channel list

gpunto:
### Goal

Clean up channel actions:
- The swipe action should only be "mute group", so we remove fallbacks for swipe actions
- The "more actions" sheet for groups should not display "mute group" as that's the swipe action
- Update the mute icon on DM channels to only reflect channel mutes and differentiate between "mute channel" (primary swipe action) and "mute user"
  - This last point reverts what I did in https://github.com/GetStream/stream-chat-android/pull/6302 in light of the decision to only show channel mutes with the icon

### Implementation

- Update swipe actions to only show mute/unmute as the primary action (remove archive and pin)
- Remove `ChannelAction.requiredCapability` (it was unused)
- Disable archive channel action by default in `ChannelOptionItemVisibility` as we don't have an archive screen in the SDK
- Revert showing muted icon on DM channels based on user mute: the icon now only reflects channel mutes
- Remove "Mute Group" / "Unmute Group" from the group channel actions sheet (the swipe action already handles channel muting)
- Swipe actions now respect `ChannelOptionItemVisibility` (e.g. `isMuteChannelVisible`)

### 🎨 UI Changes

None

### Testing

- Swipe a channel item and verify only "More" and "Mute"/"Unmute" actions are shown
- Long press a group channel → sheet should not show Mute Group
- Long press a DM channel → sheet should show Mute User, but not Mute Channel/Group
- Mute a DM channel via swipe → verify the mute icon appears on the channel item
- Set `isMuteChannelVisible = false` in `ChannelOptionItemVisibility` → verify the mute swipe action is hidden

## Summary by CodeRabbit

* **Changes**
  * Swipe actions simplified to expose only mute/unmute.
  * Archive and pin actions removed from swipe menu.
  * Group channel menus no longer show mute/unmute.
  * DM channels can show a separate “mute user” option.
  * Archive channel option hidden by default.
  * Channel muting now relies on channel-level controls rather than per-user mute state.


gpunto:
@coderabbitai review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/channels/actions/ChannelAction.kt
```diff
@@ -68,7 +65,6 @@ public class LeaveGroup(
 ) : ChannelAction {
     @DrawableRes
     override val icon: Int = R.drawable.stream_ic_action_leave
-    override val requiredCapability: String = ChannelCapabilities.LEAVE_CHANNEL
```
VelikovPetar, line 71: LGTM! Just one small question: I find it a bit strange that this was introduced in the refactor, but not actually used. Is it possible that we were missing some `requiredCapabilities` check? Or was its usage naturally removed as we were polishing the UI?

gpunto: TBH I'm not sure, maybe @aleksandar-apostolov remembers?

gpunto:
@coderabbitai review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/channels/actions/ChannelAction.kt
gpunto: Btw, I don't know the original intent, but the code checks the capabilities before instantiating an action (e.g. in `buildGroupChannelActions`), so having `requiredCapability` in the action itself seems redundant. Checking out the original commit that introduced it, it's unused there too, so probably it was an initial API design iteration

aleksandar-apostolov: The original intent was to make ChannelAction fully self-describing so consumers could just pass a list and the renderer would filter by requiredCapability automatically. No if/else, no when blocks. The action knows what it needs.

aleksandar-apostolov: Its fine to remove it as capability gating is complex and currently cannot be simplified to reduce the checks to only the requiredCapability from the action.

VelikovPetar (approved):

PR #6328, opened 2026-04-01
Update message grouping to take time difference into account

gpunto:
### Goal

Messages from the same user that are far apart in time should not be visually grouped together. The
default `MessagePositionHandler` now breaks groups when messages are more than 60 seconds apart.

### Implementation

- Update `MessagePositionHandler.defaultHandler` to check the time difference between consecutive
  messages when determining grouping
- Messages from the same user that are more than 60 seconds apart are no longer grouped together

### Testing

Open a conversation where the same user sent messages with large time gaps and verify they are no
  longer grouped

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated message footer visibility behavior to use "last in group" display as default instead of time-based calculation
  * Enhanced message grouping to enforce a 60-second time window between consecutive messages from the same user
  * Simplified edited message footer display


gpunto:
@coderabbitai review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessagePositionHandler.kt
```diff
@@ -16,12 +16,16 @@
...
+import io.getstream.chat.android.client.extensions.getCreatedAtOrDefault
+import io.getstream.chat.android.client.extensions.internal.NEVER
 import io.getstream.chat.android.client.utils.message.isError
 import io.getstream.chat.android.client.utils.message.isSystem
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.models.User
+import io.getstream.chat.android.ui.common.feature.messages.list.MessagePositionHandler.Companion.DEFAULT_MAX_GROUP_TIME_MILLIS
 import io.getstream.chat.android.ui.common.state.messages.list.MessagePosition
+import kotlin.math.abs
```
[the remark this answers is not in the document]
gpunto: but it's the same file

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageFooter.kt
```diff
@@ -121,61 +113,27 @@ public fun MessageFooter(
...
- *
- * This function adjusts the `message.messageTextUpdatedAt` time to ensure it does not exceed the current system time.
- * If the `messageTextUpdatedAt` time is slightly in the future, it resets the time to the current system time.
- *
- * @param message The message object containing the `messageTextUpdatedAt` timestamp.
- * @param modifier Modifier for styling.
- * @param formatType The type of formatting to provide. By default, it's [DateFormatType.RELATIVE].
- */
-@Composable
-internal fun MessageEditedTimestamp(
```
gpunto, line 175: Showing the edited timestamp is a feature that doesn't exist anymore in the new designs, so I'm also removing that

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageFooter.kt
```diff
@@ -121,61 +113,27 @@ public fun MessageFooter(
...
-                            .weight(1f, fill = false),
-                        text = LocalContext.current.getString(R.string.stream_compose_message_list_footnote_edited),
-                        style = timestampStyle,
-                        overflow = TextOverflow.Ellipsis,
-                        maxLines = 1,
-                    )
-                    MessageEditedTimestamp(message = message)
-                }
+        } else if (showEditLabel) {
+            Row(
```
gpunto, line 133: "Edited" should be shown even when the rest of the footer is not shown

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/messages/list/MessageFooterVisibility.kt
```diff
@@ -35,24 +35,10 @@ public sealed class MessageFooterVisibility {
         override fun toString(): String = "LastInGroup"
     }
 
-    /**
-     * The message footer will be visible to items that are sent inside a specified time duration.
-     *
-     * @param timeDifferenceMillis Time duration after which we show the message footer.
-     */
-    public data class WithTimeDifference(
```
gpunto, line 43: As mentioned [in Slack](https://getstream.slack.com/archives/C08NJKNDEC9/p1775050925653509), this now conflicted with the logic in `MessagePositionHandler` so I removed it. Note: we still have the option to remove the whole `MessageFooterVisibility` before releasing v7. It seems kind of an outdated concept now that:
- MessagePositionHandler handles time-based grouping
- the footer can be replaced through the component factory

andremion (approved):
Edge cases already discussed in https://getstream.slack.com/archives/C08NJKNDEC9/p1775050925653509

PR #6330, opened 2026-04-02
Fix unread separator not appearing on first mark-as-unread

gpunto:
### Goal

Fix unread message separator not appearing on the first mark-as-unread action. The separator only appeared if the user marked as unread a second time (and on the previously marked-as.

### Implementation

- Replace the `updateUnreadLabelState` boolean flag with a reactive approach: `observeUnreadLabelState` now recalculates whenever `lastReadMessageId` changes, using `lastProcessedReadMessageId` to skip duplicates. The flag caused a race condition: it was set in the `markUnread` success callback before the server event updated the read state, so the calculation ran on stale data and then blocked the real update.
- Make the unread label "sticky": null results (e.g. from auto-read on scroll) don't overwrite a non-null label.
- `markUnread()` no longer sets any trigger: the label appears when the `NotificationMarkUnreadEvent` updates the read state.
- Add a test covering the race condition scenario.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/8c82e3f9-0516-48be-996d-7514e5eea98f" controls="controls" muted="muted" /> | <video src="https://github.com/user-attachments/assets/e653410e-64cc-4c68-87cb-8829fb4c30dd" controls="controls" muted="muted" /> |

### Testing

1. Open a channel where all messages are read
2. Long-press a message and select "Mark as unread"
3. Verify the unread separator appears immediately (not only on the second attempt)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unread label display behavior to ensure labels persist correctly when transitioning between read and unread message states, enhancing consistency in message management workflows

* **Tests**
  * Added test case for unread label visibility during mark unread operations before server state synchronization
  * Refactored existing test code for improved maintainability and readability

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -230,7 +230,7 @@ public class MessageListController(
      */
     public val unreadLabelState: MutableStateFlow<UnreadLabel?> = MutableStateFlow(null)
     private val showUnreadButtonState = MutableSharedFlow<Boolean>(extraBufferCapacity = 1)
-    private val updateUnreadLabelState = MutableStateFlow(true)
+    private var lastProcessedReadMessageId: String? = null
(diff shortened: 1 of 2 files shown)


gpunto:
@coderabbitai review

VelikovPetar (approved):

PR #6334, opened 2026-04-03
Apply IME padding in create poll screen

gpunto:
### Goal

The create poll screen is not fully scrollable when the keyboard is open

### Implementation

We apply `imePadding` so that the content effectively resizes when the keyboard is open and can scroll. This also solves the problem occurring when adding an option that goes outside of the visible content portion -> after the changes, the column will auto scroll to show the focused input.

### 🎨 UI Changes

None

### Testing

Open the create poll screen, open the keyboard, and verify that you can scroll to the bottom. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed poll creation screen layout to properly adjust when the keyboard appears, preventing content overlap and improving user interaction.


andremion (approved):
👌🏻

PR #6335, opened 2026-04-03
Fix multiple RTL layout issues

gpunto:
### Goal

Fix multiple RTL layout issues.

### Implementation

- Fix composer text starting from the center in RTL by removing the `TextDirection.Content` text style override (also made the input max width as it wasn't)
- Fix channel item swipe actions in RTL by removing manual RTL offset/alignment flips (the framework handles this natively)
- Fix reactions placement in RTL by computing the position based on both `messageAlignment` and layout direction
- Update icon `autoMirrored` flags to match Figma specs: add mirroring to `audio`, `mute`, `search`, `sidebar`, `video`, `video_fill`; remove it from `export`, `quote`, `share`, and all `filetype_*` icons

### 🎨 UI Changes

| What | Before | After |
|--------|--------|-------|
| Positioning of composer text & message reactions | Screenshot_20260407_100118 | Screenshot_20260407_095921 |
| Channel swipe (& you can also see the auto-mirrored search icon) | <video src="https://github.com/user-attachments/assets/e1f9fa11-e8c3-469e-a1c1-e5af9714158f" controls="controls" muted="muted" /> | <video src="https://github.com/user-attachments/assets/fddb4197-f914-4373-8763-7bd7fdb1a802" controls="controls" muted="muted" /> |

### Testing

1. Switch device language to an RTL locale or use the Force RTL developer option
2. Open the channel list: verify swiping a channel item reveals actions correctly
3. Open a channel: verify the composer text is left-aligned (start-aligned) and not centered
4. Send a message with reactions: verify reactions are positioned on the correct side
5. Verify mirrored icons (audio, mute, search, sidebar, video) flip in RTL and non-mirrored icons (export, quote, share, filetype) do not

## Release Notes

* **New Features**
  * Enhanced right-to-left (RTL) layout support for improved message display and reaction positioning.

* **Bug Fixes**
  * Fixed message input field to properly utilize full available width.
  * Improved automatic icon mirroring for RTL languages across audio, mute, search, sidebar, and video icons.


stream-chat-android-compose/src/main/res/drawable/stream_design_ic_filetype_audio.xml
```diff
@@ -17,7 +17,6 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
     android:width="32dp"
     android:height="40dp"
-    android:autoMirrored="true"
```
andremion, line 20: I'm confused now. What's the convention for auto-mirror?
I thought it would be for unsymmetrical icons.
Here we have `stream_design_ic_audio` with auto-mirror on, but `stream_design_ic_filetype_audio` is OFF 🤔

gpunto: I'm not sure there's a convention per se. [In Figma we have a table column](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y/Chat-SDK-Design-System?node-id=13696-41008&m=dev) that tells whether the icon should be mirrored

andremion: Hum.. That column wasn't there when I started to work on the icons...
Where did you find that column for file type icons?

gpunto: I couldn't find it, I used the file/upload icons as references, since they show a file icon and I imagine we should be consistent
 
Screenshot 2026-04-07 at 14 45 28

andremion: I believe we can't infer consistency there; see the audio example.
I posted a comment to the thread https://getstream.slack.com/archives/C0A3JU6C6G3/p1775567041473099?thread_ts=1775039476.936139&cid=C0A3JU6C6G3

andremion: https://getstream.slack.com/archives/C0A3JU6C6G3/p1775570075617479?thread_ts=1775039476.936139&cid=C0A3JU6C6G3

gpunto: Ok, thanks! Updated those too

andremion: can you include this, too?
https://www.notion.so/stream-wiki/Android-QA-New-Design-31b6a5d7f9f680be9378c1561297dae3?source=copy_link#3356a5d7f9f6802083b6fc6ccab75125
If not, that's fine. It can be in another PR

andremion (approved):

stream-chat-android-compose/src/main/res/drawable/stream_design_ic_filetype_audio.xml
gpunto: TBH I'm still not 100% sure how to approach that problem. We're manually concatenating strings over there instead of having a row with separate components so we're kinda mixing layout direction and text direction: the icon, sender, and text seem three separate components (which should follow layout direction), but instead are just one string (which should follow text direction).

andremion: I used `TextDirection` 568d70172ca6c53714090c2ea4ff55416c396555
Let me see if that approach can also be used here

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/MessageComposerTheme.kt
```diff
@@ -87,10 +86,7 @@ public data class ComposerInputFieldTheme(
             return ComposerInputFieldTheme(
                 borderShape = RoundedCornerShape(StreamTokens.radius3xl),
                 backgroundColor = colors.backgroundCoreSurfaceDefault,
-                textStyle = typography.bodyDefault.copy(
-                    color = colors.textPrimary,
-                    textDirection = TextDirection.Content,
-                ),
+                textStyle = typography.bodyDefault.copy(color = colors.textPrimary),
```
VelikovPetar, line 89: Nice catch!
I will port this fix backwards to `develop` as well, as we have an already open ticket for this!

stream-chat-android-compose/src/main/res/drawable/stream_design_ic_filetype_audio.xml
andremion: It seems there's no way to avoid using split composables: 695a9cebb9f655c85e2265c13667eae3b490a312
Let me know what you think

andremion: I also reviewed the styling according to Figma. That's why snapshot files got updated

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/MessagePreviewFormatter.kt
```diff
@@ -164,32 +147,23 @@ private class DefaultMessagePreviewFormatter(
...
-                { msg -> appendSenderName(msg, currentUser, senderNameTextStyle, isDirectMessaging) }
-            } else {
-                null
-            }
             when {
                 message.isSystem() -> append(displayedText)
 
                 message.isDeleted() -> {
-                    appendSender?.invoke(message)
                     appendInlineContent(DefaultMessagePreviewIconFactory.DELETED)
```
gpunto, line 161: I wonder if we should also invert the icon-text order. Visually they look like separate components and their direction would be inverted by a Row, but here we hardcode them in the text. Like, if we were to build this layout anew today, we would have something like `Row { Text(sender), Icon, Text(content) }` but instead we're inheriting this formatter (+ `MessagePreviewIconFactory`) that was putting everything into a single Text

andremion: The icon direction won't cause visible issues in practice because the inline content placeholder is neutral in bidi, and the text following it is strong-directional.
You can test running on RTL. The good thing about keeping the icon in the formatter is that integrators can just override the formatter to build an AnnotatedString that might include icons.

stream-chat-android-compose/src/main/res/drawable/stream_design_ic_filetype_audio.xml
gpunto: Looks good, thanks for it!

PR #6340, opened 2026-04-08
Remove pinned channel items customization from the demo

gpunto:
### Goal

Removing the custom background for pinned channels. We'll revisit it after the state is designed and bring it to the SDK.

### Implementation

Remove the background modifier.

### 🎨 UI Changes

Pinned channels now look visually identical to regular ones

### Testing

Launch the compose sample, pin a channel and verify that the corresponding channel item looks identical.

## Summary by CodeRabbit

* **Style**
  * Removed visual styling previously applied to pinned channels.

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
@@ -16,11 +16,9 @@
 
 package io.getstream.chat.android.compose.sample.ui.component
 
-import androidx.compose.foundation.background
 import androidx.compose.foundation.lazy.LazyItemScope
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import io.getstream.chat.android.client.extensions.isPinned
 import io.getstream.chat.android.compose.sample.ui.location.LocationComponentFactory
 import io.getstream.chat.android.compose.sample.vm.SharedLocationViewModelFactory
 import io.getstream.chat.android.compose.ui.channels.list.ChannelItem
@@ -40,11 +38,6 @@ class CustomChatComponentFactory(
     override fun LazyItemScope.ChannelListItemContent(params: ChannelListItemContentParams) {
         val coordinator = LocalSwipeRevealCoordinator.current
         val swipeEnabled = ChatTheme.config.channelList.swipeActionsEnabled && coordinator != null
-        val pinnedModifier = if (params.channelItem.channel.isPinned()) {
-            Modifier.background(color = ChatTheme.colors.backgroundCoreHighlight)
-        } else {
-            Modifier
-        }
 
         if (swipeEnabled) {
             SwipeableChannelItem(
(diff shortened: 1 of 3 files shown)


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
```diff
@@ -40,11 +38,6 @@ class CustomChatComponentFactory(
     override fun LazyItemScope.ChannelListItemContent(params: ChannelListItemContentParams) {
         val coordinator = LocalSwipeRevealCoordinator.current
         val swipeEnabled = ChatTheme.config.channelList.swipeActionsEnabled && coordinator != null
-        val pinnedModifier = if (params.channelItem.channel.isPinned()) {
-            Modifier.background(color = ChatTheme.colors.backgroundCoreHighlight)
-        } else {
-            Modifier
-        }
```
VelikovPetar, line 47: Should we also remove the custom sort option from the `ChannelsActivity`?
```kotlin
ChannelListViewModelFactory(
            chatClient = chatClient,
            querySort = QuerySortByField
        --------> .descByName<Channel>("pinned_at") // pinned channels first
                .desc("last_updated"), // then by last updated
            filters = Filters.and(
                Filters.`in`("members", listOf(currentUserId)),
                Filters.or(Filters.notExists(CHANNEL_ARG_DRAFT), Filters.eq(CHANNEL_ARG_DRAFT, false)),
            ),
        )
```

andremion (commented):
As I understood from https://getstream.slack.com/archives/C0A3JU6C6G3/p1775651311234899?thread_ts=1775568258.825169&cid=C0A3JU6C6G3, we should hide the pin option in the SDK.

gpunto:
> As I understood from https://getstream.slack.com/archives/C0A3JU6C6G3/p1775651311234899?thread_ts=1775568258.825169&cid=C0A3JU6C6G3, we should hide the pin option in the SDK.

Yeah, it's disabled by default in `ChannelOptionItemVisibility`

andremion:
> > As I understood from https://getstream.slack.com/archives/C0A3JU6C6G3/p1775651311234899?thread_ts=1775568258.825169&cid=C0A3JU6C6G3, we should hide the pin option in the SDK.
> 
> Yeah, it's disabled by default in `ChannelOptionItemVisibility`

Oh, I'm seeing in the sample. Shouldn't we disable in the sample too?

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/component/CustomChatComponentFactory.kt
gpunto: Yeah, we probably should. And then I guess we also remove the option in the sample, since it's explicitly enabled for the sample, otherwise we'll have an action with no visible effect

gpunto:
> > > As I understood from https://getstream.slack.com/archives/C0A3JU6C6G3/p1775651311234899?thread_ts=1775568258.825169&cid=C0A3JU6C6G3, we should hide the pin option in the SDK.
> > 
> > 
> > Yeah, it's disabled by default in `ChannelOptionItemVisibility`
> 
> Oh, I'm seeing in the sample. Shouldn't we disable in the sample too?

Yeah, it makes sense

andremion (approved):

VelikovPetar (approved):

PR #6341, opened 2026-04-08
Remove MessageComposerTheme and its associated types

andremion:
### Goal

Remove `MessageComposerTheme` and its associated types (`ComposerInputFieldTheme`, `MentionStyleFactory`, `StreamDesign`) from the public API, consolidating composer theming under `ChatTheme` and `ChatUiConfig`.

### Implementation

- Deleted `MessageComposerTheme.kt` and all related types (`ComposerInputFieldTheme`, `MentionStyleFactory`, `StreamDesign`).
- Removed the `messageComposerTheme` parameter from `ChatTheme`.
- Removed `CustomMentionStyleFactory` from the sample app and simplified `MessagesActivity.SetupChatTheme()`.
- Moved mention styling logic from the removed `MentionStyleFactory` into `TextUtils.kt`, using the existing theme colors directly.
- Simplified `InputField.kt` and `MessageComposerInputCenterContent.kt` to no longer depend on `MessageComposerTheme`.
- Added a guard in `ChatTheme` for `ChatClient.instance()` to prevent crashes in Compose `@Preview` environments.

No UI changes.

### Testing

1. Open any conversation in the Compose sample app.
2. Verify the message composer renders correctly — input field, attachments, and send button.
3. Type `@` and verify mention suggestions appear and can be selected.
4. Open a `@Preview` for `MessageComposer` or `MessageInput` in Android Studio and verify it renders without crashing.

## Summary by CodeRabbit

* **Refactor**
  * Simplified theme architecture by removing message composer theme customization capabilities.
  * Removed mention styling customization support from input field.
  
* **Style**
  * Input field now uses default theme styling instead of custom composer theme configuration.

* **Tests**
  * Updated text utility tests to reflect removal of mention styling functionality.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
```diff
@@ -223,7 +214,14 @@ public fun ChatTheme(
     }
 
     val context = LocalContext.current
-    val cdn = remember { ChatClient.instance().cdn }
+    val cdn = remember {
+        // Guard needed because ChatClient is not initialized in Compose @Preview environments
+        try {
+            ChatClient.instance().cdn
```
gpunto: Hmmmm, I'm not sure we should be swallowing the exception in production. I'd say it's correct to crash if an integrator forgets to init ChatClient.

We could add a condition checking `LocalInspectionMode`, but even in that case we'd only catch the exception here. Any other code accessing `ChatClient` would still crash

andremion: That's why we don't always need to `ChatPreviewTheme` in Preview, because not every UI component depends on `ChatClient` directly. Those preview functions started to break after the CDN PR.

andremion: We can check for LocalInspectionMode and propagate the exception if it's false. WDYT?

gpunto: Sounds good!

andremion: 8f13914c6ad98f7955d3a9302296a6983696a263

gpunto (approved):

PR #6342, opened 2026-04-08
Introduce ThreadsScreen and ThreadListHeader composables

gpunto:
### Goal

Add a header to the thread list screen, matching the Figma specs and the existing `ChannelsScreen` / `ChannelListHeader` pattern.

### Implementation

- Add `ThreadListHeader` composable that reuses `ChannelListHeader` internally with no trailing action button
- Add `ThreadsScreen` composable that wraps `ThreadListHeader` + `ThreadList`, mirroring `ChannelsScreen`
- Add `ThreadListHeaderParams` and `ChatComponentFactory.ThreadListHeader` for customization
- Expose `user` and `connectionState` on `ThreadListController` and `ThreadListViewModel`
- Update the sample app to use `ThreadsScreen`

### 🎨 UI Changes

| Before | After |
|--------|-------|
|    Screenshot_20260408_150321    |  Screenshot_20260408_150207   |

### Testing

Launch the sample and verify that the threads screen shows the header

## Release Notes

* **New Features**
  * Added a dedicated thread list header component displaying connection status and current user information.
  * Introduced a unified ThreadsScreen composable integrating the header with the thread list interface.
  * Enhanced thread list presentation with real-time connection state indication and user session display.

* **Tests**
  * Added snapshot tests for the thread list header across multiple connection states and user scenarios.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/threads/ThreadListHeader.kt
```diff
@@ -0,0 +1,136 @@
...
+    modifier: Modifier = Modifier,
+    title: String = stringResource(R.string.stream_compose_thread_list_header_title),
+    currentUser: User? = null,
+    connectionState: ConnectionState = ConnectionState.Connected,
+    color: Color = ChatTheme.colors.backgroundCoreElevation1,
+    shape: Shape = RectangleShape,
+    elevation: Dp = 0.dp,
+    onAvatarClick: (User?) -> Unit = {},
+) {
+    ChannelListHeader(
```
VelikovPetar: Maybe a small suggestion: Instead of calling `ChannelListHeader` from the `ThreadListHeader`, maybe we can have a common `ListHeader` (or something similar) that will be called by both the channels and the threads header?

VelikovPetar (approved):
Overall looks good to me, left just one small suggestion - feel free to ignore if you think it doesn't make sense. 

If it is decided to introduce the header/screen as SDK components, let's merge this PR.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/threads/ThreadListHeader.kt
gpunto: Makes sense! I extracted a common ListHeader & related components so now Channels/Threads implementations point to those

gpunto:
> Overall looks good to me, left just one small suggestion - feel free to ignore if you think it doesn't make sense.
> 
> If it is decided to introduce the header/screen as SDK components, let's merge this PR.

So, the implementation is not consistent across mobile SDKs, but the code currently in the PR seems to be aligned with what iOS exposes (and asked in [Slack](https://getstream.slack.com/archives/C0A3JU6C6G3/p1775744000051759?thread_ts=1775724701.941179&cid=C0A3JU6C6G3)):

> Header: exposed via factory (`ThreadListHeader` / `makeThreadListHeaderViewModifier`)                                                                
> Screen: public composable/view, not in the factory (`ThreadsScreen` / `ChatThreadListView`)

WDYT?

VelikovPetar:
I would go for it, as long as we are in sync on what we provide, and which customizations we allow!

gpunto:
Marking as draft, I need to check a few things before we can merge

PR #6344, opened 2026-04-08
Add opt-in fast event parsing (`message.new`)

VelikovPetar:
### Goal

Introduce an **opt-in** direct JSON-to-Domain parsing path for WebSocket events, bypassing the intermediate DTO layer. This reduces object allocations and speeds up event deserialization on the hot path (starting with `message.new`). The new path produces domain objects identical to the existing DTO-based path.

The motivation: during livestreams, massive bursts of `message.new` events cause large numbers of object allocations (amplified by the intermediate DTO objects), triggering throttles and unnecessary garbage collections. Skipping the DTO layer results in a much smaller memory footprint and fewer temporary allocations.

To preserve behavior for existing apps on upgrade, the fast path is **disabled by default** and must be explicitly enabled by the customer via `ChatClient.Builder.fastEventParsing(true)`.

Resolves: https://linear.app/stream/issue/AND-1143/improve-messagenew-event-parsing-memory-footprint

### Implementation

**Opt-in toggle (new public API)**
- `ChatClient.Builder.fastEventParsing(enabled: Boolean)` — default `false`.
- `ChatClientConfig.fastEventParsing: Boolean` — propagated through `ChatModule`.
- When `false` (default), `DirectEventParser` is not instantiated and `MoshiChatParser` uses the legacy DTO path exclusively — same behavior as before this PR.
- When `true`, supported event types (currently `message.new`) are parsed via the direct path; unsupported event types still fall back to the DTO path, so enabling the toggle is safe even as the supported-event set grows.

**Direct JSON adapters** — Hand-written `JsonAdapter` implementations that read JSON via `JsonReader` and construct domain objects directly, skipping DTO allocation:
- `UserAdapter`, `MessageAdapter`, `NewMessageEventAdapter` for the core event chain
- Supporting adapters: `AttachmentAdapter`, `ReactionAdapter`, `ReactionGroupAdapter`, `PollAdapter`, `OptionAdapter`, `DeviceAdapter`, `PrivacySettingsAdapter`, `LocationAdapter`, `ModerationAdapter`, `MessageModerationDetailsAdapter`, `MessageReminderInfoAdapter`, `ChannelInfoAdapter`
- `JsonParsingUtils` — shared utilities for nullable fields, extra data accumulation, list/map parsing, and required-field validation with Moshi-parity error messages

**`DirectEventParser`** — Router that extracts the event `type` from raw JSON and dispatches to the appropriate direct adapter. Currently supports `message.new`; unsupported types fall back to the existing DTO path. Wired into `MoshiChatParser.parseAndProcessEvent()` as a nullable dependency that is only instantiated when `fastEventParsing` is enabled.

**`UserTransformer` / `MessageTransformer` parity** — Both transformers are applied identically in the direct path (at the end of object construction, via `.let(transformer::transform)`), matching the DTO path. `ApiModelTransformers` from `ChatClient.Builder` are passed through `ChatModule` → `DirectEventParser` → individual adapters.

**Enrichment inline** — The `NewMessageEventAdapter` performs channel info enrichment (channel name/image from `channel_custom`) during parsing, so `enrichIfNeeded()` is not needed for directly-parsed events.

### Testing

- **Dual-path parity tests** for every adapter (`UserParsingTest`, `MessageParsingTest`, `NewMessageEventParsingTest`, plus `Attachment`, `Reaction`, `Poll`, `Device`, `Location`, etc.): each test parses the same JSON through both DTO and direct paths, asserts identical domain output, and verifies identical `JsonDataException` for missing required fields.
- **Transformer parity tests**: custom `UserTransformer` and `MessageTransformer` applied to both paths, verifying identical output. Includes nested `UserTransformer` verification (message.user, mentionedUsers, threadParticipants, reaction users, poll vote/answer/createdBy users).
- **`DirectEventParserTest`**: verifies type extraction, event routing, fallback for unsupported types, and end-to-end transformer wiring.
- **`MoshiChatParserTest.FastEventParsingToggle`** (new): verifies that with `fastEventParsing = false` the parser falls back to the DTO path end-to-end for `message.new`, and that with `fastEventParsing = true` the direct and DTO paths produce equal `NewMessageEvent` output.
- **Null unread count parity**: both paths throw `JsonDataException` on `"total_unread_count": null` (Moshi default only applies when field is absent, not null).

To run the benchmark comparing DTO vs Direct parsing performance (speed + allocations), apply the patch below and run:
```
./gradlew :stream-chat-android-client:testDebugUnitTest --tests "io.getstream.chat.android.client.parser2.NewMessageEventBenchmarkTest"
```

<details>

<summary>NewMessageEventBenchmarkTest.kt — DTO vs Direct parsing benchmark</summary>

```
Subject: [PATCH] [EVENT-PARSER] BENCH TEST
---
Index: stream-chat-android-client/build.gradle.kts
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-client/build.gradle.kts b/stream-chat-android-client/build.gradle.kts
--- a/stream-chat-android-client/build.gradle.kts	(revision 6f1157564f257a6d83f298014565eb7bdab33765)
+++ b/stream-chat-android-client/build.gradle.kts	(date 1775648874178)
@@ -44,6 +44,9 @@
 
     testOptions.unitTests {
         isReturnDefaultValues = true
+        all { test ->
+            test.jvmArgs("--add-modules", "java.management,jdk.management")
+        }
     }
 }
 
@@ -69,6 +72,25 @@
     }
 }
 
+// Extract java.management + jdk.management classes from JDK jmods into a JAR for test compilation.
+// Android's bootclasspath (android.jar) excludes java.lang.management / com.sun.management,
+// but unit tests run on the host JVM where these classes exist.
+val jdkManagementJar by tasks.registering(Jar::class) {
+    archiveBaseName.set("jdk-management-stub")
+    val jdkHome = org.gradle.internal.jvm.Jvm.current().javaHome
+    from(zipTree(File(jdkHome, "jmods/java.management.jmod"))) {
+        include("classes/**")
+        eachFile { path = path.removePrefix("classes/") }
+    }
+    from(zipTree(File(jdkHome, "jmods/jdk.management.jmod"))) {
+        include("classes/**")
+        eachFile { path = path.removePrefix("classes/") }
+    }
+    includeEmptyDirs = false
+    duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+    destinationDirectory.set(layout.buildDirectory.dir("jdk-stubs"))
+}
+
 dependencies {
     api(project(":stream-chat-android-core"))
 
@@ -101,6 +123,9 @@
     implementation(libs.androidx.constraintlayout)
     implementation(libs.androidx.lifecycle.livedata.ktx)
 
+    // JDK management classes for benchmark allocation tracking (compile-time only)
+    testCompileOnly(files(jdkManagementJar.map { it.archiveFile }))
+
     // Tests
     testImplementation(project(":stream-chat-android-test"))
     testImplementation(project(":stream-chat-android-client-test"))
Index: stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/NewMessageEventBenchmarkTest.kt
IDEA additional info:
<+>UTF-8
diff --git a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/NewMessageEventBenchmarkTest.kt b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/NewMessageEventBenchmarkTest.kt
new file mode 100644
--- /dev/null	(date 1775657446715)
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/NewMessageEventBenchmarkTest.kt	(date 1775657446715)
@@ -0,0 +1,452 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *
+ */
+
+package io.getstream.chat.android.client.parser2
+
+import com.squareup.moshi.Moshi
+import io.getstream.chat.android.client.api2.mapping.DomainMapping
+import io.getstream.chat.android.client.api2.mapping.EventMapping
+import io.getstream.chat.android.client.api2.model.dto.NewMessageEventDto
+import io.getstream.chat.android.client.extensions.internal.enrichIfNeeded
+import io.getstream.chat.android.client.parser2.adapters.DateAdapter
+import io.getstream.chat.android.client.parser2.event.AttachmentAdapter
+import io.getstream.chat.android.client.parser2.event.ChannelInfoAdapter
+import io.getstream.chat.android.client.parser2.event.DeviceAdapter
+import io.getstream.chat.android.client.parser2.event.LocationAdapter
+import io.getstream.chat.android.client.parser2.event.MessageAdapter
+import io.getstream.chat.android.client.parser2.event.MessageModerationDetailsAdapter
+import io.getstream.chat.android.client.parser2.event.MessageReminderInfoAdapter
+import io.getstream.chat.android.client.parser2.event.ModerationAdapter
+import io.getstream.chat.android.client.parser2.event.NewMessageEventAdapter
+import io.getstream.chat.android.client.parser2.event.OptionAdapter
+import io.getstream.chat.android.client.parser2.event.PollAdapter
+import io.getstream.chat.android.client.parser2.event.PrivacySettingsAdapter
+import io.getstream.chat.android.client.parser2.event.ReactionAdapter
+import io.getstream.chat.android.client.parser2.event.ReactionGroupAdapter
+import io.getstream.chat.android.client.parser2.event.UserAdapter
+import io.getstream.chat.android.client.parser2.testdata.NewMessageEventTestData
+import io.getstream.chat.android.models.NoOpChannelTransformer
+import io.getstream.chat.android.models.NoOpMessageTransformer
+import io.getstream.chat.android.models.NoOpUserTransformer
+import org.junit.jupiter.api.MethodOrderer
+import org.junit.jupiter.api.Order
+import org.junit.jupiter.api.TestMethodOrder
+import java.lang.management.ManagementFactory
+import java.util.Date
+import kotlin.math.sqrt
+
+@TestMethodOrder(MethodOrderer.OrderAnnotation::class)
+internal class NewMessageEventBenchmarkTest {
+
+    // region Setup — DTO path
+
+    private val parser = ParserFactory.createMoshiChatParser()
+
+    private val domainMapping = DomainMapping(
+        currentUserIdProvider = { "" },
+        channelTransformer = NoOpChannelTransformer,
+        messageTransformer = NoOpMessageTransformer,
+        userTransformer = NoOpUserTransformer,
+    )
+
+    private val eventMapping = EventMapping(domainMapping)
+
+    // endregion
+
+    // region Setup — Direct path
+
+    private val moshi = Moshi.Builder().add(DateAdapter()).build()
+    private val dateAdapter = moshi.adapter(Date::class.java)
+
+    private val deviceAdapter = DeviceAdapter()
+    private val privacySettingsAdapter = PrivacySettingsAdapter()
+    private val userAdapter = UserAdapter(
+        deviceAdapter = deviceAdapter,
+        privacySettingsAdapter = privacySettingsAdapter,
+        dateAdapter = dateAdapter,
+        userTransformer = NoOpUserTransformer,
+    )
+    private val reactionAdapter = ReactionAdapter(
+        userAdapter = userAdapter,
+        dateAdapter = dateAdapter,
+    )
+    private val reactionGroupAdapter = ReactionGroupAdapter(
+        dateAdapter = dateAdapter,
+    )
+    private val attachmentAdapter = AttachmentAdapter()
+    private val channelInfoAdapter = ChannelInfoAdapter()
+    private val moderationDetailsAdapter = MessageModerationDetailsAdapter()
+    private val moderationAdapter = ModerationAdapter()
+    private val optionAdapter = OptionAdapter()
+    private val pollAdapter = PollAdapter(
+        userAdapter = userAdapter,
+        optionAdapter = optionAdapter,
+        dateAdapter = dateAdapter,
+        currentUserIdProvider = { "" },
+    )
+    private val reminderAdapter = MessageReminderInfoAdapter(
+        dateAdapter = dateAdapter,
+    )
+    private val locationAdapter = LocationAdapter(
+        dateAdapter = dateAdapter,
+    )
+    private val messageAdapter = MessageAdapter(
+        attachmentAdapter = attachmentAdapter,
+        channelInfoAdapter = channelInfoAdapter,
+        reactionAdapter = reactionAdapter,
+        reactionGroupAdapter = reactionGroupAdapter,
+        userAdapter = userAdapter,
+        moderationDetailsAdapter = moderationDetailsAdapter,
+        moderationAdapter = moderationAdapter,
+        pollAdapter = pollAdapter,
+        reminderAdapter = reminderAdapter,
+        locationAdapter = locationAdapter,
+        dateAdapter = dateAdapter,
+        messageTransformer = NoOpMessageTransformer,
+    )
+
+    private val directAdapter = NewMessageEventAdapter(
+        messageAdapter = messageAdapter,
+        userAdapter = userAdapter,
+    )
+
+    // endregion
+
+    // region Benchmark config
+
+    companion object {
+        private const val WARMUP_ITERATIONS = 1_000
+        private const val SPEED_ITERATIONS = 10_000
+        private const val ALLOC_ITERATIONS = 10_000
+    }
+
+    // endregion
+
+    // region Speed benchmarks
+
+    @Test
+    @Order(1)
+    fun `Speed - DTO path - allFields`() {
+        benchmarkSpeed("Speed - DTO path - allFields") { parseDtoPath(NewMessageEventTestData.jsonAllFields) }
+    }
+
+    @Test
+    @Order(2)
+    fun `Speed - Direct path - allFields`() {
+        benchmarkSpeed("Speed - Direct path - allFields") { parseDirectPath(NewMessageEventTestData.jsonAllFields) }
+    }
+
+    @Test
+    @Order(3)
+    fun `Speed - DTO path - optionalMissing`() {
+        benchmarkSpeed("Speed - DTO path - optionalMissing") { parseDtoPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+    }
+
+    @Test
+    @Order(4)
+    fun `Speed - Direct path - optionalMissing`() {
+        benchmarkSpeed("Speed - Direct path - optionalMissing") { parseDirectPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+    }
+
+    // endregion
+
+    // region Allocation benchmarks
+
+    @Test
+    @Order(5)
+    fun `Allocation - DTO path - allFields`() {
+        benchmarkAllocation("Allocation - DTO path - allFields") { parseDtoPath(NewMessageEventTestData.jsonAllFields) }
+    }
+
+    @Test
+    @Order(6)
+    fun `Allocation - Direct path - allFields`() {
+        benchmarkAllocation("Allocation - Direct path - allFields") { parseDirectPath(NewMessageEventTestData.jsonAllFields) }
+    }
+
+    @Test
+    @Order(7)
+    fun `Allocation - DTO path - optionalMissing`() {
+        benchmarkAllocation("Allocation - DTO path - optionalMissing") { parseDtoPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+    }
+
+    @Test
+    @Order(8)
+    fun `Allocation - Direct path - optionalMissing`() {
+        benchmarkAllocation("Allocation - Direct path - optionalMissing") { parseDirectPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+    }
+
+    // endregion
+
+    // region Summary
+
+    @Test
+    @Order(9)
+    fun `Summary comparison`() {
+        val threadMXBean = ManagementFactory.getThreadMXBean() as com.sun.management.ThreadMXBean
+        val threadId = Thread.currentThread().id
+
+        // Measure speed (median nanoseconds)
+        val speedDtoAll = measureMedian { parseDtoPath(NewMessageEventTestData.jsonAllFields) }
+        val speedDirectAll = measureMedian { parseDirectPath(NewMessageEventTestData.jsonAllFields) }
+        val speedDtoMin = measureMedian { parseDtoPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+        val speedDirectMin = measureMedian { parseDirectPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+
+        // Measure allocation (bytes per iteration)
+        val allocDtoAll =
+            measureAllocation(threadMXBean, threadId) { parseDtoPath(NewMessageEventTestData.jsonAllFields) }
+        val allocDirectAll =
+            measureAllocation(threadMXBean, threadId) { parseDirectPath(NewMessageEventTestData.jsonAllFields) }
+        val allocDtoMin = measureAllocation(
+            threadMXBean,
+            threadId,
+        ) { parseDtoPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+        val allocDirectMin = measureAllocation(
+            threadMXBean,
+            threadId,
+        ) { parseDirectPath(NewMessageEventTestData.jsonOptionalFieldsMissing) }
+
+        val w = 78
+        val sep = "=".repeat(w)
+        val thinSep = "-".repeat(w)
+
+        println()
+        println(sep)
+        println(centerText("NewMessageEvent Parsing Benchmark Results", w))
+        println(centerText("End-to-end: extractType + parse + enrichIfNeeded", w))
+        println(centerText("DTO (Moshi codegen + mapping) vs Direct (hand-written JsonReader)", w))
+        println(sep)
+
+        // Speed table
+        println()
+        println("  SPEED (median, $SPEED_ITERATIONS iterations after $WARMUP_ITERATIONS warmup)")
+        println("  $thinSep")
+        println("  %-24s %10s %10s %10s %12s".format("Payload", "DTO", "Direct", "Saved", "Improvement"))
+        println("  $thinSep")
+        printSpeedRow("allFields (complex)", speedDtoAll, speedDirectAll)
+        printSpeedRow("minimalFields", speedDtoMin, speedDirectMin)
+        println("  $thinSep")
+
+        // Allocation table
+        println()
+        println("  ALLOCATION (per parse, $ALLOC_ITERATIONS iterations via ThreadMXBean)")
+        println("  $thinSep")
+        println("  %-24s %10s %10s %10s %12s".format("Payload", "DTO", "Direct", "Saved", "Improvement"))
+        println("  $thinSep")
+        printAllocRow("allFields (complex)", allocDtoAll, allocDirectAll)
+        printAllocRow("minimalFields", allocDtoMin, allocDirectMin)
+        println("  $thinSep")
+
+        // Visual bars
+        println()
+        println("  VISUAL COMPARISON (normalized to DTO = 100%)")
+        println("  $thinSep")
+        printBar("Speed  | allFields", speedDtoAll, speedDirectAll)
+        printBar("Speed  | minimal  ", speedDtoMin, speedDirectMin)
+        printBar("Alloc  | allFields", allocDtoAll, allocDirectAll)
+        printBar("Alloc  | minimal  ", allocDtoMin, allocDirectMin)
+        println("  $thinSep")
+
+        println()
+        println(sep)
+        println()
+    }
+
+    // endregion
+
+    // region Summary helpers
+
+    private fun measureAllocation(
+        threadMXBean: com.sun.management.ThreadMXBean,
+        threadId: Long,
+        block: () -> Any?,
+    ): Long {
+        repeat(WARMUP_ITERATIONS) { block() }
+        @Suppress("ExplicitGarbageCollectionCall")
+        System.gc()
+        Thread.sleep(50)
+        val before = threadMXBean.getThreadAllocatedBytes(threadId)
+        var lastResult: Any? = null
+        repeat(ALLOC_ITERATIONS) { lastResult = block() }
+        val after = threadMXBean.getThreadAllocatedBytes(threadId)
+        lastResult.hashCode()
+        return (after - before) / ALLOC_ITERATIONS
+    }
+
+    private fun printSpeedRow(label: String, dtoNs: Long, directNs: Long) {
+        val dtoUs = dtoNs / 1000.0
+        val directUs = directNs / 1000.0
+        val savedUs = dtoUs - directUs
+        val improvement = dtoNs.toDouble() / directNs
+        println(
+            "  %-24s %8.1f us %8.1f us %8.1f us %10.1fx".format(
+                label,
+                dtoUs,
+                directUs,
+                savedUs,
+                improvement,
+            ),
+        )
+    }
+
+    private fun printAllocRow(label: String, dtoBytes: Long, directBytes: Long) {
+        val savedBytes = dtoBytes - directBytes
+        val improvement = dtoBytes.toDouble() / directBytes
+        println(
+            "  %-24s %10s %10s %10s %10.1fx".format(
+                label,
+                formatBytes(dtoBytes),
+                formatBytes(directBytes),
+                formatBytes(savedBytes),
+                improvement,
+            ),
+        )
+    }
+
+    private fun printBar(label: String, dtoValue: Long, directValue: Long) {
+        val barWidth = 30
+        val directPct = (directValue.toDouble() / dtoValue * 100).toInt().coerceIn(1, 100)
+        val directBarLen = (directPct * barWidth / 100).coerceAtLeast(1)
+        val dtoBar = "#".repeat(barWidth)
+        val directBar = "#".repeat(directBarLen) + " ".repeat(barWidth - directBarLen)
+        val improvement = dtoValue.toDouble() / directValue
+        println("  $label  DTO    |$dtoBar| 100%")
+        println(
+            "  %s  Direct |%s| %d%%  (%.1fx faster)"
+                .format(" ".repeat(label.length), directBar, directPct, improvement),
+        )
+        println()
+    }
+
+    private fun centerText(text: String, width: Int): String {
+        val pad = ((width - text.length) / 2).coerceAtLeast(0)
+        return " ".repeat(pad) + text
+    }
+
+    // endregion
+
+    // region Helpers
+
+    /** Real DTO flow: extractType → Moshi codegen → toDomain() → enrichIfNeeded() */
+    private fun parseDtoPath(json: String): Any {
+        DirectEventParser.extractType(json)
+        val dto = parser.fromJson(json, NewMessageEventDto::class.java)
+        val event = with(eventMapping) { dto.toDomain() }
+        return event.enrichIfNeeded()
+    }
+
+    /** Real Direct flow: extractType → hand-written JsonReader (enrichment done inline) */
+    private fun parseDirectPath(json: String): Any {
+        DirectEventParser.extractType(json)
+        return directAdapter.fromJson(json)!!
+    }
+
+    private fun benchmarkSpeed(label: String, block: () -> Any?) {
+        // Warmup
+        repeat(WARMUP_ITERATIONS) { block() }
+
+        // Measure
+        val timings = LongArray(SPEED_ITERATIONS)
+        for (i in 0 until SPEED_ITERATIONS) {
+            val start = System.nanoTime()
+            val result = block()
+            timings[i] = System.nanoTime() - start
+            // Reference result to prevent DCE
+            result.hashCode()
+        }
+        timings.sort()
+        printSpeedStats(label, timings)
+    }
+
+    private fun benchmarkAllocation(label: String, block: () -> Any?) {
+        val threadMXBean = ManagementFactory.getThreadMXBean() as com.sun.management.ThreadMXBean
+        val threadId = Thread.currentThread().id
+
+        // Warmup
+        repeat(WARMUP_ITERATIONS) { block() }
+
+        // GC to reduce interference
+        @Suppress("ExplicitGarbageCollectionCall")
+        System.gc()
+        Thread.sleep(50)
+
+        // Measure — getThreadAllocatedBytes is monotonically increasing
+        val before = threadMXBean.getThreadAllocatedBytes(threadId)
+        var lastResult: Any? = null
+        repeat(ALLOC_ITERATIONS) { lastResult = block() }
+        val after = threadMXBean.getThreadAllocatedBytes(threadId)
+
+        // Reference result to prevent DCE
+        lastResult.hashCode()
+
+        val totalBytes = after - before
+        val perIterationBytes = totalBytes / ALLOC_ITERATIONS
+
+        println()
+        println("=== BENCHMARK: $label ===")
+        println("  Iterations:      $ALLOC_ITERATIONS")
+        println("  Total allocated: ${formatBytes(totalBytes)}")
+        println("  Per iteration:   ${formatBytes(perIterationBytes)}")
+        println()
+    }
+
+    private fun measureMedian(block: () -> Any?): Long {
+        // Warmup
+        repeat(WARMUP_ITERATIONS) { block() }
+
+        val timings = LongArray(SPEED_ITERATIONS)
+        for (i in 0 until SPEED_ITERATIONS) {
+            val start = System.nanoTime()
+            val result = block()
+            timings[i] = System.nanoTime() - start
+            result.hashCode()
+        }
+        timings.sort()
+        return timings[timings.size / 2]
+    }
+
+    private fun printSpeedStats(label: String, sortedTimings: LongArray) {
+        val n = sortedTimings.size
+        val min = sortedTimings[0]
+        val median = sortedTimings[n / 2]
+        val mean = sortedTimings.sum() / n
+        val p95 = sortedTimings[(n * 0.95).toInt()]
+        val p99 = sortedTimings[(n * 0.99).toInt()]
+
+        val variance = sortedTimings.fold(0.0) { acc, t -> acc + (t - mean).toDouble().let { it * it } } / n
+        val stdDev = sqrt(variance)
+
+        println()
+        println("=== BENCHMARK: $label ===")
+        println("  Iterations:   $n")
+        println("  Min:          %.1f us".format(min / 1000.0))
+        println("  Median (p50): %.1f us".format(median / 1000.0))
+        println("  Mean:         %.1f us".format(mean / 1000.0))
+        println("  p95:          %.1f us".format(p95 / 1000.0))
+        println("  p99:          %.1f us".format(p99 / 1000.0))
+        println("  StdDev:       %.1f us".format(stdDev / 1000.0))
+        println()
+    }
+
+    private fun formatBytes(bytes: Long): String = when {
+        bytes >= 1_048_576 -> "%.2f MB".format(bytes / 1_048_576.0)
+        bytes >= 1_024 -> "%.2f KB".format(bytes / 1_024.0)
+        else -> "$bytes B"
+    }
+
+    // endregion
+}

```

</details>

## Release Notes

* **Performance Improvements**
  * Optimized event parsing for faster message and notification delivery.
  * Enhanced support for handling various chat event types including polls, reminders, and AI indicators.

* **Bug Fixes**
  * Improved message enrichment logic for thread notifications to correctly associate messages with their channels.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/MessageAdapter.kt
```diff
@@ -0,0 +1,254 @@
...
+                "pin_expires" -> pinExpires = dateAdapter.fromJson(reader)
+                "pinned" -> pinned = reader.nextBoolean()
+                "pinned_at" -> pinnedAt = dateAdapter.fromJson(reader)
+                "message_text_updated_at" -> messageTextUpdatedAt = dateAdapter.fromJson(reader)
+                "pinned_by" -> pinnedBy = userAdapter.fromJson(reader)
+                "quoted_message" -> {
+                    // Recursive parsing: pass the fallback channel info along
+                    val resolvedChannelInfo = channel ?: fallbackChannelInfo
+                    quotedMessage = fromJson(reader, resolvedChannelInfo)
+                }
```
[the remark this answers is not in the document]
VelikovPetar: done: https://github.com/GetStream/stream-chat-android/pull/6344/commits/06ab23471565c1abd65f1a39ec9a8b3c14239a6c

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
```diff
@@ -0,0 +1,172 @@
...
+                type = channelType,
+                memberCount = channelMemberCount ?: 0,
+                name = channelCustomName,
+                image = channelCustomImage,
+            )
+            message.copy(
+                channelInfo = message.channelInfo ?: fallbackChannelInfo,
+                cid = if (needsCid) cid else message.cid,
+                replyTo = if (needsReplyToCid) replyTo.copy(cid = cid) else replyTo,
+            )
```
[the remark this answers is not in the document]
VelikovPetar: Not applicable, this is the current behaviour in the existing domain mapping.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/PollAdapter.kt
```diff
@@ -0,0 +1,265 @@
...
+        return ParsedVoteDto(
+            id = id,
+            pollId = pollId,
+            optionId = optionId,
+            createdAt = createdAt,
+            updatedAt = updatedAt,
+            user = user,
+            isAnswer = isAnswer,
+            answerText = answerText,
+        )
```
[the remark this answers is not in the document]
VelikovPetar: done: https://github.com/GetStream/stream-chat-android/pull/6344/commits/06ab23471565c1abd65f1a39ec9a8b3c14239a6c

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ReactionGroupParsingTest.kt
```diff
@@ -0,0 +1,168 @@
...
+    fun `Direct path - throws on missing last_reaction_at`() {
+        assertThrows<JsonDataException> {
+            reactionGroupAdapter.parseWithType(
+                com.squareup.moshi.JsonReader.of(
+                    okio.Buffer().writeUtf8(ReactionGroupTestData.jsonMissingLastReactionAt),
+                ),
+                testType,
+            )
+        }
+    }
```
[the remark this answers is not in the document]
VelikovPetar: Not needed, as long as an exception is thrown for the same missing field, we should be ok.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/UserParsingTest.kt
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @Test
+    fun `Direct path - throws on missing online`() {
+        assertThrows<JsonDataException> {
+            userAdapter.fromJson(UserTestData.jsonMissingOnline)
+        }
+    }
+
+    // endregion
```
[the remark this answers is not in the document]
VelikovPetar: done: https://github.com/GetStream/stream-chat-android/pull/6344/commits/06ab23471565c1abd65f1a39ec9a8b3c14239a6c

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/MessageTestData.kt
```diff
@@ -0,0 +1,650 @@
...
+                    "poll_id": "poll-1",
+                    "option_id": "option-1",
+                    "created_at": "2020-01-01T02:00:00.000Z",
+                    "updated_at": "2020-01-01T02:00:00.000Z",
+                    "user": {"id": "user-1", "role": "user", "banned": false, "online": true},
+                    "is_answer": false
+                }
+            ],
+            "created_at": "2020-01-01T00:00:00.000Z",
+            "updated_at": "2020-01-01T03:00:00.000Z",
```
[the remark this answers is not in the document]
VelikovPetar: done: https://github.com/GetStream/stream-chat-android/pull/6344/commits/06ab23471565c1abd65f1a39ec9a8b3c14239a6c

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/MessageReminderInfoParsingTest.kt
```diff
@@ -0,0 +1,114 @@
...
+            parser.fromJson(MessageReminderInfoTestData.jsonMissingUpdatedAt, DownstreamReminderInfoDto::class.java)
+        }
+    }
+
+    @Test
+    fun `Direct path - throws on missing updated_at`() {
+        assertThrows<JsonDataException> {
+            adapter.fromJson(MessageReminderInfoTestData.jsonMissingUpdatedAt)
+        }
+    }
```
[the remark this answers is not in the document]
VelikovPetar: done: https://github.com/GetStream/stream-chat-android/pull/6344/commits/06ab23471565c1abd65f1a39ec9a8b3c14239a6c

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
```diff
@@ -59,8 +60,13 @@ import retrofit2.converter.moshi.MoshiConverterFactory
 internal class MoshiChatParser(
     private val eventMapping: EventMapping,
     private val dtoMapping: DtoMapping,
+    private val directEventParser: DirectEventParser?,
 ) : ChatParser {
 
+    init {
+        Log.d("X_PETAR", "DEP: $directEventParser")
```
gpunto: There's some debug logging left here and below :P

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/ReactionGroupAdapter.kt
```diff
@@ -0,0 +1,158 @@
...
+ *         reader.skipValue()
+ *     }
+ * }
+ * ```
+ *
+ * Do NOT use the standard [fromJson] method - it will throw [UnsupportedOperationException].
+ */
+internal class ReactionGroupAdapter(
+    private val dateAdapter: JsonAdapter<Date>,
+) : JsonAdapter<ReactionGroup>() {
```
gpunto: Does this need to be a JsonAdapter if it can't be used as such? 🤔 I'd make it a simple class without any parent, so it can't be used as an adapter by mistake

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/UserAdapter.kt
```diff
@@ -0,0 +1,135 @@
...
+
+                // OwnUserResponse fields — parsed to maintain parity with the DTO path.
+                // DTO declares `Int = 0`: default applies when field is absent; null throws.
+                // reader.nextInt() also throws on null, matching the DTO behavior.
+                "total_unread_count" -> totalUnreadCount = reader.nextInt()
+                "unread_channels" -> unreadChannels = reader.nextInt()
+                "unread_threads" -> unreadThreads = reader.nextInt()
+                // The following are not part of the UserResponse (they are part of OwnUserResponse):
+                // This is an intentional change from DownstreamUserDto which covers both UserResponse/OwnUserResponse
+                "mutes", "channel_mutes", "push_preferences" -> reader.skipValue()
```
gpunto, line 94: What is the downside of still trying to parse and assign these three? I'm thinking of a case where we extend the parsing to more models and we start silently dropping these.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
```diff
@@ -0,0 +1,172 @@
...
+                "cid" -> cid = reader.nextString()
+                "channel_member_count" -> channelMemberCount = JsonParsingUtils.readNullableInt(reader)
+                "channel_custom" -> {
+                    val (name, image) = parseChannelCustom(reader)
+                    channelCustomName = name
+                    channelCustomImage = image
+                }
+                "channel_type" -> channelType = reader.nextString()
+                "channel_id" -> channelId = reader.nextString()
+                "message" -> message = messageAdapter.fromJson(reader)
```
gpunto, line 81: Here we use `fromJson`, but MessageAdapter has `fromJson(reader, fallbackChannelInfo)` too. Unless I'm missing something, I think atm `fallbackChannelInfo` is always null.

When the parent has no channel, does `replyTo.channelInfo` end up null here? Is that a valid scenario? If so, we could extend the enrichment below to also handle missing `replyTo.channekInfo`. In that case, probably we can drop the `fromJson(reader, fallbackChannelInfo)` overload.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
```diff
@@ -141,8 +147,13 @@ internal class MoshiChatParser(
 
     private val chatEventDtoAdapter = moshi.adapter(ChatEventDto::class.java)
 
-    private fun parseAndProcessEvent(raw: String): ChatEvent = with(eventMapping) {
-        val event = chatEventDtoAdapter.fromJson(raw)!!.toDomain()
-        return event.enrichIfNeeded()
+    private fun parseAndProcessEvent(raw: String): ChatEvent {
+        val directEvent = directEventParser?.parse(raw)
+        if (directEvent != null) {
+            Log.d("X_PETAR", "Event from DEP: ${directEvent.type}")
```
andremion: [Blocker] Debug logs left in the code.

There are two `Log.d("X_PETAR", ...)` calls and one `android.util.Log` import that need to be removed before merging:

1. The `init { Log.d("X_PETAR", "DEP: $directEventParser") }` block (line 66).
2. The `Log.d("X_PETAR", "Event from DEP: ${directEvent.type}")` call inside `parseAndProcessEvent` (line 153).

The second one runs for every parsed event. On the livestream hot path that this PR is meant to optimize, it would cancel most of the performance gain.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
```diff
@@ -0,0 +1,172 @@
...
+                cid = if (needsCid) cid else message.cid,
+                replyTo = if (needsReplyToCid) replyTo.copy(cid = cid) else replyTo,
+            )
+        } else {
+            message
+        }
+
+        return NewMessageEvent(
+            type = type,
+            createdAt = createdAt ?: Date(0),
```
andremion, line 141: [Blocker] Malformed `created_at` is silently replaced with `Date(0)`.

`StreamDateFormatter.parse` returns null for unparseable input. The required-field check runs on `rawCreatedAt` (the string), so a JSON like `"created_at": "garbage"` passes validation. Then `createdAt = createdAt ?: Date(0)` substitutes the epoch silently.

The DTO path's `ExactDateAdapter` returns null for unparseable input, and then Moshi throws `JsonDataException` because the field is non-nullable.

Suggested fix: validate `createdAt` (not `rawCreatedAt`) with `requireField`, or throw `JsonDataException` explicitly when `rawCreatedAt != null && createdAt == null`.

Please also add a parity test with a malformed date string for both paths.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/MessageAdapter.kt
```diff
@@ -0,0 +1,274 @@
...
+        JsonParsingUtils.requireField(text, "text", reader)
+        JsonParsingUtils.requireField(type, "type", reader)
+        JsonParsingUtils.requireField(updatedAt, "updated_at", reader)
+        JsonParsingUtils.requireField(user, "user", reader)
+
+        val resolvedChannelInfo = channel ?: fallbackChannelInfo
+
+        // Enrich quoted message with parent's resolved channelInfo (matching DTO toDomain behavior
+        // where parent passes its resolved channelInfo as fallback to quoted_message.toDomain())
+        val enrichedQuotedMessage = quotedMessage?.let { qm ->
```
andremion, line 195: [Optional] `channelInfo` propagation depth.

Commit 06ab234 already addressed the field-ordering case for the direct `quoted_message`. There is still a remaining gap when nesting is deeper than one level.

The DTO path threads `channelInfo` recursively through `quoted_message?.toDomain(channelInfo)`. The current direct adapter only enriches the direct `quotedMessage` post-loop. A two-deep chain (`message → quoted_message → quoted_message`) where the inner two messages have no `channel` field will result in the innermost message having a null `channelInfo`, while the DTO path would propagate the outer message's channel down two levels.

The same applies to `enrichWithCid`: the adapter does `replyTo.copy(cid = cid)` for one level, but `Message.enrichWithCid` is recursive: `copy(replyTo = replyTo?.enrichWithCid(newCid), cid = newCid)`.

Two options:
- Replicate the full recursion.
- Document the one-level depth limit and add a regression test with a two-deep `quoted_message` to lock in the documented behavior.

I assume two-deep `quoted_message` is rare in practice, so optional rather than blocker.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/UserAdapter.kt
andremion, line 94: [Optional] PR description claim about identical output should be qualified.

The inline comment correctly documents that `mutes`, `channel_mutes`, and `push_preferences` are intentionally skipped because they are `OwnUserResponse` fields, not `UserResponse` fields. That part is fine.

The PR description still says the direct path "produces domain objects identical to the existing DTO-based path". If the backend ever includes any of these fields in a `message.new` user payload, the DTO path would populate `User.mutes`, `User.channelMutes`, and `User.pushPreference`, while the direct path would leave them empty.

Two options:
- Adjust the PR description to qualify the identity claim (e.g., "identical for `message.new` user payloads in the current backend shape").
- Add a regression test using a JSON that includes these fields, asserting the documented difference, so the carve-out cannot silently break parity later.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/DirectEventParser.kt
```diff
@@ -0,0 +1,147 @@
...
+                            } else {
+                                it.nextString()
+                            }
+                        } else {
+                            it.skipValue()
+                        }
+                    }
+                    null
+                }
+            } catch (@Suppress("TooGenericExceptionCaught", "SwallowedException") _: Exception) {
```
andremion: [Optional] `extractType` swallows exceptions with no log.

Returning null on malformed JSON correctly falls through to the DTO path, which is probably the right behavior. But there is no log at all. If a malformed `message.new` ever slips through, the DTO path will throw later and the failure will not point at the direct path.

Consider adding a verbose log inside the catch so this is debuggable.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/config/StatePluginConfig.kt
```diff
@@ -161,3 +167,54 @@ public data class ChannelMessageLimit(
...
+ *   that the underlying socket listener uses non-suspending `tryEmit`, so with `SUSPEND` an
+ *   overflowing emission is simply reported as a failed emit rather than blocking the socket.
+ * - [BufferOverflow.DROP_OLDEST]: the oldest queued event is evicted to make room for the new
+ *   one. Useful for live channels where freshness matters more than completeness.
+ * - [BufferOverflow.DROP_LATEST]: the newest event is discarded and the queued events are kept.
+ */
+public data class MessageBufferConfig(
+    public val channelTypes: Set<String> = emptySet(),
+    public val capacity: Int = Int.MAX_VALUE,
+    public val overflow: BufferOverflow = BufferOverflow.SUSPEND,
```
andremion: [Optional] `SUSPEND` as the default overflow strategy is misleading.

The KDoc explains that the socket listener uses `tryEmit`, so `SUSPEND` behaves as "drop and log error" rather than actually suspending. This means the default behavior for an opt-in performance feature is silently lossy for any caller who sets `channelTypes` but does not also set `overflow`.

Two options:
- Default to `DROP_OLDEST`, which matches the livestream use case the KDoc describes.
- Remove `SUSPEND` from the public API since it cannot be honored on this code path.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChatEvent.kt
```diff
@@ -43,5 +44,6 @@ public fun ChatEvent.enrichIfNeeded(): ChatEvent = when (this) {
     is ChannelTruncatedEvent -> copy(message = message?.enrichWithCid(cid))
     is ChannelUpdatedByUserEvent -> copy(message = message?.enrichWithCid(cid))
     is NotificationMessageNewEvent -> copy(message = message.enrichWithCid(cid))
+    is NotificationThreadMessageNewEvent -> copy(message = message.enrichWithCid(cid))
```
andremion, line 47: [Optional] Unrelated bug fix bundled into this PR.

Adding `NotificationThreadMessageNewEvent` to `enrichIfNeeded()` is a behavior fix that is not related to fast event parsing. The CodeRabbit summary mentions it briefly, but it deserves its own commit (or its own PR) with a one-line description of:
- what was broken without it,
- how the bug was found,
- a dedicated regression test.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ParserFactory.kt
```diff
@@ -25,6 +25,7 @@ internal object ParserFactory {
     fun createMoshiChatParser(
         currentUserIdProvider: () -> String = { "" },
         apiModelTransformers: ApiModelTransformers = ApiModelTransformers(),
+        fastEventParsing: Boolean = true,
```
andremion: [Nit] Default flipped to `fastEventParsing = true`.

Changing the default to `true` means every existing parser test now exercises the direct path implicitly. This catches regressions, but it also means that a failure in any parser test could be caused by the new direct path, which is confusing.

Two options:
- Keep the default at `true` and add a class-level comment in `ParserFactory.kt` that explains this.
- Default to `false`, and have only `FastEventParsingToggle` and the parity tests opt in explicitly.

andremion:
[Blocker] Two features in one PR.

The title and description only mention fast event parsing, but this PR also adds a new public `MessageBufferConfig` API and a parallel buffered `MutableSharedFlow` path in `EventHandlerSequential`. This is a separate opt-in feature (bounded buffer for `NewMessageEvent` on selected channel types).

Could you either:
- Split it into a second PR, or
- Update the title and description to describe both features and explain why they ship together.

Reviewing them as one change makes it harder to evaluate regression risk for each feature on its own.

andremion:
[Nit] Benchmark patch imports the wrong package.

The patch in the PR description uses imports like `io.getstream.chat.android.client.parser2.event.AttachmentAdapter`, but the actual package in this PR is `parser2.direct.*`. The patch will not apply cleanly as written.

andremion:
[Optional] Coverage scope versus the livestream goal.

The fast path covers `message.new` only. For a typical livestream traffic mix, what proportion of events is `message.new` versus other event types that still go through the DTO path? This determines how much the per-event gain translates into overall hot-path throughput.

If `message.new` is the dominant event, this PR delivers most of the win on its own. If not, follow-up PRs covering `reaction.new`, `message.updated`, `typing.*`, and `notification.message_new` will be needed to fully achieve the livestream performance goal.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
VelikovPetar: oops, will remove

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
VelikovPetar: Nice catch, will check it out.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/config/StatePluginConfig.kt
VelikovPetar: Ah this is part of a different branch, will remove this commit from here

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ParserFactory.kt
VelikovPetar: Will update!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
```diff
@@ -0,0 +1,172 @@
...
+                "channel_custom" -> {
+                    val (name, image) = parseChannelCustom(reader)
+                    channelCustomName = name
+                    channelCustomImage = image
+                }
+                "channel_type" -> channelType = reader.nextString()
+                "channel_id" -> channelId = reader.nextString()
+                "message" -> message = messageAdapter.fromJson(reader)
+                "watcher_count" -> watcherCount = reader.nextInt()
+                "total_unread_count" -> totalUnreadCount = reader.nextInt()
```
gpunto, line 83: `DownstreamUserDto` also has `unread_count`, what about that one?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
VelikovPetar: Removed: https://github.com/GetStream/stream-chat-android/pull/6344/commits/af32544ebfdcc5cb1e5e369c2e7b98f652e44c89

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/config/StatePluginConfig.kt
VelikovPetar: Reverted the whole commit here: https://github.com/GetStream/stream-chat-android/pull/6344/changes/c5332863b8ceab995815bf7ffd147e8436395a43

This is part of a different PR, but I added on this branch to provide a snapshot build for a customer.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/ChatEvent.kt
VelikovPetar: Not related - just a small clean-up I found. I can remove it, but I don't really see an issue in shipping this.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ParserFactory.kt
VelikovPetar: Done: https://github.com/GetStream/stream-chat-android/pull/6344/changes/6bf8637297e33cd1f6d0b5797020134790aecf2b

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/DirectEventParser.kt
VelikovPetar: Added a log here: https://github.com/GetStream/stream-chat-android/commit/6bf8637297e33cd1f6d0b5797020134790aecf2b

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/UserAdapter.kt
VelikovPetar: The thing here is that the `Message` object arrives with a `user` of type: `UserResponse` (you can check the definition [here](https://getstream.github.io/protocol/#/product%3Achat)). The `UserResponse` doesn't include the following fields: `mutes`, `channelMutes` or `pushPreference`. These fields are only included in the `OwnUserResponse`, which we receive on successful user connection. The problem in the SDK is that we user the same DTO/Domain object for both response types. 
In the current state, we would never receive a payload in which these field will be populated. But in order to parse them using this approach, we have to introduce several more mappers (including one for `Channel`), which is a massive addition, but is essentially dead code. So I decided to completely skip them. 

I will add a test, but tbh not sure how useful it would be. We will be verifying that the default parser successfully parses, and the new one ignores them - but not sure how that prevents any regressions in the future.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/MessageAdapter.kt
VelikovPetar: I documented the depth limit. I will test this tomorrow, but I don't think we can receive more 1 level of nested messages in an event.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
VelikovPetar: Added a stricter check: https://github.com/GetStream/stream-chat-android/pull/6344/changes/6bf8637297e33cd1f6d0b5797020134790aecf2b#diff-37f7bb38314c31b89f6c0b5787afb6ef1756d13ceafe5dc35153bff16a990a64

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
VelikovPetar: Removed already: https://github.com/GetStream/stream-chat-android/pull/6344/changes/af32544ebfdcc5cb1e5e369c2e7b98f652e44c89

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/ReactionGroupAdapter.kt
VelikovPetar: We can probably refactor it - I will double check it.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/UserAdapter.kt
VelikovPetar: We need to introduce several more adapters for what would be essentially dead code. I left some additional info in this [comment](https://github.com/GetStream/stream-chat-android/pull/6344#discussion_r3276101058) as well.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
VelikovPetar: This code went over several iteration, it might be redundant. I will double check.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
VelikovPetar: There was a reason I skipped this, but it was a long time ago when I implemented this, I will have to double check.

VelikovPetar:
> [Nit] Benchmark patch imports the wrong package.
> 
> The patch in the PR description uses imports like `io.getstream.chat.android.client.parser2.event.AttachmentAdapter`, but the actual package in this PR is `parser2.direct.*`. The patch will not apply cleanly as written.

I most likely added a patch from an older iteration of the benchmark, I will update the patch.

VelikovPetar:
> [Optional] Coverage scope versus the livestream goal.
> 
> The fast path covers `message.new` only. For a typical livestream traffic mix, what proportion of events is `message.new` versus other event types that still go through the DTO path? This determines how much the per-event gain translates into overall hot-path throughput.
> 
> If `message.new` is the dominant event, this PR delivers most of the win on its own. If not, follow-up PRs covering `reaction.new`, `message.updated`, `typing.*`, and `notification.message_new` will be needed to fully achieve the livestream performance goal.

The `message.new` is the most problematic one as per customer reports. For now, I think this is the most important event to cover. Hopefully, once we introduce OpenApi to the project in some of the next majors, we would get the same optimisation for all other events.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/UserAdapter.kt
andremion: Thanks for the explanation. That makes sense, so let's skip it.

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/MessageAdapter.kt
VelikovPetar: Just tested this: Only one level of quotes is delivered via `message.new`. The `message.quoted_message` references its own quoted message just by setting the `message.quoted_message.quoted_message_id`.

Also confirmed by docs: https://getstream.io/chat/docs/android/threads/#inline-replies

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
VelikovPetar: Ah this is not used even in the standard DTO path - the `User` object doesn't hold this value.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
VelikovPetar: I did some updates [here](https://github.com/GetStream/stream-chat-android/pull/6344/changes/e89bd0821338188e8943135616e07ec3ccb0740e), please take a look!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/ReactionGroupAdapter.kt
VelikovPetar: Updated: https://github.com/GetStream/stream-chat-android/pull/6344/changes/e89bd0821338188e8943135616e07ec3ccb0740e

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/NewMessageEventAdapter.kt
gpunto: Got it, thanks!

gpunto (approved):

PR #6358, opened 2026-04-10
Fix mention suggestion filter to align with Swift SDK

andremion:
### Goal

Fix the mention suggestion filter (`@mentions`) which returned unrelated users for short queries. For example, typing `@jc` in a channel with "Aleksandar Apostolov" and "Jc Minarro" showed both users instead of only "Jc Minarro". Aligns the filtering behavior with the Swift SDK.

### Implementation

**Replace fuzzy Levenshtein filtering with substring matching (`DefaultUserQueryFilter`):**
- The old `DefaultQueryFilter` filtered by Levenshtein distance < threshold, which was too permissive for short queries (e.g. distance between "jc" and "al" is 2, below the threshold of 3).
- Now filters by normalized substring containment and sorts matches by Levenshtein distance so the closest results appear first. This matches the Swift SDK approach.

**Merge `DefaultQueryFilter` into `DefaultUserQueryFilter`:**
- `DefaultQueryFilter<T>` was a generic internal class only ever used with `User`. Merged into `DefaultUserQueryFilter` (the public class) to eliminate the unnecessary abstraction.
- Moved `DefaultUserQueryFilter` from `mention` package to `query.filter` package where it belongs alongside `QueryFilter`.

**Reduce visibility of internal classes:**
- `DefaultQueryFilter` → removed (merged into `DefaultUserQueryFilter`)
- `Combine`, `Transliterate` → `internal` (only used by `DefaultUserQueryFilter`)
- `LocalUserLookupHandler` → `internal` (no external consumers)

**Fix handler defaults:**
- `LocalUserLookupHandler` and `DefaultUserLookupHandler` now use `DefaultUserQueryFilter()` as default instead of creating `DefaultQueryFilter` directly, ensuring consistent behavior.

### Testing

1. Open a channel with multiple members in the sample app.
2. Type `@` followed by a short query (e.g. `@jc`, `@al`, `@ale`) — verify only users whose name or id contains the typed substring are shown.
3. Verify that case and diacritics are ignored (e.g. `@jose` matches "José").
4. Verify that results are sorted by relevance (exact name match appears before a longer name containing the substring).
5. Run `./gradlew :stream-chat-android-ui-common:testDebugUnitTest` — all unit tests pass.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized mention filtering implementation into a dedicated package structure for improved code organization.
  * Reduced public API surface by making internal helper classes private to the module.

* **Tests**
  * Added comprehensive test coverage for mention filtering logic, including edge cases and matching behavior validation.


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/query/filter/DefaultUserQueryFilter.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    override fun filter(items: List<User>, query: String): List<User> {
+        logger.d { "[filter] query: \"$query\", items.size: ${items.size}" }
+        val formattedQuery = queryFormatter.format(query)
+        if (formattedQuery.isEmpty()) return items
+        return items
+            .mapNotNull { user ->
+                val formattedName = queryFormatter.format(query = user.name.ifBlank(user::id))
+                if (formattedName.contains(formattedQuery)) {
+                    user to levenshteinDistance(formattedQuery, formattedName)
```
gpunto: I'm a bit unsure about this, but I'm probably missing something.

Here we're saying that the formatted name already contains the formatted query, so what's the point of the distance? The typed characters are already fully contained in the user's name, so isn't the distance calculation basically just degenerating to the difference between the length of `formattedName` and `formattedQuery`? 🤔

andremion: Good catch!
It seems "overkill".
This is what the Swfit does:
```swift
    let distance: (ChatUser) -> Int = {
        min($0.id.levenshtein(searchInput), $0.name?.levenshtein(searchInput) ?? 1000)
    }
```
We could simplify to `sortedBy { name.length }`.

gpunto: Yep, or even `indexOf` if we want to prefer names where the match is earlier, e.g. query="ann" and we want to see "anne" before "johann"

andremion: 0bd9efe8ce0189d530a5e458b85cdfdfae0ed5c2

gpunto (approved):

PR #6362, opened 2026-04-13
Remove read count from message read indicator

gpunto:
### Goal

Remove the read count from the message read indicator as per [thread](https://getstream.slack.com/archives/C0A3JU6C6G3/p1771235361787799).

### Implementation

- Remove the `readCount` parameter from `MessageReadStatusIcon` and the `readCountEnabled` config from `MessageListConfig`
- Replace `IsReadCount` composable (count text + icon in a `Row`) with a simpler `IsReadIcon` (icon only)
- Update snapshot tests

### UI Changes

Please check the updated snapshots

### Testing

Open a group chat and verify that the read count is not displayed anymore

## Summary by CodeRabbit

* **Refactor**
  * Simplified message read status indicators by removing read count display
  * Removed `readCountEnabled` configuration option from message list settings
  * Updated read receipt UI to show only basic read status instead of detailed recipient counts


andremion (commented):
You might need to update some references to `Stream_MessageReadCount` in e2e tests

andremion (approved):

PR #6363, opened 2026-04-14
Fix link preview resolution and skipEnrichUrl in message composer

andremion:
### Goal

Fix two issues with link preview resolution in the message composer:

1. **Ghost preview after send** — when the user sends a message before the link preview resolves, the in-flight enrichment completes and writes a stale preview into the already-cleared composer.
2. **Missing preview on sent message** — `skipEnrichUrl` was derived from `linkPreviews.isEmpty()`, conflating "preview hasn't resolved yet" with "no preview wanted." This told the backend to skip enrichment when the user simply sent before the preview loaded, resulting in no link preview on the message.

Both fixes align the Android SDK with the iOS SDK's behavior.

### Implementation

**`TestCoroutineExtension`**
- Add `CallDispatcherProvider.set/reset` alongside the existing `DispatcherProvider` setup. This was a missing counterpart — `Call.await()` (via `MapCall`) was hopping to real `Dispatchers.IO` in tests.

**`MessageComposerController`**
- Track `linkPreviewJob` and cancel it in both the debounce site (prevents duplicate in-flight calls) and `clearData()` (prevents ghost preview after send).
- Replace the `skipEnrichUrl = linkPreviews.isEmpty()` derivation with an intent-based approach: a `linkPreviewDismissed` flag is set only when the user taps the dismiss button (`cancelLinkPreview`), and reset on any text change (`setMessageInputInternal`). At send time, `shouldSkipEnrichUrl()` combines the integrator override (`message.skipEnrichUrl` from `ChannelScreen`), user dismissal, and URL presence in the text.
- Apply `shouldSkipEnrichUrl()` to both the new-message and edit-message paths.
- Deduplicate `setMessageInput` by delegating to `setMessageInputInternal`.

**`MessageComposerControllerTest`**
- Test that in-flight link preview resolution does not leak into state after `clearData`.
- Test that `skipEnrichUrl` is `false` when sending before the preview resolves (backend should enrich).
- Test that `skipEnrichUrl` is `true` after explicit preview dismissal.
- Test that text changes reset the dismissal (backend should enrich again).

No UI changes.

### Testing

1. Open a channel in the Compose sample.
2. Type a URL (e.g. `www.meta.com`), wait for the link preview to appear, then tap Send **before** the preview shows — verify the sent message has a link preview (backend enriched it).
3. Type a URL, wait for the preview, tap the X to dismiss it, then Send — verify the message has **no** link preview.
4. Type a URL, dismiss the preview, type a different URL, Send — verify the message has the new URL's preview.
5. Type a URL, send before preview resolves — verify no ghost preview appears in the empty composer after send.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link preview handling: dismissed previews are remembered, in-flight preview fetches are cancelled on input changes, and enrichment is skipped when appropriate.
* **Refactor**
  * Composer UI now surfaces a single link preview (rather than a list), simplifying preview display.
* **Tests**
  * Added tests covering preview timing, dismissal behavior, and skip-enrichment scenarios.


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -479,12 +491,12 @@ public class MessageComposerController(
...
      */
     public fun setMessageInput(value: String) {
-        if (_messageInput.value.text == value) return
-        _messageInput.value = MessageInput(value, MessageInput.Source.External)
+        setMessageInputInternal(value, MessageInput.Source.External)
     }
 
     private fun setMessageInputInternal(value: String, source: MessageInput.Source) {
         if (_messageInput.value.text == value) return
+        linkPreviewDismissed = false
```
gpunto: If the user adds a link -> dismisses -> and then continues to type, we would reset the flag, right? As a user, I'd expect the preview to be dismissed for good 🤔

andremion: Good catch!
I will push a fix for that

andremion: Fix: 8d7df01d559ff80275abb0b519c180e25dd4907e
Refactor to align API with iOS: 440ee661f148ad451cbc759c3eb41e3b6c1336a1

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -479,12 +492,14 @@ public class MessageComposerController(
...
      */
     public fun setMessageInput(value: String) {
-        if (_messageInput.value.text == value) return
-        _messageInput.value = MessageInput(value, MessageInput.Source.External)
+        setMessageInputInternal(value, MessageInput.Source.External)
     }
 
     private fun setMessageInputInternal(value: String, source: MessageInput.Source) {
         if (_messageInput.value.text == value) return
+        if (dismissedLinkPreviewUrl != null && LinkPattern.find(value)?.value != dismissedLinkPreviewUrl) {
```
gpunto: I think this exposes us to a new edge case, i.e.
1. type url 1
2. dismiss url 1
3. type url 2
4. dismiss url 2

We end up overwriting `dismissedLinkPreviewUrl`, and with `find` we're only checking the first match.

We could keep a set of all the dismissed urls, find all matches in the string, and check against the set. But I'm also a bit worried we're doing a lot of work on each keystroke.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -1134,19 +1155,16 @@ public class MessageComposerController(
     }
 
     /**
-     * Shows link previews if necessary.
+     * Resolves and displays the link preview for the first URL in the current input.
      */
-    private suspend fun handleLinkPreviews() {
+    private suspend fun handleLinkPreview() {
```
gpunto, line 1163: Shouldn't we check here for the dismissed preview before enriching again? Otherwise I can dismiss a url, then type, and the preview re-appears in the composer.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
andremion: That's fine because we only show the first preview anyway. So, in practice, no issue happens, and a Set would be over-engineering, I guess.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
andremion: 19832a9d87e8d673ba9afe0feaadbbe318d6447c

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
andremion: I manually tested and couldn't reproduce that edge case. Let me know if I missed something

gpunto (approved):

andremion:
@coderabbitai review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
gpunto: I think I was seeing the effect of [this](https://github.com/GetStream/stream-chat-android/pull/6363#discussion_r3079845540), which is now fixed, so nothing to do here!

PR #6365, opened 2026-04-14
Trim composable parameters from public API

gpunto:
### Goal

Trim the public API surface of the Compose SDK by removing composable slot parameters (e.g. `itemContent`), styling parameters (e.g. `color`), and forwarded callback parameters from public composables. Customization should go through `ChatComponentFactory` instead.

### Implementation

- Remove composable slot parameters from `ChannelList`, `MessageList`, `PinnedMessageList`, `ThreadList`, and `MessageContent`
- Remove styling parameters from `ChannelListHeader`, `ChannelHeader`, `PollCreationHeader`
- Remove forwarded callback parameters from `MessageList` and `ChannelScreen` (e.g. `onMessageLinkClick`, `onUserAvatarClick`, `onUserMentionClick`, poll callbacks)
- Make `Messages` and `Channels` composables internal
- Move `reactionSorting` from `MessageList` parameters to `ChatUiConfig`

### Testing

Everything should be working as before

## Summary by CodeRabbit

* **Refactor**
  * Simplified ChannelList/MessageList/ThreadList: removed many per-item and slot customization points; customization now via theme component factory.
  * Header components now use fixed theme styling (color/shape/elevation removed).
  * Reaction sorting removed from many message APIs — configure via global message list config.
  * ChannelScreen API simplified (fewer visibility flags and callbacks).

* **Chores**
  * Some composables made internal/module-private.

* **Docs**
  * Samples updated to demonstrate theme-based customization.


gpunto:
@coderabbitai review

gpunto:
@coderabbitai review

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt
```diff
@@ -87,12 +85,8 @@ class MessagesActivity : ComponentActivity() {
     private fun SetupContent() {
         ChannelScreen(
             viewModelFactory = factory,
-            reactionSorting = ReactionSortingByLastReactionAt,
             onBackPressed = { finish() },
             onChannelAvatarClick = ::openChannelInfo,
-            onMessageLinkClick = { _, link ->
```
andremion, line 93: That means that the SDK now handles link clicks automatically, right? Is that intentional and aligned with iOS?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/threads/ThreadList.kt
```diff
@@ -165,57 +117,61 @@ public fun ThreadList(
...
-                    state.threads.isEmpty() -> emptyContent()
-                    else -> Threads(
-                        threads = state.threads,
-                        isLoading = state.isLoading,
-                        isLoadingMore = state.isLoadingMore,
-                        modifier = Modifier,
-                        onLoadMore = onLoadMore,
-                        itemContent = itemContent,
-                        loadingMoreContent = loadingMoreContent,
+            val contentModifier = Modifier.padding(padding)
```
andremion: I believe this has broken the UI.
Should we `fillMaxSize`?

Image

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollAnswers.kt
```diff
@@ -173,7 +170,7 @@ private fun Content(
                 .padding(top = StreamTokens.spacing2xl),
             verticalArrangement = Arrangement.spacedBy(StreamTokens.spacingMd),
         ) {
-            val showAvatar = (poll.votingVisibility == VotingVisibility.PUBLIC) || showAnonymousAvatar
+            val showAvatar = poll.votingVisibility == VotingVisibility.PUBLIC
```
andremion, line 173: It was weird to show avatars in an anonymous poll.
I hope no customer expects that 😅

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/MessagesActivity.kt
gpunto: So yes, the default is that we handle them, but the behavior can still be overridden by overriding `MessageItem` in the component factory and passing a custom `onLinkClick` (`copy`ing the params`. Basically most handlers can be customized like that, instead of passing them down layer after layer from the top level components.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/threads/ThreadList.kt
gpunto: I thought I fixed this 🤦‍♂️ let me see

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/poll/PollAnswers.kt
gpunto: Definitely! Also it's not a valid use case in the designs, so it was my original mistake because I forgot to remove this parameter when I redesigned polls

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/threads/ThreadList.kt
gpunto: Done!

andremion (approved):

PR #6366, opened 2026-04-14
Fix `ChatTheme` crashing previews when `ChatClient` is not initialized

VelikovPetar:
### Goal

Fix `ChatTheme` crashing with `ChatClient is not initialized` in Compose previews and VRT tests. This was caused by `ChatClient.instance().cdn` being called eagerly at composition time, which fails when `ChatClient` hasn't been built yet.

### Implementation

- Changed `CDNImageInterceptor` to resolve the `CDN` lazily via a `() -> CDN?` provider instead of taking a direct `CDN` instance. The default provider safely checks `ChatClient.isInitialized` before accessing the instance.
- `ChatTheme` now always installs the `CDNImageInterceptor` (with the lazy default). When `ChatClient` is not initialized, the interceptor is a no-op pass-through. When it is initialized later, the next image request picks up the CDN automatically.
- Same change applied to `StreamCoil` (XML view system path).
- Added test for the null-CDN pass-through case.
- Updated KDoc for `imageLoaderFactory` parameter to document CDN + interceptor interaction.

### UI Changes

No UI changes.

### Testing

- Existing `CDNImageInterceptorTest` tests updated and passing.
- New test `intercept passes through when cdn returns null` covers the preview/VRT scenario.
- Manual testing: verified custom single-arg `StreamCoilImageLoaderFactory` still works correctly (interceptors silently dropped by default two-arg fallback, customer's loader returned as-is).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of image loading with CDN functionality and custom image headers during app initialization.

* **Documentation**
  * Updated guidance for implementing custom image loaders with CDN or custom header configurations.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
@@ -308,9 +308,11 @@ private val LocalMediaGalleryConfig = compositionLocalOf<MediaGalleryConfig> {
  * @param durationFormatter [DurationFormatter] Used to format durations in the app.
  * @param channelNameFormatter [ChannelNameFormatter] Used throughout the app for channel names.
  * @param messagePreviewFormatter [MessagePreviewFormatter] Used to generate a string preview for the given message.
- * @param imageLoaderFactory A factory that creates new Coil [ImageLoader] instances. If used in combination with
- * [asyncImageHeadersProvider] you must override the [StreamCoilImageLoaderFactory.imageLoader] method accepting the
- * interceptors parameter.
+ * @param imageLoaderFactory A factory that creates new Coil [ImageLoader] instances. If you provide a custom factory
+ * **and** use a custom CDN (via [io.getstream.chat.android.client.ChatClient.Builder]) or
+ * [asyncImageHeadersProvider], you must override the [StreamCoilImageLoaderFactory.imageLoader] overload that accepts
+ * interceptors; otherwise those features are silently ignored. If you don't use either, overriding the single-arg
+ * method is sufficient.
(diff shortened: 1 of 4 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
```diff
@@ -458,17 +460,12 @@ public fun ChatTheme(
     }
 
     val context = LocalContext.current
-    val cdn = remember { ChatClient.instance().cdn }
```
andremion, line 461: I believe this is already fixed by https://github.com/GetStream/stream-chat-android/pull/6341/changes#diff-43ace81d3100eafa4178eb5eaa2a91790c0e8d6dc15cc2474691716bb38e5512R219 🤔

gpunto: Not on develop yet, but I guess we're going to merge v7 into develop tomorrow if we release

VelikovPetar: This was actually a bug report by a customer, whose [roborazzi](https://github.com/takahirom/roborazzi) tests started failing after introducing this. 
I am not sure that the `LocalInspectionMode` check would be enough to cover those cases, because roborazzi should run on Robolectric, in which I believe `LocalInspectionMode` would be `false`.

andremion: In that case. No need for the v7 approach. We can use this approach in v7 instead.

andremion (approved):

PR #6367, opened 2026-04-15
Group ChannelViewModelFactory params into classes

gpunto:
### Goal

Reduce the parameter count of `ChannelViewModelFactory` by grouping related parameters into dedicated configuration classes.

### Implementation

- Introduce `MessageListOptions` and `ComposerOptions` grouping message list and composer parameters
- Update all usages

### Testing

Everything should be working as before

## Summary by CodeRabbit

* **Refactor**
  * Reorganized message and composer configuration settings into structured options objects. MessageListOptions consolidates message display, threading, separator, and loading behavior settings, while ComposerOptions groups attachment limits and feature enablement settings. This change provides a cleaner, more organized approach to managing messaging customizations while maintaining full compatibility with existing implementations.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/ChannelViewModelFactory.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/ChannelViewModelFactory.kt
@@ -78,13 +78,13 @@ public data class MessageListOptions(
  * Configuration for the message composer behavior.
  *
  * @param maxAttachmentCount The maximum number of attachments that can be sent in a single message.
- * @param isLinkPreviewEnabled If the link preview is enabled in the composer.
- * @param isDraftMessageEnabled If the draft message is enabled in the composer.
+ * @param linkPreviewEnabled If the link preview is enabled in the composer.
+ * @param draftMessagesEnabled If draft messages are enabled in the composer.
  */
 public data class ComposerOptions(
     val maxAttachmentCount: Int = AttachmentConstants.MAX_ATTACHMENTS_COUNT,
-    val isLinkPreviewEnabled: Boolean = false,
-    val isDraftMessageEnabled: Boolean = true,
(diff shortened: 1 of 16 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/messages/ChannelViewModelFactory.kt
```diff
@@ -47,6 +47,46 @@ import kotlinx.coroutines.SupervisorJob
...
+ * Configuration for the message composer behavior.
+ *
+ * @param maxAttachmentCount The maximum number of attachments that can be sent in a single message.
+ * @param isLinkPreviewEnabled If the link preview is enabled in the composer.
+ * @param isDraftMessageEnabled If the draft message is enabled in the composer.
+ */
+public data class ComposerOptions(
+    val maxAttachmentCount: Int = AttachmentConstants.MAX_ATTACHMENTS_COUNT,
+    val isLinkPreviewEnabled: Boolean = false,
+    val isDraftMessageEnabled: Boolean = true,
```
VelikovPetar: What do you think about renaming this flag to `draftMessagesEnabled` - we would be more aligned with iOS, and `isDraftMessageEnabled` sounds a bit weird

Edit: Maybe we can change `isLinkPreviewEnabled` to `linkPreviewEnabled` for consistency. (`MessageListOptions` also doesn't use `is-` prefix for boolean flags)

gpunto: sounds good!

gpunto: Done! I've also removed the prefix from the booleans in `MessageComposerController.Config` for consistency, but it's `@InternalStreamChatApi` anyway.

aleksandar-apostolov (approved):

PR #6377, opened 2026-04-17
Fix attachments lost on activity recreation

VelikovPetar:
### Goal

Fix selected composer attachments being lost when the activity is destroyed and recreated (e.g., "Don't keep activities" developer option, process death). Previously, all in-memory composer state was lost on activity recreation since `MessageComposerController` stored everything in plain `MutableStateFlow` fields.

### Implementation

Introduced a `ComposerStateSaver` abstraction that persists selected attachments via `SavedStateHandle`:

- **`ComposerStateSaver`** — framework-agnostic interface the controller uses for save/restore (no Android imports in the controller)
- **`SavedStateComposerStateSaver`** — `SavedStateHandle`-backed implementation injected by the ViewModel factory
- **`NoOpComposerStateSaver`** — fallback for the legacy `create(modelClass)` factory path (no persistence)
- **`ParcelableAttachment`** — `@Parcelize` wrapper containing only fields populated at compose time (`upload`, `type`, `name`, `fileSize`, `mimeType`, `title`, `extraData`)

Key design decisions:
- **Defensive parcel safety**: Before saving, all `extraData` values are validated via `isParcelSafe()`. If any value is non-parcelable, saving is skipped entirely (falls back to current behavior — no crash, no silent data loss)
- **File existence check on restore**: Attachments whose `upload` file no longer exists on disk are silently dropped
- **Attachments only**: Text input is not persisted (handled by existing draft message mechanism). `alsoSendToChannel` is not persisted (only relevant in thread mode which isn't persisted)

### UI Changes

No UI changes.

### Testing

1. Enable "Don't keep activities" in developer options
2. Open a channel, select image/file attachments
3. Open another activity on top (e.g., channel info)
4. Return — verify attachments are still in the composer
5. Send the message — verify it sends successfully with attachments
6. Unit tests: `ParcelableAttachmentTest`, `SavedStateComposerStateSaverTest`, `MessageComposerControllerTests`

## Summary by CodeRabbit

* **New Features**
  * Message composer attachments are now persisted and automatically restored if the app is unexpectedly terminated while composing a message.


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/internal/ParcelableAttachment.kt
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Recursively checks whether a value can be safely written via [android.os.Parcel.writeValue].
+ */
+private fun Any.isParcelSafe(): Boolean = when (this) {
+    is String, is Int, is Long, is Float, is Double, is Boolean, is Byte, is Short -> true
+    is BooleanArray, is ByteArray, is FloatArray, is IntArray, is LongArray, is DoubleArray -> true
+    is List<*> -> all { it == null || it.isParcelSafe() }
+    is Map<*, *> -> all { (k, v) -> k is String && (v == null || v.isParcelSafe()) }
+    else -> false
```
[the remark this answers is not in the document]
VelikovPetar: The whitelist is intentionally conservative. I went with this approach to make sure that the built-in Stream attachments are covered, while trying to prevent any unexpected behaviour for custom attachments. This fix is anyway covering an edge case, and I would much rather have the minimal coverage needed for the supported attachments, rather than breaking existing integrations with custom attachments.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/internal/SavedStateComposerStateSaver.kt
```diff
@@ -0,0 +1,61 @@
...
+            logger.w {
+                "[saveAttachments] Skipping attachment save: extraData contains non-parcelable values. " +
+                    "Attachments will not survive process death."
+            }
+            savedStateHandle.remove<ArrayList<ParcelableAttachment>>(KEY_ATTACHMENTS)
+            return
+        }
+        savedStateHandle[KEY_ATTACHMENTS] = ArrayList(attachments.map { it.toParcelable() })
+    }
+
```
aleksandar-apostolov, line 48: Question: the PR description mentions "File existence check on restore: Attachments whose `upload` file no longer exists on disk are silently dropped" — but `restoreAttachments()` maps straight through without filtering. Is this handled downstream (e.g., the upload flow gracefully handles missing files), or should there be a `.filter { it.upload == null || it.upload.exists() }` here?

aleksandar-apostolov (approved):
LGTM, posted 1 question.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/internal/SavedStateComposerStateSaver.kt
VelikovPetar: This was initially a wrong description on the PR, but when I think about it, perhaps adding this change makes sense. While `upload == null` can never be true, the `upload.exists()` can be `false` (cache dir cleared after the process death, and before the process restoration), so the upload would fail anyway. 
I will introduce this change.

VelikovPetar: Added this filter in the `MessageComposerController`, where the restored attachments are consumed. https://github.com/GetStream/stream-chat-android/pull/6377/commits/15a99b2aa24326b99f4292901b66457293da2701

PR #6383, opened 2026-04-24
Harden E2E UiAutomator text-wait helpers against recomposition

andremion:
### Goal

Compose-driven nodes get recycled mid-poll, causing `waitForText` to hit `StaleObjectException` or NPE and fail tests deterministically on slow emulators. Tighten the helpers so assertions survive recomposition and surface timeouts with a clear error.

### Implementation

**`stream-chat-android-e2e-test / Wait.kt`**
- `waitForText` moved from `UiObject2` receiver to `BySelector`. Re-resolves the node on every iteration via `device.findObject(this)`, swallows `StaleObjectException` as "no match yet", sleeps 50 ms per poll, and returns the matched `String` (or last observed text on timeout).
- `waitToAppear` (both overloads) now throws a descriptive `IllegalStateException` on timeout instead of leaking a null NPE or `ArrayIndexOutOfBoundsException`.
- KDoc added describing throw/stale semantics.

**`stream-chat-android-compose-sample / robots`**
- Call sites in `UserRobotChannelListAsserts` and `UserRobotMessageListAsserts` migrated to the new `BySelector.waitForText(…)` signature, dropping the prior `waitToAppear().text` pattern that was susceptible to the same recomposition races.

**API** — `waitForText` receiver and return type changed (`UiObject2 → BySelector`, `UiObject2 → String`). Internal to the test-infra module; all in-repo consumers migrated.

### Testing

E2E run required (device/emulator). On `stream-chat-android-compose-sample` with the mock server:

1. **Recomposition resilience** — run the message-list asserts (`assertEditedMessage`, `assertQuotedMessage`, `assertMentionWasApplied`, `assertThreadReplyLabel`) and `ChannelListTests.test_channelPreviewShowsPreviousMessage_whenLastMessageIsDeleted`. Expect: no intermittent `StaleObjectException` in logcat; previously flaky suites pass consistently.
2. **Timeout surfaces cleanly** — temporarily break a selector (e.g. wrong resource id) at any `waitToAppear` call site and re-run. Expect: `IllegalStateException: waitToAppear timed out after 10000ms; no object matched selector: …` instead of an NPE.

### Follow-up tasks (out of scope here)

Separate PRs planned under `refactor/e2e-uiautomator-helpers` and `perf/e2e-test-runtime`:

- `BySelector.wait()` discards its boolean — either return it or throw on timeout; migrate call sites that currently do `.wait().isDisplayed()`.
- Apply the same stale-guard + poll sleep to sister helpers in `Wait.kt` (`waitForCount`, `waitForTextToChange`) and to `UiObject2.isDisplayed()` in `Element.kt`.
- `assertSystemMessage` currently discards `isDisplayed()` — add the missing `assertTrue`/`assertFalse`.
- `openChannel` hits `ArrayIndexOutOfBoundsException` when the channel list hasn't loaded; switch to `waitToAppear(withIndex = …)`.
- Scroll helpers (`scrollUp/DownUntilDisplayed`) NPE on miss — replace with descriptive error.
- Replace fixed `sleep(500/1000/5000)` with predicate waits where possible; reduce `defaultTimeout` from 10 s and add an explicit long timeout for known-slow ops.
- Route typing through `UiObject2.typeText` (direct `.text =`) instead of `adb shell input text` where the field is addressable — the shell path types character-by-character and dominates runtime for long strings.
- `RetryRule` attachment recording fires per failed attempt — record only the final attempt.

No UI changes.

## Summary by CodeRabbit

* **Tests**
  * Improved assertion reliability for channel message previews with enhanced text matching logic.
  * Added test coverage for deleted message placeholders displayed in channel preview lists.
  * Refined UI automation wait mechanisms with more accurate text polling and explicit timeout error messages.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotChannelListAsserts.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotChannelListAsserts.kt
@@ -37,7 +37,10 @@ fun UserRobot.assertMessageInChannelPreview(text: String, fromCurrentUser: Boole
         false -> "${ParticipantRobot.name}: $text"
         null -> text
     }
-    assertEquals(expectedPreview, Channel.messagePreview.waitToAppear().waitForText(expectedPreview).text.trimEnd())
+    assertEquals(
+        expectedPreview,
+        Channel.messagePreview.waitForText(expectedPreview, mustBeEqual = false).trimEnd(),
+    )
     return this
 }
 
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
@@ -119,10 +119,8 @@ fun UserRobot.assertMessageFailedIcon(isDisplayed: Boolean): UserRobot {
 
 fun UserRobot.assertEditedMessage(text: String): UserRobot {
     assertMessage(text)
-    assertEquals(
-        appContext.getString(R.string.stream_compose_message_list_footnote_edited),
-        Message.editedLabel.waitToAppear().text,
-    )
+    val expectedLabel = appContext.getString(R.string.stream_compose_message_list_footnote_edited)
+    assertEquals(expectedLabel, Message.editedLabel.waitForText(expectedLabel))
     return this
 }
 
@@ -237,8 +235,7 @@ fun UserRobot.assertMentionWasApplied(): UserRobot {
 }
 
 fun UserRobot.assertComposerText(expectedText: String): UserRobot {
-    val actualText = Composer.inputField.waitToAppear().text
-    assertEquals(expectedText, actualText)
+    assertEquals(expectedText, Composer.inputField.waitForText(expectedText))
     return this
 }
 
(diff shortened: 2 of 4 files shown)


andremion:
@coderabbit review

VelikovPetar (commented):
I believe that we should be able to uncomment the following test after merging this: `test_quotedReplyInThreadAndAlsoInChannel`

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotChannelListAsserts.kt
```diff
@@ -37,7 +37,10 @@ fun UserRobot.assertMessageInChannelPreview(text: String, fromCurrentUser: Boole
         false -> "${ParticipantRobot.name}: $text"
         null -> text
     }
-    assertEquals(expectedPreview, Channel.messagePreview.waitToAppear().waitForText(expectedPreview).text.trimEnd())
+    assertEquals(
+        expectedPreview,
+        Channel.messagePreview.waitForText(expectedPreview, mustBeEqual = false).trimEnd(),
```
VelikovPetar, line 42: Just curious: Why did `mustBeEqual` value changed in this test?

andremion: The preview text under load can render with trailing whitespace ("You: test "). Strict equality inside `waitForText` would never match "You: test" and would burn the full 10-second timeout before returning the last observed string for trimEnd() to fix. Switching to `mustBeEqual = false` lets the helper return the moment the expected substring is present, then trimEnd() + outer assertEquals still validates exact equality. Same correctness, faster on the whitespace-truncation case.

andremion:
> I believe that we should be able to uncomment the following test after merging this: `test_quotedReplyInThreadAndAlsoInChannel`

Let's see: a9d51bd63f77354f807067c0722a9f43bf57e93d

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotChannelListAsserts.kt
VelikovPetar: I must be missing something - why would the text render with a trailing space? (sorry if I am missing something obvious, but I cannot figure this out at the moment)

andremion: The text-only preview path in `MessagePreviewFormatter.appendMessageText` does `append("$messageText ")`. A literal trailing space (so attachment text can be appended after without a separator). That carries through to the `Stream_MessagePreview` row's fullPreview and into the semantics text UiAutomator reads, so a plain text preview shows up as "You: foo " rather than "You: foo".

With `mustBeEqual = true`, the inner poll never strictly equals the expected value, so the helper burns the full 10s timeout before returning the last observed string for trimEnd() to fix up. Every successful preview assertion paid that tax. mustBeEqual = false lets it return as soon as the expected text is contained in the observed text, and the outer trimEnd() + assertEquals still validates exact equality after stripping the trailing space.

VelikovPetar: Ok got it now, thanks!

VelikovPetar (approved):

PR #6391, opened 2026-04-28
Add dependabot configuration to bump Stream conventions

gpunto:
### Goal

Add dependabot configuration so that PRs bumping Stream conventions will be created automatically by dependabot.

### Implementation

Add `.github/dependabot.yml` with configuration to only update workflows/actions & stream plugins

### Testing

We'll enable dependabot in the repo settings & check that PRs are opened properly

## Summary by CodeRabbit

* **Chores**
  * Configured automated dependency updates with daily checks for GitHub Actions and project dependencies.


.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
...
+      interval: daily
+    labels:
+      - "pr:ci"
+
+  - package-ecosystem: gradle
+    directory: /
+    schedule:
+      interval: daily
+    labels:
+      - "pr:ci"
```
andremion, line 15: Shouldn't labels be "pr:dependencies"?

.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
...
+    allow:
+      - dependency-name: "io.getstream.project"
+      - dependency-name: "io.getstream.android.library"
+      - dependency-name: "io.getstream.android.application"
+      - dependency-name: "io.getstream.android.test"
+      - dependency-name: "io.getstream.java.library"
+      - dependency-name: "io.getstream.java.platform"
+      - dependency-name: "io.getstream.publish"
```
andremion, line 23: "io.getstream.android.test", "io.getstream.java.platform", and "io.getstream.publish" are not in libs.versions.toml file.
Which libraries are these?

andremion: Given that the intention is to update only GitHub actions and Stream Conventions, should all of those be listed there?

gpunto: They are all the stream conventions plugins. I've decided to list also the ones we don't use in chat yet so that if we start using them we don't forget to update the dependabot config

.github/dependabot.yml
gpunto: In my mind that's more for product dependencies rather than buildscript/CI dependencies

andremion: Make sense. In both cases, they should not be listed in the release notes, correct?
Maybe using "pr:ignore-for-release", too?

gpunto: Other CI work is listed under the "internal" section in release notes. Do we want to exclude these? Let me raise a question in Slack

andremion (approved):
Approving to unblock, as it can be refined later if needed

PR #6398, opened 2026-04-29
Set required permissions for the SDK size and DB entities checks

gpunto:
### Goal

SDK size & DB entities checks need write permissions for writing comments, so we're adding the relevant permission block. Up until now, this didn't cause problems because the token we're using has those permissions, but that doesn't work for Dependabot PRs. [They receive a read-only token](https://docs.github.com/en/code-security/reference/supply-chain-security/troubleshoot-dependabot/troubleshooting-dependabot-on-github-actions#troubleshooting-failures-when-dependabot-triggers-existing-workflows) unless we declare permissions, which is why the corresponding failures in https://github.com/GetStream/stream-chat-android/pull/6394.

### Implementation

Add the permissions block

### Testing

After merging this PR, we should update dependabot's PR and verify that the SDK size check succeeds

## Summary by CodeRabbit

* **Chores**
  * Updated GitHub Actions workflow permissions to explicitly define authorization scopes for continuous integration checks and SDK size verification tasks.

### DIFF
--- a/.github/workflows/check-entities.yml
+++ b/.github/workflows/check-entities.yml
@@ -2,6 +2,11 @@ name: Check DB Entities
 
 on: [pull_request]
 
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/check-entities.yml
+++ b/.github/workflows/check-entities.yml
@@ -5,7 +5,6 @@ on: [pull_request]
 permissions:
   contents: read
   pull-requests: write
-  issues: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/check-entities.yml
```diff
@@ -2,6 +2,11 @@ name: Check DB Entities
 
 on: [pull_request]
 
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
```
andremion: just to double-check: do we need `issues: write`?

.github/workflows/sdk-size-checks.yml
```diff
@@ -3,6 +3,11 @@ name: SDK size checks
 on:
   pull_request:
 
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
```
andremion: same here

andremion (approved):

.github/workflows/check-entities.yml
gpunto: Good catch, it's not needed. Dropped!

PR #6405, opened 2026-04-30
Expand message long-press hit area to entire cell

andremion:
### Goal

Long-press to open the message actions menu now targets the entire message cell, not the bubble alone. Aligns the long-press surface with the existing swipe-to-reply surface (already cell-wide), gives short one or two character messages an adequate gesture target, and matches the dominant convention on WhatsApp / iMessage.

### Implementation

Three commits, intentionally separate so the structural change is reviewable in isolation:

1. **`Expand message long-press hit area to entire cell`** — moves the `Modifier.combinedClickable` from the inner content row (`wrapContentWidth`) to the outer container (`fillMaxWidth`), so the click and long-click handlers fire anywhere across the row. Switches `indication = null` to `indication = ripple()` so the gesture target is visible to the user. Reaction bar and actions menu remain anchored to the bubble; only the gesture surface widens.
2. **`Extract message click predicates into named locals`** — replaces the inline boolean expressions inside the click and long-click lambdas with `canOpenThread` and `canOpenActions`, so the gesture handlers read as intent.
3. **`Collapse message item Box+Row wrapper into a single Row`** — the outer `Box(fillMaxWidth, contentAlignment)` was only providing edge alignment, which a Row achieves directly via `horizontalArrangement`. Drops the redundant `Stream_MessageItem` test tag (zero references outside this file) and keeps `Stream_MessageCell` on the merged Row, where it accurately labels the cell that the gesture surface lives on.

#### API surface

No public API changes. `apiDump` produces no diff.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/9834757a-2f37-4648-a62a-5f6c8ecbabcd" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/9d139527-92dd-4667-8bae-be194940a94c" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### Testing

Run the Compose sample app on a device and open any channel.

1. **Hit-area expansion.** Long-press in the empty space to the side of a short message bubble (e.g. a one-character "k"). Expected: the message actions menu opens — previously only long-pressing on the bubble itself worked.
2. **Ripple feedback.** Tap any message anywhere on the row. Expected: a ripple emanates from the touch point across the full row width.
3. **Bubble-anchored UI unchanged.** Open the message actions menu and the reactions picker. Expected: both remain visually anchored to the bubble, not the cell — the gesture target widened, not the visual anchor.
4. **Swipe-to-reply.** Horizontal swipe on the row. Expected: works as before — `SwipeToReply` is still the parent of the merged Row.
5. **Thread-start tap.** Tap on a message that started a thread, from outside that thread. Expected: opens the thread, exactly as before.
6. **Deleted / uploading messages.** Long-press a deleted message or one that is mid-upload. Expected: no actions menu appears (gated by `canOpenActions`).
7. **Outgoing vs incoming alignment.** Scroll through a mix of own and other-user messages. Expected: outgoing messages remain right-aligned, incoming left-aligned, with the avatar on the correct side.

## Summary by CodeRabbit

* **Style**
  * Enhanced message interactions with ripple effect feedback, improving visual response when tapping messages.
  * Improved message layout structure for better rendering consistency.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
@@ -175,12 +175,15 @@ public fun MessageContainer(
     val focusState = messageItem.focusState
     val haptic = LocalHapticFeedback.current
 
+    val canOpenThread = message.isThreadStart() && !messageItem.isInThread
+    val canOpenActions = !message.isDeleted() && !message.isUploading()
+
     val clickModifier = Modifier.combinedClickable(
         interactionSource = remember { MutableInteractionSource() },
         indication = ripple(),
-        onClick = { if (message.isThreadStart() && !messageItem.isInThread) onThreadClick(message) },
+        onClick = { if (canOpenThread) onThreadClick(message) },
         onLongClick = {
-            if (!message.isDeleted() && !message.isUploading()) {
+            if (canOpenActions) {
                 haptic.performHapticFeedback(HapticFeedbackType.LongPress)
                 onLongItemClick(message)
             }
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
@@ -181,6 +181,7 @@ public fun MessageContainer(
     val clickModifier = Modifier.combinedClickable(
         interactionSource = remember { MutableInteractionSource() },
         indication = ripple(),
+        enabled = canOpenThread || canOpenActions,
         onClick = { if (canOpenThread) onThreadClick(message) },
         onLongClick = {
             if (canOpenActions) {
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
@@ -24,6 +24,7 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.combinedClickable
 import androidx.compose.foundation.gestures.detectHorizontalDragGestures
 import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ColumnScope
@@ -38,7 +39,6 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.layout.wrapContentHeight
-import androidx.compose.foundation.layout.wrapContentWidth
 import androidx.compose.material3.ripple
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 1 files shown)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
```diff
@@ -174,12 +175,15 @@ public fun MessageContainer(
...
     val haptic = LocalHapticFeedback.current
 
+    val canOpenThread = message.isThreadStart() && !messageItem.isInThread
+    val canOpenActions = !message.isDeleted() && !message.isUploading()
+
     val clickModifier = Modifier.combinedClickable(
         interactionSource = remember { MutableInteractionSource() },
-        indication = null,
-        onClick = { if (message.isThreadStart() && !messageItem.isInThread) onThreadClick(message) },
+        indication = ripple(),
```
aleksandar-apostolov, line 183: Question on the ripple behavior: with the click surface now spanning the full row width, a short tap on the empty side of a message produces a ripple even when `canOpenThread` is false (e.g. tapping next to an own message that isn't a thread start) — `onClick` is a no-op there, but the ripple still fires. Is that intentional? If not, gating the ripple via `enabled = canOpenThread || canOpenActions` (or splitting indication per-gesture) would keep the visual feedback honest about what's actionable.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
```diff
@@ -218,89 +222,87 @@ public fun MessageContainer(
...
                 .wrapContentHeight()
                 .background(color = color)
+                .then(clickModifier)
                 .semantics { contentDescription = description },
-            contentAlignment = messageAlignment.itemAlignment,
+            horizontalArrangement = if (messageAlignment == MessageAlignment.Start) {
+                Arrangement.Start
+            } else {
+                Arrangement.End
+            },
```
aleksandar-apostolov, line 238: Nit / suggestion: `MessageAlignment` already exposes `itemAlignment` and `contentAlignment` so callers don't branch on the value. Adding a third property keeps the abstraction symmetric:

```kotlin
public enum class MessageAlignment(
    public val itemAlignment: Alignment,
    public val contentAlignment: Alignment.Horizontal,
    public val horizontalArrangement: Arrangement.Horizontal,
) {
    Start(Alignment.CenterStart, Alignment.Start, Arrangement.Start),
    End(Alignment.CenterEnd, Alignment.End, Arrangement.End),
}
```

Call site collapses to `horizontalArrangement = messageAlignment.horizontalArrangement`. Single call site today so the win is small — take it or leave it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
andremion: That makes sense. I will take a look

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
andremion: Tried that first and reverted. With the Box->Row change, `itemAlignment` is already unused after this PR, so adding `horizontalArrangement` would only add a third property to an enum that already has an unused one.
These should have been internal from the start. I'd prefer to do the cleanup in v8: make the properties internal and keep the enum public. Then we can remove `itemAlignment`, add `horizontalArrangement`, or change the enum however we want without breaking the public API.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
andremion: f645c2bee38e41c26fb2c5bfd9b20f49209bb34e

Quick note: this gate does not change the scenario you described, since `canOpenActions` is still true for a regular non-thread-start message, so enabled stays true and the ripple still fires on tap. The gate helps when both flags are false, for example, a deleted message that is not a thread start.

aleksandar-apostolov (approved):

PR #6406, opened 2026-05-01
Add `MessageBufferingConfig` to allow custom back-pressure config for `message.new` events

VelikovPetar:
### Goal

High-traffic channel types (e.g. livestreams) can produce a flood of `message.new` events that arrive faster than the sequential event-handling pipeline can process them. The current implementation funnels every socket event through a single `MutableSharedFlow` with `extraBufferCapacity = Int.MAX_VALUE`, which means there is no back-pressure: a burst of new-message events queues up unbounded memory and starves more important signals (reads, bans, member updates) of timely processing.

This PR introduces a `MessageBufferConfig` that lets integrators opt specific channel types into a bounded buffer for `NewMessageEvent`s, with a configurable overflow strategy (`SUSPEND` / `DROP_OLDEST` / `DROP_LATEST`). Signal-critical events and events for non-opted-in channel types are unaffected.

### Implementation

- Added `MessageBufferConfig` (under `MessageLimitConfig.messageBufferConfig`) exposing:
  - `channelTypes: Set<String>` — channel types whose `NewMessageEvent`s go through the bounded buffer (empty by default → feature is a no-op).
  - `capacity: Int` — buffer capacity (defaults to `Int.MAX_VALUE`).
  - `overflow: BufferOverflow` — overflow strategy (defaults to `SUSPEND`).
- `EventHandlerSequential` now allocates a secondary `MutableSharedFlow` (`bufferedNewMessageEvents`) lazily, only when buffering is enabled, so the default configuration pays no cost for it.
- Two listener variants:
  - `defaultSocketEventListener` — the existing unbuffered path; used when no channel types are opted in.
  - `bufferedSocketEventListener` — routes `NewMessageEvent`s for opted-in channel types to the bounded flow, and everything else (including non-opted-in `NewMessageEvent`s and all other event types) to the unbuffered flow.
- `startListening()` picks the listener based on `bufferConfig.channelTypes.isNotEmpty()` and only collects from `bufferedNewMessageEvents` when buffering is enabled.
- `StreamStatePluginFactory` wires the config from `StatePluginConfig.messageLimitConfig.messageBufferConfig` into `EventHandlerSequential`.

The bounded flow shares the same downstream pipeline (`socketEventCollector` → `handleBatchEvent`) as the unbuffered flow, so ordering inside each flow is preserved and back-pressure is applied independently per flow.

### Testing

- Added 200+ lines of unit tests in `EventHandlerSequentialTest` covering:
  - default behaviour (no buffering) is unchanged when `channelTypes` is empty.
  - `NewMessageEvent`s for opted-in channel types are routed through the bounded buffer.
  - `NewMessageEvent`s for non-opted-in channel types and all non-`NewMessageEvent` events keep using the unbuffered path.
  - `DROP_OLDEST` / `DROP_LATEST` / `SUSPEND` overflow strategies behave as expected when the buffer is full.
- Existing tests (`TotalUnreadCountTest`, `EventHandlerSequentialUserMessagesDeletedTest`) updated to pass the new `bufferConfig` argument.

## Release Notes

* **New Features**
  * Introduced customizable message buffering for improved event stream management. Configure buffer settings per channel type with adjustable capacity and overflow handling strategies, enabling optimized message delivery performance and efficient resource management based on specific channel requirements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6406?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/config/StatePluginConfig.kt
```diff
@@ -161,3 +167,54 @@ public data class ChannelMessageLimit(
...
+ *   that the underlying socket listener uses non-suspending `tryEmit`, so with `SUSPEND` an
+ *   overflowing emission is simply reported as a failed emit rather than blocking the socket.
+ * - [BufferOverflow.DROP_OLDEST]: the oldest queued event is evicted to make room for the new
+ *   one. Useful for live channels where freshness matters more than completeness.
+ * - [BufferOverflow.DROP_LATEST]: the newest event is discarded and the queued events are kept.
+ */
+public data class MessageBufferConfig(
+    public val channelTypes: Set<String> = emptySet(),
+    public val capacity: Int = Int.MAX_VALUE,
+    public val overflow: BufferOverflow = BufferOverflow.SUSPEND,
```
gpunto: Given that we don't honor the `SUSPEND` strategy (it basically is equivalent to `DROP_LATEST` in the current implementation), should we change the default? Or maybe even restrict what can be passed here with a different enum instead of reusing kotlin's `BufferOverflow`.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -167,12 +169,58 @@ internal class EventHandlerSequential(
...
+     * [socketEvents].
+     */
+    private val bufferedSocketEventListener: ChatEventListener<ChatEvent> = ChatEventListener { event ->
+        val target = if (event is NewMessageEvent && event.channelType in bufferConfig.channelTypes) {
+            bufferedNewMessageEvents
+        } else {
+            socketEvents
+        }
+        logEmitOutcome(event, target.tryEmit(event))
+    }
```
gpunto, line 214: In here I think we're introducing an edge case when buffering is enabled. Given that now events can go to different flows, we lose ordering of new message events vs. all other events. For example if we get `new message X` and then `reaction added to X`, we could now process them in the opposite order, which would leave us in the wrong state (the reaction wouldn't be applied).

However, maybe this is acceptable, given that the point of the PR is to allow dropping events so the integrator willingly accepts a consistency loss already. Is that the case? If we are fine with it, maybe the ordering issue is worth an explicit mention in the public KDoc.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/config/StatePluginConfig.kt
VelikovPetar: Nice catch! What do you think of defining our own enum, but allow only `DROP_LATEST` / `DROP_OLDEST` as options? (since `SUSPEND` is not really applicable)

gpunto: Sounds good to me!

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/event/handler/internal/EventHandlerSequential.kt
VelikovPetar: In my opinion this is fine - I would say that when a customer wants this behaviour, they are already aware that some messages will be missing from the channel. I would go for documenting this properly. 

(also something interesting: On V6, for `reaction.new` we actually perform an `upsert` operation with the `message`, so the message should still be inserted in the state. This is changed on V7, so the reaction will be dropped)

gpunto: Great, thanks!

gpunto (approved):

PR #6408, opened 2026-05-04
Use right chevron icon from design system

gpunto:
### Goal

`ChannelInfoOption` used `Icons.AutoMirrored.Filled.KeyboardArrowRight` from `androidx.compose.material:material-icons-core`. That artifact wasn't declared as a dependency of `stream-chat-android-compose`. It was pulled in transitively via `material3`, which used to expose it as an `api` dependency.

In recent Compose versions, `material3` no longer re-exports `material-icons-core`. Consumer apps using a newer Compose BOM than what the SDK was built against end up without the icon class on their runtime classpath. Their build still succeeds (because their own code doesn't import `Icons`), but the moment `GroupChannelInfoScreen` / `DirectChannelInfoScreen` renders the option row, the app crashes with `NoClassDefFoundError`.

Also, we should use design system icons now that we have them.

### Implementation

Replaced the Material icon with our existing design-system drawable `R.drawable.stream_design_ic_chevron_right`.

### 🎨 UI Changes

Please check the updated snapshots

### Testing

Repro steps:
1. Create a fresh Android Compose app, depend on `io.getstream:stream-chat-android-compose:7.0.1` and `androidx.compose:compose-bom:2026.02.01`.
2. Open `GroupChannelInfoScreen` or `DirectChannelInfoScreen`.
3. App crashes with `NoClassDefFoundError: androidx.compose.material.icons.automirrored.filled.KeyboardArrowRight` as soon as the option rows render.

To verify the fix you'd have to publish a local version of Stream Chat, use it in step 1 above, and verify it doesn't crash anymore

## Summary by CodeRabbit

* **Style**
  * Updated the chevron icon appearance in channel info options.


gpunto:
@coderabbitai review

VelikovPetar (approved):

PR #6412, opened 2026-05-05
Align shared header divider and height across screens

andremion:
### Goal

Three design-system inconsistencies between the Channel List, Thread List, and Message / Thread View headers — two flagged on the SDK Testing Feedback Notion board, plus one consistency lock-down picked up in code review:

- [Issue #3 — Header and composer use a dropshadow instead of `border/core/subtle`](https://www.notion.so/3516a5d7f9f680cb87c0ef0c13dec037)
- [Issue #4 — Header height in the message view is shorter than in the Channel List and Thread List headers](https://www.notion.so/3516a5d7f9f68014939cc0e5045e4f4a)
- Reviewer suggestion: extract a shared internal building block so the four screen-level headers can no longer drift apart.

### Implementation

- **Borders.** `ChannelHeader` (used by Message View and Thread View) and `MessageComposer` (non-floating mode) draw a 1.dp `borderCoreSubtle` divider instead of an elevation / dropshadow. `ListHeader` was using `borderCoreDefault` for its existing divider; updating it to `borderCoreSubtle` aligns all four screens on the same token.
- **Header height.** `ChannelScreen.DefaultTopBarContent` constrained `ChannelHeader` to a hardcoded `Modifier.height(56.dp)`. `ListHeader` renders intrinsically at 72.dp (avatar 48.dp + 2× 12.dp padding). Replaced the rigid `.height` with `.defaultMinSize(minHeight = 72.dp)` so the message-view header matches the list screens.
- **HeaderScaffold extraction.** Renamed the internal `ListHeader` to `HeaderScaffold` and lifted the shared chrome (`Surface` + inset row + bottom divider) into a single building block. `ChannelHeader`, `ChannelListHeader`, and `ThreadListHeader` all delegate the chrome to `HeaderScaffold` and only own their domain slots. Dropped the unused `color` / `shape` / `elevation` parameters from the previous signature and baked the design-system padding (`spacingSm`) and inter-slot gap (`spacingXs`) in. `ChannelHeader`'s Row gains 4 dp horizontal padding and an 8 dp inter-slot gap, matching what the list headers already used.

Snapshot baselines were re-recorded for `ChannelListHeaderTest`, `ThreadListHeaderTest`, `ChannelHeaderTest`, and `MessageComposerTest`.

#### Channel List

| Before | After |
| --- | --- |
| channels before | channel list after |

#### Channel

| Before | After |
| --- | --- |
| channel before | channel after |

#### Replies

| Before | After |
| --- | --- |
| reply before | reply after |

#### Thread List

| Before | After |
| --- | --- |
| thread before | threads after |

### Testing

1. Build and run the Compose sample app on this branch.
2. Open the Channel List.
   - Expected: a thin 1.dp `borderCoreSubtle` line under the header.
3. Open any channel.
   - Expected: the message-view header sits at the same vertical position as the Channel List header from step 2 (was ~16 dp shorter). The header ends with the same 1 dp subtle divider, not a dropshadow. Slot positions inside the header shift slightly (4 dp horizontal padding + 8 dp inter-slot gap) — this is the design-system alignment.
4. Scroll the message list and look at the composer (non-floating style).
   - Expected: a 1 dp subtle divider sits at the top of the composer instead of a dropshadow.
5. Toggle the composer to floating style (if exposed in the demo).
   - Expected: no divider or shadow on the floating composer (unchanged).
6. Open a thread.
   - Expected: the header treatment matches step 3.
7. Switch to the Thread List tab.
   - Expected: header divider matches the Channel List from step 2.

## Summary by CodeRabbit

* **Style**
  * Adjusted channel header sizing for improved layout proportions.
  * Refined divider styling and colors across chat interface components.
  * Removed shadow effects in favor of refined visual separators.


andremion:
@coderabbit review

gpunto (approved):
Looks good!

I wonder if we should introduce an internal shared header component at some point, to make it harder to end up in an inconsistent state.

andremion:
> I wonder if we should introduce an internal shared header component at some point, to make it harder to end up in an inconsistent state.

@gpunto , 616a17b

PR #6415, opened 2026-05-06
Add support for `PredefinedFilters` for `QueryChannels`

VelikovPetar:
### Goal

Add support for server-side `PredefinedFilters` to `QueryChannels`. Apps can now run a channel query by referencing a filter/sort registered on the backend (by name, plus interpolation values) instead of building the `FilterObject`/`QuerySorter` on the client. This lets product/backend tweak filtering rules without an app release.

Resolves: AND-1162

_**You can check / update / create the predefined filters for the app "Stream Mobile SDK / Stream SDK - Android" - in the "Predefined Filters" section in the Beta Dashboard**_

### Implementation

**Public API**
- `QueryChannelsRequest` gains `predefinedFilter: String?`, `filterValues: Map<String, Any>?`, `sortValues: Map<String, Any>?`. When `predefinedFilter` is set, the client-side `filter`/`querySort` are ignored by the backend.
- `QueryChannelsResult` is a new wrapper around `ChatApi.queryChannels` that exposes both `channels` and the backend-resolved `PredefinedFilter` (parsed `name` + `FilterObject` + `QuerySorter<Channel>`).
- `ChannelListViewModel` (Compose) and `ChannelListViewModel` (UI Components) get a new constructor / `ChannelListViewModelFactory` overload taking `predefinedFilterName`, `filterValues`, `sortValues`. In predefined mode, `setFilters`/`setQuerySort` are no-ops; channel search still narrows the displayed list.

**State / offline architecture**
- New `QueryChannelsIdentifier` sealed type (`Standard` / `Predefined`) is the canonical key used by `LogicRegistry`, `StateRegistry`, the listener, and the offline repository. A `QueryChannelsRequest`/`QueryChannelsSpec` resolves to one or the other depending on whether `predefinedFilter` is present.
- `QueryChannelsSpec` keeps its existing `(filter, querySort)` constructor (so JVM signatures stay stable) and adds `predefinedFilterName`/`predefinedFilterValues`/`predefinedSortValues` as `var`s. For predefined queries, those three form the row's stable identity; `filter`/`querySort` carry the *currently resolved* values.
- `QueryChannelsMutableState` starts predefined queries with placeholder filter/sort and applies the resolved spec via `applyResolvedSpec(...)` once the backend response arrives (or the row is rehydrated from disk).
- Internal `ChannelListViewModel.QueryMode` (`Standard` vs `Predefined`) drives the two flows — pagination, refresh, search, and filter-flow wiring branch on it.

**Persistence**
- `QueryChannelsEntity` gains `predefinedFilterName`/`predefinedFilterValues`/`predefinedSortValues`. `QueryChannelsRepository.selectBy(identifier)` replaces the previous filter-based lookup.
- Room DB version bumped to `104` (destructive migration, in line with the rest of the SDK).

**Wire format**
- `FilterDomainMapping` now supports a full round-trip (DTO ↔ domain) so the resolved predefined filter coming back from the server can be parsed into a `FilterObject` and stored.

### Testing

- Unit tests added/updated across the new boundaries: `QueryChannelsIdentifier` derivation, `MoshiChatApi`, `FilterDomainMapping` (round trip), `QuerySortByField` round trip, `DatabaseQueryChannelsRepository`, `QueryChannelsLogic` / `QueryChannelsDatabaseLogic` / `QueryChannelsListenerState`, `QueryChannelsMutableState`, `LogicRegistry`, and both `ChannelListViewModel`s.
- Both sample apps (Compose and UI Components) are wired to a backend-registered `android_sample_filter` so the feature can be exercised end-to-end on device.

## Summary by CodeRabbit

* **New Features**
  * Added support for server-defined predefined channel filters with template-based parameter interpolation instead of client-side filter construction.
  * Channel list UI now supports querying via predefined filter names with optional filter/sort value maps.

* **Bug Fixes & Improvements**
  * Enhanced channel query response to include resolved predefined filter metadata from the server.
  * Updated database schema (v103→v104) to persist predefined filter configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6415)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/FilterDomainMapping.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    // Multi-key map: implicit AND
+    val filters = map.entries.mapNotNull { (key, value) -> parseSingleEntry(key, value) }
+    if (filters.isEmpty()) return null
+    if (filters.size == 1) return filters.first()
+    return Filters.and(*filters.toTypedArray())
+}
+
+@Suppress("SpreadOperator")
```
gpunto, line 64: Maybe at some point we should consider introducing `Filters` overloads that take a collection, so we avoid these suppression and the intermediate array (since now we do `collection -> array -> set`). Not blocking at all for this PR.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
```diff
@@ -172,10 +209,40 @@ internal class QueryChannelsMutableState(
...
+     *  - The server response arrives carrying `QueryChannelsResult.predefinedFilter`, or
+     *  - The offline DB rehydrates a previously persisted resolved spec for the same identifier.
+     *
+     * Not relevant for standard (non-predefined) queries.
+     *
+     * Because [QueryChannelsSpec] keeps `filter` and `querySort` as `val` for binary
+     * compatibility, we replace the held [_querySpec] instance instead of mutating it in place.
+     * `cids` and the predefined identity fields are carried over from the previous instance.
+     */
+    fun applyResolvedSpec(filter: FilterObject, sort: QuerySorter<Channel>) {
```
gpunto, line 234: In the docs we're saying that this is no op in the standard case, but that responsibility is currently on the callers. I'm a bit concerned that we might be accidentally call this for standard filters in the future.

Some ideas:
- Maybe we could we use a type specific to predefined filters here to make this un-callable for standard filters? -> although we don't yet have that type as far as I can see
- We could change the function name and update the documentation to say that this should only be called for predefined filters rather than implying that the function itself is no op
- We cold make it no op by adding an early return on `identifier !is QueryChannelsIdentifier.Predefined`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/queryChannels/internal/QueryChannelsEntity.kt
```diff
@@ -26,9 +26,19 @@ import io.getstream.chat.android.models.querysort.QuerySorter
...
     val querySort: QuerySorter<Channel>,
     val cids: List<String>,
+    /**
+     * Set only for predefined-filter queries; null for standard ones. Together with the value maps
+     * below, the predefined name forms the row's stable identity (the resolved filter/sort can
+     * change between runs if the server-side template changes).
+     */
+    val predefinedFilterName: String? = null,
+    val predefinedFilterValues: Map<String, Any>? = null,
+    val predefinedSortValues: Map<String, Any>? = null,
```
gpunto, line 50: Is it ok that we define these as nullable but then they'll become empty maps when deserialized? Do you think it's worth to align the types/defaults?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/query/QueryChannelsSpec.kt
```diff
@@ -20,9 +20,46 @@ import io.getstream.chat.android.models.Channel
 import io.getstream.chat.android.models.FilterObject
 import io.getstream.chat.android.models.querysort.QuerySorter
 
+/**
+ * Spec describing a query channels operation and the channel CIDs that belong to it.
+ *
+ * Note on shape: the predefined-filter fields ([predefinedFilterName],
+ * [predefinedFilterValues], [predefinedSortValues]) are declared as body-level `var` properties
+ * rather than primary-constructor parameters specifically to keep the primary `<init>(filter,
+ * querySort)` and the synthesized `copy(filter, querySort)` JVM signatures unchanged.
```
gpunto: Could we instead introduce a secondary constructor + a copy function? Something like:

```kotlin
public data class QueryChannelsSpec(
    val filter: FilterObject,
    val querySort: QuerySorter<Channel>,
    val cids: Set<String> = emptySet(),
    val predefinedFilterName: String? = null,
    val predefinedFilterValues: Map<String, Any>? = null,
    val predefinedSortValues: Map<String, Any>? = null,
) {
    public constructor(
        filter: FilterObject,
        querySort: QuerySorter<Channel>,
    ) : this(filter, querySort, emptySet(), null, null, null)

    public fun copy(
        filter: FilterObject = this.filter,
        querySort: QuerySorter<Channel> = this.querySort,
    ): QueryChannelsSpec = QueryChannelsSpec(
        filter = filter,
        querySort = querySort,
        cids = cids,
        predefinedFilterName = predefinedFilterName,
        predefinedFilterValues = predefinedFilterValues,
        predefinedSortValues = predefinedSortValues,
    )
}
```

So we avoid the vars and we make sure `equals` and `hashCode` take all properties into account (at the moment, a predefined filter is treated as equal to a standard filter that has the same resolved filter/sort).

stream-chat-android-compose/api/stream-chat-android-compose.api
```diff
@@ -6617,9 +6619,10 @@ public final class io/getstream/chat/android/compose/viewmodel/channels/ChannelL
 
 public final class io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModelFactory : androidx/lifecycle/ViewModelProvider$Factory {
 	public static final field $stable I
-	public fun <init> ()V
```
gpunto, line 6697: Is it ok that we lost this? I believe it only affects java callers, but would break them if they just call `new ChannelListViewModelFactory()`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
VelikovPetar: Hm good point, I actually re-did this part couple of times, so the docs are a bit outdated. I will revisit this, taking your feedback into consideration.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/FilterDomainMapping.kt
VelikovPetar: I would eventually love to migrate to the type-safe filters from `core` - they would also make this parser much simpler (or not simpler, but rather safer)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/queryChannels/internal/QueryChannelsEntity.kt
VelikovPetar: Aham I missed this part. Hmm, I will need to go over the flow - I think it shouldn't be an issue, but I will double check.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/query/QueryChannelsSpec.kt
VelikovPetar: Will look into it!

stream-chat-android-compose/api/stream-chat-android-compose.api
VelikovPetar: Hmm good point, let me see what we can do about this.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
VelikovPetar: Addressed it by:
1. Early return on a `Standard` identifier
2. Update the docs to better reflect the logic

Let me know if you think it is OK now!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/queryChannels/internal/QueryChannelsEntity.kt
VelikovPetar: Implemented a new Converter for nullable maps!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/query/QueryChannelsSpec.kt
VelikovPetar: I took the approach you mentioned, looks to be ok 👍

stream-chat-android-compose/api/stream-chat-android-compose.api
VelikovPetar: I think that adding `@JvmOverloads` fixes this. (but I also realised that this is from the Compose SDK, so I doubt there would be any Java users affected)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
```diff
@@ -172,10 +209,50 @@ internal class QueryChannelsMutableState(
...
+     */
+    fun setCids(cids: Set<String>) {
+        _querySpec = QueryChannelsSpec(
+            filter = _querySpec.filter,
+            querySort = _querySpec.querySort,
+            cids = cids,
+            predefinedFilterName = _querySpec.predefinedFilterName,
+            predefinedFilterValues = _querySpec.predefinedFilterValues,
+            predefinedSortValues = _querySpec.predefinedSortValues,
+        )
```
gpunto: Nit: could this be `copy`?
```suggestion
        _querySpec = _querySpec.copy(cids = cids)
```

gpunto (approved):
LGTM!

stream-chat-android-compose/api/stream-chat-android-compose.api
gpunto: Ah right! 😓

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/querychannels/internal/QueryChannelsMutableState.kt
VelikovPetar: Done!

gpunto (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
```diff
@@ -150,20 +141,35 @@ internal class QueryChannelsLogic(
...
+                querySort = identifier.sort,
+                messageLimit = messageLimit,
+                memberLimit = memberLimit,
+            )
+            is QueryChannelsIdentifier.Predefined -> QueryChannelsRequest(
+                offset = INITIAL_CHANNEL_OFFSET,
+                limit = CHANNEL_LIMIT,
+                messageLimit = messageLimit,
+                memberLimit = memberLimit,
+                predefinedFilter = identifier.name,
```
andremion, line 165: **Sync recovery does not apply the server-resolved filter and sort for predefined-filter queries**

`queryFirstPage()` is called from `SyncManager.updateActiveQueryChannels:445` during recovery. `fetchPage()` (line 297 below) is called for pagination. Both go through `client.queryChannelsInternal(request)`, which maps the response to `it.channels` and ignores `QueryChannelsResult.predefinedFilter`. The listener path that calls `applyResolvedSpec()` does not run here. If the predefined template changed on the backend between sessions, the cached filter and sort stay outdated until the next user-initiated query.

The backend does return the resolved filter in `predefined_filter.filter`, so the data is in the response. It is just not used.

Suggested fix: route both call sites through `chatClient.queryChannels(request)`, which goes through the plugin listeners. Or add an internal variant that returns the full `QueryChannelsResult` and call `applyResolvedSpec` from the logic before forwarding channels.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -903,4 +906,32 @@ internal class DomainMapping(
...
+            }
+        }
+    }
+
+    /**
+     * Computes the default `predefined_filter.sort` the server would apply when no sort was sent.
+     * Mirrors the backend's behaviour: `{ last_updated: -1 }` by default, swapping to
+     * `{ last_message_at: -1 }` when `last_message_at` is referenced as a filter condition.
+     */
+    internal fun defaultPredefinedFilterSort(filterFields: Set<String>): QuerySorter<Channel> {
```
andremion, line 928: **`defaultPredefinedFilterSort` does not match the backend behaviour**

The fallback (added in commit `1ff4411f0f`) defaults to `last_message_at` desc when the filter references it, else `last_updated` desc. When the predefined filter has no registered sort, the backend omits `predefined_filter.sort` from the response and applies its own internal default — which is not what this heuristic does.

Effect: predefined filters registered without a sort template will use a different order on Android than the server uses for pagination. Channels can change order between pages.

Suggested fixes (any one):
- Require a sort template on every predefined filter; log if missing.
- Ask the backend to always include the sort it used in `predefined_filter.sort`.
- Drop the heuristic and use `QuerySortByField()` so the comparator does not override server order.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelsRequest.kt
```diff
@@ -18,26 +18,36 @@ package io.getstream.chat.android.client.api.models
...
  * @property memberLimit Number of members in the response. When `null`, the server-side default is used.
+ * @property predefinedFilter ID of a server-side predefined filter to use instead of [filter].
+ * When set, [filter] and [querySort] are ignored by the backend.
+ * @property filterValues Values to interpolate into the predefined filter template.
+ * @property sortValues Values to interpolate into the predefined sort template.
  */
-public data class QueryChannelsRequest(
-    public val filter: FilterObject,
+public data class QueryChannelsRequest @JvmOverloads constructor(
+    public val filter: FilterObject = Filters.neutral(),
```
andremion: **`filter` changed from required to defaulted with `Filters.neutral()`**

This change is needed for the predefined-filter path. For standard callers, it means that `QueryChannelsRequest(limit = 30)` is now a valid construction, which is probably a misuse.

Two options:
- Make `filter` required again, and add a separate constructor (or factory) for the predefined path that does not ask for a filter.
- Keep the default, but add `require(predefinedFilter != null || filter !is NeutralFilterObject) { … }` at construction time, so misuse fails immediately.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/QueryChannelsIdentifier.kt
```diff
@@ -0,0 +1,83 @@
...
+ * across runs.
+ *
+ * Two shapes are supported:
+ *  - [Standard] for classic queries where the client knows `filter` + `querySort` upfront.
+ *  - [Predefined] for server-side predefined filters where the actual `filter` and `querySort` are
+ *    only learned from the response. Identity must therefore be the predefined name plus the
+ *    interpolation values, since those are the only stable inputs available before the response.
+ */
+@InternalStreamChatApi
+public sealed interface QueryChannelsIdentifier {
```
andremion, line 39: **Avoid `public sealed interface` for `QueryChannelsIdentifier`**

Per engineering principles (`Idiomatic Kotlin > Sealed classes`): avoid `public sealed` types likely to gain new variants — once public-sealed, adding a variant is a binary + source break. `QueryChannelsIdentifier` will likely grow (saved queries, server-defined views, etc.). The `@InternalStreamChatApi` annotation is advisory only and doesn't stop external `when` usage.

Suggested shape:

```kotlin
public abstract class QueryChannelsIdentifier internal constructor() {
    public data class Standard(
        val filter: FilterObject,
        val sort: QuerySorter<Channel>,
    ) : QueryChannelsIdentifier()

    public data class Predefined(
        val name: String,
        val filterValues: Map<String, Any>?,
        val sortValues: Map<String, Any>?,
    ) : QueryChannelsIdentifier()
}
```

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelsResult.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Result wrapper for [io.getstream.chat.android.client.api.ChatApi.queryChannels].
+ * Holds both the list of channels and the optional parsed predefined filter returned by the backend.
+ *
+ * @param channels The list of channels returned by the query.
+ * @param predefinedFilter The parsed predefined filter metadata, or null if a regular filter was used.
+ */
+@InternalStreamChatApi
+public data class QueryChannelsResult(
```
andremion, line 30: **`@InternalStreamChatApi` types are part of the public `ChatApi.queryChannels` signature**

`ChatApi` is in the public package `io.getstream.chat.android.client.api` and exposes `fun queryChannels(...): Call<QueryChannelsResult>`. Both `QueryChannelsResult` and `PredefinedFilter` are annotated `@InternalStreamChatApi`, but they remain part of the public binary surface — the annotation is advisory and doesn't stop external consumers.

If they're meant to be internal:
- Make them `internal` and add an internal-only `ChatApi.queryChannelsWithMeta(...)`, keeping the public `ChatApi.queryChannels` returning `Call<List<Channel>>`, or
- Drop the `@InternalStreamChatApi` annotation since it can't be enforced here.

Small style note: nearby data classes use `@property` for constructor properties; this one uses `@param`. Worth aligning.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/PredefinedFilter.kt
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Represents a predefined filter parsed by the backend.
+ *
+ * @param name The name/identifier of the predefined filter.
+ * @param filter The parsed filter specification.
+ * @param sort The parsed sort specification, or null if no sort was provided.
+ */
+@InternalStreamChatApi
+public data class PredefinedFilter(
+    val name: String,
```
andremion: **`PredefinedFilter.name` is set but never read**

The listener at `QueryChannelsListenerState:60-65` reads `resolved.filter` and `resolved.sort`. It never reads `resolved.name`. If the field is only for debugging, please use it in the logger tag or in error breadcrumbs. If it has no use, please remove it.

Same KDoc-style note as on `QueryChannelsResult`: this file uses `@param` for constructor properties; nearby data classes use `@property`. Worth aligning.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/viewmodel/channels/ChannelListViewModel.kt
```diff
@@ -83,19 +81,84 @@ import kotlinx.coroutines.launch
...
+     *
+     * The filter and sort are identified by [predefinedFilterName] and resolved server-side;
+     * [filterValues] and [sortValues] interpolate into the predefined template. [setFilters] does not
+     * affect a view model created this way.
+     *
+     * @param predefinedFilterName The name of the predefined filter registered on the backend.
+     * @param filterValues Optional values interpolated into the predefined filter template.
+     * @param sortValues Optional values interpolated into the predefined sort template.
+     */
+    public constructor(
```
andremion, line 136: **UI-Components ViewModel has no channel-search path in predefined mode**

CodeRabbit already raised the "search escapes the backend-resolved query" problem on the Compose `ChannelListViewModel`. Worth noting the difference here: this UI-Components ViewModel has no `setSearchQuery` method at all. So search in predefined mode is "not supported" on UI-Components and "silently runs a different query" on Compose. Whichever fix is chosen for Compose, please apply the same behaviour here so both SDKs behave the same way.

andremion:
**Two test gaps worth covering (no inline anchor):**

- No test exercises sync recovery for a predefined-filter query (Finding 1 in the review). A test would catch the case where `queryChannelsInternal` ignores the resolved filter and sort.
- No test asserts DB identity stability for numeric `filterValues` across a write → clear → read cycle (Finding 2 in the review). A test would catch the Moshi `Double` round-trip.

The search-related and unknown-operator test gaps are already noted in CodeRabbit's threads on `ChannelListViewModelTest.kt:616` and `MoshiChatApiTestArguments.kt:461`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1291,7 +1297,24 @@ constructor(
...
-            ).map { response -> response.channels.map(this::flattenChannel) }
+            ).map { response ->
+                with(domainMapping) {
+                    QueryChannelsResult(
+                        channels = response.channels.map(this@MoshiChatApi::flattenChannel),
+                        predefinedFilter = response.predefined_filter?.let {
+                            val (filter, filterFields) = it.filter.toFilterDomainWithFields()
+                                ?: return@let null
+                            val sort = it.sort.toSortDomain()
+                                ?: defaultPredefinedFilterSort(filterFields)
```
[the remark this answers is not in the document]
andremion: The server-side MQ parser also accepts `$q`, `$any`, and `$path_exists` (with `$search` aliased to `$q`). `FilterDomainMapping` does not handle any of these. So this is not theoretical: any predefined filter registered with `$q` or `$any` will reach the silent-null path. Worth either teaching the parser these operators, or defining the contract when the resolved filter has a valid shape but uses an operator the parser does not know.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/query/QueryChannelsSpec.kt
```diff
@@ -20,9 +20,41 @@ import io.getstream.chat.android.models.Channel
...
+ * `QueryChannelsMutableState.applyResolvedSpec`).
+ *
+ * The 2-arg [constructor] and 2-arg [copy] are kept for binary compatibility with callers that
+ * predate the predefined-filter fields. They delegate to the primary constructor with the
+ * predefined fields defaulted to their empty/null values.
+ */
 public data class QueryChannelsSpec(
     val filter: FilterObject,
     val querySort: QuerySorter<Channel>,
+    val cids: Set<String> = emptySet(),
```
andremion: **`cids` changed from `var` to `val` — please mention in the CHANGELOG**

The setter is gone. Any external code that previously did `spec.cids = …` will fail to link. Inside the SDK this is fine — `QueryChannelsMutableState.setCids` already uses `.copy(cids = …)`. Worth a CHANGELOG note so external consumers aren't surprised.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
VelikovPetar: This is actually the info that I got from the BE team: https://getstream.slack.com/archives/CE5N802GP/p1778226798199859

andremion: Given that this was already verified with the BE team. The heuristic does match the server's actual default, so the bug framing of my comment doesn't apply.
The only residual concern (non-blocking): since the client mirrors a server-side default that isn't surfaced in the response, any backend change (extra tiebreaker, different field, etc.) would silently diverge from the Android comparator. A follow-up to have the backend always include the resolved sort in `predefined_filter.sort` would remove that risk.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/state/StateRegistry.kt
```diff
@@ -84,9 +88,34 @@ public class StateRegistry @JvmOverloads constructor(
...
+                is QueryChannelsIdentifier.Predefined -> Filters.neutral() to QuerySortByField()
+            }
+            QueryChannelsMutableState(
+                identifier = identifier,
+                initialFilter = initialFilter,
+                initialSort = initialSort,
+                scope = scope,
+                latestUsers = latestUsers,
+                activeLiveLocations = activeLiveLocations,
+            )
```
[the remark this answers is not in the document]
VelikovPetar: This should be fine, because the seed is temporary - it is either populated immediately from DB (if known), or from server for the first time. And events are not dispatched until the QueryChannels (with watch = true) completes, so the window is extremely small (get a WS event after the request was processed by the server, but before the response is returned from to the sender). I think this is safe to skip

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/FilterDomainMapping.kt
```diff
@@ -0,0 +1,151 @@
...
+            false -> Filters.notExists(fieldName)
+            null -> null
+        }
+        KEY_AUTOCOMPLETE -> {
+            val strValue = opValue as? String ?: return null
+            Filters.autocomplete(fieldName, strValue)
+        }
+        else -> null
+    }
+}
```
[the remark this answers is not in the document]
VelikovPetar: This should not be possible - I will double check with BR and get back here

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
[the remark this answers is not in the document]
VelikovPetar: The response is guaranteed to contain `filter`, but not `sort` (see [specs](https://getstream.github.io/protocol/#/product%3Achat/QueryChannels)). So I think we should fail on `filter`, but not on `sort` - and use the BE provided fallback. 

Concerning the missing filter types: I think we should do this in a follow up: We currently don't support those in the upstream case as well (`Filters`/`FilterObject`), so parsing them currently doesn't really make sense.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
```diff
@@ -269,6 +274,32 @@ internal class QueryChannelsLogic(
...
+                filterValues = identifier.filterValues,
+                sortValues = identifier.sortValues,
+            )
+        }
+        return when (val result = client.queryChannelsInternal(request).await()) {
+            is Result.Success -> result.value
+            is Result.Failure -> emptyList()
+        }
+    }
```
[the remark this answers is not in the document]
VelikovPetar: Pre-existing behavior, just relocated code. I already created a followup, because we should rethink this whole logic: https://linear.app/stream/issue/AND-1172.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTestArguments.kt
```diff
@@ -439,6 +439,27 @@ internal object MoshiChatApiTestArguments {
...
+                            hide_messages_before = randomDateOrNull(),
+                            draft = randomDownstreamDraftDto(),
+                        ),
+                    ),
+                ),
+            ).toRetrofitCall(),
+            Result.Success::class,
+        ),
+        Arguments.of(RetroError<QueryChannelsResponse>(statusCode = 500).toRetrofitCall(), Result.Failure::class),
+    )
```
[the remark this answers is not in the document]
VelikovPetar: Will update it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
```diff
@@ -131,14 +201,26 @@ public class ChannelListViewModel(
...
+     * State flow that keeps the value of the current [QuerySorter] for channels. Only meaningful in
+     * [QueryMode.Standard]; in [QueryMode.Predefined] it carries an inert default (the server owns the sort).
      */
-    private val querySortFlow: MutableStateFlow<QuerySorter<Channel>> = MutableStateFlow(initialSort)
+    private val querySortFlow: MutableStateFlow<QuerySorter<Channel>> = MutableStateFlow(
+        when (mode) {
+            is QueryMode.Standard -> mode.initialSort
+            is QueryMode.Predefined -> QuerySortByField()
+        },
+    )
```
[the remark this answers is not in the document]
VelikovPetar: This is actually intentional. We should slowly migrate away from using `member.user.name` in the default case as it is extremely expensive + creating complex filters (with multiple ANDs). Since we cannot change the existing logic as it will be a breaking change, we can migrate the new case to use the better approach (which also aligns the behaviour with iOS). It will be documented in the documentation.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModelFactory.kt
```diff
@@ -43,18 +43,81 @@ import io.getstream.chat.android.models.querysort.QuerySorter
...
+            filterValues = filterValues,
+            sortValues = sortValues,
+        ),
+        channelLimit = channelLimit,
+        memberLimit = memberLimit,
+        messageLimit = messageLimit,
+        chatEventHandlerFactory = chatEventHandlerFactory,
+        draftMessagesEnabled = draftMessagesEnabled,
+        messageSearchSort = messageSearchSort,
+    )
```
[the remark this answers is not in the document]
VelikovPetar: Will add it

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModelTest.kt
```diff
@@ -532,6 +533,128 @@ internal class ChannelListViewModelTest {
...
+            viewModel.setSearchQuery(SearchQuery.Channels("Search query"))
+            advanceUntilIdle()
+            viewModel.setSearchQuery(SearchQuery.Empty)
+            advanceUntilIdle()
+
+            val captor = argumentCaptor<QueryChannelsRequest>()
+            verify(chatClient, times(3)).queryChannels(captor.capture())
+            val revertedRequest = captor.thirdValue
+            assertEquals("vip_filter", revertedRequest.predefinedFilter)
+        }
```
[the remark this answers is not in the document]
VelikovPetar: This is actually intentional. We should slowly migrate away from using member.user.name in the default case as it is extremely expensive + creating complex filters (with multiple ANDs). Since we cannot change the existing logic as it will be a breaking change, we can migrate the new case to use the better approach (which also aligns the behaviour with iOS). It will be documented in the documentation.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
VelikovPetar: Good point - I would rather change the existing `queryChannelsInternal` to return `Result<QueryChannelsResult>`, and call `applyResolvedSpec` directly from here. I think changing the invocation to `chatClient.queryChannels(request)` could have unforeseen side-effect.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelsRequest.kt
VelikovPetar: Let me see if we can define a new constructor specific for the predefined filters case.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/QueryChannelsIdentifier.kt
VelikovPetar: Hmm I am not sure about this one. Since it was marked as @InternalStreamChatApi, I don't think we should worry about breaking changes here. Opting in using this API, means the customer is aware that this might change in the future.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelsResult.kt
VelikovPetar: Hmm But `ChatApi` is already internal... 
I think we have to keep this marked as @InternalStramChatApi because it used in the `QueryChannelsListener`, which is also public, but marked as @InternalStramChatApi

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/PredefinedFilter.kt
VelikovPetar: Yeah we can remove it, the state/db layer use the user-provided name anyway.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/viewmodel/channels/ChannelListViewModel.kt
VelikovPetar: Yes but I would say that we shouldn't introduce the search logic in the XML ChannelListViewModel. It is an existing discrepancy, plus this PR is about adding support for predefined filters, not aligning XML to Compose.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/query/QueryChannelsSpec.kt
VelikovPetar: Good catch - I will need to revert this as it is a potential breaking change.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTestArguments.kt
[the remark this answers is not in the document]
VelikovPetar: Added https://github.com/GetStream/stream-chat-android/pull/6415/commits/d2a71dd9dc208c78fa0bb6926bd8b2521c26b0f1

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModelFactory.kt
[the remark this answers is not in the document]
VelikovPetar: Added https://github.com/GetStream/stream-chat-android/pull/6415/commits/d2a71dd9dc208c78fa0bb6926bd8b2521c26b0f1

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
VelikovPetar: DOne https://github.com/GetStream/stream-chat-android/pull/6415/commits/d2a71dd9dc208c78fa0bb6926bd8b2521c26b0f1

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/QueryChannelsRequest.kt
VelikovPetar: Updated here, let me know what do you think: https://github.com/GetStream/stream-chat-android/pull/6415/commits/d2a71dd9dc208c78fa0bb6926bd8b2521c26b0f1

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/models/PredefinedFilter.kt
VelikovPetar: Removed here: https://github.com/GetStream/stream-chat-android/pull/6415/commits/d2a71dd9dc208c78fa0bb6926bd8b2521c26b0f1

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/query/QueryChannelsSpec.kt
VelikovPetar: Updated here, please have a look, I want to make sure I don't miss anything again: https://github.com/GetStream/stream-chat-android/pull/6415/commits/d2a71dd9dc208c78fa0bb6926bd8b2521c26b0f1

andremion (approved):

VelikovPetar:
> **Two test gaps worth covering (no inline anchor):**
> 
> * No test exercises sync recovery for a predefined-filter query (Finding 1 in the review). A test would catch the case where `queryChannelsInternal` ignores the resolved filter and sort.
> * No test asserts DB identity stability for numeric `filterValues` across a write → clear → read cycle (Finding 2 in the review). A test would catch the Moshi `Double` round-trip.
> 
> The search-related and unknown-operator test gaps are already noted in CodeRabbit's threads on `ChannelListViewModelTest.kt:616` and `MoshiChatApiTestArguments.kt:461`.

Hi @andremion, concerning this remark: 
1. I am adding a test for this case 👍 
2. This is very valid concern - round trip is not preserved for numeric values via the `NullableMapConverter` and could be an issue. However, we would never actually encounter (at least for now) this bug, with how the QueryChannelsLogic Currently works. We are are always either writing to the DB using the user-provided `filterValues/sortValues`, or reading from the DB using the same user-provided `filterValues/sortValues`. So in theory, unless the user starts passing something like `member_count: 2.0` instead of `member_count: 2` we would never encounter the issue. And even if this happens, the problematic case would happen only on the first switch. 
This could be a serious issue, if we start reading/writing the `QueryChannelsSpec` data to/from DB, while mixing user-provided filter/sort values, and values parsed from previous DB reads. 

Do you think we should harden this logic to also ensure proper round-trip parsing, to bypass the Moshi Double parsing limitation?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/FilterDomainMapping.kt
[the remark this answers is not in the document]
VelikovPetar: This is not really possible, the operators are `single operator per field`.

andremion:
> Hi @andremion, concerning this remark:
> 
> 1. I am adding a test for this case 👍
> 2. This is very valid concern - round trip is not preserved for numeric values via the `NullableMapConverter` and could be an issue. However, we would never actually encounter (at least for now) this bug, with how the QueryChannelsLogic Currently works. We are are always either writing to the DB using the user-provided `filterValues/sortValues`, or reading from the DB using the same user-provided `filterValues/sortValues`. So in theory, unless the user starts passing something like `member_count: 2.0` instead of `member_count: 2` we would never encounter the issue. And even if this happens, the problematic case would happen only on the first switch.
>    This could be a serious issue, if we start reading/writing the `QueryChannelsSpec` data to/from DB, while mixing user-provided filter/sort values, and values parsed from previous DB reads.
> 
> Do you think we should harden this logic to also ensure proper round-trip parsing, to bypass the Moshi Double parsing limitation?

No, I wouldn't harden it. The current code is correct as long as identity always flows from the live user-provided maps, which it does. Hardening generateId to normalize values would add complexity for a problem that doesn't exist today.
If you want a cheap defensive measure, a regression test ("write -> read with same live identifier, numeric values, lookup succeeds") would lock the invariant, so any future refactor that starts using deserialized values for identity fails loudly.

PR #6423, opened 2026-05-07
Add spacing between composer snackbar and input row

andremion:
### Goal

Address the [SDK Testing Feedback ticket](https://www.notion.so/3516a5d7f9f68000b928f912e8826133) to fix the vertical snackbar position.

### Implementation

- Add `MessageComposerSnackbar` to `ChatComponentFactory` with a matching `MessageComposerSnackbarParams` holder, mirroring the pattern used by the audio-recording snackbars. The default applies 16 dp bottom padding around the visible card.
- `StreamSnackbar.modifier` now applies to the outer `Box` instead of the inner `Surface` (idiomatic Compose; internal-only API).
- Replace the hardcoded `4.dp` shadow with `StreamTokens.elevation3`.
- Drop a redundant `padding(vertical = spacing2xs)` on the input `Row` (no Figma backing) and a `wrapContentHeight()` modifier on the `MessageComposer` call in `ChannelScreen` (had no effect — the parent `Column` already grants intrinsic vertical sizing).
- Re-record `MessageComposerTest` snapshots.

### Fixed Style

| Attachment Picker | Suggestions Popup |
| --- | --- |
| Screenshot_20260507_111852 | Screenshot_20260507_111841 |

### Floating Style

| Attachment Picker | Suggestions Popup |
| --- | --- |
| Screenshot_20260507_112550 | Screenshot_20260507_112608 |

### Testing

1. Open any channel in the Compose sample, reply to a message, then trigger an unavailable command (e.g. `/giphy` while replying).
   - Expected: snackbar above the input with a clear gap.
2. Open the attachment picker → commands tab → tap an unavailable command in the same reply state.
   - Expected: same spacing.
3. Toggle floating composer and repeat.
   - Expected: snackbar still sits above the floating composer with the gap.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed message composer snackbar display functionality.

* **Style**
  * Updated shadow elevation design tokens for improved visual consistency across snackbar components.

* **Refactor**
  * Enhanced message composer UI layout and IME window inset handling for better responsiveness.


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -330,7 +330,12 @@ public fun MessageComposer(
             )
 
             if (snackbarHostState.currentSnackbarData != null) {
-                SnackbarPopup(hostState = snackbarHostState)
+                SnackbarPopup(hostState = snackbarHostState) { snackbarData ->
+                    StreamSnackbar(
+                        modifier = Modifier.padding(bottom = StreamTokens.spacingMd),
+                        snackbarData = snackbarData,
+                    )
+                }
```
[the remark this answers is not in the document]
andremion: The code is correct and visually verified by the sample app

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
```diff
@@ -1282,6 +1282,15 @@ public data class MessageComposerAudioRecordingHintParams(
     val data: SnackbarData,
 )
 
+/**
+ * Parameters for [ChatComponentFactory.MessageComposerSnackbar].
+ *
+ * @param data The snackbar data containing the message, variant, and optional action.
+ */
+public data class MessageComposerSnackbarParams(
+    val data: SnackbarData,
```
VelikovPetar, line 1291: Would it make sense to pass a `Modifier` here as well?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1705,6 +1705,22 @@ public interface ChatComponentFactory {
...
+     * events.
+     *
+     * Override this method to provide a custom snackbar appearance for the composer.
+     *
+     * @param params Parameters for this component.
+     */
+    @Composable
+    public fun MessageComposerSnackbar(params: MessageComposerSnackbarParams) {
+        StreamSnackbar(
+            modifier = Modifier.padding(bottom = StreamTokens.spacingMd),
```
VelikovPetar, line 1719: Is it OK to always pass `Modifier.padding(bottom = StreamTokens.spacingMd)` here? Mostly concerned about a case where we would need to provide a different bottom spacing in a new feature. 
I am thinking if it would be ok to pass the padding modifier from outside, giving us more flexibility to style the snackbar.

andremion: Thanks for the suggestion. I think we should keep this as is. Here is why:
`MessageComposerSnackbar` is called in only one place: `MessageComposer`. The SDK builds the params there with `MessageComposerSnackbarParams(data = snackbarData)`. Integrators cannot change what is passed.
So adding a modifier field to the params does not help. No one outside the SDK has a way to set it. The way to customize the snackbar is to override the factory function. Inside the override, the developer writes the modifier they want directly. The field would not be used in practice.
In a future feature, we would have another factory because that one is specific to the message composer (we have `MessageComposerAudioRecordingPermissionRationale`, `MessageComposerAudioRecordingHint` as examples)
If you have a concrete case where this is a problem, please share it, and we can change the approach.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
andremion: https://github.com/GetStream/stream-chat-android/pull/6423#discussion_r3208627195

VelikovPetar (approved):

PR #6429, opened 2026-05-08
Add configurable character limits for poll creation

ryanhurststrava:
## 🎯 Goal
Add configurable character limits for poll titles and answer options to provide developers with flexibility in controlling input length during poll creation. This addresses use cases where different character limits may be needed based on UI constraints, localization requirements, or business rules.

### New Files Created
**`CreatePollDialogFragment.kt`**
- Added new dialog fragment for creating polls with configurable character limits
- Implemented `newInstance()` factory method with optional `titleCharLimit` and `answerCharLimit` parameters
- Parameters are passed via `Bundle` arguments using keys `ARG_TITLE_CHAR_LIMIT` and `ARG_ANSWER_CHAR_LIMIT`
- Applied character limits using `InputFilter.LengthFilter` on the question/title EditText field
- Added `CreatePollDialogListener` interface for handling poll creation and dismissal events
- Integrated with `CreatePollViewModel` for state management
- Full edge-to-edge UI support with proper inset handling

**`OptionsAdapter.kt`**
- Created RecyclerView adapter for poll options with configurable character limits
- Accepts optional `answerCharLimit` parameter that's applied to each option EditText
- Implements `InputFilter.LengthFilter` for answer options when limit is specified
- Includes duplicate detection and error state management
- Uses `DiffUtil` for efficient list updates
- Stable IDs implementation for better RecyclerView performance

### Key Design Decisions
- **Backward Compatibility**: Both `titleCharLimit` and `answerCharLimit` default to `null`, meaning no character limit is enforced unless explicitly specified
- **Optional Configuration**: Developers can choose to set limits on title only, answers only, both, or neither
- **InputFilter Approach**: Used Android's native `InputFilter.LengthFilter` for enforcing limits at the input level

### Usage Example
```kotlin
CreatePollDialogFragment.newInstance(
    createPollDialogListener = myListener,
    titleCharLimit = 100,
    answerCharLimit = 50
)
```

## 📱 UI Changes
Character limits are enforced at the input level using InputFilter, preventing users from exceeding the specified length. 

https://github.com/user-attachments/assets/f09ac394-c58c-4489-bfca-c2c2447f6690

### Manual Testing Steps
1. Create a poll without specifying limits - verify no restrictions are applied (backward compatibility)
2. Create a poll with `titleCharLimit = 100` - verify title input stops at 100 characters
3. Create a poll with `answerCharLimit = 50` - verify each answer option stops at 50 characters
4. Create a poll with both limits - verify both work independently
5. Test edge cases: limit of 1, very large limits, and null values

### Expected Behavior
- When limits are not specified, users can type unlimited characters (default behavior)
- When limits are specified, `InputFilter.LengthFilter` prevents input beyond the limit
- No runtime crashes or errors when switching between limited and unlimited configurations

---

### Contributor Checklist
- [ ] I have assigned a reviewer (code owner)
- [x] This PR targets the `develop` branch
- [ ] I have added unit tests for new code
- [x] I have updated documentation (KDocs included in code)

### Reviewer Checklist
- [ ] UI Components sample app has been tested
- [ ] Compose sample app has been tested (N/A - UI Components only)
- [ ] Visuals are correct
- [ ] Feature is validated
- [ ] KDocs have been reviewed
- [ ] SDK size impact has been reviewed from CI logs

---

## 🎉 GIF
_Screenshots and demo to be added_

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Poll questions now support optional character limit configuration
  * Poll answer options now support optional character limit configuration
  * These limits help maintain consistent formatting and prevent excessively long poll content

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6429)

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
@@ -17,6 +17,7 @@
 package io.getstream.chat.android.ui.feature.messages.composer.attachment.picker.poll
 
 import android.os.Bundle
+import android.text.InputFilter
 import android.view.LayoutInflater
 import android.view.MenuItem
 import android.view.View
@@ -48,7 +49,10 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
     private var createPollDialogListener: CreatePollDialogListener? = null
     private val createPollViewModel: CreatePollViewModel by viewModels()
     private val optionsAdapter: OptionsAdapter by lazy {
-        OptionsAdapter { id, text -> createPollViewModel.onOptionTextChanged(id, text) }
+        OptionsAdapter(
+            answerCharLimit = arguments?.getInt(ARG_ANSWER_CHAR_LIMIT)?.takeIf { it > 0 },
+            onOptionChange = { id, text -> createPollViewModel.onOptionTextChanged(id, text) },
+        )
     }
     private lateinit var sendMenuItem: MenuItem
 
@@ -81,6 +85,9 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
      */
     private fun setupDialog() {
         setupToolbar(binding.toolbar)
+        arguments?.getInt(ARG_TITLE_CHAR_LIMIT)?.takeIf { it > 0 }?.let { limit ->
+            binding.question.filters = arrayOf(InputFilter.LengthFilter(limit))
+        }
         binding.multipleAnswersSwitch.setOnCheckedChangeListener { _, isChecked ->
             binding.multipleAnswersCount.isVisible = isChecked
             createPollViewModel.setAllowMultipleVotes(isChecked)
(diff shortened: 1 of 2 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
```diff
@@ -158,15 +165,28 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
...
+         * @param titleCharLimit Optional character limit for the poll title. Null means no limit.
+         * @param answerCharLimit Optional character limit for poll answer options. Null means no limit.
          * @return A new instance of [CreatePollDialogFragment].
          */
-        public fun newInstance(createPollDialogListener: CreatePollDialogListener): CreatePollDialogFragment {
-            return CreatePollDialogFragment()
-                .setCreatePollDialogListener(createPollDialogListener)
+        public fun newInstance(
+            createPollDialogListener: CreatePollDialogListener,
+            titleCharLimit: Int? = null,
```
VelikovPetar: Adding the `titleChatLimit/answerChatLimit` here would also be a breaking change for Java integrations. We should add a `JvmOverloads` annotation to this method, so that the corresponding overloads are generated for the java code.
I would also rename the parameters to:
1. `questionTextLimit`
2. `optionTextLimit`
To be better aligned with the class naming.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/OptionsAdapter.kt
```diff
@@ -27,6 +28,7 @@ import io.getstream.chat.android.ui.databinding.StreamUiPollOptionBinding
 import io.getstream.chat.android.ui.utils.extensions.streamThemeInflater
 
 public class OptionsAdapter(
+    private val answerCharLimit: Int?,
```
VelikovPetar: As the `OptionsAdapter` is a public facing API, adding a mandatory argument at the start of the constructor would be a breaking change as customers who potentially instantiate the `OptionsAdapter` manually, would encounter a compilation error when updating the SDK. 

In order to prevent a breaking both for Java and Kotlin consumers of the API, we would need to introduce a secondary constructor: 
```kotlin
// Extend the default constructor with the new parameter

public class OptionsAdapter(
    private val optionTextLimit: Int?,
    private val onOptionChange: (id: Int, text: String) -> Unit,
) : ListAdapter<PollAnswer, OptionsAdapter.OptionViewHolder>(OptionDiffCallback) {

    // Introduce the new constructor, mimicking the original, primary constructor signature

    /**
     * Builds an [OptionsAdapter] instance without providing option text limit.
     *
     * @param onOptionChange Callback invoked when the option text changes.
     */
    public constructor(onOptionChange: (id: Int, text: String) -> Unit) : this(null, onOptionChange)


    // ....
}
```

I would also rename the `answerCharLimit` to `optionTextLimit`, to be a bit more aligned with the class name.

VelikovPetar (commented):
Hi @ryanhurststrava,

Thank you for the contribution!

I've left several remarks concerning breaking changes in the public API, which we are not allowed to make, unless it is a major release. 

I have an additional question: The changes that you made, indicate that you directly call `CreatePollDialogFragment.newInstance` from your code - can you confirm that my assumption is correct? 
Otherwise, there is no other public way to inject these values, which means that if you are using the XML SDK as-is (without intercepting the poll click), it will not be possible to customise these values.

ryanhurststrava:
> The changes that you made, indicate that you directly call CreatePollDialogFragment.newInstance from your code - can you confirm that my assumption is correct?

yeah this is correct 👍

ryanhurststrava:
update: I need some additional configuration here to support hiding of the comment and suggestion fields when creating a poll.

I opened a new pr https://github.com/GetStream/stream-chat-android/pull/6435 with some more customization

PR #6404, opened 2026-04-30
Composer attachment picker accessibility improvements

andremion:
### Goal

Address the accessibility audit ([Notion DB](https://www.notion.so/stream-wiki/c2599560f8774c95b7d904b5ad49d6fd?v=6a01d11916874a85908ab37c2b40afb2)). All findings concentrate on the **attachment picker**.

### Implementation

| Audit | Fix | Files |
|---|---|---|
| A1 (Critical) | Media grid items announce label / role / selected state | `ImagesPicker.kt` |
| A2 | Picker tabs use `Role.Tab` via `clearAndSetSemantics` (suppresses `ToggleableState`) | `AttachmentTypePicker.kt` |
| A3, A4(1) | Composer button uses `contentDescription = "Attachments"` + `stateDescription = "collapsed" / "expanded"`. `paneTitle` on the picker root announces the new pane on entry | `MessageComposerLeadingContent.kt`, `AttachmentPicker.kt` |
| A5(1), A8(1) | Accessibility announcements ("Photo / Video / File / Audio recording attached" / "Attachment removed") via `View.announceForAccessibility()` | `MessageComposerAttachmentAnnouncer.kt`, `MessageInput.kt` |
| A5(2), A10 | Action hint flips between "select" / "remove" based on `isSelected` | `ImagesPicker.kt`, `FilesPicker.kt` |
| A6 | Composer chip × labelled "Remove attachment" via additive `contentDescription` parameter on `ComposerCancelIcon` | `ComposerCancelIcon.kt` + 3 chip call sites |
| A9 | CreatePoll sheet announces the screen title before the auto-focused field via `paneTitle` + `heading()` | `CreatePollScreen.kt`, `PollCreationHeader.kt` |

#### Designer-aligned scope

Audit **A7** (focus → text input on chip remove) is resolved by the new "Attachment removed" announcement: the literal fix pops the keyboard mid-picker, and TalkBack's natural left-to-right / top-to-bottom progression after removal lands on the mic — the announcement provides the audible confirmation. Designer accepted; A7 is closed in Notion. The **focus halves** of A4 and A8 share the same root cause (Compose programmatic focus moves don't reliably propagate to TalkBack across our merged-semantics tab nodes) and ship in the same state pending designer alignment.

#### API surface

- `ComposerCancelIcon` gains a defaulted `contentDescription: String` parameter — additive, non-breaking per project convention. `apiDump` reflects the addition.
[...]

PR #6433, opened 2026-05-11
Disambiguate composer attachment remove buttons for TalkBack

andremion:
### Goal

Resolve [Accessibility issue 88 — Attachment Picker](https://www.notion.so/stream-wiki/Attachment-Picker-3576a5d7f9f680389978efb4165b2a9d)

The (x) remove button on each queued attachment in the Compose composer is announced as the same generic "Remove attachment" for every item. With multiple attachments queued, TalkBack users could not tell the remove buttons apart and had no way to confirm which attachment they were about to delete.

### Implementation

Replace the single `R.string.stream_compose_remove_attachment` with per-type, name-aware labels resolved locally inside each composer attachment item:

- **Photos / videos** (`MessageComposerAttachmentMediaItem`) — `"Remove photo IMG_0023"` / `"Remove video clip.mp4"`, falling back to `"Remove photo"` / `"Remove video"` when the attachment exposes no title or name. Type comes from `attachment.type == AttachmentType.VIDEO`.
- **Other files** (`MessageComposerAttachmentFileItem`) — `"Remove file report.pdf"`, falling back to `"Remove file"`.
- **Voice recordings** (`MessageComposerAttachmentAudioRecordItem`) — `"Remove voice message"` (recordings have no useful filename to surface).

Display name follows the codebase's standard pick `attachment.title ?: attachment.name`. The new string resources are added with translations for all 7 supported locales (es, fr, hi, in, it, ja, ko). The previously generic `stream_compose_remove_attachment` resource has no remaining callers and is dropped from all locales — same cleanup pattern as #6404, which removed `stream_compose_attachments` after the picker-button a11y refactor.

No public API surface changes (these composables are routed through `ChatTheme.componentFactory.MessageComposerAttachment{File,Media,AudioRecord}Item` and `apiDump` produces no diff).

### Testing

Manual steps on the Compose sample with TalkBack enabled:

1. Open a 1:1 channel and tap the attachments button.
2. Pick **two photos with distinct filenames** (e.g. from the device gallery). Wait for them to appear in the composer tray.
3. Swipe through the tray with TalkBack. Each (x) button should announce as `"Remove photo <filename>"` — the two announcements must differ.
4. Pick a **video** instead and confirm `"Remove video <filename>"`.
5. Switch to the **Files** tab in the picker, pick a PDF, and confirm `"Remove file <filename>"`.
6. **Record a voice message** (hold-to-record). The remove (x) should announce as `"Remove voice message"`.
7. Repeat steps 2–6 with the device language set to one of: Spanish, French, Hindi, Indonesian, Italian, Japanese, Korean — labels should be translated.
8. Edge case: if you have a way to inject an attachment with no `title` and no `name` (e.g. via a custom `MessageComposerController` plug-in), confirm the fallback `"Remove photo"` / `"Remove video"` / `"Remove file"` is announced.

## Release Notes

* **Improvements**
  * Enhanced accessibility descriptions for removing attachments with type-specific labels (photo, video, file, voice message).
  * Updated attachment removal labels to include the attachment name when available.
  * Improved localization support across multiple languages with context-aware messaging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6433)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/res/values-ja/strings.xml
```diff
@@ -237,7 +237,13 @@
...
   <string name="stream_compose_recent_files">"最近のファイル"</string>
   <string name="stream_compose_remaining_media_attachments_count">"+%1$d"</string>
-  <string name="stream_compose_remove_attachment">"添付ファイルを削除"</string>
+  <string name="stream_compose_remove_attachment_photo">"写真を削除"</string>
+  <string name="stream_compose_remove_attachment_photo_named">"写真 %1$s を削除"</string>
+  <string name="stream_compose_remove_attachment_video">"動画を削除"</string>
+  <string name="stream_compose_remove_attachment_video_named">"動画 %1$s を削除"</string>
+  <string name="stream_compose_remove_attachment_file">"ファイルを削除"</string>
+  <string name="stream_compose_remove_attachment_file_named">"ファイル %1$s を削除"</string>
+  <string name="stream_compose_remove_attachment_voice_message">"音声メッセージを削除"</string>
```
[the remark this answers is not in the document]
andremion: that key doesn't exist anymore

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/attachments/MessageComposerAttachmentMediaItem.kt
```diff
@@ -88,11 +88,30 @@ internal fun MessageComposerAttachmentMediaItem(
...
             onClick = { onAttachmentRemoved(attachment) },
-            contentDescription = stringResource(R.string.stream_compose_remove_attachment),
+            contentDescription = mediaAttachmentRemoveDescription(attachment),
         )
     }
 }
 
+@Composable
+private fun mediaAttachmentRemoveDescription(attachment: Attachment): String {
+    val displayName = attachment.title ?: attachment.name
```
VelikovPetar, line 98: Question: What if the attachment has an arbitrary name? For example, the captured photo attachment, has a name consisting of its timestamp - so I just want to check if we are OK reading out such text, which might not make too much sense for the users:

https://github.com/user-attachments/assets/aae209f9-42cb-4e6a-8c3e-c340e07daa4a

andremion: Screenshot 2026-05-12 at 09 11 01

That is the designer specs.
What do you propose?

VelikovPetar: To be honest this is a bit difficult thing to handle. If design is OK with this, I would say to keep it. I don't see a point in checking what the file name looks like and whether it is 'understandable', there is absolutely no way to cover everything. 

Resolving this.

VelikovPetar (approved):

PR #6435, opened 2026-05-11
Add configurable character limits and feature toggles for polls

ryanhurststrava:
### Goal
Introduce comprehensive configuration for poll creation features, allowing developers to control which features are available to users, set default values, and enforce character limits on questions and options. This provides flexibility for different use cases where certain poll features may need to be restricted, pre-configured, or character-limited based on UI constraints, business rules, or localization requirements. This loosely follows the [iOS implementation.](https://getstream.io/chat/docs/sdk/ios/swiftui/polls/)

### New Files Created
**`PollsConfig.kt`**
- New `PollsConfig` data class for configuring all poll creation features
- New `PollsEntryConfig` data class for individual feature configuration (configurable flag + default value)
- Configurable features:
  - `multipleVotes` - Allow users to select multiple options
  - `anonymousPoll` - Hide voter identities
  - `suggestAnOption` - Allow users to add their own options
  - `addComments` - Allow users to add comments to polls
- Character limits:
  - `questionTextLimit` - Optional character limit for poll questions
  - `optionTextLimit` - Optional character limit for poll options
- Both classes implement `Parcelable` for passing via Bundle arguments
- Includes `Default` and `NotConfigurable` presets

### Modified Files
**`ChatUI.kt`**
- Added `pollsConfig` property to provide global default configuration
- Defaults to `PollsConfig.Default` (all features configurable and disabled by default)

**`CreatePollDialogFragment.kt`**
- Updated `newInstance()` to accept optional `PollsConfig` parameter
- Falls back to `ChatUI.pollsConfig` when not provided
- New `configurePollFeatures()` method to show/hide UI elements based on config
- Applies character limits using `InputFilter.LengthFilter`
- Wires up "add a comment" feature that was previously in UI but not functional
- Sets initial switch states based on config default values

**`CreatePollViewModel.kt`**
- Added `allowAnswers` field and `setAllowAnswers()` method
- Passes `allowAnswers` to the poll configuration when creating polls

**`OptionsAdapter.kt`**
- Added `optionTextLimit` parameter to constructor
- Applies `InputFilter.LengthFilter` to option EditText fields when limit is specified
- Maintains backward compatibility with overloaded constructor

### Key Design Decisions
- **Backward Compatibility**: All features remain configurable by default; existing code continues to work without changes
- **Two-Level Configuration**: Features can be controlled via global `ChatUI.pollsConfig` or per-dialog via `newInstance()`
- **Visibility Control**: When a feature is not configurable, its UI elements are completely hidden
- **Default Values**: Features can be pre-configured with default on/off states
- **Optional Limits**: Character limits default to `null` (unlimited) unless explicitly specified

### Usage Examples
```kotlin
// Global configuration
ChatUI.pollsConfig = PollsConfig(
    multipleVotes = PollsEntryConfig(configurable = false, defaultValue = false),
    anonymousPoll = PollsEntryConfig(configurable = true, defaultValue = true),
    questionTextLimit = 200,
    optionTextLimit = 100
)

// Per-dialog configuration
CreatePollDialogFragment.newInstance(
    createPollDialogListener = myListener,
    pollsConfig = PollsConfig(
        suggestAnOption = PollsEntryConfig.NotConfigurable,
        addComments = PollsEntryConfig(configurable = true, defaultValue = false),
        questionTextLimit = 150,
        optionTextLimit = 75
    )
)
```

## 📱 UI Changes
- Poll features can now be hidden when not configurable
- Features can be pre-enabled with default values
- Character limits are enforced at the input level using InputFilter
- "Add a comment" switch is now functional

### Manual Testing Steps
1. Test default behavior - all features configurable, none enabled by default
2. Hide features - set `configurable = false`, verify UI elements are hidden
3. Set default values - verify switches initialize to correct state
4. Test character limits on questions and options
5. Verify "add a comment" feature now works correctly
6. Test global `ChatUI.pollsConfig` vs per-dialog config
7. Verify backward compatibility with existing `newInstance()` calls

### Expected Behavior
- Non-configurable features are hidden from the UI
- Configurable features display with correct default switch states
- Character limits prevent input beyond specified length
- Configuration can be set globally or per-dialog
- Existing code without config works unchanged

---

### Contributor Checklist
- [ ] I have assigned a reviewer (code owner)
- [x] This PR targets the `develop` branch
- [ ] I have added unit tests for new code
- [x] I have updated documentation (KDocs included in code)

### Reviewer Checklist
- [ ] UI Components sample app has been tested
- [ ] Compose sample app has been tested (N/A - UI Components only)
- [ ] Visuals are correct
- [ ] Feature is validated
- [ ] KDocs have been reviewed
- [ ] SDK size impact has been reviewed from CI logs

---

## 🎉 GIF
https://github.com/user-attachments/assets/f09ac394-c58c-4489-bfca-c2c2447f6690

[Screen_recording_20260511_173314.webm](https://github.com/user-attachments/assets/de6d4adc-8ca8-464b-bcb1-3d57908666ba)

## Release Notes

* **New Features**
  * Introduced comprehensive poll configuration system enabling control over which poll creation features are available to users
  * Added configurable text limits for poll questions and options
  * Made poll settings (multiple votes, anonymous polls, suggest-an-option, add comments) individually configurable with customizable default values
  * Added option to control whether poll responses can include answers

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6435)

### DIFF
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
@@ -17,6 +17,7 @@
 package io.getstream.chat.android.ui.feature.messages.composer.attachment.picker.poll
 
 import android.os.Bundle
+import android.text.InputFilter
 import android.view.LayoutInflater
 import android.view.MenuItem
 import android.view.View
@@ -30,6 +31,7 @@ import androidx.core.widget.addTextChangedListener
 import androidx.fragment.app.viewModels
 import androidx.lifecycle.lifecycleScope
 import io.getstream.chat.android.models.PollConfig
+import io.getstream.chat.android.ui.ChatUI
 import io.getstream.chat.android.ui.R
 import io.getstream.chat.android.ui.common.utils.PollsConstants
 import io.getstream.chat.android.ui.databinding.StreamUiFragmentCreatePollBinding
@@ -40,6 +42,8 @@ import kotlinx.coroutines.launch
 
 /**
  * Represent the bottom sheet dialog that allows users to pick attachments.
+ *
+ * Use [newInstance] to create an instance with optional [PollsConfig].
  */
 public class CreatePollDialogFragment : AppCompatDialogFragment() {
 
@@ -47,8 +51,19 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
     private val binding get() = _binding!!
     private var createPollDialogListener: CreatePollDialogListener? = null
     private val createPollViewModel: CreatePollViewModel by viewModels()
+    private val pollsConfig: PollsConfig by lazy {
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.TIRAMISU) {
+            arguments?.getParcelable(ARG_POLLS_CONFIG, PollsConfig::class.java)
+        } else {
+            @Suppress("DEPRECATION")
+            arguments?.getParcelable(ARG_POLLS_CONFIG)
+        } ?: ChatUI.pollsConfig
+    }
     private val optionsAdapter: OptionsAdapter by lazy {
-        OptionsAdapter { id, text -> createPollViewModel.onOptionTextChanged(id, text) }
+        OptionsAdapter(
+            optionTextLimit = pollsConfig.optionTextLimit,
+            onOptionChange = { id, text -> createPollViewModel.onOptionTextChanged(id, text) },
+        )
     }
     private lateinit var sendMenuItem: MenuItem
 
@@ -77,10 +92,55 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
     }
 
     /**
+     * Configures the visibility and default values of poll features based on [pollsConfig].
+     */
+    private fun configurePollFeatures() {
+        // Configure multiple votes feature
+        createPollViewModel.setAllowMultipleVotes(pollsConfig.multipleVotes.defaultValue)
+        binding.multipleAnswersLabel.isVisible = pollsConfig.multipleVotes.configurable
+        binding.multipleAnswersSwitch.isVisible = pollsConfig.multipleVotes.configurable
+        if (pollsConfig.multipleVotes.configurable) {
+            binding.multipleAnswersSwitch.isChecked = pollsConfig.multipleVotes.defaultValue
+            binding.multipleAnswersCount.isVisible = pollsConfig.multipleVotes.defaultValue
+        }
+
+        // Configure anonymous poll feature
+        createPollViewModel.setAnnonymousPoll(pollsConfig.anonymousPoll.defaultValue)
+        binding.anonymousPollLabel.isVisible = pollsConfig.anonymousPoll.configurable
+        binding.anonymousPollSwitch.isVisible = pollsConfig.anonymousPoll.configurable
+        if (pollsConfig.anonymousPoll.configurable) {
+            binding.anonymousPollSwitch.isChecked = pollsConfig.anonymousPoll.defaultValue
+        }
+
+        // Configure suggest an option feature
+        createPollViewModel.setSuggestAnOption(pollsConfig.suggestAnOption.defaultValue)
+        binding.suggestAnOptionLabel.isVisible = pollsConfig.suggestAnOption.configurable
+        binding.suggestAnOptionSwitch.isVisible = pollsConfig.suggestAnOption.configurable
+        if (pollsConfig.suggestAnOption.configurable) {
+            binding.suggestAnOptionSwitch.isChecked = pollsConfig.suggestAnOption.defaultValue
+        }
+
+        // Configure add a comment feature
+        createPollViewModel.setAllowAnswers(pollsConfig.addComments.defaultValue)
+        binding.addACommentLabel.isVisible = pollsConfig.addComments.configurable
+        binding.addACommentLabelSwitch.isVisible = pollsConfig.addComments.configurable
+        if (pollsConfig.addComments.configurable) {
+            binding.addACommentLabelSwitch.isChecked = pollsConfig.addComments.defaultValue
+        }
+    }
+
+    /**
      * Initializes the dialog.
      */
     private fun setupDialog() {
         setupToolbar(binding.toolbar)
+        pollsConfig.questionTextLimit?.takeIf { it > 0 }?.let { limit ->
+            binding.question.filters = arrayOf(InputFilter.LengthFilter(limit))
+        }
+
+        // Configure poll feature visibility and default values based on pollsConfig
+        configurePollFeatures()
+
         binding.multipleAnswersSwitch.setOnCheckedChangeListener { _, isChecked ->
             binding.multipleAnswersCount.isVisible = isChecked
             createPollViewModel.setAllowMultipleVotes(isChecked)
@@ -96,6 +156,9 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
         }
         binding.suggestAnOptionSwitch.setOnCheckedChangeListener { _, isChecked ->
             createPollViewModel.setSuggestAnOption(isChecked)
+        }
+        binding.addACommentLabelSwitch.setOnCheckedChangeListener { _, isChecked ->
+            createPollViewModel.setAllowAnswers(isChecked)
         }
         binding.addOption.setOnClickListener {
             createPollViewModel.createOption()
@@ -158,15 +221,25 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
 
     public companion object {
         public const val TAG: String = "create_poll_dialog_fragment"
+        private const val ARG_POLLS_CONFIG: String = "arg_polls_config"
 
         /**
          * Creates a new instance of [CreatePollDialogFragment].
          *
+         * @param createPollDialogListener The listener for poll creation events.
+         * @param pollsConfig Optional configuration for poll features. Defaults to [ChatUI.pollsConfig].
          * @return A new instance of [CreatePollDialogFragment].
          */
-        public fun newInstance(createPollDialogListener: CreatePollDialogListener): CreatePollDialogFragment {
-            return CreatePollDialogFragment()
-                .setCreatePollDialogListener(createPollDialogListener)
+        @JvmOverloads
+        public fun newInstance(
+            createPollDialogListener: CreatePollDialogListener,
+            pollsConfig: PollsConfig? = null,
+        ): CreatePollDialogFragment {
+            return CreatePollDialogFragment().apply {
+                arguments = Bundle().apply {
+                    pollsConfig?.let { config -> putParcelable(ARG_POLLS_CONFIG, config as android.os.Parcelable) }
+                }
+            }.setCreatePollDialogListener(createPollDialogListener)
         }
     }
 
--- a/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
+++ b/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
@@ -25,6 +25,7 @@ import android.view.ViewGroup
 import androidx.appcompat.app.AppCompatDialogFragment
 import androidx.appcompat.widget.Toolbar
 import androidx.core.content.ContextCompat
+import androidx.core.os.BundleCompat
 import androidx.core.view.WindowInsetsCompat
 import androidx.core.view.isVisible
 import androidx.core.widget.addTextChangedListener
@@ -39,6 +40,7 @@ import io.getstream.chat.android.ui.utils.extensions.applyEdgeToEdgePadding
 import io.getstream.chat.android.ui.utils.extensions.streamThemeInflater
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.launch
+import kotlin.jvm.java
 
 /**
  * Represent the bottom sheet dialog that allows users to pick attachments.
@@ -52,11 +54,8 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
     private var createPollDialogListener: CreatePollDialogListener? = null
     private val createPollViewModel: CreatePollViewModel by viewModels()
     private val pollsConfig: PollsConfig by lazy {
-        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.TIRAMISU) {
-            arguments?.getParcelable(ARG_POLLS_CONFIG, PollsConfig::class.java)
-        } else {
-            @Suppress("DEPRECATION")
-            arguments?.getParcelable(ARG_POLLS_CONFIG)
+        arguments?.let {
+            BundleCompat.getParcelable(it, ARG_POLLS_CONFIG, PollsConfig::class.java)
         } ?: ChatUI.pollsConfig
     }
     private val optionsAdapter: OptionsAdapter by lazy {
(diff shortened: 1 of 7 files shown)


stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
```diff
@@ -76,11 +91,56 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
         setupDialog()
     }
 
+    /**
+     * Configures the visibility and default values of poll features based on [pollsConfig].
+     */
+    private fun configurePollFeatures() {
```
gpunto, line 101: This function updates the state of some views, i.e. when setting `switch.isChecked = defaultValue`, which we should only do when the fragment view is created the first time (`savedInstanceState == null`). Otherwise, we are going to overwrite user selections on configuration changes.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/PollsConfig.kt
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @param multipleVotes Configuration for allowing multiple votes in a poll.
+ * @param anonymousPoll Configuration for enabling anonymous polls.
+ * @param suggestAnOption Configuration for allowing users to suggest options in a poll.
+ * @param addComments Configuration for adding comments to a poll.
+ * @param questionTextLimit Optional character limit for the poll question. Null means no limit.
+ * @param optionTextLimit Optional character limit for poll answer options. Null means no limit.
+ */
+@Parcelize
+public data class PollsConfig(
```
gpunto, line 34: We have a similar config class [in the Compose SDK for v7](https://github.com/GetStream/stream-chat-android/blob/1965ffedcdfe55eeeadf3ca1bf3eee05d65b8d18/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatUiConfig.kt#L76) (on the `develop` branch), could you align the naming to be consistent with that one?

Basically, my suggestion would be:
- `PollsEntryConfig` -> `PollFeatureConfig`
- `addComments` -> `allowComments`

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
```diff
@@ -158,15 +221,25 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
...
-            return CreatePollDialogFragment()
-                .setCreatePollDialogListener(createPollDialogListener)
+        @JvmOverloads
+        public fun newInstance(
+            createPollDialogListener: CreatePollDialogListener,
+            pollsConfig: PollsConfig? = null,
+        ): CreatePollDialogFragment {
+            return CreatePollDialogFragment().apply {
+                arguments = Bundle().apply {
+                    pollsConfig?.let { config -> putParcelable(ARG_POLLS_CONFIG, config as android.os.Parcelable) }
```
gpunto: Nit: the cast shouldn't be needed
```suggestion
                    pollsConfig?.let { config -> putParcelable(ARG_POLLS_CONFIG, config) }
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
```diff
@@ -40,15 +42,28 @@ import kotlinx.coroutines.launch
...
     private val binding get() = _binding!!
     private var createPollDialogListener: CreatePollDialogListener? = null
     private val createPollViewModel: CreatePollViewModel by viewModels()
+    private val pollsConfig: PollsConfig by lazy {
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.TIRAMISU) {
+            arguments?.getParcelable(ARG_POLLS_CONFIG, PollsConfig::class.java)
+        } else {
+            @Suppress("DEPRECATION")
+            arguments?.getParcelable(ARG_POLLS_CONFIG)
+        } ?: ChatUI.pollsConfig
```
gpunto: Nit: we could use BundleCompat to do this for us:
```suggestion
        arguments?.let { BundleCompat.getParcelable(it, ARG_POLLS_CONFIG, PollsConfig::class.java) }
            ?: ChatUI.pollsConfig
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
```diff
@@ -76,11 +91,56 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
...
 
+    /**
+     * Configures the visibility and default values of poll features based on [pollsConfig].
+     */
+    private fun configurePollFeatures() {
+        // Configure multiple votes feature
+        createPollViewModel.setAllowMultipleVotes(pollsConfig.multipleVotes.defaultValue)
+        binding.multipleAnswersLabel.isVisible = pollsConfig.multipleVotes.configurable
+        binding.multipleAnswersSwitch.isVisible = pollsConfig.multipleVotes.configurable
+        if (pollsConfig.multipleVotes.configurable) {
```
gpunto, line 106: I think we're introducing a new edge case here.

If we have `multipleVotes.defaultValue = true` + `multipleVotes.configurable = false`, then we end up in a state where:
- In the VM, `allowMultipleVotes = true` since we set it here
- `multipleAnswersCount` is hidden because of `configurable = false`
- `sendMenuItem` is disabled because of the above ([due to this logic](https://github.com/GetStream/stream-chat-android/blob/v6/stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollViewModel.kt#L125)) so the user can never create the poll

In Compose v7 we changed the behavior to allow `multipleVotes = true + maxVotes = null` which means unlimited votes, so maybe we could do something like that? It's a behavior change, though. 

Another option is to disallow the problematic combo, e.g. we could add a `check(multipleVotes.configurable == true || multipleVotes.defaultValue == false)` call in `PollsConfig`'s constructor.

cc @andremion @VelikovPetar wdyt?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/OptionsAdapter.kt
```diff
@@ -27,9 +28,17 @@ import io.getstream.chat.android.ui.databinding.StreamUiPollOptionBinding
...
+    private val optionTextLimit: Int?,
     private val onOptionChange: (id: Int, text: String) -> Unit,
 ) : ListAdapter<PollAnswer, OptionsAdapter.OptionViewHolder>(OptionDiffCallback) {
 
+    /**
+     * Builds an [OptionsAdapter] instance without providing option text limit.
+     *
+     * @param onOptionChange Callback invoked when the option text changes.
+     */
+    public constructor(onOptionChange: (id: Int, text: String) -> Unit) : this(null, onOptionChange)
```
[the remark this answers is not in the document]
gpunto: Doesn't the secondary constructor compile to the same jvm signature as the original primary constructor?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
VelikovPetar: I am more in favour of disallowing the problematic combination - I wouldn't want to change the default behaviour in a minor release. But once we port this addition to V7, I would say to align it with the compose implementation.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/PollsConfig.kt
ryanhurststrava: Oh I see there was similar functionality added to v7. I'll see if I can get upgraded to v7. 

Also, I'll create a new pr for character limits in v7.

gpunto: Have a look, but keep in mind that the changes I mentioned were for Compose only in v7, so we'd still have to port basically what you're doing in this PR over there.

My suggestion for unblocking you is to move forward with this very PR to the v6 branch so you get the functionality you need. Then the same feature can be ported to v7 and you can bump to it at your own pace, given that the v7 major includes breaking changes too.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
```diff
@@ -97,6 +157,9 @@ public class CreatePollDialogFragment : AppCompatDialogFragment() {
         binding.suggestAnOptionSwitch.setOnCheckedChangeListener { _, isChecked ->
             createPollViewModel.setSuggestAnOption(isChecked)
         }
+        binding.addACommentLabelSwitch.setOnCheckedChangeListener { _, isChecked ->
+            createPollViewModel.setAllowAnswers(isChecked)
+        }
```
ryanhurststrava, line 163: I added this, but it is more or less unrelated to this pr. I think the current functionality(before this pr) is that this switch doesn't do anything. I wasn't able to get this to work in my testing with the changes in this pr either though, so I'm curious what your thoughts are.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
ryanhurststrava: Do you have a good way to test this? when running the demo app without these changes, user input is reset so I don't have a good baseline for this. I've made some changes so that configurePollFeatures is only called when savedInstanceState is null, but I don't have a great way to validate it.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
ryanhurststrava: Can you be more explicit at what you'd like to see here? 

another option would be to remove the multiple votes configuration from this pr. I included it for completeness, but it is not a feature my team needs.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
gpunto: I think you discovered a bug. It seems we're creating an additional fragment on rotation, which is incorrect. We'll fix this on our side, your change looks good!

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
gpunto: I'd also leave it included for completeness. Basically the suggestion is to put this inside `PollsConfig` constructor:

```kotlin
init {
    require(multipleVotes.configurable || !multipleVotes.defaultValue) {
        "Invalid PollsConfig: multipleVotes cannot have defaultValue=true while " +
            "configurable=false as the user would be unable to set maxVotesAllowed."
    }
}
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
gpunto: It definitely looks like we had a bug both on the creation side and the consuming side, since we never read "allow answers" when displaying poll messages. We'll check that on our side, thanks for fixing the problem at creation time!

gpunto:
We have some failing static checks:
- Detekt complaining about a couple of long lines -> you can see the warnings with `./gradlew detekt`
- Api check -> you can regenerate the api file with `./gradlew apiDump`
- Spotless -> you can reformat the code with `./gradlew spotlessApply`

Could you fix them? 🙏

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/PollsConfig.kt
```diff
@@ -0,0 +1,48 @@
...
+    val multipleVotes: PollFeatureConfig = PollFeatureConfig.Default,
+    val anonymousPoll: PollFeatureConfig = PollFeatureConfig.Default,
+    val suggestAnOption: PollFeatureConfig = PollFeatureConfig.Default,
+    val allowComments: PollFeatureConfig = PollFeatureConfig.Default,
+    val questionTextLimit: Int? = null,
+    val optionTextLimit: Int? = null,
+) : Parcelable {
+    public companion object {
```
gpunto, line 50: Based on the [thread](https://github.com/GetStream/stream-chat-android/pull/6435#discussion_r3247232569)
```suggestion
) : Parcelable {

   init {
        require(multipleVotes.configurable || !multipleVotes.defaultValue) {
            "Invalid PollsConfig: multipleVotes cannot have defaultValue=true while " +
                "configurable=false as the user would be unable to set maxVotesAllowed."
        }
    }

    public companion object {
```

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/CreatePollDialogFragment.kt
gpunto: @ryanhurststrava I've added a [code suggestion](https://github.com/GetStream/stream-chat-android/pull/6435#discussion_r3247229223) with this, let me know if it looks ok to you (maybe you'll have to re-run `spotlessApply` after adding that code). I believe that's the only missing piece and we can merge :)

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/messages/composer/attachment/picker/poll/PollsConfig.kt
```diff
@@ -0,0 +1,56 @@
...
+public data class PollsConfig(
+    val multipleVotes: PollFeatureConfig = PollFeatureConfig.Default,
+    val anonymousPoll: PollFeatureConfig = PollFeatureConfig.Default,
+    val suggestAnOption: PollFeatureConfig = PollFeatureConfig.Default,
+    val allowComments: PollFeatureConfig = PollFeatureConfig.Default,
+    val questionTextLimit: Int? = null,
+    val optionTextLimit: Int? = null,
+) : Parcelable {
+
+   init {
```
gpunto:
```suggestion
    init {
```

gpunto (approved):

PR #6441, opened 2026-05-18
Improve message composer TalkBack accessibility: action verbs, banner merge, audio recording lifecycle

andremion:
AND-1180

### Goal

The Compose message composer was underserved on TalkBack: action buttons announced as generic "double-tap to activate" with no verb, the Edit / Reply banner read fragment-by-fragment, the audio recording flow had no lifecycle announcements, and TalkBack focus drifted out of the composer on state transitions. This PR brings the composer up to a usable TalkBack baseline.

### Implementation

Three feature-scoped commits.

**Action verbs on Send / Save / attach buttons.**
`FilledIconButton` doesn't expose an `onClickLabel`, so TalkBack defaulted to "double-tap to activate" with no verb. Layered `Modifier.semantics { onClick(label = …) { …; true } }` on each button so the verb reads as `Send message`, `Save changes`, `Open attachments` / `Close attachments` (the attachments toggle pairs the label with the existing `stateDescription`). Extracted `AttachmentPickerButton` from `MessageComposerLeadingContent` to keep the parent under detekt's method-length cap.

**Edit / Reply banner merge for a11y.**
Wrapped the edit-indicator row and reply quoted-message in `Modifier.semantics(mergeDescendants = true) {}` so TalkBack reads each banner as a single chunk. Cancel icons announce the action they undo (`Cancel editing` / `Cancel reply`) instead of a generic "Cancel". When replying to your own message, the username `Text` leaf overrides `contentDescription` to "Reply to you" so the post-preposition pronoun form is grammatically correct in es / it / fr (the visible "You" label stays compact).

**Audio recording lifecycle a11y.**
- Mic button: `contentDescription` includes the press-and-hold gesture hint, exposes `Role.Button` — TalkBack reads `"Record audio message. Press and hold to record. Button."`.
- One-shot announcements via `View.announceForAccessibility` on recording state transitions: `Idle → Hold` announces "Recording started. Slide left to cancel, up to lock."; `non-Idle → Idle` announces "Recording cancelled" when gated by a `cancelRequested` flag set by wrapped `onCancelRecording` / `onDeleteRecording` (confirm / send paths stay silent and let the attachment item announce itself).
- Focus moves on transitions into the recording UI: `Hold → Locked` lands on the merged recording-bar row, `Locked → Overview` lands on the merged playback row. Overview's outer Row is split into a separately-focusable play button + an inner focusable summary so the play action stays reachable.
- `Confirm` button label is state-aware: "Send recording" when `audioRecordingSendOnComplete` is enabled, "Confirm recording" otherwise.
- A 100 ms delay before each `requestFocus` lets Compose finish layout and the accessibility tree settle before we move focus, so it wins over TalkBack's fallback recovery.

All new strings translated across the 7 supported locales (es / fr / hi / in / it / ja / ko). No public API changes.

### Testing

Enable TalkBack on a physical device or emulator. For each scenario, expected announcement is in **bold**.

**1. Composer trailing-button verbs**
1. Open a channel, focus the trailing button while the input is empty → **"Record audio message. Press and hold to record. Button."**
2. Type any character, focus the trailing button → **"Send message"** verb (instead of "Double-tap to activate").
3. Trigger an Edit action on one of your messages, focus the trailing button → **"Save changes"** verb.

**2. Composer leading-button label**
1. Focus the leading attachments button → **"Open attachments. Collapsed."**
2. Tap to open the picker, focus again → **"Close attachments. Expanded."**

**3. Edit / Reply banner**
1. Long-press your own message → Edit. Focus the banner → reads as one chunk including the original text.
2. Focus the cancel icon → **"Cancel editing, button"**.
3. Long-press someone else's message → Reply. Focus the banner → reads "Reply to {name}, …" as one chunk.
4. Repeat reply on your own message → **"Reply to you, …"** (not just "You").
5. Focus the cancel icon → **"Cancel reply, button"**.

**4. Audio recording lifecycle**
1. Focus the mic button → **"Record audio message. Press and hold to record. Button."**
2. Double-tap and hold to start recording. As the Hold state activates → **"Recording started. Slide left to cancel, up to lock."**
3. Slide up to lock → focus moves to the recording-bar row, reads **"Audio recording in progress"**.
4. Tap the Stop button → focus moves to the playback row, reads **"Audio recording"**. Play / Pause stays separately focusable.
5. While in Locked or Overview, tap the Delete button → **"Recording cancelled"** announces.
6. While in Hold, slide left to cancel → **"Recording cancelled"** announces and focus stays on the mic button.
7. Toggle the `audioRecordingSendOnComplete` flag (`ChatTheme.config.composer`). In Overview, focus the Confirm button → reads **"Send recording"** when the flag is true, **"Confirm recording"** when false.

## Release Notes

* **New Features**
  * Enhanced accessibility for audio recording with automatic state announcements
  * Improved accessibility labels and descriptions for message composer controls
  * Better accessibility support for quoted message interactions

* **Localization**
  * Updated user-facing strings across Spanish, French, Hindi, Indonesian, Italian, Japanese, and Korean languages with new labels for audio recording and message composition actions

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6441?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingAccessibility.kt
```diff
@@ -0,0 +1,97 @@
...
+            is RecordingState.Idle -> if (!wasIdle && cancelRequested) {
+                view.announceForAccessibility(cancelledAnnouncement)
+            }
+            is RecordingState.Locked,
+            is RecordingState.Overview,
+            is RecordingState.Complete,
+            -> Unit
+        }
+        onTransitionConsumed()
+    }
```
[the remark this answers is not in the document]
andremion: da0224fb8492e0f3f1b769a3718f7ec3d5e6d16c

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingContent.kt
```diff
@@ -174,8 +180,20 @@ internal fun MessageComposerAudioRecordingLockedContent(
     recordingActions: AudioRecordingActions,
     modifier: Modifier = Modifier,
 ) {
+    val rowFocusRequester = remember { FocusRequester() }
+    LaunchedEffect(Unit) {
+        delay(FocusRequestDelayMs)
```
VelikovPetar, line 185: Just a small confirmation: I guess it is perfectly safe to apply this focus request, even if Talkback is off?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/AudioRecordAttachmentContent.kt
```diff
@@ -283,6 +284,13 @@ internal fun PlaybackToggleButton(
         } else {
             painterResource(id = R.drawable.stream_design_ic_play_fill)
         }
+    val label = stringResource(
```
VelikovPetar, line 287: Should we maybe move this to the new `AudioRecordingAccessibility` utils class? (there are also several similar `if-else` decision blocks for accessibility labels, I am thinking if should group them all together)

VelikovPetar (approved):
Left two minor questions, but I don't see them as blockers!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/AudioRecordingContent.kt
andremion: Yes. It's safe with TalkBack off. I manually tested. Feel free to double-check it, too.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/AudioRecordAttachmentContent.kt
andremion: I'd prefer to keep it where it is for now. `AudioRecordingAccessibility` is scoped to the recording lifecycle (mic button + transition announcements in the composer). `PlaybackToggleButton` runs for any audio attachment playback, including received audio messages, so moving its label there would couple attachment playback to composer-internal recording utils, and the file name would no longer fit. If similar decisions grow, a neutral `AudioAttachmentAccessibility` under `attachments.content` would be a better home.

VelikovPetar: OK fine with me!

PR #6445, opened 2026-05-19
Improve poll creation TalkBack accessibility: action verb, toggle rows, error live region, reorder actions

andremion:
AND-1180

### Goal

Bring the Compose poll creation screen to a usable TalkBack baseline.

### Implementation

- **Create button verb.** `StreamButton` accepts an `onClickLabel`; the poll header passes "create poll". TalkBack reads the verb instead of the generic hint.
- **Switch rows as a single cell.** *Multiple votes* and *Limit votes per person* rows are `toggleable` with merged semantics; inner `StreamSwitch` is non-interactive and the row owns the 48 dp touch target via `defaultMinSize`. TalkBack reads label + state.
- **Option error live region.** `PollOptionErrorRow` uses `liveRegion = Polite` so the duplicate-title error is announced as it mounts.
- **Reorder via custom actions.** Drag handle exposes boundary-aware *Move up* / *Move down* `CustomAccessibilityAction`s. Drag-settle and a11y action share one move lambda.
- **Option identity in reorder / remove.** Reorder / Remove icons announce *"Reorder option {name}"* / *"Remove option {name}"*, falling back to *"option N"* when the row is blank.

New strings: `stream_compose_poll_create_action`, `stream_compose_poll_option_move_up`, `stream_compose_poll_option_move_down`. `stream_compose_poll_option_reorder` and `_remove` take the option name as `%s`. Translated in 7 locales.

Public API: additive `onClickLabel: String? = null` on `StreamButton` (internal).

### Testing

Enable TalkBack on a physical device. Open the Compose sample, start a new channel, tap the attachments picker, select **Polls**.

1. **Create button verb** — Focus Create → **"Create, double-tap to create"**.
2. **Switch rows** — Focus *Multiple votes* / *Limit votes per person* → one cell with **"off"** / **"on"**; double-tap toggles.
3. **Error live region** — Type the same title in two options → TalkBack announces the duplicate error as the row mounts.
4. **Reorder actions** — With 3+ options, open the drag handle's local context menu (single-finger swipe down → right). First row offers only *Move down*; last row only *Move up*; middle rows both. Activate → list reorders.
5. **Option identity** — Filled-in row's drag handle / remove icon say **"Reorder option Pizza"** / **"Remove option Pizza"**; blank row says **"Reorder option 2"** / **"Remove option 2"**.

## Summary by CodeRabbit

* **New Features**
  * Poll option reordering with move up/down actions and per-row move controls.

* **Accessibility**
  * Added labeled accessibility actions for poll move operations and polite live-region updates.
  * Buttons now include optional accessibility click labels for screen readers.

* **Localization**
  * Added/updated poll-related translations (create, move up/down, remove/reorder with placeholders) across multiple languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6445?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andremion:
@coderabbit review

andremion:
@coderabbit full review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/poll/PollOptionList.kt
```diff
@@ -141,16 +158,20 @@ private fun ReorderableScope.PollOptionRow(
...
+                    }
                     .draggableHandle()
                     .minimumInteractiveComponentSize(),
             ) {
                 Icon(
                     painter = painterResource(id = R.drawable.stream_design_ic_reorder),
-                    contentDescription = stringResource(R.string.stream_compose_poll_option_reorder),
+                    contentDescription = null,
                     tint = colors.inputTextIcon,
-                    modifier = Modifier.size(20.dp),
```
gpunto, line 153: I'd still specify the expected size here, so we guard against changes of the size declared in the drawable

gpunto (approved):
Looks good, just a suggestion

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/attachments/poll/PollOptionList.kt
andremion: If the drawable's intrinsic size matches the design spec, I'd avoid hardcoding it so integrators can override the drawable and have the icon size update automatically. I'd hardcode only when the intrinsic size doesn't match the spec.

PR #6447, opened 2026-05-19
Let integrators supply channel header click labels for TalkBack

andremion:
### Goal

Give TalkBack a concrete verb when the user focuses the channel header trailing avatar or the title section, instead of the generic "double-tap to activate" hint. Because the click actions are integrator-defined, the SDK exposes the labels as paired parameters and ships no default label of its own — the action's owner also owns the label.

### Implementation

**SDK plumbing.** Every layer that already exposes `onChannelAvatarClick` / `onHeaderTitleClick` now exposes a paired `onChannelAvatarClickLabel` / `onHeaderTitleClickLabel: String? = null`:

- `ChannelScreen` (public): new defaulted params, threaded into `DefaultTopBarContent`.
- `ChannelHeaderParams` (factory): new defaulted fields `onHeaderTitleClickLabel`, `onChannelAvatarClickLabel`.
- `ChannelHeader` (public): new defaulted params, threaded into the trailing- and center-content params.
- `ChannelHeaderCenterContentParams.onClickLabel`, `ChannelHeaderTrailingContentParams.onClickLabel`: new defaulted fields (additive — non-breaking for Compose data-class consumers).
- `DefaultChannelHeaderTrailingContent` and `DefaultChannelHeaderCenterContent`: pass the supplied label straight to `Modifier.clickable(onClickLabel = …, role = Role.Button)`. When the integrator doesn't supply one, `null` flows through and TalkBack falls back to its generic verb hint.

**No SDK default label.** The SDK doesn't know what the integrator's callback actually does (open settings? show a sheet? something else?), so it must not hard-code a label like "Open conversation info". The integrator owns both the action and its announced verb.

**Sample app demonstrates the integrator side.** `ChannelActivity` (the renamed sample messages screen) passes `onChannelAvatarClickLabel = stringResource(R.string.messages_open_channel_info)` alongside `::openChannelInfo`, with the label defined in the sample's own `strings.xml`.

**Sample refactor.** While in this file, `MessagesActivity` is renamed to `ChannelActivity` and moved to the `sample/ui/channel` package (alongside `DirectChannelInfoActivity` and `GroupChannelInfoActivity`). All callers (`AndroidManifest`, `ChannelsActivity`, `StartupActivity`, channel info / drafts / reminders / add-channel activities) are updated.

Public API: additive trailing-defaulted params on `ChannelScreen`, `ChannelHeader`, `ChannelHeaderParams`, `ChannelHeaderCenterContentParams`, `ChannelHeaderTrailingContentParams`. API dump regenerated.

Note: the scanner also flagged the avatar's 40 dp touch-target height. Deliberately not addressed here — any layout-level expansion either resizes the avatar (visual change) or shifts the centered title text. To be revisited separately.

### Testing

Enable TalkBack on a physical device, run the sample.

1. Open a 1:1 or group channel.
2. Focus the trailing avatar in the header. Expected: TalkBack announces **"&lt;avatar contents&gt;, button. Double-tap to open channel info"** — the verb comes from the sample's `messages_open_channel_info` string, demonstrating the integrator-supplied label.
3. Focus the channel title section. Expected: no click verb is announced because `MessagesActivity`/`ChannelActivity` doesn't wire `onHeaderTitleClick` in the sample (the SDK doesn't make assumptions on its behalf). TalkBack reads only the title text. If a future sample integration wires `onHeaderTitleClick`, the matching `onHeaderTitleClickLabel` is exposed and ready to use.
4. (For SDK consumers: pass a custom `onChannelAvatarClick = { doSomethingElse() }` without `onChannelAvatarClickLabel`. Confirm TalkBack falls back to the generic "double-tap to activate" hint — the SDK doesn't announce a misleading default.)

## Summary by CodeRabbit

* **New Features**
  * Enhanced accessibility support with semantic labels for header title and channel avatar click actions.

* **Refactor**
  * Restructured navigation flow to use unified channel activity for improved app navigation consistency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6447?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/ChannelHeader.kt
```diff
@@ -263,12 +264,17 @@ internal fun DefaultChannelHeaderTrailingContent(
     currentUser: User?,
     onClick: ((Channel) -> Unit)?,
 ) {
+    val openInfoLabel = stringResource(R.string.stream_compose_channel_header_open_info)
```
gpunto: I just realized something that might apply to other cases too. We are resolving action descriptions close to the click itself but, in theory, we don't know what the `onClick` will do, because it can be overridden by integrators so the talkback could be misleading if they do that 🤔

andremion: Good point!
I will investigate the impact and how to handle it.
Thanks!

andremion: b6b2ed91e4ed73f6188ff66ba50ebccb64fd0330
4ae29a03b08ec4af364f80507cf2d8a14da177e4
I will consider this in the open PRs and open follow-ups if needed.
I also did this small refactoring 00996007d590c25f74345d8baf9a5ce960500019

andremion:
@coderabbit full review

gpunto (approved):

PR #6462, opened 2026-05-21
Accept null file_size in AttachmentDto

### LINKED ISSUE #6428: AttachmentDto / Moshi fails when API returns file_size: null on attachments (watch/query surfaces as NETWORK_FAILED / code 1000)
**Environment**
- Library: io.getstream:stream-chat-android-client 
- Version: 5.17.23 (likely affects other versions with the same DTO)
- Integration: ChannelClient.watch() / channel query responses that include messages with image (or other) attachments

**Problem**
The HTTP API returns 200 and valid JSON, but the Android SDK reports failure with:
- ChatErrorCode.NETWORK_FAILED (stream code 1000)
- Message: Response is failed. See cause
- statusCode: -1
So the failure is client-side deserialization, not a failed HTTP call or empty messages.

**Root cause (observed)**
DownstreamMessageDto / AttachmentDto models file_size as a non-null Int (default 0). When the server JSON includes an attachment with an explicit null:

Image

```
"attachments": [
  {
    "type": "image",
    "image_url": "https://example.com/image.jpeg",
    "file_size": null
  }
]
```
Moshi fails to parse (JsonDataException: expected int, got null). That exception is wrapped as NETWORK_FAILED / 1000, which is misleading when debugging.

**Expected behavior**
The SDK should tolerate file_size: null the same way it tolerates a missing field—e.g. treat null as “unknown” and coerce to 0 or use Int? in AttachmentDto with safe mapping to domain models.

gpunto:
### Goal

It seems that backend can return `null` for `file_size` on attachments, but `AttachmentDto.file_size` was non-nullable so explicit nulls in the JSON caused deserialization to fail.

This also aligns [with iOS](https://github.com/GetStream/stream-chat-swift/blob/8c7d807a5f3f6ccf60ada1262e0e1ef0d50f3ab2/Sources/StreamChat/Models/Attachments/AttachmentTypes.swift#L200-L206), where `AttachmentFile` decodes a missing or non-decodable `file_size` as `0`.

Closes #6428

### Implementation

- `AttachmentDto.file_size` is now `Int?`, so an explicit `null` from the API deserializes cleanly.
- `DomainMapping` maps `null` to `0` when mapping to the domain `Attachment.fileSize`.

### Testing

- New unit test `Deserialize JSON attachment with null file_size` covers the parser path.
- Existing attachment parser and mapping tests continue to pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attachment data handling by making file size processing null-safe. When file size information is unavailable, the system now defaults to zero instead of propagating null values, ensuring more robust behavior.

* **Tests**
  * Expanded test coverage with new test cases validating correct deserialization and mapping of attachments when file size values are null or missing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6462?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/AttachmentDto.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/AttachmentDto.kt
@@ -33,7 +33,7 @@ internal data class AttachmentDto(
     val author_name: String?,
     val author_link: String?,
     val fallback: String?,
-    val file_size: Int = 0,
+    val file_size: Int? = 0,
     val image: String?,
     val image_url: String?,
     val mime_type: String?,
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/AttachmentDto.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/AttachmentDto.kt
@@ -33,7 +33,7 @@ internal data class AttachmentDto(
     val author_name: String?,
     val author_link: String?,
     val fallback: String?,
-    val file_size: Int? = 0,
+    val file_size: Int?,
     val image: String?,
     val image_url: String?,
     val mime_type: String?,
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
@@ -558,7 +558,7 @@ internal class DomainMappingTest {
             authorName = attachmentDto.author_name,
             authorLink = attachmentDto.author_link,
             fallback = attachmentDto.fallback,
-            fileSize = attachmentDto.file_size,
+            fileSize = attachmentDto.file_size ?: 0,
             image = attachmentDto.image,
             imageUrl = attachmentDto.image_url,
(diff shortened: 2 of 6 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
```diff
@@ -558,7 +558,7 @@ internal class DomainMappingTest {
             authorName = attachmentDto.author_name,
             authorLink = attachmentDto.author_link,
             fallback = attachmentDto.fallback,
-            fileSize = attachmentDto.file_size,
+            fileSize = attachmentDto.file_size ?: 0,
```
andremion, line 561: Maybe it's worth it asserting the case when the file_size is null

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/AttachmentDto.kt
```diff
@@ -33,7 +33,7 @@ internal data class AttachmentDto(
     val author_name: String?,
     val author_link: String?,
     val fallback: String?,
-    val file_size: Int = 0,
+    val file_size: Int? = 0,
```
andremion: What about dropping the default value here, to be consistent with the other nullable fields, and letting the domain mapping handle the fallback?

andremion (approved):

PR #6464, opened 2026-05-21
Improve channel header TalkBack accessibility

andremion:
AND-1180

### Goal

The channel header has three accessibility gaps that prevent TalkBack users from interacting with it:

1. The trailing avatar and the title `Column` are clickable with `Role.Button` but expose no readable name. TalkBack treats them as malformed buttons and skips them during swipe navigation.
2. The channel name announces only when focus lands on it. There is no `heading` semantic, so TalkBack heading-navigation can't jump to it.
3. On group channels with overflow members, the avatar's `+N` `CountBadge` and the channel formatter's `+N more` text both announce — TalkBack hears the count twice.

### Implementation

- **Name the trailing avatar.** Add `Modifier.semantics(mergeDescendants = true) { contentDescription = avatarLabel }` on the clickable, where `avatarLabel` comes from `ChatTheme.channelNameFormatter.formatChannelName(channel, currentUser)` — the same source the header title uses. TalkBack announces the avatar as `"<channel name>, button. Double-tap to <onClickLabel>."`.
- **Name the title `Column` when clickable.** Add `Modifier.semantics(mergeDescendants = true) {}` after the clickable. The merge folds the title + subtitle `Text` content into the button's label, so TalkBack announces the column as one node.
- **Mark the title as a heading.** Add `Modifier.semantics { heading() }` to the channel-name `Text`. TalkBack heading-navigation can now jump to the channel name.
- **Hide the overflow count badge.** Add `Modifier.semantics { hideFromAccessibility() }` to the `+N` `CountBadge` inside `StackedGroupAvatar`. The badge is visual emphasis; the formatter's `+N more` text already conveys the count to TalkBack.

### 🎨 UI Changes

None. All changes live in the accessibility tree only.

### Testing

1. **1:1 DM** — Open a 1:1 channel (e.g. the Padmé Amidala DM) and swipe through the header:
   - Back button: `"Back"`.
   - Title: `"<channel name>, heading"`. Heading-navigation jumps to it.
   - Avatar: `"<channel name>, button. Double-tap to open conversation info."` (the verb comes from `onChannelAvatarClickLabel`).
2. Double-tap the avatar — the conversation info screen opens.
3. **Group with overflow** — Open a group channel with 5+ members so the avatar renders the `+N` overflow badge. Confirm the avatar's announcement includes `+N more` once, not twice.
4. **Group with image** — Open a group channel with a `channel.image` set (uses `SimpleGroupAvatar`, no `CountBadge`). Avatar still names itself correctly.
5. **Title click** — If a downstream integration wires `onHeaderTitleClick` + `onHeaderTitleClickLabel`, TalkBack focuses the title `Column` as one node and announces `"<title>. <subtitle>. Button. Double-tap to <label>."`.

## Summary by CodeRabbit

* **Accessibility**
  * Enhanced screen reader support in channel headers with improved semantic descriptions
  * Better accessibility labels for channel avatars to help users navigate more effectively

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6464?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/ChannelHeader.kt
```diff
@@ -274,6 +280,7 @@ internal fun DefaultChannelHeaderTrailingContent(
     onClick: ((Channel) -> Unit)?,
     onClickLabel: String? = null,
 ) {
+    val avatarLabel = ChatTheme.channelNameFormatter.formatChannelName(channel, currentUser)
```
gpunto: Nit: we could move this inside the `ifNotNull` below so it's computed only if used.

gpunto (approved):
Looks good, just a nitpick you can safely ignore

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/header/ChannelHeader.kt
andremion: e29e813875c607e9b70a1b21900b72bf19248c7f

PR #6465, opened 2026-05-22
Read media durations as natural language for TalkBack

andremion:
AND-1180

### Goal

TalkBack used to read media durations character by character ("zero colon two three"), which is slow and hard to parse. This PR makes durations spoken as natural language ("23 seconds", "1 minute 23 seconds") across every surface that displays a duration in the Compose SDK.

### Implementation

- New `spokenDuration` helper in `PlaybackTimer.kt` produces a speech-friendly duration string via `MeasureFormat` (API 24+), falling back to the existing `ChatTheme.durationFormatter` on older API levels. A non-Composable variant is exposed for call sites that build strings outside Compose (`QuotedMessageBodyBuilder`).
- Applied as `contentDescription` on every visible `mm:ss` `Text` while keeping the displayed text unchanged:
  - Audio recording playback timer (`PlaybackTimerText`)
  - Composer audio recording — active, locked, and overview playback rows (`AudioRecordingContent`)
  - Video player playback time (`VideoPlaybackControls`)
  - Photos / Videos thumbnails — `VideoBadge` and `AudioBadge` (`MediaBadges`)
  - Quoted-message preview of an audio recording (`QuotedMessageBody.spokenText` populated by `QuotedMessageBodyBuilder`, consumed by `QuotedMessageText`)

No visible UI changes; the rendered text and layout are unchanged.

### Testing

1. Enable TalkBack.
2. Record and send a voice message. While recording, focus the duration counter — it should announce "X seconds" / "X minutes Y seconds" instead of digit-by-digit.
3. Lock the recording and let it run for a few seconds. Focus the overview playback row — same natural-language announcement.
4. Play back the attached audio in the message list — same announcement on the playback timer.
5. Quote a voice-message message. The quoted preview should announce "Voice message X seconds".
6. Send a video. On the Photos & Videos list, focus the thumbnail badge — it should announce the duration as natural language.
7. Open the full-screen video player — same announcement on the time counter.

## Summary by CodeRabbit

* **New Features**
  * Added voice-accessible duration descriptions for audio and media playback components, including playback controls, progress timers, media badges, quoted messages, and audio recording interfaces. Descriptions are locale-aware and integrate with screen readers and voice assistants.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6465?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/audio/PlaybackTimer.kt
```diff
@@ -40,12 +51,72 @@ internal fun PlaybackTimerText(
...
+ * Uses `MeasureFormat` (API 24+) for locale-aware natural-language output. On older API
+ * levels falls back to the visible clock format — same TalkBack experience as today.
+ *
+ * @param durationInMs Duration to format, in milliseconds.
+ */
+@Composable
+internal fun spokenDuration(durationInMs: Int): String =
+    spokenDuration(
+        durationInMs = durationInMs,
+        locale = requireNotNull(ConfigurationCompat.getLocales(LocalConfiguration.current)[0]) {
```
VelikovPetar: Should we maybe fallback to a reasonable default instead of crashing here? (I know it is most likely impossible to fail here, but I think we should not have such strict guards in code which is only for internal use)

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/audio/PlaybackTimer.kt
```diff
@@ -40,12 +51,72 @@ internal fun PlaybackTimerText(
...
+ * `contentDescription`. Visible UI keeps the compact `mm:ss` form; this helper produces the
+ * speech-friendly equivalent so TalkBack doesn't read raw `0:05` character by character.
+ *
+ * Uses `MeasureFormat` (API 24+) for locale-aware natural-language output. On older API
+ * levels falls back to the visible clock format — same TalkBack experience as today.
+ *
+ * @param durationInMs Duration to format, in milliseconds.
+ */
+@Composable
+internal fun spokenDuration(durationInMs: Int): String =
```
VelikovPetar: Should we maybe extract this to a separate file? I can already see it is used in several different packages, I think it will be easier to locate if we isolate it in a dedicated file.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/QuotedMessageBodyBuilder.kt
```diff
@@ -258,13 +270,17 @@ internal fun rememberBodyBuilder(): QuotedMessageBodyBuilder {
     val autoTranslationEnabled = ChatTheme.config.translation.enabled
     val durationFormatter = ChatTheme.durationFormatter
     val streamCdnImageResizing: StreamCdnImageResizing = ChatTheme.streamCdnImageResizing
+    val locale = requireNotNull(ConfigurationCompat.getLocales(LocalConfiguration.current)[0]) {
```
VelikovPetar: Same here, I am a bit hesitant to have such strict guard here.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/audio/PlaybackTimer.kt
```diff
@@ -40,12 +51,72 @@ internal fun PlaybackTimerText(
...
+    } else {
+        fallbackFormatter.format(durationInMs)
+    }
+
+@RequiresApi(Build.VERSION_CODES.N)
+private fun spokenDurationIcu(durationInMs: Int, locale: Locale): String {
+    val totalSeconds = (durationInMs / MillisPerSecond).coerceAtLeast(0L)
+    val minutes = totalSeconds / SecondsPerMinute
+    val seconds = totalSeconds % SecondsPerMinute
+    val formatter = MeasureFormat.getInstance(locale, MeasureFormat.FormatWidth.WIDE)
```
gpunto: Should we `remember` this? We're calling `spokenDuration` at every recomposition, which happens constantly during playback.

Maybe it's worth introducing a tiny internal class to handle this kind of formatting, so we can `remember` the class instance and only call `format(time)` on it.

We could probably also guard it on accessibility being actually enabled, so we don't compute anything at all if it's not.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/audio/PlaybackTimer.kt
andremion: 3c66ed7a94b8bab8c5d4c8b43ade24c4feede91f

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/audio/PlaybackTimer.kt
andremion: 6e7f4f6eb188af3eabdfbbf025c3f81f17c1308a

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/QuotedMessageBodyBuilder.kt
andremion: 3c66ed7a94b8bab8c5d4c8b43ade24c4feede91f

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/audio/PlaybackTimer.kt
andremion: 11d79548d43eb1e1f6cb1220ae29383aac75f318

VelikovPetar (approved):

gpunto (approved):

PR #6474, opened 2026-05-25
Channel pinning: list indicator and sheet header icons

andremion:
Closes AND-1185

### Goal

Implement the channel pinning visual redesign across the Compose SDK: render mute and pin icons on the channel list item and the `SelectedChannelMenu` header, surface Pin/Unpin Chat as a first-class action in the channel-options sheet, and expose a configurable indicator position so integrators can mix inline and trailing-bottom layouts per the Figma cookbook.

docs https://github.com/GetStream/docs-content/pull/1312

### Implementation

**Channel-options sheet (`ChannelOptions.kt`)**

- Moved Pin/Unpin Chat to slot 2 in the DM action list to match the redesign: `View Info → Pin/Unpin Chat → Mute/Unmute User → Block/Unblock User → Archive → Delete`. Group order already matched; refreshed the stale KDoc.
- `isPinChannelVisible = false` default preserved — the feature stays opt-in.
- Adopted the new `Channel.dmCounterpartId` helper to remove duplicated member iteration in `buildDmChannelActions`.

**`SelectedChannelMenu` header icons**

- Added `isMuted` and `isPinned` defaulted parameters to `SelectedChannelMenu`, `ChannelMenuParams`, and `ChannelMenuHeaderContentParams`.
- The default header renders inline mute and pin icons next to the channel name in a fixed `name → mute → pin` order when either flag is on. Each icon owns its own `contentDescription` at the leaf (no parent-level semantic override).
- `ChannelsScreen` composes `isChannelMuted(cid) || isUserMuted(dmCounterpartId)` at the call site — same OR the channel-list row uses — so the sheet header mirrors the row state.

**Channel list item pin indicator (`ChannelItem.kt`, `ChatUiConfig.kt`)**

- New public `PinIndicatorPosition` enum (`InlineTitle`, `TrailingBottom`) and a new `pinIndicatorPosition` field on `ChannelListConfig` (default value `InlineTitle`). Mirrors the existing `MuteIndicatorPosition` shape so the two attributes are configured independently — integrators can mix positions (mute inline + pin trailing, etc.).
- Extracted `MutedIcon` / `PinnedIcon` private composables to share rendering between the inline and trailing branches and keep `TitleRow` under detekt's `LongMethod` threshold.
- Preview and snapshot test config overrides now use `CompositionLocalProvider(LocalChatUiConfig provides …)` (matching the `MessageComposerInputLink` pattern). This also fixes a pre-existing dark-mode rendering bug in `muted_channel_trailing_bottom` caused by the previous nested-`ChatTheme` approach defaulting `isInDarkMode` back to the system value.

**Helpers**

- New `internal fun Channel.dmCounterpartId(currentUser: User?): String?` in `ChannelUtils.kt`. Used by both `buildDmChannelActions` and `ChannelsScreen`. Covered by `ChannelUtilsTest` (5 cases: DM with counterpart, non-distinct channel, distinct 3-member, distinct 2-member without current user, null current user).

**Localization**

- `stream_compose_channel_item_pinned` added to the eight supported locale files (default + es, fr, hi, in, it, ja, ko).

**Compose sample**

- New `CustomSettings.isChannelPinningEnabled` boolean pref (default `false`).
- Surfaced as a `FeatureFlag` toggle in the Custom Login settings panel, ordered between Adaptive layout and the composer flags to follow the app's screen-surface flow.
- `ChannelsActivity` and `ChatsActivity` read the flag to drive `ChannelOptionsVisibility.isPinChannelVisible`, and both compose a pinned-first sort `QuerySortByField<Channel>().desc("pinned_at").desc("last_updated")` as the recipe integrators can mirror.
- Both activities adopt the `private val settings by lazy { customSettings() }` field convention (matches `CustomLoginActivity`, `ChannelActivity`).

**API surface**

Additive only — `PinIndicatorPosition` enum, `pinIndicatorPosition` field, `isMuted`/`isPinned` fields on the two params data classes, and `SelectedChannelMenu` defaulted params. All new fields are trailing parameters with default values, so existing callers compile unchanged.

### 🎨 UI Changes

[Screen_recording_20260525_155105.webm]()

<table>
<thead>
<tr>
<th>Default</th>
<th>PinIndicatorPosition.TrailingBottom</th>
<th>MuteIndicatorPosition.TrailingBottom</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/413cc748-df2b-4a9e-830e-9674f469362e" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/6c0c1331-b953-45f0-83ff-1d3e71056de0" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/6c28c6cf-1770-4f09-8c47-134b9ad65e27" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### Testing

1. Run the `stream-chat-android-compose-sample` app and open **Custom Login**.
2. In the settings panel, enable the new **Channel pinning** flag, then log in with a user that has both one-to-one and group channels.
3. **Sheet header**:
   - Long-press a DM. Confirm the header shows the other user's name only (no icons yet).
   - Tap **Pin Chat**. Long-press the same DM again. Confirm the pin icon appears inline next to the user's name in the header.
   - Tap **Mute User**. Long-press again. Confirm the mute icon appears too, rendered before the pin icon (order: name → mute → pin).
   - Repeat with a group channel; the group sheet should show `View Info → Pin Chat → Leave Group → Delete Group` (assuming owner capabilities).
4. **Channel-options actions**:
   - Confirm the DM sheet shows actions in the order: `View Info → Pin/Unpin Chat → Mute/Unmute User → Block/Unblock User → Delete Chat`.
   - Confirm the action label toggles between **Pin Chat** / **Unpin Chat** based on the current state.
5. **Channel list item indicator — default inline position**:
   - On the channel list, confirm the pin icon appears inline next to the channel name for pinned channels.
   - Confirm pinned channels surface at the top of the list (pinned-first sort recipe).
6. **Trailing-bottom position**:
   - Temporarily override `ChannelListConfig(pinIndicatorPosition = PinIndicatorPosition.TrailingBottom)` in `ChannelsActivity`. Confirm the pin icon now renders at the trailing end of the message-preview row.
7. **Mixed positions**:
   - Configure `muteIndicatorPosition = InlineTitle` and `pinIndicatorPosition = TrailingBottom` on the same `ChannelListConfig`. Confirm the muted icon stays inline with the name while the pin icon moves to the trailing-bottom corner.
8. **Opt-in preserved**:
   - Turn off the **Channel pinning** flag. Confirm Pin/Unpin Chat disappears from the sheet (default opt-in behavior preserved; no breaking change for integrators).

Snapshot baselines for the five new pin variants plus the refreshed `muted_channel_trailing_bottom` baseline are recorded in `stream-chat-android-compose/src/test/snapshots/images/`.

## Release Notes

* **New Features**
  * Channel pinning: pin/unpin important channels for quick access
  * Configurable pin indicator positioning (inline or trailing bottom)
  * Channel muted and pinned state badges displayed in channel headers
  * Feature flag toggle to enable/disable channel pinning

* **Tests**
  * Added test coverage for pinned channel UI variants and states

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6474?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andremion:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1805,6 +1805,8 @@ public interface ChatComponentFactory {
             channelActions = params.channelActions,
             onChannelOptionConfirm = params.onChannelOptionConfirm,
             onDismiss = params.onDismiss,
+            isMuted = params.isMuted,
```
VelikovPetar: Just one small concern: Should we pass these values? Isn't it simple for the consumer to calculate the `isMuted`/`isPinned` state? Plus the `ismuted` might be a bit deceptive, because it can either reference the user mute (for DMs), or the channel mute.

andremion: I agree. Dropped both flags. The menu now derives them internally from `selectedChannel.isPinned()` and the user's mute lists (channel mute + DM-counterpart user mute), so consumers don't pass anything beyond what they already have. Pushed in 0678778d367116c4b163b3d8e14be50ece7db342.

VelikovPetar: Looks good, thank you!

VelikovPetar (approved):

PR #6479, opened 2026-05-29
Wire enhanced-mention fields through Message

gpunto:
### Goal

Add the new enhanced-mention fields and `Config.pushLevel` to the SDK's data layer so apps can read and write them end-to-end. UI lands in later PRs.

This is PR 1 in the enhanced-mentions series.

### Implementation

- `Message` gets `mentionedHere`, `mentionedChannel`, `mentionedRoles: List<String>`, `mentionedGroups: List<UserGroup>`.
- New `UserGroup` / `UserGroupMember` models; `Config.pushLevel: String?`.
- DTOs: upstream sends `mentioned_group_ids` (id list); downstream receives hydrated `mentioned_groups`. `DomainMapping` / `DtoMapping` round-trip the new fields.
- Room: `MessageEntity` / `ReplyMessageEntity` get a column per field; `mentionedGroups` is persisted as JSON via a new `UserGroupConverter`. `ChatDatabase` bumps its schema version.

### Testing

- `MessageMapperTest`, `DtoMappingTest`, `DomainMappingTest`, `DownstreamMessageDtoAdapterTest` round-trip the new fields.
- Manual: send a message with `mentionedHere = true` and confirm it round-trips through Room.

## Release Notes

* **New Features**
  * Added support for mentioning user groups in messages alongside existing mention types.
  * Enhanced message mention tracking to include `@here`, `@channel`, user groups, and roles.
  * Added push level configuration option for channels.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6479?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


gpunto:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/ConfigDto.kt
```diff
@@ -47,4 +47,5 @@ internal data class ConfigDto(
     val user_message_reminders: Boolean?,
     val shared_locations: Boolean?,
     val mark_messages_pending: Boolean,
+    val push_level: String? = null,
```
andremion, line 50: Looks like `pushLevel` isn't covered by any test. `randomConfig` and `randomConfigDto` don't set it, so it just stays `null` everywhere and the round-trip never really exercises it. Mind adding `push_level` to `randomConfigDto` (and `pushLevel` to `randomConfig`) so the existing config mapping tests actually check a non-null value?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/database/converter/internal/UserGroupConverter.kt
```diff
@@ -0,0 +1,36 @@
...
+internal class UserGroupConverter {
+
+    @OptIn(ExperimentalStdlibApi::class)
+    private val listAdapter = moshi.adapter<List<UserGroup>>()
+
+    @TypeConverter
+    fun userGroupListToString(groups: List<UserGroup>?): String? = groups?.let(listAdapter::toJson)
+
+    @TypeConverter
+    fun stringToUserGroupList(data: String?): List<UserGroup>? {
```
andremion, line 32: Every other Room converter has its own little `*ConverterTest` (`UserMuteConverterTest`, `ListConverterTest`, etc.), and this one has the trickiest logic of the bunch — the `"null"` guard, the empty handling, and the Date serialization. Worth a small round-trip test here too? Or are you planning to add it later?

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/UserGroup.kt
```diff
@@ -0,0 +1,60 @@
...
+ * @param createdBy User ID of the creator, when known.
+ * @param createdAt When the group was created.
+ * @param updatedAt Last time the group's metadata changed.
+ */
+@Immutable
+public data class UserGroup(
+    val id: String,
+    val name: String,
+    val description: String? = null,
+    val team: String = "",
```
andremion, line 39: Small thing: `team` defaults to `""` but `description` and `createdBy` are nullable. The mapper does `team_id.orEmpty()`, so "no team" ends up as an empty string instead of `null` like the others. Is the empty string on purpose (iOS alignment maybe?), or would `null` be more consistent with the rest of the model?

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/database/converter/internal/UserGroupConverter.kt
gpunto: Definitely should've added it. Done now!

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/UserGroup.kt
gpunto: I used the empty string as default because we are using that in `Channel`, but I have no strong opinion tbh.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/ConfigDto.kt
gpunto: Good catch, added!

PR #6480, opened 2026-05-29
Align modal bottom sheets with Figma

andremion:
Closes AND-1203

### Goal

Align every modal bottom sheet in the Compose SDK with the Figma design. The audit started from a dark-mode regression on the attachment command picker (it inherited M3's light `surfaceContainerLow` because the callsite didn't pass a `containerColor`) and revealed that modal bottom sheets across the SDK had each chosen their own values for shape, scrim, container color, and drag handle — none matching the Figma tokens, and three sheets bypassing M3 entirely with `Popup+AnimatedVisibility` or `Box+Surface`. Three deliverables fall out:

1. **Two opinionated wrappers — `StreamCardBottomSheet` and `StreamScreenBottomSheet`** — bake the Figma tokens (corner radii, elevation/app surface, scrim, drag handle, max width) in one place. Every `ModalBottomSheet` callsite in the SDK now goes through them.
2. **`ChannelActionsSheet`** replaces `SelectedChannelMenu` (deprecated, removed in v8) as a more discoverable, Figma-aligned channel-actions surface.
3. **`@Preview` and Paparazzi rendering** — every sheet has light + dark snapshot coverage so future Figma drift surfaces at PR review instead of in production.

### Implementation

Two internal wrappers consolidate sheet construction:

- `StreamCardBottomSheet` — card sitting above the app: 32dp top corners, `backgroundCoreElevation1` container, `backgroundCoreScrim` scrim, M3 drag handle.
- `StreamScreenBottomSheet` — full-takeover surface: `sheetMaxWidth = Dp.Unspecified`, rectangular, no drag handle, `backgroundCoreApp` container.

Every modal bottom sheet in the SDK now goes through one of those two wrappers, including three that were previously built from `Box+Surface` or `Popup+AnimatedVisibility` (`MediaGalleryPhotosMenu`, `PollAnswers`, `PollMoreOptionsDialog`) — they inherit drag handle, swipe-to-dismiss, and scrim from M3 for free.

`SelectedChannelMenu` and `SimpleMenu` are marked `@Deprecated` (removal in v8) in favour of `ChannelActionsSheet`, a public `@Composable` built on `StreamCardBottomSheet`. The factory hook `ChatComponentFactory.ChannelMenu` delegates to the new sheet, so integrators using `ChannelsScreen` pick up the new rendering automatically; the sample's `MyCustomUi` is also migrated. `SelectedChannelMenu` keeps its existing public surface and develop-aligned defaults (16dp shape, `overlayColor`, slot params) — the Figma redesign lives on `ChannelActionsSheet`, while the deprecated path stays visually close to develop until removal.

`@Preview` calls wrap each sheet in `Box(Modifier.fillMaxSize())` so the layout has bounds for M3 to compute the sheet's `Expanded` anchor against — without that, M3's internal show `LaunchedEffect` skips and the preview stays blank. For Paparazzi (single-frame capture; M3's show animation doesn't tick to completion in one frame), a private `rememberStreamSheetState` returns a pre-expanded `SheetState` under `LocalInspectionMode`; the IDE preview pane runs the animation on its own, so that swap is exclusively a Paparazzi enabler. Both behaviours are documented in the wrappers' KDoc, including the [Google Issue Tracker reference](https://issuetracker.google.com/issues/283843380).

Also fixes a small visual misalignment: `ChannelInfoMemberInfoModalSheetTopBar` was stacking a hardcoded `top = spacingMd` on top of the M3 drag handle's inset.

### 🎨 UI Changes

The rendered output for every modal bottom sheet is captured by the Paparazzi snapshots under `stream-chat-android-compose/src/test/snapshots/images/` — reviewing the diff there is the most reliable way to see the before/after across light and dark modes.

### Testing

Run the Compose sample in dark mode.

1. Open the message composer and trigger the attachment command picker (the original dark-mode bug). The sheet now renders with the dark elevation token, not M3's default light `surfaceContainerLow`.
2. Long-press a channel in the channel list. The channel actions sheet renders with 32dp rounded top corners, drag handle, and the Figma-aligned scrim — both in light and dark modes.
3. From the channel actions sheet, tap **Channel Info** → long-press a member. The member info sheet renders with the avatar aligned directly under the drag handle (no extra top padding stacking on top of M3's inset).
4. Open a poll message and trigger each of the poll bottom sheets from its overflow menu (poll results, poll comments, more options). Each takes over the full screen with a rectangular sheet, no drag handle, system bar inset applied.
5. Open a media attachment and trigger the gallery options menu and the photos menu. Both render as card sheets with the Stream scrim.
6. Uncomment `// MyCustomUi()` at `ChannelsActivity.kt:262` and re-run. Long-press a channel — confirm it opens `ChannelActionsSheet` (the new API), not the deprecated `SelectedChannelMenu`.

Automated coverage: Paparazzi snapshots (light + dark) for every modal bottom sheet in the SDK. `./gradlew :stream-chat-android-compose:verifyPaparazziDebug` runs all 820 tests green. IDE-preview rendering was spot-checked on `ChannelActionsSheet` and `PollMoreOptionsDialog`; the same wrapping pattern (`Box(Modifier.fillMaxSize())`) is applied to every other affected preview.

## Release Notes

* **New Features**
  * Introduced improved channel actions interface with a new sheet-based presentation for channel menu options.

* **Improvements**
  * Enhanced consistency across bottom sheet components throughout the app for a unified user experience.
  * Simplified channel menu presentation with cleaner, more streamlined interface interactions.
  * Modernized dialog and menu components for improved visual consistency.

* **Deprecations**
  * Deprecated legacy channel menu and action components in favor of new sheet-based alternatives (removal planned for v8).

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-android/pull/6480?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/info/ChannelActionsSheet.kt
```diff
@@ -0,0 +1,361 @@
...
+ * @param onActionClick Invoked when the user clicks an action. Destructive actions route through
+ * a separate confirmation step before executing.
+ * @param onDismiss Invoked when the sheet is dismissed.
+ * @param modifier Modifier applied to the sheet container.
+ * @param currentUser The currently logged-in user. Used by the default header to derive inline
+ * state icons (muted, pinned).
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+public fun ChannelActionsSheet(
```
VelikovPetar, line 82: Is it possible that we are introducing a naming disparity with the other platforms here?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -1792,19 +1792,19 @@ public interface ChatComponentFactory {
...
     /**
-     * Factory method for creating the full content of the SelectedChannelMenu.
+     * Factory method for creating the channel actions sheet shown on channel long-press.
      *
      * @param params Parameters for this component.
      */
     @Composable
     public fun ChannelMenu(params: ChannelMenuParams) {
-        SelectedChannelMenu(
+        ChannelActionsSheet(
```
VelikovPetar, line 1811: I don't fully understand why this changed - wasn't it possible to apply the new modals inside the SelectedChannelMenu?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/info/ChannelActionsSheet.kt
andremion: The name was chosen to reduce the disparity, not introduce one.
SwiftUI calls the equivalent [MoreChannelActionsView](https://github.com/GetStream/stream-chat-swiftui/blob/main/Sources/StreamChatSwiftUI/ChatChannelList/MoreChannelActionsView.swift).
`ChannelActionsSheet` is a lot closer to that than the previous `SelectedChannelMenu` was. I swapped the `View` suffix for `Sheet` to match Compose's semantics (Sheet vs Dialog vs Screen) and dropped the `More` prefix.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: The goal of this PR is to make every modal bottom sheet in the SDK go through two internal wrappers (StreamCardBottomSheet and StreamScreenBottomSheet). These wrappers consolidate the UI as the Figma specs.

`SelectedChannelMenu` has parts of its public API that do not fit this approach:
- `modifier` is used for positioning, including the centered-dialog rendering used in the sample app. The M3 ModalBottomSheet we use under the wrappers only renders at the bottom of the screen.
- `shape` and `overlayColor` are per-call overrides, but the goal of the wrappers is to define them once for the whole SDK.

Migrating `SelectedChannelMenu` in place would either drop these parameters and break existing consumers, or keep two different rendering paths inside the same component. Introducing `ChannelActionsSheet` as a smaller, new API and deprecating the old one was the simpler path. `SelectedChannelMenu` stays until v8, so existing consumers do not break.

VelikovPetar: Ah OK I see. In that case I guess we should be OK - I just didn't want to deprecate something that was introduced couple of versions ago.

VelikovPetar (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
andremion: Sorry, what do you mean by `introduced a couple of versions ago`?
`SelectedChannelMenu` was introduced in 2021 🤔 
Screenshot 2026-06-03 at 11 32 55

gpunto (approved):

PR #6483, opened 2026-06-02
Improve Giphy preview TalkBack accessibility

andremion:
AND-1180

### Goal

The ephemeral Giphy preview (rendered after a `/giphy` command) had three accessibility gaps:

1. **No signal that the preview appeared.** TalkBack focus stayed on the composer's voice-record button; users had to swipe blindly into the message list to find the bubble.
2. **Bubble announced only "Giphy".** No alt text, no "Only visible to you" context, no preview-state framing — the Giphy image had `contentDescription = null`, the inner `combinedClickable` created a nested accessibility merge boundary, and the visible-to-you row was a sibling focus.
3. **No confirmation after Send / Shuffle / Cancel.** TalkBack focus drifted unpredictably after the bubble unmounted, with no signal that the action took effect.

### Implementation

Three feature-scoped commits.

**1. Name the Giphy image for TalkBack.**

Set `contentDescription` on the inner `StreamAsyncImage` in `GiphyAttachmentContent` to `attachment.title?.takeIf(String::isNotBlank)`. When the attachment has a non-blank title (typically the search query supplied by the Giphy API), TalkBack reads it together with the existing "Giphy" badge — e.g. "Hello, giphy". Improves both the ephemeral preview path and sent Giphy messages.

**2. Focus the preview bubble on appearance.**

Wrap the visible-to-you row + Giphy image in `GiphyMessageContent` in a focusable `Column` with `Modifier.semantics(mergeDescendants = true) {}`, and `requestFocus()` on it 100 ms after the message id is composed. Mirrors the precedent in `MessageComposerAudioRecordingLockedContent`.

Adds `interactive: Boolean = true` to `GiphyAttachmentContent` and `GiphyAttachmentContentParams`. The ephemeral preview passes `false`, which skips the inner `combinedClickable` — its handlers are no-ops in the ephemeral case, and the inner clickable was a nested accessibility merge boundary that prevented the image alt and "Giphy" badge from merging into the bubble's focus.

TalkBack now announces the bubble as one focus, e.g. "Only visible to you, Hello, Giphy".

> [!NOTE]
> **The `interactive` workaround.** The cleaner design is to make `onItemClick` (and `AttachmentState.onLongItemClick`) nullable and gate `combinedClickable` on `handler != null`, which removes the need for a separate flag. That's a binary-breaking type change on a published API, so this PR takes the additive route. An inline comment in `GiphyAttachmentContent.kt` marks the spot; v8 should refactor to nullable handlers and drop `interactive`.

**3. Announce action state changes.**

Wrap each Send / Shuffle / Cancel button's `onClick` to fire `view.announceForAccessibility` before forwarding to `onGiphyActionClick`:

- `SendGiphy` → "Giphy sent"
- `ShuffleGiphy` → "Giphy shuffled"
- `CancelGiphy` → "Giphy preview cancelled"

The announce queues before the state change, so users hear the confirmation regardless of where TalkBack focus lands afterward (the focus drift itself is a TalkBack fallback-recovery race after the bubble unmounts, not deterministically fixable from this composable).

New strings translated across the 7 supported locales. No public API breaking changes; `interactive` is an additive defaulted param.

### Testing

1. Open a channel and type `/giphy hello`, then submit the command.
2. The ephemeral preview bubble appears in the message list. Expected: TalkBack focus moves to the bubble within ~100 ms and announces "Only visible to you, Hello, Giphy" as a single focus.
3. Swipe right to reach the **Send** button. Double-tap to send.
4. Expected: TalkBack announces "Giphy sent". The sent Giphy appears in the message list; focus drifts to a nearby message (the drift itself is a TalkBack fallback timing behaviour and may vary).
5. Repeat with `/giphy welcome`. On the preview, swipe to **Shuffle** and double-tap.
6. Expected: TalkBack announces "Giphy shuffled". The bubble stays mounted with a new Giphy. Swipe back to the bubble; the alt text reflects the new content.
7. Repeat with `/giphy something`. On the preview, swipe to **Cancel** and double-tap.
8. Expected: TalkBack announces "Giphy preview cancelled". The bubble unmounts.
9. Disable TalkBack and re-test all three actions to confirm the visible behaviour is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Giphy attachments now support configurable interactivity control
  * Enhanced accessibility for Giphy message interactions with TalkBack announcements and improved focus handling

* **Localization**
  * Added translations for Giphy action states (cancelled, sent, shuffled) across Spanish, French, Hindi, Indonesian, Italian, Japanese, and Korean


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
```diff
@@ -68,52 +76,75 @@ public fun GiphyMessageContent(
...
+        Column(
             modifier = Modifier
-                .fillMaxWidth()
-                .height(40.dp)
-                .padding(horizontal = StreamTokens.spacingSm, vertical = StreamTokens.spacingXs),
-            verticalAlignment = Alignment.CenterVertically,
-            horizontalArrangement = Arrangement.spacedBy(StreamTokens.spacingXs),
+                .focusRequester(previewFocusRequester)
+                .focusable()
+                .semantics(mergeDescendants = true) {},
```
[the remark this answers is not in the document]
andremion: Good catch — confirmed the regression for non-SR users (Compose's unified focus would yank the IME away from the composer after `/giphy …`).

Gated `focusRequester` + `focusable()` and the `requestFocus()` call behind a new private `rememberIsTouchExplorationEnabled()` helper that mirrors `SpokenDurationFormatter.rememberIsAccessibilityEnabled()` — same observer pattern, with `TouchExplorationStateChangeListener` for the more precise TalkBack check. The `view.announceForAccessibility` calls stay unconditional since they are no-ops when no SR service is active.

Fixed in cd9c6ba38164a2f1c58586cd71d6da89f74cd4c9 (amended commit 2).

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
```diff
@@ -68,52 +90,78 @@ public fun GiphyMessageContent(
...
 ) {
     val colors = ChatTheme.colors
+    val view = LocalView.current
+    val sentAnnouncement = stringResource(R.string.stream_compose_message_list_giphy_sent)
+    val cancelledAnnouncement = stringResource(R.string.stream_compose_message_list_giphy_cancelled)
+    val shuffledAnnouncement = stringResource(R.string.stream_compose_message_list_giphy_shuffled)
+
+    val isTouchExplorationEnabled = rememberIsTouchExplorationEnabled()
+    val previewFocusRequester = remember { FocusRequester() }
+    LaunchedEffect(message.id) {
```
gpunto: Do we re-focus if the message goes out of view and then back (so it's removed and readded by the lazy column)? If so, is that intended?

gpunto (approved):
LG, just a question

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
andremion: Good catch!
41ba7ad5c352e8418e23a7840323da7a51581c7d

PR #6466, opened 2026-05-22
Improve poll dialogs TalkBack accessibility

andremion:
AND-1180

### Goal

The poll-results bottom sheet, the per-option votes screen, the comments dialog, and the more-options dialog were hard to navigate with TalkBack: the screen title wasn't announced on open, sub-section labels weren't navigable as headings, vote/comment rows produced one focus per child element (avatar, name, date), and the radio/checkbox in the more-options dialog was unlabeled. This PR fixes all of that so each row is a single focus with a meaningful announcement, and dialog titles identify their screen.

### Implementation

- `PollDialogHeader` (shared by all four poll dialogs): title `Text` carries `heading()` semantics, and the header sends a programmatic `view.announceForAccessibility(title)` on first composition. `ModalBottomSheet` swallows the standard `paneTitle` window-state-changed event, so programmatic announce is the reliable workaround.
- `PollVoteItem` (shared by results + per-option votes screens): the row merges its descendants so each vote announces as one focus (avatar + name + date).
- `PollViewResultDialog`: option header (`Option N`) + summary row (option text + winner badge + vote count) wrapped in a merged `Column` with `heading()` — one focus per option summary, navigable via TalkBack's heading shortcut. Question section also merged with `heading()`.
- `PollOptionVotesDialog`: the trophy-and-vote-count header row above the votes list is merged.
- `PollAnswers`: each comment card merges the answer text with the user/timestamp row into one focus; the "Update Your Comment" button stays as its own separate focus.
- `PollMoreOptionsDialog`: applies the same pattern as the just-merged inline poll (`PollMessageContent`, #6461): each option row is `toggleable` with `Role.RadioButton` (single-vote polls) or `Role.Checkbox` (multi-vote), the `RadioCheck` is hidden from accessibility since the toggleable row already conveys state, the vote count gets a plural `contentDescription`, and the `LinearProgressIndicator` is removed from the semantics tree (the row already carries the count). Question section also merged with `heading()`.

No visible UI changes; layout, colours, and behaviour are unchanged.

### Testing

1. Enable TalkBack.
2. Long-press a poll message and open **View Results**:
   - Sheet open should announce the screen title ("Poll Results").
   - Swiping right should focus the back button, then the question section (one focus, announced as a heading), then each option summary ("Option 1, A, 3 votes" / "Option 1, A, Winner badge, 3 votes"), then each vote row as one focus, then the "Show all" button (when present), then the total votes count.
   - TalkBack's heading-navigation shortcut should jump between the question and each option.
[...]

PR #6440, opened 2026-05-15
Improve message list TalkBack accessibility

andremion:
### Goal

Resolve TalkBack accessibility gaps on the Compose message list (channel screen, threads screen). Issues a screen-reader user hits today:

1. Every message row announces as a generic *"Message item, …"*, **the sender name is silently hidden** (a `clearAndSetSemantics` wrapper kept only the testTag), and tap/long-press hints have no verb.
2. Date separators (*"Today"*, *"Yesterday"*) and the unread divider are plain text — TalkBack's swipe-by-heading gesture skips them. The unread divider also renders *"1 unread messages"* for `n = 1` (no `<plurals>`).
3. The typing indicator is visual-only (avatars + animated dots). TalkBack picks nothing up — no focus, no announcement when someone starts typing.
4. Reactions: the long-press reaction picker shows a checked background for the user's own reactions but exposes no role or selected state, so a TalkBack user can't tell which reactions they already chose. In-bubble reaction chips have no action verb.
5. Image / video / file / link attachments announce *"double-tap to activate"* — no concrete verb for the action.

### Implementation

Five self-contained commits, each scoped to one concern:

1. **`Make message rows accessible to TalkBack`** — drop the static *"Message item"* placeholder on the outer `Row`; `combinedClickable` gains `onClickLabel = "Open thread"` (when `canOpenThread`) and `onLongClickLabel = "Show message options"` (when `canOpenActions`). `MessageFooter` sender-name `Text` drops its destructive `clearAndSetSemantics { testTag = … }` for `Modifier.testTag(…)` so the user name is spoken again. Drops the orphaned `stream_compose_cd_message_item` placeholder across all 8 locales.
2. **`Make message list separators accessible as headings`** — add `Modifier.semantics { heading() }` to both the date and unread separators in `DefaultMessageDateSeparatorContent` / `DefaultMessageUnreadSeparatorContent`. Convert `stream_compose_message_list_unread_separator` from `<string>` to `<plurals>` across all 8 locales so n=1 reads as a singular.
3. **`Announce typing indicator to TalkBack`** — wrap the indicator row in a polite live region with a localized description (*"X is typing"* / *"X and Y are typing"* / *"N people are typing"*, reusing the existing `stream_compose_channel_list_typing_*` strings). The description is built by a small private helper to keep the composable under the method-length cap.
[...]

PR #6484, opened 2026-06-02
Improve message actions menu TalkBack accessibility

andremion:
AND-1180

### Goal

The long-press message actions menu and the emoji reactions surfaces had four screen-reader gaps:

1. **Reactions announced as checkboxes.** `ReactionToggle` used `Modifier.toggleable(role = Role.Checkbox)`, so TalkBack read "[emoji] sign, checkbox, not checked, double tap to toggle". The Figma a11y principle calls this out explicitly as the *Don't* case; reactions are toggle buttons (tap to add / tap to remove), not checkboxes.
2. **No status confirmation on toggle.** Tapping a reaction emitted no screen-reader signal that the toggle state changed.
3. **Menu opening leaked host activity label.** Opening the long-press actions menu announced the host activity's `android:label` ("Chat Sample Compose") before the focused element, with no signal of the new surface.
4. **Emoji picker sheet focused the drag handle.** Opening the picker via "+" in the reactions bar landed TalkBack on the drag handle, with no signal that the picker surface had appeared.

### Implementation

Four feature-scoped commits.

**1. Expose reactions as toggle buttons to TalkBack.**

In `ReactionToggle`, switch the toggleable role from `Role.Checkbox` to `Role.Button` and attach an explicit `stateDescription` of "Pressed" / "Not pressed". TalkBack now reads "[emoji], button, pressed / not pressed, double tap to toggle" — the canonical Android toggle-button pattern, since Compose has no `Role.ToggleButton`. Internal composable; no public API change. This overrides #6440's deliberate `Role.Checkbox` decision in favour of the design source-of-truth.

**2. Announce reaction add and remove.**

Wrap `ReactionToggle.onValueChange` to fire `view.announceForAccessibility("[emoji] reaction added" / "removed")` before forwarding the new state. Both the long-press reactions bar and the full emoji picker grid benefit automatically because the picker's `ReactionMenuOptionItem` factory delegates to the same `ReactionToggle`. The string carries the emoji character as `%1$s`; TalkBack reads the character with its locale-native name, so the announce reads naturally as "Thumbs up sign reaction added" without maintaining a per-reaction name table.

**3. Announce the message actions menu pane title.**

Declare `Modifier.semantics { paneTitle = "Message actions" }` on the inner `Column` inside the `Dialog`. TalkBack treats `paneTitle` changes as a window-state event and announces the title on dialog entry, replacing the previously-leaked host activity label.

**4. Announce the emoji picker sheet.**

`ModalBottomSheet` swallows the standard `paneTitle` window-state event (verified in #6466 / `PollDialogHeader`), so fall back to programmatic announce via `view.announceForAccessibility` in a `LaunchedEffect`. Hide the sheet's drag handle from the accessibility tree via `Modifier.semantics { hideFromAccessibility() }` — the handle stays visually present and remains draggable for sighted users; TalkBack focus skips it and lands on the first emoji in the picker grid.

Migrated `ReactionsPicker` to the `StreamCardBottomSheet` wrapper introduced in #6480 so the emoji picker matches the rest of the SDK's card sheets (32dp top corners, `backgroundCoreElevation1` container, Stream scrim). Added an optional `dragHandle` parameter to `StreamCardBottomSheet` (internal API; defaults to `BottomSheetDefaults.DragHandle()`) so the emoji picker can pass its accessibility-hidden handle. All other callers of `StreamCardBottomSheet` keep the default and are unaffected.

New strings translated across the 7 supported locales:
- `stream_compose_reactions_pressed`, `stream_compose_reactions_not_pressed`
- `stream_compose_reactions_added`, `stream_compose_reactions_removed`
- `stream_compose_message_actions_menu_title`
- `stream_compose_emoji_picker_title`

No public API changes.

### Testing

1. Long-press a message in the chat. Expected: TalkBack announces "Message actions" on dialog open, then reads the focused reaction as "[emoji], button, not pressed, double tap to toggle".
2. Double-tap a reaction in the reactions bar. Expected: TalkBack announces "[emoji] reaction added" (live region). Double-tap again to remove: "[emoji] reaction removed".
3. From the reactions bar, double-tap the "+" button to open the full emoji picker sheet. Expected: TalkBack announces "Emoji picker" then bottom sheet context. Focus does **not** land on the drag handle. Swipe to navigate the grid; tap an emoji to fire the same add/remove announce as step 2.
4. Disable TalkBack and re-test all three surfaces to confirm visible behaviour is unchanged (the drag handle is still visible and draggable).

## Summary by CodeRabbit

* **New Features**
  * Enhanced accessibility for the message actions menu with proper semantic labeling.
  * Improved reactions picker with better screen reader support and accessibility announcements.
  * Added accessible state descriptions for reaction buttons to better communicate interaction states.

* **Localization**
  * Added new UI text strings across 8+ languages (Spanish, French, Hindi, Indonesian, Italian, Japanese, Korean, and English) for emoji picker, message actions menu, and reaction states.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionpicker/ReactionsPicker.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionpicker/ReactionsPicker.kt
@@ -17,9 +17,16 @@
 package io.getstream.chat.android.compose.ui.components.reactionpicker
 
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.material3.BottomSheetDefaults
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalView
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.hideFromAccessibility
+import androidx.compose.ui.semantics.semantics
+import io.getstream.chat.android.compose.R
 import io.getstream.chat.android.compose.ui.components.StreamCardBottomSheet
 import io.getstream.chat.android.compose.ui.components.reactionoptions.ExtendedReactionsOptions
 import io.getstream.chat.android.compose.ui.theme.ChatTheme
@@ -45,9 +52,20 @@ public fun ReactionsPicker(
     modifier: Modifier = Modifier,
     onDismiss: () -> Unit = {},
 ) {
+    val view = LocalView.current
+    val emojiPickerTitle = stringResource(R.string.stream_compose_emoji_picker_title)
+    // ModalBottomSheet swallows the standard `paneTitle` window-state-changed event, so
+    // announce the title programmatically. See PollDialogHeader for the same workaround.
+    LaunchedEffect(emojiPickerTitle) { view.announceForAccessibility(emojiPickerTitle) }
+
     StreamCardBottomSheet(
         onDismissRequest = onDismiss,
         modifier = modifier,
+        dragHandle = {
+            BottomSheetDefaults.DragHandle(
+                modifier = Modifier.semantics { hideFromAccessibility() },
+            )
+        },
     ) {
         ChatTheme.componentFactory.MessageReactionsPickerContent(
             params = MessageReactionsPickerContentParams(
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionpicker/ReactionsPicker.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionpicker/ReactionsPicker.kt
@@ -17,15 +17,12 @@
 package io.getstream.chat.android.compose.ui.components.reactionpicker
 
 import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.material3.BottomSheetDefaults
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.semantics.hideFromAccessibility
-import androidx.compose.ui.semantics.semantics
 import io.getstream.chat.android.compose.R
 import io.getstream.chat.android.compose.ui.components.StreamCardBottomSheet
 import io.getstream.chat.android.compose.ui.components.reactionoptions.ExtendedReactionsOptions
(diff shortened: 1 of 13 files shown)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionpicker/ReactionsPicker.kt
```diff
@@ -45,9 +52,20 @@ public fun ReactionsPicker(
...
+    LaunchedEffect(emojiPickerTitle) { view.announceForAccessibility(emojiPickerTitle) }
+
     StreamCardBottomSheet(
         onDismissRequest = onDismiss,
         modifier = modifier,
+        dragHandle = {
+            BottomSheetDefaults.DragHandle(
+                modifier = Modifier.semantics { hideFromAccessibility() },
+            )
+        },
```
gpunto: Should we make this the default directly in `StreamCardBottomSheet`? If you're using talkback you're not dragging anyway, right?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactions/ReactionToggle.kt
```diff
@@ -60,6 +66,13 @@ internal fun ReactionToggle(
 ) {
     emoji?.let {
         val containerSize = size.toContainerSize()
+        val selectedStateDescription = stringResource(R.string.stream_compose_reactions_state_selected)
+        val notSelectedStateDescription = stringResource(R.string.stream_compose_reactions_state_not_selected)
+        val selectActionLabel = stringResource(R.string.stream_compose_reactions_action_select)
+        val unselectActionLabel = stringResource(R.string.stream_compose_reactions_action_unselect)
+        val addedAnnouncement = stringResource(R.string.stream_compose_reactions_added, emoji)
+        val removedAnnouncement = stringResource(R.string.stream_compose_reactions_removed, emoji)
```
gpunto: WDYT about using more generic terms than added/removed, e.g. selected/unselected? From the reaction toggle itself, we don't know that the caller is using the toggle to add/remove.

gpunto (approved):
Looks good, just a couple of suggestions

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactionpicker/ReactionsPicker.kt
andremion: Good call. Applied in f9d5581ff7e87c30545de997ea157737b7a5238c: `StreamCardBottomSheet` now bakes `Modifier.semantics { hideFromAccessibility() }` into its default drag handle, and the per-call-site override here is gone.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/reactions/ReactionToggle.kt
andremion: Renamed in f9d5581ff7e87c30545de997ea157737b7a5238c. Variables and `stream_compose_reactions_*` keys are now `selected` / `unselected`; translations updated across the 7 supported locales.

PR #6488, opened 2026-06-03
Improve message list TalkBack accessibility

andremion:
AND-1180

### Goal

Two TalkBack issues on the message list, bundled because they share a screen and a verification path.

1. **Quoted reply context is invisible to TalkBack.** When focus lands on a quoted preview inside a reply, the bubble announces only the quoted sender and text. Screen-reader users have no signal about *who* replied to *whom* — they have to swipe to the reply row to piece the relationship together.
2. **Message rows announce "double-tap to activate" but the tap does nothing.** For regular messages (no thread started yet, the common case) the row's `combinedClickable` exposes a click action to the accessibility tree even though `onClick` is guarded to a no-op. TalkBack promises an action that doesn't fire.

### Implementation

**Quoted reply context** (`QuotedMessage.kt`)

Extend the existing `accessibilityName` override on `QuotedMessageUserName` to cover the in-list case (`replyMessage != null`). The quoted bubble's sender label now reads as one of:

- `"You replied to your message"` — replier and quoted are both the current user (self-thread).
- `"You replied to {name}'s message"` — current user replied to someone else.
- `"{name} replied to your message"` — someone else replied to the current user.
- `"{name} replied to {other name}'s message"` — someone else replied to a third party.

TalkBack then reads the bubble as `"{relationship}, {quoted text}, double-tap to activate"` — context first, content second. The composer-banner case (`replyMessage == null`) is unchanged. Two new string keys (`stream_compose_quoted_message_replied_to_your_message`, `stream_compose_quoted_message_replied_to_their_message`) translated across the 7 supported locales.

**Row click hint** (`MessageContainer.kt`)

Split the row's modifier into three branches by capability so the accessibility tree only carries the actions the row actually performs:

- `canOpenThread`: keep `combinedClickable` with both `onClick` (open thread) and `onLongClick` (open actions, when available).
- `canOpenActions` only: `Modifier.pointerInput { detectTapGestures(onLongPress = …) }` + `Modifier.semantics(mergeDescendants = true) { onLongClick(label) }`. Long-press still opens the actions menu; TalkBack reads only `"double-tap and hold to show message options"`; descendants merge into a single row focus.
- Neither (deleted, uploading): empty `Modifier.semantics(mergeDescendants = true) {}` — the row stays a single focus target with no interaction hints.

Trade-off: the `canOpenActions`-only branch loses the press-down ripple that `combinedClickable` applied through its default `indication`. Sighted users still get feedback from the actions menu opening; the ripple loss is the smaller of the two regressions to keep.

Public API unchanged. The modifier is a local `val` in `MessageContainer`.

### Testing

Enable TalkBack on the device and open a channel in the Compose sample.

**Quoted reply announcement**

1. Have user A send a message; reply to it as user B. Focus the quoted preview inside B's reply.
    - **Expected**: `"{A} replied to your message, {quoted text}, double-tap to activate"`.
2. As user A, reply to one of A's own messages (self-thread). Focus the quoted preview.
    - **Expected**: `"You replied to your message, {quoted text}, …"`.
3. As user A, reply to one of user B's messages. Focus the quoted preview.
    - **Expected**: `"You replied to {B}'s message, {quoted text}, …"`.
4. As user C, look at a reply where user A replied to user B. Focus the quoted preview.
    - **Expected**: `"{A} replied to {B}'s message, {quoted text}, …"`.
5. Open the composer reply banner (tap **Reply** on a message). Focus the quoted preview in the banner.
    - **Expected**: announcement unchanged from before this PR (the in-list relational copy is not applied here).

**Row click hint**

1. Find a message with **no thread started** in the list. Focus the row.
    - **Expected**: TalkBack reads the message content followed by `"double-tap and hold to show message options"`. **No** `"double-tap to activate"` hint.
    - Long-press still opens the actions menu.
2. Find a message with **an existing thread** (reply count > 0). Focus the row.
    - **Expected**: TalkBack reads the message followed by `"double-tap to open thread"` and the long-press hint.
    - Tap opens the thread; long-press opens actions.
3. Find a **deleted** message in the list. Focus the row.
    - **Expected**: row receives a single TalkBack focus with no double-tap hint and no action.
4. Send a new message so it briefly shows as **uploading**. Focus the row before it resolves.
    - **Expected**: same as deleted — single focus, no hint.

**Localization smoke test**

Switch the device language to one of the supported locales (es, fr, it, ja, ko, hi, in) and repeat the four quoted-reply cases. Verify the translated strings render with the names substituted in the correct positions.

## Summary by CodeRabbit

* **New Features**
  * Enhanced accessibility for quoted message replies with improved screen reader support, including context-aware descriptions for who replied to whose message
  * Improved touch interaction handling for message actions and thread navigation

* **Localization**
  * Added quoted message reply translations for Spanish, French, Hindi, Indonesian, Italian, Japanese, and Korean


andremion:
@coderabbit review

gpunto (approved):

PR #6492, opened 2026-06-08
Align Channel List feedback states with Figma (loading error, action failure, deletion)

andremion:
### Goal

The Channel List in the Compose SDK had no feedback for failed or completed actions. This aligns it with the three **Locked** Figma states on *Chat SDK Design System → Android → Channel List* (AND-1215):

1. **Loading Error** — a pagination failure showed nothing.
2. **User-triggered Action Failure** — channel actions (mute, pin, archive, delete, leave, block, …) failed silently.
3. **Channel Deleted** — a successful delete gave no confirmation.

### Implementation

- **Loading-error banner (item 1):** added `ChannelsState.loadingError`, set on a load-more failure and cleared on the next successful load. The stateless `ChannelList` renders a new `ChannelListBanner` ("Couldn't load new chats. Tap to retry") pinned below the header; tapping it retries via `loadMore()`. Exposed through a `ChatComponentFactory.ChannelListBanner` slot + params for customization. Mirrors the Thread List "Tap to retry" pattern.
- **Action feedback (items 2 & 3):** `ChannelListViewModel` now exposes an **internal** `events: SharedFlow<ChannelListEvent>`. The 12 channel/user action handlers emit `ActionError(action)` on failure; delete emits `ChannelDeleted` on success. `ChannelsScreen` hosts a `StreamSnackbarHost` and maps events to snackbars — an error pill ("Error muting channel.", etc.) and a "Chat deleted." confirmation — reusing the existing `StreamSnackbar`.
- Per-action error strings + "Chat deleted.", translated into all 7 supported locales.
- No public API beyond item 1's `loadingError` field/handler (the events API is internal).

### 🎨 UI Changes

_Before: a pagination failure and failed/succeeded actions produced no visible feedback._

| Loading-error banner | Action-failure snackbar |
| --- | --- |
| and1215_run_states | and1215_run_snackbar |

### Testing

Unit tests cover the new behavior (`ChannelListViewModelTest`: `loadingError` set on failure / cleared on successful retry / kept on failed retry; action failure emits `ActionError`; delete emits `ChannelDeleted`). A Paparazzi snapshot covers `ChannelListBanner`; the snackbar variants are covered by `StreamSnackbarTest`.

Manual steps (compose sample):
- **Banner:** on a list with more than one page of channels, scroll to the bottom while the next-page request fails (e.g. via a proxy) → the banner appears below the header; tap it to retry; on success it dismisses.
- **Action-failure snackbar:** trigger a channel action that fails (e.g. mute while the request fails) → "Error muting channel." appears at the bottom.
- **Delete confirmation:** long-press a channel → **Delete Chat** → confirm → "Chat deleted." appears.

<details>

<summary>Force the loading-error banner and an action-failure snackbar on open</summary>

```diff
// ChannelListViewModel.kt — replay the events flow and emit a sample action error on open
-    private val _events = MutableSharedFlow<ChannelListEvent>(extraBufferCapacity = 1)
+    private val _events = MutableSharedFlow<ChannelListEvent>(replay = 1, extraBufferCapacity = 1)

         viewModelScope.launch { init() }
+        viewModelScope.launch {
+            _events.tryEmit(
+                ChannelListEvent.ActionError(
+                    ChannelListAction.MuteChannel,
+                    io.getstream.result.Error.GenericError("debug"),
+                ),
+            )
+        }

// ChannelList.kt (stateless ChannelList) — always render the banner
-                if (channelsState.loadingError) {
+                if (true) {
```

</details>

## Release Notes

* **New Features**
  * Tap-to-retry error banner when channel list fails to load
  * Action error notifications for channel management (mute, pin, archive, delete, leave group) and user interactions (mute, block, unblock)
  * Localized error messages across multiple languages

* **Tests**
  * Added test coverage for error handling and retry banner components


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/channels/list/ChannelListEvent.kt
```diff
@@ -0,0 +1,57 @@
...
+ */
+internal sealed interface ChannelListEvent {
+
+    /**
+     * A user-triggered channel action failed.
+     *
+     * @param action The action that failed, used to pick the feedback message.
+     * @param error The error that caused the failure.
+     */
+    data class ActionError(val action: ChannelListAction, val error: Error) : ChannelListEvent
```
gpunto: Do we need `error`? it seems we're not reading it 🤔

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelList.kt
```diff
@@ -114,6 +117,7 @@ public fun ChannelList(
     onChannelLongClick: (Channel) -> Unit = remember(viewModel) { { viewModel.selectChannel(it) } },
     onSearchResultClick: (Message) -> Unit = {},
     onStartChatClick: (() -> Unit)? = null,
+    onLoadingErrorClick: () -> Unit = remember(viewModel) { { viewModel.loadMore() } },
```
gpunto: This is a silent behavior change that would break integrators using trailing lambda syntax. If they had something like

```kotlin
ChannelList(other params...) { handle on start chat click }
```

this would still compile, but now does something different: it triggers on `onLoadingErrorClick` rather than `onStartChatClick`. I think we should avoid it. Probably we need a new ChannelList overload. Any other idea?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
```diff
@@ -908,14 +942,18 @@ public class ChannelListViewModel internal constructor(
         }
         lastNextQuery = nextQuery
         logger.v { "[loadMoreQueryChannels] offset: ${nextQuery.offset}, limit: ${nextQuery.limit}" }
+        // Preserve loadingError until the outcome is known, so a failing retry doesn't clear and re-set it.
```
gpunto, line 945: Should we clear the errors in `observeSearchMessages`? Otherwise we risk that an error in e.g. pagination, would persist when the user searches.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/ChannelListActionMessages.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.annotation.VisibleForTesting
+import io.getstream.chat.android.compose.R
+import io.getstream.chat.android.compose.state.channels.list.ChannelListAction
+
+/**
+ * The snackbar message shown when [this] channel action fails.
+ */
+@StringRes
+@VisibleForTesting
+internal fun ChannelListAction.errorMessageResId(): Int = when (this) {
```
gpunto: WDYT about putting the string res directly as a property of the enum (`val errorMessageResId`) so we avoid this code?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelListBanner.kt
```diff
@@ -0,0 +1,91 @@
...
+ * @param onClick Action invoked when the banner is tapped.
+ */
+@Composable
+internal fun ChannelListBanner(
+    modifier: Modifier = Modifier,
+    onClick: () -> Unit = {},
+) {
+    val color = ChatTheme.colors.chatTextSystem
+
+    Row(
```
gpunto, line 57: Do we need to add any accessibility semantics here? E.g. this banner has an onClick so maybe it should have the button role or some on click label.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelList.kt
andremion: What about moving `onLoadingErrorClick` before `onStartChatClick` so `onStartChatClick` stays in the trailing slot?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/channels/list/ChannelListEvent.kt
andremion: I initially introduced it because `ChannelListEvent` was public at first. Then I made it internal, so it doesn't make sense now.
I will remove it

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/ChannelListActionMessages.kt
andremion: Good idea!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelList.kt
gpunto: I thought about that, but it still has a problem (albeit a smaller one). Formally, it would break consumers that pass **all** parameters and **do not** use named parameters nor trailing lambda 😅 like:

```kotlin
ChannelList(modifier, contentPadding, ..., {}, {})
```

I'm not sure how common this scenario is though 🤷‍♂️

andremion: I believe that's the least probable: passing all ~9 params positionally with no names and no trailing lambda (so onStartChatClick's position shifts) 😅

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/state/channels/list/ChannelListEvent.kt
andremion: 0bb3116fb0d9704210852b5ded8a70887d181fdd

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
andremion: 54390ba6ca0dbcb6972e5885d4349f21b88f8af1

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/ChannelListActionMessages.kt
andremion: 43baf3d3906683f8d0d576409241a56f674c38ce

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelListBanner.kt
andremion: 977ee19caa1dcceaffad1ffd7e7a21e324fe2062

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channels/list/ChannelList.kt
andremion: b8a544c6ebe54a841ad88b45ee7d5b043024eeee

gpunto (approved):

PR #6493, opened 2026-06-09
Add ChatClient endpoints for user groups, roles, and chat preferences

gpunto:
### Goal

Add the `ChatClient` endpoints the enhanced-mentions UI needs: user-group CRUD/search, role search, and granular chat-preference toggles. No UI yet.

This is PR 2 in the enhanced-mentions series.

### Implementation

- User groups: `createUserGroup`, `updateUserGroup`, `deleteUserGroup`, `getUserGroup`, `queryUserGroups`, `searchUserGroups`, `addUserGroupMembers`, `removeUserGroupMembers`. New Retrofit interface `UserGroupApi` with matching request/response DTOs.
- Roles: `searchRoles` + new `Role` model and `DownstreamRoleDto`. New Retrofit interface `RoleApi`.
- Chat preferences: `setUserChatPreferences(prefs)` and `setChannelChatPreferences(cid, prefs)`. New `ChatPreferences` model with a `ChatPreferenceToggle` per mention type. `PushPreference` gains a `chatPreferences` field so existing push-preference reads surface the new toggles.
- `ChatApi` / `MoshiChatApi` wire the new endpoints; `ChatModule` registers `RoleApi` and `UserGroupApi`.

### Testing

- `MoshiChatApiTest` covers every new endpoint: request shape, response mapping, and error paths.
- API dump regenerated for `stream-chat-android-client` and `stream-chat-android-core`.

## Summary by CodeRabbit

* **New Features**
  * Comprehensive user group management: create, query/search, retrieve, update, delete, and manage group members
  * Role search to discover available roles
  * Granular per-user and per-channel chat preferences for category-specific push notification controls (mention/thread defaults)


gpunto:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1730,6 +1733,228 @@ internal constructor(
...
+     * channel-assignable roles. Both kinds are returned when null.
+     * @param includeGlobalRoles When `true`, include cross-team operator roles whose name starts
+     * with `global_`. Defaults to `false` server-side.
+     * @param nameGt Cursor: roles whose name sorts after this value.
+     */
+    @CheckResult
+    public fun searchRoles(
+        query: String,
+        limit: Int? = null,
+        roleType: String? = null,
```
andremion: Small thing: `roleType` is only ever `"user"` or `"channel"`, but it's a plain String here. You wrap the other closed sets in this PR (`ChatPreferenceToggle`), and `PushPreferenceLevel` does the same, so this one is inconsistent. Since it's a public API, changing to a typed value later would be a breaking change. Want to wrap it now, or keep it as a String on purpose?

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/ChatPreferences.kt
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.models
+
+/**
+ * Per-category push toggles. A null field means "use the server default". When set on a
+ * [PushPreference], takes precedence over [PushPreference.level].
+ */
+public data class ChatPreferences(
```
andremion, line 26: `Role` and `UserGroup` from PR 1 both have `@Immutable`, but `ChatPreferences` and `ChatPreferenceToggle` don't. They are immutable value types that end up in Compose state via `PushPreference.chatPreferences`, so they would benefit from it too. Is it on purpose?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
```diff
@@ -2924,6 +3022,241 @@ internal class MoshiChatApiTest {
...
+        result `should be instance of` expected
+        verify(api, times(1)).removeUserGroupMembers(
+            id = id,
+            body = RemoveUserGroupMembersRequest(member_ids = memberIds, team_id = teamId),
+        )
+    }
+
+    @ParameterizedTest
+    @MethodSource("io.getstream.chat.android.client.api2.MoshiChatApiTestArguments#searchRolesInput")
+    fun testSearchRoles(call: RetrofitCall<SearchRolesResponse>, expected: KClass<*>) = runTest {
```
andremion, line 3228: These check the request shape and Success/Failure, but nothing asserts the actual mapped `Role` / `UserGroup` fields, so a mistake in `DownstreamRoleDto.toDomain()` would not be caught. The mappings are trivial 1:1, so the risk is low. Maybe add one assertion on a mapped field to cover it?

andremion (approved):

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
gpunto: Great point, I actually should've added standalone domain mapping tests. Added them in `DomainMappingTest` in https://github.com/GetStream/stream-chat-android/pull/6493/changes/738a18ceeb3117fc5f80a927db0554180b74b896.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/ChatPreferences.kt
gpunto: Good point, added! (https://github.com/GetStream/stream-chat-android/pull/6493/changes/9b75a221b17abc450c06d21ea2e11816369f3514)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
gpunto: I think it makes sense to also introduce `RoleType` 👍  done https://github.com/GetStream/stream-chat-android/pull/6493/changes/9b75a221b17abc450c06d21ea2e11816369f3514.

andremion: Small follow-up on `RoleType`: the constructor is public, so a caller can build any value, like `RoleType("whatever")`, not just `User` / `Channel`. I think that is intentional and matches `ChatPreferenceToggle` and `PushPreferenceLevel`, where an unknown value coming from the server is kept instead of dropped, and `fromValue` maps the known ones. Just confirming the open construction is what you want, or would you rather limit creation to the constants plus `fromValue`?

gpunto: We could make the constructor internal, but with the current `fromValue` implementation, it would still be possible to construct `RoleType("anything goes")`. Do you think we should limit `fromValue` to only accept user + channel and return null for other strings? Maybe it would actually be good 🤔 we can always relax it later if needed. WDYT?

andremion: If we go for strictness, I believe `RoleType.User` and `RoleType.Channel` are enough, so we don't even need the `fromValue` function. Currently, there's no usage from the SDK, right?

gpunto: done!

PR #6494, opened 2026-06-09
Open message actions menu on poll option long-press

andremion:
### Goal

Long-press on a poll option now opens the message actions menu, the same as long-press anywhere else on the message row.

Resolves [AND-1220](https://linear.app/stream/issue/AND-1220).

### Implementation

`MessageContainer` exposes its long-press handler to descendants through a new `LocalMessageOnLongClick` CompositionLocal — provided when the message supports actions, `null` otherwise.

`PollOptionVotingRow` swaps the `toggleable` modifier for `combinedClickable`. The new `onLongClick` reads `LocalMessageOnLongClick.current` so TalkBack's double-tap-and-hold opens the message actions menu instead of being intercepted by the toggle. The `toggleable`'s "checked" / "not checked" announce is restored via an explicit `toggleableState` semantic.

### Testing

With TalkBack enabled:

1. Open a channel containing an open poll with at least one option.
2. Swipe right until focus lands on a poll option.
3. Double-tap to vote — the option toggles and TalkBack announces "checked" / "not checked".
4. Double-tap-and-hold on the same option — the message actions menu opens.
5. Repeat on a closed poll: options are no longer interactive, and long-press on the message bubble still opens the actions menu.

Without TalkBack:

1. Tap a poll option — vote toggles as before.
2. Long-press a poll option — actions menu opens.

## Release Notes

* **New Features**
  * Enhanced poll voting interactions with long-press support on poll options.

* **Improvements**
  * Improved gesture handling for poll option selection.
  * Better accessibility support for screen readers when voting on polls.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
@@ -41,7 +41,6 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableFloatStateOf
@@ -49,7 +48,6 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.setValue
-import androidx.compose.runtime.staticCompositionLocalOf
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clipToBounds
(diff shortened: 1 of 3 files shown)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/MessageContainer.kt
```diff
@@ -120,6 +122,18 @@ import kotlinx.coroutines.launch
...
+ * Forwards a long-press from an interactive descendant of the message bubble (poll option,
+ * attachment item, etc.) up to the message row's actions-menu handler. Provided by
+ * [MessageContainer] when long-press is supported for the current message; `null` otherwise.
+ *
+ * Compose's nested gesture handling means a child `toggleable` consumes the long-press as a tap
+ * before the row's `combinedClickable` can react. Children inside composite content read this
+ * local and pass it as `onLongClick` to their own `combinedClickable`, so the long-press routes
+ * to the actions menu regardless of which leaf the user holds.
+ */
+internal val LocalMessageOnLongClick = staticCompositionLocalOf<(() -> Unit)?> { null }
```
gpunto: Do we need this? `PollMessageContent` already receives `onLongItemClick` 🤔

andremion: Good catch. You're right.
Done in b0bddb73e2f4421fcb15f24a162f372926f94018

gpunto (approved):

PR #6495, opened 2026-06-10
Fix overly strict cid validation

gpunto:
### Goal

Align `validateCid` with iOS so the Android client accepts the same set of cids. The current regex (`^([a-zA-z0-9]|!|-)+:([a-zA-z0-9]|!|-)+$`) over-restricts valid characters (e.g. `_`, `.`) and has a typo: `a-zA-z` (lowercase `z` to uppercase `Z`) accidentally matches `[`, `\`, `]`, `^`, `_`, `` ` ``.

iOS [`ChannelId.init(cid:)`](https://github.com/GetStream/stream-chat-swift/blob/main/Sources/StreamChat/Models/ChannelId.swift#L27-L39) does not use a character-set regex. It splits on `:`, requires exactly 2 parts, and requires each part to be non-empty after stripping whitespace.

Closes [AND-1227](https://linear.app/stream/issue/AND-1227/align-cid-validation-with-ios)

### Implementation

- `Validation.kt`: replaced `cidPattern` with `cid.split(":")` and `parts.size == 2 && parts[0].isNotBlank() && parts[1].isNotBlank()`. The empty/blank pre-checks are no longer needed since they're subsumed by the split check.

### Testing

- `ValidationKtTest`:
  - Moved `"mess aging:123"` to the valid set (matches iOS, which permits internal spaces).
  - Added `"messaging:id_with_underscore"` and `"messaging:id.with.dots"` to the valid set (previously rejected by the regex).
  - Added `"   :   "`, `"messaging:   "`, and `"   :123"` to the invalid set (blank parts).
  - Updated `""` and `"   "` expectations to the unified format error message.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced channel ID validation logic to properly handle edge cases with whitespace and blank values, ensuring more reliable validation across all scenarios.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/internal/Validation.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/internal/Validation.kt
@@ -19,9 +19,6 @@ package io.getstream.chat.android.client.utils.internal
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import io.getstream.result.Error
 import io.getstream.result.Result
-import java.util.regex.Pattern
-
-private val cidPattern = Pattern.compile("^([a-zA-z0-9]|!|-)+:([a-zA-z0-9]|!|-)+$")
 
 /**
  * Validates a cid. Verifies it's not empty and in the format messaging:123.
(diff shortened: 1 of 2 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/internal/Validation.kt
```diff
@@ -33,9 +30,8 @@ private val cidPattern = Pattern.compile("^([a-zA-z0-9]|!|-)+:([a-zA-z0-9]|!|-)+
 @Throws(IllegalArgumentException::class)
 @InternalStreamChatApi
 public fun validateCid(cid: String): String = cid.apply {
```
gpunto, line 32: I considered aligning to the backend validation in [`ValidChannelCIDRe`](https://github.com/GetStream/chat/blob/efd2429785aaac1182ff6f7b509439f0226d0ab7/lib/combined/internal/validator/registry.go#L22), but decided against taht. If we encode the format too strictly, the backend can't change the cid format in the future without affecting released SDKs. Let me know if you think we should align with them instead.

gpunto:
@coderabbitai review

andremion (approved):

PR #6496, opened 2026-06-10
Replace reflection-based resolveDependency with typed variants

gpunto:
### Goal

Remove the reflection-based dispatcher from `ChatClient.resolveDependency` now that `StatePlugin` and `OfflinePlugin` live in `stream-chat-android-client`. The call site already statically knows whether it wants a `Plugin` or a `PluginFactory`, so the runtime `isSubclassOf` branch was dead weight.

Closes [AND-1003](https://linear.app/stream/issue/AND-1003/remove-reflection-based-instantiation-from-plugins)

### Implementation

- Deleted the `ChatClient.resolveDependency<DR, T>()` dispatcher and the `kotlin.reflect.full.isSubclassOf` call it relied on.
- Promoted the two underlying helpers to `public inline` with tighter type bounds: `resolvePluginDependency<P : Plugin, T>()` and `resolveFactoryDependency<F : PluginFactory, T>()`. Wrong-side calls now fail at compile time instead of at the old `error("Unsupported...")`.
- Updated the four call sites in `ChatClientStateExtensions.kt` and `internal/state/extensions/internal/ChatClient.kt` to pick the matching helper.

Note: `kotlin-reflect` is still needed by `NotificationHandlerFactory.primaryConstructor` so we can't remove it yet.

### Testing

- `DependencyResolverTest` updated to call the typed helpers; all existing cases (plugin not found, factory not found, dependency not resolved, init-state guard, plugin-cleared-during-resolution race) still pass.
- `:stream-chat-android-client:apiCheck`, `spotlessKotlinCheck`, `detekt` all green.
- Manual: exercise channel list, channel details, message list pagination, reply, and Giphy cancel in the sample app to confirm `state`, `globalState`, `chatClientConfig`, and `logic` all still resolve.

## Summary by CodeRabbit

* **API Changes**
  * Simplified dependency resolution by replacing the generic method with two specific, type-safe helpers for plugin factories and plugin dependencies.

* **Tests**
  * Updated tests to reflect new dependency resolution APIs.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -135,7 +135,6 @@ import io.getstream.chat.android.client.persistance.repository.factory.Repositor
 import io.getstream.chat.android.client.persistance.repository.noop.NoOpRepositoryFactory
 import io.getstream.chat.android.client.persistence.db.ChatClientDatabase
 import io.getstream.chat.android.client.persistence.repository.ChatClientRepository
-import io.getstream.chat.android.client.plugin.DependencyResolver
 import io.getstream.chat.android.client.plugin.MessageDeliveredPluginFactory
 import io.getstream.chat.android.client.plugin.Plugin
 import io.getstream.chat.android.client.plugin.factory.PluginFactory
@@ -261,7 +260,6 @@ import java.util.Calendar
 import java.util.Date
 import java.util.concurrent.atomic.AtomicReference
 import kotlin.coroutines.CoroutineContext
-import kotlin.reflect.full.isSubclassOf
 import kotlin.time.Duration.Companion.days
 
 /**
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -366,30 +364,14 @@ internal constructor(
...
-            else -> error("Unsupported dependency resolver: ${DR::class}")
-        }
-    }
-
-    @PublishedApi
-    @InternalStreamChatApi
-    @Throws(IllegalStateException::class)
-    @Suppress("ThrowsCount")
-    internal inline fun <reified F : DependencyResolver, reified T : Any> resolveFactoryDependency(): T {
+    public inline fun <reified F : PluginFactory, reified T : Any> resolveFactoryDependency(): T {
```
andremion: Can this be `internal` and remove the `@InternalStreamChatApi` usage?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -402,11 +384,17 @@ internal constructor(
...
+     * This method can't be called before user is connected because plugins are added only after user
+     * connection is completed.
+     *
+     * @throws IllegalStateException if the plugin was not added or the dependency is not found.
+     */
     @InternalStreamChatApi
     @Throws(IllegalStateException::class)
     @Suppress("ThrowsCount")
-    internal inline fun <reified P : DependencyResolver, reified T : Any> resolvePluginDependency(): T {
+    public inline fun <reified P : Plugin, reified T : Any> resolvePluginDependency(): T {
```
andremion: Can this be `internal` and remove the `@InternalStreamChatApi` usage?

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
gpunto: That's true 🤔 my bad, fixed!

PR #6497, opened 2026-06-11
Prevent tracking channel states for malformed CIDs

gpunto:
### Goal

Prevent the SDK from tracking `ChannelState`s by a malformed cid. The original symptom is in `SyncManager.updateActiveChannels`: it issues a `Filters.in("cid", missingCids)` query on reconnect, and a single malformed entry in the list fails the whole request, blocking re-watch for every valid channel in the same batch. The underlying problem is that `StateRegistry`/`LogicRegistry` happily store state keyed by an unvalidated `(channelType, channelId)` pair, so malformed cids can leak in through any code path that bypasses the existing `validateCidWithResult` gates (sync recovery, `channelFromMessage`, etc.).

Closes [AND-1187](https://linear.app/stream/issue/AND-1187/prevent-tracking-channelstates-by-malformed-cid).

### Implementation

- New internal `ChannelId` value class with `fromCid` / `fromTypeAndId` factories that return `null` on invalid input.
- `StateRegistry` and `LogicRegistry` re-key their channel maps on `ChannelId`. Their public `channel(type, id)` entry points validate at the boundary; malformed input returns a fresh, non-cached state/logic.
- `StateRegistry.getActiveChannelStates()` now returns `Map<ChannelId, ChannelState>`, so `SyncManager.updateActiveChannels` and `EventHandlerSequential` iterate validated keys directly.
- `QueryChannelsStateLogic.addChannelsState` filters out malformed server channels upfront so they never reach `setCids` / `setChannels` / `logicRegistry`.

### Testing

- New `ChannelIdTest` covers `fromCid` / `fromTypeAndId`.
- `StateRegistryTest` and `LogicRegistryTest`: malformed cid returns a non-cached state/logic and the registry maps stay empty.
- `QueryChannelsStateLogicTest`: `getActiveChannelState` returns null on malformed cid; `addChannelsState` drops malformed channels from `setCids` / `setChannels` / `logicRegistry`.

## Release Notes

* **Bug Fixes**
  * Improved handling of malformed channel identifiers to prevent sync failures and data inconsistencies.
  * Enhanced channel state recovery during reconnection to ensure all active channels are properly restored.

* **Improvements**
  * Optimized channel state management and user presence event dispatching for better reliability.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
@@ -34,7 +34,6 @@ import io.getstream.chat.android.models.querysort.QuerySorter
 import io.getstream.log.taggedLogger
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.async
-import kotlinx.coroutines.awaitAll
 
 @Suppress("TooManyFunctions")
 internal class QueryChannelsStateLogic(
@@ -172,7 +171,7 @@ internal class QueryChannelsStateLogic(
                     isChannelsStateUpdate = true,
                 )
             }
-        }.awaitAll()
+        }.forEach { it.await() }
     }
 
     private fun Channel.joinMessages(existingChannel: Channel?): Channel =
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
@@ -22,12 +22,11 @@ import io.getstream.chat.android.client.api.state.QueryChannelsState
 import io.getstream.chat.android.client.api.state.StateRegistry
 import io.getstream.chat.android.client.channel.state.ChannelState
 import io.getstream.chat.android.client.events.ChatEvent
-import io.getstream.chat.android.client.extensions.cidToTypeAndId
-import io.getstream.chat.android.client.extensions.internal.toCid
 import io.getstream.chat.android.client.extensions.internal.users
 import io.getstream.chat.android.client.internal.state.plugin.logic.internal.LogicRegistry
 import io.getstream.chat.android.client.internal.state.plugin.state.querychannels.internal.QueryChannelsMutableState
 import io.getstream.chat.android.client.query.QueryChannelsSpec
+import io.getstream.chat.android.client.utils.internal.ChannelId
 import io.getstream.chat.android.models.Channel
 import io.getstream.chat.android.models.FilterObject
 import io.getstream.chat.android.models.User
@@ -35,6 +34,7 @@ import io.getstream.chat.android.models.querysort.QuerySorter
 import io.getstream.log.taggedLogger
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
 
 @Suppress("TooManyFunctions")
 internal class QueryChannelsStateLogic(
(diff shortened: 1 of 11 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/internal/SyncManagerTest.kt
```diff
@@ -191,7 +193,7 @@ internal class SyncManagerTest {
             _syncEvents.test {
                 syncManager.onEvent(connectingEvent)
 
-                assertEquals(listOf(mockedChatEvent), awaitItem())
+                expectNoEvents()
```
aleksandar-apostolov, line 196: Curious about this one and the `@TestInstance(PER_CLASS)` removal — was `PER_CLASS` reusing one instance across tests and leaking `_syncEvents` emissions, so the old `assertEquals(listOf(mockedChatEvent), …)` was passing on stale state rather than on what `onEvent(connectingEvent)` actually emits?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
```diff
@@ -153,26 +153,26 @@ internal class QueryChannelsStateLogic(
...
+                logicRegistry.channel(id).updateDataForChannel(
                     channel = channel,
                     messageLimit = channel.messages.size,
                     isChannelsStateUpdate = true,
                 )
             }
-        }.map {
-            it.await()
-        }
+        }.awaitAll()
```
aleksandar-apostolov: Was the switch from `.map { it.await() }` to `.awaitAll()` intentional re: cancellation? `awaitAll()` cancels the remaining `Deferred`s if one throws, whereas the old form would let the others finish — is fail-fast the desired behavior here?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/internal/ChannelId.kt
```diff
@@ -0,0 +1,38 @@
...
+package io.getstream.chat.android.client.utils.internal
+
+/**
+ * A channel identifier whose underlying `type:id` cid has been validated.
+ */
+@JvmInline
+internal value class ChannelId private constructor(val cid: String) {
+
+    val type: String get() = cid.substringBefore(':')
+    val id: String get() = cid.substringAfter(':')
```
aleksandar-apostolov, line 26: If `validateCid` accepts a cid like `a:b:c`, this resolves to `type = "a"`, `id = "b:c"` — is that consistent with what `cidToTypeAndId()` does today?

aleksandar-apostolov (approved):
LGTM - Left some comments to double check before merging.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/internal/SyncManagerTest.kt
gpunto: Yep, exactly. The test function name says `when one event of exact same raw time of last sync arrive, it should not be propagated` but actually was asserting for events because they were leaking from previous tests.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
gpunto: So, actually I didn't know there was this difference. I applied a warning fix thinking they were equals, but they're not. Reverting

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/internal/ChannelId.kt
gpunto: `validateCid` splits by `:` and then validates that there are only two resulting parts, so we make sure we don't get into a case like `id = "b:c"`

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
gpunto: Changed in https://github.com/GetStream/stream-chat-android/pull/6497/changes/641344158df41827985662a58399e55c02deea30. I realized that we are not using the result of the `map` so I replaced it with `forEach` instead. This means AS will not show the warning anymore, so we don't make the same mistake again.

VelikovPetar (approved):

PR #6498, opened 2026-06-11
Clean up TalkBack accessibility semantics

andremion:
### Goal

The TalkBack sweep (AND-1180) left the Compose SDK's accessibility code uneven: a few parent-level overrides re-derive what their children already announce, and some empty merge-semantics blocks add noise. This consolidates the remaining cleanup into one pass, and folds in the verified media-grid traversal change and the QA reorder feedback on the Giphy preview, so we stop opening a PR per finding.

### Implementation

Four independent changes, one per commit:

- **AudioRecordingContent** (locked + overview rows): the parent `contentDescription` under `mergeDescendants` concatenated with the child descriptions (duration, progress), so any new child silently shifted the announce. The label now lives on a child leaf (the leading icon, the waveform) and the merge stays only as the single TalkBack focus stop.
- **Empty merge-semantics blocks**: removed the empty `Modifier.semantics(mergeDescendants = true) {}` on Rows that already have a `clickable` or `toggleable` (the merge boundary is implicit there), and kept plus commented the load-bearing ones (containers with no click handler, such as the deleted-message fallback).
- **Giphy preview**: the focused ephemeral preview announced "Only visible to you, [alt text], Giphy". It now announces as one node leading with "Giphy preview", then the alt text, then "Only visible to you" (composed via `clearAndSetSemantics`). Added the `stream_compose_giphy_preview_label` string in all supported locales.
- **Media attachment grid**: added `isTraversalGroup` so a TalkBack swipe walks every tile in order before leaving the grid.

No public API change (`apiCheck` passes). The changes are semantics and string only.

### Testing

Manual TalkBack testing on a device or emulator:

1. **Giphy preview**: open a channel and send a `/giphy <query>` command. With TalkBack on, focus the rendered preview. It should read as a single stop: "Giphy preview, [search query / alt text], Only visible to you" (leading with "Giphy preview"). Verified on-device: the merged accessibility node exposes `contentDescription="Giphy preview, <query>, Only visible to you"`.
2. **Voice recording**: start a voice recording in the composer and lock it, then stop to reach the overview. With TalkBack, the locked and overview rows each read as a single stop with the recording label plus duration (plus progress for the locked state).
3. **Media grid**: open a message with multiple image attachments. With TalkBack, once focus enters the grid, swiping walks every tile in order before moving past the grid.
4. **Poll switches, poll dialogs, quoted reply, deleted message**: confirm each still reads as a single focus stop (no regression from the merge-block cleanup).

## Summary by CodeRabbit

* **Improvements**
  * Enhanced accessibility for media attachments, Giphy previews, polls, and audio recording with improved TalkBack descriptions and traversal behavior.

* **Localization**
  * Added Giphy preview label translations across Spanish, French, Hindi, Indonesian, Italian, Japanese, and Korean locales.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
@@ -123,10 +123,7 @@ public fun GiphyMessageContent(
                 .applyIf(isTouchExplorationEnabled) {
                     focusRequester(previewFocusRequester).focusable()
                 }
-                // Announce the focused preview as a single TalkBack stop that leads with the
-                // "Giphy preview" label. mergeDescendants prepends it to the children's natural
-                // announcements (the only-visible banner, alt text, Giphy label), keeping their
-                // test tags and any integrator overrides intact.
+                // Lead the preview's merged TalkBack announcement with the "Giphy preview" label.
                 .semantics(mergeDescendants = true) { contentDescription = giphyPreviewLabel },
         ) {
             Row(
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
@@ -48,14 +48,13 @@ import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.semantics.clearAndSetSemantics
 import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.core.content.getSystemService
 import coil3.ColorImage
 import coil3.compose.LocalAsyncImagePreviewHandler
-import io.getstream.chat.android.client.utils.attachment.isGiphy
 import io.getstream.chat.android.compose.R
 import io.getstream.chat.android.compose.state.messages.attachments.AttachmentState
 import io.getstream.chat.android.compose.ui.components.button.StreamButtonStyleDefaults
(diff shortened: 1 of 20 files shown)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
```diff
@@ -120,7 +136,7 @@ public fun GiphyMessageContent(
                 .applyIf(isTouchExplorationEnabled) {
                     focusRequester(previewFocusRequester).focusable()
                 }
-                .semantics(mergeDescendants = true) {},
+                .clearAndSetSemantics { contentDescription = previewDescription },
```
gpunto: Not sure if we should worry about this, but if an integrator overrides `GiphyAttachmentContent` with custom semantics we are clearing that. I guess the recommendation in that case would be to override `MessageGiphyContent`, right?

gpunto (approved):
LG, just one question

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/GiphyMessageContent.kt
andremion: Right, that's a deliberate tradeoff. As per [notion issue](https://app.notion.com/p/stream-wiki/Giphy-preview-lacks-alt-text-only-visible-to-you-context-3676a5d7f9f6801da97cc5f37b115c6a?v=6a01d11916874a85908ab37c2b40afb2&source=copy_link)m, the preview has to announce "Giphy preview" first, then the alt text, then "Only visible to you", which is the reverse of the visual top-to-bottom order, so it can't be composed at the leaves.
To get that as a single focus stop, I compose the description on the container and clear the descendants, which does mean a custom GiphyAttachmentContent loses its semantics inside the preview. The alt text is read from the same `attachment.title` the default uses, so the announcement stays correct unless that value is changed.
So yes: to customize the preview's semantics, they need to override `MessageGiphyContent` rather than `GiphyAttachmentContent`.

andremion: I wrote a Kdoc note 5d65c001736275e7f5d762f6dac1adb3829c1919

andremion: Clearing the whole subtree was too heavy: it dropped your override case and also the content's test tag the e2e relies on. So I switched to just prepending a "Giphy preview" label on the container with `mergeDescendants`, which leaves the children (and any override) untouched.

eba3d54305bfd00c5349b6e25547820df94f04e4

andremion: I also trimmed the inline comments

2810befb74f3d598f70217bead052156875a5741

PR #6499, opened 2026-06-12
Disable composer during slow mode

andremion:
Resolves AND-1177.

### Goal

When slow mode is active on a channel (`coolDownTime > 0`), the whole composer should be disabled, not just the send action. This follows the design decision DS-035 ("Slow Mode Locks the Entire Composer"). Before this change, only the send button was replaced by the countdown. The input field and the attachment "+" button stayed interactive and the composer border kept its default colour, which contradicted the spec.

### Implementation

While `coolDownTime > 0`:

- The input field is disabled (`MessageComposerInputCenterContent`), and its placeholder shows the remaining seconds ("Slow mode, wait 9s…") via the new `stream_ui_message_composer_placeholder_slow_mode` string, with translations for all supported locales.
- The attachment "+" button stays visible but is disabled, including its border (`MessageComposerLeadingContent`). Its click action is removed from the accessibility tree while disabled.
- The composer border switches to `borderUtilityDisabled` and returns to `borderCoreDefault` once the timer reaches zero (`MessageInput`). The composer node uses `border/utility/disabled` here (confirmed with design), so no new colour token was added.
- The trailing countdown indicator is unchanged.

The composer re-enables automatically when the timer reaches zero, because every slot reads from `MessageComposerState.coolDownTime`. Factory overrides that only replace the trailing content keep working without changes.

Slow mode activation is also relaxed (`MessageComposerController`): it now triggers on `cooldown > 0` and the absence of the `skip-slow-mode` capability, instead of also requiring the `slow-mode` own-capability. The backend only grants `slow-mode` when `cooldown > 0` and the user cannot skip it, so this is equivalent today, but it keeps the client correct if the backend later tightens how that capability is granted, without needing a new client release. It also aligns the activation rule with the iOS and Flutter SDKs.

### 🎨 UI Changes

Verify the snapshots diff

### Testing

Manual steps:

1. Enable slow mode on a channel (set a cooldown, for example from the dashboard or with `channelClient.enableSlowMode(seconds)`).
2. Open the channel and send a message.
3. Confirm: the input field is greyed and not editable, the "+" button is greyed and not tappable, the composer border uses the lighter disabled colour, the placeholder reads "Slow mode, wait Ns…", and the trailing countdown shows the seconds.
4. Wait for the countdown to reach zero. The composer re-enables automatically (input editable, "+" tappable, border back to default) with no user action.

Automated:

- Compose: Robolectric interaction tests in `MessageComposerScreenTest` (input and "+" disabled during cooldown, slow-mode placeholder, both enabled without cooldown) and Paparazzi snapshots in `MessageComposerInputTest` and `MessageComposerTest`.
- Controller: unit tests in `MessageComposerControllerTest` assert the cooldown activates without the `slow-mode` capability and stays off when the user can skip slow mode.
- E2E: `SlowModeTests` in the compose sample (cooldown shown after sending, composer locked during cooldown), mirroring the iOS `SlowMode_Tests`. The e2e suite needs the emulator and the mock-server driver, so it runs in CI.

## Summary by CodeRabbit

* **New Features**
  * Added slow mode feature with cooldown indicator when message sending is restricted
  * Message composer input and attachments button disabled during cooldown period
  * Cooldown timer displays in placeholder text with countdown
  * Multi-language support for slow mode messaging

* **Tests**
  * Added E2E tests for slow mode behavior
  * Added unit tests for composer UI state during cooldown


andremion:
@coderabbit review

gpunto (approved):

PR #6500, opened 2026-06-12
Fix WaveformHandle crash on parent narrower than handle

gpunto:
### Goal

`WaveformSlider`'s `WaveformHandle` crashes with `IllegalArgumentException: Cannot coerce value to an empty range` when the parent is narrower than `handleSize - handleBorderSize` (14dp - 2dp = 12dp). The current `left.coerceIn(-handleBorderSize, parentWidth - handleSize)` produces a max bound below the min bound in that case.

Closes [AND-1232](https://linear.app/stream/issue/AND-1232/fix-waveformhandle-crash-when-parent-is-narrower-than-the-handle)

### Implementation

- `WaveformSlider.kt`: clamp the max offset to be at least `-handleBorderSize` before calling `coerceIn`, so the range is always non-empty. The handle stays pinned to the start when the parent is too small to lay it out.

### Testing

- Manual: a Compose preview / waveform sized below 12dp no longer crashes.

Tested on a custom reproduction screen. Both the examples crashed before the fix:

https://github.com/user-attachments/assets/395102ea-fa0d-49b8-94f4-e55ea6eaf5bb

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced waveform audio slider functionality. The audio slider thumb now correctly constrains and positions itself in compact layouts and smaller containers. This fixes edge cases where the thumb positioning could become unreliable when the parent container dimensions were constrained, ensuring smooth and predictable slider behavior across various layout sizes.


aleksandar-apostolov (approved):
LGTM

PR #6501, opened 2026-06-12
Render enhanced mentions in incoming messages

gpunto:
### Goal

Render `@channel`, `@here`, role, and user-group tokens in incoming messages so users see them styled and clickable, like user mentions. No composer autocomplete yet.

This is PR 3 in the enhanced-mentions series.

### Implementation

- `ui-common`: `Mention` sealed interface gains `Channel`, `Here`, `Role`, `Group` cases; new `mentionRegex(display)` helper uses Unicode-aware boundaries so non-ASCII display names match correctly.
- Compose: `TextUtils` spans every mention type with a per-type annotation tag; `StreamDesign` exposes `chatTextMention*` / `chatBgMention*` color tokens; `MessageText` / `MessageContent` / `MessageContainer` / `MessageItem` / `MessageList` plumb `onMentionClick: (Mention) -> Unit`.
- ui-components: `Linkify`, `MentionSpan` (renamed from `UserSpan`), `TextViewLinkHandler`, and `LongClickFriendlyLinkMovementMethod` now span every mention type and dispatch through a new `MessageListView.OnMentionTokenClickListener`. The legacy user-only `OnMentionClickListener` stays in place and is deprecated.

### UI Changes

| Before | After |
| ------ | ----- |
| Screenshot_20260615_122835 | Screenshot_20260615_122524 |

### Testing

- `MentionRegexTest` covers Unicode boundaries, escape behavior, and the start/end-of-string cases.
- `TextUtilsKtTest` and `MessageTextHelpersTest` assert that each mention type produces the right annotation/span.
- API dump regenerated for `stream-chat-android-compose`, `stream-chat-android-ui-common`, and `stream-chat-android-ui-components`.

## Release Notes

* **New Features**
  * Added support for channel, group, role, and here mention types in messages alongside user mentions.
  * Enhanced mention interaction with improved handling and selection callbacks.

* **Refactor**
  * Refactored mention system infrastructure across message display and composition components for improved handling of all mention types.


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
```diff
@@ -809,9 +827,15 @@ public data class MessageContentParams(
...
     val onQuotedMessageClick: (Message) -> Unit = {},
+    @Deprecated(
+        message = "Use onMentionClick; it also fires for other mention types.",
+        replaceWith = ReplaceWith("onMentionClick"),
+        level = DeprecationLevel.WARNING,
+    )
     val onUserMentionClick: (User) -> Unit = {},
     val onMediaGalleryPreviewResult: (MediaGalleryPreviewResult?) -> Unit = {},
     val onLinkClick: ((Message, String) -> Unit)? = null,
+    val onMentionClick: (Mention) -> Unit = {},
```
andremion, line 838: `onMentionClick` is added to the params and composables, but `ChatComponentFactory` never forwards to it. The factory passes only `onUserMentionClick`.
Is that intentionally postponed for the follow-up PR? Are you planing blocking the new release until we get this feature fully merged?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/TextUtils.kt
```diff
@@ -215,33 +291,27 @@ private fun AnnotatedString.Builder.linkify(
     }
 }
 
-private fun AnnotatedString.Builder.tagUser(
+/**
+ * Tags every word-bounded `@<token>` occurrence in [text] with [annotationTag] and styles it
+ * with [mentionsColor].
```
andremion: minor: KDoc references `[mentionsColor]` and `[annotationTag]`, but the params are `text` and `mention`.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/mention/Mention.kt
```diff
@@ -58,12 +78,48 @@ public interface Mention {
...
+        override val type: MentionType = MentionType.role
+        override val display: String = role
+    }
+
+    /**
+     * A user-group mention.
+     *
+     * @param group The group being mentioned.
+     */
+    public data class Group(public val group: io.getstream.chat.android.models.UserGroup) : Mention {
```
andremion: minor: Can you import `io.getstream.chat.android.models.UserGroup`?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
```diff
@@ -1004,6 +1051,50 @@ public data class MessageComposerUserSuggestionItemParams(
     val onUserSelected: (User) -> Unit,
 )
 
+/**
+ * Parameters for [ChatComponentFactory.MessageComposerSuggestionItem].
+ *
+ * @param mention The non-user [Mention] for which the suggestion is rendered.
+ * @param onMentionSelected Action invoked when the mention is selected.
+ */
+public data class MessageComposerSuggestionItemParams(
```
andremion: These new classes are unused.
Are they for the follow-up PR?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
gpunto: Good catch, I made a mistake when splitting the changes in different branches. 😅 Fixed it in https://github.com/GetStream/stream-chat-android/pull/6501/changes/a226576322413f2edac81f48129bb479ce3f19fb

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/TextUtils.kt
gpunto: Fixed!

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
gpunto: Same as https://github.com/GetStream/stream-chat-android/pull/6501#discussion_r3420162587. I moved them to pr4. Thanks!

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/mention/Mention.kt
```diff
@@ -58,12 +80,48 @@ public interface Mention {
...
+    }
+
+    /**
+     * A user-group mention.
+     *
+     * @param group The group being mentioned.
+     */
+    public data class Group(public val group: UserGroup) : Mention {
+        override val type: MentionType = MentionType.group
+        override val display: String = group.name
```
aleksandar-apostolov: The spec example in the Notion doc shows:

```
"text": "Hey @here, @backendsupport, and @admin, check this out!"
"mentioned_group_ids": ["backendsupport"]
```

`backendsupport` is what's in `mentioned_group_ids`, so it reads as the group **ID**, not the display name (which in the spec is `"Backend Support Team"`). With `display = group.name` here, the regex will miss the match whenever `name ≠ id` (e.g., after a rename).

Two questions worth resolving before this API locks:

1. What does the backend actually put in `message.text` — `@<group.id>` or `@<group.name>`?
2. On rename, do stored messages get the token rewritten, or do they silently lose the highlight?

Depending on the answer, this might need to either flip to `group.id`, or accept both as candidates.

aleksandar-apostolov: FYI React hedged on this exact ambiguity — their `getMentionDisplayTexts` returns *both* `name` and `id` as candidate display tokens for `user_group` (and `role`), so the regex matches `@<id>` OR `@<name>` in `message.text`:

So if backend can emit either form (or we just don't want to find out the hard way), matching both is the safer move and gives us parity with React without committing to one or the other.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/mention/MentionRegex.kt
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * The boundaries use Unicode-aware lookaround, because Java's default word-boundary classes only
+ * recognize ASCII `[A-Za-z0-9_]`. The lookbehind `(?<![\p{L}\p{N}_])` also naturally handles the
+ * start-of-string case.
+ *
+ * @param display The literal display text of the mention (without the leading `@`).
+ */
+@InternalStreamChatApi
+public fun mentionRegex(display: String): Regex =
+    Regex("(?<![\\p{L}\\p{N}_])@${Regex.escape(display)}(?![\\p{L}\\p{N}_])")
```
aleksandar-apostolov: With `display = ""` this collapses to `(?<![\p{L}\p{N}_])@(?![\p{L}\p{N}_])` and matches every bare `@`. Today `Linkify` and `TextUtils` each guard against this — push it down so it can't be misused, and add a `MentionRegexTest` case for empty display.

```suggestion
public fun mentionRegex(display: String): Regex =
    if (display.isEmpty()) Regex("(?!)")
    else Regex("(?<![\\p{L}\\p{N}_])@${Regex.escape(display)}(?![\\p{L}\\p{N}_])")
```

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/mention/Mention.kt
gpunto: > 1. What does the backend actually put in message.text — @<group.id> or @<group.name>?
> 2. On rename, do stored messages get the token rewritten, or do they silently lose the highlight?

1. What's in the message text is on us. Following how user mentions work in production, we are going to insert the name if not empty, and the id otherwise.
2. No rewrite, so we'd lose the highlight like for user mentions. I raised this concern as part of [this Slack message](https://getstream.slack.com/archives/C07EV8DFMRN/p1780309335792609) and decided to keep it as is since we had no complaints for user mentions.

In any case, I agree it's better if we do the matching on both id and name at rendering time, so we handle both formats. Done in https://github.com/GetStream/stream-chat-android/pull/6501/changes/0a7370e5272d1e718026a0aa94460a3b9807ecaa.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/mention/MentionRegex.kt
gpunto: Addressed in https://github.com/GetStream/stream-chat-android/pull/6501/changes/0a7370e5272d1e718026a0aa94460a3b9807ecaa

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
andremion: Thanks. The four classes are gone, but two related bits are still here: `MessageComposerParams.onMentionSelected`, which has no consumer yet, and the `@Deprecated` on `onUserSelected` that points to it. So `onUserSelected` is now deprecated in favor of something that does nothing in this PR. Should you move those to PR4 too, so the deprecation lands with a working replacement?

andremion (approved):

gpunto:
Just made a draft to prevent accidental merging right now. We're going to do a release today and I prefer **not** to include this PR.

aleksandar-apostolov (approved):
LGTM

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactoryParams.kt
gpunto: True! Moved to PR4 in https://github.com/GetStream/stream-chat-android/pull/6501/changes/01be104b631f2099a84a7ba5b136e77918a51cd9

PR #6502, opened 2026-06-16
Enforce poll maxVotesAllowed on the XML voting side

gpunto:
### Goal

The XML UI SDK never checks `Poll.maxVotesAllowed` when the user taps an option to vote. Both the inline poll on the message (`PollView`) and the "see all options" bottom sheet (`AllPollOptionsDialogFragment`) call `ChatClient.castPollVote` unconditionally, so a user can vote past the configured limit.

Closes [AND-1237](https://linear.app/stream/issue/AND-1237/xmlpolls-maxvotesallowed-is-not-enforced-on-the-voting-side)

### Implementation

- New internal `Poll.canCastVote()` extension in `ui-common`.
- XML: both voting sites (`PollView`, `AllPollOptionsDialogFragment`) skip the cast when `canCastVote()` is false. Un-voting always goes through.
- Compose: replaced `PollOptionVotingRow`'s `checkedCount` parameter with a direct `poll.canCastVote()` call. Both callers passed `poll.ownVotes.size`, so no behaviour change.

### Testing

- `PollExtensionsTest`: three new cases cover `canCastVote()` for `maxVotesAllowed = null`, `ownVotes.size < maxVotesAllowed`, and `ownVotes.size == maxVotesAllowed`.
- Manual on the sample app: created polls with `maxVotesAllowed = 1`, `2`, and unlimited in both XML and Compose. Verified that tapping a third option in a 2-vote poll is now a no-op (XML inline + "see all options" sheet, and Compose).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vote casting validation in polls to consistently enforce vote limits across the app.

* **Tests**
  * Added unit tests for poll vote limit validation behavior.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -2368,6 +2368,9 @@ internal constructor(
     /**
      * Cast a vote for a poll in a message.
      *
+     * Note: This function does not validate that the vote can be cast (e.g. that the poll is open or that
+     * the user is still within the [Poll.maxVotesAllowed] limit). Callers are responsible for these checks.
+     *
      * @param messageId The message id where the poll is.
      * @param pollId The poll id.
      * @param optionId The id of the option to vote for.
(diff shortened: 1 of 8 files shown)


gpunto:
@coderabbitai review

andremion (approved):
LGTM.
Just a small optional comment:
The guard sits at the view layer. A direct caller or a custom poll view can still vote past `maxVotesAllowed`. This matches how Compose already works, so it's not a regression, and the scope here is fine.
Would it be worth a short KDoc saying the caller owns the limit check, so integrators using the public API directly aren't surprised? Not sure where exactly to put it.
Happy either way.

gpunto:
> LGTM. Just a small optional comment: The guard sits at the view layer. A direct caller or a custom poll view can still vote past `maxVotesAllowed`. This matches how Compose already works, so it's not a regression, and the scope here is fine. Would it be worth a short KDoc saying the caller owns the limit check, so integrators using the public API directly aren't surprised? Not sure where exactly to put it. Happy either way.

Makes sense! Just added a note to `castPollVote`'s KDoc, which is what integrators would use if they don't go through our components.

PR #6503, opened 2026-06-17
Add composer autocomplete for enhanced mentions in Compose SDK

gpunto:
### Goal

Offer `@channel`, `@here`, role, and user-group suggestions in the Compose composer alongside user mentions, gated by channel capabilities.

This is PR 4 in the enhanced-mentions series.

### Implementation

- `@channel`, `@here`, role, and group suggestions are gated on `NOTIFY_CHANNEL`, `NOTIFY_HERE`, `NOTIFY_ROLE`, and `NOTIFY_GROUP` respectively. User suggestions stay ungated: on Permissions V1 regular members lack `CREATE_MENTION` but can still mention users, since the server only enforces it on send under V2.
- `MentionLookupHandler` aggregates users, roles, and groups for a given query.
- Compose: `UserSuggestionItem` becomes `MentionSuggestionItem`, rendering every mention type; exposed as a `ChatComponentFactory` slot.

### Behavior change

Default mention suggestions now use word-prefix matching and alphabetical sorting (via `DefaultUserQueryFilter`), aligned with the other SDKs, instead of substring matching sorted by match position. Integrators relying on the previous user-mention ordering will see different results.

### UI Changes

| Before | After |
| ------ | ----- |
| Screenshot_20260617_125735       |     Screenshot_20260617_125618  |

### Testing

- `MentionLookupHandlerTest` covers capability gating, ordering, query matching, and empty paths.
- `MessageComposerControllerTest` covers the new `filterMentions` survival rules across every mention type.
- `MessageComposerViewModelTest` exercises the suggestion-list flow end to end.
- Paparazzi: new `MentionSuggestionListTest` baseline; refreshed `MessageComposerTest` snapshots that now render the broader suggestion item.

Manual (on a channel with the new capabilities granted, e.g. any messaging channel in the demo app):

1. Open the Compose sample and tap the composer.
2. Type `@` and confirm `@channel` and `@here` appear at the top.
3. Type a few more characters and confirm matching roles, groups, and users show up.
4. Pick one of each type and confirm the inserted token survives `Send`.

## Summary by CodeRabbit

* **New Features**
  * Message composer mention suggestions now support `@channel`, `@here`, role-based, and group mentions alongside user mentions.

* **Documentation**
  * Added localized UI labels for mention suggestion types in Spanish, French, Hindi, Indonesian, Italian, Japanese, and Korean.


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
```diff
@@ -124,14 +129,15 @@ public fun MessageComposer(
     onCancelAction: () -> Unit = { viewModel.dismissMessageActions() },
     onLinkPreviewClick: ((LinkPreview) -> Unit)? = null,
     onCancelLinkPreviewClick: (() -> Unit)? = { viewModel.cancelLinkPreview() },
-    onUserSelected: (User) -> Unit = { viewModel.selectMention(it) },
+    onUserSelected: (User) -> Unit = {},
```
andremion, line 134: The default for `onUserSelected` moved from `{ viewModel.selectMention(it) }` to `{}`, and `onMentionSelected` now defaults to `viewModel::selectMention`. For a caller who passes a custom `onUserSelected` but no `onMentionSelected`, the default `selectMention` still runs, so for user taps both their handler and `selectMention` fire. With `selectedMentions` being a `Set` and the text insert being idempotent this is harmless in the common case, but a caller who overrode `onUserSelected` to replace or suppress the default selection would now get the mention inserted anyway. Is that intended? Might be worth a note for integrators who customized `onUserSelected`.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/suggestions/UserSuggestionList.kt
```diff
@@ -19,60 +19,71 @@ package io.getstream.chat.android.compose.ui.messages.composer.internal.suggesti
...
             )
         }
     }
 }
 
+private fun mentionKey(index: Int, mention: Mention): String = when (mention) {
+    is Mention.User -> "user:${mention.user.id}"
+    is Mention.Channel -> "channel"
+    is Mention.Here -> "here"
+    is Mention.Role -> "role:${mention.role}"
```
andremion, line 70: `itemsIndexed` throws if two items produce the same key. Roles key on name (`role:<name>`), so if `searchRoles` ever returned two entries with the same name the list would crash. Role names are unique server-side so this is low risk, but a `distinct()` on the role results (or folding the index into the key) would remove the chance entirely. Worth guarding?

andremion (commented):
Good job!

One small heads-up: DefaultUserQueryFilter now does word-prefix matching and sorts alphabetically, instead of substring matching sorted by match position. That seems intentional and matches the spec, so no change needed, but it does change existing user-mention results, so maybe we should call it out somewhere for integrators.

Also, left two small comments.

andremion (approved):

andremion (commented):
I just noticed that this user gate looks risky for Permissions V1 apps.
By default on V1, regular members don't get `create-mention` in `own_capabilities` (only admins do). But the user mentions work fine there because the server enforces `create-mention` on send only under V2. So gating user suggestions on it here would hide all user-mention suggestions for regular users on every V1 app, which is a regression from the current always-show behavior.
The iOS is keeping user suggestions ungated for exactly this reason.

Maybe we could drop the `create-mention` gate for users (matching the other SDKs), or only apply it when we know the channel is on V2?
The @channel/@here/role/group gates are fine; the issue might be with the user gate.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/internal/suggestions/UserSuggestionList.kt
gpunto: Done in https://github.com/GetStream/stream-chat-android/pull/6503/changes/a3ec49bb61252215b8ce6223b4c8bceac86933e0. We now only show distinct roles, as showing duplicates would give us no benefit.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
gpunto: Added a note to the KDoc in https://github.com/GetStream/stream-chat-android/pull/6503/changes/a3ec49bb61252215b8ce6223b4c8bceac86933e0. Would you put it also somewhere else?

gpunto:
> One small heads-up: DefaultUserQueryFilter now does word-prefix matching and sorts alphabetically, instead of substring matching sorted by match position. That seems intentional and matches the spec, so no change needed, but it does change existing user-mention results, so maybe we should call it out somewhere for integrators.

Any suggestions as to where to call this out? Do you think adding it to `DefaultUserQueryFilter` would be enough?

> I just noticed that this user gate looks risky for Permissions V1 apps.

Great catch, I didn't know about this. I checked with the other folks working on this feature and for now I'm removing the gate so we preserve behavior for user mentions. We can always introduce it later if it becomes safe to do so.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/composer/MessageComposer.kt
andremion: I believe the KDoc is the right place and should be enough.

andremion:
> Any suggestions as to where to call this out? Do you think adding it to `DefaultUserQueryFilter` would be enough?

The KDoc on DefaultUserQueryFilter is a good spot, and it already describes the new matching and sorting well.
I would keep it as a description of the current behavior rather than a "changed from X" note.
The one thing the KDoc does not cover is upgrade visibility, since integrators will not necessarily read it. We do not keep a manual changelog, and the release notes are generated from the PR, so I would add one line to the PR description, something like: "Behavior change: default mention suggestions now use word-prefix matching and alphabetical sorting, aligned with the other SDKs." That way, it is reachable from the release notes.

andremion (approved):

PR #6505, opened 2026-06-18
client: Count thread replies toward the slow mode cooldown

andremion:
### Goal

When slow mode is active, sending a thread-only reply did not start the composer cooldown. The backend cooldown is channel-wide (per app, channel, user) and counts thread replies, so the user replied in a thread, saw no countdown, sent again, and the server rejected the message. iOS already counts thread replies; this brings Android (both Compose and XML) to the same behavior. Reported by the Flutter team.

Linear: AND-1230

### Implementation

The composer cooldown is driven only by `ChannelState.lastSentMessageDate`, whose only consumer is the cooldown observer in the shared `MessageComposerController` (used by both the Compose and XML composers). That observer is not gated by message mode, so once `lastSentMessageDate` reflects thread replies, the countdown works in both channel mode and thread mode with no composer change.

`lastSentMessageDate` was derived from the channel message list filtered to the current user. A thread-only reply (`parentId != null && showInChannel == false`) never contributed, for different reasons in the two channel-state implementations:

- New `ChannelStateImpl` (default): `shouldIgnoreUpsertion` drops thread-only replies at every write entry point, so they never enter the message list.
- Legacy `ChannelStateLegacyImpl`: they enter the raw map but are filtered out before `lastSentMessageDate` reads the visible list.

This change keeps the existing channel-message derivation untouched and adds a thread-reply contribution, then exposes `lastSentMessageDate` as the later of the two. The contribution is a small per-channel value updated at the message write entry points with the current user's most recent thread-only reply date, and it only moves forward. The optimistic send path feeds it, so the countdown starts on send. The same change is applied to both state implementations, so behavior is consistent whether `useLegacyChannelLogic` is `false` (default) or `true`.

The visible message list, `shouldIgnoreUpsertion`, and `lastMessageAt` are unchanged, so no other consumer is affected. There is no public API change.

Cross-session resume needs no extra work: on channel init the offline load reads messages (including thread replies) and feeds the same value, which only moves forward, so it survives the later server refresh.

Parity note: this matches iOS, where the cooldown comes from `channel.lastMessageFromCurrentUser` (which counts thread replies) via `currentCooldownTime()`, used by both the UIKit and SwiftUI composers.

### Testing

Manual steps (the Compose and XML sample apps behave the same):

1. Enable slow mode on a channel from the dashboard (set a cooldown, for example 10s).
2. Open the channel as a member without the skip-slow-mode capability.
3. Open a thread on any message and send a reply.
4. Expected: the composer shows the countdown and the send button is disabled until it reaches zero, the same as sending a message in the channel.
5. While the countdown is active, go back to the channel. Expected: the channel composer also shows the remaining countdown.
6. Open a different thread in the same channel. Expected: it also shows the remaining countdown.
7. Force-stop and reopen the app while a countdown is active. Expected: the countdown resumes.

Automated: unit tests in `ChannelStateImplLastSentMessageDateTest` and the additions to `ChannelStateLegacyImplTest` cover a thread reply updating `lastSentMessageDate` (and not updating it for another user's reply), in both state implementations, plus the channel-wide and cross-refresh cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tracking of the current user's most recent message timestamp to now properly include thread-only replies, ensuring the last sent message date reflects both channel messages and threaded conversations.

* **Tests**
  * Added comprehensive test coverage for message timestamp tracking with thread replies.


andremion:
@coderabbit review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -99,6 +99,12 @@ internal class ChannelStateImpl(
     private val _quotedMessagesMap = MutableStateFlow<Map<String, Set<String>>>(emptyMap())
     private val _messages = MutableStateFlow<List<Message>>(emptyList())
     private val localOnlyMessages = MutableStateFlow<List<Message>>(emptyList())
+
+    /**
+     * Tracks the creation date of the current user's most recent thread-only reply. Thread-only
+     * replies are excluded from [_messages], so the cooldown derivation tracks them here separately.
+     */
+    private val _lastSentThreadReplyDate = MutableStateFlow<Date?>(null)
```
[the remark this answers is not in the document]
andremion: 232aad04cd674142a4e1369fe6e559016e1ffdf6

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -1500,6 +1528,7 @@ internal class ChannelStateImpl(
         _repliedMessage.value = null
         _quotedMessagesMap.value = emptyMap()
         _messages.value = emptyList()
+        _lastSentThreadReplyDate.value = null
```
gpunto, line 1531: Should this be done in the legacy impl too?

gpunto (approved):
LG, just one small divergence with the legacy implementation

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
andremion: It's safe, but it would not add value in the legacy impl, so I left it out.

`ChannelStateImpl.destroy()` resets its message values (`_messages.value = emptyList()`, and so on), so I reset `_lastSentThreadReplyDate` there to keep it consistent.

`ChannelStateLegacyImpl.destroy()` does not reset values; it only sets the field references to null. I checked: after `destroy()`, `lastSentMessageDate` still returns the retained channel message date. So nulling the thread date there would not clear the cooldown, it would only change which retained date is returned. It also does not matter either way, because a destroyed instance is removed from the registry (`clear()` and `removeChanel()`) and is never reused.

PR #6506, opened 2026-06-18
XML: render dark/light modes in a single paparazzi snapshot instance.

kanake10:
### Goal
Dark and light mode snapshots in `PaparazziViewTest` were not rendering correctly. Both halves of the combined snapshot appeared identical because `createConfigurationContext` is ignored by Paparazzi, so the night mode flag had no effect.

### Implementation

Replace `createConfigurationContext` with `paparazzi.unsafeUpdateConfig`, passing a `DeviceConfig` with the correct `NightMode` before each view is inflated. 

### Testing

<details>

```
Provide the patch code here
```

</details>

#### General
- [ ] Thread with the PR link started in a respective Slack channel (required internally)

#### Code & documentation

### ☑️Reviewer Checklist

### 🎉 GIF

## Summary by CodeRabbit

* **Tests**
  * Improved snapshot testing efficiency by generating light and dark mode snapshots within a single test instance. Night-mode configuration now updates at runtime, reducing test redundancy and streamlining testing workflows.


andremion (commented):
Thanks for working on this. The approach is right: `createConfigurationContext` is ignored by Paparazzi, so calling `unsafeUpdateConfig` before each inflation is the correct way to do it, and the regenerated header golden confirms light and dark now render together.

One blocking issue: the device config is set in two places that can differ, so one test renders on the wrong device. I left the details inline. After the fix, the affected goldens need re-recording. Nothing else is required to merge.

stream-chat-android-ui-components/src/test/kotlin/io/getstream/chat/android/ui/PaparazziViewTest.kt
```diff
@@ -42,23 +44,11 @@ internal abstract class PaparazziViewTest : MockedChatClientTest {
...
-    private val darkContext get() = paparazzi.context
-    // TODO Figure out how to properly support dark/light modes in a single paparazzi snapshot instance.
-    /*.createConfigurationContext(Configuration().apply {
-        uiMode = Configuration.UI_MODE_NIGHT_YES
-    })*/
-
     @get:Rule
     val instantAnimations = InstantAnimationsRule()
 
+    open val baseDeviceConfig: DeviceConfig = DeviceConfig.PIXEL_4A
```
andremion: `baseDeviceConfig` defaults to PIXEL_4A, but `ChannelListHeaderViewTest` builds its rule with `Paparazzi(deviceConfig = DeviceConfig.PIXEL_2)` and does not override `baseDeviceConfig`. So the first `applyNightMode` call replaces PIXEL_2 with PIXEL_4A, and the header is snapshotted on the wrong device. You can see it in the regenerated golden: the header image went from 562x1000 (PIXEL_2 ratio) to 461x1000 (PIXEL_4A ratio).

The cause is that the device config is now defined in two places that can differ: the `Paparazzi` rule and `baseDeviceConfig`. Could we define the device config once and reuse it in the rule, so the two stay the same? Something like:

Base:
```kotlin
abstract val deviceConfig: DeviceConfig

private fun applyNightMode(isDark: Boolean) {
    paparazzi.unsafeUpdateConfig(
        deviceConfig = deviceConfig.copy(
            nightMode = if (isDark) NightMode.NIGHT else NightMode.NOTNIGHT,
        ),
    )
}
```

Subclass (declare `deviceConfig` before `paparazzi` so the rule reads the set value):
```kotlin
override val deviceConfig = DeviceConfig.PIXEL_2

@get:Rule
override val paparazzi = Paparazzi(deviceConfig = deviceConfig)
```

This keeps each subclass in control of its own `Paparazzi(...)` call (so a test that also passes `renderingMode` still works), removes the default device that can be inherited by mistake, and keeps one definition of the device. The header goldens will need re-recording back to PIXEL_2 after the change (`./gradlew :stream-chat-android-ui-components:recordPaparazziDebug`).

stream-chat-android-ui-components/src/test/kotlin/io/getstream/chat/android/ui/PaparazziViewTest.kt
```diff
@@ -86,9 +76,13 @@ internal abstract class PaparazziViewTest : MockedChatClientTest {
...
-        val lightView = viewFactory(lightContext)
-        // TODO Figure out how to properly support dark/light modes in a single paparazzi snapshot instance.
-        // val darkView = viewFactory(darkContext)
+        applyNightMode(false)
+        val lightView = viewFactory(paparazzi.context)
+
+        applyNightMode(true)
+        val darkView = viewFactory(paparazzi.context)
+
+        applyNightMode(false)
```
andremion, line 86: Optional: a one line comment here would help. Resetting to light before building the container is intentional (so the container and final snapshot render in light mode after the dark view was inflated), but it reads as redundant without a note.

andremion:
@coderabbitai review

stream-chat-android-ui-components/src/test/kotlin/io/getstream/chat/android/ui/PaparazziViewTest.kt
kanake10: done !!

stream-chat-android-ui-components/src/test/kotlin/io/getstream/chat/android/ui/PaparazziViewTest.kt
kanake10: Good catch 12ad61f4cdc8bacbac7cccb2b61c9d5075ff4429

andremion: Sorry, the `@get:Rule` on the override in my snippet is what broke CI. It changed the JUnit rule order so `InstantAnimationsRule` runs before Paparazzi sets up layoutlib, which is the `UnsatisfiedLinkError` in all 10 tests. Unit tests passed on `cec8609` and only broke on `12ad61f`, and that annotation is the one rule-related change.

Fix: keep the `deviceConfig` change, but drop `@get:Rule` from the override (and the now-unused `import org.junit.Rule`). Leave it only on the base `abstract val paparazzi`, which is the wiring that passed before.

```kotlin
override val deviceConfig = DeviceConfig.PIXEL_2
override val paparazzi = Paparazzi(deviceConfig = deviceConfig)
```

Same for `ChannelListViewTest`.

kanake10: updated fa07259ece4a5d54197df5fc1386fb4cf114e261

kanake10: I also think the  PR CI check is failing since this contribution comes from a fork, I suspect it may be related to GitHub's permission restrictions on forked PRs.

andremion: That's fine. But we can check in the logs what's actually failing in that check:
```
body: '{"body":"<!-- pr-quality-anchor -->\\n### PR checklist ❌\\n\\nThe following issues were detected:\\n\\n\\n- **Linked issue missing**. Add a Linear ticket reference (e.g. `AND-123`, `Closes AND-123`, or a https://linear.app/... link) or a GitHub issue (`Closes #123`) to the PR description.\\n\\n**What we check**\\n1. Title is concise (5–18 words) unless labeled `pr:ignore-for-release`.\\n2. At least one `pr:` label exists (e.g., `pr:bug`, `pr:new-feature`).\\n3. Sections `### Goal`, `### Implementation`, and `### Testing` contain content. Bot-authored PRs are exempt.\\n4. PR description references an issue (Linear ticket like `AND-123`, a Linear URL, or a GitHub `Closes #N`). Bot-authored PRs are exempt."}',
```

External contributors don't need to ticket reference

andremion (approved):

stream-chat-android-ui-components/src/test/kotlin/io/getstream/chat/android/ui/PaparazziViewTest.kt
kanake10: Could this be the root cause:

In build-conventions , the “Linked issue” validation is currently implemented as:
https://github.com/GetStream/stream-build-conventions-android/blob/25ee4c2fc921410d61475c0da18a94defb175845/.github/workflows/pr-quality.yml#L95-L104
From this, the check only excludes `is_bot_author`, meaning it still applies to all non-bot PRs  including external contributors from forks.

To align the implementation with the intended policy, the condition could be extended to also exclude forked PRs (or external contributors), for example:

```bash
IS_FORK_PR="${{ github.event.pull_request.head.repo.fork }}"

if ! is_bot_author && [ "$IS_FORK_PR" = "false" ]; then
```

PR #6509, opened 2026-06-19
Reduce redundant DB writes for users and channel configs

gpunto:
### Goal

Persisting a channel response fans out into redundant SQLite writes: the same user gets inserted once per reference in the payload, and the channel list rewrites each channel's config once per channel. A customer flagged the volume. This dedups before the DB write, with no behaviour change.

Closes [AND-1247](https://linear.app/stream/issue/AND-1247/reduce-redundant-db-writes-for-users-and-channel-configs)

### Implementation

- `insertUsers`: dedup users by id (`Channel.users()` lists the same user once per member/author/watcher/etc.).
- `insertChannelConfigs`: dedup configs by type (one row per type, not per channel).

Both keep the last occurrence, matching the existing `INSERT OR REPLACE` behaviour.

### Testing

- New unit tests in `UserRepositoryTests` and `ChannelConfigRepositoryTest` covering the dedup.
- Manual: Compose sample with SQLite logging, confirmed write counts drop to the distinct counts on chat open and channel list load.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed duplicate channel configuration and user data being saved to local storage. The system now properly deduplicates entries before storing, retaining the most recent version.

* **Tests**
  * Added test coverage to verify deduplication behavior for channel configurations and user data.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/user/internal/DatabaseUserRepository.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/user/internal/DatabaseUserRepository.kt
@@ -58,15 +58,16 @@ internal class DatabaseUserRepository(
      */
     override suspend fun insertUsers(users: Collection<User>) {
         if (users.isEmpty()) return
-        val usersToInsert = users
-            // Use associateBy instead of distinctBy to keep the *last* occurrence of each user
-            .associateBy(User::id)
-            .values
+        // Use associateBy instead of distinctBy to keep the *last* occurrence of each user
+        val uniqueUsers = users.associateBy(User::id).values
+        val usersToInsert = uniqueUsers
             .filter { it != userCache[it.id] }
             .map { it.toEntity() }
-        cacheUsers(users)
+        cacheUsers(uniqueUsers)
         scope.launchWithMutex(dbMutex) {
-            logger.v { "[insertUsers] inserting ${usersToInsert.size} entities on DB, updated ${users.size} on cache" }
+            logger.v {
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/user/internal/DatabaseUserRepository.kt
```diff
@@ -59,6 +59,9 @@ internal class DatabaseUserRepository(
     override suspend fun insertUsers(users: Collection<User>) {
         if (users.isEmpty()) return
         val usersToInsert = users
+            // Use associateBy instead of distinctBy to keep the *last* occurrence of each user
+            .associateBy(User::id)
+            .values
             .filter { it != userCache[it.id] }
             .map { it.toEntity() }
         cacheUsers(users)
```
aleksandar-apostolov: Question on the log line just below this (`logger.v { ... users.size ... }`): after the dedup, `users.size` no longer matches what the message claims — it's the number of references walked, not the number of distinct users updated in cache. If anyone uses this log to verify the customer's write-volume drop, the second count won't move with the optimization. Worth swapping to a deduped count, or is `users.size` here deliberate? Non-blocking.

aleksandar-apostolov (approved):
LGTM

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/user/internal/DatabaseUserRepository.kt
gpunto: Updated in the latest commit!

PR #6511, opened 2026-06-22
compose: Focus the message composer on entry for screen readers

andremion:
[AND-1248](https://linear.app/stream/issue/AND-1248)

### Goal

When a screen reader user opens a message view, the reading cursor should land on the message composer input, so they can start typing right away (WCAG 2.4.3 Focus Order, 1.3.1). Today the cursor lands somewhere in the middle of the message list, and where it lands varies with the content.

The root cause is that the message list uses `reverseLayout = true`, which scrambles the accessibility traversal order (Compose bug b/232293357). So traversal-order solutions (`traversalIndex` / `isTraversalGroup`) cannot reliably hoist the composer to the top, and Compose exposes no public API to set initial accessibility focus.

### Implementation

`ComposerScreenReaderEntryFocus` in `ChannelScreen` moves the reading cursor onto the composer input when a message view opens. It only runs when an explore-by-touch service (TalkBack) is active, and it moves the reading cursor only, not input focus, so the keyboard stays closed and the user double-taps to type.

- It dispatches `ACTION_ACCESSIBILITY_FOCUS` to the node tagged as the composer input. There is no public Compose API for this, so the node id is resolved reflectively via `AndroidComposeView.getSemanticsOwner()`. The lookup fails safe: if anything cannot be resolved, focus is simply left untouched. A `consumer-rules.pro` keep-rule preserves that method in R8-minified release builds.
- The message list loads asynchronously and the screen reader re-asserts its own initial focus when content appears. To not lose that race, the cursor is re-applied while the list settles (keyed on the item count) within a short bounded window, after which it stops so later incoming messages do not pull focus back.
- A thread is also a message view. The logic re-arms when the view switches between the channel and a thread (keyed on `parentMessageId`), so opening a thread in place focuses the composer too.

Shared accessibility helpers live in a new `AccessibilityUtils.kt`. The existing private touch-exploration helper in `GiphyMessageContent` was consolidated into it.

### 🎨 UI Changes

[Screen_recording_20260622_161048.webm](https://github.com/user-attachments/assets/3274e88b-c41c-4685-9db6-9f4e4b940e95)

### Testing

Requires TalkBack on a device or emulator (with the soft keyboard enabled so it behaves like a physical device).

1. Turn TalkBack on.
2. From the channel list, open a channel. The reading cursor should land on the message composer input, and the keyboard should stay closed.
3. Double-tap. The keyboard should open and the input should be ready to type.
4. Open a thread in place from a message's reply footer. The cursor should land on the composer input again.
5. Open a channel from the Threads tab. The cursor should land on the composer input.
6. Send or receive a few messages after the screen settles. The cursor should not jump back to the composer.
7. Turn TalkBack off and confirm normal use is unchanged (the composer is not auto-focused and no keyboard opens on entry).

Automated coverage: `AccessibilityUtilsTest` (touch-exploration state, semantics node lookup, fail-safe behavior) and the existing `MessageComposerScreenTest` / `ChannelScreenTest`.

## Release Notes

* **Bug Fixes**
  * Improved screen reader (TalkBack) focus management: message composer input now receives proper accessibility focus when messages open, enhancing the experience for screen reader users.

* **Refactor**
  * Consolidated accessibility utilities into a shared module for improved code reusability and consistency.
  * Updated test infrastructure with standardized composer input test tagging.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/ChannelScreen.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/ChannelScreen.kt
@@ -43,9 +43,12 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
+import androidx.compose.runtime.withFrameNanos
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.LocalResources
+import androidx.compose.ui.platform.LocalView
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
@@ -59,6 +62,7 @@ import io.getstream.chat.android.compose.ui.components.moderatedmessage.Moderate
 import io.getstream.chat.android.compose.ui.components.poll.PollAnswersDialog
 import io.getstream.chat.android.compose.ui.components.poll.PollMoreOptionsDialog
 import io.getstream.chat.android.compose.ui.components.poll.PollViewResultDialog
+import io.getstream.chat.android.compose.ui.messages.composer.ComposerInputTestTag
 import io.getstream.chat.android.compose.ui.messages.composer.MessageComposer
 import io.getstream.chat.android.compose.ui.messages.list.LocalSelectedMessageSnapshot
 import io.getstream.chat.android.compose.ui.messages.list.MessageList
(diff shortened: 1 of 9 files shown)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/ChannelScreen.kt
```diff
@@ -275,6 +284,43 @@ public fun ChannelScreen(
     }
 }
 
+/**
+ * When a screen reader is active, moves the reading cursor onto the message composer input as a
+ * message view opens, without moving input focus, so the keyboard stays closed and the user
+ * double-taps to type. A thread is also a message view, so this re-arms when the view switches
```
gpunto: nit: Should these be message list/messages? 🤔 

```suggestion
 * message list view opens, without moving input focus, so the keyboard stays closed and the user
 * double-taps to type. A thread is also a message list view, so this re-arms when the view switches
```

gpunto (approved):
Looks good, I just left a nit on a kdoc

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/ChannelScreen.kt
andremion: cd1fb0c798c33b942464e071bb1dc345a0303485

PR #6515, opened 2026-06-24
compose: Show member role badge and muted indicator in group info

andremion:
### Goal

On the Group Info members list, each member's role was shown as plain tertiary text and there was no muted indicator. This aligns the member row with the new design: the role is shown as a pill badge and a mute icon is shown for muted members.

Resolves AND-1256.

### Implementation

- Render the member role as a pill badge (`RoleBadge`) instead of plain text. The SDK shows a single role per member, resolved by priority: owner (derived from the channel creator, primary label color), then the global user role (admin), then the channel role (moderator or a custom role). The default `channel_member` role is not labeled.
- Show a mute icon next to the member's name and last seen text when the current user has muted that member.
- Split the member row into `Leading`, `Center` and `Trailing` component-factory slots so integrators can override each part. The mute icon lives in the SDK-owned center content, so it stays even when the trailing roles are overridden.
- Add internal label color tokens (primary and neutral, background and text) and a public reusable `RoleBadge` component.
- The multi-role layout (several badges per member, for example Owner and Admin side by side) is demonstrated in the Compose sample app by overriding the trailing content, not in the SDK.
- Add the `Admin` role string with translations for all 7 locales.

Roles come from three independent fields: ownership is derived from the channel creator, admin comes from the global user role (`member.user.role`), and the channel role holds moderator or a custom role. The channel role never holds admin. This matches the React Native and backend models.

### 🎨 UI Changes

The role pill and mute icon are visible in the Group Info members list. Updated Paparazzi snapshots are included in the diff (`GroupChannelInfoContentTest`: collapsed and expanded, light and dark).

### Testing

Manual steps:

1. Open a group channel and go to Group Info.
2. Confirm each member with an elevated role shows a single pill: owner shows "Owner" with the primary (light blue) color, admin/moderator/custom roles show the neutral (gray) color. Members with only the default role show no badge.
3. Mute one of the members (or use a channel where the current user has muted a member) and confirm a mute icon appears next to that member's name.
4. Run the Compose sample app, open Group Info, and confirm a member with more than one role (for example an owner who is also an admin) shows multiple badges side by side. This is the sample-only multi-role override.
5. Check both light and dark mode.

## Summary by CodeRabbit

* **New Features**
  * Channel member rows now show richer role badges, including owner, admin, moderator, and custom roles.
  * Member entries can now display a mute indicator when applicable.
  * Added localized admin labels across multiple languages.
* **UI Improvements**
  * Updated channel info member styling to use compact pill badges with distinct emphasis for primary roles.
  * Improved member row layout for clearer avatar, name, status, and role display.


andremion:
@coderabbit review

stream-chat-android-compose/api/stream-chat-android-compose.api
```diff
@@ -3495,8 +3499,11 @@ public abstract interface class io/getstream/chat/android/compose/ui/theme/ChatC
...
 	public fun GroupChannelInfoAvatarContainer (Lio/getstream/chat/android/compose/ui/theme/GroupChannelInfoAvatarContainerParams;Landroidx/compose/runtime/Composer;I)V
 	public fun GroupChannelInfoExpandMembersItem (Lio/getstream/chat/android/compose/ui/theme/GroupChannelInfoExpandMembersItemParams;Landroidx/compose/runtime/Composer;I)V
+	public fun GroupChannelInfoMemberCenterContent (Lio/getstream/chat/android/compose/ui/theme/GroupChannelInfoMemberCenterContentParams;Landroidx/compose/runtime/Composer;I)V
 	public fun GroupChannelInfoMemberItem (Lio/getstream/chat/android/compose/ui/theme/GroupChannelInfoMemberItemParams;Landroidx/compose/runtime/Composer;I)V
+	public fun GroupChannelInfoMemberLeadingContent (Lio/getstream/chat/android/compose/ui/theme/GroupChannelInfoMemberLeadingContentParams;Landroidx/compose/runtime/Composer;I)V
 	public fun GroupChannelInfoMemberSection (Lio/getstream/chat/android/compose/ui/theme/GroupChannelInfoMemberSectionParams;Landroidx/compose/runtime/Composer;I)V
+	public fun GroupChannelInfoMemberTrailingContent (Lio/getstream/chat/android/compose/ui/theme/GroupChannelInfoMemberTrailingContentParams;Landroidx/compose/runtime/Composer;I)V
```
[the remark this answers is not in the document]
andremion: Thanks. This is binary-safe here and follows the SDK's existing pattern, so no rebuild is forced on existing implementers:

- The module builds with Kotlin 2.2.0, where the default `-jvm-default` mode emits real JVM default methods on the interface plus compatibility bridges. Adding default methods to `ChatComponentFactory` is binary-compatible, existing custom implementations keep working without recompiling.
- `ChatComponentFactory` is an additive extension point by design: it already exposes ~196 default factory methods, and new components extend it in most PRs. These three follow the same shape (a default body plus a params holder).

Resolving as not applicable.

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/MemberRolesComponentFactory.kt
```diff
@@ -0,0 +1,113 @@
...
+import io.getstream.chat.android.ui.common.R as UiCommonR
+
+/**
+ * Sample [ChatComponentFactory] that shows multiple role badges per member in the group channel
+ * info screen. The SDK shows a single role badge, this override demonstrates the multi-role layout.
+ *
+ * The mute icon stays provided by the SDK (in the center content), this override only replaces the
+ * trailing role badges.
+ */
+internal object MemberRolesComponentFactory : ChatComponentFactory {
```
gpunto, line 39: Just for confirmation, are we aligned with the other SDKs in having the sample using a configuration different than the default?

gpunto (approved):

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/MemberRolesComponentFactory.kt
andremion: Yes. The designer has confirmed that the samples should handle the multi-role logic. Anything else is on the SDKs.

PR #6517, opened 2026-06-24
compose: Restore location sharing in the sample behind a feature flag

gpunto:
### Goal

While working on the OpenAPI migration, I realized the demo location sharing feature was still unused (we removed it while testing the v7 redesign). I'm readding it behind a flag so it's easier to test location (which is what I needed to do for testing the app with generated DTOs).

Closes [AND-1259](https://linear.app/stream/issue/AND-1259/restore-location-sharing-in-the-compose-sample-behind-a-feature-flag)

### Implementation

- Added an `isLocationSharingEnabled` flag to `CustomSettings` (SharedPreferences, defaults to `false`) and a toggle for it on the login/settings screen.
- `ChannelActivity` now passes `LocationComponentFactory` when the flag is on, and a no-op factory otherwise.

### Testing

- Manual: with the flag off, the attachment picker is unchanged; with it on, a Location option appears and sharing static/live locations works. Required location permissions are already declared in the sample manifest.

## Summary by CodeRabbit

* **New Features**
  * Added a new location sharing option in the sample app’s login settings.
  * Channel screens now adapt to the location sharing setting and can enable location-related chat components.
  * Theme customization now supports supplying a component factory for chat UI extensions.

* **Bug Fixes**
  * Improved toolbar and feature flag text styling to better match the app theme.

### DIFF
--- a/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/ChannelActivity.kt
+++ b/stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/ChannelActivity.kt
@@ -23,13 +23,17 @@ import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
 import androidx.compose.ui.res.stringResource
 import io.getstream.chat.android.compose.sample.R
 import io.getstream.chat.android.compose.sample.data.customSettings
 import io.getstream.chat.android.compose.sample.feature.channel.isGroupChannel
 import io.getstream.chat.android.compose.sample.ui.SampleChatTheme
+import io.getstream.chat.android.compose.sample.ui.location.LocationComponentFactory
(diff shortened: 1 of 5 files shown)


stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/ChannelActivity.kt
```diff
@@ -72,6 +76,13 @@ class ChannelActivity : ComponentActivity() {
 
     @Composable
     private fun SetupChatTheme() {
+        val componentFactory = remember(cid) {
```
andremion: I believe remembering `cid` is not necessary because it is constant for the activity's lifetime.

andremion (approved):
Looks good.
I just left a NIT on `remember(cid)`

stream-chat-android-compose-sample/src/main/java/io/getstream/chat/android/compose/sample/ui/channel/ChannelActivity.kt
gpunto: Right! Removed

PR #6518, opened 2026-06-24
compose: Introduce AvatarPresenceIndicator and hide offline avatar indicator by default

andremion:
### Goal

Today the Compose SDK always draws a presence dot on user avatars when the indicator is requested: green when the user is online and grey when offline. This does not match the design system ("Common / Avatar Behavior") or the other Stream SDKs, which show the indicator only when the user is online.

This change hides the indicator when the user is offline by default, and gives integrators per-call control over the presence state. It is aligned with the iOS SwiftUI `AvatarIndicator`.

Linear: https://linear.app/stream/issue/AND-1257

### Implementation

- Added a public `AvatarPresenceIndicator` enum with `Online`, `Offline`, and `None`, mirroring the iOS SwiftUI `AvatarIndicator`.
- `UserAvatar` and `ChannelAvatar` now take `indicator: AvatarPresenceIndicator = None`. The shared `OnlineIndicator` draws green for `Online`, grey for `Offline`, and nothing for `None`. This is the single place the dot is rendered, so every screen that shows a presence indicator is covered.
- The default for the SDK screens changes: an online user shows a green dot, an offline user shows nothing. The grey offline dot is now an opt-in: pass `AvatarPresenceIndicator.Offline`, or override the avatar in `ChatComponentFactory`. There is no theme flag, which matches how iOS does it.
- Non-breaking. The previous `showIndicator: Boolean` parameter is deprecated and kept on both composables and on `UserAvatarParams` / `ChannelAvatarParams`, alongside the new `indicator`. The deprecated path preserves today's behavior (green online, grey offline) so existing integrations do not change silently. The factory resolves `params.indicator`, falling back to the deprecated `showIndicator` when `indicator` is not set.
- Migrated every SDK call site that showed the indicator (channel info, member info, members list, channel actions, search results, thread items) to the online-only mapping.

### 🎨 UI Changes

Avatars that show a presence indicator no longer draw the grey dot when the user is offline; they show nothing. Online avatars are unchanged (green dot). This is visible on the channel and contact info screens, the members list, the channel actions sheet, search results, and thread items.

The updated Paparazzi snapshots in this PR capture the before/after (the grey offline dot is removed from `UserAvatarTest`, `ThreadListTest`, and `MentionListTest`).

### Testing

Manual steps in the Compose sample app:

1. Open a 1:1 channel where the other user is offline. The avatar in the channel header and on the channel info screen shows no presence dot. When that user is online, it shows a green dot.
2. Open Group Info for a group with both online and offline members. Online members show a green dot; offline members show no dot (previously grey).
3. Long-press a 1:1 channel to open the channel actions sheet. The dot is shown only when the other user is online.
4. Open search results and the thread list. Author avatars show the dot only when the author is online.
5. To verify the opt-in: in a `ChatTheme` `componentFactory` override of `UserAvatar`, pass `indicator = AvatarPresenceIndicator.Offline` (or `if (user.online) AvatarPresenceIndicator.Online else AvatarPresenceIndicator.Offline`) and confirm the grey dot returns for offline users.

`AvatarPresenceIndicatorTest` covers the resolver mapping (online, offline with and without the opt-in, the channel "any member online" rule, and the null current user case).

## Summary by CodeRabbit

* **New Features**
  * Added richer avatar presence states, including online, offline, and hidden/none.
  * Avatars across chat screens now show presence more accurately based on user or channel status.

* **Bug Fixes**
  * Replaced several always-on or always-off presence indicators with status-driven behavior.
  * Updated avatar display in lists, member views, search results, threads, pinned messages, polls, and channel info screens.

* **Tests**
  * Added coverage for presence indicator behavior across users and channels.


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/UserAvatar.kt
```diff
@@ -84,6 +85,42 @@ public fun UserAvatar(
...
+ * @param showIndicator Whether to overlay a status indicator to show whether the user is online.
+ * @param showBorder Whether to draw a border around the avatar to provide contrast against the background.
+ */
+@Deprecated(
+    message = "Use the overload that takes an AvatarPresenceIndicator. showIndicator showed a grey dot when " +
+        "the user was offline; pass an explicit indicator to control the online, offline, and hidden states.",
+    replaceWith = ReplaceWith(
+        "UserAvatar(user, modifier, if (user.online) AvatarPresenceIndicator.Online " +
+            "else AvatarPresenceIndicator.Offline, showBorder)",
+    ),
```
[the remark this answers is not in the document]
andremion: 46a5b3e271cc2437191ce21b3b52ea0a85418639

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPresenceIndicator.kt
```diff
@@ -0,0 +1,71 @@
...
+ * @param currentUser The user currently logged in, excluded when determining the channel's online state.
+ * @param showWhenOffline Whether to show the grey [AvatarPresenceIndicator.Offline] dot when the channel is offline.
+ * When `false` (the default), an offline channel resolves to [AvatarPresenceIndicator.None] and no dot is drawn.
+ */
+internal fun Channel.avatarPresenceIndicator(
+    currentUser: User?,
+    showWhenOffline: Boolean = false,
+): AvatarPresenceIndicator =
+    resolveAvatarPresenceIndicator(
+        online = members.any { it.user.id != currentUser?.id && it.user.online },
```
gpunto, line 62: Do you think it's worth a `remember`? Might be overkill tho, the only "expensive" operation is traversing `members` 🤔

gpunto (approved):
LG, just left a doubt but probably it can be ignored

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/avatar/AvatarPresenceIndicator.kt
andremion: Yes. It runs only in the channel actions sheet and in the deprecated path. I think remember would be overkill here.

PR #6524, opened 2026-06-26
Fall back to a video frame when the server thumbnail is not ready

andremion:
### Goal

A sent video shows a blank preview in the message list (and in the shared-media grids and quoted previews) until the chat is reopened. Photos work fine.

The cause is a race with server-side thumbnail generation. The upload-intent response returns a `thumbUrl` before the server has finished generating the thumbnail. The client requests that URL right away and gets a 404, so the preview is blank. A few seconds later the thumbnail is ready, and the next time the item is rendered (scroll away and back, or reopen) it loads. The existing `?: upload` fallback does not help, because `thumbUrl` is present; only the load fails.

iOS already handles this: when the thumbnail load fails, it extracts a frame from the video itself. This change brings Android to the same behavior.

Resolves AND-1267

### Implementation

Core, in `stream-chat-android-ui-common`:

- `VideoThumbnailImageData(thumbnailUrl, videoUrl)` is a small request model used as the Coil data for a video preview.
- `VideoThumbnailFallbackInterceptor` loads the server thumbnail first and, when it is missing or fails, falls back to a frame from the video. It rewrites the request to a plain URL before proceeding, so CDN signing, headers, and caching are reused. It is registered as the outermost interceptor, so the CDN interceptor still signs the fallback video URL.
- `VideoFrameFetcher` extracts the fallback frame with `MediaMetadataRetriever`, which seeks with HTTP range requests and reads only the bytes needed for the frame. The full video is never written to the image disk cache. This matches the cost profile of iOS (`AVAssetImageGenerator` on a remote asset). Local files keep using Coil's `VideoFrameDecoder`.

Call sites now pass `VideoThumbnailImageData(thumbUrl, assetUrl)` for videos, on both kits and on the same surfaces iOS covers:

- Message list: `MediaAttachmentContent` (Compose), `MediaAttachmentView` (XML).
- Shared-media grid: `ChannelMediaAttachmentsScreen` (Compose), `MediaAttachmentAdapter` (XML).
- Quoted and file previews: `imagePreviewData` (Compose, shared by quoted and file content), `loadAttachmentThumb` (XML).

The new types are `@InternalStreamChatApi`, so there is no public API change.

### 🎨 UI Changes

Message list with a video whose thumbnail is not yet available.

| Before | After |
| --- | --- |
| before-blank-video-preview | after-video-frame-fallback |

### Testing

1. Open the Compose sample and open a channel that has a video message.
2. With a normal thumbnail, the preview shows as usual.
3. To exercise the fallback deterministically, apply the patch below to force the thumbnail load to fail, reinstall, and open the channel. The preview should still render, now from a frame of the video instead of a blank box.

<details>

<summary>Patch to force the thumbnail to fail</summary>

```kotlin
// AttachmentExtensions.kt, imagePreviewData, video branch
isVideo() && ChatTheme.videoThumbnailsEnabled -> {
    val thumbnailUrl = thumbUrl?.let { "https://invalid.invalid/nonexistent-thumbnail.jpg" }
    if (thumbnailUrl != null || assetUrl != null) {
        VideoThumbnailImageData(thumbnailUrl = thumbnailUrl, videoUrl = assetUrl)
    } else {
        upload
    }
}
```

</details>

Verified on the emulator: the frame renders through the fallback, and the full video is not written to the image disk cache (after clearing the cache and reloading, the largest cache entry stays in the hundreds of KB).

## Summary by CodeRabbit

* **New Features**
  * Improved video attachment previews across the app: load a server thumbnail first, and automatically fall back to a preview frame from the video when needed.

* **Bug Fixes**
  * Fixed cases where video attachments could show blank previews or fail to load in gallery, message lists, and attachment views—especially when thumbnail/upload data is missing or incomplete.

* **Tests**
  * Added coverage for video frame preview selection and thumbnail fallback behavior.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/extensions/internal/AttachmentExtensions.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/extensions/internal/AttachmentExtensions.kt
@@ -21,6 +21,7 @@ import io.getstream.chat.android.client.utils.attachment.isImage
 import io.getstream.chat.android.client.utils.attachment.isVideo
 import io.getstream.chat.android.compose.ui.theme.ChatTheme
 import io.getstream.chat.android.models.Attachment
+import io.getstream.chat.android.ui.common.images.internal.VideoThumbnailImageData
 import io.getstream.chat.android.ui.common.images.resizing.applyStreamCdnImageResizingIfEnabled
 
 /**
@@ -30,7 +31,9 @@ import io.getstream.chat.android.ui.common.images.resizing.applyStreamCdnImageRe
  * Otherwise, it returns null.
  *
  * For image attachments, [Attachment.imageUrl] is used.
- * For video attachments when thumbnails are enabled, [Attachment.thumbUrl] is used.
+ * For video attachments when thumbnails are enabled, the server [Attachment.thumbUrl] is used,
+ * falling back to a frame extracted from the [Attachment.assetUrl] video when the thumbnail is
+ * missing or fails to load (see [VideoThumbnailImageData]).
  */
 @get:Composable
 internal val Attachment.imagePreviewData: Any?
@@ -39,9 +42,13 @@ internal val Attachment.imagePreviewData: Any?
             imageUrl
                 ?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
                 ?: upload
-        isVideo() && ChatTheme.videoThumbnailsEnabled ->
-            thumbUrl
-                ?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
-                ?: upload
+        isVideo() && ChatTheme.videoThumbnailsEnabled -> {
+            val thumbnailUrl = thumbUrl?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
+            if (thumbnailUrl != null || assetUrl != null) {
+                VideoThumbnailImageData(thumbnailUrl = thumbnailUrl, videoUrl = assetUrl)
+            } else {
+                upload
+            }
+        }
         else -> null
     }
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/extensions/internal/AttachmentExtensions.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/extensions/internal/AttachmentExtensions.kt
@@ -21,7 +21,7 @@ import io.getstream.chat.android.client.utils.attachment.isImage
 import io.getstream.chat.android.client.utils.attachment.isVideo
 import io.getstream.chat.android.compose.ui.theme.ChatTheme
 import io.getstream.chat.android.models.Attachment
-import io.getstream.chat.android.ui.common.images.internal.VideoThumbnailImageData
+import io.getstream.chat.android.ui.common.images.internal.videoThumbnailImageData
 import io.getstream.chat.android.ui.common.images.resizing.applyStreamCdnImageResizingIfEnabled
 
 /**
@@ -33,7 +33,7 @@ import io.getstream.chat.android.ui.common.images.resizing.applyStreamCdnImageRe
  * For image attachments, [Attachment.imageUrl] is used.
  * For video attachments when thumbnails are enabled, the server [Attachment.thumbUrl] is used,
  * falling back to a frame extracted from the [Attachment.assetUrl] video when the thumbnail is
- * missing or fails to load (see [VideoThumbnailImageData]).
+ * missing or fails to load (see [videoThumbnailImageData]).
  */
 @get:Composable
 internal val Attachment.imagePreviewData: Any?
@@ -44,11 +44,7 @@ internal val Attachment.imagePreviewData: Any?
                 ?: upload
         isVideo() && ChatTheme.videoThumbnailsEnabled -> {
             val thumbnailUrl = thumbUrl?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
-            if (thumbnailUrl != null || assetUrl != null) {
-                VideoThumbnailImageData(thumbnailUrl = thumbnailUrl, videoUrl = assetUrl)
-            } else {
-                upload
-            }
+            videoThumbnailImageData(thumbnailUrl) ?: upload
         }
         else -> null
     }
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/internal/VideoFrameFetcher.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/internal/VideoFrameFetcher.kt
@@ -77,7 +77,7 @@ internal class VideoFrameFetcher(
         try {
             retriever.setDataSource(data.toString(), options.requestHeaders())
             val bitmap = retriever.extractFrame()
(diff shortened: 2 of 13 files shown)


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/extensions/internal/AttachmentExtensions.kt
```diff
@@ -39,9 +42,14 @@ internal val Attachment.imagePreviewData: Any?
...
-                ?: upload
+        isVideo() && ChatTheme.videoThumbnailsEnabled -> {
+            // TEMP TEST: force the thumbnail to fail so the MediaMetadataRetriever fallback runs.
+            val thumbnailUrl = thumbUrl?.let { "https://invalid.invalid/nonexistent-thumbnail.jpg" }
+            if (thumbnailUrl != null || assetUrl != null) {
+                VideoThumbnailImageData(thumbnailUrl = thumbnailUrl, videoUrl = assetUrl)
+            } else {
+                upload
+            }
+        }
```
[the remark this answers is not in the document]
andremion: bc99d1bbbb7f72dce78112ef6ac069d2cfa594be

andremion:
@coderabbit review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/internal/VideoFrameFetcher.kt
```diff
@@ -0,0 +1,117 @@
...
+import coil3.network.httpHeaders
+import coil3.request.ImageRequest
+import coil3.request.Options
+import coil3.size.pxOrElse
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+
+/**
+ * Offset of the extracted frame, kept slightly above zero so the very first frame
+ * (often black) is skipped. Matches the iOS preview offset.
```
VelikovPetar: Very small remark, feel free to ignore: Not a big fan of referencing other platforms in the KDocs 😄 - I think it can introduce some confusion when reading the code.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/gallery/overview/internal/MediaAttachmentAdapter.kt
```diff
@@ -95,10 +96,20 @@ internal class MediaAttachmentAdapter(
 
             val imageData = if (shouldLoadImage) {
                 val attachment = attachmentGalleryItem.attachment
-                val url = if (attachment.isImage()) attachment.imageUrl else attachment.thumbUrl
-                url?.applyStreamCdnImageResizingIfEnabled(
-                    streamCdnImageResizing = ChatUI.streamCdnImageResizing,
-                )
+                if (attachment.isVideo()) {
+                    val thumbnailUrl = attachment.thumbUrl?.applyStreamCdnImageResizingIfEnabled(
```
VelikovPetar: Thinking loud here - maybe it would make sense to encapsulate this wrapping logic as an `Attachment` extension? I see it (almost) duplicated in several places, maybe we can unify it as a single extension method/property

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/extensions/internal/AttachmentExtensions.kt
```diff
@@ -39,9 +42,13 @@ internal val Attachment.imagePreviewData: Any?
             imageUrl
                 ?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
                 ?: upload
-        isVideo() && ChatTheme.videoThumbnailsEnabled ->
-            thumbUrl
-                ?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
-                ?: upload
+        isVideo() && ChatTheme.videoThumbnailsEnabled -> {
+            val thumbnailUrl = thumbUrl?.applyStreamCdnImageResizingIfEnabled(ChatTheme.streamCdnImageResizing)
+            if (thumbnailUrl != null || assetUrl != null) {
```
VelikovPetar: Small thing to discuss here: previously, we favoured the data sources in the order: `thumbUrl > `upload`. Now it will be `thumbUrl OR assetUrl > upload`. I am thinking if it still worth to prioritise `thumbUrl > upload > assetUrl`, so that the thumbnail is extracted from the local file, instead off accessing the network?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/internal/VideoFrameFetcher.kt
```diff
@@ -0,0 +1,117 @@
...
+    private val data: Uri,
+    private val options: Options,
+) : Fetcher {
+
+    override suspend fun fetch(): FetchResult = withContext(Dispatchers.IO) {
+        val retriever = MediaMetadataRetriever()
+        try {
+            retriever.setDataSource(data.toString(), options.requestHeaders())
+            val bitmap = retriever.extractFrame()
+                ?: error("Could not extract a preview frame from video: $data")
```
VelikovPetar: Thinking if we should not include the whole `data` object in the error message - it can probably have some sensitive info like signing headers (or some data that customers would not like to be potentially visible in the logs.)

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/internal/VideoFrameFetcher.kt
andremion: Agreed, removed. 5c3b4636b33098c92272770ced47945b670e127d

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/gallery/overview/internal/MediaAttachmentAdapter.kt
andremion: Done. Extracted into `Attachment.videoThumbnailImageData` and collapsed the 5 call sites. 06a503fcc7660fd6bb16bf410da294976d1126c8

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/internal/VideoFrameFetcher.kt
andremion: Good point, dropped the URL from the message. a9d4e1a9f2aca72c85c11fff359147be2a5829a3

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/util/extensions/internal/AttachmentExtensions.kt
andremion: Traced it: the local `upload` file is still present right after send, so a local-first fallback is possible. But iOS uses the remote `videoURL` for the fallback (it sets `videoURL` to the remote URL after upload), so it does not prefer the local file either. I kept the remote `assetUrl` to match iOS, which is what this fix targets. A local-first fallback would be an Android-only optimization beyond parity, happy to do it as a follow-up if we want.

VelikovPetar: OK let's keep it aligned then!

VelikovPetar (approved):

PR #6526, opened 2026-06-26
ui-common: Ignore mention suggestions that start with whitespace

gpunto:
### Goal

In the message composer, typing `@` followed by spaces keeps the mention suggestion active with a whitespace-only query, which makes the group/role lookup return everything.

`TypingSuggester.typingSuggestion` lets a suggestion token start with whitespace, so `@ ` produces a `" "` query that is forwarded to the lookup. The fix rejects suggestions that start with whitespace, while still allowing internal spaces so multi-word names (e.g. `@John Doe`) remain searchable.

Closes [AND-1268](https://linear.app/stream/issue/AND-1268/composer-typing-spaces-after-reveals-all-groupsroles)

### Implementation

- `TypingSuggester` now returns no suggestion when the token starts with whitespace. Internal and trailing spaces are still part of the token, so multi-word names keep working.

### Testing

- New `TypingSuggesterTest`: empty token, single-word and mid-text tokens, multi-word and trailing-space tokens, and the leading-whitespace cases that now produce no suggestion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mention suggestions so matches are no longer shown when the suggested text starts with whitespace, while still allowing valid internal spacing.
* **Tests**
  * Added coverage for mention suggestion behavior across common typing cases and invalid whitespace patterns.

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/typing/TypingSuggester.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/typing/TypingSuggester.kt
@@ -58,6 +58,11 @@ internal class TypingSuggester(private val options: TypingSuggestionOptions) {
         }
 
         val suggestionText = text.substring(suggestionStart, caretLocation)
+        // A suggestion cannot start with whitespace: typing a space right after the symbol ends it (e.g. "@ "),
+        // while internal spaces stay allowed so multi-word names like "@John Doe" remain searchable.
+        if (suggestionText.firstOrNull()?.isWhitespace() == true) {
+            return null
+        }
         if (suggestionText.length < options.minimumRequiredCharacters) {
             return null
         }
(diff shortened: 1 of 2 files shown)


andremion (approved):
LGTM!

One thing on cross-SDK parity: No iOS SDK has this leading-whitespace guard. Looks like the same bug exists on iOS. Should we file equivalent fixes on iOS so the behavior stays aligned?

Minor, not blocking: the Android check uses `isWhitespace()` (covers tabs and other Unicode whitespace), while iOS only treats " " and "\n" as separators. That difference is pre-existing, just noting it in case we want the platforms to classify edge inputs the same way.

gpunto:
> LGTM!
> 
> One thing on cross-SDK parity: No iOS SDK has this leading-whitespace guard. Looks like the same bug exists on iOS. Should we file equivalent fixes on iOS so the behavior stays aligned?
> 
> Minor, not blocking: the Android check uses `isWhitespace()` (covers tabs and other Unicode whitespace), while iOS only treats " " and "\n" as separators. That difference is pre-existing, just noting it in case we want the platforms to classify edge inputs the same way.

Great point, I wrote to the team on Slack, thanks!

PR #6527, opened 2026-06-26
compose: Render composer input and message text in content direction

andremion:
### Goal

The Compose composer input and message bubbles render RTL text (for example Arabic or Hebrew) with the wrong alignment when the device locale is LTR, or when RTL and LTR text are mixed. The input field and message text align to the left, and sentence-final punctuation appears at the visual start of the line.

This re-implements the v6 fix from #6487 (AND-993) on `develop`. The v6 patch could not be cherry-picked because the v7 composer and message theme were restructured.

Resolves AND-1270

### Implementation

- `MessageComposerInputCenterContent`: resolve the layout direction from the input content (the first strong directional character, falling back to the device locale) and wrap the `BasicTextField` in a `CompositionLocalProvider` that provides it. This is the v7 home of the old `MessageInput` change, because the text field now renders through the `MessageComposerInputCenterContent` component factory.
- `MessageStyling.textStyle`: set `textDirection = TextDirection.Content` on the message text style so bubbles render in the direction of their content. This covers both the direct render path and the default `MessageTextFormatter`.
- `QuotedMessage`: set `textDirection = TextDirection.Content` on the quoted text style.
- Added Paparazzi snapshots for the RTL composer input and RTL message text (dark and light).

All changes are internal or private. `apiCheck` reports no public API change.

Difference from v6: the v6 fix wrapped the whole composer input row, so the send button mirrored too. Here only the text field follows the content direction and the send button keeps its position. Happy to switch to full-row mirroring if you prefer parity with v6.

### 🎨 UI Changes

| Before | After |
| --- | --- |
| before_composer_input | after_composer_input |
| before_sent_message | after_sent_message |

### Testing

Tested on an emulator with an LTR device locale, which is the AND-993 scenario (RTL keyboard, not device language):

1. Open a channel.
2. Type an RTL sentence with trailing punctuation, for example `مرحبا، هذه رسالة تجريبية.`
3. The composer input aligns to the right, and the period stays at the logical end (visual left), not at the start.
4. Send the message. The bubble renders the text right-aligned with the period at the logical end.
5. Type LTR text and confirm it still aligns to the left.

I also checked the send path for the punctuation reorder reported in AND-993: the sent string is `state.inputValue` passed straight to `buildNewMessage`, which only prepends an active command and trims whitespace. No character reordering happens, so the symptom was a rendering issue, fixed by the message text direction change.

Validation:

- `./gradlew :stream-chat-android-compose:detekt` pass
- `./gradlew :stream-chat-android-compose:apiCheck` pass (no public API change)
- `./gradlew :stream-chat-android-compose:verifyPaparazziDebug` pass (full module)

## Summary by CodeRabbit

* **New Features**
  * Improved right-to-left (RTL) support in message input and message text previews, including Arabic sample content for better visual validation.

* **Bug Fixes**
  * Fixed text alignment and direction handling so mixed, neutral, and RTL content displays more naturally in the composer and quoted messages.


andremion:
@coderabbit review

gpunto (approved):

PR #6531, opened 2026-06-29
Fall back to a video frame when the server thumbnail is not ready

andremion:
### Goal

A sent video shows a blank preview in the message list (and in the shared-media grids and quoted previews) until the chat is reopened. The upload-intent response returns a `thumbUrl` before server-side thumbnail generation has finished, so the client requests it, gets a 404, and the preview is blank until the item is rendered again. When the thumbnail fails, we now extract a preview frame from the video itself, matching the iOS behavior.

This is the develop (v7) port of the v6 fix in #6524.

Resolves AND-1271

### Implementation

Core, in `stream-chat-android-ui-common`:

- `VideoThumbnailImageData(thumbnailUrl, videoUrl)` plus the `Attachment.videoThumbnailImageData(...)` builder.
- `VideoThumbnailFallbackInterceptor` loads the server thumbnail first and, when it is missing or fails, falls back to a frame from the video. Registered as the outermost interceptor so CDN signing still applies to the fallback URL.
- `VideoFrameFetcher` extracts the fallback frame with `MediaMetadataRetriever`, which seeks with HTTP range requests and reads only the bytes needed for the frame; the full video is never written to the image disk cache.

Call sites route video previews through the shared extension on both kits: message list (`MediaAttachmentContent`, `MediaAttachmentView`), shared-media grids (`ChannelMediaAttachmentsScreen`, `MediaAttachmentAdapter`), and file/quoted previews (`AttachmentExtensions.imagePreviewData`, `AttachmentUtils`, `QuotedMessageBodyBuilder`).

Notable differences from the v6 PR, due to develop divergence:
- The video-thumbnails flag moved into `ChatTheme.config.messageList.videoThumbnailsEnabled`.
- The quoted preview now goes through `QuotedMessageBodyBuilder` (the old `MediaAttachmentQuotedContent` no longer exists), thumbnail-first.

The new types are `@InternalStreamChatApi`, so there is no public API change.

### 🎨 UI Changes

Message list with a video whose thumbnail is not yet available.

| Before | After |
| --- | --- |
| before-blank-video-preview | after-video-frame-fallback |

### Testing

1. Open the Compose sample and open a channel that has a video message.
2. With a normal thumbnail, the preview shows as usual.
3. To exercise the fallback deterministically, apply the patch below to force the thumbnail load to fail, reinstall, and open the channel. The preview should still render, now from a frame of the video instead of a blank box.

<details>

<summary>Patch to force the thumbnail to fail</summary>

```kotlin
// AttachmentExtensions.kt, imagePreviewData, video branch
isVideo() && ChatTheme.config.messageList.videoThumbnailsEnabled ->
    videoThumbnailImageData("https://invalid.invalid/nonexistent-thumbnail.jpg")
        ?: upload
```

</details>

Verified on the emulator: the frame renders through the fallback, and the full video is not written to the image disk cache.

## Summary by CodeRabbit

* **New Features**
  * Improved video attachment previews across chat, gallery, quoted messages, and message lists.
  * Video thumbnails can now fall back to a frame from the video when a dedicated thumbnail isn’t available.

* **Bug Fixes**
  * Fixed video preview loading when thumbnail URLs are missing or fail to load.
  * Improved consistency for attachment preview images, including better handling for video and image fallback cases.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/attachments/ChannelMediaAttachmentsScreen.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/channel/attachments/ChannelMediaAttachmentsScreen.kt
@@ -53,6 +53,7 @@ import io.getstream.chat.android.compose.viewmodel.channel.ChannelAttachmentsVie
 import io.getstream.chat.android.compose.viewmodel.channel.ChannelAttachmentsViewModelFactory
 import io.getstream.chat.android.previewdata.PreviewMessageData
 import io.getstream.chat.android.ui.common.feature.channel.attachments.ChannelAttachmentsViewAction
+import io.getstream.chat.android.ui.common.images.internal.videoThumbnailImageData
 import io.getstream.chat.android.ui.common.state.channel.attachments.ChannelAttachmentsViewState
 import io.getstream.result.Error
 import io.getstream.chat.android.ui.common.R as UiCommonR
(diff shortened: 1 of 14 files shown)


andremion:
@coderabbit review

aleksandar-apostolov (approved):
LGTM

VelikovPetar (approved):

PR #6533, opened 2026-06-29
Add opt-in video disk cache for ExoPlayer playback

VelikovPetar:
### Goal

Media attachments currently re-stream from the CDN on every replay, seek, or scroll-back. This wastes customer bandwidth and forces users to wait through buffering on content they have already watched. This PR introduces an opt-in on-disk media cache so previously watched videos (and other ExoPlayer-driven media, including audio) replay from local bytes.

### Implementation

**Public API**

A new `io.getstream.chat.android.client.cache` package exposes the opt-in surface:

- `StreamCacheConfig(video: VideoCacheConfig? = null)` — aggregate config; future cache types can be added as additional nullable properties without breaking the call site.
- `VideoCacheConfig(maxSizeBytes: Long)` — soft cap on cache size; defaults to 150 MB. LRU eviction kicks in once exceeded. Rejects non-positive sizes at construction.
- `ChatClient.Builder.cacheConfig(StreamCacheConfig)` — single entry point. Default is off; existing apps see no behavior change.

Usage:
```kotlin
ChatClient.Builder("api-key", context)
    .cacheConfig(StreamCacheConfig(video = VideoCacheConfig()))
    .build()
```

> Note: the config is named `VideoCacheConfig` because video was the original motivation, but a single cache now backs all ExoPlayer media. We intentionally kept one config surface rather than splitting video/audio granularity in this first iteration.

**Internals**

- `VideoMediaCache` owns a Media3 `SimpleCache` + `StandaloneDatabaseProvider`. A process-wide registry keyed by the cache directory's absolute path prevents the "two `SimpleCache` instances against the same directory" crash if a customer rebuilds `ChatClient` in the same process.
- `VideoCacheDataSourceFactory` wraps Media3's `CacheDataSource.Factory` and composes _outside_ any custom CDN — on a hit the bytes are served from disk and the customer's `CDN.fileRequest` is not consulted; on a miss the CDN signs the URL just-in-time. The cache key strips URL query parameters, so rotating pre-signed signature/query values still resolve to the same cached entry.
- Wired into all four Media3 playback sites, each passing the shared `videoCache`:
  - Compose: `MediaPreviewActivity` → `StreamMediaPlayerContent`
  - UI components: `AttachmentMediaActivity`, `AttachmentGalleryVideoPageFragment`
  - Audio playback: `StreamAudioPlayer` pipeline built in `ChatClient.Builder.build()`
- All media routed through these sites is cached — audio included. No content-type gating.
- Cache clearing is **in-place**: `clearCacheAndTemporaryFiles(context)` clears every live cache in the process via `VideoMediaCache.clearAll()` (which empties the `SimpleCache` while keeping it alive and its directory lock held), and only falls back to deleting the cache directory when no live cache owns it. This avoids corrupting Media3's on-disk index/lock by deleting a directory out from under a live `SimpleCache` — including the rebuild-without-cache case.

### UI Changes

No UI changes.

### Testing

**Automated**
- `StreamMediaDataSourceCacheIntegrationTest` — four-cell matrix `{custom CDN, no CDN} × {hit, miss}`; rotated-query cache-hit case; LRU eviction (three videos into a 2-video-capacity cache, verifying the least-recently-used entry is dropped); in-place `clear`; and `clearAll` (empties live caches and reports whether any existed, so the fallback to directory deletion only runs when none do).
- `VideoMediaCacheTest` — directory creation, `release()` idempotency, per-directory instance registry.
- `VideoCacheConfigTest` — default/positive/zero/negative `maxSizeBytes` validation.
- `StreamFileManagerTest` / `ChatClientCacheAndTemporaryFilesTest` — cache directory lookup and cleanup wiring.

```
./gradlew :stream-chat-android-client:testDebugUnitTest \
    --tests "io.getstream.chat.android.client.cache.internal.*" \
    --tests "io.getstream.chat.android.client.ChatClientCacheAndTemporaryFilesTest"
```

**Manual smoke**

1. Enable via `ChatClient.Builder.cacheConfig(StreamCacheConfig(video = VideoCacheConfig()))` in the sample app.
2. Send a video attachment, scroll away, scroll back, tap to play.
3. Close, then tap the same video again.
4. **Expected:** the second play has no network traffic in Profiler; the cache directory under `<cacheDir>/stream_video_cache/` contains files.
5. Repeat with an audio attachment — it is cached the same way.
6. Kill and restart the app; playing the same media should still serve from cache (no network).
7. Trigger `clearCacheAndTemporaryFiles(context)`; the cached bytes are removed and playback of the same media re-fetches from the network, with the cache staying functional afterwards.

## Summary by CodeRabbit

* **New Features**
  * Added optional video caching for streamed media, improving playback performance and reducing repeated downloads.
  * Expanded media preview and gallery playback to use the new caching flow automatically for video content.
  * Added a new configuration option to control cache settings, including cache size.

* **Bug Fixes**
  * Improved cache clearing so video cache data is removed correctly.
  * Fixed cache behavior for media URLs with changing query parameters, helping reuse cached video content more reliably.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -1523,12 +1523,20 @@ internal constructor(
     public fun clearCacheAndTemporaryFiles(context: Context): Call<Unit> =
         CoroutineCall(clientScope) {
             logger.d { "[clearCacheAndTemporaryFiles] Clearing all cache and temporary files" }
+            // Clear video cache: in-place via the live cache when opted in (keeps the SimpleCache
+            // alive so playback continues to work), or by deleting the directory when no live
+            // cache owns it.
+            val videoCacheResult = videoCache?.let {
+                it.clear()
+                Result.Success(Unit)
+            } ?: fileManager.clearVideoCache(context)
             // Clear all cache directories
             val cacheResult = fileManager.clearAllCache(context)
             // Clear external (temporary) storage files - always run regardless of cache result
             val externalStorageResult = fileManager.clearExternalStorage(context)
             // Return the first failure if any, otherwise success
             when {
+                videoCacheResult is Result.Failure -> videoCacheResult
                 cacheResult is Result.Failure -> cacheResult
                 externalStorageResult is Result.Failure -> externalStorageResult
                 else -> Result.Success(Unit)
@@ -5217,7 +5225,7 @@ internal constructor(
             val videoCache = cacheConfig?.video?.let {
                 VideoMediaCache.create(appContext, StreamFileManager().getVideoCache(appContext), it)
             }
-            val mediaDataSourceFactory = StreamMediaDataSource.factory(appContext, cdn)
+            val mediaDataSourceFactory = StreamMediaDataSource.factory(appContext, cdn, videoCache)
             val audioPlayer: AudioPlayer = StreamAudioPlayer(
                 mediaPlayer = NativeMediaPlayerImpl(mediaDataSourceFactory) {
                     ExoPlayer.Builder(appContext)
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -72,6 +72,8 @@ import io.getstream.chat.android.client.attachment.AttachmentsSender
 import io.getstream.chat.android.client.audio.AudioPlayer
 import io.getstream.chat.android.client.audio.NativeMediaPlayerImpl
 import io.getstream.chat.android.client.audio.StreamAudioPlayer
+import io.getstream.chat.android.client.cache.StreamCacheConfig
+import io.getstream.chat.android.client.cache.internal.VideoMediaCache
 import io.getstream.chat.android.client.cdn.CDN
 import io.getstream.chat.android.client.cdn.internal.StreamMediaDataSource
 import io.getstream.chat.android.client.channel.ChannelClient
@@ -300,6 +302,8 @@ internal constructor(
     internal val messageReceiptManager: MessageReceiptManager,
     @InternalStreamChatApi
     public val cdn: CDN? = null,
+    @InternalStreamChatApi
+    public val videoCache: VideoMediaCache? = null,
 ) {
     private val logger by taggedLogger(TAG)
     private val fileManager = StreamFileManager()
@@ -4810,6 +4814,7 @@ internal constructor(
         private var fileTransformer: FileTransformer = NoOpFileTransformer
         private var apiModelTransformers: ApiModelTransformers = ApiModelTransformers()
         private var cdn: CDN? = null
+        private var cacheConfig: StreamCacheConfig? = null
         private var appName: String? = null
         private var appVersion: String? = null
 
@@ -5023,6 +5028,15 @@ internal constructor(
         }
 
         /**
+         * Configures the SDK's user-configurable on-disk caches.
+         *
+         * @param config The per-cache configurations.
+         */
+        public fun cacheConfig(config: StreamCacheConfig): Builder = apply {
+            this.cacheConfig = config
+        }
+
+        /**
          * Sets the CDN URL to be used by the client.
          */
         public fun cdnUrl(value: String): Builder = apply {
@@ -5200,6 +5214,9 @@ internal constructor(
             val api = module.api()
             val appSettingsManager = AppSettingManager(api)
 
+            val videoCache = cacheConfig?.video?.let {
+                VideoMediaCache.create(appContext, StreamFileManager().getVideoCache(appContext), it)
+            }
             val mediaDataSourceFactory = StreamMediaDataSource.factory(appContext, cdn)
             val audioPlayer: AudioPlayer = StreamAudioPlayer(
                 mediaPlayer = NativeMediaPlayerImpl(mediaDataSourceFactory) {
@@ -5255,6 +5272,7 @@ internal constructor(
                     api = api,
                 ),
                 cdn = cdn,
+                videoCache = videoCache,
             ).apply {
                 attachmentsSender = AttachmentsSender(
                     context = appContext,
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -1523,13 +1523,15 @@ internal constructor(
     public fun clearCacheAndTemporaryFiles(context: Context): Call<Unit> =
         CoroutineCall(clientScope) {
             logger.d { "[clearCacheAndTemporaryFiles] Clearing all cache and temporary files" }
-            // Clear video cache: in-place via the live cache when opted in (keeps the SimpleCache
-            // alive so playback continues to work), or by deleting the directory when no live
-            // cache owns it.
-            val videoCacheResult = videoCache?.let {
-                it.clear()
+            // Clear video cache: in-place via any live cache in the process (keeps the SimpleCache
+            // alive so playback continues to work), or by deleting the directory when no live cache
+            // owns it. The registry is process-wide, so this covers caches from a prior ChatClient
+            // build even if the current client was built without a cacheConfig.
+            val videoCacheResult = if (VideoMediaCache.clearAll()) {
                 Result.Success(Unit)
-            } ?: fileManager.clearVideoCache(context)
+            } else {
+                fileManager.clearVideoCache(context)
+            }
             // Clear all cache directories
             val cacheResult = fileManager.clearAllCache(context)
             // Clear external (temporary) storage files - always run regardless of cache result
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/StreamCacheConfig.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/StreamCacheConfig.kt
@@ -43,6 +43,10 @@ public data class StreamCacheConfig(
 public data class VideoCacheConfig(
     public val maxSizeBytes: Long = DEFAULT_MAX_SIZE_BYTES,
 ) {
+    init {
+        require(maxSizeBytes > 0) { "maxSizeBytes must be > 0, got $maxSizeBytes" }
+    }
+
     public companion object {
         /** Default cap of 150 MB. */
         public const val DEFAULT_MAX_SIZE_BYTES: Long = 150L * 1024 * 1024
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/StreamCacheConfig.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/StreamCacheConfig.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.cache
+
+/**
+ * Bundles the per-cache configurations exposed by the Stream Chat SDK.
+ *
+ * Pass an instance to [io.getstream.chat.android.client.ChatClient.Builder.cacheConfig] to configure
+ * the on-disk caches.
+ *
+ * @param video Configuration for the video playback cache used by SDK.
+ */
+public data class StreamCacheConfig(
+    public val video: VideoCacheConfig? = null,
+)
+
+/**
+ * Configuration for the on-disk cache used when streaming video attachments.
+ *
+ * Wrap an instance in [StreamCacheConfig] and pass it to
+ * [io.getstream.chat.android.client.ChatClient.Builder.cacheConfig] to opt in. When the cache is
+ * enabled, replaying or seeking within a previously watched video reuses cached byte ranges
+ * instead of re-downloading from the CDN.
+ *
+ * @param maxSizeBytes Soft cap on cache size; LRU eviction kicks in once exceeded. Files larger
+ * than this cap are not effectively cached. Size [maxSizeBytes] to comfortably exceed the
+ * largest expected video.
+ */
+public data class VideoCacheConfig(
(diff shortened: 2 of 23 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
```diff
@@ -62,12 +63,16 @@ public class MediaPreviewActivity : AppCompatActivity() {
...
         val title = intent.getStringExtra(KEY_TITLE) ?: ""
+        val mimeType = intent.getStringExtra(KEY_MIME_TYPE)
+        val type = intent.getStringExtra(KEY_TYPE)
 
         if (url.isNullOrEmpty() || ChatClient.isInitialized.not()) {
             finish()
             return
         }
 
+        val enableVideoCache = isVideoContent(mimeType, type)
```
gpunto: I'd suggest renaming this val and similar parameters to `isVideo` or something along those lines. With the current naming, a reader might think that this enables caching, but caching is still gated by the config down in `createPlayer`. WDYT?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cdn/internal/StreamMediaDataSource.kt
```diff
@@ -39,12 +43,24 @@ public object StreamMediaDataSource {
...
      * for URL rewriting and header injection. Local URIs (file://, content://) pass through unchanged.
      *
+     * When a [videoCache] is provided, it wraps the resulting factory so video byte ranges are
+     * served from disk on subsequent reads. The cache layer is the outermost wrapper, keyed by the
+     * raw `dataSpec.uri`, so cache lookups are not affected by CDN URL rewriting.
+     *
      * @param context The context used to create the base data source.
      * @param cdn Optional custom CDN for transforming network requests.
+     * @param videoCache Optional disk cache for video playback. Callers should pass `null` for
+     * non-video content (e.g. audio attachments, voice recordings) so audio bytes do not occupy
```
gpunto: I'm wondering if this is actually desirable feature-wise. Audio is generally lighter than video, but isn't it still good to cache it vs redownloading it every time? 🤔 

Possibly out of scope, since the original inquiry was about video.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
```diff
@@ -423,6 +438,22 @@ public class StreamFileManager {
...
+                Result.Success(Unit)
+            } else if (directory.deleteRecursively()) {
+                Result.Success(Unit)
+            } else {
+                Result.Failure(Error.GenericError("Could not clear video cache directory."))
+            }
+        } catch (e: Exception) {
+            Result.Failure(Error.ThrowableError("Could not clear video cache directory.", e))
+        }
+    }
```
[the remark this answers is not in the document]
gpunto: @VelikovPetar is this a valid concern?

gpunto (commented):
Looks good, just a couple of doubts and a nitpick

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cdn/internal/StreamMediaDataSource.kt
VelikovPetar: To be honest, I don't see it as a big issue to cache them as well. My main concern is that I would like to avoid adding granularity of the public config API (ex. `cacheConfig.video` vs `cacheConfig.audio`). Maybe in this first iteration we can re-use the video config for these other files which are using the ExoPlayer?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
VelikovPetar: Yeah we can rename it - if we decide to split the caching (video vs non-video)

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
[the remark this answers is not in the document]
VelikovPetar: I believe that we shouldn't fully release the VideoCache - if we release it, it means that our `VideoMediaCache` instance will reference an outdated `SimpleCache`. I think we should instead just clear the cache dir via the `cache.removeResource(key)` method, and not directly via `deleteRecursively`. Currently running some tests related to this.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cdn/internal/StreamMediaDataSource.kt
gpunto: That sounds good to me!

VelikovPetar: Ok, reworking it like this

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/internal/VideoMediaCache.kt
```diff
@@ -0,0 +1,122 @@
...
+            instances.remove(dirPath)
+            try {
+                cache.release()
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                logger.e(e) { "[release] failed to release SimpleCache" }
+            }
+            try {
+                databaseProvider.close()
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                logger.e(e) { "[release] failed to close StandaloneDatabaseProvider" }
```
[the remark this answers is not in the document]
VelikovPetar: `release()` is annotated `@VisibleForTesting` and only called from `@After` blocks between test cases — production keeps the cache alive for the process lifetime, so the described failure mode isn't reachable outside tests. In the test path, an eager `instances.remove()` is actually preferable — it guarantees the next test's `create()` builds a fresh cache instead of reusing a partially-torn-down one. Given this is test-only code, I'd prefer to leave it as-is. Thanks for flagging!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/internal/VideoMediaCache.kt
```diff
@@ -0,0 +1,122 @@
...
+                    }
+                    return@synchronized existing
+                }
+                val dbProvider = StandaloneDatabaseProvider(appContext)
+                val simpleCache = SimpleCache(
+                    cacheDir,
+                    LeastRecentlyUsedCacheEvictor(config.maxSizeBytes),
+                    dbProvider,
+                )
+                VideoMediaCache(simpleCache, dbProvider, key).also { instances[key] = it }
```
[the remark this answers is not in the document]
VelikovPetar: The only caller of `create()` is `ChatClient.build()`, which always passes `StreamFileManager().getVideoCache(appContext)` — that's `context.cacheDir.resolve("stream_video_cache")` — which is already a canonical, alias-free path. The singleton guarantee holds for our usage, and since `create()` is internal (`@InternalStreamChatApi`) I'd prefer not to add a `canonicalFile` normalization for a scenario that isn't reachable from the SDK's own wiring. Thanks for flagging!

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/StreamCacheConfig.kt
```diff
@@ -0,0 +1,50 @@
...
+ * than this cap are not effectively cached. Size [maxSizeBytes] to comfortably exceed the
+ * largest expected video.
+ */
+public data class VideoCacheConfig(
+    public val maxSizeBytes: Long = DEFAULT_MAX_SIZE_BYTES,
+) {
+    public companion object {
+        /** Default cap of 150 MB. */
+        public const val DEFAULT_MAX_SIZE_BYTES: Long = 150L * 1024 * 1024
+    }
```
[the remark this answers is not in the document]
VelikovPetar: Fixed in 372ab0f9a5 — added `init { require(maxSizeBytes > 0) }` on `VideoCacheConfig`, with unit tests in `VideoCacheConfigTest` covering the default, positive, zero, and negative cases.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cdn/internal/StreamMediaDataSource.kt
```diff
@@ -39,12 +43,24 @@ public object StreamMediaDataSource {
      * When a [CDN] is provided, HTTP/HTTPS requests are transformed through [CDN.fileRequest]
      * for URL rewriting and header injection. Local URIs (file://, content://) pass through unchanged.
      *
+     * When a [videoCache] is provided, it wraps the resulting factory so video byte ranges are
+     * served from disk on subsequent reads. The cache layer is the outermost wrapper, keyed by the
+     * raw `dataSpec.uri`, so cache lookups are not affected by CDN URL rewriting.
```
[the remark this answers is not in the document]
VelikovPetar: Fixed in 372ab0f9a5 — KDoc now states the cache is keyed by `dataSpec.uri` with query parameters stripped, so cache lookups are unaffected by CDN URL rewriting and by rotating pre-signed query/signature values.

gpunto (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/file/StreamFileManager.kt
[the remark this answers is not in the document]
VelikovPetar: Yes, valid concern — fixed in 372ab0f9a5. Added `VideoMediaCache.clear()` which iterates `cache.keys` and calls `SimpleCache.removeResource(key)` for each entry; this empties the cache **in place** without touching the directory Media3's live `SimpleCache` still owns (deleting the directory out from under it would corrupt the on-disk index and lock).

Wiring:
- `StreamFileManager.clearAllCache` no longer touches the video cache directory (the invariant is guarded by a new test).
- `ChatClient.clearCacheAndTemporaryFiles` routes video-cache clearing through `videoCache?.clear()` when a live cache is present, falling back to `fileManager.clearVideoCache(context)` when no cache is configured (so leftover files from a prior opt-in still get cleaned).

This keeps subsequent playback working immediately after the clear (verified end-to-end via the new `clear removes cached bytes from disk and keeps the cache functional` test).

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/MediaPreviewActivity.kt
VelikovPetar: Resolving because we decided to include other media as well.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cdn/internal/StreamMediaDataSource.kt
VelikovPetar: Done in 372ab0f9a5 — audio/voice-recording playback is now routed through the same `videoCache` (wired via `ChatClient.build()` passing `videoCache` into `StreamMediaDataSource.factory(...)` for the `AudioPlayer` pipeline as well). Kept the config surface as-is (single `VideoCacheConfig`), no new granularity introduced.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/gallery/AttachmentMediaActivity.kt
```diff
@@ -138,8 +139,9 @@ public class AttachmentMediaActivity : AppCompatActivity() {
 
     @OptIn(UnstableApi::class)
     private fun createPlayer(): Player {
-        val cdn = ChatClient.instance().cdn
-        val dataSourceFactory = StreamMediaDataSource.factory(this, cdn)
+        val client = ChatClient.instance()
+        val videoCache = client.videoCache.takeIf { isVideoContent() }
```
andremion: This still gates the cache to video only, but the audio player pipeline and the Compose `MediaPreviewActivity` now cache all media after the earlier discussion on audio. So an audio file opened in this XML viewer would skip the cache while the same file is cached elsewhere. Is the `isVideoContent()` gate still meant to be here, or is it a leftover from the video-only version?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
```diff
@@ -1519,12 +1523,20 @@ internal constructor(
     public fun clearCacheAndTemporaryFiles(context: Context): Call<Unit> =
         CoroutineCall(clientScope) {
             logger.d { "[clearCacheAndTemporaryFiles] Clearing all cache and temporary files" }
+            // Clear video cache: in-place via the live cache when opted in (keeps the SimpleCache
+            // alive so playback continues to work), or by deleting the directory when no live
+            // cache owns it.
+            val videoCacheResult = videoCache?.let {
+                it.clear()
+                Result.Success(Unit)
+            } ?: fileManager.clearVideoCache(context)
```
andremion: Small edge case here: this fallback assumes that if the current client has no `videoCache`, no live cache owns the directory. But the `instances` registry keeps a previous build's `SimpleCache` alive for the whole process, so if someone builds with the cache on, then rebuilds `ChatClient` without `cacheConfig`, this branch would `deleteRecursively` the directory while the old cache still holds it. That's the same corruption the in-place `clear()` was added to avoid. Probably rare, but do you think it's worth guarding, or is rebuild-without-cache out of scope for now?

andremion (commented):
Left two small comments inline. Also the PR description still says audio bypasses the cache, which no longer matches the code after the audio discussion, might be worth updating.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/gallery/AttachmentMediaActivity.kt
VelikovPetar: Good catch — it was a leftover from the video-only version. Fixed in 79764579fe: removed the `isVideoContent()` gate (and the now-unused helper) so this viewer passes `client.videoCache` directly, matching the audio pipeline, the Compose `MediaPreviewActivity`, and the gallery video fragment. Audio opened here is now cached too.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
VelikovPetar: Valid concern, worth guarding — fixed in 79764579fe. The check now keys off the process-wide registry instead of the current client's `videoCache`: added `VideoMediaCache.clearAll()`, which clears every live cache in place and returns whether any existed. `clearCacheAndTemporaryFiles` only falls back to `deleteRecursively` when `clearAll()` reports no live cache owns the directory, so the rebuild-without-cache case you described can no longer delete a directory a prior build's `SimpleCache` still holds. Covered by two new tests in `StreamMediaDataSourceCacheIntegrationTest`.

andremion (approved):

PR #6535, opened 2026-07-02
Prevent `MessageList` item jump during IME animation

VelikovPetar:
### Goal

Fix [AND-1285]. When opening the on-screen keyboard on `MessagesScreen` with only a few messages in the channel (items fit in the viewport), the messages visibly "hopped" as the keyboard animated in — and again, more subtly, when it animated out.

### Implementation

Root cause: `Modifier.animateItem()` animates item **placement changes**. With `reverseLayout = true` + `Arrangement.Top` (defaults for the message list), items shift position slightly on every frame while the LazyColumn's viewport shrinks/grows in step with the IME animation. Without `animateItem` each frame's placement is committed instantly and stays in sync with the IME animation; with `animateItem` its placement spring lays its own animation on top of the per-frame deltas, which surfaces as the visible hop.

Fix: suppress **only** the placement portion of `Modifier.animateItem()` during the IME animation window; leave fade-in / fade-out and everything else at rest untouched.

- New `IsImeAnimating.kt` with an internal `@Composable` predicate `isImeAnimating()`. It reads `WindowInsets.imeAnimationSource` / `imeAnimationTarget` (`@ExperimentalLayoutApi`) — they diverge for the full IME animation and converge again once it settles, giving a synchronous "is animating" boolean that flips on the same frame the animation starts (a coroutine-debounced `snapshotFlow` variant was tried and lagged by one frame at the close-start edge, letting a small hop through).
- `ChatComponentFactory.messageListItemModifier()` calls `isImeAnimating()` and returns `Modifier.animateItem(placementSpec = null)` while `true`, or the original `Modifier.animateItem()` otherwise. `LocalInspectionMode.current` still short-circuits to plain `Modifier` for Paparazzi.

`Messages` is not touched — the fix lives entirely inside the factory method. Consumers who override `messageListItemModifier()` and want the same behaviour can call `isImeAnimating()` themselves once we promote it to `public`.

No changes to public API: the new declaration is `internal`, and `messageListItemModifier()`'s signature is unchanged. `./gradlew :stream-chat-android-compose:apiDump` produced no diff.

### Trade-offs

- Placement animation is suppressed for the entire ~300 ms IME animation window. If a new message arrives inside that window, its slide-in won't animate (fade-in still does). Uncommon in practice.
- Only patches the default `messageListItemModifier()`. Consumers who override it and use their own `Modifier.animateItem()` will still see the hop; `isImeAnimating()` is currently `internal`, so they can't opt in. Ready to promote to `public` if this becomes a real ask.
- Non-IME viewport shrinks (rotation, split-screen resize) are not covered by this gate.

### Performance

Measured on-device on an Android 15 emulator with a 3-item channel:
- `isImeAnimating` recomposes exactly **twice per IME animation** (start + settle), not per frame.
- Item modifier scope re-runs ~4 × N per animation (N = visible items) — the second pass is one frame later and is the LazyColumn re-measure after the IME inset changes. Only the modifier lambda re-runs; `itemContent(item)` has its own restart scope and is not recomposed.

### UI Changes

No UI changes at rest. The visible difference is that message items no longer hop during keyboard open/close on channels where items fit in the viewport.

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/01aa9f38-45d1-4113-8577-366e67f80b35"></video> | <video src="https://github.com/user-attachments/assets/206f0029-0c5d-46fc-abfe-147fdd0b0e7d"></video> |

### Testing

- [ ] Open a channel with 2–5 messages via `MessagesScreen`. Tap composer → keyboard opens smoothly, no item hop. Tap outside → keyboard closes, no hop.
- [ ] Repeat on a channel with many messages (viewport full) — behaviour unchanged.
- [ ] Send a new message with `Arrangement.Top` on a small channel — placement animation still works at rest.
- [ ] `./gradlew :stream-chat-android-compose:spotlessCheck :stream-chat-android-compose:detekt :stream-chat-android-compose:apiDump :stream-chat-android-compose:testDebugUnitTest` — all green locally.
- [ ] Paparazzi snapshots unchanged (`messageListItemModifier()` still returns plain `Modifier` under `LocalInspectionMode.current`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message list behavior while the keyboard is animating, reducing unwanted item movement.
  * Kept list item animations disabled during keyboard transitions for a smoother chat experience.
  * Preserved existing behavior in preview mode and after the keyboard finishes animating.

### DIFF
--- a/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
+++ b/stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
@@ -28,6 +28,7 @@ import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyItemScope
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.derivedStateOf
@@ -49,7 +50,9 @@ import androidx.compose.ui.unit.dp
 import io.getstream.chat.android.compose.handlers.LoadMoreHandler
 import io.getstream.chat.android.compose.ui.components.LoadingIndicator
 import io.getstream.chat.android.compose.ui.theme.ChatTheme
+import io.getstream.chat.android.compose.util.LocalIsImeAnimating
 import io.getstream.chat.android.compose.util.isAppInForegroundAsState
+import io.getstream.chat.android.compose.util.rememberIsImeAnimating
 import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.ui.common.state.messages.list.HasMessageListItemState
 import io.getstream.chat.android.ui.common.state.messages.list.MessageFocused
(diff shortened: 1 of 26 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
```diff
@@ -144,77 +147,80 @@ public fun Messages(
     val isLoadingMoreOldMessages = messagesState.isLoadingOlderMessages
 
     val density = LocalDensity.current
+    val isImeAnimating = rememberIsImeAnimating()
```
gpunto: Will this not work if we put it directly inside `messageListItemModifier`?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/util/IsImeAnimatingAsState.kt
```diff
@@ -0,0 +1,56 @@
...
+ * Returns `true` while an IME animation is in progress.
+ *
+ * `imeAnimationSource` and `imeAnimationTarget` diverge for the whole duration of the
+ * animation and converge again once it settles. Reading them synchronously in composition
+ * means the value flips on the same frame the animation starts, avoiding a one-frame lag
+ * that a coroutine-based debounce approach would incur at the transition edges.
+ */
+@Composable
+@OptIn(ExperimentalLayoutApi::class)
+internal fun rememberIsImeAnimating(): Boolean {
```
gpunto: Should it be called just `isImeAnimating` since we don't have any `remember`? 🤔

VelikovPetar: Ugh this was named `remember...` in a previous iteration of the fix, completely forgot to remove it after removing the actual remember

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
VelikovPetar: Yes it will 😮‍💨 
Tried several other fixes and some of them used some stuff from the `Messages` composable, so I had to use a `LocalComposition...` to pass the value to the `Modifier` here. Then I didn't realise that we can clean it up...

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/util/IsImeAnimatingAsState.kt
VelikovPetar: addressed: https://github.com/GetStream/stream-chat-android/pull/6535/commits/5a8774fb75f77d71f6fd95c16a196c1e3443d203

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/messages/list/Messages.kt
VelikovPetar: addressed: https://github.com/GetStream/stream-chat-android/pull/6535/commits/5a8774fb75f77d71f6fd95c16a196c1e3443d203

gpunto (approved):

PR #6536, opened 2026-07-02
compose: Improve message view screen-reader focus order

andremion:
### Goal

Make the TalkBack focus order in the message view predictable and match the visual reading order, without changing the reversed message list layout that chat relies on. Before this change the reading cursor could enter the header on the subtitle, jump from the composer straight to the header and skip the messages, and, moving backwards from the composer, land inside a reaction or an attachment of a neighbouring message instead of on the message itself.

Resolves AND-1280

### Implementation

- **Header**: group the title and subtitle so the header reads back button, heading, subtitle, avatar.
- **Regions**: mark the header, message list, and composer as ordered traversal regions (header, then list, then composer), because a Scaffold does not reliably sort its scrollable content between the top and bottom bars for screen readers.
- **Message list (the core fix)**: TalkBack rebuilds its own traversal from the Compose semantics tree order and silently drops any ordering hint that points from a node to its own descendant (a message cell hinted before its inner button). A reversed list composes newest-first but reads oldest-first, so those hints were constantly fought and the message cells got stranded, which is what made backward navigation skip messages and jump around. Each lazy item now carries a `zIndex` that mirrors its traversal index, so the semantics tree order equals the reading order and TalkBack has nothing to rearrange. The reactions bubble is pinned to its tree position for the same reason.
- **Docs**: add `stream-chat-android-compose/ACCESSIBILITY.md`, a playbook covering the leaf-first principles, the semantics decision tree, and the traversal-order model with the diagnostics used to verify it.

The interactive parts of a message (open attachment, vote, View Results, reactions) stay individually reachable. They are not collapsed into a single stop, so poll voting and attachment actions remain usable.

No public API changes.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th></th>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>Bottom-up (swipe back from the composer)</td>
<td><video src="https://github.com/user-attachments/assets/74c59343-f449-4ef2-ae9f-9eec95abe835" controls muted /></td>
<td><video src="https://github.com/user-attachments/assets/a30a3fe3-c7cc-4074-a3a1-6df27328b1d3" controls muted /></td>
</tr>
<tr>
<td>Top-down (swipe forward from the header)</td>
<td><video src="https://github.com/user-attachments/assets/427d4c83-5d3f-4b92-ac64-4768844e196d" controls muted /></td>
<td><video src="https://github.com/user-attachments/assets/57ed4c68-36d4-4d61-b0f5-cb8b5c215488" controls muted /></td>
</tr>
</tbody>
</table>

### Testing

Manual, with TalkBack enabled on a device or emulator:

1. Open a channel with a mix of messages (for example a group that has a poll, an image message, and a message with reactions).
2. On entry, focus lands on the composer input (existing behaviour).
3. Sweep through the screen and confirm the order matches the visual order: the header reads back button, heading, subtitle, avatar; the regions read header, then messages, then composer.
4. Swipe left (backwards) from the composer: attachment picker, then the newest message, then older messages in order, and the header only after the oldest loaded message. Focus should not jump between messages or land inside a neighbour's reaction or attachment.
5. On a poll message, focus "View Results" and swipe left: focus lands on the poll message, not on a neighbouring message.
6. Confirm the interactive parts stay reachable and actionable: double-tap a reactions row, a poll option, "View Results", and an attachment.

One Compose platform limitation is out of scope and was accepted with QA: TalkBack speaks a region-name fragment when the reading cursor crosses between the header, list, and composer regions, which cannot be suppressed.

## Summary by CodeRabbit

* **Accessibility**
  * Improved screen-reader navigation across channel screens, message lists, headers, reactions, and the message composer.
  * Refined reading order so content is announced more consistently, including in reversed message lists.
  * Added accessibility guidance and verification notes for future updates.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
@@ -38,6 +38,7 @@ import io.getstream.chat.android.compose.uiautomator.tapOnScreenCenter
 import io.getstream.chat.android.compose.uiautomator.typeText
 import io.getstream.chat.android.compose.uiautomator.wait
 import io.getstream.chat.android.compose.uiautomator.waitToAppear
+import io.getstream.chat.android.compose.uiautomator.waitToAppearBottomUp
 import io.getstream.chat.android.compose.uiautomator.waitToDisappear
 import io.getstream.chat.android.e2e.test.mockserver.AttachmentType
 import io.getstream.chat.android.e2e.test.mockserver.ReactionType
@@ -76,8 +77,7 @@ class UserRobot {
     }
 
     fun openContextMenu(messageCellIndex: Int = 0): UserRobot {
-        MessageList.messages.waitToAppear()
-        val messages = MessageList.messages.findObjects()
+        val messages = MessageList.messages.waitToAppearBottomUp()
         val message = if (messages.size < messageCellIndex + 1) messages.last() else messages[messageCellIndex]
         message.longPress()
         return this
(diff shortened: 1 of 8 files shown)


andremion:
@coderabbit review

gpunto (approved):

PR #6538, opened 2026-07-03
compose: Show mute icon and role badge in channel member detail sheet

andremion:
### Goal

In the Channel Info member detail sheet, the header showed only the member's name. It did not show whether the member was muted, and it did not show the member's role. The member list rows already show a mute icon and a role badge, so the sheet was inconsistent with the list and with the design.

This change makes the sheet header match the list rows and the Figma design: a mute icon when the member is muted, and a role badge (Owner, Admin, Moderator, or a custom channel role such as VIP) inline trailing the name.

The mute icon in the member list rows was added earlier in #6515. This completes the same behavior in the detail sheet.

Resolves AND-1287

### Implementation

- `ChannelInfoMemberInfoModalSheetTopBar` now renders the name through a small `ChannelInfoMemberInfoModalSheetName` composable that adds the mute icon (shown when muted) and the role badge inline after the name.
- The role badge reuses the existing `GroupChannelInfoMemberTrailingContent` component factory, so the role resolution (owner, admin, moderator, custom role) lives in one place and integrator overrides apply to both the list and the sheet.
- `ChannelInfoMemberViewState.Content` gains an `isOwner` field, derived in `ChannelInfoMemberViewController` from the channel creator (`channelData.createdBy.id == memberId`). This is the same source the member list uses for its owner badge, and it is needed because owner is not a field on `Member`.
- `ChannelInfoMemberInfoModalSheetTopBarParams` gains `isMuted` and `isOwner`. Both are defaulted so the change stays additive for integrators who construct the params.

### 🎨 UI Changes

The member detail sheet header now shows the mute icon and role badge inline after the name. New Paparazzi snapshots cover a muted member with a custom role and a channel owner, in light and dark mode. See the snapshot images added in this diff.

### Testing

1. Open a group channel where you are a member with permission to view members.
2. Open Channel Info and tap a member who is not you to open the member detail sheet.
3. Tap "Mute User". Reopen the sheet for the same member. A mute icon appears inline after the name in the header.
4. Tap "Unmute User". Reopen the sheet. The mute icon is gone.
5. Open the detail sheet for the channel owner. An "Owner" badge appears after the name.
6. Open the detail sheet for a member with a role (admin, moderator, or a custom channel role). The matching role badge appears after the name.
7. Confirm the member list row and the sheet header show the same mute icon and role badge for the same member.

## Summary by CodeRabbit

* **New Features**
  * Added owner and mute status to member info displays, including the channel owner indicator.
  * Expanded the member info sheet to show additional member role details and updated previews.

* **Bug Fixes**
  * Improved member info state handling so owner status is consistently reflected across the UI.

* **Tests**
  * Added snapshot coverage for muted members and channel owners in light and dark modes.
  * Updated state tests to account for the new owner flag.


andremion:
@coderabbit review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/state/channel/info/ChannelInfoMemberViewState.kt
```diff
@@ -38,11 +38,13 @@ public sealed interface ChannelInfoMemberViewState {
...
      * @param isMuted Whether the member's user is muted by the current user.
      * @param isBlocked Whether the member's user is blocked by the current user.
+     * @param isOwner Whether the member is the channel owner (its creator).
      */
     public data class Content(
         val member: Member,
         val capabilities: Set<String>,
         val isMuted: Boolean,
         val isBlocked: Boolean,
+        val isOwner: Boolean,
```
gpunto: This is formally a source breaking change, should we add a default?

andremion: Good catch, added a default. Thanks!

gpunto (approved):

PR #6539, opened 2026-07-03
Prevent `MessageList` item jump during IME animation

VelikovPetar:
### Goal

Port of v6 commit `13a8be56c7` — [Prevent `MessageList` item jump during IME animation (#6535)](https://github.com/GetStream/stream-chat-android/pull/6535) — to `develop`.

Fixes **AND-1285**. When opening the on-screen keyboard on `MessagesScreen` with only a few messages (items fit in the viewport), the Compose `MessageList` items visibly "hop" as the keyboard animates in — and again, more subtly, when it animates out.

Root cause: `Modifier.animateItem()` animates item **placement** changes. With `reverseLayout = true` + `Arrangement.Top`, items shift position slightly on every frame while the LazyColumn viewport shrinks/grows in step with the IME animation; the placement spring lays its own animation on top of those per-frame deltas, which surfaces as the visible hop.

### Implementation

Suppress **only** the placement portion of `Modifier.animateItem()` during the IME animation window; fade in/out and normal list-change placement animations (insertions, deletions, height changes) are untouched.

- New internal `isImeAnimating()` `@Composable` in `util/IsImeAnimating.kt`. It reads `WindowInsets.imeAnimationSource` / `imeAnimationTarget` (`@ExperimentalLayoutApi`), which diverge for the full IME animation and converge once it settles — a synchronous "is animating" boolean that flips on the same frame the animation starts (a coroutine-debounced variant lagged by one frame at the close-start edge).
- `ChatComponentFactory.messageListItemModifier(params)` now returns `Modifier.animateItem(placementSpec = null)` while `isImeAnimating()` is `true`, the original `Modifier.animateItem()` otherwise, and plain `Modifier` under `LocalInspectionMode.current` (Paparazzi).

**Port note:** This was a **manual port**, not a clean cherry-pick. On `develop`, `messageListItemModifier` takes a `MessageListItemModifierParams` argument that v6 lacks, so a raw `git cherry-pick` conflicts on the method signature line. The fix body is identical to v6; `IsImeAnimating.kt` was added verbatim.

**No public API change:** `isImeAnimating()` is `internal` and the `messageListItemModifier(params)` signature is unchanged. `./gradlew :stream-chat-android-compose:apiDump` produced no diff.

### UI Changes

No UI changes at rest. The visible difference is that message items no longer hop during keyboard open/close on channels where items fit in the viewport. See the before/after videos on the original v6 PR #6535.

### Testing

- `./gradlew :stream-chat-android-compose:detekt` — green
- `./gradlew :stream-chat-android-compose:spotlessApply` — no reformatting
- `./gradlew :stream-chat-android-compose:apiDump` — no API diff
- `./gradlew :stream-chat-android-compose:testDebugUnitTest` — green (includes Paparazzi snapshots, unchanged)

Manual: open a channel with 2–5 messages via `MessagesScreen`, tap the composer → keyboard opens smoothly with no item hop; tap outside → keyboard closes with no hop. Repeat on a channel with a full viewport (behaviour unchanged), and send a new message on a small channel (placement animation still works at rest).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Improved message list animations during keyboard transitions for a smoother chat experience.
  * Keyboard-aware behavior now adjusts item motion while the on-screen keyboard is opening or closing.
* **Bug Fixes**
  * Reduced awkward list movement when the keyboard changes height, especially during animated transitions.
  * Kept animations stable in inspection mode for more predictable preview behavior.


andremion (commented):
LGTM. One optional note below.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/util/IsImeAnimating.kt
```diff
@@ -0,0 +1,42 @@
...
+ * (either open or closed).
+ *
+ * `imeAnimationSource` and `imeAnimationTarget` diverge for the whole duration of the
+ * animation and converge again once it settles. Reading them synchronously in composition
+ * means the value flips on the same frame the animation starts, avoiding a one-frame lag
+ * that a coroutine-based debounce approach would incur at the transition edges.
+ */
+@Composable
+@OptIn(ExperimentalLayoutApi::class)
+internal fun isImeAnimating(): Boolean {
```
andremion, line 41: The neighbouring `isKeyboardVisibleAsState()` avoids `WindowInsets.ime` because it can report wrong values on some OEM ROMs (Xiaomi/MIUI), see its KDoc. This helper reads `imeAnimationSource` / `imeAnimationTarget` from the same IME inset channel, so on those ROMs (or when the host window isn't set up for edge-to-edge inset dispatch) it may never report `source != target` and the fix just won't apply. It degrades gracefully, so not blocking. Worth a one-line note in the KDoc to document the dependency?

andremion (approved):

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/util/IsImeAnimating.kt
VelikovPetar: Fixed in ac44c10124 — added a KDoc note explaining that on OEM ROMs (e.g. Xiaomi/MIUI) that don't dispatch IME inset animations, `source` and `target` always match, so the function always returns `false` and the suppression silently degrades to a no-op.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/util/IsImeAnimating.kt
```diff
@@ -0,0 +1,42 @@
...
+    val density = LocalDensity.current
+    val source = WindowInsets.imeAnimationSource.getBottom(density)
+    val target = WindowInsets.imeAnimationTarget.getBottom(density)
+    return source != target
+}
```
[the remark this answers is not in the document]
andremion: Fair observation, but I don't think it's worth acting on in this PR.

The read does recompose per visible item per frame during the IME animation, but that's a handful of items for a short window. Per the root cause described here, the list placement is already changing every frame while the viewport tracks the keyboard animation, so the marginal cost of recomputing the modifier is small next to the layout work already happening. It's also smallest exactly where this fix applies: small channels where items fit in the viewport have the fewest visible items.

Hoisting doesn't remove the per-frame recomposition, it moves it to the ancestor. To make it a net win you'd have to isolate the inset read in a minimal scope and propagate only the distinct boolean. The coroutine/distinct-until-changed variant was already rejected on purpose because it lagged one frame at the close edge, which is a real timing tradeoff.

This is also a straight port of the merged v6 PR #6535, so I'd rather not change the shape here. If profiling shows actual jank during the animation, I'll raise it as a follow-up against both branches.

andremion (approved):

PR #6543, opened 2026-07-03
Migrate CompletableResponse to generated Response

gpunto:
### Goal

Replace the hand-written `CompletableResponse` DTO with the generated `Response` model. First slice of the incremental OpenAPI model migration: generated models keep their generated package `io.getstream.chat.android.network.models` as `internal` classes inside the client module (no separate module, mirroring the generator output so a future regen is drag-and-drop), one endpoint/model at a time.

### Implementation

- Add the generated `Response` under `io.getstream.chat.android.network.models` (`internal`).
- Swap every `CompletableResponse` usage to `Response` across the api2 endpoint interfaces and `RetrofitCdnApi`, and delete the two hand-written `CompletableResponse` classes.
- Update the affected tests and drop the stale `CompletableResponse` baseline-profile entries.

### Testing

- Project-wide `spotlessApply`, `apiDump` (no API drift), `lint`, `detekt`, and `testDebugUnitTest` all pass.
- Verified on device: `markAllRead` (backed by `RetrofitCall<Response>`) parses the real wire response and completes successfully.


stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/Response.kt
```diff
@@ -14,11 +14,23 @@
...
+ * Basic response information
+ */
 
-@JsonClass(generateAdapter = true)
-internal data class CompletableResponse(
-    val duration: String,
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class Response(
+    @Json(name = "duration")
+    val duration: kotlin.String,
```
andremion, line 35: Small behavior note on the merge: the old `api.models.CompletableResponse` used by `RetrofitCdnApi` had `duration: String = ""`, so a CDN delete response that omitted `duration` still parsed. The generated `Response` has no default, so with the Moshi adapter `duration` is now required, and a delete response without it would fail to parse instead of returning success. The api2 endpoints already had no default, so only the CDN delete paths (`deleteFile`, `deleteImage`) change here.                                                                                                                                                                                                                         
Do the delete file/image responses always include `duration`? The on-device check covered `markAllRead` (api2), which was already non-defaulted, so it does not exercise this path. If the CDN responses are guaranteed to send `duration`, this is fine as is.
Another point: do we need the full package declaration in `kotlin.String`?

andremion: It seems the backend will always send the duration.
I'm approving it now.

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/Response.kt
gpunto: Yep! I also double checked the backend now and ran a manual test to confirm the behavior.

For the full package declaration: it's the generator output as it works today. I haven't touched that part, but I'll see if we can do that (at least for stdlib types) as part of the generator improvements I'm working on.

PR #6547, opened 2026-07-06
Migrate device endpoints to generated models

gpunto:
### Goal

Migrate the `/devices` endpoints to generated models. Part of the incremental OpenAPI model migration.

Built on top of the Serializer foundation PR (this is the first slice with a generated enum, `PushProvider`).

### Implementation

- Add generated `DeviceResponse`, `ListDevicesResponse`, and `CreateDeviceRequest` (with the typed `PushProvider` sealed enum); delete the hand-written `DevicesResponse` and `AddDeviceRequest`.
- Register `CreateDeviceRequest.PushProvider.PushProviderAdapter()` in `Serializer` (the one-line-per-enum pattern the foundation enables).
- Point the `/devices` endpoints at the generated types (real names, no aliases): `getDevices` returns `ListDevicesResponse`, `addDevice` sends `CreateDeviceRequest`.
- Keep the hand-written `DeviceDto` for user-embedded devices (parsed from WS events / user payloads via `DeviceDto.toDomain()`, serialized via `Device.toDto()`): the generated `DeviceResponse` requires `created_at`/`user_id`, which the lean WS device shape omits, so switching those to `DeviceResponse` is deferred to the user migration.

### Testing

- Project-wide `spotlessApply`, `apiDump` (no API drift), `lint`, `detekt`, and `testDebugUnitTest` all pass.
- Verified on device: `addDevice` (generated request + `PushProvider` enum) succeeds, `getDevices` parses 22 real `DeviceResponse`s (`roundTrip=true`), `deleteDevice` succeeds.


gpunto:
@coderabbitai review

andremion (commented):
Thanks for this, reads clean overall. A couple of small non-blocking notes inline, and one on the description: it mentions `Device.toDto()` supplying placeholder `createdAt`/`userId`, but in the final diff `toDto()` still returns the 3-field `DeviceDto` and `CreateDeviceRequest` doesn't have those fields. Looks like a leftover from an earlier pass, maybe worth dropping to keep it clear.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/CreateDeviceRequest.kt
```diff
@@ -0,0 +1,89 @@
...
+                else -> Unknown(s)
+            }
+        }
+        object Apn : PushProvider("apn")
+        object Firebase : PushProvider("firebase")
+        object Huawei : PushProvider("huawei")
+        object Xiaomi : PushProvider("xiaomi")
+        data class Unknown(val unknownValue: kotlin.String) : PushProvider(unknownValue)
+
+        class PushProviderAdapter : JsonAdapter<PushProvider>() {
```
andremion, line 76: Nice to see the enum adapter land. One thing: I don't think it's exercised through Moshi yet. `MoshiChatApiTest` only checks the request object against a mock, so `fromString` and the JSON output aren't covered. Since the next slices build on this pattern, would a small round-trip test here be worth adding?

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/DeviceResponse.kt
```diff
@@ -0,0 +1,60 @@
...
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class DeviceResponse(
+    @Json(name = "created_at")
+    val createdAt: java.util.Date,
+
+    @Json(name = "id")
+    val id: kotlin.String,
+
+    @Json(name = "push_provider")
+    val pushProvider: kotlin.String,
```
andremion, line 41: Small question: any reason this stays a `String` while `CreateDeviceRequest.pushProvider` uses the typed enum? I'm guessing it just mirrors the spec, so no change needed, just wanted to check my understanding.

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/DeviceResponse.kt
gpunto: It's exactly as you say, it's just how the spec is defined in the [go source](https://github.com/GetStream/chat/blob/e37961be6e313fb015ff0814aed5dc7500bdea24/lib/combined/controllers/common/commonpayloads/devices.go#L53) (lacks the `oneof` that the request instead has).

gpunto:
> Thanks for this, reads clean overall. A couple of small non-blocking notes inline, and one on the description: it mentions `Device.toDto()` supplying placeholder `createdAt`/`userId`, but in the final diff `toDto()` still returns the 3-field `DeviceDto` and `CreateDeviceRequest` doesn't have those fields. Looks like a leftover from an earlier pass, maybe worth dropping to keep it clear.

Yep, leftovers from an earlier, wrong implementation

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/CreateDeviceRequest.kt
gpunto: Added in https://github.com/GetStream/stream-chat-android/pull/6547/changes/48e71e7fc85b45a8fe1644b93b0f4c54f50027c7

Edit: also parameterized them in the next commit

PR #6548, opened 2026-07-06
Introduce Serializer foundation for generated models

gpunto:
### Goal

Introduce the `network.infrastructure.Serializer` foundation that generated models are (de)serialized through, and route `MoshiChatParser` via it. Foundational for the incremental OpenAPI model migration: every generated-model slice builds on this, and each enum-bearing slice adds one adapter registration to `Serializer`.

### Implementation

- Vendor `io.getstream.chat.android.network.infrastructure` (`internal`, verbatim generator output): `Serializer` + `IsoDateAdapter`. The generator emits base adapters only for types the spec uses, and chat has no `BigDecimal`/`BigInteger`/`ByteArray`/`URI`/`UUID` fields, so `IsoDateAdapter` is the only one. `Serializer` starts with it + `KotlinJsonAdapterFactory`; enum adapters are added one per slice as models are migrated.
- All Moshi-based wire date handling now routes through the generated `IsoDateAdapter`: `MoshiChatParser` builds from `Serializer.moshi.newBuilder()` (layering the existing hand-written adapters on top), and `DirectEventParser` + the offline Room DB converter (`ConverterMoshiInstance`) register it directly. The hand-written `DateAdapter` — functionally identical (same ITU parser, formats, null/cache handling), so parsing is unchanged — is removed, and its unit test becomes `IsoDateAdapterTest`. (`StreamDateFormatter` stays for its non-Moshi-adapter uses — `ExactDateAdapter`, `NewMessageEventAdapter`, and `ChatClient` sync formatting — which converge on `IsoDateAdapter` as their DTOs adopt `Date` fields.)
- The `KotlinJsonAdapterFactory` is emitted by the generator's `Serializer` (added last, as a fallback), so it's kept as-is rather than adding a chat-specific exclusion. It only affects a non-codegen Kotlin class with no registered adapter — those now parse reflectively instead of throwing; all codegen DTOs and hand-written adapters take precedence, so existing types are unaffected. Longer term it should be dropped from all SDKs so we rely solely on codegen adapters.

### Testing

- Project-wide `spotlessApply`, `apiDump` (no API drift), `lint`, `detekt`, and `testDebugUnitTest` all pass — the full parser test suite validates the `DateAdapter -> IsoDateAdapter` switch across every existing DTO.

## Summary by CodeRabbit

* **New Features**
  * Improved date handling in network data processing, with more consistent ISO-8601 formatting and parsing.

* **Bug Fixes**
  * Increased reliability when reading and writing date values, including better handling of empty or malformed inputs.
  * Improved compatibility across different Android versions for date serialization.


stream-chat-android-client/src/main/java/io/getstream/chat/android/network/infrastructure/Serializer.kt
```diff
@@ -0,0 +1,38 @@
...
+
+package io.getstream.chat.android.network.infrastructure
+
+import com.squareup.moshi.Moshi
+
+internal object Serializer {
+    @JvmStatic
+    internal val moshiBuilder: Moshi.Builder = Moshi.Builder()
+        .add(io.getstream.chat.android.network.infrastructure.IsoDateAdapter())
+        .addLast(com.squareup.moshi.kotlin.reflect.KotlinJsonAdapterFactory())
```
andremion, line 32: Small thing: this pulls a reflective `KotlinJsonAdapterFactory` into `MoshiChatParser`, which didn't have one before. It sits after the custom adapters and codegen DTOs are unaffected, so nothing changes today. The one new behavior is that a non-codegen Kotlin class with no adapter now parses reflectively instead of throwing. Maybe worth a line in the description so it's clearly intentional?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
```diff
@@ -63,8 +63,7 @@ internal class MoshiChatParser(
 ) : ChatParser {
 
     private val moshi: Moshi by lazy {
-        Moshi.Builder()
-            .addAdapter(DateAdapter())
+        Serializer.moshi.newBuilder()
```
andremion, line 66: Any reason `DirectEventParser` and `ConverterMoshiInstance` stay on `DateAdapter`? It means two date implementations run for now, which is fine since they're equivalent. A later slice?

andremion (approved):
LGTM, the date-adapter swap looks equivalent and the parser tests cover it. Two optional notes inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/MoshiChatParser.kt
gpunto: Good catch. The idea is to converge all wire date handling on the generated IsoDateAdapter and drop StreamDateFormatter/DateAdapter. I think I can include it here, no strong reason to defer it 👍

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/infrastructure/Serializer.kt
gpunto: Yeah, let me add a line in the description. It's automatically added by the generator and I'd prefer not to introduce an exclusion for chat. Ideally at some point we should remove it completely from all SDKs and only rely on codegen adapters

andremion (approved):

PR #6549, opened 2026-07-06
compose: Anchor the audio recording hint above the composer in threads

andremion:
### Goal

When recording a voice message inside a thread, the recording hint ("Hold to record. Release to save.") and the microphone permission rationale appeared over the composer content instead of above the composer. In a thread the composer is taller (it shows the parent message and the "Also send in Channel" row), so the hint popup, which was anchored to the microphone button, landed too low and overlapped the composer.

Resolves AND-1292

### Implementation

The recording hint and the permission rationale are now rendered by the composer itself, using snackbar hosts provided through composition locals, so they anchor above the composer regardless of the composer height. This mirrors how the validation-error snackbar is already rendered by the composer.

- `MessageComposer` creates the two hosts, provides them via `LocalMessageComposerRecordingHintHostState` and `LocalMessageComposerRecordingRationaleHostState`, and renders all three composer snackbars (validation error, recording hint, permission rationale) in one place.
- `AudioRecordingButton` reads the hint host from the local instead of creating its own host and rendering the snackbar itself.
- `AudioRecordingPermission` reads the rationale host from the local.

#### Docked style

| Before | After |
| --- | --- |
| before_thread_recording_hint_docked | after_thread_recording_hint_docked |

#### Floating style

| Before | After |
| --- | --- |
| before_thread_recording_hint | after_thread_recording_hint |

### Testing

1. Open a thread (reply to a message) so the composer shows the parent message and the "Also send in Channel" row.
2. Tap and release the microphone button quickly (a tap, not a hold). The recording hint should appear directly above the composer, not overlapping it.
3. Revoke the microphone permission for the app, then tap the microphone button. The permission rationale snackbar should also appear above the composer.
4. Repeat both steps in a normal (non-thread) channel to confirm the position is unchanged there.

Unit tests: added `audio recording hint is rendered by the composer` and `audio recording permission rationale is rendered by the composer` to `MessageComposerScreenTest`, verifying the composer renders content from the provided hosts.

## Summary by CodeRabbit

* **New Features**
  * Added two composer snackbar messages for audio recording: a recording hint and a microphone permission rationale.
  * These messages are now displayed above the message composer, alongside existing composer alerts.

* **Bug Fixes**
  * Improved snackbar placement so audio recording prompts appear in the composer area consistently.
  * Updated recording permission handling to keep the rationale prompt working as expected.

* **Tests**
  * Added UI coverage for both new audio recording snackbar messages.

### DIFF
--- a/stream-chat-android-compose/api/stream-chat-android-compose.api
+++ b/stream-chat-android-compose/api/stream-chat-android-compose.api
@@ -2035,15 +2035,17 @@ public final class io/getstream/chat/android/compose/ui/messages/composer/Compos
 	public final fun getLambda$-1267828661$stream_chat_android_compose_release ()Lkotlin/jvm/functions/Function2;
 	public final fun getLambda$-1344661276$stream_chat_android_compose_release ()Lkotlin/jvm/functions/Function2;
 	public final fun getLambda$-1483541199$stream_chat_android_compose_release ()Lkotlin/jvm/functions/Function2;
+	public final fun getLambda$-17156589$stream_chat_android_compose_release ()Lkotlin/jvm/functions/Function3;
 	public final fun getLambda$-173232017$stream_chat_android_compose_release ()Lkotlin/jvm/functions/Function2;
+	public final fun getLambda$-1908203637$stream_chat_android_compose_release ()Lkotlin/jvm/functions/Function3;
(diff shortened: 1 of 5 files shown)


andremion:
@coderabbit review

gpunto (approved):

PR #6550, opened 2026-07-06
Migrate read-state request bodies to generated models

gpunto:
### Goal

Migrate the read-state request bodies (`markRead`, `markUnread`) to generated models. Part of the incremental OpenAPI model migration.

### Implementation

- Add generated `internal MarkReadRequest` and `MarkUnreadRequest` in `network.models`; delete the hand-written ones. Point `ChannelApi`/`MoshiChatApi`/tests at them.
- Field names move to camelCase with `@Json` names preserved (`message_id` -> `messageId`, `message_timestamp` -> `messageTimestamp`, `thread_id` -> `threadId`), so the wire format is unchanged; only positional/named args at the call sites change.
- `markAllRead` sends no body, so there is no `MarkChannelsReadRequest` to migrate here.

### Testing

- Project-wide `spotlessApply`, `apiDump` (no API drift), `detekt`, and `testDebugUnitTest` all pass.
- Verified on device: `markMessageRead` -> `200`, `markUnread` -> `200`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of read and unread status updates, including threaded conversations.
  * Updated request data handling for better compatibility and consistency across the app.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1161,7 +1161,7 @@ constructor(
         return channelApi.markRead(
             channelType = channelType,
             channelId = channelId,
-            request = MarkReadRequest(thread_id = threadId),
+            request = MarkReadRequest(threadId = threadId),
```
andremion, line 1164: The old `message_id` was `String = ""` (non-null), the new `messageId` is `String? = null`. Since the Retrofit converter uses `MoshiConverterFactory.create(moshi)` without null serialization, Moshi drops null fields. So this call now sends `{"thread_id":"..."}` where before it sent `{"message_id":"","thread_id":"..."}`.

It is probably fine (an empty `message_id` on a thread read was likely meaningless), but it is a real body change, not just a rename, and it is the one path not in the "Verified on device" list. Can you confirm `markThreadRead` still returns 200 with the new body? If you want to keep the exact old request body, defaulting `messageId` to `""` would do it.

andremion (approved):
One thing to verify inline before merge.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Yes! Tested with a thread:


```
I  (OkHttp https://chat.stream-io-api.com/...:102) --> POST https://chat.stream-io-api.com/channels/messaging/!members-YIUUn_4yrXqDaNlIeebmSQg2243g5FZg6ZAZ8mVaLEU/read (58-byte body)
I  (OkHttp https://chat.stream-io-api.com/...:102) {"thread_id":"filip-3cc72b9d-3a21-46ea-bc4b-9834b867694b"}
I  (OkHttp https://chat.stream-io-api.com/...:102) --> END POST (58-byte body)
I  (OkHttp https://chat.stream-io-api.com/...:102) curl -X POST -H "Content-Type:application/json; charset=UTF-8" -d '{"thread_id":"filip-3cc72b9d-3a21-46ea-bc4b-9834b867694b"}' "https://chat.stream-io-api.com/channels/messaging/!members-YIUUn_4yrXqDaNlIeebmSQg2243g5FZg6ZAZ8mVaLEU/read"
I  (OkHttp https://chat.stream-io-api.com/...:102) <-- 201 https://chat.stream-io-api.com/channels/messaging/!members-YIUUn_4yrXqDaNlIeebmSQg2243g5FZg6ZAZ8mVaLEU/read?api_key=qx5us2v6xvmh (157ms, unknown-length body)
I  (OkHttp https://chat.stream-io-api.com/...:102) <-- END HTTP (4086-byte body omitted)
D  markThreadRead messaging:!members-YIUUn_4yrXqDaNlIeebmSQg2243g5FZg6ZAZ8mVaLEU/filip-3cc72b9d-3a21-46ea-bc4b-9834b867694b: OK
```

andremion (approved):

PR #6551, opened 2026-07-07
compose: Derive Giphy ratio from the downloaded GIF when dimensions are missing

gpunto:
### Goal

When a Giphy attachment has no dimensions metadata (e.g. GIFs synced from another platform), `GiphyAttachmentContent` rendered a hardcoded container size with `ContentScale.Crop`, cropping portrait and landscape GIFs. Derive the container aspect ratio from the downloaded GIF instead.

Closes [AND-1296](https://linear.app/stream/issue/AND-1296/derive-giphy-aspect-ratio-from-the-downloaded-gif-when-dimensions-are)

### Implementation

- When a Giphy attachment has no dimensions (`giphyInfo == null`), measure the downloaded GIF's intrinsic size via Coil and size the container to that aspect ratio, clamped to the configured max dimensions.
- While the GIF loads, use a 1:1 container so both portrait and landscape are covered.
- Only this no-dimensions path changes; Giphy attachments that carry dimensions render as before.

### UI Changes

| Before | After |
| --- | --- |
| Screenshot_20260707_105827 | Screenshot_20260707_111156 |

### Testing

- Updated the `giphy attachment content` Paparazzi snapshot and `verifyPaparazziDebug` passes.
- Spotless and Detekt pass on the module.
- Manually verified in the compose sample: Giphy attachments without dimensions render at their real aspect ratio instead of a fixed-size box.

## Summary by CodeRabbit

* **New Features**
  * Improved Giphy attachment previews by sizing fallback images more accurately when extra metadata isn’t available.
  * Added a smoother loading experience for fallback previews, including a placeholder until the image is ready.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
```diff
@@ -233,12 +251,38 @@ public fun GiphyAttachmentContent(
...
+                    ShimmerProgressIndicator(modifier = Modifier.matchParentSize())
+                } else {
+                    Image(
+                        painter = painter,
+                        contentDescription = null,
+                        modifier = Modifier.matchParentSize(),
+                        contentScale = contentScale,
+                    )
+                }
+            }
```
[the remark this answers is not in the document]
gpunto: The behavior is aligned with what we do today, out of scope for this fix

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
```diff
@@ -178,6 +183,8 @@ public fun GiphyAttachmentContent(
     val width = ChatTheme.dimens.attachmentsContentGiphyWidth
     val height = ChatTheme.dimens.attachmentsContentGiphyHeight
 
+    var downloadedRatio by remember(previewUrl) { mutableStateOf<Float?>(null) }
```
andremion: The ratio resets on `previewUrl`, but the image below loads from `attachment.giphyFallbackPreviewUrl`. These resolve in a different order: `previewUrl` is `titleLink ?: thumbUrl ?: ogUrl`, and `giphyFallbackPreviewUrl` is `thumbUrl ?: titleLink ?: ogUrl`. So if both `titleLink` and `thumbUrl` are set and differ, the reset key and the loaded image are different URLs. The old ratio can then stay while the image changes. Should the key be `attachment.giphyFallbackPreviewUrl`, so it tracks the image that produced the ratio?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
```diff
@@ -233,12 +251,38 @@ public fun GiphyAttachmentContent(
                 onLongClick = { onLongItemClick(message) },
             ),
     ) {
-        StreamAsyncImage(
-            data = giphyInfo?.url ?: attachment.giphyFallbackPreviewUrl,
-            modifier = Modifier.fillMaxSize(),
-            contentDescription = null,
-            contentScale = contentScale,
-        )
+        if (giphyInfo == null) {
```
andremion: This branch repeats what the `StreamAsyncImage(data, contentDescription, ...)` overload already does: shimmer when the painter is null, otherwise `Image` with `matchParentSize`. The only extra is the `LaunchedEffect` reading `intrinsicSize`. If that shared overload changes its loading or success rendering later, this copy will not reflect the change. Is there a way to read the intrinsic size without re-implementing the whole loading path?

andremion (approved):
LGTM. Left a couple of questions

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
gpunto: Good point, added a callback to avoid duplicating the implementation

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
gpunto: You're right, fixed it!

PR #6552, opened 2026-07-07
compose: Render no-caption voice message replies on the bubble background

andremion:
### Goal

When a voice message reply has no caption, the player rendered inside the darker attachment card and carried the attachment section padding plus the bubble's bottom spacer, instead of sitting on the message bubble background like a standalone voice message. The finalized Figma Reply States show a no-caption voice reply on the bubble background with no card, so this aligns the Compose SDK with the design.

The quoted (replied-to) preview keeps its darker attachment background, which is correct per the same spec, only the reply's own player was wrong.

Resolves AND-1288

### Implementation

- `AudioRecordAttachmentContent`: gate the player's darker attachment card and its section padding on caption presence (`hasCaption`). A no-caption reply renders on the bubble background with only the player's own content padding; a captioned reply keeps the card, and the caption provides the bottom padding. Extracted a stateless `AudioRecordAttachmentContentItems` so the reply states can be previewed and snapshot-tested.
- `MessageContent`: skip the bubble's bottom spacer when the trailing content is a no-caption voice recording, which already supplies its own bottom padding. Media and file sections have no bottom padding of their own, so they still rely on the shared spacer.

### 🎨 UI Changes

Voice message reply without a caption, in the message list. Before, the player sits in the darker attachment card with extra inset and a doubled bottom padding; after, it sits on the bubble background like a standalone voice message. A captioned voice reply is unchanged (it keeps the card).

| Before | After |
| --- | --- |
| before_voice_reply | after_voice_reply |

### Testing

1. Open a channel and reply to a voice message by recording and sending a voice message with no caption.
2. The reply's player should render on the message bubble background (no darker card), aligned like a standalone voice message, with no extra bottom padding.
3. Reply again with a voice message that includes a caption. The player should keep the darker card, and the caption should sit below it.
4. Confirm a standalone voice message (not a reply) is unchanged.

Unit tests: added `audio record reply without caption` and `audio record reply with caption` Paparazzi snapshots to `AttachmentsContentTest`, covering the no-card and card states.

## Summary by CodeRabbit

* **New Features**
  * Added reply-style audio recording attachment previews for both caption and no-caption message layouts.
  * Improved audio attachment rendering so playback state and controls are passed through consistently.

* **Bug Fixes**
  * Adjusted spacing around messages with recordings to better match text and attachment combinations.
  * Updated audio attachment layout padding to respond to caption presence more accurately.

* **Tests**
  * Added snapshot coverage for the new audio reply attachment states in both light and dark mode.


andremion:
@coderabbit review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/components/messages/MessageContent.kt
```diff
@@ -303,7 +303,8 @@ internal fun DefaultMessageRegularContent(
             )
         }
 
-        if (!info.displaysFullSizeAttachment) {
+        val recordingSuppliesBottomPadding = info.hasRecordings && message.text.isEmpty()
+        if (!info.displaysFullSizeAttachment && !recordingSuppliesBottomPadding) {
```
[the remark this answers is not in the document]
andremion: e26fc58c6b645e8ce75f3f6241e4092d8e81c6c3

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/AudioRecordAttachmentContent.kt
```diff
@@ -94,16 +94,44 @@ public fun AudioRecordAttachmentContent(
...
     val audioRecordings = attachmentState.message.attachments
         .filter { attachment ->
             val attachmentUrl = attachment.assetUrl ?: attachment.upload?.toUri()?.toString()
             attachment.isAudioRecording() && attachmentUrl != null
         }
 
-    val playerState by viewModel.state.collectAsStateWithLifecycle()
-
-    val shouldBeFullSize = attachmentState.message.shouldBeDisplayedAsFullSizeAttachment()
+    val hasCaption = attachmentState.message.text.isNotEmpty()
```
gpunto, line 134: Does this mean that now voice recording attachments diverge from other attachment types? e.g. a file attachment will still be rendered with the background in the quoted message case, right?

gpunto (commented):
Just a doubt if this change should be propagated to other attachment types

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/AudioRecordAttachmentContent.kt
andremion: Yes, on purpose, only the voice player drops the background when there's no caption; file and media replies keep theirs.
Here are the Figma specs https://www.figma.com/design/BJppI3JYlPy3XeEu5MZY1E/SDK-Components-Library?node-id=4011-115610&m=dev

gpunto (approved):

PR #6553, opened 2026-07-07
compose: Derive Giphy ratio from the downloaded GIF when dimensions are missing

gpunto:
### Goal

When a Giphy attachment has no dimensions (e.g. GIFs synced from another platform), `GiphyAttachmentContent` rendered a hardcoded container size with `ContentScale.Crop`, cropping portrait and landscape GIFs. Derive the container aspect ratio from the downloaded GIF instead.

Ports #6551 from `v6`.

Closes [AND-1296](https://linear.app/stream/issue/AND-1296/derive-giphy-aspect-ratio-from-the-downloaded-gif-when-dimensions-are)

### Implementation

- When a Giphy attachment has no dimensions (`giphyInfo == null`), measure the downloaded GIF's intrinsic size via Coil and size the container to that aspect ratio, clamped to the configured max dimensions.
- While the GIF loads, use a 1:1 container so both portrait and landscape are covered.
- Only this no-dimensions path changes; Giphy attachments that carry dimensions render as before.
- Prevent the "GIPHY" label from wrapping onto multiple lines on narrow (portrait) attachments.

### Testing

- Updated the affected `AttachmentsContentTest` Paparazzi snapshots and `verifyPaparazziDebug` passes.
- Spotless and Detekt pass on the module.

## Summary by CodeRabbit

* **New Features**
  * Giphy attachments now size more accurately when image details aren’t immediately available, improving how they fit within the chat view.
  * Giphy labels are now kept on a single line for cleaner display.

* **Bug Fixes**
  * Fixed cases where Giphy previews could appear with awkward fallback dimensions before the image finished loading.
  * Improved image loading behavior so sizing can update once the downloaded image dimensions are known.


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
```diff
@@ -227,8 +244,16 @@ private fun calculateSize(
                     )
                 }
             }
+        } else if (downloadedRatio == null) {
+            val side = minOf(maxWidth, maxHeight)
+            DpSize(side, side)
         } else {
-            DpSize(maxWidth, maxHeight)
+            calculateResultingDimensions(
```
andremion, line 252: The preview here is a 200x150 (landscape) image, but the updated snapshots come out as a 1:1 square instead of the landscape ratio we'd derive from it. So `downloadedRatio` is still null when the snapshot is captured, and the resize this adds isn't really covered. Might be worth a small unit test on the sizing instead (e.g. `calculateResultingDimensions` for a known ratio)?

andremion (commented):
One question on the snapshots inline.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/GiphyAttachmentContent.kt
gpunto: Yeah, the frame is captured before settling. Added unit tests in https://github.com/GetStream/stream-chat-android/pull/6553/commits/16946aedfba8c3ba9f937b3dfb84d61f5900d481

andremion: Tests look good, thanks. Small thing: calculateResultingDimensions is now internal only for the test, so want to tag it @VisibleForTesting to make that explicit?
I'm approving anyway

andremion (approved):

PR #6557, opened 2026-07-08
Start the composer in thread mode when opened for a thread

andremion:
### Goal

Opening the thread view from the thread list made the message composer content jump during the opening transition. The composer was built in the channel's Normal mode, so it briefly rendered the channel draft and then switched to the thread once the message list loaded, changing its content while the transition animation ran.

This change makes the composer show its final thread state from the first frame, so it stays in place during the whole transition.

Resolves AND-1289

### Implementation

`MessageComposerController` takes a new `parentMessageId` parameter. When it is set (the screen is opened directly for a thread), the composer starts in thread mode instead of Normal, and its initial draft fetch targets the thread rather than the channel. The Compose `ChannelViewModelFactory` passes `parentMessageId` through to the controller.

The controller is `@InternalStreamChatApi` and the new parameter is additive with a default value, so there is no public API change.

### 🎨 UI Changes

Behavior change during the thread-list to thread-view transition: the composer now shows its thread state (the "Also send to channel" checkbox and any thread draft) from the first frame, instead of briefly showing the channel content and jumping to the thread state when the transition finishes.

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/0e904e97-fec6-4956-a862-58a014e6df51" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/5918b4af-6ae4-4153-a43a-53dfed71b766" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### Testing

1. Open the Threads tab.
2. Tap a thread to open it.
3. Watch the message composer while the opening transition runs.

Expected: the composer is in its final thread state for the whole transition, with no mid-animation jump from channel content to thread content.

## Summary by CodeRabbit

* **New Features**
  * Opening a message thread now starts directly in thread mode when a parent message is provided.
  * The message composer now initializes in the correct mode and loads the appropriate thread draft immediately.

* **Bug Fixes**
  * Improved composer state handling so thread replies don’t incorrectly use channel draft state at startup.

* **Documentation**
  * Updated guidance clarifying thread-mode vs channel-mode behavior when a parent message is set.

* **Tests**
  * Added coverage for thread-mode initialization with an existing parent message draft.


andremion:
@coderabbit review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -266,7 +269,15 @@ public class MessageComposerController(
 
     private val sessionRepository = ComposerSessionRepository(savedStateHandle)
 
-    private val _state = MutableStateFlow(MessageComposerState())
+    // Render the final thread state from the first frame, so the composer does not change during the
+    // opening transition and its initial draft fetch targets the thread, not the channel.
+    private val _state = MutableStateFlow(
+        MessageComposerState(
+            messageMode = initialParentMessageId
+                ?.let { MessageMode.MessageThread(Message(id = it)) }
```
gpunto: In `setMessageMode` we guard updates on `Message.id`. This means that when it is called for a thread, the id will match with the one in this stub and so we won't update the message mode in the state to contain the full `Message`. Is that a intended?

andremion: Good catch, it was not intended. The stub is only meant as a starting point so the composer opens straight into thread mode. It should be replaced by the full `Message` once the list provides it.

I moved the `_state.messageMode` update to the top of `setMessageMode` so the mode is always refreshed, and kept the `isSameMessageMode` guard only for the draft save/fetch, which we don't need to redo when the thread is unchanged. So the state now holds the full parent while the draft is still not reloaded on same-mode calls.

Fixed in cb61f76e005e0a95f9e7bb59d6d6e9f506586e2a

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/composer/MessageComposerController.kt
```diff
@@ -266,7 +269,15 @@ public class MessageComposerController(
...
-    private val _state = MutableStateFlow(MessageComposerState())
+    // Render the final thread state from the first frame, so the composer does not change during the
+    // opening transition and its initial draft fetch targets the thread, not the channel.
+    private val _state = MutableStateFlow(
+        MessageComposerState(
+            messageMode = parentMessageId
+                ?.let { MessageMode.MessageThread(Message(id = it)) }
+                ?: MessageMode.Normal,
+        ),
+    )
```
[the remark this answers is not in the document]
andremion: cb61f76e005e0a95f9e7bb59d6d6e9f506586e2a

gpunto (approved):

PR #6559, opened 2026-07-09
compose: Show poll option text in full and align option layout with the design

andremion:
### Goal

In the Compose SDK, long poll option text in the message bubble was truncated to 2 lines with an ellipsis. On iOS the full text is shown, so this was a parity gap. This change shows the full option text and, while on the same option row, aligns the layout with the design spec.

Resolves AND-1274

### Implementation

- Remove the `maxLines = 2` / `TextOverflow.Ellipsis` on the option text so it renders in full.
- Top-align the option row content (radio, voter avatars, vote count) so they sit on the first line of a wrapped option, matching the design's `items-start`. A `LineHeightStyle` with `Trim.None` keeps the gap from the text to the progress bar consistent for one-line and multi-line options.
- Give each option a rounded 12dp tap frame with a Material ripple (`clip` + `combinedClickable(indication = ripple())`), replacing the flat state overlay that came from the fallback indication. An 8dp padding makes the ripple wrap the whole option (radio, text, votes, progress bar).
- Match the design spacing: options are grouped with an 8dp horizontal inset and a 2dp gap between them, in both the message bubble (`PollMessageContent`) and the more-options sheet (`PollMoreOptionsDialog`).
- Reserve the voter-avatar height (`avatarSize + AvatarStackBorderSize`) for the votes area on non-anonymous polls, so an option with a voter avatar keeps the same height as one without.
- Keep a consistent 8dp gap between the option text and the vote count even when an option has no avatars.

### 🎨 UI Changes

Visual changes on the poll message and the more-options sheet:
- Long option text now wraps and shows in full instead of being cut to 2 lines.
- Tapping an option shows a rounded ripple instead of a flat grey background.
- Tighter, spec-aligned spacing between options.
- Options keep the same height whether or not they have a voter avatar.

Before and after are captured in the updated Paparazzi snapshots in this PR (`PollMessageContentTest`, `PollMoreOptionsDialogTest`, `MessageListTest`).

### Testing

1. Open a channel and send a poll with at least one option whose text is long enough to wrap past 2 lines.
2. Confirm the long option shows in full (no ellipsis), matching iOS.
3. Tap an option and confirm the press shows a rounded ripple, not a flat grey fill.
4. Vote on one option and confirm its row does not become taller than the others (the voter avatar appears without changing the option height).
5. Open the more-options sheet ("See all options") and confirm the option spacing matches the message bubble.

## Summary by CodeRabbit

* **New Features**
  * Improved poll option and avatar stack layouts for a cleaner, more consistent chat experience.
  * Poll options now maintain steadier spacing and alignment, including better handling of longer option text.

* **Bug Fixes**
  * Fixed inconsistent poll row heights so options display more uniformly.
  * Improved poll interaction behavior and vote count presentation for clearer, more reliable UI feedback.


andremion:
@coderabbit review

gpunto (approved):

PR #6560, opened 2026-07-09
client: Guard SimpleCache construction and document cache key behaviour

VelikovPetar:
## Summary

Backport of the same fixes applied to #6542 (port/v6-to-develop/opt-in-video-disk-cache).

- **`VideoMediaCache.create()` now returns `VideoMediaCache?`** — wraps `SimpleCache` construction in a `try-catch`. A stale directory lock left by a prior crash no longer aborts `ChatClient` initialisation; video caching is silently disabled instead. `ChatClient.kt` already uses `?.let {}` so the nullable return is handled transparently.
- **`VideoCacheConfig` KDoc** — added one sentence noting that cache entries are keyed by URL path only (query parameters are stripped).
- **LRU eviction test** — comment extended to acknowledge the wall-clock dependency of `Thread.sleep` spacing and what to do if the test becomes flaky.

## Test plan

- [ ] `./gradlew :stream-chat-android-client:testDebugUnitTest --tests "*.cache.*"` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved video cache startup reliability by handling cache creation failures gracefully instead of crashing.
  * URLs that differ only by query parameters now share the same video cache entry, reducing duplicate cache usage.
  * Added clearer guidance around cache behavior and updated tests to better catch cache initialization issues.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/StreamCacheConfig.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/cache/StreamCacheConfig.kt
@@ -36,6 +36,9 @@ public data class StreamCacheConfig(
  * enabled, replaying or seeking within a previously watched video reuses cached byte ranges
  * instead of re-downloading from the CDN.
  *
+ * Cache entries are keyed by URL path only; query parameters are stripped. URLs that differ
+ * only in their query string share the same cache entry.
+ *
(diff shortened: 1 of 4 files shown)


VelikovPetar:
Closing — will be reopened via the correct workflow.

PR #6566, opened 2026-07-10
Migrate sendReaction request body to generated ReactionRequest and SendReactionRequest models

gpunto:
### Goal

Migrate the `sendReaction` request body (`POST /messages/{id}/reaction`) to the generated `SendReactionRequest` + `ReactionRequest` network models. First patch-free generated model that carries a flattened `custom` field. Part of the incremental OpenAPI model migration.

### Implementation

- Add generated `internal SendReactionRequest` (the `{reaction, enforce_unique, skip_push}` wrapper) and `ReactionRequest` (the reaction object) in `network.models`. Remove the hand-written `api2.model.requests.ReactionRequest` wrapper and the hand-written `UpstreamReactionDto`. Real names, no alias.
- Enhance `CustomObjectDtoAdapter` (backward-compatibly) with an `extraDataPropertyName` parameter (defaults to `extraData`) and `@Json`-aware member-name resolution, so generated DTOs that hold the overflow map as `custom` flatten correctly. Existing hand-written adapters are unchanged. `UpstreamReactionDtoAdapter` now targets the generated `ReactionRequest` with `extraDataPropertyName = "custom"`.
- `Reaction.toDto()` now emits the lean backend shape: it drops `message_id` (comes from the URL path), `user`/`user_id` (set server-side), and folds `emoji_code` into `custom`. Verified against the `/chat` `SendReactionRequest`/`ReactionRequest` Go structs.

### Testing

- `spotlessApply`, `apiCheck` (no API drift), `detekt`, `lint`, and the full client `testDebugUnitTest` suite pass.
- Verified on device: `POST /messages/{id}/reaction` sends `{"reaction":{"type":"like","score":1,"migrate_probe":"v1","emoji_code":"👍"},"enforce_unique":false,"skip_push":false}` (lean object, no `message_id`/`user`/`user_id`; `custom` flattened to root), returns `201`, and the custom field + `emoji_code` round-trip on the returned reaction.

### Note on `CustomObjectDtoAdapter`

The `memberNames` change (from `kClass.members` to the primary-constructor parameters, resolving `@Json` names) affects every adapter extending `CustomObjectDtoAdapter`, not just reactions. Previously `kClass.members` also included data-class functions (`copy`, `toString`, `componentN`), so a custom field whose key collided with one of those names was dropped from `extraData`; it is now correctly retained. Pinned by a test in `DownstreamReactionDtoAdapterTest`.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/CustomObjectDtoAdapter.kt
```diff
@@ -16,28 +16,35 @@
...
-     * handled. These will not be copied into extraData when parsing with
-     * [parseWithExtraData].
+     * Wire-format names of the declared properties on [Value]. These will not be copied into the
+     * overflow map when parsing with [parseWithExtraData]. Reads `@Json(name = ...)` first so
+     * camelCase properties map to their snake_case wire names, falling back to the Kotlin
+     * parameter name when `@Json` is absent.
      */
     private val memberNames: List<String> by lazy {
-        kClass.members.map { member -> member.name }.minus(EXTRA_DATA)
+        kClass.primaryConstructor?.parameters.orEmpty().mapNotNull { param ->
```
andremion, line 45: Heads up that this changes behavior for every adapter extending `CustomObjectDtoAdapter`, not just the reaction one. `kClass.members` also included the data class functions (`copy`, `toString`, `componentN`, ...), so a custom field named `copy` was dropped from `extraData` before, and now it's kept. I think the new behavior is better, but maybe worth a test to pin it, and a mention in the PR description?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DtoMappingTest.kt
```diff
@@ -243,16 +243,14 @@ internal class DtoMappingTest {
...
             type = reaction.type,
-            updated_at = reaction.updatedAt,
-            user = reaction.user?.let { with(mapping) { it.toDto() } },
-            user_id = reaction.userId,
-            extraData = reaction.extraData,
-            emoji_code = reaction.emojiCode,
+            createdAt = reaction.createdAt,
+            score = reaction.score,
+            updatedAt = reaction.updatedAt,
+            custom = reaction.emojiCode
```
andremion: The expected `custom` here is built with the same expression as the implementation, so it can't catch a bug in the folding logic. Also `randomReaction()` defaults `emojiCode` to a non-null value and `extraData` to an empty map, so the `emojiCode == null` branch never runs and the merge is always over an empty map. Maybe split into two fixed cases, one with `emojiCode` and extra data set, and one with `emojiCode = null`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ReactionDtoAdapters.kt
```diff
@@ -40,17 +40,17 @@ internal object DownstreamReactionDtoAdapter :
 }
 
 internal object UpstreamReactionDtoAdapter :
-    CustomObjectDtoAdapter<UpstreamReactionDto>(UpstreamReactionDto::class) {
+    CustomObjectDtoAdapter<ReactionRequest>(ReactionRequest::class, extraDataPropertyName = "custom") {
```
andremion, line 44: Nit: this `"custom"` needs to stay in sync with the `@Json(name = "custom")` on `ReactionRequest`. The old DTO had a `@StreamHandsOff` note for exactly this kind of coupling. Maybe a short comment here too?

andremion (approved):
LGTM. Left a few small comments.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/CustomObjectDtoAdapter.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/808cffee42

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DtoMappingTest.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/41589edcae

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ReactionDtoAdapters.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/a7fa3dee1e

andremion (approved):

PR #6570, opened 2026-07-13
e2e: Fix the flaky test races and re-enable the composer giphy test

andremion:
### Goal

Remove the two client-side race mechanisms behind the flaky e2e tests, harden the RetryRule screen recording, and re-enable the first of the ignored tests. Part of AND-1303; the remaining flake family (mock server behavior under WebSocket reconnects) is addressed in GetStream/stream-chat-test-mock-server#43.

### Root causes

Diagnosed from the per-attempt artifacts (logcat, screenshot, hierarchy, step timings) that AND-1302's retry reporting attached to every flake of the 2026-07-13 nightly, plus the step durations of historic failures in TestOps:

| Mechanism | Signature | Affected tests |
|---|---|---|
| Wait-once-read-once race in `waitToAppear` | "timed out after 10000ms" that actually failed in 272 to 740ms, or `StaleObjectException`; screenshots show the correct app state | `threadIsNotLocked...`, `userAddsReactionWhileOffline` |
| Same race via `wait().isDisplayed()` | fast `AssertionError` (460 to 572ms) | both `deliveryStatusShown...` tests |
| Mock server under reconnects (not client bugs) | genuine 10s timeouts | `threadReplyIsRemovedEverywhere...`, `originalQuoteIsDeleted...`, `channelPreviewShowsPreviousMessage...`, `deletesReaction` — tracked in the mock repo |

### Implementation

- `waitToAppear` (both overloads) polls `findObject` at the existing 50ms interval within the same timeout budget and absorbs `StaleObjectException` between polls.
- Visibility asserts go through a new selector-level `waitDisplayed` predicate: the find-and-check happens inside the poll, so asserts never hold a `UiObject2` across reads (a held node can be recycled by recomposition before `visibleCenter` is read, which CI demonstrated after the first two fixes). `assertMessageDeliveryStatus` and `assertMessageFailedIcon` route through the same helper (the READ branch keeps its 30s timeout).
- `waitToAppearBottomUp` gets the same treatment: the find and the bounds-reading sort happen inside the poll, because the sort reads `visibleBounds` on every element and any of them can be recycled mid-sort. This surfaced while running the currently ignored quoted-reply tests, the only callers that reach the crashing branch. The wait helper KDocs now state only the contract; the race rationale lives in this description.
- `RetryRule`: the recording file is named by `methodName` (the display name's parentheses can break when the recording commands go through the device shell), and `stopRecordingSafely` makes sure a recording infrastructure failure cannot change a test result or skip the failure reporting.
- `test_userObservesAnimatedGiphy_afterAddingGiphyThroughComposerMenu` re-enabled: its blocking ticket (AND-1018, composer redesign) shipped in January. The flow is adapted to the redesigned composer (typing `/` opens the command suggestion list; the dedicated commands button no longer exists).
- The shell-based `UiDevice.typeText` is removed: on this uiautomator version the shell command's wrapping quotes reach the input field as literal text, and nothing uses it after the flow fix.

### Testing

All on a local emulator against the mock server:

1. `MessageListTests#test_threadIsNotLocked_afterParentMessageDeletedByParticipant` (the race-prone test): 3/3 green, first attempt each time.
2. `MessageDeliveryStatusTests#test_singleCheckmarkShown_whenMessageIsSent` and `#test_deliveryStatusShownForTheLastMessageInGroup`: green through the new assert routing.
3. `GiphyTests#test_userObservesAnimatedGiphy_afterAddingGiphyThroughComposerMenu` (re-enabled): 2/2 green on the adapted flow.

4. The two ignored quoted-reply tests that crash inside `waitToAppearBottomUp` (`test_quotedReplyNotInList_whenUserAddsQuotedReply` and its thread variant), run with their `@Ignore` stripped: the `StaleObjectException` is gone and both reach their real assertions (they still fail on product behavior, which is expected and tracked in AND-1304).

To reproduce any of them: `./gradlew :stream-chat-android-compose-sample:connectedE2eDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.class=<class>#<method>` with the mock server driver running.

No UI changes.

## Summary by CodeRabbit

- **New Features**
  - Composer command suggestions can now be opened by typing `/` in the message field.
  - Giphy commands can be selected and inserted directly from the suggestion list.

- **Bug Fixes**
  - Improved reliability when waiting for messages, delivery statuses, and error indicators.
  - Reduced UI automation failures caused by stale or changing screen elements.
  - Improved handling of test recordings so recording issues do not obscure test results.

- **Tests**
  - Re-enabled coverage for animated Giphy messages added through composer commands.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
@@ -65,7 +65,6 @@ open class MessageListPage {
             val cooldownIndicator get() = By.res("Stream_ComposerCooldownIndicator")
             val saveButton get() = By.res("Stream_ComposerSaveButton")
             val recordAudioButton get() = By.res("Stream_ComposerAudioRecordingButton")
-            val commandsButton get() = By.res("Stream_ComposerCommandsButton")
             val commandSuggestionList get() = By.res("Stream_CommandSuggestionList")
             val commandSuggestionListTitle get() = By.res("Stream_CommandSuggestionListTitle")
             val userSuggestion get() = By.res("Stream_SuggestionItem")
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
@@ -31,7 +31,6 @@ import io.getstream.chat.android.e2e.test.mockserver.ReactionType
 import io.getstream.chat.android.e2e.test.robots.ParticipantRobot
 import io.getstream.chat.android.e2e.test.uiautomator.defaultTimeout
 import io.getstream.chat.android.e2e.test.uiautomator.device
-import io.getstream.chat.android.e2e.test.uiautomator.findObject
 import io.getstream.chat.android.e2e.test.uiautomator.findObjects
 import io.getstream.chat.android.e2e.test.uiautomator.isDisplayed
 import io.getstream.chat.android.e2e.test.uiautomator.longPress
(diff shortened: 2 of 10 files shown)


andremion:
@coderabbit review

VelikovPetar (approved):

PR #6572, opened 2026-07-14
e2e: Port the iOS-only tests for slow mode, truncation, scroll, read events, and attachments

andremion:
### Goal

Close the iOS e2e parity gap for every topic that needs no new delivery infrastructure: slow mode, channel truncation, message list scroll, read events, and attachment upload recovery. Each ported test binds to its pre-seeded Allure TestOps case id via `@AllureId`, so the case flips from manual to automated and keeps its history. Part of AND-1304; push notifications follow separately because they need a new delivery path in the mock server.

### Implementation

One commit per topic, in the order they were verified:

- **Slow mode** (4 tests, ids 5788, 5790, 5791, 5794). The "message can't be sent" scenarios assert Android's own contract (cooldown indicator shown, composer disabled) because Android disables the composer during cooldown since #6499, while iOS keeps typing enabled. `quoteMessage` is split into `selectReplyFromContextMenu` plus send, since two tests need the reply action alone. The cooldown duration is 60 seconds (iOS uses 15) so it cannot expire between the message send and the assert on a slow CI emulator; no test waits for the cooldown to end.
- **Channel truncation** (2 tests, ids 5797, 5827). iOS truncates through its test app's debug menu; the Android sample has none, so `BackendRobot.truncateChannel` triggers it server-side through a new mock-server route and the app reacts to the `channel.truncated` event. New `assertMessageCount` helper.
- **Message list scroll and commands** (3 tests, ids 5795, 5831, 5720). Adds `ParticipantRobot.sendMultipleMessages` (mirrors iOS: `text-index` suffix, 300ms spacing), a testTag on the scroll-to-bottom unread badge plus its assert helper, and the zero-count backfill on the truncation tests. The command test asserts Android's behavior: the attachments button hides when a command is selected from the suggestions, while on iOS typing `/` already hides the left buttons.
- **Read events and member removal** (11 tests, ids 5746, 5756, 5759-5766, 5771). Adds `BackendRobot.setReadEvents`/`addMember`/`removeMember` (server-side member triggers in the mock server) and `ParticipantRobot.id`. Only 2 land active: the other 9 assert behavior the Compose SDK does not implement yet and are `@Ignore`d against AND-1309 (delivery indicators ignore `read_events`) and AND-1310 (READ does not fall back to SENT when the reading member is removed). Both tickets list their tests; the fix PRs remove the `@Ignore` and the tests are the acceptance check.
- **Attachment upload recovery** (1 test, id 5928). iOS restarts a failed image upload with a dedicated button; the Compose SDK has no restart control, and the offline module uploads the image automatically on reconnect, so the test asserts that recovery under an Android-accurate name.

The truncation and member trigger routes this PR relies on shipped in GetStream/stream-chat-test-mock-server#44 (merged).

Not ported, with reasons verified per case: 5786, 5696, 5913, 5924, 5926 are already covered by existing active Android tests; 5829, 5708, 5929 no longer exist in the iOS suite; 5914 asserts "loaded but off-screen", which Compose's LazyColumn does not expose to UiAutomator (the observable behavior is covered by active test 5912). The full disposition list is on AND-1304.

### Testing

All new active tests pass locally against the mock server, first attempt each:

```
./gradlew :stream-chat-android-compose-sample:connectedE2eDebugAndroidTest \
  -Pandroid.testInstrumentationRunnerArguments.class=io.getstream.chat.android.compose.tests.SlowModeTests
```

Same for `ChannelListTests` (truncation), `MessageListTests` (scroll and command), `MessageDeliveryStatusTests` (the 2 active read-events tests), and `AttachmentsTests#test_imageUploadRecovers_whenUserComesBackOnline`. The suite needs the attachment fixtures on the emulator (`bundle exec fastlane upload_attachments`) and recent modification times on them, so they appear in the picker's first screen.

On CI, the full 6-API-level matrix ran from this branch against the new mock routes ([run](https://github.com/GetStream/stream-chat-android/actions/runs/29345514457)): the truncation and member trigger tests passed on every API level. The only failures were the two slow mode tests on the 3 slower API levels, which is what the 60-second cooldown commit fixes.

No UI changes (the only SDK change is a testTag).

## Summary by CodeRabbit

* **Improvements**
  * Added a UI selector/tag for the “scroll to bottom” unread-count badge, enabling more precise message list validations.
  * Improved end-to-end coverage for message list UI: composer attachment button state, scroll-to-bottom behavior, and reloading skipped messages.
  * Improved reliability testing for image upload recovery after connectivity returns.
  * Improved channel preview/message list correctness after truncation (with and without system messages).
  * Strengthened message delivery-status and preview indicators for participant changes and when read events are disabled.
  * Enhanced slow-mode end-to-end validation for replies, thread replies, context-menu actions, and message editing.
  * Improved composer command flow for Giphy suggestion selection.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
@@ -95,6 +95,7 @@ open class MessageListPage {
             val unreadMessagesBadge get() = By.res("Stream_UnreadMessagesBadge")
             val typingIndicator get() = By.res("Stream_MessageListTypingIndicator")
             val scrollToBottomButton get() = By.res("Stream_ScrollToBottomButton")
+            val scrollToBottomButtonUnreadCount get() = By.res("Stream_ScrollToBottomButtonUnreadCount")
             val systemMessage get() = By.res("Stream_SystemMessage")
         }
 
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
@@ -201,9 +201,14 @@ class UserRobot {
     }
 
     fun quoteMessage(text: String, messageCellIndex: Int = 0): UserRobot {
+        selectReplyFromContextMenu(messageCellIndex)
+        sendMessage(text)
+        return this
+    }
+
+    fun selectReplyFromContextMenu(messageCellIndex: Int = 0): UserRobot {
         openContextMenu(messageCellIndex)
         ContextMenu.reply.waitToAppear().click()
-        sendMessage(text)
         return this
     }
 
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
@@ -319,13 +319,18 @@ class UserRobot {
         return this
     }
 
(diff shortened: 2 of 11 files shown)


andremion:
@coderabbit review

andremion:
@coderabbit full review

VelikovPetar (approved):

PR #6573, opened 2026-07-15
e2e: Re-enable the recovered tests and settle the ignored-test decisions

andremion:
### Goal

Settle the fate of the remaining ignored e2e tests, based on the per-test verification recorded on AND-1304. Eleven tests come back to the active suite, and the ones that stay ignored now point at real, open tickets instead of canceled ones. Part of AND-1304.

### Implementation

- **6 tests re-enabled as they were**: their canceled bug tickets described behavior that is correct on current develop (verified by running each test locally on the fixed harness). Two of them only needed the current empty-state copy (`No messages yet`, no sender prefix).
- **1 test re-enabled with a rewritten assertion**: the channel preview deliberately shows `Message deleted` when the only message is deleted (documented intent in `getLastMessageIncludingDeleted`, and iOS shows the same placeholder), so `test_channelPreviewShowsNoMessages_whenTheOnlyMessageInChannelIsDeleted` became `test_channelPreviewShowsMessageDeleted_...` and expects the placeholder.
- **4 quoted-reply tests fixed and re-enabled**: their failures were port bugs in the tests, not product bugs. The expected quoted text said `30` where the mock quotes the oldest message (`1`, matching the active sibling test and iOS), and the attachment variants asserted reply text that their robot calls never send (iOS asserts only the quote plus the attachment). The user variants now quote deterministically by message text after scrolling to the top, instead of by an index that Compose's LazyColumn cannot reach.
- **1 test deleted**: `test_commandsMenuCloses_whenUserTapsOnMessageList` asserted old-composer behavior; the redesigned suggestion list is driven by the input content and iOS has no counterpart. Its orphaned robot helper went with it.
- **3 tests parked on reopened AND-960**: tapping a quote in a thread does not jump to the quoted reply unless its cell was previously composed; reproduced deterministically with artifacts. The fix PR for AND-960 re-enables them.
- Harness: one more stale-read fix (`assertMessage`'s absence branch now checks at the selector level), `assertQuotedMessage`'s `text` parameter is optional to mirror the iOS helper shape, and a `quoteMessage(text, quotedMessageText)` overload for by-text quoting.

Ticket trail: AND-256, AND-272, and AND-960 are reopened with their tests as the acceptance check; the other linked tickets stay canceled with closing comments.

### Testing

Every re-enabled test passed locally against the mock server on the first attempt, verified per batch while making the changes (retry data on the device checked to rule out pass-on-retry). Reproduce with:

```
./gradlew :stream-chat-android-compose-sample:connectedE2eDebugAndroidTest \
  -Pandroid.testInstrumentationRunnerArguments.class=io.getstream.chat.android.compose.tests.QuotedReplyTests
```

No UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved end-to-end validation for quoted replies, deleted messages, empty channels, ephemeral messages, and message delivery states.
  * Updated channel previews to display “No messages yet” for empty channels and a deleted-message indicator when applicable.

* **Tests**
  * Re-enabled coverage for scrolling, quoted replies, delivery status, and ephemeral-message scenarios.
  * Added support for composing replies to specifically selected quoted messages.
  * Removed an obsolete message-list interaction test.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/MessageListTests.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/MessageListTests.kt
@@ -223,26 +223,6 @@ class MessageListTests : StreamTestCase() {
         }
     }
 
-    @Ignore("https://linear.app/stream/issue/AND-181")
-    @AllureId("5717")
-    @Test
-    fun test_commandsMenuCloses_whenUserTapsOnMessageList() {
-        step("GIVEN user opens the channel") {
-            userRobot.login().openChannel()
-        }
-        step("AND user opens attachments menu") {
-            userRobot
-                .openComposerCommands()
-                .assertComposerCommandsMenu(isDisplayed = true)
-        }
-        step("WHEN user taps on message list") {
-            userRobot.tapOnMessageList()
-        }
-        step("THEN command suggestions disappear") {
-            userRobot.assertComposerCommandsMenu(isDisplayed = false)
-        }
-    }
-
     @AllureId("5720")
     @Test
     fun test_addingCommandHidesAttachmentsButton() {
@@ -391,7 +371,6 @@ class MessageListTests : StreamTestCase() {
 
     // MARK: Scrolling
 
-    @Ignore("https://linear.app/stream/issue/AND-76")
     @AllureId("5792")
     @Test
     fun test_messageListScrollsDown_whenMessageListIsScrolledUp_andUserSendsNewMessage() {
@@ -906,7 +885,6 @@ class MessageListTests : StreamTestCase() {
     }
 
     @AllureId("6608")
-    @Ignore("https://linear.app/stream/issue/AND-212")
     @Test
     fun test_messageRendersTimestampAgain_whenMessageLastInGroupIsSoftDeleted() {
         step("GIVEN user opens the channel") {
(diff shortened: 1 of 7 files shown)


andremion:
@coderabbit review

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/MessageListTests.kt
```diff
@@ -223,26 +223,6 @@ class MessageListTests : StreamTestCase() {
         }
     }
 
-    @Ignore("https://linear.app/stream/issue/AND-181")
-    @AllureId("5717")
-    @Test
-    fun test_commandsMenuCloses_whenUserTapsOnMessageList() {
```
testableapple, line 229: is this no longer needed?

testableapple (approved):

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/MessageListTests.kt
andremion: Correct. This tested the old commands menu, a popup that dismissed on a tap to the message list (AND-181). The redesign replaced it with an inline suggestion list tied to the input text, which by design does not dismiss on an outside tap. So the assertion no longer holds, and I removed the test rather than leave it ignored forever.
One thing worth flagging: iOS still has the equivalent (test_commandsPopupDisappear_whenUserTapsOnMessageList, case 5717). This might be a good candidate for the clean-up too.

VelikovPetar (approved):

PR #6576, opened 2026-07-20
Run e2e on Android 16

rahul-lohra:
### Goal

Run e2e on Android 16

### Implementation

Run e2e on Android 16

### Testing

None

## Summary by CodeRabbit

* **Tests**
  * Updated Android end-to-end testing to run against Android API level 37.


.github/workflows/e2e-test.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           - batch: 2
       fail-fast: false
     env:
-      ANDROID_API_LEVEL: 34
+      ANDROID_API_LEVEL: 37
```
andremion: This is why the three `Test compose` shards fail in ~30s. There is no `platforms;android-37` package in the SDK repository. Android 17 uses the new major.minor naming, so the available packages are `platforms;android-37.0` and `platforms;android-37.1`.

There is a second problem behind it: the step does not set `target:`, so `android-emulator-runner` defaults to `default` (AOSP). No `default` system image exists for 37 at all. The only x86_64 images are `google_apis_ps16k` and `google_apis_playstore_ps16k`.

Could you change it to this?

```yaml
env:
  ANDROID_API_LEVEL: "37.0"
```

and add the target to the emulator step:

```yaml
          api-level: ${{ env.ANDROID_API_LEVEL }}
          target: google_apis_ps16k
```

That resolves to `platforms;android-37.0` and `system-images;android-37.0;google_apis_ps16k;x86_64`, both present in the stable channel. One caveat: the 37 images are 16 KB page-size builds, so we only learn from a run whether the tests pass on them.

andremion (commented):
Thanks for setting this up. The CI failures here are not test failures, the emulator never boots. Details in the inline comment.

The same issue affects the cron workflow after #6575 was merged: the `android_api_level: 37` leg in `e2e-test-cron.yml` will fail the same way every night, and it will trigger the nightly Slack alert each time. The 36 leg is fine (`platforms;android-36` and `system-images;android-36;default;x86_64` both exist). Could you include the cron fix in this PR too, or in a small follow-up? Since the matrix legs need different targets, something like:

```yaml
matrix:
  include:
    - android_api_level: "37.0"
      target: google_apis_ps16k
    - android_api_level: 36
    ...
```

and in the emulator step:

```yaml
          target: ${{ matrix.target || 'default' }}
```

andremion (approved):

PR #6577, opened 2026-07-20
Make the channel preview deterministic when the last message is deleted

andremion:
### Goal

The e2e test that covers the channel preview after deleting the last message fails intermittently (3 of the last 200 nightly runs). The investigation showed two problems. First, the preview selection in `getLastMessageIncludingDeleted` picks the first of the messages that tie on `createdAt`, so with second-precision timestamps (the mock server) the shown preview depends on whether two messages land in the same second. Second, the test still expects the previous message, an expectation that predates the channel list redesign. Since the redesign, the preview intentionally shows the "Message deleted" placeholder when the newest message is soft-deleted, and current iOS behaves the same (its e2e suite asserts this exact scenario).

Resolves AND-1314

### Implementation

- `getLastMessageIncludingDeleted` now prefers the later list element when creation times are equal, so the selection is deterministic. Against the real backend, timestamps carry sub-second precision and ties do not occur, so production behavior does not change.
- The e2e test is renamed to `test_channelPreviewShowsMessageDeleted_whenLastMessageIsDeleted` and expects the placeholder, mirroring the sibling only-message test and the iOS suite.
- `PreviewMessageData` fixture dates are now strictly increasing instead of init-time `Date()`, so Paparazzi goldens no longer depend on initialization timing. The updated goldens now show `message2` as the newest message, which matches the fixture order.
- New coverage: two `ChannelItem` snapshots (deleted last message with an older message present, and only message deleted) and three `ChannelUtilsTest` unit tests, including the tie case.

### 🎨 UI Changes

No UI changes. The selection change only affects messages with identical creation times, which does not happen against the real backend.

### Testing

1. Run the two channel preview e2e tests: open a channel, let the second user send two messages and delete the last one, go back to the channel list. The preview shows "Message deleted" with the timestamp, both when an older message exists and when the deleted message was the only one.
2. `./gradlew :stream-chat-android-compose:testDebugUnitTest --tests "*.ChannelUtilsTest"` passes.
3. `./gradlew :stream-chat-android-compose:verifyPaparazziDebug` passes.

Local verification: both e2e tests ran 10 times each against the mock server after the fix, 20 of 20 passed. Before the fix, the failure reproduced whenever the two messages crossed a second boundary (verified in 6 of 6 instrumented runs and in all 3 nightly failures).

## Summary by CodeRabbit

* **Bug Fixes**
  * Channel previews now display “Message deleted” when the latest message has been deleted.
  * Corrected latest-message selection when messages share the same timestamp.

* **Tests**
  * Added coverage for deleted-message channel states and timestamp tie-breaking.
  * Added visual previews for channels with deleted latest or only messages.


andremion:
@coderabbit review

VelikovPetar (approved):

PR #6579, opened 2026-07-20
Update the message read status when the reading member is removed from the channel

andremion:
### Goal

When the only member who read a message is removed from the channel, the message delivery status should fall back from READ (double checkmark) to SENT (single checkmark), because no remaining member has read it. Currently the status stays READ after the `member.removed` event: the member is removed from the state, but their read entry is kept, so both the Compose and XML message lists keep showing the double checkmark. The iOS SDK already behaves correctly by deleting the removed member's read state.

Resolves AND-1310

### Implementation

On `member.removed`, the removed member's `ChannelUserRead` is now deleted together with the member, in all three places that hold it:

- In-memory state: `ChannelEventHandlerImpl` calls the new `ChannelStateImpl.removeRead(userId)`.
- Legacy state: `ChannelEventHandlerLegacyImpl` calls the new `ChannelStateLogic.deleteRead(userId)`, which delegates to `ChannelStateLegacyImpl`.
- Offline storage: the `Channel.removeMember` extension (used by `EventHandlerSequential` to persist the event) now also filters the removed user's entry out of `Channel.read`, so a cold start does not restore the stale READ status.

The scope is `member.removed` only, matching the iOS SDK. `notification.removed_from_channel` and `notification.invite_rejected` are unchanged.

Since `MessageListController` computes the read status from `ChannelState.reads`, the fix applies to both the Compose and XML UI kits.

The two e2e tests ported in #6572 are no longer ignored and act as the acceptance check:
- `test_readByDecremented_whenParticipantIsRemoved`
- `test_readByDecrementedInThreadReply_whenParticipantIsRemoved`

### Testing

Manual steps:
1. Log in with user A and open a channel that also has member B.
2. Send a message as user A.
3. Read the message as user B. User A sees a double checkmark below the message.
4. Remove user B from the channel (for example, from the channel info screen or via a server-side call).
5. The indicator below the message falls back to a single checkmark.

The same steps apply to a thread reply (step 2: reply in a thread instead).

Automated coverage:
- New unit tests for `ChannelStateImpl.removeRead`, `ChannelStateLegacyImpl.deleteRead`, both event handlers, and the `Channel.removeMember` extension. `:stream-chat-android-client:testDebugUnitTest` passes.
- The two previously ignored e2e tests listed above.

## Summary by CodeRabbit

- **Bug Fixes**
  - Removed users’ read-state entries when they leave or are removed from a channel.
  - Prevented removed members from continuing to appear as channel readers.
  - Preserved read states for remaining channel members.

- **Tests**
  - Added coverage for member removal and read-state cleanup.
  - Re-enabled end-to-end tests for message and thread read-status updates after member removal.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
@@ -1166,6 +1166,15 @@ internal class ChannelStateImpl(
     }
 
     /**
+     * Removes the read state of the given user, so they no longer count as a reader of the channel.
+     *
+     * @param userId The id of the user whose read state should be removed.
+     */
+    fun removeRead(userId: UserId) {
+        _reads.update { current -> current - userId }
+    }
+
+    /**
      * Updates the delivered status for a user's read state.
      *
      * @param read The [ChannelUserRead] containing the delivered status to update.
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Channel.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Channel.kt
@@ -105,6 +105,7 @@ public fun Channel.updateLastMessage(
 
 /**
  * Removes member from the [Channel.members] and aligns [Channel.memberCount].
+ * Also removes the member's read state from [Channel.read], as a removed member no longer counts as a reader.
  *
  * @param memberUserId User id of the removed member.
  */
@@ -112,6 +113,7 @@ public fun Channel.updateLastMessage(
 public fun Channel.removeMember(memberUserId: String?): Channel = copy(
     members = members.filterNot { it.user.id == memberUserId },
     memberCount = memberCount - (1.takeIf { members.any { it.user.id == memberUserId } } ?: 0),
+    read = read.filterNot { it.user.id == memberUserId },
 )
 
 /**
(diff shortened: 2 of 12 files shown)


andremion:
@coderabbit review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/state/channel/internal/ChannelStateImpl.kt
```diff
@@ -1165,6 +1165,15 @@ internal class ChannelStateImpl(
         }
     }
 
+    /**
+     * Removes the read state of the given user, so they no longer count as a reader of the channel.
+     *
+     * @param userId The id of the user whose read state should be removed.
+     */
+    fun removeRead(userId: UserId) {
```
VelikovPetar, line 1173: Just curious: Why do we call the method `remove-` here, but `delete-` in the legacy implementation? Shouldn't we align them?

andremion: Each name follows the verb family of its own file: `ChannelStateImpl` has `removeMember` and `removeWatcher`, so I used `removeRead` there. The legacy impl has `deleteMember` and `deleteWatcher`, so I used `deleteRead`. Aligning them across the two files would break the local naming in one of them, and the legacy implementation should go away at some point anyway. But I can rename one of them if you prefer a single name.

VelikovPetar: Oh I see, I didn't see the surrounding context. All good with me thank you!

VelikovPetar (approved):

PR #6581, opened 2026-07-20
Re-enable the failed-message channel preview e2e test

andremion:
### Goal

Re-enable the e2e test `test_errorIndicatorShownInPreview_whenMessageFailedToBeSent`, ignored since January 2025, and fix the selector bug that made it fail. The SDK behavior is correct: the channel preview shows the failed delivery icon for a message with `SyncStatus.FAILED_PERMANENTLY` (confirmed by the CI failure screenshot and UI hierarchy dump, which show the icon rendered with the expected resource id). The test failed because it asserted the wrong tag.

Resolves AND-256

### Implementation

- The channel preview renders the failed state via `MessageReadStatusIcon`, tagged `Stream_MessageReadStatus_isError`. The page object reused the message list's `Stream_MessageFailedIcon` tag, which never exists in the channel list, so the assertion could only pass when a leftover message-list node was still present during the pane transition. The read, pending, and sent tags are shared between the two surfaces, which is why only the failed variant was broken.
- `ChannelListPage.Channel.deliveryStatusIsFailed` now points to `Stream_MessageReadStatus_isError`, and the test asserts it through a new `assertFailedMessageDeliveryStatusInPreview` helper.
- Removed the `@Ignore` from the test. No production code changes.

### 🎨 UI Changes

No UI changes.

### Testing

1. In the app: open a channel, force message sends to fail, send a message, and go back to the channel list. The preview shows the message with the failed delivery icon.
2. The fixed test passed 3 of 3 normal local runs, plus 1 run against a mock patched to delay the failure response by 3 seconds, so the assertion also holds when the failure lands after the navigation back.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/ChannelListPage.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/ChannelListPage.kt
@@ -17,6 +17,7 @@
 package io.getstream.chat.android.compose.pages
 
 import androidx.test.uiautomator.By
+import androidx.test.uiautomator.BySelector
 import io.getstream.chat.android.compose.pages.MessageListPage.MessageList.Message
 
 class ChannelListPage {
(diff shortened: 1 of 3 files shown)


andremion:
@coderabbit review

VelikovPetar (approved):

PR #6584, opened 2026-07-21
Update the quoted message preview in threads when the quoted message is deleted or edited

andremion:
### Goal

When a message quoted by a thread reply is deleted or edited, the quoted preview inside the thread is not updated. The channel message list already handles this case, but the thread message list keeps showing the old content. This change applies the same update to the thread state.

Resolves AND-272

### Implementation

The channel path refreshes quote previews when the `message.updated` / `message.deleted` WebSocket event arrives: `ChannelEventHandlerImpl` calls `ChannelStateImpl.updateQuotedMessageReferences`, which rewrites the `replyTo` of every quoting message. The thread path (`EventHandlerSequential.updateThreadState` -> `ThreadLogic.handleMessageEvents`) only upserted the changed message itself, so quoting thread replies kept a stale `replyTo`.

- `ThreadStateLogic`: new `updateQuotedMessageReferences(quotedMessage)` and `deleteQuotedMessageReferences(quotedMessageId)`. They scan the in-memory thread messages for `replyTo?.id` / `replyMessageId` matches instead of maintaining a `quotedMessagesMap` like the channel state does, since the thread list is small and fully in memory.
- `ThreadLogic.handleMessageEvents`: after the existing upsert, `MessageUpdatedEvent` and soft `MessageDeletedEvent` refresh the `replyTo` of quoting messages with the enriched message, and a hard delete clears the `replyTo`. This mirrors the channel event handler contract. `ThreadLogic` is shared by the new and legacy channel logic modes, so both are covered.
- Removed the `@Ignore` from `test_originalQuoteIsDeletedByUser_deletedMessageIsShown_InThread`, as required by the ticket.
- Added a `randomMessageDeletedEvent` builder to the shared test fixtures.

One intentional difference from the channel counterpart: attachment URLs of the old `replyTo` are not preserved when rewriting it. Thread state upserts never preserve attachment URLs today, so this follows the existing thread module behavior.

This also aligns with the iOS SDK, where the quote is a database relationship: quoting messages always render the current content of the quoted message, and a hard delete nullifies the reference.

### Testing

Covered by new unit tests (`ThreadLogicTest`, `ThreadStateLogicTest`) and by the re-enabled e2e test `QuotedReplyTests#test_originalQuoteIsDeletedByUser_deletedMessageIsShown_InThread`. The e2e test passes locally with this change and fails without it (verified by running both builds against the mock server).

Manual steps:
1. Open a channel and start a thread on one of its messages.
2. Inside the thread, quote the root message (long press > Reply) and send the reply.
3. Delete the original quoted message.
4. Expected: the quoted bubble inside the thread reply shows "Message deleted". Before this change it kept the old text.
5. Editing works the same way: edit the quoted message and the quoted bubble shows the updated text.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread consistency when quoted messages are edited or deleted.
  * Updated quoted replies appropriately for soft deletions and removed references for permanent deletions.
  * Ensured quoted-message updates are reflected across thread messages.

* **Tests**
  * Added coverage for quoted-message updates and deletions.
  * Enabled end-to-end coverage for displaying deleted quoted messages in threads.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/thread/internal/ThreadStateLogic.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/thread/internal/ThreadStateLogic.kt
@@ -64,6 +64,32 @@ internal class ThreadStateLogic(private val mutableState: ThreadMutableState) {
         )
     }
 
+    /**
+     * Updates each thread message that quotes the given message with its new content.
+     *
+     * @param quotedMessage The message whose quoting messages should be updated.
+     */
+    fun updateQuotedMessageReferences(quotedMessage: Message) {
+        updateQuotingMessages(quotedMessage.id) { it.copy(replyTo = quotedMessage) }
+    }
+
+    /**
+     * Clears the quoted message reference from each thread message quoting the given message.
+     *
+     * @param quotedMessageId The ID of the quoted message to remove references for.
+     */
+    fun deleteQuotedMessageReferences(quotedMessageId: String) {
+        updateQuotingMessages(quotedMessageId) { it.copy(replyTo = null) }
+    }
+
+    private fun updateQuotingMessages(quotedMessageId: String, update: (Message) -> Message) {
+        val quotingMessages = mutableState.rawMessage.value.values
+            .filter { it.replyTo?.id == quotedMessageId || it.replyMessageId == quotedMessageId }
+        if (quotingMessages.isNotEmpty()) {
+            mutableState.upsertMessages(quotingMessages.map(update))
+        }
+    }
+
     private fun isMessageNewerThanCurrent(currentMessage: Message?, newMessage: Message): Boolean {
         return if (newMessage.syncStatus == SyncStatus.COMPLETED) {
             (currentMessage?.lastUpdateTime() ?: NEVER.time) <= newMessage.lastUpdateTime()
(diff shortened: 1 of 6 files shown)


andremion:
@coderabbit review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/channel/thread/internal/ThreadStateLogic.kt
```diff
@@ -64,6 +64,32 @@ internal class ThreadStateLogic(private val mutableState: ThreadMutableState) {
...
+        updateQuotingMessages(quotedMessage.id) { it.copy(replyTo = quotedMessage) }
+    }
+
+    /**
+     * Clears the quoted message reference from each thread message quoting the given message.
+     *
+     * @param quotedMessageId The ID of the quoted message to remove references for.
+     */
+    fun deleteQuotedMessageReferences(quotedMessageId: String) {
+        updateQuotingMessages(quotedMessageId) { it.copy(replyTo = null) }
```
VelikovPetar, line 82: Should we also set `message.replyMessageId = null` here?

andremion: I kept it on purpose, for consistency in two directions. It mirrors the channel implementation (`ChannelStateImpl.deleteQuotedMessageReferences` also clears only `replyTo`). And it matches the server state: after a hard delete, queries still return the quoting message with `quoted_message_id` set, only the `quoted_message` object is gone. If we cleared it locally, the next sync would set it back anyway. The UI renders the quote bubble from `replyTo` only, so the remaining id has no visible effect.

andremion: If I'm missing something here, I'm happy to change it.

VelikovPetar: Aham I see, I thought that `message.replyMessageId` would be cleared as well server-side. If it is preserved, let's not clear it!

VelikovPetar (approved):

PR #6587, opened 2026-07-24
Treat the duplicate message id send error as a successful delivery

andremion:
### Goal

When a message send is retried after its response was lost (for example, the connection drops before the ack is processed), the backend rejects the retry with a 400 input error: code 4, "a message with ID x already exists". The message was in fact delivered by the first attempt, but the SDK marked it as `FAILED_PERMANENTLY`. The user then sees a delivered message as failed, and a manual resend fails again with the same error, so the message can never leave the failed state. The iOS SDK already treats this rejection as a successful send.

This is the SDK side of the fix for the message list crash found in the e2e nightly (`test_userAddsReactionWhileOffline`): the mock server used to accept the duplicate send and store a second copy with the same id, and the channel query response then crashed the Compose message list with a duplicate `LazyColumn` key. The mock server now mirrors the backend rejection (see https://github.com/GetStream/stream-chat-test-mock-server/pull/51), which makes this SDK path reachable in e2e.

Resolves AND-1320

### Implementation

- Added an internal `Error.isDuplicateMessageError()` extension that detects the backend rejection: `Error.NetworkError` with server error code 4 (`ChatErrorCode.VALIDATION_ERROR`) and a message containing "already exists". This is the same detection the iOS SDK uses in `MessageRepository`.
- `SendMessageListenerState` and `SendMessageListenerDatabase` now route this failure to their existing success handling, so the message is stored with `SyncStatus.COMPLETED` instead of `FAILED_PERMANENTLY`.

### Testing

Unit tests: new cases in `SendMessageListenerStateTest` and `SendMessageListenerDatabaseTest` cover the duplicate rejection (message marked as completed) and a different validation error (message still marked as failed permanently).

Manual steps:
1. Open a channel and send a message, then cut connectivity immediately so the send response is lost (on an emulator: `adb shell svc wifi disable` and `adb shell svc data disable` right after tapping send). The message must stay in the pending state.
2. Restore connectivity and wait for the reconnect sync to retry the send.
3. Before this change, the retried message could show the failed indicator even though the first attempt delivered it. After this change, it shows as sent.

E2e: `test_userAddsReactionWhileOffline` exercises this flow when the send races the offline switch, together with the mock server PR above. Verified locally against the updated mock server (test passes; the duplicate send returns 400 and the message list contains the message once).

## Summary by CodeRabbit

* **Bug Fixes**
  * Duplicate message-send responses are now treated as successful, preventing messages from being incorrectly marked as permanently failed.
  * Messages affected by duplicate IDs are stored with a completed sync status.
  * Other validation errors continue to be handled as permanent failures.

* **Tests**
  * Added coverage for duplicate message IDs and other validation-error scenarios.


andremion:
@coderabbit review

gpunto (approved):

PR #6589, opened 2026-07-24
Fix jumping to a quoted message in a thread when the reply is not loaded yet

andremion:
### Goal

Tapping a quoted message inside a thread did nothing when the quoted reply was not loaded in the thread yet. The controller only fetched the target `Message` object and set it as focused, but the focus produces a scroll only when the message is already part of the loaded thread page, and nothing loaded the missing page. The channel list does not have this problem because its focus path falls back to `loadMessageById`, which loads the page around the target.

Resolves AND-960

### Implementation

- **client**: new `ChatClient.getRepliesAround(parentId, aroundId, limit)` that fetches the replies page around a given reply (`id_around` pagination, the same call the iOS SDK uses for this feature). It mirrors the `getRepliesMore`/`getNewerReplies` contract: same plugin hook order, precondition, and call sharing. New `ThreadQueryListener.onGetRepliesAroundRequest/Result` hooks have default implementations, so the change is additive. The state listener updates the fetched bounds and derives the end-of-older/newer flags from where the target sits in the page, following the iOS pagination state handler. The initial replies page now marks the newer end as reached so the flag is reliable in both load orders.
- **ui-common**: `focusThreadMessage` loads the page around the target when it is not in the thread state, and only then sets the focus. After such a jump the thread can page newer replies again: `endOfNewMessagesReached` of the thread list mirrors `ThreadState.endOfNewerMessages` instead of a hardcoded `true`, and `loadNewerMessagesInThread` is gated by the pagination flags instead of `threadLoadOrderOlderToNewer`. Also fixes `removeMessageFocus`, which wrote the focus-removed items into the channel list state while in thread mode.
- **e2e**: re-enables the four `QuotedReplyTests` thread jump tests that were parked on this ticket.

The only public API change is the new `ChatClient.getRepliesAround` function.

### 🎨 UI Changes

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
<video src="https://github.com/user-attachments/assets/b11f8e03-9808-402e-b8f7-22ded971dd9b" controls="controls" muted="muted" />
</td>
<td>
<video src="https://github.com/user-attachments/assets/d56be388-391b-484f-9a0b-024d9e1507b1" controls="controls" muted="muted" />
</td>
</tr>
</tbody>
</table>

### Testing

Manual steps in the Compose sample app (demo flavor):

1. Log in with any predefined user and open a channel.
2. Send a message, open its thread, and send at least 31 replies.
3. Quote the first reply and send it.
4. Log out and log in again, so the thread state is fresh.
5. Open the thread again. Only the newest 30 replies load.
6. Tap the quoted preview inside the newest reply.

Expected: the thread loads the page around the quoted reply, scrolls up to it, and highlights it. After the jump, the scroll to bottom button shows, and scrolling down pages the newer replies back in.

Automated coverage:

- The four re-enabled e2e tests pass locally against the mock server: `test_quotedReplyNotInList_whenUserAddsQuotedReply_InThread`, `test_quotedReplyNotInList_whenParticipantAddsQuotedReply_Message_InThread`, `test_quotedReplyNotInList_whenParticipantAddsQuotedReply_File_InThread`, `test_quotedReplyNotInList_whenParticipantAddsQuotedReply_Giphy_InThread`. The whole `QuotedReplyTests` class was also run locally.
- New unit tests: `ChatClient` hook order for `getRepliesAround`, `DistinctChatApi` deduplication, and the end-flag calculation in `ThreadQueryListenerState` (target in the middle, first half, second half, and page smaller than the limit), plus persistence in `ThreadQueryListenerDatabase`.

## Summary by CodeRabbit

* **New Features**
  * Added support for loading thread replies surrounding a specific message.
  * Focusing a missing thread reply now automatically loads nearby replies first.
  * Improved thread pagination state tracking for older and newer replies.

* **Bug Fixes**
  * Thread replies loaded around a message are now stored for offline access.
  * Corrected message focus updates when viewing threaded conversations.
  * Re-enabled end-to-end coverage for quoted replies.


andremion:
@coderabbit review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/listener/internal/ThreadQueryListenerState.kt
```diff
@@ -98,6 +100,53 @@ internal class ThreadQueryListenerState(
...
+        val midPoint = Math.round(messages.size / 2.0).toInt() - 1
+        val secondHalf = messages.subList(midPoint + 1, messages.size)
+        when {
+            messages[midPoint].id == aroundId -> {
+                threadLogic.setEndOfOlderMessages(false)
+                threadLogic.setEndOfNewerMessages(false)
+            }
+            secondHalf.any { it.id == aroundId } -> threadLogic.setEndOfNewerMessages(true)
+            else -> threadLogic.setEndOfOlderMessages(true)
+        }
```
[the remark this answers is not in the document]
andremion: 563a89e0e469466d9bc7a9fe8647f4b6b3442420

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/DistinctChatApi.kt
```diff
@@ -91,6 +92,16 @@ internal class DistinctChatApi(
...
+    override fun getRepliesAround(parentId: String, aroundId: String, limit: Int): Call<List<Message>> {
+        val uniqueKey = GetRepliesAroundHash(parentId, aroundId, limit).hashCode()
+        StreamLog.d(TAG) {
+            "[getRepliesAround] parentId: $parentId, aroundId: $aroundId, limit: $limit, uniqueKey: $uniqueKey"
+        }
+        return getOrCreate(uniqueKey) {
+            delegate.getRepliesAround(parentId, aroundId, limit)
+        }
+    }
+
```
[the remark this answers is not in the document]
andremion: getOrCreate is a pre-existing helper shared by all the distinct calls, and this PR only adds one more caller in the same style. Changing the concurrency behavior of every distinct call deserves its own change, so I would keep it out of this fix.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -1484,7 +1498,11 @@ public class MessageListController(
                         it
                     }
                 }
-                setMessageListState(_messageListState.value.copy(messageItems = messages))
+                if (isInThread) {
+                    _threadListState.value = _threadListState.value.copy(messageItems = messages)
+                } else {
+                    setMessageListState(_messageListState.value.copy(messageItems = messages))
+                }
```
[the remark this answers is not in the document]
andremion: Evaluating the mode at fire time keeps the read and the write on the same list state. If we captured the mode before the delay, leaving the thread within those 2 seconds would write the captured items into a thread state that enterNormalMode already reset. The highlight itself is cleared through the focusedMessage flow in both cases, so the current shape looks safer to me.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api/internal/DistinctChatApi.kt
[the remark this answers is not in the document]
andremion: Tracked in AND-1327.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/listener/internal/ThreadQueryListenerState.kt
```diff
@@ -98,6 +97,58 @@ internal class ThreadQueryListenerState(
...
+    private fun setEndFlagsFromAroundPage(threadLogic: ThreadLogic, messages: List<Message>, aroundId: String) {
+        if (messages.isEmpty()) return
+        val midPoint = (messages.size / 2.0).roundToInt() - 1
+        val middleSize = if (messages.size % 2 == 0) 2 else 1
+        val middleEnd = (midPoint + middleSize).coerceAtMost(messages.size)
+        val middle = messages.subList(midPoint, middleEnd)
+        val secondHalf = messages.subList(middleEnd, messages.size)
+        when {
+            middle.any { it.id == aroundId } -> {
+                threadLogic.setEndOfOlderMessages(false)
```
VelikovPetar, line 144: Question: Since now we can end up in a state where we are in the middle of the thread without having the newest messages loaded - shouldn't we update the logic around inserting new messages as well? Wouldn't it be possible to end up with a gap in the messages, if a new message arrives, while we are in the middle of a thread?

andremion: Good question. A new message cannot create a new gap here. In the default load order the newest page is always loaded when the thread is opened (the jump entry also enters the thread first and only then loads the around page), and socket replies are upserted next to that newest window, so the tail stays contiguous.

The gap that does exist is the one the jump itself creates, between the around page and the previously loaded window. It fills through the newer-replies paging: after the jump `newestInThread` points at the around page, `endOfNewMessagesReached` becomes false, and reaching the bottom of the list triggers `getNewerReplies` until the thread reports no more pages. To be fair, while scrolling through that region there is no loading indicator, so the missing replies only load once that trigger fires. The iOS SDK has the same accumulate-and-fill model for thread jumps.

The only mode where the tail itself can be missing while replies arrive is `threadLoadOrderOlderToNewer`, and that insert behavior is unchanged by this PR.

If we want gap-aware inserts or a windowed thread state on top of this, that would be new behavior on both platforms (iOS has the same model today), so I would prefer to raise it as a cross-SDK discussion instead of changing only the Android state model. WDYT?

VelikovPetar: Hmm interesting, so the gap created by the jump, would be filled when we scroll to the bottom of the thread? (we would be using the jump-page anchor message ID) 

IMO we should have proper gap-aware inserts, but if we would be introducing a parity gap, I would say to raise a discussion before committing to this.

andremion: Yes, exactly. One small precision: the cursor is not the tapped message id but the newest message of the around page (`newestInThread` is updated to the last message of each fetched page). So every time the list reaches the bottom it loads the page after the current cursor and moves it forward, until the backend returns a short page and `endOfNewerMessages` becomes true.

Agreed on the discussion, and to be clear about scope: this PR follows the iOS model, so it does not introduce a parity gap by itself. The parity gap would only appear if we add gap-aware inserts on Android alone, so the cross-SDK discussion is for that follow-up, not a blocker for this PR. I will start that discussion separately. Does that match what you meant?

andremion: Follow-up tracked in AND-1331.

VelikovPetar (approved):

PR #6590, opened 2026-07-27
Build on compileSdk/targetSdk 37 (Android 17): AGP 9.1 + Gradle 9.3 + Kotlin 2.2

rahul-lohra:
### Goal

Build against compileSdk / targetSdk 37 (Android 17) and align with the shared Android toolchain (AGP 9 / Gradle 9), matching the Video SDK.

## Summary

Getting to compileSdk 37 required moving the whole toolchain to **AGP 9.1.1 / Gradle 9.3.1 / JDK 21** (AGP 8.10.1 doesn’t officially support API 37, and the shared-toolchain goal targets AGP 9). That, in turn, cascaded into a set of dependency, CI, build-script and test-infrastructure fixes — Gradle 9 removed several implicit conveniences the old build relied on, and Paparazzi/Shot needed rework for AGP 9 + API 37.

**Notable:** **no production source code changes were required** — chat was already on Kotlin 2.2, so none of the Kotlin-2 strictness fixes the Video SDK needed applied here. Every change below is build config, CI, or test infrastructure.

### 1. Toolchain updates

| Item | From | To | Reason |
|---|---|---|---|
| `compileSdk` | 35 | **37** | Build against Android 17 |
| `targetSdk` (library lint/test) | 34 | **37** | Align lint & instrumentation target |
| `sampleTargetSdk` (sample apps) | 35 | **37** | Opt sample apps into API-37 runtime behavior |
| Android Gradle Plugin | 8.10.1 | **9.1.1** | 8.10.1 doesn't officially support compileSdk 37; 9.1.1 does + shared-toolchain parity |
| Gradle | 8.11.1 | **9.3.1** | Required by AGP 9 |
| JDK (build & CI) | 17 | **21** | Required by AGP 9 tooling **and** Paparazzi 2.0.0-alpha05 |
| Kotlin / KSP | 2.2.0 / 2.2.0-2.0.2 | **unchanged** | Already on Kotlin 2.x — no bump needed (this is why no source changes were required) |

---

### 2. Dependency updates

| Dependency | From | To | Reason |
|---|---|---|---|
| `io.getstream:streamConventions` (build-logic plugin) | 0.13.1 | **0.13.2-SNAPSHOT** | AGP-9-compatible build conventions (0.13.1 breaks on AGP 9's non-generic `CommonExtension`) |
| `androidx.baselineprofile` / benchmark (`macroBenchmark`) | 1.2.3 | **1.5.0-alpha07** | Older versions can't resolve AGP 9's `TestExtension` |
| `app.cash.paparazzi` | 1.3.5 | **2.0.0-alpha05** | Only Paparazzi line compatible with AGP 9 / Gradle 9 (1.3.5 fails on Gradle 9) |
| `androidx.test:core` | 1.5.0 | **1.6.1** | Align androidx.test family (consistent-resolution conflict on Gradle 9) |
| `androidx.test.ext:junit` | 1.1.5 | **1.2.1** | ↑ same family alignment |
| `androidx.test:runner` | 1.5.2 | **1.6.2** | ↑ same family alignment |
| `androidx.test:{core, monitor}`, `androidx.test.services:storage` (forced) | 1.5.0 / 1.6.0 / 1.4.2 | **1.6.1 / 1.7.2 / 1.5.0** | Force-aligned across all configs — Shot drags old versions onto the main classpath; AGP consistent resolution then can't reconcile main vs androidTest |
| `androidx.collection` | (transitive, < 1.4.0) | **1.5.0 (explicit)** | `AudioPlayerState` uses `IntFloatMap`/`intFloatMapOf` (added in 1.4.0); the new graph resolved it too low — declared explicitly as `api` |
| `org.junit.platform:junit-platform-launcher` | (none) | **1.11.3** | Gradle 9 no longer auto-provisions the JUnit Platform launcher for `useJUnitPlatform()` modules |
| `com.pascalwelsch.gitversioner` | 0.5.0 | **removed** | Calls the Gradle-9-removed `Project.exec(Action)`; unmaintained, no Gradle-9 release |

---

### 3. CI updates

| Item | From | To | Reason |
|---|---|---|---|
| Shared conventions reusable workflows — `android-ci.yml` (pr-checks), `sdk-size-checks.yml`, `pr-quality.yml` | `@v0.13.1` / pinned SHA | **`@feature/rahullohra/build-on-sdk-37`** | Point at the JDK-21 / AGP-9 conventions branch (old refs provisioned JDK 17 → build failed) |
| pr-checks `detekt` job `setup-gradle` action | `@v0.13.1` | **`@feature/rahullohra/build-on-sdk-37`** | Same — JDK 21 |
| `setup-java` across workflows | JDK 17 / `@main` | **JDK 21** | AGP 9 tooling & Paparazzi 2.0.0-alpha05 require JDK 21 |
| `.github/workflows/record-paparazzi.yml` | (none) | **added** | Record Paparazzi goldens on the CI runner (Linux/amd64) — layoutlib pixels are OS/arch-dependent, so goldens must be produced where they're verified |
| `scripts/paparazzi-pull-goldens.sh` | (none) | **added** | Download CI-recorded goldens locally (via `gh`) and overlay them for review + commit |

---

### 4. Build script updates

| Location | From | To | Reason |
|---|---|---|---|
| `gradle/wrapper/gradle-wrapper.properties` | `gradle-8.11.1` | **`gradle-9.3.1`** | Gradle 9 (required by AGP 9) |
| `gradle.properties` | `android.suppressUnsupportedCompileSdk=34` | **`=37`** | Match compileSdk (silence the "untested compileSdk" warning) |
| `gradle.properties` | (none) | **`android.builtInKotlin=false`** | AGP 9 registers its own `kotlin` extension; opt out so the standalone KGP keeps working (avoids duplicate-extension crash) |
| `gradle.properties` | (none) | **`android.newDsl=false`** | Restore AGP's old DSL so standalone KGP's `BaseExtension` cast works (fixes `TestExtension → BaseExtension` crash) |
| `build.gradle.kts` (root) | `alias(libs.plugins.gitversioner)` | **removed** | gitversioner removed |
| `build.gradle.kts` (root) | (none) | **`resolutionStrategy.force(androidx.test:*)`** | Force-align androidx.test across all configs (consistent-resolution fix) |
| `settings.gradle.kts` | jitpack pluginManagement repo + gitversioner `resolutionStrategy` | **removed** | gitversioner removed (jitpack entry existed only for it) |
| `settings.gradle.kts` | (none) | **Central snapshots repo (`io.getstream.*`)** | Resolve the `streamConventions` 0.13.2-SNAPSHOT plugin |
| `benchmark/build.gradle.kts` | `kotlinOptions { jvmTarget = "11" }` | **`kotlin { compilerOptions { jvmTarget.set(JVM_11) } }`** | AGP 9 hard-errors on the old DSL |
| `benchmark/build.gradle.kts` | `testOptions.managedDevices.devices {}` | **`…managedDevices.allDevices {}`** | AGP 9 DSL rename |
| `ui-guides` & `ui-uitests/build.gradle.kts` | `packagingOptions {}` | **`packaging {}`** | AGP 9 DSL rename |
| `ui-uitests/build.gradle.kts` | `lintOptions {}` | **`lint {}`** | AGP 9 DSL rename |
| `compose/build.gradle.kts` | `composeCompiler { enableStrongSkippingMode = true; stabilityConfigurationFile = … }` | **removed `enableStrongSkippingMode`; `stabilityConfigurationFiles.add(…)`** | Deprecated/renamed Compose-compiler DSL |
| `client`, `ui-common`, `ui-components`, `compose` `build.gradle.kts` | `baselineProfileOutputDir = "."` | **removed** | AGP 9's `prepareReleaseArtProfile` recursively parses `src/main` (manifest, res) as ART-profile rules → `Illegal token '<'` |
| `ui-common/build.gradle.kts` | (none) | **`api(libs.androidx.collection)`** | `IntFloatMap` leaks into `AudioPlayerState`'s public API |
| `core/build.gradle.kts` | (none) | **`testRuntimeOnly(libs.junit.platform.launcher)`** | Gradle 9 launcher (manual `useJUnitPlatform()` module) |
| `compose` & `ui-components/build.gradle.kts` | (none) | **`systemProperty("paparazzi.compileSdk", …)`** in `Test` task | Forward `-Ppaparazzi.compileSdk` to the forked test worker |
| `scripts/sample-app-versioner.gradle` | `gitVersioner { … }` block | **inline `git rev-list --count` / `rev-parse --short` + `.toString()`** | gitversioner removed; `.toString()` coerces the Groovy `GString` so the sample's `as String` cast doesn't fail |

---

## 5. Source code (production) updates

| Location | From | To | Reason |
|---|---|---|---|
| — | — | — | **None.** No `src/main`/release source changes were required for compileSdk 37 — chat was already on Kotlin 2.2, so no Kotlin-2 strictness or API migrations were needed, and no new `NewApi` lint surfaced. |

---

## 6. Test code (test / debug infrastructure) updates

| Location | From | To | Reason |
|---|---|---|---|
| `PaparazziComposeTest.kt` / `PaparazziViewTest.kt` | inline `Paparazzi(...)` per test | **`createPaparazzi(...)` factory** (render-SDK configurable via `-Ppaparazzi.compileSdk`, default 36) | Centralize Paparazzi construction into one place |
| ~90 compose tests + 2 view tests | `override val paparazzi = Paparazzi(...)` | **`= createPaparazzi(...)`** (and dropped the now-unused `Paparazzi` import) | Route every test through the factory |
| `PollMoreOptionsDialogTest.kt` | `override val paparazzi: Paparazzi = …` | **`override val paparazzi = …`** (dropped `: Paparazzi`) | Its explicit type annotation orphaned the removed import |
| `PaparazziComposeTest.kt` | (none) | **error-swallowing helper: `try/catch` + reflective logger-clear** | Swallow the two known Paparazzi 2.0.0-alpha05 / layoutlib-16.2.1 crashes — #2342 (`Thread.setPosixNicenessInternal` on a background `HandlerThread`) and #2373 (`WindowManagerImpl.removeView` NPE on Popup/Dialog teardown); the snapshot is captured before both throw |
| `ui-common` & `compose` `src/test/resources/robolectric.properties` | (none) | **`sdk=28`** | Robolectric 4.10.3 caps at API 34; compileSdk 37 makes AGP feed it `targetSdk=37` → pin the default low (matches `client`) |
| `compose/src/test/java/com/android/org/conscrypt/TrustManagerImpl.java` and `.../src/debug/...` | (none) | **added empty stub** | layoutlib lacks conscrypt's `TrustManagerImpl`; when Coil builds an OkHttp client during render, OkHttp throws an uncaught `NoClassDefFoundError`. The stub lets it fall back gracefully (test = Paparazzi, debug = Compose Preview) |

---
### Known temporary items (remove when upstream catches up)

- **`android.builtInKotlin=false` / `android.newDsl=false`** — AGP 9 escape hatches; removed in AGP 10. Migrate to built-in Kotlin before then.
- **Paparazzi crash workarounds** (`TrustManagerImpl` stub, error-swallow + logger-clear) — fixed upstream by **layoutlib 16.2.3** (Paparazzi alpha05.2 / alpha06); remove when adopted.
- **`streamConventions 0.13.2-SNAPSHOT`** and the **`@feature/rahullohra/build-on-sdk-37`** CI refs — move to a released tag once the shared conventions repo publishes AGP-9 support.
- **androidx.test `force(...)` block** — can be dropped if Shot is retired/updated (it's the source of the old transitive versions).

### 🎨 UI Changes

None

### Testing

Smoke Test the app 

## Summary by CodeRabbit

- **New Features**
  - Added an automated workflow for recording and sharing visual snapshot results.
  - Added a helper for retrieving the latest visual snapshots for local review.

- **Bug Fixes**
  - Improved compatibility with newer Android SDK and build tooling.
  - Increased reliability of Compose previews, screenshot tests, and release builds.

- **Tests**
  - Standardized visual test rendering across supported devices and configurations.
  - Updated continuous integration checks to use Java 21 for consistent results.


gradle/libs.versions.toml
```diff
@@ -77,7 +78,7 @@ room = "2.7.2"
 shimmer = "0.5.0"
 shimmerCompose = "1.3.1"
 shot = "6.1.0"
-streamConventions = "0.13.1"
+streamConventions = "0.13.2-SNAPSHOT"
```
gpunto, line 81: Snapshots are pruned after some time, so this commit will start failing once that happens.

I'd publish a stable conventions version. We can use a version suffix if we want to keep the version separated from the regular ones for now.

.github/workflows/record-paparazzi.yml
gpunto, line 1: Can't we regenerate them locally like today? I checked on my machine and:
- When recording: some snapshots change, but the image diff shows 0 or close to 0 pixel changes
- When running validation: the check passes

I'd avoid having to rely on CI for regenerating snapshots if possible, as it will slow us down compared to today.

Also one less workflow + script to maintain.

.github/workflows/attach-release-apk.yml
```diff
@@ -21,7 +21,9 @@ jobs:
           ref: ${{ github.event.release.tag_name }}
           fetch-depth: 0
           persist-credentials: false
-      - uses: GetStream/android-ci-actions/actions/setup-java@main
+      - uses: GetStream/stream-build-conventions-android/.github/workflows/android-ci.yml@feature/rahullohra/build-on-sdk-37
```
gpunto: I wouldn't use temporary branch names. We can publish a release with a version suffix (e.g. 0.14.0-sdk37, 1.0.0-beta, or something along those lines) and point to that. Adding support in https://github.com/GetStream/stream-build-conventions-android/pull/73

Aside from that, this seems incorrect. The step was using an action but now points to a workflow 🤔

scripts/sample-app-versioner.gradle
```diff
@@ -8,17 +8,16 @@ static Boolean gitIsAvailable() {
...
-        formatter = { gitVersioner ->
-            "1.${gitVersioner.versionCode}.${gitVersioner.featureBranchCommitCount}.${gitVersioner.currentSha1Short}"
-        }
-    }
-
-    ext["sampleAppVersionCode"] = gitVersioner.versionCode
-    ext["sampleAppVersionName"] = gitVersioner.versionName
+    // Derive the demo-app version from git. Replaces the gitVersioner plugin, which relied on
+    // the Gradle-9-removed Project.exec(Action) API. Groovy's String.execute() is plain JVM and
+    // Gradle-9-safe. versionCode = commit count on HEAD (monotonic); versionName = 1.<count>.<sha>.
```
gpunto, line 17: I'd remove this comment tbh. gitVersioner won't be in the repo anymore so it's not something a reader will care about + we don't have to justify that the `execute` function works.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/PaparazziComposeTest.kt
```diff
@@ -71,18 +73,99 @@ private fun DeviceConfig.atHdpi(): DeviceConfig = copy(
...
+        environment = detectEnvironment().copy(compileSdkVersion = paparazziCompileSdk),
+        deviceConfig = deviceConfig,
+        renderingMode = renderingMode,
+    )
+
+    /**
+     * Wraps [Paparazzi.snapshot], swallowing the known alpha05/layoutlib-16.2.1 teardown bugs
+     * (see [isPaparazziAlpha05LayoutlibBug]). The snapshot is captured before they throw.
+     */
+    private fun snapshotSwallowingLayoutlibBugs(content: @Composable () -> Unit) {
```
gpunto, line 151: This turns a golden mismatch into a green test.

I had Claude dig into it. It instrumented this method and ran the full compose suite on macOS: 398/400 snapshots take the clean path, 2 take the swallow path (`AudioRecordingButtonTest > button hold` and `> button locked`, both cashapp/paparazzi#2373). It then perturbed all four snapshots in that class by ~50% of pixels and re-ran `verifyPaparazziDebug`:

```
button idle      FAIL  Images differ (by 48.199162%)
button overview  FAIL  Images differ (by 49.103992%)
button hold      PASS
button locked    PASS
```

Paparazzi wrote a delta image for all four, so it caught the mismatch every time — but on the swallowed ones the teardown NPE masks the `AssertionError` and this `catch` eats it. The KDoc premise holds for the capture, not for the assertion: those two tests can no longer fail.

**Splitting the two mechanisms keeps the coverage.** They carry different risk:
- `dropKnownLayoutlibBugErrors()` is safe to keep — it only filters Paparazzi's internal logger list by those two signatures, and a golden-mismatch `AssertionError` never lands there.
- The `try/catch` around `snapshot()` is the part that eats assertions.

So: drop the `try/catch`, and `@Ignore("cashapp/paparazzi#2373")` the two tests that genuinely can't render (`button hold`, `button locked`). That leaves 398 snapshots asserting and 2 visibly disabled, instead of 400 that all look green.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.11.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.3.1-bin.zip
```
gpunto, line 3: It seems Gradle 9 silently disables our Gradle Enterprise plugin. The build on this branch prints:

```
Gradle Enterprise plugin 3.7 has been disabled as it is incompatible with this version of Gradle.
Upgrade to Gradle Enterprise plugin 3.13.1 or newer to restore functionality.
```

We should likely bump it (I think it's called Develocity now).

build.gradle.kts
```diff
@@ -98,6 +97,23 @@ subprojects {
...
+        resolutionStrategy {
+            force(
+                "androidx.test:core:1.6.1",
+                "androidx.test:core-ktx:1.6.1",
+                "androidx.test:runner:1.6.2",
+                "androidx.test:monitor:1.7.2",
+                "androidx.test.services:storage:1.5.0",
+            )
+        }
+    }
```
gpunto: It isn't Shot — it's `debugImplementation(libs.androidx.fragment.testing)` at `stream-chat-android-ui-components-sample/build.gradle.kts:176`. That's the pre-fragment-1.6 pattern (before the `fragment-testing-manifest` split), so it lands `fragment-testing` and its old `androidx.test:core:1.5.0` directly on `demoDebugRuntimeClasspath` — which is where the strict constraint AGP can't reconcile comes from.

It's also unused: that module has no `androidTest` source set and never touches `FragmentScenario`.

I removed that line and this block together and re-resolved: ui-components-sample, compose-sample and ui-uitests androidTest classpaths all resolve, and the `{strictly androidx.test:*}` constraints disappear. ui-uitests does use `FragmentScenario`, but its existing `exclude(module = "monitor")` already covers it.

Worth preferring that: `force` silently overrides every configuration in every module, and this block leaves the debug APK shipping fragment-testing anyway (just at 1.6.1). It also means "remove when Shot is retired/updated" can never actually fire.

stream-chat-android-compose/src/test/resources/robolectric.properties
gpunto, line 1: Is this config needed because of some failure? I see we're setting it in code in test classes to sdk 33. Same for the ui-common counterpart

stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.attachments.preview_MediaGalleryPreviewScreenTest_media_gallery_screen_with_share_large_file_prompt.png
VelikovPetar, line 1: Hmm this test looks suspicious - could it be an actual regression?

stream-chat-android-compose/src/test/snapshots/images/io.getstream.chat.android.compose.ui.attachments.preview_MediaGalleryPreviewScreenTest_media_gallery_screen_with_share_large_file_prompt_in_dark_mode.png
VelikovPetar, line 1: Same here.

gradle/libs.versions.toml
```diff
@@ -1,16 +1,17 @@
...
-compileSdk = "35"
-targetSdk = "34"
-sampleTargetSdk = "35"
+compileSdk = "37"
+targetSdk = "37"
+sampleTargetSdk = "37"
 minSdk = "21"
 allureKotlin = "2.4.0"
-androidGradlePlugin = "8.10.1"
+androidGradlePlugin = "9.1.1"
```
VelikovPetar, line 7: I understand that this is necessary - but could this be a potential breaking change - breaking customer builds?

gradle/libs.versions.toml
```diff
@@ -1,16 +1,17 @@
 [versions]
-compileSdk = "35"
-targetSdk = "34"
-sampleTargetSdk = "35"
+compileSdk = "37"
+targetSdk = "37"
+sampleTargetSdk = "37"
```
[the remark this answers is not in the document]
VelikovPetar: I assume there were no behavioural changes caused by the targetSdk bump? - Should we maybe do thorough manual testing before releasing this?

gpunto (commented):
Some comments by me + Claude review.

Also, I had it create a task based on the deferred follow ups (i.e. stuff that we can't do now or are out of scope). Let me know if anything looks wrong!
https://linear.app/stream/issue/AND-1371/follow-ups-from-the-agp-9-gradle-9-compilesdk-37-migration

gradle/libs.versions.toml
rahul-lohra: Correct.

The reason I’m still using a snapshot is that I’m finalizing the changes needed for the Android 37 build support, and I won’t know the final stream-conventions requirements until both the Video and Chat PRs are approved.

My plan is:

**Step 1** — Get the build-support PRs approved

1. Chat: build support for Android 37
2. Video: [build support for Android 37](https://github.com/GetStream/stream-video-android/pull/1756)

**Step 2** — Roll out stream-conventions
Once both PRs are finalized, I’ll make the required changes to stream-conventions and publish a stable conventions version.

**Step 3** — Switch to the stable version
Update the Chat and Video Android 37 PRs to use the stable stream-conventions version instead of the snapshot.

This way, we avoid publishing a stable version prematurely while the required conventions are still changing.

Does it sound good to you? I am happy to change anything you suggest

.github/workflows/record-paparazzi.yml
rahul-lohra: > Can't we regenerate them locally like today?

Yes — and your measurements match what I'd expect: Paparazzi's cross-platform rendering diff is currently tiny, so locally-recorded goldens pass. Local recording is fine as the default.

The one caveat: Paparazzi renders via a native layoutlib, and native text/anti-aliasing can differ slightly by OS + CPU arch (macOS/arm64 vs the Ubuntu/x86_64 CI). Today that difference is ~0, but a complex snapshot — or a future toolchain bump — could push a diff above maxPercentDifference, and then CI verify (which runs on Ubuntu) would fail against Mac-recorded goldens. If that ever happens, the fix is to bump the tolerance or re-record on the same OS as CI 

Separately: I need to re-record the goldens now regardless, because the newer layoutlib changed its color output — so this re-record has to happen either way.

I'm happy to drop the **record-paparazzi** workflow + pull script once the PR is approved (one less thing to maintain). We can always re-add CI recording if a cross-platform diff ever exceeds tolerance.

.github/workflows/attach-release-apk.yml
rahul-lohra: Thanks for pointing that I have replaced an action with workflow. I will review this change across all workflow and  see if I have mistakenly replaced an action with workflow

scripts/sample-app-versioner.gradle
rahul-lohra: okay, will follow that

build.gradle.kts
rahul-lohra: Thank you for the investigation, will correct this

stream-chat-android-compose/src/test/resources/robolectric.properties
rahul-lohra: > Is this config needed because of some failure?

I added it to default Robolectric to SDK 33 because, if we don’t explicitly specify the SDK programmatically, Robolectric will pick up the project’s targetSdk, which is currently 37. Robolectric hasn’t published support for SDK 37 yet.

> I see we’re setting it in code in test classes to SDK 33. Same for the ui-common counterpart.

Okay, in that case we can remove ⁠[stream-chat-android-compose/src/test/resources/robolectric.properties](https://github.com/GetStream/stream-chat-android/pull/6590/files/8f9b979fb1ec14bb89a6e44be64e07efaa317842#diff-4218689fd0f29a3efff30af9ce2c213ec256ad53aefcf99c48174d2f2e5ce8da), since the SDK is already being explicitly configured in the test classes.

.github/workflows/attach-release-apk.yml
rahul-lohra: Code is reverted and no action is required so closing this conversation

build.gradle.kts
rahul-lohra: Correct, code is reverted

gradle/wrapper/gradle-wrapper.properties
rahul-lohra: After mutual discussion, we are removing gradle enterprise for now. We will revisit it later when needed

stream-chat-android-compose/src/test/resources/robolectric.properties
rahul-lohra: After mutual discussion, it is decided to globally keep `robolectric.properties` and sdk=33 there.

rahul-lohra:
Closing this PR after internal discussion. We are postponing the rollout for the following reasons:

1. Compiling the SDK with Compile SDK 37 will require integrators to bump their app’s Compile SDK to 37 as well.
2. We introduced a breaking change in Spring, so it’s too early to roll out another breaking change that would force consumers to upgrade to Compile SDK 37.

PR #6595, opened 2026-07-28
Migrate user groups to generated models

gpunto:
### Goal

Migrate the whole user-groups surface (`/usergroups` CRUD + members) to generated `network.models`. First slice vendored from the current `/chat` master generator, and the first response-side migration. Unblocked by the upstream fix for the same-name struct collapse (#12/#15/#16): `UserGroupResponse` now carries `members` and no longer needs the local rename patch. Part of the incremental OpenAPI model migration.

### Implementation

- Vendor 13 generated `internal` models: the 4 request bodies (`Create`/`Update`/`AddUserGroupMembers`/`RemoveUserGroupMembers`), the 7 per-endpoint response wrappers (`Create`/`Get`/`Update`/`Add`/`RemoveUserGroupMembersResponse`, `ListUserGroupsResponse`, `SearchUserGroupsResponse`), plus the shared `UserGroupResponse` object and `UserGroupMember`.
- `UserGroupApi` returns the generated per-endpoint wrappers; request `@Body` types use the generated requests. `MoshiChatApi` call sites pass camelCase args and map via generic helpers `mapUserGroup { it.userGroup }` / `mapUserGroups { it.userGroups }`. `DomainMapping` gains `UserGroupResponse.toDomain()` + `UserGroupMember.toDomain()`.
- Remove the hand-written `UserGroupRequests.kt` and `UserGroupResponses.kt`. **Keep** `UserGroupDtos.kt` (`DownstreamUserGroupDto`) — still used by `Message.mentioned_groups`, which is part of the deferred message migration.
- Tests: add `Mother.randomUserGroupResponse()` + one arg provider per endpoint; retype the 8 user-group test functions.

### Testing

- `apiCheck` (no API drift — all `internal`), `detekt`, `spotlessCheck`, and the full client `testDebugUnitTest` suite pass.
- Verified on device against the real backend, full CRUD cycle: create -> get -> update -> add members -> remove members -> query -> search -> delete. Every per-endpoint response wrapper (and the nested `UserGroupResponse` object + `members`) parsed cleanly (no `JsonDataException`); request bodies serialize the expected `member_ids`/`name`/`id` shapes.

## Release Notes

* **Chores**
  * Reorganized internal API models for user group operations to improve data handling and consistency. User group management functionality remains unchanged.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -659,17 +664,22 @@ constructor(
...
+        return userGroupApi.removeUserGroupMembers(id = id, body = body)
+            .mapUserGroup(RemoveUserGroupMembersResponse::userGroup)
     }
 
-    private fun RetrofitCall<UserGroupResponse>.mapUserGroup() =
-        map { response -> with(domainMapping) { response.user_group.toDomain() } }
+    private fun <T : Any> RetrofitCall<T>.mapUserGroup(extract: (T) -> UserGroupResponse?) =
+        map { response ->
+            with(domainMapping) {
+                requireNotNull(extract(response)) { "usergroup response missing user_group" }.toDomain()
```
andremion: This changes the failure mode for a 2xx response without `user_group`. Before, the field was required in Moshi, so a missing field became a parse error and `RetrofitCall` converted it to `Result.Failure`. Now parsing succeeds and `requireNotNull` throws inside the `map` lambda. `MapCall` doesn't catch mapper exceptions (`Result.map` is `flatMap { Success(f(it)) }`, no try/catch), so on the `enqueue` path this throws inside the SDK coroutine scope, and on `await` it propagates to a caller that expects a `Result`.

The backend should always send `user_group` on success, so this is unlikely to trigger. But do we want a throw here instead of a `Result.Failure`?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
```diff
@@ -1425,6 +1426,17 @@ internal object Mother {
         updated_at = updatedAt,
     )
 
+    fun randomUserGroupResponse(): UserGroupResponse = UserGroupResponse(
```
andremion: This helper always uses `members = emptyList()` and has no override parameters, so `UserGroupMember.toDomain()` is never executed by any test. The existing `DomainMappingTest` cases only cover the legacy `DownstreamUserGroupDto` path. Since this is the first response-side migration, would it make sense to add a `DomainMappingTest` pair for `UserGroupResponse.toDomain()` / `UserGroupMember.toDomain()`, and give this helper named overrides like the sibling helpers?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -596,7 +600,7 @@ constructor(
                 createdAtGt = createdAtGt,
                 teamId = teamId,
             )
-            .mapUserGroups()
+            .mapUserGroups { it.userGroups }
```
andremion: nit: `mapUserGroup` gets member references but `mapUserGroups` gets lambdas. `ListUserGroupsResponse::userGroups` / `SearchUserGroupsResponse::userGroups` would work here too.

andremion (commented):
Nice cleanup! Left a question about the failure mode of `mapUserGroup` and a note on test coverage.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/8120358b2b

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/7371ec161c and https://github.com/GetStream/stream-chat-android/commit/56cc9486b0

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/e77be3bf89

andremion (approved):

PR #6597, opened 2026-07-29
Migrate query channels, users, and grouped channels request bodies to generated models

gpunto:
### Goal

Migrate the query-* request bodies to their generated OpenAPI models.

### Implementation

- `queryChannels` -> generated `QueryChannelsRequest` (sort now via `toSortParams()`; `filterValues`/`sortValues` via `.orEmpty()`).
- `queryUsers` -> generated `QueryUsersPayload`.
- `queryGroupedChannels` -> generated `GroupedQueryChannelsRequest` + `GroupedChannelsGroupRequest`.
- Removed the hand-written `QueryChannelsRequest`, `QueryUsersRequest`, `QueryGroupedChannelsRequest` request DTOs.
- `queryBannedUsers` left hand-written on purpose: the generated `QueryBannedUsersPayload` omits the `created_at_after/before(_or_equal)` filter fields the current request sends, so migrating it would drop those filters.

### Testing

- Updated `MoshiChatApiTest` payload assertions to the generated models.
- `spotlessApply`, `apiCheck` (no public-API change), `detekt`, full client `testDebugUnitTest`, and `lint` all pass.
- Device probe (compose-sample) confirmed the live wire bodies:
  - `queryChannels`: `sort:[{direction,field}]`, `filter_conditions`, and the new empty `filter_values:{}`/`sort_values:{}`/`member_custom_include:[]` accepted by the backend (201, real channels).
  - `queryUsers`: `filter_conditions`/`sort`/`limit`/`offset`/`presence` serialized correctly (200, real user).
  - `queryGroupedChannels`: body `{"groups":{"messaging":{}},...}` serialized correctly (endpoint is enterprise-gated, so it returns 403 on this app; serialization verified).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved channel, grouped-channel, and user query requests for more reliable API communication.
  * Preserved connection-dependent behavior when channel queries require watching or presence updates.
  * Improved handling of query filters, sorting, pagination, and message/member limits.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1463,16 +1461,16 @@ constructor(
...
-            filter_conditions = if (query.predefinedFilter != null) null else query.filter.toMap(),
-            sort = if (query.predefinedFilter != null) null else query.sort,
-            predefined_filter = query.predefinedFilter,
-            filter_values = query.filterValues,
-            sort_values = query.sortValues,
+        val request = io.getstream.chat.android.network.models.QueryChannelsRequest(
+            filterConditions = if (query.predefinedFilter != null) null else query.filter.toMap(),
+            sort = if (query.predefinedFilter != null) null else query.querySort.toSortParams(),
+            predefinedFilter = query.predefinedFilter,
+            filterValues = query.filterValues.orEmpty(),
```
andremion: The generated params are nullable, so passing `query.filterValues` / `query.sortValues` through directly would keep the old wire behavior (null key omitted). With `.orEmpty()` we now always send `filter_values: {}` and `sort_values: {}`. Was that intentional, to match the generated defaults? If so, was the predefined-filter path also probed with the empty maps? The probe in the description looks like the standard path only.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
```diff
@@ -1919,13 +1919,16 @@ internal class MoshiChatApiTest {
         sut.setConnection(userId = userId, connectionId = connectionId)
         val result = sut.queryChannels(query).await()
         // then
-        val expectedPayload = io.getstream.chat.android.client.api2.model.requests.QueryChannelsRequest(
-            filter_conditions = query.filter.toMap(),
-            sort = query.sort,
+        val expectedPayload = io.getstream.chat.android.network.models.QueryChannelsRequest(
+            filterConditions = if (query.predefinedFilter != null) null else query.filter.toMap(),
```
andremion: This expectation copies the production conditional (`if (query.predefinedFilter != null) null else ...`), so a regression in that branch logic would not fail this test. Since this query has no predefined filter, could we assert the concrete values instead (`filterConditions = query.filter.toMap()`, `sort = query.querySort.toSortParams()`)? The predefined-filter test below already covers the null side explicitly.

andremion (approved):
LGTM. Two small non-blocking comments inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/071d0f8b37

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
gpunto: done in https://github.com/GetStream/stream-chat-android/commit/03907bbe12

andremion (approved):

PR #6598, opened 2026-07-29
ci: Trigger required checks on merge_group for merge queue

gpunto:
### Goal

Closes [AND-1335](https://linear.app/stream/issue/AND-1335/enable-github-merge-queue-for-the-android-chat-repo)

Prepare the CI to support GitHub's merge queue. When a PR is added to a merge queue, GitHub fires a `merge_group` event (not `pull_request`) against a temporary queue branch. Any required status check must run on that event or the queued PR hangs indefinitely waiting for a check that never reports.

All 7 checks currently required on `develop` are produced by two workflows:
- `pr-checks.yml` → `base-android-ci / Build`, `/ Run actionlint`, `/ Run static checks`, `/ Run unit tests`
- `e2e-test.yml` → `Test compose (0)`, `(1)`, `(2)`

### Implementation

- Added a `merge_group:` trigger to `pr-checks.yml` and `e2e-test.yml` so the required checks report on merge-queue runs.
- Fixed `pr-checks.yml` concurrency key: `github.head_ref` is empty on `merge_group` events, which would collapse all queued entries into one group and cancel each other. Falls back to `github.ref` (unique per queue entry). `e2e-test.yml` already keyed on `github.ref`, so it was left as-is.

### Testing

`actionlint` passes on both files. Full validation requires the merge queue to be enabled on `develop` (a repo-settings change made after this merges); a throwaway PR should then be run through the queue to confirm all 7 checks report on the `merge_group` run.

> [!NOTE]
> `e2e-test.yml` sets `GITHUB_PR_NUM: github.event.pull_request.number`, which is empty under `merge_group`. Verified this is safe — every consumer in the `stream_actions` plugin degrades gracefully: `is_check_required` treats an empty PR number as "always run" (so E2E still executes), `allure_create_launch` falls back to naming the launch from the commit message, and `current_branch` falls back to `git_branch`. No path fails the build; the only effect is cosmetic in Allure TestOps (launch named from the commit message instead of the PR title, no PR-label tags). No follow-up guard needed.

### Rollout

Merge-queue trigger must be present on `develop` **before** the queue is enabled (merge_group workflows run from the base branch's file). Sequence: merge this PR normally → enable "Require merge queue" on the `develop` branch protection rule (keep the same 7 required checks) → validate with a test PR.

### ☑️Contributor Checklist
- [x] PR is linked to the Linear issue it resolves

## Summary by CodeRabbit

* **Chores**
  * Expanded automated checks to run for merge queue events.
  * Improved workflow concurrency handling to prevent conflicts when merge queue events lack a pull request branch reference.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -1,9 +1,11 @@
 name: Pull request CI checks
 
-on: [ pull_request ]
+on:
+  pull_request:
+  merge_group:
 
 concurrency:
-  group: ${{ github.head_ref }}
+  group: ${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -5,7 +5,7 @@ on:
   merge_group:
 
 concurrency:
-  group: ${{ github.head_ref || github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -5,7 +5,7 @@ on:
   merge_group:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -13,7 +13,7 @@ on:
         default: main
 
 concurrency:
-  group: ${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 env:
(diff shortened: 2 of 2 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -1,9 +1,11 @@
 name: Pull request CI checks
 
-on: [ pull_request ]
+on:
+  pull_request:
+  merge_group:
```
andremion, line 6: Concurrency groups are shared across workflows, and on `merge_group` this falls back to the same `github.ref` that `e2e-test.yml` uses. Since both have `cancel-in-progress: true`, one run would cancel the other. Should we prefix with the workflow name like the other workflows do?

```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
```

andremion (commented):
Left one comment about the concurrency group. The rest looks good.

.github/workflows/pr-checks.yml
gpunto: Good catch, thanks. Went a bit further and dropped the `head_ref` fallback entirely: both workflows now use `${{ github.workflow }}-${{ github.ref }}`, matching the convention in the other repos' `android.yml`. `github.ref` is never empty and is already unique per PR (`refs/pull/N/merge`) and per queue entry, so the fallback wasn't needed — and this also avoids two fork PRs with the same branch name sharing a group.

andremion (approved):

PR #6599, opened 2026-07-29
e2e: Cover the unread messages flows backed by the mock server read state

andremion:
### Goal

Cover the unread messages flows now that the mock server maintains per-user read state: the unread separator, the jump-to-unread button and its dismissal, and the mark-as-unread message action.

The new tests exposed two real client-side defects in the unread label pipeline, fixed here as well. First, on channel open the automatic mark-read zeroes the read state optimistically, and the label pipeline collects a conflating flow, so on a slow device it could observe only the zeroed state. Second, the pipeline marked a read state as processed before calculating, so a first emission arriving while the message list was still empty produced no label and permanently blocked recalculation for that read state. Either way the unread separator sometimes never rendered.

Built on the read state that GetStream/stream-chat-test-mock-server#53 added to the mock server (merged).

Related: AND-1329

### Implementation

- `MessageListController` now computes the unread label from the current read state right before triggering the mark-read call. The controller owns both sides of the race, so this makes the order deterministic. The label calculation is extracted into `computeUnreadLabel`, shared by the existing flow observation and the new capture, and `showUnreadButtonState` becomes a `StateFlow` so the capture can read its current value. No public API change; the XML message list shares the same controller, so both UIs are covered.
- `lastProcessedReadMessageId` is recorded only when the calculator produces a label, so a transient null result (for example, the message list has not loaded yet) no longer blocks recalculation. The capture before mark-read is the natural recovery point: mark-read only fires when the last message is visible, so the messages are always loaded there.
- `UnreadMessagesTests`: the three tests `@Ignore`d in #6591 (Allure 11453-11455) run again, and a new `test_userMarksMessageAsUnread` (Allure 6073) covers the mark-as-unread context menu action end to end, asserting the unread separator and the channel list badge.
- All four tests wait for the user's message delivery before the participant acts, and the two count-asserting tests also wait for the channel list badge to settle before reopening the channel, so the assertions never race the message order or the event processing.
- New `assertChannelUnreadCount`: the badge exposes only a content description (`clearAndSetSemantics` in `UnreadCountIndicator` removes the child text and tag from the semantics tree), so the assertion matches the `stream_compose_channel_item_unread` plural. The stale `Stream_UnreadCountIndicator` selector is removed.

### Testing

- Two new unit tests in `MessageListControllerTests`: one pins the ordering contract (when the mark-read call goes out, the label is already computed, and it keeps its count after the read state is zeroed), and one reproduces the empty-list poisoning deterministically (read state arrives before the messages, then the messages load and the label must appear; it fails against the old code). All 903 ui-common unit tests pass, plus detekt and apiCheck.
- Ran the four e2e tests locally against the mock server PR branch (API 35 emulator, per-test data isolation): 6/6 including repeats.
- A conflation race cannot be reproduced under a serial test scheduler, so the empirical evidence is the full nightly matrix, where the label race reproduced in every earlier run (roughly one of the four tests per run): https://github.com/GetStream/stream-chat-android/actions/runs/30550107723 is the run with the complete fix against the merged mock main.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
@@ -186,8 +186,8 @@ class UserRobot {
         return this
     }
 
-    fun markMessageAsUnread(messageCellIndex: Int = 0): UserRobot {
-        openContextMenu(messageCellIndex)
+    fun markMessageAsUnread(text: String): UserRobot {
+        openContextMenu(text)
         ContextMenu.markAsUnread.waitToAppear().click()
         return this
     }
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
@@ -186,6 +186,12 @@ class UserRobot {
         return this
     }
 
+    fun markMessageAsUnread(messageCellIndex: Int = 0): UserRobot {
+        openContextMenu(messageCellIndex)
+        ContextMenu.markAsUnread.waitToAppear().click()
+        return this
+    }
+
     fun pinMessage(messageCellIndex: Int = 0): UserRobot {
         openContextMenu(messageCellIndex)
         ContextMenu.pin.waitToAppearAndClick()
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
@@ -188,7 +188,7 @@ class UserRobot {
 
     fun markMessageAsUnread(text: String): UserRobot {
         openContextMenu(text)
-        ContextMenu.markAsUnread.waitToAppear().click()
+        ContextMenu.markAsUnread.waitToAppearAndClick()
         return this
     }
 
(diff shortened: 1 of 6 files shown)


andremion:
@coderabbit review

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/UnreadMessagesTests.kt
```diff
@@ -79,14 +92,41 @@ class UnreadMessagesTests : StreamTestCase() {
...
+        }
+        step("AND the message is delivered") {
+            userRobot.assertMessageDeliveryStatus(MessageDeliveryStatus.SENT)
+        }
+        step("AND participant sends messages") {
+            participantRobot.sendMultipleMessages(text = "New", count = unreadCount)
+            userRobot.assertMessage("New-$unreadCount")
+        }
+        step("WHEN user marks the first participant message as unread") {
+            userRobot.markMessageAsUnread(messageCellIndex = 1)
```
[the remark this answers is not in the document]
andremion: c3d194b3e777188853062cfcf0ea598570c33245

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
```diff
@@ -186,6 +186,12 @@ class UserRobot {
         return this
     }
 
+    fun markMessageAsUnread(text: String): UserRobot {
+        openContextMenu(text)
+        ContextMenu.markAsUnread.waitToAppear().click()
```
gpunto: Should this also use the stale-safe impl `waitToAppearAndClick`?

gpunto (approved):
LG, just one question

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
andremion: Good catch! 045a6b48b0f1b2261c3f4158271472874a202a3f

PR #6600, opened 2026-07-29
Migrate queryThreads, queryReminders, queryDrafts, and partial-update-thread request bodies to generated models

gpunto:
### Goal

Migrate the `queryThreads`, `queryReminders`, and `queryDrafts` request bodies (`POST /threads`, `POST /reminders/query`, `POST /drafts/query`) to the generated `QueryThreadsRequest` / `QueryRemindersRequest` / `QueryDraftsRequest` network models. Completes the query-payload request-body cluster. Part of the incremental OpenAPI model migration.

### Implementation

- Add the three generated `internal` request models in `network.models`; remove the hand-written `api2.model.requests` versions. `MoshiChatApi`/`ThreadsApi`/`RemindersApi`/`MessageApi`/tests use the generated types by their real names.
- Each `sort` field is now `List<SortParamRequest>` (was `List<Map<String, Any>>`), so the call sites switch from `sort.toDto()` to `sort.toSortParams()`. `filter?.toMap()` fits the generated `Map<String, Any?>` field. `queryThreads`' call site switches its named args to the generated camelCase params (`memberLimit`/`participantLimit`/`replyLimit`). No wire-shape change.

### Testing

- `spotlessCheck`, `apiCheck` (no API drift), `detekt`, and the full client `testDebugUnitTest` suite pass.
- Verified on device against the real backend, all three endpoints with a `created_at desc` sort: `POST /threads`, `POST /reminders/query`, and `POST /drafts/query` each send `"sort":[{"direction":-1,"field":"created_at"}]` and return `201`. `queryThreads` also correctly serializes `member_limit`/`participant_limit`/`reply_limit`/`watch`.

_Also folds in the `partialUpdateThread` request-body migration (`PATCH /threads/{id}` → generated `UpdateThreadPartialRequest`) — bundled here to keep the thread-request migrations together._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request handling for draft, reminder, and thread queries.
  * Ensured sorting and partial thread updates are serialized consistently.
  * Preserved correct behavior for default and optional query parameters.

* **Tests**
  * Added coverage for complete and default request payloads.
  * Verified serialized query and thread-update data against expected JSON.


stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/QueryRemindersRequest.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.squareup.moshi.Json
+import kotlin.collections.List
+import kotlin.collections.Map
+
+/**
+ *
+ */
+
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class QueryRemindersRequest(
```
andremion, line 35: Sonar flags this file and `QueryDraftsRequest.kt` as identical.

Worth excluding generated `network/models` from Sonar duplication, since the series will keep producing these?

andremion (commented):
LGTM. Three notes, none blocking.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
```diff
@@ -2404,8 +2411,19 @@ internal class MoshiChatApiTest {
         val request = Mother.randomQueryThreadsRequest()
         val result = sut.queryThreads(request).await()
         // then
+        val expectedBody = io.getstream.chat.android.network.models.QueryThreadsRequest(
```
andremion, line 2414: `Mother.randomQueryThreadsRequest()` doesn't set `filter`, so `filter?.toMap()` never runs here. Add a `filter` parameter to the helper and pass a non-null one?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -327,7 +327,7 @@ constructor(
                 filter = filter.toMap(),
                 limit = limit,
                 next = next,
-                sort = sort.toDto(),
+                sort = sort.toSortParams(),
```
andremion, line 330: `toSortParams()` keeps only `field` and `direction`, with `as? String` / `as? Number` falling back to null. A custom `QuerySorter` returning anything else used to reach the server untouched. Nothing in the SDK does that, and `queryChannels`/`queryUsers` already behave this way, so no change needed. Just flagging it now covers threads, reminders and drafts too.

andremion (approved):

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
gpunto: Addressed in `bf075aab51` — `Mother.randomQueryThreadsRequest()` now takes a `filter`, and the test passes a non-null one so `filter?.toMap()` is exercised.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/QueryRemindersRequest.kt
gpunto: Definitely! However this needs a change in the conventions plugin + a release so I'm going to do it as a follow up. Tracked in AND-1362.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Ack — intentional and consistent with `queryChannels`/`queryUsers`, so no change.

PR #6601, opened 2026-07-29
Migrate searchMessages request body to the generated SearchPayload model

gpunto:
### Goal

Migrate the `searchMessages` request body to the generated `SearchPayload` model.

### Implementation

- `searchMessages` request -> generated `SearchPayload` (+ vendored `MessageOptions`, referenced by it); `sort` via `toSortParams()`. Mapper ported from the reference branch.
- `GeneralApi.searchMessages` `@UrlQueryPayload` param retyped to `SearchPayload`.
- Removed the hand-written `SearchMessagesRequest` DTO. Response side (`SearchMessagesResponse`) is unchanged.
- `SearchPayload` is a superset of the old DTO: it adds only optional `query`/`force_default_search`/`force_sql_v2_backend`/`message_options`, all default-null, so they are omitted from the wire — no shape change.

### Testing

- Updated the `MoshiChatApiTest` payload assertion to `SearchPayload`.
- Device probe (compose-sample) confirmed the live wire body: `{"filter_conditions":...,"message_filter_conditions":{"text":{"$autocomplete":"a"}},"sort":[{"direction":-1,"field":"created_at"}],"offset":0,"limit":5}` — the optional fields were omitted, backend returned 200 with results parsed into `SearchMessagesResult`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated message search request handling to use the correct payload format.
  * Preserved support for search filters, sorting, pagination, and message-specific options.
  * Improved compatibility with the messaging API while maintaining existing search behavior.

* **Tests**
  * Updated message search coverage to validate the revised request payload.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
@@ -1449,13 +1449,13 @@ constructor(
         next: String?,
         sort: QuerySorter<Message>?,
     ): Call<SearchMessagesResult> {
-        val newRequest = io.getstream.chat.android.client.api2.model.requests.SearchMessagesRequest(
-            filter_conditions = channelFilter.toMap(),
-            message_filter_conditions = messageFilter.toMap(),
+        val newRequest = io.getstream.chat.android.network.models.SearchPayload(
+            filterConditions = channelFilter.toMap(),
+            messageFilterConditions = messageFilter.toMap(),
             offset = offset,
             limit = limit,
             next = next,
-            sort = sort?.toDto(),
+            sort = sort?.toSortParams(),
         )
         return generalApi.searchMessages(newRequest)
             .mapDomain { response ->
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/SearchPayload.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/SearchPayload.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:Suppress(
+    "ArrayInDataClass",
+    "EnumEntryName",
+    "RemoveRedundantQualifierName",
+    "UnusedImport",
+)
+
+package io.getstream.chat.android.network.models
+
+import com.squareup.moshi.Json
+
+/**
+ *
+ */
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class SearchPayload(
+    @Json(name = "filter_conditions")
+    internal val filterConditions: Map<String, Any?> = emptyMap(),
+
+    @Json(name = "force_default_search")
+    internal val forceDefaultSearch: Boolean? = null,
+
+    @Json(name = "force_sql_v2_backend")
+    internal val forceSqlV2Backend: Boolean? = null,
+
+    @Json(name = "limit")
+    internal val limit: Int? = null,
+
+    @Json(name = "next")
+    internal val next: String? = null,
+
+    @Json(name = "offset")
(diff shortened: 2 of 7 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/SearchPayload.kt
```diff
@@ -0,0 +1,62 @@
...
+
+package io.getstream.chat.android.network.models
+
+import com.squareup.moshi.Json
+
+/**
+ *
+ */
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class SearchPayload(
```
andremion, line 32: The test asserts the payload object, so nothing locks the JSON shape. Worth adding a serialization test? #6564 did that with `QueryReactionsRequestAdapterTest` and `QueryReactionsRequestTestData`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1449,13 +1449,13 @@ constructor(
...
-            filter_conditions = channelFilter.toMap(),
-            message_filter_conditions = messageFilter.toMap(),
+        val newRequest = io.getstream.chat.android.network.models.SearchPayload(
+            filterConditions = channelFilter.toMap(),
+            messageFilterConditions = messageFilter.toMap(),
             offset = offset,
             limit = limit,
             next = next,
-            sort = sort?.toDto(),
+            sort = sort?.toSortParams(),
```
andremion, line 1458: Not blocking. `toSortParams()` keeps only `field` and `direction`, while the old code forwarded the raw map from `QuerySorter.toDto()`. Both are public, so a custom sorter returning extra keys loses them here. `queryChannels` and `queryUsers` already work this way, so consistency probably wins. Flagging it in case search was meant to be the loose one.

andremion (commented):
LGTM. Just two inline notes.

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/SearchPayload.kt
gpunto: Addressed in 5f3e419c8b — added `SearchPayloadAdapterTest` + `SearchPayloadTestData` (serialize all-fields and defaults), locking the JSON shape.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Intentional, and consistent with `queryReactions` (#6597), `queryChannels`, and `queryUsers`. The backend sort contract is `[{field, direction}]`; extra keys a custom sorter emits aren't consumed server-side, so normalizing drops nothing that worked. Keeping as-is.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/SearchPayload.kt
andremion: Thanks, that covers it. Tiny thing: the test data uses `type = "message"`, but the backend reads `type` as the field type, so the accepted values are `number`, `boolean`, or empty. Nothing validates it in a serialization test, it just may read as if `"message"` were valid.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
andremion: Fine to keep as-is. One correction on the detail: `type` is read by the backend, it picks number or boolean sort semantics, and the shorthand form `{"created_at": -1}` is accepted too. Neither was ever produced by `BaseQuerySort`, so nothing that worked before breaks, only a custom `toDto()` override would lose them. Possible follow-up for another PR: `toSortParams()` never fills `SortParamRequest.type` on any of the three endpoints.

andremion (approved):
LGTM. Left a small note on the test data in the other thread.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/SearchPayload.kt
gpunto: Good catch — dropped `type` from the fixture in 6b5dfc30b1. Built-in sorters never emit it, so `{field, direction}` is the realistic shape.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Thanks for the correction — noted that `type` and the `{field: dir}` shorthand are backend-read, just never emitted by `BaseQuerySort`, so nothing that worked breaks. Agreed on leaving it here; I'll track filling `SortParamRequest.type` in `toSortParams()` across the three endpoints as a separate follow-up.

PR #6602, opened 2026-07-29
Migrate queryPolls, queryPollVotes, and partial-update-poll request bodies to generated models

gpunto:
### Goal

Migrate the `queryPolls` and `queryPollVotes` request bodies (`POST /polls/query`, `POST /polls/{id}/votes`) to the generated `QueryPollsRequest` + `QueryPollVotesRequest` network models. First consumers of the merged `SortParamRequest`/`toSortParams` helper. Part of the incremental OpenAPI model migration.

### Implementation

- Add generated `internal QueryPollsRequest` + `QueryPollVotesRequest` in `network.models`; remove the hand-written `api2.model.requests` versions. `MoshiChatApi`/`PollsApi`/tests use the generated types by their real names (no alias).
- The generated `sort` field is `List<SortParamRequest>` (was `List<Map<String, Any>>`), so the two call sites switch from `sort?.toDto()` to `sort?.toSortParams()`. `filter?.toMap()` fits the generated `Map<String, Any?>` field. No wire-shape change: still `{"sort":[{"field","direction"}],"filter":{…},"limit":…}`.

### Testing

- `spotlessApply`, `apiCheck` (no API drift), `detekt`, `lint`, and the full client `testDebugUnitTest` suite pass (the `queryPolls`/`queryPollVotes` tests assert the exact request body).
- Verified on device against the real backend: `POST /polls/query` and `POST /polls/{id}/votes` each send `{"sort":[{"direction":-1,"field":"created_at"}]}` and return `201`, confirming the `SortParamRequest`-serialized `sort` is accepted on both endpoints.

_Also folds in the `partialUpdatePoll` request-body migration (`PATCH /polls/{id}` → generated `UpdatePollPartialRequest`) — bundled here to keep all poll-request migrations in one PR._

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved poll query request serialization for sorting and pagination.
  * Updated partial poll updates, including closing polls, to use the correct request format.
* **Tests**
  * Updated poll-related coverage to verify partial updates and query request payloads.


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
```diff
@@ -2751,7 +2751,7 @@ internal class MoshiChatApiTest {
             filter = filter.toMap(),
             limit = limit,
             next = next,
-            sort = sort.toDto(),
+            sort = sort.toSortParams(),
```
andremion: `QuerySortByField<Poll>()` has no fields added, so `toSortParams()` returns an empty list on both sides of the assertion, and `Filters.neutral()` maps to an empty map. The test passes regardless of how `SortParamRequest` is built.

Add a field, for example `QuerySortByField<Poll>().descByName("created_at")`, so the test pins the `field` and `direction` mapping? Same for `testQueryPollVotes` on line 2781.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1854,7 +1854,7 @@ constructor(
             filter = filter?.toMap(),
             limit = limit,
             next = next,
-            sort = sort?.toDto(),
+            sort = sort?.toSortParams(),
```
andremion, line 1857: `queryPolls` and `queryPollVotes` take any `QuerySorter`, including integrator implementations. `toDto()` went to the wire as is, `toSortParams()` keeps only `field` and `direction`:

- `{"field":"created_at","direction":-1,"type":"custom"}` becomes `{"direction":-1,"field":"created_at"}`
- `{"name":"created_at","dir":-1}` becomes `{}`

Already the case for `queryReactions`, `queryChannels`, and `queryUsers` since #6597. Also, `SortParamRequest` has a `type` field that `toSortParams()` never sets.

andremion (commented):
LGTM. Two non-blocking comments.

andremion (approved):

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
gpunto: Addressed in 63d4b4010e — both tests now use `descByName("created_at")` and assert an explicit `SortParamRequest(field = "created_at", direction = -1)` instead of `sort.toSortParams()`, so the mapping is pinned.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Intentional, and consistent with `queryReactions` (#6597), `queryChannels`, and `queryUsers`. The backend sort contract is `[{field, direction}]`; extra keys a custom sorter emits aren't consumed server-side, so normalizing drops nothing that worked. Keeping as-is.

andremion: Same point as the thread on #6601, answered there. Fine to keep as-is here too.

andremion (approved):

PR #6603, opened 2026-07-29
e2e: Make the remaining UserRobot clicks stale-safe

andremion:
### Goal

The last nightly run failed on API 35 because `GiphyTests.test_userObservesAnimatedGiphy_afterAddingGiphyThroughComposerMenu` failed all 3 retry attempts with `StaleObjectException`: https://github.com/GetStream/stream-chat-android/actions/runs/30414531872. The click landed on a composer button node that went stale between the find and the click.

#6588 introduced the stale-safe `waitToAppearAndClick`, but some `UserRobot` actions still click a found `UiObject2` directly, so any of them can fail the same way when Compose recomposes at tap time. This PR routes every remaining click through the stale-safe helper.

Resolves AND-1344

### Implementation

- `tapOnComposerConfirmButton` now taps one combined selector, `By.res(Pattern.compile("Stream_Composer(Send|Save)Button"))`, through `waitToAppearAndClick`, replacing the custom loop that polled the send and save buttons and clicked without stale protection. On timeout the error message now comes from `waitToAppear` instead of the custom text; the exception type is unchanged.
- The 10 `waitToAppear().click()` call sites convert to `waitToAppearAndClick()`: copy, pin, unpin, message reaction, extended reactions picker (2 taps), scroll-to-first-unread button and its dismiss icon, channel swipe "more", and view channel info.
- `tapOnMessage` is removed: it was the last unprotected click, and it has no callers.

### Testing

Ran locally against the mock server (`main`) on an API 35 emulator with direct instrumentation:

- `GiphyTests`: 10/10 passed in 106s, including the test that failed the nightly.
- `ReactionsTests`, `PinnedMessagesTests`, `ChannelActionsTests`, `MessageActionsTests` (the classes that exercise the converted click sites): 21/21 passed in 257s.

No retries were needed in either run. To reproduce, build `assembleE2eDebug` and `assembleE2eDebugAndroidTest` of `stream-chat-android-compose-sample`, start the mock server (`bundle exec fastlane start_mock_server`), install both APKs, and run the classes above with `adb shell am instrument`.

No UI changes.

## Summary by CodeRabbit

- **Tests**
  - Improved automated coverage for message composition, including both sending messages and saving commands.
  - Increased reliability of UI interactions for pinning, reactions, unread indicators, and channel information.
  - Standardized interaction handling to reduce flaky end-to-end test behavior in the Compose sample app.

### DIFF
--- a/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
+++ b/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
@@ -19,6 +19,7 @@ package io.getstream.chat.android.compose.pages
 import androidx.test.uiautomator.By
 import androidx.test.uiautomator.BySelector
 import io.getstream.chat.android.e2e.test.mockserver.ReactionType
+import java.util.regex.Pattern
 
 open class MessageListPage {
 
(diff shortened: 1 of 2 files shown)


andremion:
@coderabbit review

gpunto (approved):

PR #6606, opened 2026-07-31
Forward sort type from custom sorters in toSortParams

gpunto:
### Goal

`QuerySorter.toSortParams()` mapped only `field`/`direction` into the generated `SortParamRequest`, silently dropping the `type` the backend reads for number/boolean sort semantics. Forward it so a custom `QuerySorter` whose `toDto()` emits a `type` keeps it on the wire. Built-in sorters never emit `type`, so their output is unchanged.

Closes [AND-1354](https://linear.app/stream/issue/AND-1354/tosortparams-never-fills-sortparamrequesttype-on-sort-request-bodies)

### Implementation

- `toSortParams()` now also reads the `type` key from each `toDto()` map into `SortParamRequest.type` (null when absent).

### Testing

- New parameterized `ToSortParamsTest`: custom sorter with `type` forwards it, custom sorter without `type` leaves it null, built-in `descByName` maps to `field`/`direction` only.
- `spotlessCheck`, `detekt`, `apiDump` (no API change), and the full client `testDebugUnitTest` suite pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Custom sorting now correctly forwards an optional sort type when querying data.
* **Tests**
  * Added coverage for custom sorters with and without a type, as well as built-in field sorting.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
@@ -2045,5 +2045,9 @@ internal fun QuerySorter<*>.toSortParams(): List<SortParamRequest> =
         SortParamRequest(
             field = it[QuerySorter.KEY_FIELD_NAME] as? String,
             direction = (it[QuerySorter.KEY_DIRECTION] as? Number)?.toInt(),
+            // Built-in sorters don't emit a type; forward it when a custom sorter does.
+            type = it[KEY_SORT_TYPE] as? String,
         )
     }
+
+private const val KEY_SORT_TYPE = "type"
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
@@ -2046,8 +2046,6 @@ internal fun QuerySorter<*>.toSortParams(): List<SortParamRequest> =
             field = it[QuerySorter.KEY_FIELD_NAME] as? String,
             direction = (it[QuerySorter.KEY_DIRECTION] as? Number)?.toInt(),
             // Built-in sorters don't emit a type; forward it when a custom sorter does.
-            type = it[KEY_SORT_TYPE] as? String,
+            type = it[QuerySorter.KEY_TYPE] as? String,
         )
     }
-
-private const val KEY_SORT_TYPE = "type"
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/ToSortParamsTest.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/ToSortParamsTest.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.api2
+
+import io.getstream.chat.android.models.querysort.ComparableFieldProvider
+import io.getstream.chat.android.models.querysort.QuerySortByField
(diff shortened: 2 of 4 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -2045,5 +2045,9 @@ internal fun QuerySorter<*>.toSortParams(): List<SortParamRequest> =
         SortParamRequest(
             field = it[QuerySorter.KEY_FIELD_NAME] as? String,
             direction = (it[QuerySorter.KEY_DIRECTION] as? Number)?.toInt(),
+            // Built-in sorters don't emit a type; forward it when a custom sorter does.
+            type = it[KEY_SORT_TYPE] as? String,
         )
     }
+
+private const val KEY_SORT_TYPE = "type"
```
aleksandar-apostolov: `KEY_FIELD_NAME`/`KEY_DIRECTION` are `public const val` on `QuerySorter` so custom sorters can reference them in `toDto()` — here `"type"` is a private local const. Intentional (keeping it off the public `.api`, or cross-platform parity), or should it live on `QuerySorter` too so custom sorters have a canonical key to emit rather than hardcoding `"type"`?

aleksandar-apostolov (approved):
LGTM

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -2045,5 +2045,9 @@ internal fun QuerySorter<*>.toSortParams(): List<SortParamRequest> =
         SortParamRequest(
             field = it[QuerySorter.KEY_FIELD_NAME] as? String,
             direction = (it[QuerySorter.KEY_DIRECTION] as? Number)?.toInt(),
+            // Built-in sorters don't emit a type; forward it when a custom sorter does.
+            type = it[KEY_SORT_TYPE] as? String,
```
andremion: Today the only way to send a `type` is extending `BaseQuerySort` and overriding `toDto()`, since `SortSpecification` has no type field. The map to sorter direction drops it too (`QuerySortConverter`, `QuerySortParser.fromRawInfo`, `DomainMapping.toSortDomain` all read `field`/`direction` only), though those feed offline state and not the request body, so nothing on the wire changes.

Is a follow-up planned to make the type part of the sorter model, for example an `asc(field, type)` overload? That is what someone sorting numerically on a custom field would reach for.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/ToSortParamsTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        Assertions.assertEquals(expected, sorter.toSortParams())
+    }
+
+    companion object {
+
+        @JvmStatic
+        fun arguments(): List<Arguments> = listOf(
+            // Custom sorter emitting a type: field, direction and type are all forwarded.
+            Arguments.of(
+                FakeSorter(mapOf("field" to "created_at", "direction" to -1, "type" to "number")),
```
andremion: Nit: the fake hardcodes `"field"` and `"direction"` while the production code reads them through `QuerySorter.KEY_*`. Deliberate, to pin the wire keys?

andremion (approved):
LGTM. One scope question and one nit below.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Good call, made it consistent: promoted it to a public `QuerySorter.KEY_TYPE` alongside `KEY_FIELD_NAME`/`KEY_DIRECTION`, so custom sorters have a canonical key to emit instead of hardcoding `"type"`. `toSortParams()` now reads `QuerySorter.KEY_TYPE`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Right, this PR only carries a `type` when a sorter already emits one (which today still needs a `BaseQuerySort`/`toDto()` override). Making `type` a first-class part of the sorter model with an `asc(field, type)` overload is a bigger, separate change, especially the offline side (`QuerySortConverter`/`QuerySortParser`/`toSortDomain` read only field/direction, so the local comparator would need to honor `type` for online/offline parity). Filed as AND-1361.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/ToSortParamsTest.kt
gpunto: Fixed, the fake now uses `QuerySorter.KEY_FIELD_NAME`/`KEY_DIRECTION`/`KEY_TYPE` too.

PR #6609, opened 2026-07-31
Support local unread counts for channels with read events disabled

gpunto:
### Goal

Closes [AND-1358](https://linear.app/stream/issue/AND-1358/support-local-client-side-unread-counts-for-channels-with-read-events)

Provide an unread mechanism for channels where server-side read events are disabled (e.g. livestreams). Android port of [stream-chat-swift#4102](https://github.com/GetStream/stream-chat-swift/pull/4102).

Adds an opt-in `ChatClientConfig.isLocalUnreadCountEnabled` flag (default `false`). When on, channels with `readEventsEnabled = false` get a per-channel unread count tracked on-device. The user-level total unread count is unaffected.

### Scope: best-effort, live-events only

By design this is a **best-effort** count of live events, not an authoritative unread count. These channels turn off server-side read tracking, so there is no read state to reconcile against — which means:

- it counts messages that arrive **while the app is running**; it does not backfill messages received while the app was closed,
- it is **per-device** (two devices can disagree), and
- it **resets on a fresh install** (no prior local state to derive from).

This matches the intended behavior of the original feature, confirmed with the iOS team (author of stream-chat-swift#4102). Making it accurate across restarts/offline would require server-tracked reads, which is exactly what `read_events = false` disables.

### Implementation

- **Increment**: for a locally tracked channel, an incoming live message creates the current user's read on the fly (the server never sends one) and increments it. Skip rules (own/muted/silent/shadowed/thread-reply/already-read) apply, plus a skip for messages already in the channel state so a sync-replayed event isn't counted twice.
- **Mark read**: `markRead` on a locally tracked channel resets the count in state without a network request (`markRead()` returns a new `MarkReadResult` describing whether a remote request is still needed) and refreshes the channel-list rows. It advances `lastRead` just past the last message, mirroring the server's mark-read, so the message list doesn't render a spurious unread separator for the user's own next message.
- **Server-data guard**: the in-memory read merge and the DB-layer merge both keep the locally tracked read; server payloads only contribute user info and delivered-receipt fields. `updateLastMessage` also leaves the local read untouched, so the on-device count is the single writer (it would otherwise increment the persisted read with weaker guards).
- **Persistence**: local reads are written to Room through a new internal `ChannelRepository.upsertChannelReads` (default no-op) that bypasses the server-data merge, so the last-known count survives an app restart.
- Both the new and legacy channel-logic paths are covered.

### Demo app

The compose sample gets a "Local unread count" toggle (advanced login options). When enabled it also includes livestream channels in the channel list so the feature is testable.

### Testing

- `spotlessCheck`, `detekt`, `apiCheck` (only additive: the new `ChatClientConfig`/`StateRegistry` trailing flag; source-compatible), and the full client `testDebugUnitTest` suite pass.
- Unit tests cover the increment, local mark-read, the server-data guard (including tying-event-date and cold-start cases), and the sync-replay skip, on both channel-logic paths.

Manual testing:

1. Enable **Local unread count** in the login advanced options and log in as a user who's a member of the livestream channel, e.g. Leia, Padmé and any other member of "Local Unread Test".
2. With the app foregrounded and that channel closed, send a few messages from another user.
3. The channel row's unread badge climbs by the number of messages received.
4. Open the channel: the badge clears with no `/read` request.

Best-effort by design (see Scope): messages arriving while the app is backgrounded/killed, offline, or before a fresh install are not counted.

## Summary by CodeRabbit

* **New Features**
  * Added an optional local unread-count setting for channels without server read events; it is disabled by default.
  * Local unread counts persist across offline updates and remain consistent with channel state.
  * Added sample-app controls and localized descriptions for enabling the feature.

* **Bug Fixes**
  * Improved read-state merging, duplicate message handling, and channel-list refresh behavior.
  * Read actions now correctly distinguish remote, local, and unnecessary operations.


gpunto:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/listener/internal/ChannelMarkReadListenerState.kt
```diff
@@ -41,11 +42,16 @@ internal class ChannelMarkReadListenerState(private val logic: LogicRegistry) :
...
+            MarkReadResult.RemoteRequired -> Result.Success(Unit)
+            MarkReadResult.HandledLocally -> {
+                // No server read event follows a local mark-read; refresh the channel-list queries directly.
+                val cid = "$channelType:$channelId"
+                logic.getActiveQueryChannelsLogic().forEach { it.refreshChannelState(cid) }
+                Result.Failure(Error.GenericError("Channel $channelId marked as read locally"))
+            }
+            MarkReadResult.NotNeeded ->
+                Result.Failure(Error.GenericError("Can not mark channel as read with channel id: $channelId"))
         }
```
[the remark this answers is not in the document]
gpunto: Keeping `Result.Failure` here on purpose. `Call.precondition` (ChatClient.kt) short-circuits on the first `Result.Failure` and skips the underlying `api.markRead` request; `Result.Success` lets it proceed. For a channel with read events disabled that `/read` request is rejected by the server with `400` ("does not support events of type `message.read`"), so `Failure` is required to suppress it - which is also why the pre-existing `NotNeeded` branch returns `Failure`. Returning `Success` would reintroduce that failed request. The channel-list refresh already runs before we return.

gpunto:
@coderabbitai review

gpunto:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/plugin/logic/querychannels/internal/QueryChannelsStateLogic.kt
```diff
@@ -187,6 +187,12 @@ internal class QueryChannelsStateLogic(
                 )
             }
         }.forEach { it.await() }
+        // For read-events-disabled channels the locally tracked read is authoritative over the raw
+        // payload set above, so re-derive those channels from the merged per-channel state.
+        validated
+            .mapNotNull { (id, channel) -> id.cid.takeUnless { channel.config.readEventsEnabled } }
+            .takeIf { it.isNotEmpty() }
+            ?.let(::refreshChannels)
```
aleksandar-apostolov: This reconciliation isn't gated on `isLocalUnreadCountEnabled` — it re-derives every `read_events = false` channel from per-channel state for all apps, feature on or off.

It looks harmless for non-adopters (`refreshChannels` only touches active channels, and with the flag off the state read equals the server read), but it's a behavior change with a wider blast radius than the opt-in flag implies. Is the flag-independence intentional? Gating it behind `isLocalUnreadCountEnabled` would keep the feature's footprint at zero for apps that query livestream channels without opting in.

gpunto: Good catch, gated!

aleksandar-apostolov (approved):

PR #6613, opened 2026-08-03
Migrate user request bodies (partialUpdateUser, getGuestUser, updateUsers) to generated models

gpunto:
### Goal

Migrate the `partialUpdateUser`, `getGuestUser`, and `updateUsers` request bodies to the generated `UpdateUsersPartialRequest`/`UpdateUserPartialRequest`, `CreateGuestRequest`, `UpdateUsersRequest`, and `UserRequest` network models.

### Implementation

- `partialUpdateUser` -> `UpdateUsersPartialRequest(UpdateUserPartialRequest(id/set/unset))`.
- `getGuestUser` -> `CreateGuestRequest(UserRequest(id, name))`; the generated `UserRequest` flattens `custom` to the wire root via a new `UserRequestAdapter`.
- `updateUsers` -> `UpdateUsersRequest(Map<String, UserRequest>)` via a new `User.toUserRequest()` mapper.
- Removed hand-written `PartialUpdateUsersRequest`, `GuestUserRequest`, `PartialUpdateUserDto`, and `UpdateUsersRequest`.

**Note on `updateUsers`:** the generated `UserRequest` is leaner than the old `UpstreamUserDto` — it omits `role`/`teams`/`teams_role`/`banned`/`devices`. That's intentional and safe: the backend marks `role`/`teams`/`teams_role` as `ignore_if_client_side` (stripped for client-token requests) and doesn't bind `banned`/`devices` on this endpoint, so the old DTO was shipping fields the server already discarded. Verified on the wire — an `updateUsers` upsert returns `200` and `role`/`teams` are preserved server-side even though the client no longer sends them.

### Testing

- New `UserRequestsAdapterTest` + testdata lock the wire shape (incl. `UserRequest` custom flatten and `privacy_settings`).
- `MoshiChatApiTest` mapper assertions updated.
- Device-probed against the live backend: `partialUpdateUser` (set + unset paths) and `updateUsers` (custom sentinel round-trips; `role`/`teams` preserved) confirmed.

## Summary by CodeRabbit

- **Improvements**
  - Improved guest-user creation and user update requests for more consistent handling of profiles, custom data, and privacy settings.
  - Added support for partial and bulk user updates, including setting and removing user fields.
  - Improved request processing for typing indicators, read receipts, and delivery receipts.
  - Custom user data is now correctly included in outgoing requests.

- **Reliability**
  - Added coverage for guest creation, partial updates, bulk updates, custom data, and privacy settings.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/PrivacySettingsResponse.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/PrivacySettingsResponse.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:Suppress(
+    "ArrayInDataClass",
+    "EnumEntryName",
+    "RemoveRedundantQualifierName",
+    "UnusedImport",
+)
+
+package io.getstream.chat.android.network.models
+
+import com.squareup.moshi.Json
+
+/**
+ *
+ */
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class PrivacySettingsResponse(
+    @Json(name = "delivery_receipts")
+    internal val deliveryReceipts: io.getstream.chat.android.network.models.DeliveryReceiptsResponse? = null,
+
+    @Json(name = "read_receipts")
+    internal val readReceipts: io.getstream.chat.android.network.models.ReadReceiptsResponse? = null,
+
+    @Json(name = "typing_indicators")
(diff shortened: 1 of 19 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/PrivacySettingsResponse.kt
```diff
@@ -0,0 +1,41 @@
...
+
+package io.getstream.chat.android.network.models
+
+import com.squareup.moshi.Json
+
+/**
+ *
+ */
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class PrivacySettingsResponse(
```
VelikovPetar, line 32: Should this really be `-Response`? - isn't it used as part of a `request`? (I know that this is auto-generated, but I still wanted to double check it)

VelikovPetar (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/PrivacySettingsResponse.kt
gpunto: Yeah, it's odd but the backend reuses the same model, if they ever split it we will pick it at the following regen. We could propose them to do the split, wdyt?

The good part, anyway, it's that in Chat these models are internal, so they can be removed/renamed without breaking changes.

VelikovPetar: I am fine keeping it like this, not sure if it is worth it, as it will most likely affect other SDKs as well (for whom it might be a breaking change).

gpunto: Sounds good, let's keep it 👍 And if it ever changes, we're going to get it anyway

VelikovPetar (approved):

PR #6614, opened 2026-08-04
Migrate slow mode cooldown request to the generated UpdateChannelPartialRequest model

gpunto:
### Goal

Migrate the slow-mode cooldown request body (`enableSlowMode`/`disableSlowMode`, via the private `updateCooldown`) to the generated `UpdateChannelPartialRequest` network model.

### Implementation

- `updateCooldown` now builds `UpdateChannelPartialRequest(set = mapOf("cooldown" to N))`; `ChannelApi.updateCooldown`'s `@Body` type is retyped accordingly.
- Remove the hand-written `UpdateCooldownRequest` (was `internal`, so no public-API change).

### Testing

- New `UpdateChannelPartialRequestAdapterTest` locks the wire shape (`{"unset":[],"set":{"cooldown":N}}`).
- Device-probed enable/disable slow mode on the wire: request body byte-exact and backend-accepted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved slow-mode updates by sending cooldown settings through the standard partial channel update flow.
  * Preserved reliable enabling and disabling of channel slow mode.

* **Tests**
  * Added coverage for serializing partial channel updates, including populated, default, set, and unset fields.
  * Updated slow-mode request tests to reflect the unified update format.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/endpoint/ChannelApi.kt
```diff
@@ -111,7 +110,7 @@ internal interface ChannelApi {
     fun updateCooldown(
         @Path("type") channelType: String,
         @Path("id") channelId: String,
-        @Body body: UpdateCooldownRequest,
+        @Body body: UpdateChannelPartialRequest,
```
VelikovPetar: Thinking if we should drop the `updateCooldown` method -> and use the `updateChannelPartial` directly instead? 
(just a suggestion, if you think we benefit from having the 2 api endpoints, let's keep it as it is)

gpunto: Good point, let me actually double check against the generated API class since that's our end goal anyway 👀

gpunto: Confirmed, no trace of a specific function in the generated API and indeed it would have no benefit, so I collapsed it to just one function!

VelikovPetar: Awesome - I think it is better this way, one less method to maintain 👍

VelikovPetar (approved):

PR #6621, opened 2026-08-06
client: Keep persisted channel lastMessageAt consistent with its latest message

### LINKED ISSUE #6604: Channel list uses stale lastMessageAt after loading a newer message
## Describe the bug

While comparing the Android and iOS channel lists, I noticed that Android can show a newer message in a channel row without updating the channel's local `lastMessageAt`.

This leads to an inconsistent list: the row shows the latest message and its timestamp, but its position is still based on the older `lastMessageAt`.

The iOS SDK avoids this by updating the local value when it saves a newer message:

```swift
channelDTO.lastMessageAt = max(
    channelDTO.lastMessageAt?.bridgeDate ?? payload.createdAt,
    payload.createdAt
).bridgeDate
```

See [`MessageDTO.swift`, lines 1048–1052](https://github.com/GetStream/stream-chat-swift/blob/1ee90970846764d27cde8210d62c8b85a7035571/Sources/StreamChat/Database/DTOs/MessageDTO.swift#L1048-L1052).

Android appears to update the latest messages in the channel state, but leaves `Channel.lastMessageAt` unchanged.

## SDK version

- `stream-chat-android-client:7.6.0`
- `stream-chat-android-compose:7.6.0`

## To Reproduce

1. Query channels sorted by `last_message_at`:

   ```kotlin
   val sort = QuerySortByField
       .descByName<Channel>("pinned_at")
       .desc("last_message_at")
   ```

2. Load a channel whose `lastMessageAt` is older than its latest message's
   `createdAt`.
3. Display the result using the Compose channel list.
4. The row shows the newer message, but the channel remains sorted using the
   older `lastMessageAt`.

## Expected behavior

I would expect Android to update the local `Channel.lastMessageAt` when it
processes a newer message, as the iOS SDK does.

Is the difference between the two SDKs intentional? If so, what is the
recommended way to keep channel ordering consistent across Android and iOS?

## Device

- Vendor and model: Android Emulator
- Android version: Not device-specific

gpunto:
### Goal

Channels could be sorted by a stale `lastMessageAt` when the persisted value lagged the channel's newest visible message, so a row would show a newer message than its position in the list reflected.

Closes #6604

### Implementation

When persisting a channel, keep `lastMessageAt` consistent with its last message instead of trusting the stored field, so it never lags behind (nor regresses below) the newest visible message. This matches the iOS SDK's behavior.

### Testing

- `ChannelMapperTest`: added coverage for advancing to a newer message, not regressing when the stored value is newer, and the no-last-message case; updated the existing mapping test.
- On device: confirmed the SDK heals real channels on launch and that `stream_chat_channel_state.lastMessageAt` matches the channel's last message on this build versus staying stale on `develop`.
- `spotlessCheck`, `detekt`, `apiCheck` (no public API change), and the client `testDebugUnitTest` suite pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved channel activity tracking by recording the most recent timestamp between the channel update and its latest message.
  - Preserved the channel timestamp when it is newer than the last message.
  - Correctly handles channels without a last message.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/channel/internal/ChannelMapper.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/channel/internal/ChannelMapper.kt
@@ -16,6 +16,7 @@
 
 package io.getstream.chat.android.client.internal.offline.repository.domain.channel.internal
 
+import io.getstream.chat.android.client.extensions.getCreatedAtOrNull
 import io.getstream.chat.android.client.extensions.internal.lastMessage
 import io.getstream.chat.android.client.extensions.syncUnreadCountWithReads
 import io.getstream.chat.android.client.internal.offline.repository.domain.channel.member.internal.MemberEntity
@@ -26,6 +27,7 @@ import io.getstream.chat.android.client.internal.offline.repository.domain.chann
 import io.getstream.chat.android.client.internal.offline.repository.domain.channel.userread.internal.toModel
 import io.getstream.chat.android.client.internal.offline.repository.domain.message.internal.toEntity
 import io.getstream.chat.android.client.internal.offline.repository.domain.message.internal.toModel
+import io.getstream.chat.android.core.utils.date.max
 import io.getstream.chat.android.models.Channel
 import io.getstream.chat.android.models.ChannelUserRead
 import io.getstream.chat.android.models.DraftMessage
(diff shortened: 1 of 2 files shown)


aleksandar-apostolov (approved):
LGTM — does this also need a backport to v6?

gpunto:
> LGTM — does this also need a backport to v6?

Good point, backporting it in https://github.com/GetStream/stream-chat-android/pull/6623

PR #6623, opened 2026-08-07
offline: Keep persisted channel lastMessageAt consistent with its latest message

gpunto:
### Goal

Backport of #6621 to `v6`. Channels could be sorted by a stale `lastMessageAt` when the persisted value lagged the channel's newest visible message, so a row would show a newer message than its position in the list reflected.

Original issue: #6604

### Implementation

When persisting a channel, keep `lastMessageAt` consistent with its last message instead of trusting the stored field, so it never lags behind (nor regresses below) the newest visible message. Same change as #6621, applied to `stream-chat-android-offline` where the mapper lives on this branch.

### Testing

- `ChannelMapperTest`: added coverage for advancing to a newer message, not regressing when the stored value is newer, and the no-last-message case; updated the existing mapping test.
- `spotlessCheck`, `detekt`, `apiCheck` (no public API change), and the offline `testDebugUnitTest` suite pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved channel message timestamps to retain the most recent available time.
  * Corrected timestamp handling when the latest message is newer than the channel timestamp.
  * Added coverage for older, newer, and missing message timestamps.


aleksandar-apostolov (approved):
LGTM

PR #6626, opened 2026-08-10
Expose the channel truncatedAt, disabled and blocked state as properties

gpunto:
### Goal

Domain `Channel` had no property for the `truncated_at`, `disabled` and `blocked` wire fields, so callers could only reach them by digging into `Channel.extraData`, and only by accident. Expose them as real properties, matching iOS (`truncatedAt`/`isDisabled`/`isBlocked` on `ChatChannel`), and map them end to end. The keys stay readable from `extraData` too, so this is not a behavioural change.

This also unblocks the OpenAPI migration: the generated `ChannelResponse` already declares these fields, so adopting it today would silently drop them.

Closes [AND-1375](https://linear.app/stream/issue/AND-1375/expose-the-channel-truncatedat-disabled-and-blocked-state-as)

### Implementation

- Add the three properties to `Channel`, with defaults, plus `Builder` and `getComparableField` support.
- Declare them on `DownstreamChannelDto` and map them in `toDomain()`. `blocked` is nullable because some endpoints and all events omit it.
- Carry them through `ChannelData` so the state layer keeps them, and merge them from events like iOS: always take `disabled`, keep `blocked` when omitted, keep the later `truncatedAt`.
- Persist as `ChannelEntity` columns and bump the DB version (no migration, per `fallbackToDestructiveMigration`).
- Add an opt-in keep-list to `CustomObjectDtoAdapter` so the three keys are still copied into `extraData` with their raw wire values, keeping the old reads working. Removal is tracked on AND-1375 for the next major, and `ChannelResponseAdapter` inherits it after the OpenAPI migration.

### Testing

- Unit: mapping, Room round trip, sort fields and merge rules, plus a parser test asserting the keys stay in `extraData` with their raw wire types.
- On device against the live API: `queryChannels`, `queryChannel`, `createChannel`, partial update, grouped channels and five event types. Forced `blocked=true` and a non-null `truncatedAt` server-side to confirm both.
- Also on device: `extraData["truncated_at"]` keeps the raw String and its sub-millisecond precision (`2021-11-10T10:30:35.99922Z`), which a reformat from the parsed `Date` would have truncated, and stays absent on channels that were never truncated.

## Summary by CodeRabbit

* **New Features**
  * Added channel state information for truncation date, disabled status, and blocked status.
  * Exposed these properties through channel data, builders, and conversion APIs.
  * Preserved channel state consistently during event updates and offline storage.

* **Bug Fixes**
  * Improved channel updates to retain the latest truncation date and existing blocked status when updates omit values.

* **Tests**
  * Added coverage for parsing, mapping, persistence, event merging, and state conversion.


stream-chat-android-core/src/main/java/io/getstream/chat/android/models/Channel.kt
```diff
@@ -108,6 +113,9 @@ public data class Channel(
     val pushPreference: PushPreference? = null,
     val filterTags: List<String> = emptyList(),
     val lastMessageAt: Date? = null,
+    val truncatedAt: Date? = null,
```
VelikovPetar, line 116: One very small concern here: Could this be considered a behavioural breaking change? 
Assume the case where a customer used to read `channel.extraData["disabled"]` - after updating, they will no longer be able to read that field properly. 

TBH it is entirely possible that we have unknowingly shipped something similar to this in the past, so _most likely_ this is fine - I just wanted to bring this up just in case.

gpunto: Good catch, these values would stop being accessible from `extraData`. An alternative we have is to copy them also in `extraData` and then remove this logic in the major. I think we should be able to do that by extending what `CustomObjectDtoAdapter` does, trying that 👀

gpunto: Done in https://github.com/GetStream/stream-chat-android/pull/6626/changes/cbebd2bfc8fbbbaadfe1a4f0a2c842655c87d1db, it was simpler than I expected!

And the mechanism can be reused in case any other similar case appears during the OpenAPI migration.

VelikovPetar: Looks pretty good - for an ugly workaround 😅 

Thanks for the update - approving the PR!

VelikovPetar (approved):

VelikovPetar (approved):

PR #6627, opened 2026-08-10
Migrate poll option responses to the generated PollOptionResponse model

gpunto:
### Goal

Migrate the poll option responses from the hand-written `PollOptionResponse` to the generated network models.

### Implementation

- Add generated `PollOptionResponse` and `PollOptionResponseData`; remove the hand-written `PollOptionResponse`.
- Add `PollOptionResponseDataAdapter` so the custom data the v1 endpoints flatten to the root is collected into `custom`.
- Retype the option endpoints in `PollsApi` and map `PollOptionResponseData` to the domain `Option` in `DomainMapping`.

### Testing

- `PollOptionResponseParsingTest` covers the option and its collected custom fields. Removing the adapter registration makes the custom assertions fail.
- Device-probed the option endpoints on the wire, with custom data round-tripping through create and update (`sentiment`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved poll option response handling, including duration, identifiers, text, and custom fields.
  * Ensured poll options are correctly converted into the app’s domain model.
  * Added support for responses with or without custom fields.

* **Tests**
  * Added coverage for poll option parsing and custom-field handling.


stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/PollOptionResponse.kt
```diff
@@ -14,19 +14,25 @@
...
- * Response for creating/updating a poll option.
  *
- * @param duration The duration of the request.
- * @param poll_option The poll option object.
  */
-@JsonClass(generateAdapter = true)
+@com.squareup.moshi.JsonClass(generateAdapter = true)
 internal data class PollOptionResponse(
-    val duration: String,
-    val poll_option: DownstreamPollOptionDto,
```
VelikovPetar, line 31: Is the `DownstreamPollOptionDto` not delete-able yet?

VelikovPetar (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/PollOptionResponse.kt
gpunto: Yeah, not yet. It's still referenced by `DownstreamPollDto.options` (which is mostly used in events but also in some other models).

PR #6631, opened 2026-08-11
Resize image attachments on by default with a 2MP cap

aleksandar-apostolov:
### Goal
Fixes [AND-1384](https://linear.app/stream/issue/AND-1384/image-attachments-not-resized-by-default-2x-cdn-bandwidth-vs-ios) — image attachments were downloaded at full resolution because CDN resizing was off by default, costing ~2x the bytes of iOS (which caps attachments at 2MP by default, always on).

### Implementation
Introduces a `StreamCdnImageResizer` abstraction (`stream-chat-android-ui-common`) and turns image resizing on by default with a 2MP cap, matching iOS:
- `StreamCdnMaxPixelsImageResizer` (default, 2MP) and `NoOpStreamCdnImageResizer` (explicit opt-out) implement the public `StreamCdnImageResizer` interface; a new `createResizedStreamCdnImageUrl(maxImagePixels)` client overload caps by total pixels (preserves aspect ratio, never upscales, Stream-CDN URLs only).
- Wired into both kits via `ChatUI.streamCdnImageResizer` and `ChatTheme(streamCdnImageResizer = …)`, defaulting to the 2MP resizer.
- Precedence keeps existing integrations intact: when the legacy `StreamCdnImageResizing` is explicitly enabled its exact behavior is honored; otherwise the resizer applies. Disabling now requires `NoOpStreamCdnImageResizer`.
- Deprecates `StreamCdnImageResizing`. There is no built-in percentage successor — integrators who need percentages implement their own `StreamCdnImageResizer`.
- Behavior change: the fullscreen gallery now applies the 2MP cap (previously served full-resolution originals).

### Testing
Added `StreamCdnImageResizerTest` (resolver precedence, 2MP cap, no-op) and client `StringExtensionsKtTest` cases (over/under budget). Run:
`./gradlew :stream-chat-android-client:testDebugUnitTest --tests "*extensions.StringExtensionsKtTest" :stream-chat-android-ui-common:testDebugUnitTest --tests "*images.resizing.StreamCdnImageResizerTest"`

## Summary by CodeRabbit

- **New Features**
  - Added configurable Stream CDN image resizing with a default 2-megapixel limit.
  - Supports preserving aspect ratio, optional resize and crop modes, and disabling resizing when needed.
  - Applied resizing consistently to avatars, galleries, attachments, thumbnails, and quoted media.
  - Added a convenient maximum-pixel image URL option that avoids upscaling.

- **Deprecations**
  - Legacy image-resizing configuration remains available but is deprecated in favor of the new resizer settings.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
@@ -25,10 +25,15 @@ import io.getstream.chat.android.models.streamcdn.image.StreamCdnCropImageMode
 import io.getstream.chat.android.models.streamcdn.image.StreamCdnOriginalImageDimensions
 import io.getstream.chat.android.models.streamcdn.image.StreamCdnResizeImageMode
 import io.getstream.log.StreamLog
+import kotlin.math.sqrt
 
 private val snakeRegex = "_[a-zA-Z]".toRegex()
 private val camelRegex = "(?<=[a-zA-Z])[A-Z]".toRegex()
 private val baseUrlRegex = "^(?:https?://)?(.*?)/*$".toRegex()
+
+// Mirrors AttachmentHelper's STREAM_CDN_HOST_PATTERN — resize only Stream-hosted images.
+private val streamCdnHostRegex =
+    "stream-chat-+.+\\.imgix.net$|.+\\.stream-io-cdn.com$".toRegex(RegexOption.IGNORE_CASE)
 
 /**
  * Converts string written in snake case to String in camel case with the first symbol in lower case.
@@ -84,6 +89,10 @@ public fun String.cidToTypeAndId(): Pair<String, String> {
 public fun String.getStreamCdnHostedImageDimensions(): StreamCdnOriginalImageDimensions? {
     return try {
         val imageUri = this.toUri()
+
+        if (imageUri.host?.let(streamCdnHostRegex::matches) != true) {
+            return null
+        }
 
         val width = imageUri.getQueryParameter("ow")
             ?.toInt()
@@ -174,6 +183,30 @@ public fun String.createResizedStreamCdnImageUrl(
         }
         this
     }
+}
+
+/**
+ * Generates a URL with Stream CDN resizing parameters that cap the image to [maxImagePixels] total
+ * pixels (width × height), preserving aspect ratio and never upscaling. Mirrors the iOS SDK's
+ * `imageAttachmentMaxPixels`. Only affects Stream CDN hosted images that carry original width (ow)
+ * and height (oh) params; any other URL, or one already at/under the budget, is returned unchanged.
+ */
+public fun String.createResizedStreamCdnImageUrl(
+    maxImagePixels: Long,
+    resizeMode: StreamCdnResizeImageMode? = null,
+    cropMode: StreamCdnCropImageMode? = null,
+): String {
+    if (maxImagePixels <= 0L) return this
+    val dimensions = getStreamCdnHostedImageDimensions() ?: return this
+    val totalPixels = dimensions.originalWidth.toLong() * dimensions.originalHeight.toLong()
+    if (totalPixels <= 0L || totalPixels <= maxImagePixels) return this
+    val scale = sqrt(maxImagePixels.toDouble() / totalPixels.toDouble()).toFloat()
+    return createResizedStreamCdnImageUrl(
+        resizedWidthPercentage = scale,
+        resizedHeightPercentage = scale,
+        resizeMode = resizeMode,
+        cropMode = cropMode,
+    )
 }
 
 /**
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
@@ -31,9 +31,31 @@ private val snakeRegex = "_[a-zA-Z]".toRegex()
 private val camelRegex = "(?<=[a-zA-Z])[A-Z]".toRegex()
 private val baseUrlRegex = "^(?:https?://)?(.*?)/*$".toRegex()
 
-// Mirrors AttachmentHelper's STREAM_CDN_HOST_PATTERN — resize only Stream-hosted images.
-private val streamCdnHostRegex =
-    "stream-chat-+.+\\.imgix.net$|.+\\.stream-io-cdn.com$".toRegex(RegexOption.IGNORE_CASE)
+/**
+ * Matches the default Stream CDN hosts (imgix and stream-io-cdn). Shared with [AttachmentHelper] so the two
+ * can't drift. Anchored on the host only — resizing of custom/proxied Stream CDN domains is opted into via the
+ * `cdnHost` parameter of [createResizedStreamCdnImageUrl], mirroring iOS' `StreamCDNRequester(cdnHost:)`.
+ */
+@InternalStreamChatApi
+public val STREAM_CDN_HOST_PATTERN: Regex =
+    "stream-chat-+.+\\.imgix\\.net$|.+\\.stream-io-cdn\\.com$".toRegex(RegexOption.IGNORE_CASE)
+
+/**
+ * Returns whether this URL is served from a Stream CDN host that understands resizing query parameters.
+ *
+ * @param cdnHost An optional custom Stream CDN host (e.g. a proxied domain). When provided, the URL host is
+ * matched against it with a substring check, mirroring iOS' `StreamCDNRequester(cdnHost:)`; when null, the
+ * default [STREAM_CDN_HOST_PATTERN] is used.
+ */
+@InternalStreamChatApi
+public fun String.isStreamCdnHosted(cdnHost: String? = null): Boolean {
+    val host = this.toUri().host ?: return false
+    return if (cdnHost != null) {
+        host.contains(cdnHost, ignoreCase = true)
+    } else {
+        STREAM_CDN_HOST_PATTERN.matches(host)
+    }
+}
 
 /**
  * Converts string written in snake case to String in camel case with the first symbol in lower case.
@@ -89,10 +111,6 @@ public fun String.cidToTypeAndId(): Pair<String, String> {
 public fun String.getStreamCdnHostedImageDimensions(): StreamCdnOriginalImageDimensions? {
     return try {
         val imageUri = this.toUri()
-
-        if (imageUri.host?.let(streamCdnHostRegex::matches) != true) {
-            return null
-        }
 
         val width = imageUri.getQueryParameter("ow")
             ?.toInt()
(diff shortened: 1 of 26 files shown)


stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
```diff
@@ -103,7 +105,13 @@ internal fun MediaGalleryImagePage(
         modifier = Modifier.fillMaxSize(),
         contentAlignment = Alignment.Center,
     ) {
-        val data = attachment.imageUrl
+        // Custom StreamCdnImageResizer instances don't propagate to the standalone gallery Activity
+        // (interfaces aren't serializable across the Intent boundary), so the legacy resizing is honored from the
+        // Intent when enabled and otherwise the default 2MP StreamCdnMaxPixelsImageResizer is applied inline.
+        val data = attachment.imageUrl?.applyStreamCdnImageResizing(
+            ChatTheme.streamCdnImageResizing,
+            StreamCdnMaxPixelsImageResizer(),
```
gpunto: Hardcoding the resizer means `NoOpStreamCdnImageResizer` does nothing here, contradicting the `@param streamCdnImageResizer` KDoc in `ChatTheme.kt`. Capping in fullscreen is the right call, UIKit does it too, but it reads the configurable `components.imageAttachmentMaxPixels` there (`ImageAttachmentGalleryCell.swift:38`). It caps and still lets you change the number; this is the one place ours can't be changed.

`maxImagePixels` is a `Long`, so the Intent boundary doesn't force a hardcode. `MediaGalleryInjector` already exists for exactly this problem, carrying the `ImageLoader` into this Activity.

Parity: the XML kit doesn't resize in its fullscreen viewer at all (`AttachmentGalleryImagePageFragment` loads `imageUrl` raw), only in the grid (`MediaAttachmentAdapter`, which honours `ChatUI.streamCdnImageResizer` because it's a singleton). Views is the outlier there rather than Compose, since UIKit caps and SwiftUI sizes to the screen, so the fix is Views catching up and not Compose backing off.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
```diff
@@ -177,6 +189,7 @@ public fun ChatTheme(
     imageLoaderFactory: StreamCoilImageLoaderFactory = StreamCoilImageLoaderFactory.defaultFactory(),
     messageAlignmentProvider: MessageAlignmentProvider = MessageAlignmentProvider.defaultMessageAlignmentProvider(),
     streamCdnImageResizing: StreamCdnImageResizing = StreamCdnImageResizing.defaultStreamCdnImageResizing(),
+    streamCdnImageResizer: StreamCdnImageResizer = StreamCdnMaxPixelsImageResizer(),
```
gpunto: This creates a new instance per call and, given that we later use `staticCompositionLocalOf`, every `ChatTheme` recomposition invalidates the whole subtree.

We  could have a `default` object so the instance is always the same.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -85,6 +90,10 @@ public fun String.getStreamCdnHostedImageDimensions(): StreamCdnOriginalImageDim
     return try {
         val imageUri = this.toUri()
 
+        if (imageUri.host?.let(streamCdnHostRegex::matches) != true) {
+            return null
+        }
```
gpunto: This narrows a shipped public function. `getStreamCdnHostedImageDimensions` parsed `ow`/`oh` from any URL before, and now returns null for anything off the allowlist. Three things inherit that:

- the legacy percentage overload of `createResizedStreamCdnImageUrl` is gated on it, so it stops resizing as well
- integrators calling it directly to read the original dimensions, for placeholder aspect ratios say, get null with no resizing involved
- anyone serving Stream images from a proxied or custom domain silently loses resizing they had before

Worth confirming nobody is on a custom domain, and a release-note line either way. iOS treats that as a supported setup: `StreamCDNRequester(cdnHost:)` takes the host as a parameter, and matches it with `host.contains(cdnHost)` rather than an anchored pattern, so a hardcoded regex here leaves no escape hatch.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -25,11 +25,16 @@ import io.getstream.chat.android.models.streamcdn.image.StreamCdnCropImageMode
...
 import io.getstream.log.StreamLog
+import kotlin.math.sqrt
 
 private val snakeRegex = "_[a-zA-Z]".toRegex()
 private val camelRegex = "(?<=[a-zA-Z])[A-Z]".toRegex()
 private val baseUrlRegex = "^(?:https?://)?(.*?)/*$".toRegex()
 
+// Mirrors AttachmentHelper's STREAM_CDN_HOST_PATTERN — resize only Stream-hosted images.
+private val streamCdnHostRegex =
+    "stream-chat-+.+\\.imgix.net$|.+\\.stream-io-cdn.com$".toRegex(RegexOption.IGNORE_CASE)
```
gpunto: Preexisting in `AttachmentHelper` too: dots before `net` and `com` aren't escaped, so `foo.stream-io-cdnXcom` matches. Also, is it worth it to share a constant with `AttachmentHelper.kt:44` so we make sure they don't drift?

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/resizing/StreamCdnImageResizer.kt
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.ui.common.images.resizing
+
+import io.getstream.chat.android.client.extensions.createResizedStreamCdnImageUrl
+import io.getstream.chat.android.models.streamcdn.image.StreamCdnCropImageMode
+import io.getstream.chat.android.models.streamcdn.image.StreamCdnResizeImageMode
+
+public fun interface StreamCdnImageResizer {
```
gpunto, line 33: Should we add some KDoc since this is public code?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -176,6 +177,29 @@ public fun String.createResizedStreamCdnImageUrl(
...
+    val dimensions = getStreamCdnHostedImageDimensions() ?: return this
+    val totalPixels = dimensions.originalWidth.toLong() * dimensions.originalHeight.toLong()
+    if (totalPixels <= 0L || totalPixels <= maxImagePixels) return this
+    val scale = sqrt(maxImagePixels.toDouble() / totalPixels.toDouble()).toFloat()
+    return createResizedStreamCdnImageUrl(
+        resizedWidthPercentage = scale,
+        resizedHeightPercentage = scale,
+        resizeMode = resizeMode,
+        cropMode = cropMode,
+    )
```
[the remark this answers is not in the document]
aleksandar-apostolov: Fixed in `8fb919a0e4`. `getStreamCdnHostedImageDimensions` now validates the host before returning dimensions (mirrors `AttachmentHelper.STREAM_CDN_HOST_PATTERN` — `*.stream-io-cdn.com` and Stream imgix), so non-Stream URLs are never resized. Added a regression test for an external URL carrying `ow`/`oh`.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/resizing/StreamCdnImageResizer.kt
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/** Caps images to [maxImagePixels] total pixels (default 2MP), preserving aspect ratio; on by presence. */
+public class StreamCdnMaxPixelsImageResizer(
+    private val maxImagePixels: Long = StreamCdnImageResizer.DEFAULT_MAX_IMAGE_PIXELS,
+    private val resizeMode: StreamCdnResizeImageMode? = null,
+    private val cropMode: StreamCdnCropImageMode? = null,
+) : StreamCdnImageResizer {
+    override fun resizeUrl(imageUrl: String): String =
+        imageUrl.createResizedStreamCdnImageUrl(maxImagePixels, resizeMode, cropMode)
```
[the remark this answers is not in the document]
aleksandar-apostolov: Fixed in `8fb919a0e4`. `StreamCdnMaxPixelsImageResizer` now `require`s `maxImagePixels > 0`, and the `createResizedStreamCdnImageUrl(maxImagePixels)` overload returns the URL unchanged for a non-positive budget. Added zero and negative budget tests.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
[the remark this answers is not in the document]
aleksandar-apostolov: `StreamCdnImageResizing` is already `@Deprecated` at the class level, so passing it to `ChatTheme(streamCdnImageResizing = …)` already emits a deprecation warning. Kotlin doesn't allow `@Deprecated` on an individual function parameter, so there is nothing further to annotate on the composable param itself — the type-level deprecation covers the migration signal.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/preview/internal/MediaGalleryPage.kt
aleksandar-apostolov: Good catch — fixed. The resizer now travels through `MediaGalleryInjector` alongside the `ImageLoader`, so a custom resizer (or `NoOpStreamCdnImageResizer`) is honored in the fullscreen viewer instead of the hardcoded 2MP. On the parity point: the XML kit not resizing its fullscreen viewer at all is a real gap — I'll file a follow-up for Views to catch up there, out of scope for this PR.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatTheme.kt
aleksandar-apostolov: Fixed. Added a stable internal `DefaultStreamCdnImageResizer` instance used as the default argument, so the `staticCompositionLocalOf` no longer sees a fresh instance each call and stops invalidating the subtree on recomposition. Kept it internal — not a public `.Default`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
aleksandar-apostolov: Agreed, reverted. `getStreamCdnHostedImageDimensions` parses `ow`/`oh` from any URL again — the public contract is restored, so integrators reading dimensions and the legacy percentage overload are unaffected. The host gate moved to the resize site (`createResizedStreamCdnImageUrl(maxImagePixels, …)`), which now takes an optional `cdnHost` matched with `contains` — iOS `StreamCDNRequester(cdnHost:)` parity — exposed via `StreamCdnMaxPixelsImageResizer(cdnHost = …)` so proxied/custom domains have an escape hatch.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
aleksandar-apostolov: Done both. Escaped the dots and promoted the pattern to a single `STREAM_CDN_HOST_PATTERN` (`@InternalStreamChatApi`), now referenced by both `AttachmentHelper` and the resize extensions so they can't drift.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/images/resizing/StreamCdnImageResizer.kt
aleksandar-apostolov: Added a class-level KDoc to the `StreamCdnImageResizer` interface — what it does, the two built-ins, and the opt-out.

gpunto (approved):

PR #6632, opened 2026-08-11
Migrate file upload and message translation to generated models

gpunto:
### Goal

Migrate the file upload response and the message translation request body to the generated network models.

### Implementation

- Add generated `FileUploadResponse`; remove the hand-written `UploadFileResponse` and rename its mapper and
  test to match. It covers all four upload endpoints (`/channels/{type}/{id}/file`,
  `/channels/{type}/{id}/image`, `/uploads/file`, `/uploads/image`).
- The generated `file` is nullable, matching the wire: Go tags it `omitempty`, so an empty asset URL is
  omitted and the old non-null `file: String` would have failed to parse. A response without a URL has
  nothing to attach, so `toUploadedFile()` now returns a `Result` and reports it as a failure;
  `StreamFileUploader` uses `flatMap` to carry that through.
- Add generated `TranslateMessageRequest`; remove the hand-written one. `ChatClient.translate` still takes
  a `String`, converted with `Language.fromString`, which falls back to `Unknown` and sends the value
  through unchanged, so an unsupported language is still rejected by the server rather than locally.
- Exclude the generated `network/**` path from the `LongMethod` detekt rule: the models map each wire enum
  in one exhaustive `when` (57 entries for `Language`), which a formatter cannot split and which would be
  regenerated anyway.

### Testing

- `UploadFileResponseMappingTest` covers the mapping, a missing thumbnail, and the missing-URL failure.
- `TranslateMessageRequestAdapterTest` asserts the serialized body for a known language, for one the models
  do not know, and for every one of the 57 languages the endpoint accepts. Removing the `Language` adapter
  registration makes them fail.
- Device-probed all four upload endpoints plus translate. Every upload resolved a CDN URL, and all four
  returned no `thumb_url`, so the nullable thumbnail path is the normal shape rather than an edge case.
  `translate` sent `{"language":"fr"}`, unchanged from the hand-written body, and an invalid language was
  rejected by the server with its list of accepted values. That list matches the generated enum exactly,
  all 57 values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file and image upload handling, including support for optional thumbnails and missing file URLs.
  * Added upload duration information to processed upload responses.
  * Improved message translation requests for supported and unrecognized language codes.

* **Tests**
  * Added coverage for upload response parsing, upload failures, translation language handling, and serialization behavior.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/FileUploadResponseMapping.kt
```diff
@@ -16,20 +16,22 @@
...
-        val expected = UploadedFile(
-            file = dto.file,
-            thumbUrl = dto.thumb_url,
+/**
+ * The upload endpoints omit `file` when the asset URL is empty, which leaves nothing to attach, so
+ * that is reported as a failure rather than an upload with a blank URL.
+ */
+internal fun FileUploadResponse.toUploadedFile(): Result<UploadedFile> =
+    when (val uploadedFileUrl = file) {
+        null -> Result.Failure(Error.GenericError(message = "Missing file URL in the upload response"))
```
andremion: The KDoc says a blank URL is reported as a failure, but the guard only matches `null`. If the response carries `"file": ""`, this returns `Result.Success(UploadedFile(file = ""))`, so the attachment is marked as uploaded with an empty asset URL and renders broken instead of failing.

Not a regression, the previous non-null `file: String` behaved the same way for `""`. Would `file.isNullOrEmpty()` be closer to what the KDoc describes?

```kotlin
internal fun FileUploadResponse.toUploadedFile(): Result<UploadedFile> =
    when {
        file.isNullOrEmpty() -> Result.Failure(
            Error.GenericError(message = "Missing file URL in the upload response"),
        )
        else -> Result.Success(UploadedFile(file = file, thumbUrl = thumbUrl))
    }
```

andremion (commented):
LGTM. One question on the blank-URL guard.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/FileUploadResponseMapping.kt
gpunto: Agreed, updated to also handle empty strings.

andremion (approved):

Issue #59, opened 2019-09-23
Bug: Concurrent Modification Exception


josheroff-livly:
Steps to Reproduce:
1. Load channel list
2. Wait an arbitrary amount of time (I waited around 20 seconds)
3. Close out of app
4. Open app again

Reproducibility: ~10%

Stacktrace:
> 2019-09-23 11:05:34.074 17365-17448/com.osh.streamchatpoc E/AndroidRuntime: FATAL EXCEPTION: WSS - event handler thread
    Process: com.osh.streamchatpoc, PID: 17365
    java.util.ConcurrentModificationException
        at java.util.ArrayList$Itr.next(ArrayList.java:860)
        at com.getstream.sdk.chat.rest.core.Client.connectionResolved(Client.java:379)
        at com.getstream.sdk.chat.rest.WebSocketService$EchoWebSocketListener.lambda$onMessage$1$WebSocketService$EchoWebSocketListener(WebSocketService.java:272)
        at com.getstream.sdk.chat.rest.-$$Lambda$WebSocketService$EchoWebSocketListener$f9N2ARRqFxybqCam0RKr38hIg2U.run(Unknown Source:4)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at com.getstream.sdk.chat.rest.WebSocketService$EventHandlerThread.run(WebSocketService.java:323)

Version: 1.10.0

tschellenbach:
@tbarbugli interesting bug above.

tbarbugli:
I am unable to reproduce this; @josheroff-livly can you upgrade to latest version and try again?

tbarbugli:
I never managed to reproduce this; @josheroff-livly please re-open if you can still manage to

PR #6634, opened 2026-08-13
Migrate channel update and member request bodies to the generated UpdateChannelRequest model

gpunto:
### Goal

Migrate the channel update and member request bodies to the generated `UpdateChannelRequest` model.

### Implementation

- Replace six hand-written request bodies with the single generated `UpdateChannelRequest`:
  `AcceptInviteRequest`, `RejectInviteRequest`, `AddMembersRequest`, `RemoveMembersRequest`,
  `InviteMembersRequest` and the old `UpdateChannelRequest`. The endpoint takes one shape, so the
  separate wrappers collapse into it, along with the transitive models the body needs.
- Collapse the six `ChannelApi` declarations for `POST /channels/{type}/{id}` into a single
  `updateChannel`, which is the shape the generated API interface has. The `ChatApi` methods keep their
  names; only the endpoint function they call changes.
- Add `Message.toMessageRequest()` for the system message these endpoints can carry, plus
  `MessageRequestAdapter`, `AttachmentRequestAdapter`, `ChannelMemberRequestAdapter` and
  `ChannelInputRequestAdapter` so custom data is flattened to the root on the way out.
- Let `UserResponseAdapter` write a null. `ChannelMemberRequest` is the first outgoing model to embed a
  `UserResponse`, and the read-only adapter declared its `toJson` value non-null, so serializing a member
  threw instead of omitting the absent field. A real value is still refused.
- `SharedLocation` generates `Double` coordinates now that the spec publishes `format: double`, so the
  mapper hands the domain value straight through with no narrowing.

### Testing

- `UpdateChannelRequestAdapterTest` covers the serialized body for each of the collapsed calls, including
  the flattened custom data.
- Device-probed all six calls against throwaway channels, with a second sample user for the member and
  invite paths: `updateChannel` (channel data plus a nested system message), `addMembers`, `removeMembers`,
  `inviteMembers`, then `acceptInvite` and `rejectInvite` answered from the invited user's session. The
  server moved that member from `pending` to `member` with `invite_accepted_at` set.
- Sent a message carrying a shared location with coordinates that a 32-bit float cannot represent
  (`37.7749295`, `-122.4194155`). They came back byte-identical, so the coordinates keep full precision.
- `add_members` serializes as objects and `remove_members` as plain ids, both unchanged from the bodies
  they replace. `invites` does change, from a list of ids to a list of member objects
  (`[{"user_id": "..."}]`), since the endpoint takes one member shape for all three fields. The server
  accepts it and marks the member invited.
- Every action list is now always present, so the body carries `"remove_members":[]` even when adding.
  The E2E mock server picked its add/remove branch by key presence, and an empty array is truthy in Ruby,
  so it took the remove branch on every call and broadcast no `member.added`. Nothing is red today because
  no E2E test adds a member from the app. Fixed by mock-server
  [#59](https://github.com/GetStream/stream-chat-test-mock-server/pull/59), which selects the branch by
  list contents; the real backend is unaffected either way.
- Re-probed the two calls carrying a `ChannelMemberRequest` after the adapter change: both bodies omit
  `user` entirely rather than sending `"user":null`, the server accepted them, and the member's custom data
  still flattened to the root.


gpunto:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1216,16 +1217,16 @@ constructor(
...
             channelId = channelId,
             body = with(dtoMapping) {
-                AddMembersRequest(
-                    add_members = members.map { it.toDto() },
-                    message = systemMessage?.toDto(),
-                    hide_history = hideHistory,
-                    hide_history_before = hideHistoryBefore,
-                    skip_push = skipPush,
+                UpdateChannelRequest(
+                    addMembers = members.map { it.toChannelMemberRequest() },
```
andremion, line 1225: The new body always carries `"remove_members":[]` (see `EMPTY_LISTS` in `UpdateChannelRequestTestData`), and the mock server picks the branch by presence, not size: in `src/helpers/members.rb` it does `remove_members = json['remove_members'] ? true : false`. An empty array is truthy in Ruby, so every call to this endpoint now takes the remove branch and `add_members` is ignored, with no `member.added` event.

Nothing is red today because no E2E test adds a member from the app, but the next one would hit it. Could you change the mock to check `&.any?` instead, and drop the "mock server is unaffected" line from the description?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
```diff
@@ -170,6 +174,87 @@ internal class DtoMapping(
...
+        latitude = latitude,
+        longitude = longitude,
+        createdByDeviceId = deviceId,
+        endAt = endAt,
+    )
+
+    /**
+     * Transforms the domain [Message] to the generated network [MessageRequest] model.
+     */
+    internal fun Message.toMessageRequest(): MessageRequest =
```
andremion, line 232: `toMessageRequest` and `toRequest` have no test in `DtoMappingTest`, and the type coercion just below is a copy of the one in `toDto` that `Message toDto coerces type to allowed upstream values` already covers. The adapter test builds `MessageRequest` by hand, so nothing checks the mapping itself.

Could you add the mirror tests next to the existing `toDto` ones? That should also cover most of the Sonar gap.

andremion (commented):
Nice cleanup, the single request shape reads much better than six wrappers. Two things inline, one of them needs a change in the mock server repo.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
```diff
@@ -170,6 +174,87 @@ internal class DtoMapping(
         end_at = endAt,
     )
 
+    /**
+     * Maps the domain [Attachment] to the generated network Attachment model.
+     */
+    internal fun Attachment.toRequest(): io.getstream.chat.android.network.models.Attachment {
```
andremion: Small thing: `toRequest` reads a bit inconsistent next to `toMessageRequest` and `toChannelMemberRequest`. `toAttachmentRequest` would match, and an import alias would let you drop the two fully qualified type names. Not blocking.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Good catch, and it would have bitten the next E2E test that adds a member. Fixed in [mock-server #59](https://github.com/GetStream/stream-chat-test-mock-server/pull/59), which picks the branch by list contents instead of key presence.

That change also needs a nil guard on `member_ids`: flipping the branch means a body carrying `remove_members: []` and no `add_members` key now lands on the add branch and finds `nil`, so `.each` would raise. Checked both branches across the payload shapes.

I also dropped the "mock server is unaffected" line from the description. It was wrong: I checked which fields the mock reads and never evaluated its control flow against our actual body.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
gpunto: Added `Attachment`, `MemberData` and `Message` mapper tests next to the existing `toDto` ones, plus a parameterized one reusing `messageTypeCoercionInput` for the coercion.

I mutation-checked them rather than trusting green: swapping `authorName`/`fallback`, dropping the `file_size` fold, and removing the coercion each make them fail, so they exercise the mapping instead of restating it.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
gpunto: Renamed to `toAttachmentRequest`, and added `import ...network.models.Attachment as AttachmentRequest` which drops both fully qualified names.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DtoMappingTest.kt
```diff
@@ -217,6 +218,108 @@ internal class DtoMappingTest {
...
+        request.custom shouldBeEqualTo memberData.extraData
+        // The endpoint takes the id; sending a whole user is neither needed nor serializable.
+        request.user shouldBeEqualTo null
+        request.channelRole shouldBeEqualTo null
+    }
+
+    @Test
+    fun `Message is correctly mapped to the generated request model`() {
+        val message = randomMessage(type = MessageType.REGULAR)
+        val messageTransformer = MessageTransformer { it }
```
andremion: Small thing: `MessageTransformer { it }` is what the default fixture already uses, and there is no `verify` here, so dropping `messageTransformer.transform(this)` from `toMessageRequest` would not fail anything. The `toDto` test guards that with `spy(NoOpMessageTransformer)` plus a `verify`. Could you do the same here? Transformers are a public hook for outgoing messages, so it feels worth pinning. Not blocking.

andremion (approved):
Looks good, thanks for the mock server fix too. One optional note inline.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DtoMappingTest.kt
gpunto: Right, that one could not fail. Swapped it to `spy(NoOpMessageTransformer)` with a `verify`, matching the `toDto` test, and checked it bites: with `messageTransformer.transform(this)` stripped from `toMessageRequest` the test fails, and it passes again once restored. `6113227833`.

PR #6635, opened 2026-08-13
Migrate the thread channel to the generated ChannelResponse model

gpunto:
### Goal

Adopt the generated `ChannelResponse` for the channel embedded in a thread, replacing the hand-written
`DownstreamChannelDto` on that path.

### Implementation

- `DownstreamThreadDto.channel` and `DownstreamThreadInfoDto.channel` now use the generated
  `ChannelResponse`, along with the transitive models it needs.
- Add `ChannelResponse.toDomain()`, `ChannelResponse.toChannelInfo()`, `ChannelConfigWithInfo.toDomain()`
  and `ChannelMemberResponse.toDomain()`. `name` and `image` are custom data on the wire, so they are
  promoted out of `custom` and excluded from `extraData`.
- Add `ChannelResponseAdapter` and `ChannelMemberResponseAdapter` to collect root-level custom data, plus
  the generated sealed-class enum adapters they need.
- Fix a false positive in `CustomAdapterCoverageTest`: the marker needs its trailing colon, otherwise it
  also matches `customEvents`.

### Notes

- The generated models declare wire fields their hand-written predecessors did not, which drops those keys
  from `extraData`, since the collecting adapter only sweeps keys that are not constructor parameters. They
  are mapped to their domain properties where one exists, and the adapters also pass them to
  `alsoKeepInExtraData` so they stay in the overflow map with their raw wire values, as
  `DownstreamChannelDtoAdapter` does for the legacy keys. Two sets, both dropped with AND-1398:
  `GENERATED_CHANNEL_EXTRA_DATA_KEYS` (`auto_translation_enabled`, `auto_translation_language`, `hidden`,
  `hide_messages_before`, `mute_expires_at`, `muted`, `truncated_by`) and
  `GENERATED_MEMBER_EXTRA_DATA_KEYS` (`deleted_at`, `deleted_messages`, `is_moderator`, `role`, `user_id`).
  They are separate from the legacy set because adding them there would be a no-op for
  `DownstreamChannelDto`, which never declared them.
- One field cannot be made consistent here: `Config.messageRetention` falls back to its domain default
  `"infinite"`, so a channel type with numeric retention reports `"infinite"` inside a thread and its real
  value from `queryChannel`. The wire does send `message_retention`, but Go tags it `openapi:"-"` so it is
  absent from the generated model, and it is nested inside `config`, where the root-level `custom` sweep
  cannot reach it. Declaring it on the vendored model would be undone by the next re-vendor, so this needs
  a spec change; tracked separately. Nothing in the SDK reads the field, and channel configs are persisted
  only from `queryChannel`/`queryChannels`, so a stored value cannot be overwritten with the default.
- Go's `payload.ChannelResponse` declares no `messages`, `watchers`, `read`, `membership` or
  `pinned_messages`, so those were never sent on this path and stay empty as before.
- `hidden` and `blocked` reach `extraData` from `queryChannel` but not from a thread. Both paths serialise
  the channel with the same Go struct, where the two fields are `*bool` with `omitempty` and documented as
  the current user's state for that channel. The threads endpoint has no per-user channel session to fill
  them, so they stay nil and are omitted from the payload entirely. Nothing arrives, so no keep set can
  hold them, and the hand-written DTO behaved the same way on this path.

### Testing

- `DomainMappingTest` covers the state fields on both the properties and `extraData`.
  `GeneratedExtraDataParityTest` covers both keep sets, asserting every declared key still reaches
  `extraData` while also being parsed into its own field; both of its tests fail if either keep set is
  removed. The thread adapter tests cover the wiring end to end through a full thread payload.
- Device-probed against a throwaway channel, truncated before the thread was created so `truncated_at` is
  actually populated. On `queryThreads`, `getThread` and `partialUpdateThread` the channel came back with
  `name` promoted out of `custom`, the sentinel intact, `truncatedAt` set, `extraData["truncated_at"]`
  holding the raw wire string, and `disabled` agreeing between the property and `extraData`.
- Probed both keep sets against a real payload rather than only the fixtures. For the channel set that
  means `truncated_by`, the only one of the seven the threads endpoint sends without app-level
  configuration; it arrives as a whole user object, custom fields included, and the keep set holds all of
  it. For the member set, `role` and `user_id` match what `queryChannel` reports for the same member.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
@@ -819,9 +819,7 @@ internal class DomainMapping(
     )
 
     /**
-     * Transforms [ChannelConfigWithInfo] to [Config]. The wire carries `message_retention`, but Go
-     * tags it `openapi:"-"` so it is absent from the generated model; we keep the domain default
-     * until the spec describes it.
+     * Transforms [ChannelConfigWithInfo] to [Config].
      */
     internal fun ChannelConfigWithInfo.toDomain(): Config = Config(
         createdAt = createdAt,
@@ -841,6 +839,7 @@ internal class DomainMapping(
         pushNotificationsEnabled = pushNotifications,
         skipLastMsgUpdateForSystemMsgs = skipLastMsgUpdateForSystemMsgs,
         pollsEnabled = polls,
+        messageRetention = messageRetention,
         maxMessageLength = maxMessageLength,
         automod = automod.value,
         automodBehavior = automodBehavior.value,
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
@@ -130,7 +130,11 @@ import io.getstream.chat.android.models.querysort.SortDirection
 import io.getstream.chat.android.network.models.AppResponseFields
 import io.getstream.chat.android.network.models.BlockUsersResponse
 import io.getstream.chat.android.network.models.BlockedUserResponse
+import io.getstream.chat.android.network.models.ChannelConfigWithInfo
+import io.getstream.chat.android.network.models.ChannelMemberResponse
+import io.getstream.chat.android.network.models.ChannelOwnCapability
 import io.getstream.chat.android.network.models.ChannelPushPreferencesResponse
+import io.getstream.chat.android.network.models.ChannelResponse
 import io.getstream.chat.android.network.models.ChatPreferencesResponse
 import io.getstream.chat.android.network.models.DeviceResponse
 import io.getstream.chat.android.network.models.GetApplicationResponse
@@ -226,6 +230,53 @@ internal class DomainMapping(
             name = name,
             type = type,
             image = image,
+        )
+
+    /**
+     * Transforms [ChannelResponse] into [Channel]. Channel-level wire fields only: the type carries
+     * no `messages`, `watchers`, `read`, `pinned_messages`, `membership` or `active_live_locations`,
+     * so those stay empty. `name` and `image` are custom data on the wire.
+     */
+    internal fun ChannelResponse.toDomain(): Channel =
+        Channel(
+            id = id,
+            type = type,
+            name = custom["name"] as? String ?: "",
+            image = custom["image"] as? String ?: "",
+            filterTags = filterTags.orEmpty(),
+            frozen = frozen,
+            createdAt = createdAt,
+            deletedAt = deletedAt,
+            updatedAt = updatedAt,
+            memberCount = memberCount ?: 0,
+            members = members.orEmpty().map { it.toDomain() },
+            hidden = hidden,
+            hiddenMessagesBefore = hideMessagesBefore,
+            truncatedAt = truncatedAt,
+            disabled = disabled,
+            blocked = blocked,
+            config = config?.toDomain() ?: Config(),
+            createdBy = createdBy?.toDomain() ?: User(),
+            team = team.orEmpty(),
+            cooldown = cooldown ?: 0,
+            ownCapabilities = ownCapabilities.orEmpty().mapTo(mutableSetOf(), ChannelOwnCapability::value),
+            messageCount = messageCount,
+            lastMessageAt = lastMessageAt,
+            extraData = custom.mapNotNull { (key, value) -> value?.let { key to it } }
+                .toMap()
+                .minus(listOf("name", "image"))
+                .toMutableMap(),
+        ).syncUnreadCountWithReads(currentUserIdProvider())
+            .let(channelTransformer::transform)
+
+    internal fun ChannelResponse.toChannelInfo(): ChannelInfo =
+        ChannelInfo(
+            cid = cid,
+            id = id,
+            memberCount = memberCount ?: 0,
+            name = custom["name"] as? String,
+            type = type,
+            image = custom["image"] as? String,
         )
 
     /**
@@ -449,6 +500,25 @@ internal class DomainMapping(
             pinnedAt = pinned_at,
             archivedAt = archived_at,
             extraData = extraData,
+        )
+
+    internal fun ChannelMemberResponse.toDomain(): Member =
+        Member(
+            user = user?.toDomain() ?: User(id = userId.orEmpty()),
+            createdAt = createdAt,
+            updatedAt = updatedAt,
+            isInvited = invited,
+            inviteAcceptedAt = inviteAcceptedAt,
+            inviteRejectedAt = inviteRejectedAt,
+            shadowBanned = shadowBanned,
+            banned = banned,
+            channelRole = channelRole,
+            notificationsMuted = notificationsMuted,
+            status = status,
+            banExpires = banExpires,
+            pinnedAt = pinnedAt,
+            archivedAt = archivedAt,
+            extraData = custom.mapNotNull { (key, value) -> value?.let { key to it } }.toMap(),
         )
 
     internal fun UserResponse.toDomain(): User =
@@ -746,6 +816,40 @@ internal class DomainMapping(
         sharedLocationsEnabled = shared_locations ?: false,
         markMessagesPending = mark_messages_pending,
         pushLevel = push_level,
+    )
+
+    /**
+     * Transforms [ChannelConfigWithInfo] to [Config]. The wire carries `message_retention`, but Go
+     * tags it `openapi:"-"` so it is absent from the generated model; we keep the domain default
+     * until the spec describes it.
+     */
+    internal fun ChannelConfigWithInfo.toDomain(): Config = Config(
+        createdAt = createdAt,
+        updatedAt = updatedAt,
+        name = name,
+        typingEventsEnabled = typingEvents,
+        readEventsEnabled = readEvents,
+        deliveryEventsEnabled = deliveryEvents,
+        connectEventsEnabled = connectEvents,
+        searchEnabled = search,
+        isReactionsEnabled = reactions,
+        isThreadEnabled = replies,
+        muteEnabled = mutes,
+        uploadsEnabled = uploads,
+        urlEnrichmentEnabled = urlEnrichment,
+        customEventsEnabled = customEvents,
+        pushNotificationsEnabled = pushNotifications,
+        skipLastMsgUpdateForSystemMsgs = skipLastMsgUpdateForSystemMsgs,
+        pollsEnabled = polls,
+        maxMessageLength = maxMessageLength,
+        automod = automod.value,
+        automodBehavior = automodBehavior.value,
+        blocklistBehavior = blocklistBehavior?.value.orEmpty(),
+        commands = commands.map { it.toDomain() },
+        messageRemindersEnabled = userMessageReminders,
+        sharedLocationsEnabled = sharedLocations,
+        markMessagesPending = markMessagesPending,
+        pushLevel = pushLevel?.value,
     )
 
     /**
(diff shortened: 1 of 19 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelResponseAdapter.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.squareup.moshi.ToJson
+import io.getstream.chat.android.network.models.ChannelResponse
+
+// Downstream (read-only) adapter for the generated ChannelResponse: collects root-level custom fields
+// into `custom`, matching the wire's flattened extra data. extraDataPropertyName is its @Json name.
+internal object ChannelResponseAdapter :
+    CustomObjectDtoAdapter<ChannelResponse>(
+        ChannelResponse::class,
+        extraDataPropertyName = "custom",
+        alsoKeepInExtraData = LEGACY_CHANNEL_EXTRA_DATA_KEYS,
```
andremion: `LEGACY_CHANNEL_EXTRA_DATA_KEYS` covers `disabled`, `blocked` and `truncated_at`, but `ChannelResponse` declares more fields that `DownstreamChannelDto` does not: `auto_translation_enabled`, `auto_translation_language`, `muted`, `mute_expires_at`, `truncated_by`, `hidden` and `hide_messages_before`.

`hidden` and `hide_messages_before` are fine, `toDomain()` maps them onto `Channel.hidden` and `Channel.hiddenMessagesBefore`. The other five have no domain property and are not in the keep set, so `parseWithExtraData` removes them from `custom` and nothing reads them again.

They are declared on the wire type, and one of the two server-side thread response paths does populate `muted`, `mute_expires_at` and `truncated_by`. Your probe reports `blocked` and `hidden` as not sent, which matches the other path, so which endpoint you hit decides whether the value is there.

So an app reading `thread.channel.extraData["muted"]` can get a value today and null after this change, while the same channel from `queryChannel` still has it. That is the opposite of what the Notes section says.

Would a `ChannelResponse`-specific key set work better here than reusing the legacy one? The legacy set is sized for `DownstreamChannelDto`, and adding these keys to it would be a no-op there, so a separate set keeps the two intentions readable.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelMemberResponseAdapter.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import com.squareup.moshi.ToJson
+import io.getstream.chat.android.network.models.ChannelMemberResponse
+
+// Downstream (read-only) adapter for the generated ChannelMemberResponse: collects root-level custom fields
+// into `custom`, matching the wire's flattened extra data. extraDataPropertyName is its @Json name.
+internal object ChannelMemberResponseAdapter :
+    CustomObjectDtoAdapter<ChannelMemberResponse>(ChannelMemberResponse::class, extraDataPropertyName = "custom") {
```
andremion: Same situation one level down. `ChannelMemberResponse` declares `user_id`, `is_moderator`, `role`, `deleted_at` and `deleted_messages`, and `DownstreamMemberDto` declares none of them, so they are in `Member.extraData` today. This adapter passes no `alsoKeepInExtraData`, and `toDomain()` maps none of them onto `Member` (except `user_id`, and only as a fallback for the user id), so all five disappear for members inside a thread's channel.

The server sends all five for embedded channel members. Could you add a keep set here too?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -718,6 +788,40 @@ internal class DomainMapping(
         pushLevel = push_level,
     )
 
+    /**
+     * Transforms [ChannelConfigWithInfo] to [Config]. The wire carries `message_retention`, but Go
+     * tags it `openapi:"-"` so it is absent from the generated model; we keep the domain default
+     * until the spec describes it.
+     */
+    internal fun ChannelConfigWithInfo.toDomain(): Config = Config(
```
andremion, line 824: Worth being explicit that this is a wrong value rather than a missing one. `message_retention` is still on the wire, and it is nested inside `config`, so the root-level `custom` sweep cannot recover it either. The result is that `thread.channel.config.messageRetention` always reports `"infinite"`, while the same channel from `queryChannel` reports the real value. For a channel type set to a numeric retention, the two disagree.

Declaring the field on the model would fix it, but the spec deliberately excludes it, so a hand-edit gets undone on the next regeneration. Could you open a ticket for the spec side and link it from this KDoc, so it does not stay as a comment only?

andremion (commented):
Two `extraData` parity gaps, details inline. `alsoKeepInExtraData` covers `disabled`, `blocked` and `truncated_at`, but `ChannelResponse` and `ChannelMemberResponse` declare more keys in the same situation, and those are dropped.

This is easy to miss because `ThreadMapper` does not store the channel on the thread entity and rebuilds it with `getChannel(cid)` (`ThreadMapper.kt:62`). A cached thread keeps the full `extraData`, the same thread from `queryThreads` does not.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
gpunto: Good point, I'll hold off this PR until we sort that out on the specs. I'll address the other findings in the meantime

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelResponseAdapter.kt
gpunto: Added `GENERATED_CHANNEL_EXTRA_DATA_KEYS` with all seven, passed alongside the legacy set. Kept it separate because adding them to `LEGACY_CHANNEL_EXTRA_DATA_KEYS` would be a no-op for `DownstreamChannelDto`, which never declared them.

`GeneratedExtraDataParityTest` asserts every declared key still reaches `extraData` while also being parsed into its own field, and fails if the set is removed. Three existing fixture tests were failing on `hidden` and `hide_messages_before` too, so the fixture now carries them.

On the wire the threads endpoint sends only `truncated_by` of the seven, the rest being per-user session fields or app settings. It arrives as a whole user object with its own custom data, and the keep set holds all of it.

Also repointed the removal milestone: `AND-1375` is the ticket that added the legacy set and it is already done, so both sets now reference `AND-1398`, filed for the removal along with the `alsoKeepInExtraData` plumbing.

Separately on `hidden` and `blocked`: they reach `extraData` from `queryChannel` but not from a thread. Both paths serialise the channel with the same Go struct, where the two are `*bool` with `omitempty` and documented as the current user's state for that channel. The threads endpoint has no per-user channel session to fill them, so they stay nil and are omitted from the payload entirely. Nothing arrives, so no keep set can recover them, and the hand-written DTO behaved the same way on that path.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelMemberResponseAdapter.kt
gpunto: Done, `GENERATED_MEMBER_EXTRA_DATA_KEYS` with all five, covered by the same parity test. Device-probed as well: `role` and `user_id` match what `queryChannel` reports for the same member.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
gpunto: Sorted on the spec side, so this is no longer a wrong value.

chat#15920 (CHA-4534) dropped the `openapi:"-"` from `payload.ChannelConfigFields.MessageRetention` and pruned the stale `required` entries that went with it. A fresh regen off master declares `message_retention` on `ChannelConfigWithInfo` as a required non-null `String`, so the model carries it and the mapper reads it instead of falling back. The KDoc note is gone rather than linked, since there is nothing left to track.

The test that pinned the old behaviour asserted `"infinite"`; it now asserts the fixture's non-default value, which is the case the fallback got wrong.

Device-probed too, and the required-ness makes that check meaningful: an absent `message_retention` now fails the whole `ChannelConfigWithInfo` parse rather than defaulting, so a thread arriving with a config is itself proof the field is on the wire. Both paths report the same value. Worth being explicit about one limit: every channel type reachable with a user token reports `infinite`, so the numeric case is covered by the fixture rather than on device.

One consequence to flag, since it is wider than this field: a producer that omits `message_retention` now fails to parse the config at all. The mock server is fine, its thread-path channel comes from `http_channels.json` whose config carries the key.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelResponseAdapter.kt
andremion: Small thing on the parity test: the JSON sends `truncated_by_id`, which `ChannelResponse` doesn't declare, so it lands in `custom` as plain custom data and doesn't exercise the keep set. `truncated_by` itself isn't in the JSON or the asserted keys, so the one key you say the endpoint actually sends is the only untested one, and it's also the only nested value in the set. Could you add it?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelMemberResponseAdapter.kt
andremion: Two small ones. `deleted_at` is in the keep set but not in the parity test JSON or its assertions, so four of the five are covered.

The KDoc also says `role` and `user_id` are mapped onto the domain model, but `Member` has no `role` and `toDomain()` doesn't read it, so `user_id` is the only one (as the user id fallback). Worth rewording?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
andremion: Fixing it on the spec side is much better than keeping the fallback, thanks for taking it there. Resolving.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelResponseAdapter.kt
gpunto: You're right, and it was the worst key to have missed: `truncated_by_id` is not declared on `ChannelResponse`, so it was only ever ordinary custom data and exercised none of the keep set, leaving the one key the endpoint actually sends as the only untested one.

Added `truncated_by` to the payload as a full user object, and the assertion reads the nested value back rather than just checking the key is present, since surviving as a marker would not prove much for the only nested entry in the set. Mutation-checked as well: dropping `truncated_by` from the keep set now fails the test, which it did not before.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/ChannelMemberResponseAdapter.kt
gpunto: Both fixed.

`deleted_at` is in the payload and the assertions now, so all five keys are covered instead of four. Same mutation check as the channel one: removing it from the keep set fails the test.

On the KDoc you're right, and it was plainly wrong about `role`: `Member` declares `channelRole`, which is mapped from `channel_role`, and `toDomain()` never reads `role` at all. Reworded to say that only `user_id` is read, and only as the fallback for the user id when the payload carries no user.

andremion (approved):
LGTM, thanks for taking the retention fix to the spec instead of working around it.

PR #6636, opened 2026-08-13
Migrate the live location update request body to the generated UpdateLiveLocationRequest model

gpunto:
### Goal

Migrate the live location update request body to the generated `UpdateLiveLocationRequest` model.

### Implementation

- Replace the hand-written `UpdateLiveLocationRequest` with the generated model, used by both
  `updateLiveLocation` and `stopLiveLocation`.
- Stop sending `created_by_device_id`. The endpoint does not accept it: the server payload has no such
  field, and the update writes only the columns the request supplies (`latitude`, `longitude`,
  `end_at`), so the stored device id is never affected. It is set when the location is created, as part
  of the message's `shared_location`, which is unchanged here.
- Note on `updateLiveLocation` and `stopLiveLocationSharing` that their `deviceId` parameter has no effect
  and goes away in the next major. Nothing reads it: the throttling plugin keys on `messageId` and the state
  listener acts on the response, so it only ever reached a request field the endpoint ignores.
  `sendStaticLocation` and `startLiveLocationSharing` keep theirs, since the create path does store it.

The live location responses stay hand-written for now. `SharedLocationResponse` embeds `MessageResponse`
and `ChannelResponse`, which pull in models that still need spec fixes.

### Testing

- `UpdateLiveLocationRequestAdapterTest` pins both bodies this endpoint emits: a coordinate update and
  a stop that carries only `message_id` and `end_at`.
- Device-probed both paths against a throwaway channel. The update went out without a device id and the
  server returned the original one, both in the response and on a follow-up `GET /users/live_locations`.
  Coordinates a 32-bit float cannot represent (`45.4642035`, `9.1899865`) round-tripped unchanged, and
  the location left the active set after the stop.
- Ran the reverse check on `develop`, which still sends the field, since only that build can show what the
  server does with it. The update and stop bodies both carried a fresh `created_by_device_id`, both calls
  took effect (new coordinates, `end_at` set), and both responses still reported the device id from
  creation. The endpoint ignores the field, so dropping it is not a behaviour change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved live-location update and stop-sharing request handling.
  * Removed unnecessary device identifier data from live-location requests.
  * Ensured location coordinates and stop times are serialized correctly.

* **Tests**
  * Added coverage for updating coordinates and stopping live-location sharing.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -2001,9 +2000,8 @@ constructor(
     override fun stopLiveLocation(location: Location): Call<Location> =
         userApi.updateLiveLocation(
             UpdateLiveLocationRequest(
-                message_id = location.messageId,
-                created_by_device_id = location.deviceId,
-                end_at = location.endAt,
+                messageId = location.messageId,
+                endAt = location.endAt,
```
andremion, line 2005: Not blocking: now that neither body carries the device id, the `deviceId` params on `ChatClient.updateLiveLocation` and `stopLiveLocationSharing` are unused. `ThrottlingPlugin` keys on `messageId` and the state listener only reads the response, so nothing consumes it. Worth deprecating them, or would you rather keep them until the rest of live location moves over?

andremion (approved):
Looks good, the device id removal checks out against the endpoint. One non-blocking question inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Good catch, and it holds. Nothing reads them: `ThrottlingPlugin` keys on `location.messageId`, and every `LiveLocationListenerState` callback acts on `result` rather than the passed-in location.

I went one step further to be sure the parameter was not carrying weight on the wire before this PR. On `develop`, which still sends the field, the update and stop bodies each went out with a fresh `created_by_device_id`, both calls took effect (new coordinates, `end_at` set), and both responses still reported the device id from creation. The endpoint has no field for it: the handler decodes into a request without one and its `ToSharedLocation` builds from `message_id`, `latitude`, `longitude` and `end_at` only. So the parameter never had an effect, and this PR only stopped sending something the server discarded.

Rather than a full deprecation I documented it on both functions, since Kotlin cannot deprecate a single parameter and doing it properly means replacement overloads plus an api dump, which is more than this PR should carry. The note says the parameter has no effect and goes away in the next major.

Kept `sendStaticLocation` and `startLiveLocationSharing` as they are: the create path sends the device id nested in the message's `shared_location` and the server does store it, so theirs is real. Same for `Location.deviceId`, which is returned by the server, parsed from events and persisted.

andremion: Thanks for checking it on the wire, that settles it. One small thing on the note: could it carry a ticket, like the `TODO(AND-1375)` in `DownstreamChannelDtoAdapter.kt:47`? A plain "next major" line in KDoc is easy to miss when v8 is cut, and a ticket makes it show up in the cleanup list.

gpunto: Good call, done: AND-1402. Both notes now cite it, and I kept the `DEPRECATED:` prefix so a grep for deprecated code finds them next to the real `@Deprecated` annotations. The ticket is scoped to these two parameters only, since `sendStaticLocation`, `startLiveLocationSharing` and `Location.deviceId` all carry a device id the server actually stores.

PR #6637, opened 2026-08-14
Surface the message author's member custom data on Message.member

gpunto:
### Goal

Surface the message author's channel-member custom data, which the backend can now project onto message payloads, and keep it fresh when the membership changes. The SDK previously read only `channel_role` from `message.member` and dropped the rest.

The projection is gated by the app-level `member_custom_on_messages_enabled` toggle, off by default, so nothing changes for apps that do not enable it.

Closes [AND-1352](https://linear.app/stream/issue/AND-1352/surface-authors-channelmember-custom-on-message-payloads-member-custom)

### Implementation

- Add `MemberInfo` (`channelRole`, `notificationsMuted`, `extraData`) as `Message.member`, deprecating `Message.channelRole` in favour of `member?.channelRole`. Both stay populated.
- Persist it as `MemberInfoEntity` on the message and reply tables, replacing the `channelRole` column, and bump the DB version.
- Refresh channel state, thread replies and the repository on `member.updated`, since the backend sends no `message.updated` for a membership change. The repository write is a targeted column update, deferred under the same mutex as the message inserts.
- Drop `user_id` from the snapshot: the member DTO does not declare it, so the overflow adapter would otherwise pass it off as customer custom data.

### Testing

- Parser parity for both wire shapes and an absent member, Room round trip, targeted update scoping, and the state refresh for both channel state implementations and thread replies.
- Device verified against a live backend: projection present on a wiped database, surviving react and un-react, cleared on unset, no `message.updated` throughout.
- Send and update message responses still omit the projection server side, tracked by the backend team. Not patched here, since an absent value cannot be told apart from a removal, and `message.new` carries it anyway.

### Manual testing

The demo is already set up: the toggle is on in Nessy and `Silent Active` already has flair on two members.

1. Run the sample, log in as any user who is a member of `Silent Active`, and open it.
2. Messages from `Padmé Amidala` carry `member.extraData = {flair={tier=gold}}` and those from `C-3PO` `{tier=bronze}`; everyone else's is empty. Check `Message.member` in the debugger, or `adb logcat | grep Chat:Http | grep flair` for the wire.
3. Change a member's custom data (e.g. with the getstream cli) and confirm messages already on screen pick it up, with no refetch and no `message.updated`:

```
getstream api UpdateMemberPartial --type messaging --id sample-app-channel-8 \
  --request '{"user_id":"leandro","set":{"flair":{"tier":"silver"}}}'
```

## Summary by CodeRabbit

* **New Features**
  * Added `MemberInfo` to messages, including channel role, notification status, and custom data.
  * Added message builder support for setting member information.
  * Added parsing support for both legacy inline and nested custom member data.

* **Bug Fixes**
  * Member updates now refresh message and thread author details in real time and offline caches.
  * Preserved compatibility with the deprecated `channelRole` field.

* **Refactor**
  * Improved persistence and synchronization of member snapshots for messages and replies.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
@@ -35,9 +35,10 @@ public fun Collection<Member>.updateUsers(userMap: Map<String, User>): Collectio
 /**
  * Narrows a full [Member] down to the slim [MemberInfo] carried by [Message.member].
  *
- * [Member.extraData] holds every key the member DTO does not declare, which includes `user_id`. The projection the
- * backend puts on `message.member` never carries it, so it is dropped here to keep [MemberInfo.extraData] identical
- * no matter whether it came from a message payload or from a member event.
+ * [Member.extraData] holds every key the member DTO does not declare, which includes `user_id` and the deprecated
+ * member-level `role` that channel payloads send next to `channel_role`. The projection the backend puts on
+ * `message.member` carries neither, so both are dropped here to keep [MemberInfo.extraData] identical no matter
+ * whether it came from a message payload or from a member event.
  */
 @InternalStreamChatApi
 public fun Member.toMemberInfo(): MemberInfo = MemberInfo(
@@ -47,4 +48,4 @@ public fun Member.toMemberInfo(): MemberInfo = MemberInfo(
 )
 
 /** Keys that reach [Member.extraData] only because the member DTO does not declare them. */
-private val NON_CUSTOM_MEMBER_KEYS = setOf("user_id")
+private val NON_CUSTOM_MEMBER_KEYS = setOf("user_id", "role")
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
@@ -35,10 +35,10 @@ public fun Collection<Member>.updateUsers(userMap: Map<String, User>): Collectio
 /**
  * Narrows a full [Member] down to the slim [MemberInfo] carried by [Message.member].
  *
- * [Member.extraData] holds every key the member DTO does not declare, which includes `user_id` and the deprecated
- * member-level `role` that channel payloads send next to `channel_role`. The projection the backend puts on
- * `message.member` carries neither, so both are dropped here to keep [MemberInfo.extraData] identical no matter
- * whether it came from a message payload or from a member event.
+ * [Member.extraData] holds every key the member DTO does not declare, which includes stored member fields such as
+ * `role` and `deleted_messages`. The projection the backend puts on `message.member` carries none of them, so they
+ * are dropped here to keep [MemberInfo.extraData] identical no matter whether it came from a message payload or from
+ * a member event.
  */
 @InternalStreamChatApi
 public fun Member.toMemberInfo(): MemberInfo = MemberInfo(
@@ -47,5 +47,16 @@ public fun Member.toMemberInfo(): MemberInfo = MemberInfo(
     extraData = extraData - NON_CUSTOM_MEMBER_KEYS,
 )
 
-/** Keys that reach [Member.extraData] only because the member DTO does not declare them. */
-private val NON_CUSTOM_MEMBER_KEYS = setOf("user_id", "role")
+/**
+ * Keys that reach [Member.extraData] only because the member DTO does not declare them: every `ChannelMemberResponse`
+ * field the DTO leaves out, apart from `custom`, which is where member custom data itself arrives.
+ */
+private val NON_CUSTOM_MEMBER_KEYS = setOf(
+    "user_id",
+    "role",
+    "is_moderator",
+    "deleted_messages",
+    "deleted_at",
+    "ban_from_future_channels",
+    "future_channel_ban_expires",
+)
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
@@ -18,6 +18,8 @@ package io.getstream.chat.android.client.extensions.internal
 
 import io.getstream.chat.android.core.internal.InternalStreamChatApi
 import io.getstream.chat.android.models.Member
+import io.getstream.chat.android.models.MemberInfo
+import io.getstream.chat.android.models.Message
 import io.getstream.chat.android.models.User
 
 /** Updates collection of members with more recent data of [users]. */
@@ -29,3 +31,20 @@ public fun Collection<Member>.updateUsers(userMap: Map<String, User>): Collectio
         member
     }
 }
+
+/**
+ * Narrows a full [Member] down to the slim [MemberInfo] carried by [Message.member].
+ *
+ * [Member.extraData] holds every key the member DTO does not declare, which includes `user_id`. The projection the
+ * backend puts on `message.member` never carries it, so it is dropped here to keep [MemberInfo.extraData] identical
+ * no matter whether it came from a message payload or from a member event.
+ */
+@InternalStreamChatApi
+public fun Member.toMemberInfo(): MemberInfo = MemberInfo(
+    channelRole = channelRole,
+    notificationsMuted = notificationsMuted ?: false,
+    extraData = extraData - NON_CUSTOM_MEMBER_KEYS,
+)
+
+/** Keys that reach [Member.extraData] only because the member DTO does not declare them. */
+private val NON_CUSTOM_MEMBER_KEYS = setOf("user_id")
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Message.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Message.kt
@@ -29,15 +29,13 @@ import java.util.Date
 /**
  * Replaces the [Message.member] snapshot of this message, keeping the deprecated [Message.channelRole] in sync with it.
  *
- * A member payload that carries no `channel_role` must not wipe the role already known for the message: the backend
- * always assigns one, so an absent role means "not sent" rather than "no role".
+ * The snapshot is taken verbatim: the same value reaches the in-memory state, the repository cache and the database, so
+ * the three cannot disagree. A blanket column update cannot preserve a previously known role, so neither does this.
  */
 @InternalStreamChatApi
 @Suppress("DEPRECATION")
-public fun Message.withMemberInfo(memberInfo: MemberInfo?): Message {
-    val resolved = memberInfo?.let { it.copy(channelRole = it.channelRole ?: member?.channelRole) }
-    return copy(member = resolved, channelRole = resolved?.channelRole)
-}
+public fun Message.withMemberInfo(memberInfo: MemberInfo?): Message =
+    copy(member = memberInfo, channelRole = memberInfo?.channelRole)
 
 /** Updates collection of messages with more recent data of [users]. */
 @InternalStreamChatApi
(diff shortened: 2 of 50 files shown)


gpunto:
@coderabbitai review

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/offline/repository/domain/message/internal/DatabaseMessageRepository.kt
```diff
@@ -84,6 +86,29 @@ internal class DatabaseMessageRepository(
...
+    override suspend fun updateChannelUserMessagesMember(cid: String, userId: String, member: MemberInfo?) {
+        val entity = member?.toEntity()
+        // The caches are read before the database, so they follow the edit straight away.
+        patchCachedMember(messageCache, cid, userId, member)
+        patchCachedMember(replyMessageCache, cid, userId, member)
+        // Deferred under the same mutex as the inserts, so this cannot run ahead of a message still being written.
+        scope.launchWithMutex(dbMutex) {
+            messageDao.updateMemberByCidAndUserId(cid, userId, entity)
+            replyMessageDao.updateMemberByCidAndUserId(cid, userId, entity)
+        }
```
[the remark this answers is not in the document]
gpunto: The race is real, and the comment now describes it instead of claiming ordering.

Awaiting this write doesn't fix it though: an insert launched earlier and not yet run still contends for the same `dbMutex`, so awaiting only this one changes nothing about their relative order. Real ordering needs a single serialized writer for the repository, since `insertMessages` and the delete paths are fire-and-forget too. That's out of scope here, and awaiting just this write would make it the outlier while fixing nothing.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/event/handler/internal/EventHandlerSequential.kt
```diff
@@ -582,6 +583,15 @@ internal class EventHandlerSequential(
...
             }
 
+        // A membership change carries no message id, so it cannot be grouped like the events above: every active
+        // thread is asked to refresh the author's replies, which the channel state refresh does not reach.
+        sortedEvents.filterIsInstance<MemberUpdatedEvent>().forEach { event ->
+            val memberInfo = event.member.toMemberInfo()
+            logicRegistry.getActiveThreadsLogic().forEach { thread ->
+                thread.updateMessagesMemberInfo(event.cid, event.member.getUserId(), memberInfo)
+            }
+        }
```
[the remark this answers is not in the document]
gpunto: Real, but not specific to this change: any local state mutation racing an in-flight fetch behaves this way. `UserUpdatedEvent` refreshing the user on stored messages has the same exposure, as do reads and reactions. Nothing about the member snapshot makes it worse, and it self-corrects on the next fetch.

Retaining and replaying pending snapshots would add state with its own invalidation problem to work around a general property of the state-versus-fetch model, so accepting it here.

One correction to my own earlier reasoning while I'm here: the projection cap is not a related case. CHA-4513 aligned it to the 5120 write cap, so anything a customer can store in member custom can also be projected.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/offline/repository/domain/message/internal/MessageMemberInfoDaoTest.kt
```diff
@@ -0,0 +1,162 @@
...
+    @Test
+    fun `updating the member clears it when null is stored`(): Unit = runBlocking {
+        val author = randomUser(id = "author")
+        val member = MemberInfo(channelRole = "channel_member", extraData = mapOf("flair" to "gold"))
+        val message = randomMessage(cid = CID, user = author, member = member, replyTo = null, poll = null)
+        messageDao.insert(message.toEntity())
+
+        messageDao.updateMemberByCidAndUserId(CID, author.id, null)
+
+        messageDao.select(message.id)?.messageInnerEntity?.member.shouldBeNull()
```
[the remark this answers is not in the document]
gpunto: Good catch, done.

The constraint I would have cited (`Mother.kt:396`) is about `runBlockingTest`, which eagerly hijacked dispatchers and deadlocked against Room's separate transaction executor. `runTest` does not, and all 8 tests pass under it across repeated runs. `MessageMemberRefreshRepositoryTest` in this PR was already driving a real Room database from `runTest`.

Not taking the second half: a virtual-time test of insert-versus-refresh ordering would assert an ordering the repository does not provide, so it would either encode accidental behaviour or flake. Worth adding if we introduce a single serialized writer.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/channel/controller/attachment/UploadAttachmentsIntegrationTests.kt
```diff
@@ -231,6 +232,10 @@ internal class MockMessageRepository : MessageRepository {
         TODO("Not yet implemented")
     }
 
+    override suspend fun updateChannelUserMessagesMember(cid: String, userId: String, member: MemberInfo?) {
+        TODO("Not yet implemented")
+    }
```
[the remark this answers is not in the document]
gpunto: `MockMessageRepository` has 18 `TODO("Not yet implemented")` methods, including `selectAllChannelUserMessages`, so this matches the file's convention. No `member.updated` reaches this repository in the upload tests.

stream-chat-android-core/src/main/java/io/getstream/chat/android/models/Message.kt
```diff
@@ -267,8 +267,18 @@ public data class Message(
...
+        message = "Use member?.channelRole instead.",
+        replaceWith = ReplaceWith("member?.channelRole"),
+        level = DeprecationLevel.WARNING,
+    )
     val channelRole: String? = null,
 
+    /**
+     * Data about the channel membership of the user who sent the message.
+     */
+    val member: MemberInfo? = null,
```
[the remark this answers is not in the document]
gpunto: This repo guarantees source compatibility, not ABI, so binary-only breaks are not blocking here.

Appending `member` last would also achieve little: the constructor and `copy` signatures change wherever it goes, and it would only preserve `component48` onwards for a 50-property data class that nothing destructures. Keeping `member` next to `channelRole`, the field it replaces, is worth more than that.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
```diff
@@ -29,3 +31,20 @@ public fun Collection<Member>.updateUsers(userMap: Map<String, User>): Collectio
...
+ */
+@InternalStreamChatApi
+public fun Member.toMemberInfo(): MemberInfo = MemberInfo(
+    channelRole = channelRole,
+    notificationsMuted = notificationsMuted ?: false,
+    extraData = extraData - NON_CUSTOM_MEMBER_KEYS,
+)
+
+/** Keys that reach [Member.extraData] only because the member DTO does not declare them. */
+private val NON_CUSTOM_MEMBER_KEYS = setOf("user_id")
```
andremion: `user_id` is not the only key that reaches here. The member object on `member.updated` is the full member payload, so it also carries `role` (the deprecated one) next to `channel_role`, plus `is_moderator` and `deleted_messages` when they are set. None of those are declared on `DownstreamMemberDto`, so they all reach `Member.extraData`. Our own recorded fixture shows it: every member in `stream-chat-android-ui-uitests/src/androidTest/resources/http_channel.json` has `role` and `user_id` next to `channel_role`.

I parsed a member payload through the Moshi parser to check and got `{user_id=leandro, role=member, flair={tier=gold}}`. So after a `member.updated` this produces `extraData = {role=member, flair=...}`, while the message payload for the same author gives `{flair=...}`. That is the invariant the KDoc above promises, and it also means `hasOutdatedMemberInfo` reports outdated again after the next refetch.

`role` can never be real custom data, since a partial member update rejects reserved field names. Maybe add the remaining names to `NON_CUSTOM_MEMBER_KEYS`, or declare them on `DownstreamMemberDto` so they stop reaching `extraData` at all? The new tests would not catch this either, since they all set `extraData` directly rather than going through the parser. Happy to be told otherwise if you already looked at this.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Message.kt
```diff
@@ -20,11 +20,48 @@ import io.getstream.chat.android.client.extensions.getCreatedAtOrDefault
...
+public fun Message.withRefreshedMemberInfo(userId: String, memberInfo: MemberInfo?): Message {
+    val refreshed = if (user.id == userId) withMemberInfo(memberInfo) else this
+    val quoted = refreshed.replyTo
+    return when {
+        quoted == null || quoted.user.id != userId -> refreshed
+        else -> refreshed.copy(replyTo = quoted.withMemberInfo(memberInfo))
+    }
+}
+
+private fun Message.isAuthoredBy(userId: String, memberInfo: MemberInfo?): Boolean =
```
andremion: Small thing: `isAuthoredBy` also checks that the snapshot differs, so at the two call sites it reads like a plain authorship check. Maybe `hasOutdatedMemberInfoFor` or similar?

andremion (commented):
Nice work, this reads well. One thing inline I think is a real gap, plus a naming nit.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
gpunto: Right about channel payloads: `role` is a sibling of `channel_role` there, undeclared on the DTO, so it reaches `Member.extraData`. Your fixture shows it, and so do 29 of 39 live `members[]` bodies in a device capture I had.

The `member` on `member.updated` differs though. Seven of those events in that capture (four users, toggle on) parse to `{user_id}` or `{user_id, flair}`, never `role` - the `role` in those frames is inside the nested `user` object, so it goes to `DownstreamUserDto`. Your `role=member` is a membership value, so that looks like a `members[]` entry. All four `toMemberInfo()` call sites take `MemberUpdatedEvent.member`, so the invariant holds today.

Your last point stands though: nothing pinned it. 7d537f522f adds `role` to the set anyway (reserved, and `toMemberInfo` is public) plus a parser-path test carrying both `user_id` and a stray `role`. Left `is_moderator`/`deleted_messages` out since neither appears on either payload; declaring them on the DTO is the real fix, and it will be done when migrating query-members to the generated OpenAPI models.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Message.kt
gpunto: Fair, renamed to `hasOutdatedMemberInfoFor` in 7d537f522f.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/internal/Member.kt
andremion: Good catch on the payload shape, you are right. My JSON was a `members[]` entry. The reason it does not show up on your path is that the member-level `role` is not stored or cached at all, it is computed only on the channel read paths, so a partial member update leaves it empty and `omitempty` drops it.

One correction: `member.updated` has more than one emitter server side, and another one does fill `role` before sending, so events from that path do carry a member-level `role`. Nothing to do about it, 7d537f522f33bfc9161bd0716c925d8329a126c7 covers both.

On `is_moderator` and `deleted_messages`, I agree on deferring to the OpenAPI migration, but the reason is slightly off. They never appear on `message.member`, which only carries `channel_role`, `notifications_muted` and `custom`. They can appear on the `member.updated` member though, since both are stored member fields sent with `omitempty`, so your capture would only show them once one is true or non-empty. `deleted_messages` is the likelier one, it fills as soon as a member deletes a message for themselves. Still fine to leave for the migration, just so the deferral is not resting on them never being sent.

gpunto: Thanks, both corrections are useful. Good to know a second emitter fills `role`, that makes the blocklist entry load-bearing rather than defensive.

You're right that my deferral reasoning was wrong, so I stopped guessing at keys and took the set from the generated model instead: 1c85f0174d drops every `ChannelMemberResponse` field the DTO leaves out, which covers `is_moderator` and `deleted_messages` along with the rest. `custom` is deliberately kept, since that is where the custom data itself arrives on v2.

Deferring the DTO-level fix to the OpenAPI migration still.

andremion (approved):

PR #6640, opened 2026-08-14
Fix CDN image resizer host matching and dimension edge cases

aleksandar-apostolov:
### Goal

Closes [AND-1393](https://linear.app/stream/issue/AND-1393/cdn-image-resizer-review-follow-up-host-match-negative-dims) — follow-up correctness fixes for the CDN image resizer (shipped in #6631), from code review.

### Implementation

- `isStreamCdnHosted` now matches the default `STREAM_CDN_HOST_PATTERN` **or** a custom `cdnHost` (additive), matching its KDoc — a custom host no longer disables resizing of default Stream CDN images. Blank `cdnHost` values are treated as absent, since `String.contains("")` is always `true` and an explicit `cdnHost = ""` otherwise resized off-CDN URLs carrying `ow`/`oh`.
- The non-positive dimension guard now lives in the **percentage** overload rather than the `maxImagePixels` one. Both overloads are public, and the percentage overload computed `originalWidth * percentage` unchecked, so a direct call on a URL with `ow=-4000` still appended a negative `w`. `maxImagePixels` delegates to it, so one guard now covers both entry points; the `maxImagePixels` path keeps an early return to skip the scale maths.
- Removed the invalid `@Deprecated` `ReplaceWith("streamCdnImageResizer")` on `ChatUI.streamCdnImageResizing` (it generated uncompilable code — a `StreamCdnImageResizing` value assigned to a `StreamCdnImageResizer` property); documented manual migration instead.
- Dropped **every** iOS reference from the resizer's public surface, not just the `StreamCDNRequester(cdnHost:)` one flagged in review: `StreamCdnImageResizer` (interface KDoc + `DEFAULT_MAX_IMAGE_PIXELS` + `StreamCdnMaxPixelsImageResizer`), `ChatTheme`, `ChatUI`, and the two `StreamCdnImageResizing` **deprecation messages**, which surfaced "iOS parity" straight into the IDE. See the parity note below.

### Testing

Client tests: additive host (a default Stream CDN URL still resizes when a custom `cdnHost` is set); blank/whitespace `cdnHost` does not resize an off-CDN URL; non-positive original dimensions via **both** overloads, covering zero, one-negative, and both-negative (the old `totalPixels <= 0` check caught zero and single-negative, so those were unpinned once it was replaced).

Both new guards were verified red before the fix — reverting the blank-host check appends `&w=2000&h=1000` to an off-CDN URL, and removing the percentage-overload guard appends `&w=-2000&h=-1000`.

`spotlessApply`, `detekt`, `apiDump` (no `.api` change), and `StringExtensionsKtTest` (18 tests) all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Image resizing now continues to work for default Stream CDN images when a custom CDN host is configured.
  * Images with invalid or non-positive dimensions are left unchanged instead of being resized incorrectly.

* **Documentation**
  * Updated the migration guidance for the deprecated image-resizing setting to clarify that manual migration is required.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -43,18 +43,14 @@ public val STREAM_CDN_HOST_PATTERN: Regex =
 /**
  * Returns whether this URL is served from a Stream CDN host that understands resizing query parameters.
  *
- * @param cdnHost An optional custom Stream CDN host (e.g. a proxied domain). When provided, the URL host is
- * matched against it with a substring check, mirroring iOS' `StreamCDNRequester(cdnHost:)`; when null, the
- * default [STREAM_CDN_HOST_PATTERN] is used.
+ * @param cdnHost An optional custom Stream CDN host (e.g. a proxied domain), matched with a substring check
+ * in addition to the default [STREAM_CDN_HOST_PATTERN], mirroring iOS' `StreamCDNRequester(cdnHost:)`.
```
andremion: This says the additive matching mirrors iOS' `StreamCDNRequester(cdnHost:)`, but iOS is not additive. It holds a single `cdnHost` that defaults to `"stream-io-cdn.com"` and gates on `host.contains(cdnHost)` alone, with no fallback to a default pattern. Its default does not cover imgix hosts either, unlike `STREAM_CDN_HOST_PATTERN`.

Being additive is the right call for us, because `cdnHost` here is an optional extra rather than the whole config. But could we drop the iOS reference from the KDoc and put the parity note in the PR description instead? Same for the references on line 37 and on the `maxImagePixels` overload, if you want to clean them up in one go.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -223,8 +219,9 @@ public fun String.createResizedStreamCdnImageUrl(
     if (maxImagePixels <= 0L) return this
     if (!isStreamCdnHosted(cdnHost)) return this
     val dimensions = getStreamCdnHostedImageDimensions() ?: return this
+    if (dimensions.originalWidth <= 0 || dimensions.originalHeight <= 0) return this
```
andremion, line 231: The guard is only on this overload. The percentage overload above is public as well (it is in the `.api` dump) and still does `(originalWidth * resizedWidthPercentage).toInt()` with no dimension check, so a URL with `ow=-4000` still gets a negative `w` appended when called directly.

Since this overload delegates to that one, could the check move down into the percentage overload? One check would then cover both entry points.

andremion (commented):
Two comments below. I also agree with CodeRabbit on the blank `cdnHost`: this PR rewrites that expression anyway, so it is a good moment to add the `isNotBlank()` check.

The same applies to #6641, so whatever lands here should land there too.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/extensions/StringExtensionsKtTest.kt
```diff
@@ -239,6 +239,26 @@ internal class StringExtensionsKtTest {
...
+            maxImagePixels = 2_000_000L,
+            cdnHost = "images.example.com",
+        )
+        (resized != streamUrl) shouldBeEqualTo true
+        (resized.toUri().getQueryParameter(QUERY_PARAMETER_KEY_RESIZED_WIDTH) != null) shouldBeEqualTo true
+    }
+
+    @Test
+    fun `Given non-positive original dimensions Should return the url unchanged`() {
+        val negativeDimsUrl = createStreamCdnImageLink(originalWidth = -4000, originalHeight = -2000)
```
andremion: This covers both dimensions negative, which is the case the old `totalPixels <= 0L` check missed. Zero and one negative with the other positive used to be caught by that check, and now nothing pins them. Would you add those two cases here?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
andremion: Done in 08812eb2deaef47e06c9e2ea777f209a5d8a7517. Thanks for also cleaning the references in `ChatTheme`, `StreamCdnImageResizer`, `StreamCdnImageResizing` and `ChatUI`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
andremion: Done in 209db02b55c0ff746b23329c62529a6acffee869, both entry points are covered now.

One nit, not blocking: the guard that stayed here is redundant, because the percentage overload already returns the URL unchanged for the same inputs. Keeping it also means this path skips the new `logger.w`, so someone debugging a non-resized image gets the warning through the percentage overload but not through this one. Would you drop it, or do you prefer keeping the early return to skip the scale maths?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/extensions/StringExtensionsKtTest.kt
andremion: Done in 209db02b55c0ff746b23329c62529a6acffee869. The table covers zero and the mixed sign cases, and the percentage overload has its own test now.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
```diff
@@ -43,18 +43,14 @@ public val STREAM_CDN_HOST_PATTERN: Regex =
...
 @InternalStreamChatApi
 public fun String.isStreamCdnHosted(cdnHost: String? = null): Boolean {
     val host = this.toUri().host ?: return false
-    return if (cdnHost != null) {
-        host.contains(cdnHost, ignoreCase = true)
-    } else {
-        STREAM_CDN_HOST_PATTERN.matches(host)
-    }
+    return STREAM_CDN_HOST_PATTERN.matches(host) ||
+        (cdnHost != null && host.contains(cdnHost, ignoreCase = true))
```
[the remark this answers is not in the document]
andremion: 209db02b55c0ff746b23329c62529a6acffee869

andremion (approved):

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/extensions/StringExtensions.kt
aleksandar-apostolov: https://github.com/GetStream/stream-chat-android/pull/6641#discussion_r3803454023

PR #6644, opened 2026-08-17
Prevent markRead when the last message is a server-rejected local-only echo

gpunto:
### Goal

Sending into a frozen channel returns 201 with the message echoed back as `type: "error"`, which the SDK stores as `COMPLETED` while the server persists nothing. The mark-read guard from #6431 only checks `syncStatus != COMPLETED`, so every subsequent channel open calls `markRead` on a channel the server sees as empty, emitting `message.read` events with no `last_read_message_id` (customer-escalated). Aligns with the iOS fix in [stream-chat-swiftui#1452](https://github.com/GetStream/stream-chat-swiftui/pull/1452).

Port of #6643 to develop. Closes [AND-1395](https://linear.app/stream/issue/AND-1395/mark-read-fires-on-server-rejected-message-echoes-emitting-messageread)

### Implementation

- Promote the existing internal `Message.isLocalOnly()` to `@InternalStreamChatApi` public: not yet synced, or a type the server never persists (error, ephemeral). Matches iOS `ChatMessage.isLocalOnly`.
- `MessageListController.markLastMessageReadInternal`: widen the own-message guard from `syncStatus != COMPLETED` to `isLocalOnly()`. The guard returns before updating `lastSeenMessageId`, so mark-read recovers naturally once a real message arrives.
- `MessageComposerController.enqueueSendMessage`: skip the post-send `markMessageRead` when the echoed message is local-only, since its id doesn't exist server side.

### Testing

- Unit tests: error/ephemeral last message doesn't mark read (with a peer message underneath so filtering can't make the test pass vacuously), a synced message after an error one still marks read, post-send `markMessageRead` skipped for an error echo, `isLocalOnly()` truth table (existing tests folded into one parameterized test).
- The same change was verified on device on the v6 branch (#6643): before the fix each channel open fired `POST /read {"message_id":""}` and the server emitted `message.read` without `last_read_message_id`; after the fix the guard rejects the echo and no call is made.

#### Manual testing steps

1. Create a frozen channel with the logged-in user as its only member and no messages, e.g. with the Stream CLI:
   ```
   getstream api chat GetOrCreateChannel --type messaging --id frozen-test \
     --request '{"data":{"created_by_id":"<user-id>","frozen":true,"members":[{"user_id":"<user-id>"}]}}'
   ```
2. Send a message into it with `ChatClient.instance().sendMessage(...)` from a temporary debug button (the UI blocks the composer on frozen channels). The call succeeds and the result is the echo with `type: "error"` and `syncStatus: COMPLETED`.
3. Open the channel a few times, watching logcat with tag `MessageListController` and the HTTP log.
4. Without the fix, each open fires `POST /channels/.../read` with `{"message_id":""}` and the server emits `message.read` with no `last_read_message_id`. With the fix, each open logs `[markLastMessageRead] ... rejected (own local-only)` and makes no call.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved read-status handling for locally created, pending, error, and ephemeral messages.
  * Prevented messages from being marked as read when delivery is unsuccessful or the message is not eligible.
  * Ensured subsequent synchronized messages are marked as read correctly.

* **New Features**
  * Exposed local-only message detection through an internally supported public API.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
@@ -254,9 +254,9 @@ internal fun DraftMessage.ensureId(): DraftMessage =
 internal fun fallbackMessageId(): String = UUID.randomUUID().toString().lowercase()
 
 /**
- * @return If the message is not part of the server's message list: it is not synced yet, or has a
- * type the server leaves out of message queries and read state (ephemeral previews, error messages
- * such as rejected or moderation-bounced sends).
+ * @return If the message is not part of the server's message list: it is unsynced, or has a type
+ * the server leaves out of message queries and read state (ephemeral previews, error messages such
+ * as rejected or moderation-bounced sends).
  */
 @InternalStreamChatApi
 public fun Message.isLocalOnly(): Boolean =
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
@@ -254,9 +254,9 @@ internal fun DraftMessage.ensureId(): DraftMessage =
 internal fun fallbackMessageId(): String = UUID.randomUUID().toString().lowercase()
 
 /**
- * @return If the message exists only on this client and is not persisted server-side: it is
- * still in flight, failed to send, or has a type the server never persists (ephemeral previews,
- * error messages such as rejected or moderation-bounced sends).
+ * @return If the message is not part of the server's message list: it is not synced yet, or has a
+ * type the server leaves out of message queries and read state (ephemeral previews, error messages
+ * such as rejected or moderation-bounced sends).
  */
 @InternalStreamChatApi
 public fun Message.isLocalOnly(): Boolean =
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
@@ -254,20 +254,12 @@ internal fun DraftMessage.ensureId(): DraftMessage =
 internal fun fallbackMessageId(): String = UUID.randomUUID().toString().lowercase()
 
 /**
- * Returns true if this message is local-only and must be preserved across server message
- * window replacements. Local-only messages are never returned by the server after the
- * initial send attempt completes.
- *
- * Covers:
- * - Pending sends: SYNC_NEEDED, IN_PROGRESS
- * - Attachment upload in-flight: AWAITING_ATTACHMENTS
- * - Send failed: FAILED_PERMANENTLY (user must see to retry or dismiss)
- * - Ephemeral: type == "ephemeral" (e.g. Giphy previews — not re-delivered by server)
- * - Error type: type == "error" (client-generated, not re-delivered by server)
- *
- * DOES NOT include COMPLETED messages — those are already in the server response.
+ * @return If the message exists only on this client and is not persisted server-side: it is
+ * still in flight, failed to send, or has a type the server never persists (ephemeral previews,
+ * error messages such as rejected or moderation-bounced sends).
  */
-internal fun Message.isLocalOnly(): Boolean =
+@InternalStreamChatApi
+public fun Message.isLocalOnly(): Boolean =
     syncStatus in LocalOnlySyncStatuses || type in LocalOnlyMessageTypes
 
 internal val LocalOnlySyncStatuses = setOf(
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -44,6 +44,7 @@ import io.getstream.chat.android.client.utils.message.isDeleted
 import io.getstream.chat.android.client.utils.message.isError
 import io.getstream.chat.android.client.utils.message.isGiphy
 import io.getstream.chat.android.client.utils.message.isLocalOnly
+import io.getstream.chat.android.client.utils.message.isMine
 import io.getstream.chat.android.client.utils.message.isModerationBounce
 import io.getstream.chat.android.client.utils.message.isModerationError
 import io.getstream.chat.android.client.utils.message.isSystem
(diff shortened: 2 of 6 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -1759,13 +1759,11 @@ public class MessageListController(
...
-        // Skip when our own message is at the bottom and hasn't been confirmed by the server.
-        // Without this, marking read on an empty channel (only an in-flight optimistic message
-        // exists) causes the server to persist last_read_message_id = "" because its view of
-        // the channel is empty.
+        // Marking read while our own local-only message is at the bottom makes a channel the
+        // server sees as empty emit message.read with no last_read_message_id.
         val currentUserId = clientState.user.value?.id
-        if (message != null && message.user.id == currentUserId && message.syncStatus != SyncStatus.COMPLETED) {
-            logger.v { "[markLastMessageRead] cid: $cid; rejected[$isInThread] (own unsynced): $messageId" }
+        if (message != null && message.user.id == currentUserId && message.isLocalOnly()) {
```
andremion: The empty `last_read_message_id` only happens when the server's view of the channel has no messages at all. This gate keys off something wider: "the last item is my own local-only message". So it also skips `markRead` in channels that do have server-side messages.

Since the gate returns before `lastSeenMessageId` is set, nothing retries it directly. In Compose it usually recovers, because the `LaunchedEffect` in `Messages.kt` restarts on every new `messages` list and re-emits the bottom item. But `error` and `ephemeral` are terminal, unlike `IN_PROGRESS`, so the gate can stay closed for as long as that message is last.

Case I have in mind: the user is scrolled up in a channel with 3 unread messages from the other side, sends a message, moderation bounces it, then scrolls to the bottom. The bounced message is the last item, so `markRead` is skipped and those 3 messages stay unread even though the server has all of them.

Would it work to skip only when the list has no server-side message at all? Something like:

```kotlin
val hasServerMessage = messagesState.messageItems
    .filterIsInstance<HasMessageListItemState>()
    .any { !(it.message.user.id == currentUserId && it.message.isLocalOnly()) }
if (!hasServerMessage) return
```

That still covers the frozen channel (only the echo is in the list) and keeps the `user.id == currentUserId` condition the peer-message test relies on. It is only a sketch, `currentUserId` would need to move above the check. For reference, the iOS fix takes the retry route instead: it stores the pending message id and re-fires the read event once that message is no longer local-only, and `ChannelController` elsewhere uses `messages.last { !$0.isLocalOnly }?.id` as a fallback.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
```diff
@@ -254,20 +254,12 @@ internal fun DraftMessage.ensureId(): DraftMessage =
...
- * Covers:
- * - Pending sends: SYNC_NEEDED, IN_PROGRESS
- * - Attachment upload in-flight: AWAITING_ATTACHMENTS
- * - Send failed: FAILED_PERMANENTLY (user must see to retry or dismiss)
- * - Ephemeral: type == "ephemeral" (e.g. Giphy previews — not re-delivered by server)
- * - Error type: type == "error" (client-generated, not re-delivered by server)
- *
- * DOES NOT include COMPLETED messages — those are already in the server response.
+ * @return If the message exists only on this client and is not persisted server-side: it is
+ * still in flight, failed to send, or has a type the server never persists (ephemeral previews,
```
andremion: "a type the server never persists" is true for `error` but not for `ephemeral`. I checked the send path on the backend: only the error case returns before the message is stored. An ephemeral message goes through the normal insert, the row exists, and the run-message-action endpoint loads it by id. Ephemeral and error rows are left out of message queries and read state by a type filter, not by being absent.

Could we say "not returned in message queries" instead? That is accurate for both types.

This also affects the gate in `MessageListController`. Since the server's read state already ignores ephemeral rows, including `EPHEMERAL` there does not prevent the reported symptom, it only costs a legitimate `markRead` while an unanswered giphy preview sits at the bottom of a channel or a thread. In the composer it is more defensible, because marking read by an ephemeral message id would point `last_read_message_id` at a row that never shows up in the query window.

andremion (commented):
LGTM. Two things I would like your view on: how wide the mark-read gate should be, and the ephemeral part of the new KDoc.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
gpunto: You are right, and the moderation-bounce case is a regression I introduced. Changed the gate to yours in f5e75ac.

I checked the read-state write path to be sure about the boundary:

```go
func MessageContributesToUnreadCount(msg *types.Message) bool {
	return !msg.Silent && !msg.Shadowed && !msg.IsEphemeral() && !msg.IsError()
}
```

`HandleMessageCreated` skips `ChannelMessagesAdd` when that returns false, and `MarkRead` with an empty `msgID` resolves the id via `ChannelMessagesLastMessage` on that store. So the empty `last_read_message_id` happens exactly when the store has nothing for the channel, and when it does have something the server resolves the read state itself and my gate was pure loss. Terminal `error` made it worse than the in-flight case, as you say.

The gate is now:

```kotlin
val hasServerSideMessage = messageItems.any { item ->
    !(item.message.isMine(currentUserId) && item.message.isLocalOnly())
}
if (!hasServerSideMessage) return
```

`currentUserId` moved above it and the item list is computed once. Tests: the error and ephemeral cases now hold only our own local-only message (the actual frozen-channel shape), plus a new test for your scenario, a peer message followed by our own error echo, asserting `markRead` fires and `lastSeenMessageId` lands on the error message. The empty-list, peer-in-flight and own-in-flight cases are unchanged.

I did not take the iOS retry route: with this gate there is nothing left to retry, since the only skipped case is a channel the server considers empty, where any later message opens the gate on the next emission.

One thing that predicate also tells us: `Silent` is in there too, so a channel whose only messages are silent ones can produce the same empty `last_read_message_id` without any local-only message involved. Out of scope here, happy to file it separately if you think it is worth it.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
gpunto: Confirmed and fixed in f5e75ac. `state/messages/pg.go` selects the ephemeral row by PK and updates it in place when the action makes it visible, so the row exists; it is the type filters that leave it out. New wording:

```
@return If the message is not part of the server's message list: it is not synced yet, or has a
type the server leaves out of message queries and read state (ephemeral previews, error messages
such as rejected or moderation-bounced sends).
```

On `EPHEMERAL` in the gate: keeping it, but the reason is narrower than what I had. Read state excludes ephemeral rows the same way it excludes error ones (`MessageContributesToUnreadCount` checks both), so a channel whose only message is an unanswered giphy preview also resolves to an empty `last_read_message_id`. What was wrong was blocking `markRead` when the server does have messages, and the new gate no longer does that, so the cost you describe is gone. The composer keeps both types for the reason you give: it passes an explicit message id, and on `ErrNotFound` the backend still stores the id it was given.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
andremion: Correction to my first line: the old error and ephemeral tests would have failed under the new gate, not passed, since both asserted no `markRead` with a peer message in the list. Your split is the accurate one.

On silent messages, yes, please file it. This gate does not catch that case, since those messages are neither ours nor local-only.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
andremion: Nit, take it or leave it: "not synced yet" also covers `FAILED_PERMANENTLY`, which is closer to never than to yet.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
gpunto: Filed as AND-1403, with the predicate, why this gate cannot catch it (silent messages are ordinary server-persisted messages, so the client has no signal that the read-state store is empty), a repro sketch, and a note that the fix is likely server-side: letting `MarkRead` fall back to the last query-visible message when the store has nothing. Flagged as not-yet-reproduced, since I only read it out of the backend source. It also covers shadowed messages, which sit in the same predicate.

On the tests, agreed, and the split is worth being explicit about: the error and ephemeral negatives fail if the gate is removed altogether, but they pass under both the old over-wide gate and this one, so on their own they would have locked the regression in. The new positive (peer message, then our own error echo) is the one that fails under the over-wide gate.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/utils/message/MessageUtils.kt
gpunto: Fair, "yet" implies it eventually gets there. Changed to "unsynced" in c1672c81a1:

```
@return If the message is not part of the server's message list: it is unsynced, or has a type the
server leaves out of message queries and read state (ephemeral previews, error messages such as
rejected or moderation-bounced sends).
```

andremion (approved):
LGTM. Both points addressed, and thanks for filing AND-1403 for the silent-message case.

PR #6645, opened 2026-08-18
Migrate the guest user response to the generated CreateGuestResponse model

gpunto:
### Goal

Migrate the `getGuestUser` response to the generated `CreateGuestResponse` model.

### Implementation

- Replace the hand-written `TokenResponse` on the guest endpoint with the generated `CreateGuestResponse`,
  mapping its `UserResponse` to the domain user.

### Testing

- `CreateGuestResponseParsingTest` parses a payload captured from the endpoint, with only the access token
  replaced, so it pins what the server actually sends: an empty `language`, `name` at the root where the
  model declares it, and `nickname` beside it for the custom sweep. `UserResponseAdapterTest` covers the
  adapter that collects the user's custom data, which had no dedicated test before.
- A success case asserts the response maps onto the domain `GuestUser`. `access_token` and `duration` are
  both required strings, so swapping them compiled and passed everything before this; the test fails on
  that swap now.
- Device-probed on the wire with `getGuestToken`: the requested `name` round-trips, `extraData` stays empty,
  the returned id is the server's `guest-<uuid>-<requested id>` form, and the access token comes back as a
  JWT rather than another of the response's required strings.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -1428,7 +1428,7 @@ constructor(
         ).mapDomain { response ->
             GuestUser(
                 response.user.toDomain(),
-                response.access_token,
+                response.accessToken,
```
andremion, line 1431: Nothing asserts this mapping. I swapped `response.accessToken` for `response.duration` here and the full client unit test suite still passed, since `testGetGuestUser` only checks the result type and `ChatClientGuestApiTests` mocks `ChatApi`. Both are required strings on the new model, so it is an easy one to get wrong. Could the device check you mentioned become an assertion, maybe a small success-case test next to `testGetGuestUser`?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/CreateGuestResponseParsingTest.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    fun `Deserialize guest response and map the embedded user with custom fields`() {
+        val response = parser.fromJson(GUEST_JSON, CreateGuestResponse::class.java)
+
+        response.accessToken shouldBeEqualTo "guest-token"
+        response.user.id shouldBeEqualTo "guest-123"
+        response.user.role shouldBeEqualTo "guest"
+        response.user.language shouldBeEqualTo "en"
+        response.user.custom shouldBeEqualTo mapOf("nickname" to "Ghost")
```
andremion, line 36: Small thing: `duration` is the only new field here without an assertion. Add `response.duration shouldBeEqualTo "12ms"`?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/CreateGuestResponseParsingTest.kt
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    companion object {
+        private const val GUEST_JSON =
+            """{
```
andremion, line 49: `ThreadParticipantParsingTest` uses a payload captured from the real endpoint. Any reason not to use the one you captured on device here? It would also cover the `name` round trip.

andremion (approved):
Looks good. One test gap and two nits inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Good catch, and the mutation was the convincing part. Added a success case next to `testGetGuestUser` that asserts the mapped `GuestUser`, and checked it bites: with `response.accessToken` swapped for `response.duration` it fails with `Expected: <the-access-token> but was: <12ms>`. The device check also now asserts the token is a JWT rather than just logging it.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/CreateGuestResponseParsingTest.kt
gpunto: Added, as `"12.65ms"` from the captured payload.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/CreateGuestResponseParsingTest.kt
gpunto: Done, re-probed to capture it. The hand-written one was wrong in small ways that matter: the server sends `"language": ""` rather than a locale, includes `teams` and `blocked_user_ids` as empty arrays, and orders `user` before `access_token`. I replaced the access token with a JWT-shaped placeholder rather than commit a real one, and shortened the guest id; the rest is verbatim. `name` is asserted now too, though note it never reaches `custom`, since the generated `UserResponse` declares it.

PR #6646, opened 2026-08-18
Skip mark-read only when the server has no message in the channel

gpunto:
### Goal

The mark-read guard added in #6643 skips whenever the last item in the list is the current user's own local-only message, which also suppresses legitimate mark-read in channels the server does have messages for. An `error` message is terminal, so a rejected or moderation-bounced send stays the last item and keeps those channels unread indefinitely.

Skip only when the server tracks none of the loaded messages, which is the case that produces the malformed read event. Read state leaves out our own not-yet-persisted messages, and silent and shadowed ones from anyone, so none of them can resolve a mark-read call.

v6 port of #6644, plus the silent and shadowed cases. Closes [AND-1395](https://linear.app/stream/issue/AND-1395/mark-read-fires-on-server-rejected-message-echoes-emitting-messageread)

### Implementation

- `MessageListController.markLastMessageReadInternal`: replace the "last item is our own local-only message" check with "no loaded message is one the server tracks", via a new `Message.isInServerReadState` predicate covering local-only, `silent` and `shadowed`.
- Reword the `isLocalOnly` KDoc: an ephemeral row is persisted server side, it is left out of message queries and read state.
- Tests: error and ephemeral cases now hold only our own local-only message, plus new cases for a silent-only channel, a shadowed-only channel, a silent message following a tracked one, and a server-side message preceding the echo.
- `randomMessage` randomises `silent`, which the gate now keys on, so the mark-read tests and the `nowMessage` helper pin it explicitly.

### Testing

- Unit tests as above, rerun to confirm they are deterministic now that the gate reads a randomised field.
- Verified on device against two frozen channels. With no server-side message the guard rejects and no read request is sent, and the channel read state stays untouched. With server-side messages the request goes out, the unread count clears, and the emitted event carries a populated `last_read_message_id`.

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -1724,13 +1724,10 @@ public class MessageListController(
         val messageText = message?.text
         logger.d { "[markLastMessageRead] cid: $cid, msgId($isInThread): $messageId, msgText: \"$messageText\"" }
 
-        // The server keeps our own local-only messages out of its read state, so marking read
-        // without any other message makes it emit message.read with no last_read_message_id.
+        // Marking read with nothing the server tracks makes it emit message.read with no
+        // last_read_message_id.
         val currentUserId = clientState.user.value?.id
-        val hasServerSideMessage = messageItems.any { item ->
-            !(item.message.isMine(currentUserId) && item.message.isLocalOnly())
-        }
-        if (!hasServerSideMessage) {
+        if (messageItems.none { it.message.isInServerReadState(currentUserId) }) {
             logger.v { "[markLastMessageRead] cid: $cid; rejected[$isInThread] (no server-side message)" }
             return
         }
@@ -1748,6 +1745,11 @@ public class MessageListController(
             markChannelAsRead()
         }
     }
+
+    // The server keeps our own local-only messages out of its read state, and silent and shadowed
+    // ones from anyone, so none of them can resolve a mark-read call.
+    private fun Message.isInServerReadState(currentUserId: String?): Boolean =
+        !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed
 
     private fun markChannelAsRead() {
         val (channelType, channelId) = cid.cidToTypeAndId()
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -33,6 +33,7 @@ import io.getstream.chat.android.client.utils.message.isDeleted
 import io.getstream.chat.android.client.utils.message.isError
 import io.getstream.chat.android.client.utils.message.isGiphy
 import io.getstream.chat.android.client.utils.message.isLocalOnly
+import io.getstream.chat.android.client.utils.message.isMine
 import io.getstream.chat.android.client.utils.message.isModerationBounce
 import io.getstream.chat.android.client.utils.message.isModerationError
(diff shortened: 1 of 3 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -1717,19 +1718,20 @@ public class MessageListController(
...
 
-        // Marking read while our own local-only message is at the bottom makes a channel the
-        // server sees as empty emit message.read with no last_read_message_id.
+        // The server keeps our own local-only messages out of its read state, so marking read
+        // without any other message makes it emit message.read with no last_read_message_id.
         val currentUserId = clientState.user.value?.id
-        if (message != null && message.user.id == currentUserId && message.isLocalOnly()) {
-            logger.v { "[markLastMessageRead] cid: $cid; rejected[$isInThread] (own local-only): $messageId" }
+        val hasServerSideMessage = messageItems.any { item ->
+            !(item.message.isMine(currentUserId) && item.message.isLocalOnly())
```
andremion: Not blocking, just noting it while this is fresh: the gate treats any message that is not our own local-only one as proof the server has something to resolve, but the server also leaves silent and shadowed messages out of its read state. Both fields are on `Message` and neither is filtered out of `messageItems`, so a channel whose only message is a silent system message (the text passed to `addMembers`, say) still reaches `markRead` with nothing to resolve, and the same goes for a shadow-banned user's own message. The old check let both through too, so nothing got worse here. Any appetite for adding silent and shadowed to the predicate, or would you rather keep that separate?

andremion (approved):
Looks good. One non-blocking note inline.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -1717,19 +1718,20 @@ public class MessageListController(
...
+        }
```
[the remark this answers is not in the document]
andremion: For what it's worth, `isMine(null)` returned false before this PR too, so a null user already let the gate pass. Doesn't look like a regression from this change.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
gpunto: Added in 4d5430e. The gate now asks whether the server tracks a message rather than whether it is our own local-only one:

```kotlin
// The server keeps our own local-only messages out of its read state, and silent and shadowed
// ones from anyone, so none of them can resolve a mark-read call.
private fun Message.isInServerReadState(currentUserId: String?): Boolean =
    !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed
```

`silent` and `shadowed` apply regardless of sender, unlike the local-only condition, since the backend excludes them for everyone.

What convinced me it is free rather than a trade-off: the docs define `silent` as "do not increment unread counts or mark channels as unread", so a channel holding only silent messages has an unread count of zero by definition and nothing for a cursor to advance to. Shadowed is the same shape, "their messages are only visible to themselves". A channel that mixes a normal message with a trailing silent one still marks read, since the normal message qualifies.

Two things worth knowing:

- `randomMessage` randomises `silent`, so the gate reading that field made the peer-message tests flaky, and one existing test (`When system message arrives, markRead should be invoked for that channel`) failed immediately through the `nowMessage` helper. Both are pinned now, and I reran the class to confirm determinism.
- I also went looking for a backend defect here and did not find one: excluding silent messages from read state is intentional per the docs, so there is nothing to file on that side. Only the client-side call needed fixing.

gpunto: Forgot to mention: this only covers v6. #6644 already merged, so develop needs the same change as a separate PR, and we will take care of that too.

PR #6647, opened 2026-08-18
Migrate the queryBlockedUsers response to the generated GetBlockedUsersResponse model

gpunto:
### Goal

Migrate the `queryBlockedUsers` response to the generated `GetBlockedUsersResponse` model.

### Implementation

- Replace the hand-written `QueryBlockedUsersResponse` and `DownstreamUserBlockDto` with the generated
  `GetBlockedUsersResponse` and `BlockedUserResponse`.
- Map `BlockedUserResponse` to the domain `UserBlock`.

### Testing

- `BlockedUsersParsingTest` covers the response and the nested user objects. `DomainMappingTest` covers the
  mapping to `UserBlock`.
- Device-probed on the wire: blocked two users, read the list back with `queryBlockedUsers`, then unblocked
  both and confirmed the list returned to its starting state. The response carries `blocked_user` and `user`
  as complete `UserResponse` objects, which the generated model requires non-null, so the parse depends on
  them being sent in full.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blocked-user list parsing and mapping to correctly handle the current API response format.
  * Preserved blocker, blocked-user, and creation-time details when loading blocked users.
  * Added support for the response duration field.

* **Tests**
  * Added coverage for parsing and converting real blocked-user response data.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/BlockedUserResponse.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/BlockedUserResponse.kt
@@ -1,0 +1,47 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:Suppress(
+    "ArrayInDataClass",
+    "EnumEntryName",
+    "RemoveRedundantQualifierName",
+    "UnusedImport",
+)
+
+package io.getstream.chat.android.network.models
+
+import com.squareup.moshi.Json
+
+/**
+ *
+ */
+@com.squareup.moshi.JsonClass(generateAdapter = true)
+internal data class BlockedUserResponse(
+    @Json(name = "blocked_user_id")
+    internal val blockedUserId: String,
+
+    @Json(name = "created_at")
+    internal val createdAt: java.util.Date,
+
+    @Json(name = "user_id")
+    internal val userId: String,
+
+    @Json(name = "blocked_user")
+    internal val blockedUser: io.getstream.chat.android.network.models.UserResponse,
+
+    @Json(name = "user")
+    internal val user: io.getstream.chat.android.network.models.UserResponse,
+)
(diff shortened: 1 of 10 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/BlockedUserResponse.kt
```diff
@@ -0,0 +1,47 @@
...
+    internal val blockedUserId: String,
+
+    @Json(name = "created_at")
+    internal val createdAt: java.util.Date,
+
+    @Json(name = "user_id")
+    internal val userId: String,
+
+    @Json(name = "blocked_user")
+    internal val blockedUser: io.getstream.chat.android.network.models.UserResponse,
```
andremion, line 43: The mock server sends only `user_id`, `blocked_user_id` and `created_at` per block, so with these two required the response no longer parses there (`Required value 'blockedUser' (JSON name 'blocked_user') missing`). Nothing calls `queryBlockedUsers` outside the client module today, so nothing breaks now, but the first E2E test on blocking would fail there. Worth a small mock PR to add the two user objects?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/BlockedUsersParsingTest.kt
```diff
@@ -0,0 +1,77 @@
...
+          ]
+        }"""
+
+    @Test
+    fun `deserializes the blocked-users wire shape and maps it to UserBlock`() {
+        val dto = parser.fromJson(json, GetBlockedUsersResponse::class.java)
+
+        val blocks = with(domainMapping) { dto.blocks.toDomain() }
+
+        assertEquals(
```
andremion, line 81: Small thing: the only assertion here is the mapped `UserBlock`, which `DomainMappingTest` already covers with the same fields. `birthland` and the empty `language` are just setup at the moment. Any reason not to assert `user.custom` and `user.language` as well, the way `CreateGuestResponseParsingTest` does?

andremion (approved):
Looks good. One thing about the mock server and one test nit inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/BlockedUserResponse.kt
gpunto: Good catch, confirmed: the stored block carries only the three ids, and the generated model requires both user objects non-null. Including the fix in [#59](https://github.com/GetStream/stream-chat-test-mock-server/pull/59).

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/BlockedUsersParsingTest.kt
gpunto: Fair, the fixture was setting up more than it asserted. It now checks `duration`, both nested ids, `user.custom` holding `birthland`, and both `language` values including the empty one, so the parse is pinned rather than just the mapping. Folded into the existing commit.

andremion (approved):

PR #6649, opened 2026-08-19
e2e: Cover poll creation and voting

andremion:
### Goal

Cover polls end to end in the Compose sample: creating a poll from the attachment picker, voting and removing a vote in the message list, a participant vote arriving over the websocket, and ending a poll.

Requires the mock server poll support from GetStream/stream-chat-test-mock-server#62. The workflow runs linked below pass `mock_server_branch` until it merges.

Resolves AND-1330

### Implementation

- New `PollsTests` with five tests, plus the poll page objects (`CreatePollPage` and a `Poll` section in `MessageListPage`), the `UserRobot` actions (`createPoll`, `castPollVote`, `removePollVote`, `openPollResults`, `endPoll`) and `UserRobotPollAsserts`.
- `ParticipantRobot` gains `castPollVote(option)` and `addPollAnswer(answer)`, backed by the new mock robot route `participant/poll_vote`.
- The poll UI had no test tags, so the SDK now exposes them: the creation dialog (`Stream_PollQuestionInput`, `Stream_PollOptionInput`, `Stream_PollAddOptionButton`, `Stream_PollCreateButton`), the poll bubble (`Stream_PollOptionVotingRow` plus one tag per action button), the end-poll confirmation (`Stream_PollEndConfirmButton`) and the picker button (`Stream_AttachmentPickerCreatePollButton`). `FullscreenDialog` and the end-poll `AlertDialog` render in their own windows, so they now set `testTagsAsResourceId`, the same way `StreamModalBottomSheet` already does. No behavior change; `PollButtons` had its answers branch extracted into a private `PollAnswersButton` to stay under the detekt method length limit.
- Selector notes, learned from local device runs: the text inputs are selected as the `EditText` inside the tagged container, because setting text on any other node is silently ignored. A vote is asserted through the option row's toggle state, which waits out the vote round trip. The default `PollPickerMode` auto-opens the creation dialog when the Polls tab is selected, so the picker button is never tapped.
- The participant vote test waits for the poll message delivery before the vote (a vote for a message the server does not have yet is rejected), and asserts the vote count in the message list before opening the results dialog, because the dialog renders the poll as selected at tap time.

### Testing

- All five `PollsTests` are green locally against the mock server PR branch (API 35 emulator, per-test data isolation, five out of five), re-run after rebasing onto the latest develop.
- `spotlessCheck`, `detekt`, `apiCheck` (no public API changes, the touched composables are internal or private) and `:stream-chat-android-compose:verifyPaparazziDebug` all pass. The tags do not change rendering.
- E2E workflow runs on this branch with the mock server branch: [PollsTests, 3 batches](https://github.com/GetStream/stream-chat-android/actions/runs/32230522574) and the [full API matrix](https://github.com/GetStream/stream-chat-android/actions/runs/32230526593).

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for creating polls, displaying poll messages, voting, removing votes, viewing results, and ending polls.
  * Added test support for creating polls and validating options, vote counts, results, and closed states.
* **Chores**
  * Improved UI test accessibility for poll controls, inputs, dialogs, and voting rows.
  * Added participant poll interactions, including voting and submitting answers.


andremion:
@coderabbit review

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/components/FullscreenDialogTest.kt
```diff
@@ -0,0 +1,51 @@
...
+        composeTestRule.setContent {
+            FullscreenDialog(onDismissRequest = {}) {
+                Text(
+                    text = "dialog content",
+                    modifier = Modifier.testTag("DialogContent"),
+                )
+            }
+        }
+
+        composeTestRule.onNodeWithTag("DialogContent").assertIsDisplayed()
```
gpunto: This assertion passes with or without the change under test — content already rendered inside the dialog before the `semantics { testTagsAsResourceId = true }` wrapper was added. So deleting that block again would leave this test green and only show up as an E2E failure later. Asserting the tag resolves as a resource id on the dialog subtree would actually pin it.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/components/messages/PollMessageContentInteractionsTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    fun prepare() {
+        whenever(mockClientState.user) doReturn MutableStateFlow(randomUser())
+        whenever(mockClientState.connectionState) doReturn MutableStateFlow(ConnectionState.Connected)
+    }
+
+    // A dialog window opened by a later recomposition does not register with the compose test
+    // rule under Robolectric, so the end-poll flow is split: this test pins that the End Poll
+    // button never closes the poll directly, and the dialog itself is composed directly below.
+    @Test
+    @UiThread
```
gpunto: `@UiThread` is a lint/documentation annotation with no runtime effect — JUnit won't act on it, and the Compose test rule already handles dispatch. Same on lines 89 and 102. Harmless, but it reads like it's doing something.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/components/FullscreenDialogTest.kt
gpunto, line 53: This passes with the change it covers removed — I deleted the `semantics { testTagsAsResourceId = true }` block and the test still reported `tests=1 failures=0`. `onNodeWithTag` reads the semantics tree directly, so it never needed the resource-id export.

Either assert the tag resolves as a resource id on the dialog subtree, or drop this and leave the behaviour to E2E.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
```diff
@@ -42,6 +43,7 @@ open class MessageListPage {
             val filesTab get() = By.res("Stream_AttachmentPickerFilesTab")
             val mediaCaptureTab get() = By.res("Stream_AttachmentPickerMediaCaptureTab")
             val pollsTab get() = By.res("Stream_AttachmentPickerPollsTab")
+            val createPollButton get() = By.res("Stream_AttachmentPickerCreatePollButton")
```
gpunto: `createPollButton` has no caller and can't get one — the polls tab auto-opens the dialog (`autoShowCreateDialog = true` by default, not overridden in the sample), so the picker button is never tapped. Its production tag in `AttachmentPollPicker.kt` is therefore unreachable too.

Also unconsumed: `CreatePollPage.backButton` and four of the new `PollMessageContent` tags (`SeeAllOptions`, `SuggestOption`, `ViewComments`, `AddComment`). Take or leave, but I'd drop the picker one.

stream-chat-android-e2e-test/src/main/kotlin/io/getstream/chat/android/e2e/test/robots/ParticipantRobot.kt
```diff
@@ -208,6 +209,28 @@ public class ParticipantRobot(
...
+        mockServer.postRequest("participant/poll_vote?option=${option.urlEncoded()}")
+        return this
+    }
+
+    /**
+     * Adds an answer (comment) to the poll of the newest poll message.
+     *
+     * @param answer The answer text.
+     */
+    public fun addPollAnswer(answer: String): ParticipantRobot {
```
gpunto, line 227: No caller in this PR. The contract is right (route and `answer` param match the mock server), just untested — fine to keep if the comments test is close behind.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/components/messages/PollMessageContentInteractionsTest.kt
gpunto: `@UiThread` is `RetentionPolicy.CLASS`, so JUnit can't see it, and the Compose rule already handles dispatch. Same on 89 and 102.

gpunto (approved):

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/components/FullscreenDialogTest.kt
andremion: Right, the assertion never depended on the flag. Fixed in 5f3b7685c0af808c8de517892ee745b068d1c4d8: a second test asserts the dialog content has an ancestor with SemanticsPropertiesAndroid.TestTagsAsResourceId set, and I mutation-tested it (removing the semantics block fails the test, restoring it passes).

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/MessageListPage.kt
andremion: Dropped the picker selector and its production tag in 5f3b7685c0af808c8de517892ee745b068d1c4d8, you are right that the auto-open makes it unreachable. I kept backButton and the four content tags: those elements are reachable, and the comments and suggest-option tests that consume them are the natural follow-up to this suite.

stream-chat-android-e2e-test/src/main/kotlin/io/getstream/chat/android/e2e/test/robots/ParticipantRobot.kt
andremion: Keeping it for that reason: the mock route is already on main and verified, and the poll comments test is the next addition to PollsTests.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/components/messages/PollMessageContentInteractionsTest.kt
andremion: Confirmed, the annotation is @Retention(BINARY) so JUnit never sees it; the tests pass because the dialog is composed in the initial composition. Removed all three in 5f3b7685c0af808c8de517892ee745b068d1c4d8.

PR #6650, opened 2026-08-19
e2e: Cover channel search, direct message blocking and the moderation state

andremion:
### Goal

Cover the QA cases that needed mock server changes first: channel search (5934) and blocking a user in a direct message channel (6070, 6072). Also restore the six mute and block state tests that were descoped from #6610 (11563, 11566, 6071, 11575, 11560, 11567), since the mock server now keeps that state stable.

Requires the mock server support from GetStream/stream-chat-test-mock-server#63. This PR's own e2e checks run against the mock server main, so they will fail on the new tests until that PR merges; the workflow run linked below passes `mock_server_branch` instead.

Resolves AND-1359

### Implementation

- The six descoped tests come back by reverting the descope commit, unchanged.
- The Compose sample gains a `Channel search` flag: a switch on the custom login screen, stored in `CustomSettings`, that makes `ChannelsActivity` and `ChatsActivity` configure the channel list with `SearchMode.Channels` instead of `SearchMode.Messages`. The e2e flavor's `StartupActivity` assigns the flag from a `CHANNEL_SEARCH` launch extra on every start, so a value left by a previous test cannot leak into the next one. `StreamTestCase` exposes it as `useChannelSearch`.
- `BackendRobot.generateChannels` gains `channelNames` (searchable names for the seeded channels) and `withDirectMessageChannel` (a two-member channel with a member-based cid and no name, shown under the participant's name).
- New `ChannelSearchTests` (5934): searching by name shows only the matching channel. New robot pieces: `searchForChannel`, `openChannel(channelName)`, `assertChannelCount` and `assertChannelWithName`.
- Two new tests in `ModerationTests`: blocking the participant in the direct message channel offers the unblock option (6070), and hides the channel from the channel list until the unblock, asserted from the channel list in both directions (6072). The direct message channel uses seeded messages because `ParticipantRobot` posts to the group channel; seeded messages alternate authors starting with the user, so the message with text "2" belongs to the participant.

### Testing

- All tests were run locally on an API 35 emulator against the mock server PR branch, one method at a time with data cleared between runs: the 3 new tests, the 6 restored tests, and a regression set (both `SearchTests`, the giphy ephemeral cancel test, two `ChannelListTests` preview tests, and all 11 `AuthTests` for the token error health check paths). All pass with no retry attempts.
- `spotlessCheck`, `detekt` and `apiCheck` pass.
- E2E full API matrix run on this branch with the mock server branch: [e2e-test-cron run](https://github.com/GetStream/stream-chat-android/actions/runs/32261908243).

## Summary by CodeRabbit

- **New Features**
  - Added optional channel search, allowing users to find channels by name from the channel list.
  - Added a channel search toggle to the custom login settings.
  - Expanded moderation actions to support muting/unmuting and blocking/unblocking message authors.
  - Added channel swipe actions for muting and unmuting channels.

- **Tests**
  - Added end-to-end coverage for channel search, channel muting, moderation actions, and direct-message behavior.


andremion:
@coderabbit review

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
```diff
@@ -552,3 +553,27 @@ fun UserRobot.assertFlagMessageDialog(isDisplayed: Boolean): UserRobot {
...
+
+/**
+ * Opens the message menu of the message with [messageText] and asserts the block option it offers
+ * for the author. The sample has no blocked users list, so the option label is the only place the
+ * block state is visible.
+ */
+fun UserRobot.assertBlockMessageAuthorOption(messageText: String, isAuthorBlocked: Boolean): UserRobot {
+    val expectedOption = if (isAuthorBlocked) ContextMenu.unblock else ContextMenu.block
+    openContextMenuWithOption(messageText, expectedOption)
+    assertTrue(expectedOption.isDisplayed())
```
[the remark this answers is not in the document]
andremion: d2c2ca458b7640811266b62e94062a99e93eaba9

stream-chat-android-compose-sample/src/e2e/java/io/getstream/chat/android/compose/sample/ui/StartupActivity.kt
```diff
@@ -47,6 +47,8 @@ class StartupActivity : AppCompatActivity() {
                 ChatClient.instance().disconnect(flushPersistence = true).await()
                 ChatHelper.initializeSdk(applicationContext, PredefinedUserCredentials.API_KEY, baseUrl)
                 customSettings().isComposerLinkPreviewEnabled = true
+                // Always assigned, so a flag left by a previous test cannot leak into this one.
+                customSettings().isChannelSearchEnabled = intent.getBooleanExtra("CHANNEL_SEARCH", false)
```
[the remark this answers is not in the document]
andremion: d2c2ca458b7640811266b62e94062a99e93eaba9

stream-chat-android-e2e-test/src/main/kotlin/io/getstream/chat/android/e2e/test/robots/BackendRobot.kt
```diff
@@ -32,14 +32,24 @@ public class BackendRobot(
...
+        } else {
+            ""
+        }
+        val dmQueryParam = if (withDirectMessageChannel) "dm=true&" else ""
         mockServer.postRequest(
             "mock?" +
                 messagesTextQueryParam +
                 repliesTextQueryParam +
+                channelNamesQueryParam +
+                dmQueryParam +
```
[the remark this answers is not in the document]
andremion: d2c2ca458b7640811266b62e94062a99e93eaba9

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
```diff
@@ -202,6 +209,24 @@ class UserRobot {
...
+            openContextMenu(text)
+            if (option.waitDisplayed(timeOutMillis = 5.seconds)) {
+                return this
+            }
+            if (attempt < contextMenuOpenAttempts - 1) {
+                pressBack()
+            }
+        }
+        return this
+    }
```
gpunto: The options aren't actually frozen while the menu is open: `ChannelScreen` recomputes `defaultMessageOptionsState` from `user.collectAsState()`, and `ContextualMenu` builds the tag as `Stream_ContextMenu_$label` — so the tag flips in place. State arrives promptly too (mute via `notification.mutes_updated`; block with no event at all, `ChatClient.blockUser` updates local state in the HTTP callback). Since opening re-fetches nothing, a single open + longer `waitDisplayed` should be strictly better.

Two side effects of the loop: `pressBack()` assumes the menu is open, so if a `longPress` misses, `ChannelScreen.backAction` falls through to `onBackPressed()` and leaves the message list — the next `openContextMenu` then dies on the message lookup. And exhausting all attempts returns `this`, indistinguishable from success.

Possibly a leftover from before the mock's `live_own_user_state` fix, when the health check's `setUser` wiped the state. Did you need the reopen for something specific?

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/ChannelSearchTests.kt
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    fun test_userSearchesForChannel() {
+        step("GIVEN channels exist, one with a searchable name") {
+            backendRobot.generateChannels(channelsCount = 3, channelNames = listOf(channelName))
+        }
+        step("AND user logs in") {
+            userRobot
+                .login()
+                .waitForChannelListToLoad()
+        }
```
gpunto: Can pass without the search having filtered: `waitForChannelListToLoad()` returns on the first `Stream_ChannelItem`, `waitForCount(1)` returns on the first poll where count is 1, and "Blue Team" is already the top row (newest `last_message_at`). Adding `.assertChannelCount(3)` here makes the 3 -> 1 transition the actual assertion.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
```diff
@@ -552,3 +553,31 @@ fun UserRobot.assertFlagMessageDialog(isDisplayed: Boolean): UserRobot {
...
+ * mute state is visible.
+ */
+fun UserRobot.assertMuteMessageAuthorOption(messageText: String, isAuthorMuted: Boolean): UserRobot {
+    val expectedOption = if (isAuthorMuted) ContextMenu.unmuteUser else ContextMenu.muteUser
+    val oppositeOption = if (isAuthorMuted) ContextMenu.muteUser else ContextMenu.unmuteUser
+    openContextMenuWithOption(messageText, expectedOption)
+    assertTrue(expectedOption.isDisplayed())
+    assertFalse(oppositeOption.isDisplayed())
+    return this
+}
```
gpunto: These open the menu and never dismiss it, so they mutate state despite reading as asserts. Harmless today since every call is a terminal step, but a trailing `pressBack()` would make them chainable.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/ChannelListPage.kt
```diff
@@ -61,6 +62,10 @@ class ChannelListPage {
             companion object {
                 val avatar = By.res("Stream_ChannelAvatar")
                 val name = By.res("Stream_ChannelName")
+
+                // A fresh selector per call: BySelector chaining mutates the instance,
+                // so narrowing the shared `name` selector would poison later uses.
+                fun name(text: String): BySelector = By.res("Stream_ChannelName").text(text)
```
gpunto: `Stream_ChannelName` is also the message-list header's tag, so this can match outside the channel list. I couldn't find a live failure path in this PR — hardening only. `.hasAncestor(By.res("Stream_ChannelItem"))` closes it.

Side note: `MessageListPage` already avoids the mutation hazard with `val text get() = ...`. A `get()` here would cover the plain `name` val too and let the comment go.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
```diff
@@ -446,6 +505,12 @@ class UserRobot {
         return this
     }
 
+    /** Same input as [searchForMessage]; what it searches depends on the app's search mode. */
+    fun searchForChannel(text: String): UserRobot {
+        ChannelListPage.Header.searchField.waitToAppear().typeText(text)
+        return this
+    }
```
gpunto: Identical to `searchForMessage`. `fun searchForChannel(text: String) = searchForMessage(text)` keeps them from diverging.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
gpunto: **Suggestion: drop the loop — open once and wait longer.**

The reopen is there to catch the label flipping Mute -> Unmute, but it flips on its own without reopening: `ChannelScreen` recomputes the options from `user.collectAsState()`, and `ContextualMenu` derives the tag from the label (`Stream_ContextMenu_$label`), so an open menu re-tags itself as soon as the state lands. And it lands quickly — mute via `notification.mutes_updated`, block with no event at all since `ChatClient.blockUser` updates local state in the HTTP callback.

Opening the menu re-fetches nothing, so each retry only throws away up to 5s.

It also adds a risk: `pressBack()` assumes the menu is open. If a `longPress` misses, `ChannelScreen.backAction` sees no overlay and calls `onBackPressed()`, leaving the message list — the next `openContextMenu` then fails on the message lookup instead of on the assertion.

Possibly a leftover from before the mock's `live_own_user_state` fix, when the health check's `setUser` wiped the state?

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/ChannelSearchTests.kt
gpunto, line 50: Can pass without the search having filtered: `waitForChannelListToLoad()` returns on the first `Stream_ChannelItem`, `waitForCount(1)` returns on the first poll where count is 1, and "Blue Team" is already the top row (newest `last_message_at`). Adding `.assertChannelCount(3)` here makes the 3 -> 1 transition the actual assertion.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
gpunto, line 570: Worth a trailing `pressBack()`: the one inside `openContextMenuWithOption` only runs when the option *isn't* found, so on the success path these return with the menu still open. Harmless today since every call is a terminal step, but it makes them non-chainable.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/ChannelListPage.kt
gpunto: `MessageListPage` handles this same mutation hazard with a getter — `val text get() = By.res(...)`. Doing that here would cover the plain `name` val too, and the comment wouldn't be needed.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
gpunto: These two are the same function, and what gets searched is decided by the app's `SearchMode` rather than by the robot — so the names imply a specialisation that doesn't exist. Worth collapsing into a single `search(text)`.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
```diff
@@ -636,3 +701,5 @@ class UserRobot {
         return this
     }
 }
+
+private const val contextMenuOpenAttempts = 3
```
gpunto: Nit: `CONTEXT_MENU_OPEN_ATTEMPTS`, cf. `PUSH_RECEIVER_CLASS`. Detekt's `constantPattern` requires it; it only passes because this source set isn't scanned. Moot if the loop goes.

gpunto (approved):
Looks good, nothing blocking. One substantive comment (the context-menu retry loop), rest are nits.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
andremion: You are right, and it also holds up empirically: with the helper opening once and waiting 15s, all six moderation state tests and both direct message tests pass locally on the first attempt. The reopen was indeed a leftover from the static health check era. Fixed in 36da6dccd88d0ca7f8862339a1c8765ca8a5a05e.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/ChannelSearchTests.kt
andremion: Good catch, the count could momentarily be 1 while the list composes. The login step now asserts 3 channels before searching, so the 3 to 1 transition is the assertion. Fixed in 36da6dccd88d0ca7f8862339a1c8765ca8a5a05e.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobotMessageListAsserts.kt
andremion: Added the trailing pressBack in both asserts. Fixed in 36da6dccd88d0ca7f8862339a1c8765ca8a5a05e.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/pages/ChannelListPage.kt
andremion: Done, both are getters now and the comment is gone. Fixed in 36da6dccd88d0ca7f8862339a1c8765ca8a5a05e.

stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/robots/UserRobot.kt
andremion: Collapsed into a single search(text). Fixed in 36da6dccd88d0ca7f8862339a1c8765ca8a5a05e.

PR #6651, opened 2026-08-19
Debounce short search queries for longer

gpunto:
### Goal

Message search, the XML search input and the add-members user search send a query as typed, from the first character. Queries of 1-2 characters match a large portion of the data set, so they are the slowest ones to serve and some of them time out, while they are usually just a step towards the query the user is after. Debounce them for at least 500ms instead of the configured 300ms, matching iOS ([stream-chat-swift#4198](https://github.com/GetStream/stream-chat-swift/pull/4198)).

Closes [AND-1409](https://linear.app/stream/issue/AND-1409/adaptive-search-debounce-based-on-query-length)

### Implementation

- Add `SearchDebounce` in ui-common, marked `@InternalStreamChatApi` since the thresholds are not meant to be configured by integrators: 500ms up to 2 characters, the configured debounce from 3 on, and never shorter than the configured one. No public API changes, the API dumps are untouched.
- `Debouncer`: internal `submit`/`submitSuspendable` overloads taking the debounce period for a single piece of work.
- Apply it to the Compose message search in `ChannelListViewModel`, the debounced input listener of `SearchInputView`, and the user search in `AddMembersViewController`.
- `SearchInputView.clear()` now cancels a debounce still pending from the last keystroke, which notified the listener with the query being cleared and re-ran the search the user had just dismissed.

Channel search is left alone: it only queries from 3 characters on, so it never sends the short autocomplete `queryChannels` this targets. Mention autocomplete is left alone too, as it queries the members of a single channel.

### Testing

- `SearchDebounceTest` covers the thresholds, including a configured debounce longer than 500ms winning for short queries.
- `ChannelListViewModelTest`: a 2 character message search reaches `searchMessages` only after 500ms, a 3 character one after 300ms, and a `searchDebounceMs` above 500ms is honoured for short queries.
- `AddMembersViewControllerTest`: a 2 character query starts no search before 500ms. Reverting the controller to the flat 300ms debounce fails this test.
- `DebouncerTest`: the new overloads debounce by the period passed per submission, and a later submission cancels pending work.
- The `SearchInputView.clear()` fix has no test: ui-components has no Robolectric setup in its unit tests, so there is no harness to drive the view's `TextWatcher`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message and member search responsiveness with query-sensitive debounce timing.
  * Short searches of one or two characters now wait at least 500 ms before triggering results, reducing unnecessary searches.
  * Whitespace-padded short queries follow the same behavior.
  * Clearing a search field now cancels pending searches and immediately resets results.
* **Enhancements**
  * Longer or regular queries continue using the configured debounce interval.


gpunto:
@coderabbitai review

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
```diff
@@ -805,8 +809,9 @@ public class ChannelListViewModel internal constructor(
     }
 
     private fun handleSearchQuery(query: String) {
-        logger.d { "[handleSearchQuery] query: '$query'" }
-        searchDebouncer.submitSuspendable {
+        val debounceMs = SearchDebounce.debounceMsFor(query, searchDebounceMs)
+        logger.d { "[handleSearchQuery] query: '$query', debounceMs: $debounceMs" }
+        searchDebouncer.submitSuspendable(debounceMs) {
```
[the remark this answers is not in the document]
gpunto: We already opt in at the module level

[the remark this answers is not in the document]
gpunto: Confirming for the record: no change made here, and the annotations were removed rather than added.

`core`, `ui-common`, `compose` and `ui-components` all pass `-opt-in=io.getstream.chat.android.core.internal.InternalStreamChatApi` in their `build.gradle.kts`, so per-file `@OptIn` is inert in every module this PR touches. An earlier revision of the branch did carry those annotations on `ChannelListViewModel`, `SearchInputView` and the three test classes; `003a88a445` drops the last of them.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModelTest.kt
```diff
@@ -594,6 +597,82 @@ internal class ChannelListViewModelTest {
...
+                .get(this)
+
+            viewModel.setSearchQuery(SearchQuery.Messages("ab"))
+            advanceTimeBy(SearchDebounce.SHORT_QUERY_DEBOUNCE_MS + 50)
+
+            verify(chatClient, never()).searchMessages(any(), any(), anyOrNull(), anyOrNull(), anyOrNull(), anyOrNull())
+
+            advanceTimeBy(searchDebounceMs)
+
+            verify(chatClient).searchMessages(any(), any(), anyOrNull(), anyOrNull(), anyOrNull(), anyOrNull())
```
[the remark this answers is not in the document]
gpunto: Good catch, addressed in `75bc0b6e91`. The assertions did only pin a window (nothing at 350ms, fired by 850ms), so a 400ms debounce would have passed.

Each interval is now pinned exactly. `advanceTimeBy` runs tasks scheduled strictly before the target tick, so advancing to the interval and asserting no work proves the debounce is not shorter, and `runCurrent()` then executes the work scheduled at that tick:

```kotlin
advanceTimeBy(SearchDebounce.SHORT_QUERY_DEBOUNCE_MS)
verify(chatClient, never()).searchMessages(...)
runCurrent()
verify(chatClient).searchMessages(...)
```

Applied to all three files listed: the short, default and configured-long message-search tests in `ChannelListViewModelTest`, both custom-interval overload tests plus the replacement-submission test in `DebouncerTest` (`delay(499)` then `delay(1)`, since that file drives virtual time with `delay`), and both short-query tests in `AddMembersViewControllerTest`.

Both halves were mutation checked: adding 100ms to the short-query branch of `SearchDebounce.debounceMsFor` fails the post-`runCurrent()` assertion, and collapsing that branch to the configured debounce fails the pre-`runCurrent()` one.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/AddMembersViewControllerTest.kt
```diff
@@ -166,6 +168,54 @@ internal class AddMembersViewControllerTest {
...
+            // one character search.
+            sut.onViewAction(AddMembersViewAction.QueryChanged("a  "))
+            skipItems(1) // Skip the query-only state update
+
+            advanceTimeBy(SearchDebounce.SHORT_QUERY_DEBOUNCE_MS - 50)
+            expectNoEvents() // Still debounced, no search started
+
+            advanceTimeBy(100)
+            assertTrue(awaitItem().isLoading) // Search triggered
+            assertEquals(users, awaitItem().searchResult)
```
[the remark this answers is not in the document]
gpunto: Valid, addressed in `f5c32b4c49`. The test pinned only the debounce period, so the trim on the outbound request was unasserted: the fixture answers `queryUsers` by call order, so a request carrying `"a  "` would have passed.

The padded-query test now captures the request and asserts the autocomplete filter values:

```kotlin
val requestCaptor = argumentCaptor<QueryUsersRequest>()
verify(fixture.chatClient, times(2)).queryUsers(requestCaptor.capture())
val filter = requestCaptor.secondValue.filter as OrFilterObject
assertEquals(setOf("a"), filter.filterObjects.map { (it as AutocompleteFilterObject).value }.toSet())
```

That closes the loop on the fix, since the debounce is resolved on the trimmed query precisely because the request is trimmed. Mutation checked: dropping `.trim()` in `searchUsers` now fails this test, while the timing assertions alone still passed it.

andremion (commented):
Looks good. A few notes inline, the first one is the only one I would act on.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/search/SearchInputView.kt
```diff
@@ -173,6 +174,9 @@ public class SearchInputView : FrameLayout {
             return false
         }
 
+        // A debounce pending from the last keystroke would notify with the query being cleared.
+        inputDebouncer.cancelLastDebounce()
```
andremion, line 178: I think this one is testable. ui-components has no Robolectric, but the Paparazzi harness in `PaparazziViewTest` builds views in plain unit tests (`ChannelListHeaderViewTest` does it), and `DispatcherProvider.set` gives you control over the `Debouncer`'s dispatcher.

I tried it: build the view with `paparazzi.context`, set a debounced listener, `setQuery("abc")`, advance 100ms, `clear()`, advance to idle. It passes, and with `cancelLastDebounce()` removed the listener gets `["", "abc"]`. Worth adding?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/search/SearchInputView.kt
```diff
@@ -116,7 +117,7 @@ public class SearchInputView : FrameLayout {
 
             val newQuery = query
             continuousInputChangedListener?.onInputChanged(newQuery)
-            inputDebouncer.submit {
+            inputDebouncer.submit(SearchDebounce.debounceMsFor(newQuery, TYPING_DEBOUNCE_MS)) {
```
andremion, line 120: Small thing: this changes what the debounced listener promises for every integrator wiring it to a search, and there is no way to opt out. The KDoc on `setDebouncedInputChangedListener` (line 210) still only says the input has to be stable "for a short while". Could you mention the longer wait for 1-2 characters there, like you did for `searchDebounceMs` on the Compose side?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
```diff
@@ -805,8 +809,9 @@ public class ChannelListViewModel internal constructor(
     }
 
     private fun handleSearchQuery(query: String) {
-        logger.d { "[handleSearchQuery] query: '$query'" }
-        searchDebouncer.submitSuspendable {
+        val debounceMs = SearchDebounce.debounceMsFor(query, searchDebounceMs)
```
andremion, line 812: Any reason not to trim here too? The query goes into `Filters.autocomplete("text", ...)` untrimmed, so this is consistent, but `"a  "` has length 3 and takes the 300ms path while the search term is one character. Same case you fixed for the user search in db8e6ddd7874f4948686a4ef2e8169262c4b4774. Happy to be told the whitespace case is not worth covering.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
```diff
@@ -143,7 +144,8 @@ public class ChannelListViewModel internal constructor(
      * @param messageLimit How many messages are fetched for each channel item when loading channels.
      * When `null`, the server-side default is used.
      * @param chatEventHandlerFactory The instance of [ChatEventHandlerFactory] used to create [ChatEventHandler].
-     * @param searchDebounceMs The debounce time for search queries.
+     * @param searchDebounceMs The debounce time for search queries. Message search queries of 1-2 characters
```
andremion, line 147: Nit: the same param on the primary constructor (line 98) still has the old one-line text. Worth updating for consistency?

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/AddMembersViewControllerTest.kt
```diff
@@ -166,6 +175,63 @@ internal class AddMembersViewControllerTest {
...
+            .givenQueryUsers(users = users) // search for "Al"
+            .get(backgroundScope)
+
+        sut.state.test {
+            skipItems(2) // Skip initial state and empty-query search result
+
+            sut.onViewAction(AddMembersViewAction.QueryChanged("Al"))
+            skipItems(1) // Skip the query-only state update
+
+            advanceTimeBy(SearchDebounce.SHORT_QUERY_DEBOUNCE_MS)
```
andremion, line 192: `advanceTimeBy` and `runCurrent` need `ExperimentalCoroutinesApi`, and the file has no `@OptIn` for it, so the build now logs four warnings here (lines 192, 195, 217, 220). The Compose test class carries `@ExperimentalCoroutinesApi`. Add the same one here?

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/search/SearchInputView.kt
gpunto: Added in `29c09b3b9d`, thanks for the pointer to that harness. `SearchInputViewTest` extends `PaparazziViewTest`, builds the view with `paparazzi.context`, and points `DispatcherProvider` at a `StandardTestDispatcher` shared with `runTest`, so the debounce runs on virtual time.

Three cases: the debounced listener fires at the regular interval, short input is held for the longer one, and clearing drops the pending debounce. Mutation checked as you described: with `cancelLastDebounce()` removed, the last one fails with `expected:<[]> but was:<[, abc]>`.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/search/SearchInputView.kt
gpunto: Fair point, the promise did change without the KDoc saying so. Added to `setDebouncedInputChangedListener`:

> Input of 1-2 characters is held for longer than that, as such queries are the most expensive ones to search for.

On the opt-out: there is none by design. The thresholds live in an `@InternalStreamChatApi` helper rather than a public config, so they stay tunable by us and not part of the API surface. If an integrator needs the old flat interval we would rather hear about it than ship a knob for it now.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
gpunto: No good reason, and it is the same defect. Fixed in `29c09b3b9d`: the debounce now resolves on `query.trim()`, with a test for `"a  "` taking the 500ms path.

One difference worth noting versus the user search: there the request itself trims, so the debounce simply mirrors the payload. Here the query still goes into `Filters.autocomplete("text", ...)` untrimmed, so the debounce is deliberately resolved on the trimmed length while the untrimmed term is sent. The trimmed length is what drives the cost, so it is the right input for the decision. Changing what gets sent felt out of scope for this PR.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
gpunto: Updated in `29c09b3b9d`, the primary constructor now carries the same two-line text as the three public ones.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/AddMembersViewControllerTest.kt
gpunto: You are right, and I had been filtering `w:` lines out of my Gradle output, so I never saw them. All four are gone in `29c09b3b9d`.

Used `@OptIn(ExperimentalCoroutinesApi::class)` rather than the propagating `@ExperimentalCoroutinesApi` the Compose test class carries, matching `DebouncerTest` in core, which is the closest sibling. Happy to switch it to the Compose form if you prefer one convention across the two.

stream-chat-android-ui-components/src/main/kotlin/io/getstream/chat/android/ui/feature/search/SearchInputView.kt
andremion: Makes sense, keeping the thresholds internal is the right call. Thanks for the KDoc line.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/viewmodel/channels/ChannelListViewModel.kt
andremion: Agreed, the trimmed length is the right input for the decision, and leaving what gets sent alone is fine by me.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/channel/info/AddMembersViewControllerTest.kt
andremion: `@OptIn` is the better fit, no reason to propagate it from a test class. Matching DebouncerTest works for me.

andremion (approved):
All five addressed, thanks. LGTM.

PR #6652, opened 2026-08-20
Skip oversized /sync event replay and refresh watched channels

aleksandar-apostolov:
### Goal

Closes [AND-1196](https://linear.app/stream/issue/AND-1196/address-sync-events-limit-in-android-sdk)

Every event returned by `/sync` on reconnect is replayed through the shared state and persistence pipeline. A payload can sit below the backend cap and still be expensive enough to hold that pipeline, delaying the other requests that need it. The existing HTTP 400 "too many events" branch advances the last-synced timestamp and refreshes nothing, leaving visible channels stale.

### Implementation

- `SyncManager` skips event replay when a `/sync` payload exceeds `SYNC_EVENT_REPLAY_MAX_COUNT` (250) and refreshes the affected channels through `queryChannels` instead. The `isTooManyEventsToSyncError()` (HTTP 400) branch takes the same path, where previously it refreshed nothing.
- The refresh targets `getTrackedWatchedChannels()` — channels an explicitly opened screen subscribed to — and falls back to the cids `/sync` was asked about when nothing has subscribed yet, which is the usual cold-start case. It is deliberately not `getActiveChannelsLogic()`, which holds an entry for every channel any query has paged through. Batched by 30, the `queryChannels` page cap.
- The last-synced date only advances once that refresh succeeds. On failure the checkpoint is kept, so the next reconnect requests the same range and retries; the cids that did refresh are still reported so recovery does not query them twice, and the batch loop stops at the first failed request.
- `SYNC_MAX_AGE_DAYS` (30) is the ceiling under that hold: past it `/sync` is skipped and the checkpoint abandoned, so a persistently failing refresh cannot retry the same range forever. Worth noting that any `/sync` validation error takes this branch, not only a genuinely oversized range.
- `MarkAllReadEvent` is still applied when replay is skipped, since the refresh does not carry read state.
- The threshold is internal — a performance guard, not a configuration knob. No public API change, `apiCheck` passes.

Known limitation: a message hard-deleted during a skipped range survives in local state, because the refresh upserts rather than replaces. Refreshing messages instead would reset the user's scroll position on every reconnect.

### 🎨 UI Changes

No visual changes — reconnect recovery behaviour only.

### Testing

`:stream-chat-android-client:testDebugUnitTest --tests "*SyncManagerTest"`, covering:

- an oversized payload skips replay and refreshes, while a payload equal to the limit still replays
- the HTTP 400 branch refreshes instead of silently advancing
- the checkpoint is held when the refresh fails, and when one batch of two fails
- the batch loop stops at the first failed request
- 35 cids are split into batches of 30 and 5
- the refresh ignores channels that are only in the paged-channel cache, and falls back to the synced cids when nothing is watched yet
- `MarkAllReadEvent` is applied on the skip path
- sync is skipped entirely past the max age

Each new behaviour was verified red against a control that removes it, so none of the assertions are vacuous.

`detekt`, `spotlessCheck` and `apiCheck` are green, and `compileDebugUnitTestKotlin` passes across all modules.

### ☑️ Contributor Checklist
- [x] Signed the Stream CLA
- [x] PR is linked to the issue it resolves
- [x] Affected KDocs updated
- [x] PR targets `develop`


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -314,17 +339,24 @@ internal class SyncManager(
             chatClient.getSyncHistory(cappedCids, lastSyncAt).await()
         }
         if (result.isTooManyEventsToSyncError()) {
+            // Event replay is impossible here, so refresh the watched surfaces instead. The
+            // payload is unavailable, hence the timestamp moves to now to avoid retrying forever.
             logger.e { "[performSync] failed (too many events to sync): $result" }
+            val refreshedCids = refreshActiveWatchedChannels()
             updateLastSyncedDate(latestEventDate = Date(now()), rawLatestEventDate = null)
-            return
+            return refreshedCids
```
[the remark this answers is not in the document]
aleksandar-apostolov: recheck

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -314,17 +339,29 @@ internal class SyncManager(
             chatClient.getSyncHistory(cappedCids, lastSyncAt).await()
         }
         if (result.isTooManyEventsToSyncError()) {
```
gpunto: **Blocking.** This branch should keep advancing the checkpoint unconditionally.

The server *refused* the payload here, so unlike `skipEventReplay` there is nothing to re-fetch — holding `lastSyncedAt` buys nothing and guarantees the identical 400 next reconnect. There's also no backstop: every `lastSyncedAt` write is in this function, and `syncMaxThreshold` only gates the message/reaction retries, so nothing heals a stuck checkpoint. (iOS does have one — `SyncRepository.Constants.maximumDaysSinceLastSync = 30` advances unconditionally.)

Suggest keeping the hold-the-checkpoint rule in `skipEventReplay` only.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -609,6 +682,83 @@ internal class SyncManager(
...
+     *
+     * @param cidsToExclude Channel ids already refreshed elsewhere in the recovery flow.
+     *
+     * @return The channel ids that were refreshed, or a failure when any request failed.
+     */
+    private suspend fun refreshActiveWatchedChannels(
+        cidsToExclude: Set<String> = emptySet(),
+    ): Result<Set<String>> {
+        val online = clientState.isOnline
+        val watchedCids = buildSet {
```
gpunto: **Blocking.** This set is unbounded, and it isn't "actively watched channels".

`getActiveChannelsLogic()` is `channels.values.toList()` over a `ConcurrentHashMap` (`LogicRegistry.kt:81`) filled by `channel(...)`/`getOrPut` and pruned only on channel-deleted events (`EventHandlerSequential.kt:504`) — so it's every channel the session has touched. `performSync` caps the same source at `SYNC_MAX_CIDS = 100` and `updateActiveChannels` at `take(30)` for exactly this reason.

Uncapped, a long session becomes dozens of sequential `state=true`/`watch=true` requests, all inside `mutex.withLock`. Worth capping, and preferring `getTrackedWatchedChannels()` as the primary source since that one is genuinely weak-ref'd to open screens. Did you profile this against the 250-event replay?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -337,14 +374,43 @@ internal class SyncManager(
...
+     * `queryChannels` instead.
+     *
+     * The last-synced timestamp only moves to the newest skipped event once that refresh succeeded:
+     * it is what makes discarding the payload safe. When the refresh fails the checkpoint is kept,
+     * so the next reconnect asks for the same range again and can retry - at the cost of re-fetching
+     * a payload we discard, which is a request and a parse rather than a stalled pipeline.
+     *
+     * @return The channel ids that were refreshed, empty when the refresh failed.
+     */
+    private suspend fun skipEventReplay(sortedEvents: List<ChatEvent>): Set<String> {
```
gpunto, line 402: Not blocking, and I suspect this is an iOS-side design question rather than something to fix here.

The fallback is inert in exactly the case that triggers it. The `"no active cids found"` branch above is only reached when `getActiveChannelsLogic()` is empty — so when the fallback re-reads that same source it contributes nothing by construction, and `watchedCids` reduces to `getTrackedWatchedChannels()`, itself empty before any screen has subscribed. Cold start after a long offline stretch therefore gives `Success(emptySet())`: the checkpoint jumps to the newest skipped event and nothing is refreshed.

The impact is mild — those channels have no state objects and aren't on screen, so the cost is a stale Room cache until the next query refreshes them. But it does mean the KDoc's "the actively watched channels are refreshed with `queryChannels` instead" doesn't hold for the main trigger of an oversized payload, which is worth saying out loud somewhere.

iOS is inert here too: `refreshActiveWatchedChannels` collects from the same kind of weak live-controller set (`SyncRepository.swift:320-323`), so this is a shared gap rather than an Android deviation. Probably one to raise with them rather than diverge on quietly.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -337,14 +374,43 @@ internal class SyncManager(
...
+        logger.i {
+            "[performSync] skipping event replay; events.size: ${sortedEvents.size} exceeds $eventReplayMaxCount"
+        }
+        val refresh = refreshActiveWatchedChannels()
+        if (refresh !is Result.Success) {
+            logger.e { "[skipEventReplay] fallback refresh failed, keeping lastSyncedAt: ${refresh.errorOrNull()}" }
+            return emptySet()
+        }
+        val skippedLatestEvent = sortedEvents.last()
```
gpunto: `MarkAllReadEvent` handling is dropped here.

The replay path scans `sortedEvents` for it and calls `updateAllReadStateForDate`; this path returns before that, so a "mark all read" from another device inside the skipped window is lost. Scanning the sorted list is cheap and needs no replay.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -609,6 +682,83 @@ internal class SyncManager(
...
+                .onError {
+                    logger.e { "[refreshActiveWatchedChannels] request failed: $it" }
+                    failed.set(it)
+                }
+                .onSuccessSuspend { queryResult ->
+                    val foundChannels = queryResult.channels
+                    foundChannels.forEach { channel ->
+                        ChannelId.fromTypeAndId(channel.type, channel.id)
+                            ?.let(logicRegistry::channel)
+                            ?.updateDataForChannel(channel, channel.messages.size)
```
gpunto: `shouldRefreshMessages` defaults to `false`, so this merges instead of replacing.

With a non-empty local window `ChannelLogicImpl.updateDataForChannel` takes the `upsertMessages` branch, which only adds/updates — a message hard-deleted inside the skipped window survives in state and in Room indefinitely. Since this path exists to *replace* event replay, it should replace the window.

Also, `messageLimit = channel.messages.size` makes the `if (messageLimit > 0)` guard skip message updates entirely for channels returned with no messages, and makes `setEndOfOlderMessages(size < messageLimit)` permanently `false`. Copied from `updateActiveChannels` so pre-existing, but newly load-bearing here.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -609,6 +682,83 @@ internal class SyncManager(
...
+            logicRegistry.getActiveChannelsLogic().mapTo(this) { it.cid }
+        } - cidsToExclude
+        logger.d {
+            "[refreshActiveWatchedChannels] watchedCids.size: ${watchedCids.size}, online: $online, " +
+                "cidsToExclude.size: ${cidsToExclude.size}"
+        }
+        if (watchedCids.isEmpty()) {
+            return Result.Success(emptySet())
+        }
+        if (!online) {
```
gpunto: Offline is a hard failure here, but a silent no-op in both siblings (`rewatchTrackedWatchedChannels`, `updateActiveChannels`). Since it blocks the checkpoint, it compounds the first comment. Intentional?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -609,6 +682,83 @@ internal class SyncManager(
...
+     * Having nothing to refresh is a success: with no watched channels there is nothing visible that
+     * can go stale. A failed request is not, and the caller must not advance the last-synced date on
+     * it, or the events it skipped become unrecoverable.
+     *
+     * @param cidsToExclude Channel ids already refreshed elsewhere in the recovery flow.
+     *
+     * @return The channel ids that were refreshed, or a failure when any request failed.
+     */
+    private suspend fun refreshActiveWatchedChannels(
+        cidsToExclude: Set<String> = emptySet(),
```
gpunto: `cidsToExclude` is dead — no caller passes it, and nothing can: Android runs `performSync()` before `restoreActiveChannels()`, so there are no earlier-refreshed cids yet. (On iOS the equivalent is fed from steps that run *before* `/sync` — `SyncOperations.swift:163`.)

Either drop it, or wire up the other direction: `restoreActiveChannels` threads `alreadyRefreshedCids` into `updateActiveChannels` but not into `rewatchTrackedWatchedChannels()`, which re-issues a `watch` per tracked cid the fallback just refreshed — one request each, in the path this PR exists to make cheaper.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -609,6 +682,83 @@ internal class SyncManager(
...
+            return Result.Success(emptySet())
+        }
+        if (!online) {
+            logger.w { "[refreshActiveWatchedChannels] rejected (offline)" }
+            return Result.Failure(Error.GenericError("Cannot refresh watched channels while offline"))
+        }
+
+        val failed = AtomicReference<Error>()
+        val refreshedCids = mutableSetOf<String>()
+        watchedCids.chunked(QUERY_CHANNELS_MAX_LIMIT).forEach { batch ->
```
gpunto, line 724: A failed batch neither breaks the loop nor rolls back. The remaining `ceil(n/30)` requests still fire after a known failure, `failed.set()` keeps only the last error, and — because the function then returns `Result.Failure` — the batches that *did* succeed are dropped from the return value, so `restoreActiveChannels` re-queries them.

iOS breaks on the first error instead of recursing into the remaining batches: `startWatchingChannelsInBatches` returns `.failure(.failedFetchingChannels)` immediately (`SyncRepository.swift:478-482`). Doing the same here avoids both the wasted requests and the redundant re-query — in the path this PR exists to make cheaper.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/internal/SyncManagerTest.kt
```diff
@@ -367,6 +599,12 @@ internal class SyncManagerTest {
         whenever(repositoryFacade.selectSyncState(any())) doReturn testSyncState
         whenever(chatClient.getSyncHistory(any(), any<String>())) doReturn TestCall(result)
         whenever(chatClient.getSyncHistory(any(), any<Date>())) doReturn TestCall(result)
+        // The timestamp only advances once the watched-channels fallback succeeded, so the client
```
gpunto: This comment doesn't match the suite — no test sets `isOnline` to `false` (all 15 occurrences are `doReturn true`), so the offline branch is uncovered. Worth adding, given it gates the checkpoint.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/internal/SyncManagerTest.kt
```diff
@@ -1194,8 +1432,10 @@ internal class SyncManagerTest {
     private fun TestScope.buildSyncManager(
         isAutomaticSyncOnReconnectEnabled: Boolean = true,
         syncMaxThreshold: TimeDuration = TimeDuration.seconds(5),
+        eventReplayMaxCount: Int = 250,
```
gpunto: Duplicates `SYNC_EVENT_REPLAY_MAX_COUNT`; the tests will drift silently if the constant changes.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -314,17 +339,29 @@ internal class SyncManager(
             chatClient.getSyncHistory(cappedCids, lastSyncAt).await()
         }
         if (result.isTooManyEventsToSyncError()) {
+            // Event replay is impossible here, so refresh the watched surfaces instead. The payload
+            // is unavailable, hence the timestamp moves to now - but only once the refresh
+            // succeeded, otherwise the skipped events would become unrecoverable.
             logger.e { "[performSync] failed (too many events to sync): $result" }
+            val refresh = refreshActiveWatchedChannels()
+            if (refresh !is Result.Success) {
```
andremion, line 347: This can leave a permanent 400 stuck. Before, the branch always advanced `lastSyncedAt`, so the next `/sync` asked for a fresh range. Now if the refresh fails nothing moves, and since `isTooManyEventsToSyncError()` is just 400 plus validation error, the same request keeps failing on every reconnect. I ran three reconnects locally with a failing `queryChannelsInternal` and `lastSyncedAt` never moved.

The hold-on-failure rule makes sense in `skipEventReplay`, where you can ask for the payload again. Here the server already refused it, so could this branch keep the unconditional `updateLastSyncedDate(now())`? Happy to be told otherwise.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
andremion, line 411: Not blocking: the replay path calls `updateAllReadStateForDate` for each `MarkAllReadEvent`, this one drops them. Looks low impact since `markedAllReadAt` is only stored and compared, but a quick scan of `sortedEvents` before discarding would keep the two paths the same. Worth it?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -433,7 +506,7 @@ internal class SyncManager(
                 is Result.Success -> {
                     val updatedCids = result.value
                     logger.v { "[restoreActiveChannels] standardCids.size: ${result.value.size}" }
-                    updateActiveChannels(recoverAll, updatedCids + groupedHandledCids)
+                    updateActiveChannels(recoverAll, updatedCids + groupedHandledCids + alreadyRefreshedCids)
```
andremion, line 509: `alreadyRefreshedCids` only gets to `updateActiveChannels`. `rewatchTrackedWatchedChannels()` on line 497 still goes over the whole tracked set and does one `queryChannel` per cid, so on the grouped path the channels the fallback just refreshed get fetched again one by one. The `cidsToExclude` param on `refreshActiveWatchedChannels` is also unused right now.

Could `alreadyRefreshedCids` be passed there too?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -609,6 +682,83 @@ internal class SyncManager(
...
+     *
+     * @return The channel ids that were refreshed, or a failure when any request failed.
+     */
+    private suspend fun refreshActiveWatchedChannels(
+        cidsToExclude: Set<String> = emptySet(),
+    ): Result<Set<String>> {
+        val online = clientState.isOnline
+        val watchedCids = buildSet {
+            addAll(stateRegistry.getTrackedWatchedChannels())
+            logicRegistry.getActiveChannelsLogic().mapTo(this) { it.cid }
```
andremion, line 708: `getActiveChannelsLogic()` is `channels.values` (`LogicRegistry.kt:300`), and that cache gets an entry for every channel in every `queryChannels` result (`QueryChannelsStateLogic.kt:184`). So this is every channel the list has paged through, not the ones on screen, and there is no cap. With 300 cids I get 10 sequential `queryChannelsInternal` calls, each with `watch=true` and a full state write.

That can cost more than the 251 events it avoids, and it re-watches channels the user is not viewing. Any reason not to cap it at 30 like `updateActiveChannels` does, or limit it to `getTrackedWatchedChannels()` plus channels with `recoveryNeeded`?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
andremion, line 742: Small thing: `shouldRefreshMessages` stays `false` here, so this takes the `upsertMessages` branch (`ChannelLogicImpl.kt:346`), which adds and updates but never removes. A message hard-deleted during the skipped range stays in local state, and the checkpoint has already moved past the `message.deleted` event. `rewatchTrackedWatchedChannels` sets `shouldRefresh = true` for this reason.

A full replace resets scroll position though, so not obvious either way. Which do you prefer?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -99,6 +115,7 @@ internal class SyncManager(
     private val userPresence: Boolean,
     private val isAutomaticSyncOnReconnectEnabled: Boolean,
     private val syncMaxThreshold: TimeDuration,
+    private val eventReplayMaxCount: Int = SYNC_EVENT_REPLAY_MAX_COUNT,
```
andremion, line 118: Small thing: only tests pass this. Could it get `@VisibleForTesting` like `performSync`?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/internal/state/internal/SyncManagerTest.kt
```diff
@@ -258,6 +259,237 @@ internal class SyncManagerTest {
...
+                syncManager.performSync(cids = listOf("1", "2"))
+
+                /* Then */
+                expectNoEvents()
+            }
+            argumentCaptor<QueryChannelsRequest> {
+                verify(chatClient).queryChannelsInternal(capture())
+                val filter = firstValue.filter as InFilterObject
+                assertEquals(setOf(watchedChannel.cid), filter.values)
+                assertEquals(true, firstValue.watch)
```
andremion, line 285: Small thing: `state` and `watch` both default to true on `QueryChannelsRequest`, so these two pass whatever the fallback does. If they matter here, maybe set them explicitly in the request?

andremion (commented):
Nice, the skip logic reads clearly. One concern on the 400 branch and a scope question on the fallback, both inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
gpunto: Answering the "happy to be told otherwise": I don't think unconditional advance is right either, though the stuck-400 concern is real.

iOS holds the checkpoint on fallback failure in this same 400 path. `syncMissingEvents` routes the 400 into `handleSyncEventReplayFallback` (`SyncRepository.swift:417`), and on `.failure` that completes without ever calling `updateLastSyncAt` (`:459-461`). So the hold isn't an Android deviation.

What keeps it from wedging over there is two things we don't have:

- `maximumDaysSinceLastSync = 30`: when `lastSyncAt` is older than 30 days it skips `/sync` altogether and advances the checkpoint unconditionally (`SyncRepository.swift:291-304`). That's the backstop.
- `SyncError.failedFetchingChannels.shouldRetry == true`, so the operation retries twice within the same reconnect before giving up (`:18-19`, `SyncOperations.swift:20`).

On Android there's no equivalent. Every `lastSyncedAt` write is on the `/sync` path, and `syncMaxThreshold` only gates the channel/message/reaction retries (lines 768-930), never the checkpoint. So I'd rather add the max-age backstop than make this branch advance unconditionally — otherwise we diverge from the implementation we're porting and *still* have no floor under a stuck checkpoint.

One thing that makes your concern worse than it looks: `isTooManyEventsToSyncError()` is just 400 + `serverErrorCode == VALIDATION_ERROR` (`NetworkError.kt:36`). Any `/sync` validation error — bad cid, malformed date — now holds the checkpoint, not only genuinely oversized ranges.

gpunto (commented):
Most of what I had queued up is already covered by @andremion's review, so I've trimmed this down to what isn't.

Two inline notes: the fallback being inert at cold start (non-blocking, and probably an iOS-side design question), and the batch loop not breaking or rolling back on a failed request.

Plus a reply on the 400-branch thread — iOS holds the checkpoint there too, so I think the fix is the max-age backstop we're missing rather than an unconditional advance.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
andremion: The 30 day ceiling covers it, and it is a better answer than advancing here. Thanks both. 61b7f3835da42d20f44c7942bcd918715d7e7953

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
andremion: Tracked set first, and the synced cids as the floor are capped at SYNC_MAX_CIDS, so it stays bounded either way. Works for me. 61b7f3835da42d20f44c7942bcd918715d7e7953 and 30d4b4de83c7f3957f7e4f64258e03b26c077186

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
andremion: The KDoc works just as well, no need for the annotation. 61b7f3835da42d20f44c7942bcd918715d7e7953

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
andremion: By checking you last pushes I can consider that you intend to keep the upsert. Fair enough. Losing scroll position on a channel the user is looking at is worse than a hard-deleted message staying until the next full query. If it ever comes up from a customer, `shouldRefreshMessages` is where to look. Resolving this one.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/internal/state/sync/internal/SyncManager.kt
```diff
@@ -849,6 +1046,14 @@ internal class SyncManager(
         return this == null || diff(now()) > syncMaxThreshold
     }
 
+    /**
+     * Whether the last-synced date is too old for `/sync` to serve. Distinct from
+     * [exceedsSyncThreshold], which gates retrying local entities, not the sync range.
+     */
+    private fun Date.exceedsMaxSyncAge(): Boolean {
+        return diff(now()) > TimeDuration.days(SYNC_MAX_AGE_DAYS)
```
gpunto, line 1054: Not a big deal — flagging for the record rather than asking for a change.

`diff` is `abs(time - otherTime)` (`DateUtils.kt:72-75`), so this is symmetric: a `lastSyncedAt` more than 30 days in the *future* trips the ceiling too. Reaching that means the device clock is set over a month behind real time while the checkpoint holds a server date, which is basically an unreachable state — so this is about the expression matching the intent, not something anyone will hit.

For the record, iOS's version is signed and wouldn't trip: `Calendar.current.dateComponents([.day], from: self, to: Date()).day` (`SyncRepository.swift:524-526`).

If you're ever touching this line anyway, `now() - time > TimeDuration.days(SYNC_MAX_AGE_DAYS).millis` says "too old" without covering the future case. The `abs` in `exceedsSyncThreshold` is fine where it is — that one gates retrying local entities, where either direction means "stale enough to give up on".

gpunto (approved):

andremion (approved):

PR #6654, opened 2026-08-20
Skip mark-read when the server tracks none of the loaded messages

gpunto:
### Goal

Develop port of #6646. The mark-read guard from #6644 skips whenever the last item in the list is the current user's own local-only message, which also suppresses legitimate mark-read in channels the server does have messages for. An `error` message is terminal, so a rejected or moderation-bounced send stays the last item and keeps those channels unread indefinitely.

Skip only when the server tracks none of the loaded messages. Read state leaves out our own not-yet-persisted messages, and silent, shadowed and deleted ones from anyone, so none of them can resolve a mark-read call. Silent is documented as not marking a channel unread, and a deleted message is removed from read state by `HandleMessageDeleted`, so a channel holding only those has nothing to mark read. Deleted for the current user only does not count, since the message stays in the channel for everyone else.

Part of [AND-1395](https://linear.app/stream/issue/AND-1395/mark-read-fires-on-server-rejected-message-echoes-emitting-messageread)

### Implementation

- `MessageListController.markLastMessageReadInternal`: replace the "last item is our own local-only message" check with "no loaded message is one the server tracks", via a new `Message.isInServerReadState` predicate covering local-only, `silent` and `shadowed`.
- Tests: error and ephemeral cases now hold only our own local-only message, plus new cases for a silent-only channel, a shadowed-only channel, a silent message following a tracked one, and a server-side message preceding the echo.
- `randomMessage` randomises both `silent` and `deletedAt`, which the gate now keys on, so gate-dependent tests build their messages through a `trackedMessage` helper that pins them, and the `nowMessage` helper pins them too. This includes the two unread-label tests that only exist here, one of which asserts state computed inside the branch the gate guards.
- The shadowed case cannot occur today (the server clears the flag for the author, and `ChannelStateImpl.shouldIgnoreUpsertion` drops other users' shadowed messages). The check stays as defensive cover and its test says so.

The deleted case will be ported to v6 separately, since #6646 merged without it.

### Testing

- Unit tests as above, with the `ui-common` suite rerun to confirm the gate no longer depends on a randomised field.
- The same change was verified on device on the v6 branch (#6646): with no server-side message the guard rejects and no read request is sent, and with server-side messages the request goes out and the emitted event carries a populated `last_read_message_id`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message read tracking to exclude silent and shadowed messages from server read state.
  * Corrected read-status handling when silent messages, shadowed messages, or error echoes appear in the message list.
* **Tests**
  * Added coverage for read tracking and unread indicators across these message scenarios.

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -1791,9 +1791,9 @@ public class MessageListController(
         }
     }
 
-    // The server keeps our own local-only messages out of its read state, along with silent,
-    // shadowed and deleted ones, so none of them can resolve a mark-read call. Deleted for the
-    // current user only does not count: the message is still there for everyone else.
+    // The server keeps our own local-only messages out of its channel read state, along with
+    // silent, shadowed and deleted ones, so none of them can resolve a mark-read call. Deleted for
+    // the current user only does not count: the message is still there for everyone else.
     private fun Message.isInServerReadState(currentUserId: String?): Boolean =
         !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed && deletedAt == null
 
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -1758,13 +1758,10 @@ public class MessageListController(
         val messageText = message?.text
         logger.d { "[markLastMessageRead] cid: $cid, msgId($isInThread): $messageId, msgText: \"$messageText\"" }
 
-        // The server keeps our own local-only messages out of its read state, so marking read
-        // without any other message makes it emit message.read with no last_read_message_id.
+        // Marking read with nothing the server tracks makes it emit message.read with no
+        // last_read_message_id.
         val currentUserId = clientState.user.value?.id
-        val hasServerSideMessage = messageItems.any { item ->
-            !(item.message.isMine(currentUserId) && item.message.isLocalOnly())
-        }
-        if (!hasServerSideMessage) {
+        if (messageItems.none { it.message.isInServerReadState(currentUserId) }) {
             logger.v { "[markLastMessageRead] cid: $cid; rejected[$isInThread] (no server-side message)" }
             return
         }
@@ -1793,6 +1790,11 @@ public class MessageListController(
             markChannelAsRead()
         }
     }
+
+    // The server keeps our own local-only messages out of its read state, and silent and shadowed
+    // ones from anyone, so none of them can resolve a mark-read call.
+    private fun Message.isInServerReadState(currentUserId: String?): Boolean =
+        !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed
 
     private fun markChannelAsRead() {
         val (channelType, channelId) = cid.cidToTypeAndId()
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -1791,10 +1791,11 @@ public class MessageListController(
         }
     }
 
-    // The server keeps our own local-only messages out of its read state, and silent and shadowed
-    // ones from anyone, so none of them can resolve a mark-read call.
+    // The server keeps our own local-only messages out of its read state, along with silent,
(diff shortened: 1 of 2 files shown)


stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -1794,6 +1791,11 @@ public class MessageListController(
         }
     }
 
+    // The server keeps our own local-only messages out of its read state, and silent and shadowed
+    // ones from anyone, so none of them can resolve a mark-read call.
+    private fun Message.isInServerReadState(currentUserId: String?): Boolean =
+        !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed
```
andremion: Not blocking, but deleted messages have the same problem. The server drops a deleted message from its read state, and a soft-deleted one stays in `messageItems` (`filterMessagesToShow` only drops system and error messages), so a channel whose only message was deleted still passes the gate and gets an empty `last_read_message_id` back. Would a `deletedAt != null` check here make sense, or is that better as a follow-up? `isDeleted()` would also catch `deletedForMe`, which probably needs to stay.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
```diff
@@ -461,8 +475,89 @@ internal class MessageListControllerTests {
...
+        val chatClient: ChatClient = mock()
+        val messagesState = MutableStateFlow(
+            listOf(
+                randomMessage(
+                    id = "1",
+                    user = user1,
+                    type = MessageType.REGULAR,
+                    syncStatus = SyncStatus.COMPLETED,
+                    silent = false,
+                    shadowed = true,
```
andremion, line 545: Small thing: I don't think this state can occur. The server clears `shadowed` before sending a message back to its own author, so with `user1` as the current user the flag is always false, and other users' shadowed messages are already filtered out in `ChannelStateImpl.shouldIgnoreUpsertion` and `ChannelStateLegacyImpl.messagesTransformation`. Fine to keep `!shadowed` as a defensive check, but could the test note that, so it doesn't read as covering a live path?

andremion (approved):
Looks good. One scope question and one nit inline, both optional.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
gpunto: Verified and added in 0483fa68. `HandleMessageDeleted` calls `ChannelMessagesDelete`, so the message leaves the read-state store, and nothing on our side keeps a soft-deleted message out of `messageItems`, so the gate did pass with nothing to resolve.

You were right about `isDeleted()` too, so the check is `deletedAt == null`. A message deleted for the current user only is still in the channel for everyone else, so the server can resolve the read state and skipping would leave the channel unread. There is a test for that case alongside the deleted-only one.

`randomMessage` randomises `deletedAt` as well as `silent`, which would have made the same six tests flaky, so they now go through a `trackedMessage` helper that pins both. That also keeps the next randomised field from catching us out.

We will port this to v6 as well, since #6646 merged without it.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
gpunto: Confirmed: `User.HideShadowBan` clears the flag when the viewer is the author, and `ChannelStateImpl.shouldIgnoreUpsertion` drops other users' shadowed messages, so the state cannot reach the list. Kept the check as defensive and the test now says so:

```kotlin
// Defensive only: the server clears shadowed for the author, and ChannelStateImpl drops
// other users' shadowed messages, so this state does not reach the list today.
```

Happy to drop both the check and the test instead if you would rather not carry an unreachable branch. I leaned towards keeping it so the predicate still mirrors the server's read-state rule in full.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
andremion: Checked the delete-for-me path, it never reaches the read-state delete, so `deletedAt == null` is the right check. Thanks for pulling the fixtures into a helper.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
andremion: Keeping it sounds right to me, the predicate is easier to read as the full server rule. The comment covers it, so no need to drop anything.

PR #6655, opened 2026-08-20
Migrate the queryMembers and push preference request bodies to generated models

gpunto:
### Goal

Migrate the `queryMembers` and push-preference request bodies to their generated models.

### Implementation

- Replace the hand-written `QueryMembersRequest` with the generated `QueryMembersPayload`, and
  `UpsertPushPreferencesRequest` plus its `UpstreamPushPreferenceInputDto` and `UpstreamChatPreferencesDto`
  with the generated request, `PushPreferenceInput` and `ChatPreferencesInput`. The push-preference
  response was migrated earlier, so this finishes that endpoint.
- Add `Member.toChannelMemberRequest()` and replace `ChatPreferences.toDto()` with
  `ChatPreferences.toChatPreferencesInput()`, converting each toggle to its generated sealed type.
- Register the 17 sealed-class enum adapters the three new models declare. Moshi resolves an adapter for
  every declared property type when it builds the enclosing adapter, so a missing one fails at runtime
  rather than at compile time, even for properties that are always null.
- The `DownstreamPushPreferenceDto` and `DownstreamChatPreferencesDto` stay: they are still used by the
  user DTOs.

### Notes

- `members` narrows from the full `UpstreamMemberDto` to `ChannelMemberRequest`, which drops nothing the
  endpoint reads: the handler only calls `Members.HashUserIDs()` to resolve a distinct channel. The role
  and custom data are still carried, since the domain member has them.
- `PushPreferenceInput` also declares `call_level`, `feeds_level`, `feeds_preferences` and `user_id`. The
  domain input types have no counterpart for any of them, so they are absent rather than defaulted.
- `queryBannedUsers` was in scope and is deferred: the generated `QueryBannedUsersPayload` has no
  counterpart for the four `created_at_*` filters the SDK sends. Go's payload embeds `*types.BansPager`
  which declares them, but each date field carries a `query` tag, and a tagged field is left out of the
  schema. `limit` and `offset` have no such tag, which is why they survive. Adopting it would silently
  return unfiltered bans. Tracked in
  [AND-1406](https://linear.app/stream/issue/AND-1406/spec-omits-the-querybannedusers-date-filters-blocking-that-request).

### Testing

- `RequestBodiesAdapterTest` pins the serialized body for the members query and for the three distinct
  push-preference shapes (`chat_level`, `disabled_until`, `chat_preferences`). The sort is a
  `SortParamRequest` literal rather than the output of the same conversion the payload uses.
- The `MoshiChatApiTest` cases for the six push-preference entry points now assert the generated request.
- Device-probed every entry point, since each emits a different body and one standing in for the rest
  would not cover the others. All six round-tripped: the levels and snooze timestamps came back as set,
  and the chat toggles came back populated, which is the field a previous slice silently dropped. Absent
  values are omitted from the body rather than sent as `null`. `queryMembers` went out as a URL query
  payload and returned the member with its role.
- Not routed in the E2E mock server, so no mock-server change is needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of member queries and push-notification preference updates.
  * Preserved support for chat, feed, channel, and user notification preference values, including unknown settings.

* **Tests**
  * Added coverage for request serialization, member query payloads, snoozing notifications, channel settings, and chat preference updates.
  * Strengthened validation of member roles and custom data in requests.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
```diff
@@ -216,6 +216,20 @@ internal class DtoMapping(
...
     )
 
+    /**
+     * Maps the domain [Member] to the generated network [ChannelMemberRequest] model.
+     *
+     * The query endpoint hashes the user ids to resolve a distinct channel and reads nothing else, but the role
+     * and custom data are carried anyway since the domain member has them. `user` stays absent: the outgoing
+     * model embeds a read-only [io.getstream.chat.android.network.models.UserResponse].
+     */
+    internal fun Member.toChannelMemberRequest(): ChannelMemberRequest = ChannelMemberRequest(
```
andremion, line 204: Nothing asserts this mapping. `testQueryMembers` only does `verify(api).queryMembers(any())`, and `DtoMappingTest` has a case for the `MemberData` version but not this one. I set `channelRole = null` and `custom = emptyMap()` locally and the client suite still passed.

Given a previous slice dropped member `extraData` here, could we add a `DtoMappingTest` case and have `testQueryMembers` assert the payload the way `testQueryUsers` does?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/requests/QueryMembersRequest.kt
```diff
@@ -1,31 +0,0 @@
...
-
-@JsonClass(generateAdapter = true)
-internal data class QueryMembersRequest(
-    val type: String,
-    val id: String,
-    val filter_conditions: Map<*, *>,
-    val offset: Int,
-    val limit: Int,
-    val sort: List<Map<String, Any>>,
-    val members: List<UpstreamMemberDto>,
```
andremion, line 30: Small thing: this was the last production user of `UpstreamMemberDto`. What's left in `src/main` is the declaration, `Member.toDto()`, and `UpstreamMemberDtoAdapter` with its registration, and the tests still cover all three.

Remove it here, or track it for a later slice? Happy either way, just easier to spot now.

andremion (commented):
One on the description rather than the code. The reason `queryBannedUsers` is missing the `created_at_*` filters isn't that `OpenAPIInfo()` hand-lists the properties, that list only sets metadata. A field drops out of the spec when it carries a `query` tag. You can see it here: `QueryMembersPayload`'s `OpenAPIInfo()` only lists `filter_conditions` and `sort`, but the model still has `type`, `id`, `limit`, `offset` and `members`.

Same conclusion, different reason. Worth putting on AND-1406, because `QueryMembersPayload` is missing `user_id_gte`/`gt`/`lte`/`lt` and the four `created_at_*` fields for the same reason. Fine here since we don't send them, but it will come up again on every payload with a pager.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
gpunto: Confirmed, and thanks for the mutation check. The gap was hiding behind the overload: `DtoMappingTest` covered the `MemberData` version, so the mapper name looked covered.

Added a `DtoMappingTest` case for the `Member` overload and made `testQueryMembers` assert the payload the way `testQueryUsers` does, with explicit literals for the sort and the member so it cannot pass by round-tripping the mappers. Your mutation now fails 3 tests.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/requests/QueryMembersRequest.kt
gpunto: Checked and it is fully dead: `ChatClient` line 4602 is the `MemberData.toDto()` overload, so nothing in `src/main` called `Member.toDto()` after this slice.

Removed it here, in its own commit: the declaration, `Member.toDto()`, `UpstreamMemberDtoAdapter` and its registration, plus the three test fixtures. 189 deletions.

gpunto:
You're right, and `QueryMembersPayload` settles it: `OpenAPIInfo()` lists only `filter_conditions` and `sort`, yet the generated model has seven fields. So the hand-list is metadata and does not gate inclusion. What drops a field is the `query` tag: in `MembersPagerRequest` the four `user_id_*` and four `created_at_*` all carry one and are all absent, while untagged `limit` and `offset` survive. Same in `BansPager`.

Fixed the description here, and also on #6659 where I had written the same wrong reason. Updated the migration notes so the check is "grep the Go struct and its embedded pager for `query` tags", not "read `OpenAPIInfo`" - that heuristic would have given a false negative on this very payload.

Worth stressing your last point: nothing regresses today, since the old `QueryMembersRequest` had the same seven fields as the generated payload. But `queryBannedUsers` exposes the four `created_at_*` parameters publicly and sends them, so migrating that request would silently stop filtering while the call still succeeds. Carrying the `query` tag detail onto AND-1406.

andremion (approved):
Both addressed, thanks. One follow-up for AND-1406 rather than here: `QueryMembersPayload` is missing the same pager fields for the same reason, so the ticket may want to cover the pattern and not just the bans endpoint.

PR #6656, opened 2026-08-21
Migrate the queryMembers response to the generated MembersResponse model

gpunto:
### Goal

Migrate the `queryMembers` response to the generated `MembersResponse`.

### Implementation

- Replace the hand-written `QueryMembersResponse` with the generated `MembersResponse`, and point the
  endpoint at it.
- The member itself needs nothing here: `ChannelMemberResponse`, its collecting adapter and the mapper to
  the domain `Member` all landed with the thread channel slice, so this one only adopts the wrapper.

### Notes

`user_id`, `role`, `is_moderator`, `deleted_at` and `deleted_messages` are declared on the generated model
but were not on the hand-written DTO, so they would stop reaching `Member.extraData`. The adapter passes
them to `alsoKeepInExtraData`, which keeps them in the overflow map alongside their own fields, and
`ChannelMemberResponseParsingTest` pins that: a member's `custom` carries both its genuine custom data and
the kept `user_id`. Both go away with AND-1398.

### Testing

- `ChannelMemberResponseParsingTest` covers the collected custom data on the member and its nested user,
  including the keys the keep set holds.
- Device-probed on the wire against a throwaway channel seeded with three different member shapes: the
  creator, a member added with custom data, and a pending invite. All three parsed, including the fields
  the generated model requires non-null. The nested user resolved with its `name` promoted out of `custom`,
  and the pending invite came back as `status=pending invited=true`. That run predates the keep set, so it
  observed `extraData` holding only genuine custom data; the parsing test covers the current behaviour,
  where the declared keys are kept as well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved member-query response handling for more consistent channel member data.
  * Preserved standard and custom member fields during response parsing.
  * Added support for response duration information.

* **Tests**
  * Expanded coverage for member responses, randomized member data, and custom fields.
  * Updated member-query parsing tests to reflect the revised response format.

### DIFF
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
@@ -92,6 +92,7 @@ import io.getstream.chat.android.models.querysort.QuerySorter
 import io.getstream.chat.android.network.models.AppResponseFields
 import io.getstream.chat.android.network.models.BlockUsersResponse
 import io.getstream.chat.android.network.models.BlockedUserResponse
+import io.getstream.chat.android.network.models.ChannelMemberResponse
 import io.getstream.chat.android.network.models.ChannelResponse
 import io.getstream.chat.android.network.models.CreateGuestResponse
 import io.getstream.chat.android.network.models.DeviceResponse
@@ -726,6 +727,20 @@ internal object Mother {
         sum_scores = sumScores,
         first_reaction_at = firstReactionAt,
         last_reaction_at = lastReactionAt,
+    )
+
+    fun randomChannelMemberResponse(
+        user: UserResponse = randomUserResponse(),
+        channelRole: String = randomString(),
+    ): ChannelMemberResponse = ChannelMemberResponse(
+        user = user,
+        userId = user.id,
+        channelRole = channelRole,
+        createdAt = randomDate(),
+        updatedAt = randomDate(),
+        banned = randomBoolean(),
+        shadowBanned = randomBoolean(),
+        notificationsMuted = randomBoolean(),
     )
 
     fun randomDownstreamMemberDto(
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ChannelMemberResponseParsingTest.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ChannelMemberResponseParsingTest.kt
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2014-2026 Stream.io Inc. All rights reserved.
+ *
+ * Licensed under the Stream License;
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://github.com/GetStream/stream-chat-android/blob/main/LICENSE
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.getstream.chat.android.client.parser2
+
+import io.getstream.chat.android.network.models.MembersResponse
+import org.amshove.kluent.shouldBeEqualTo
+import org.amshove.kluent.shouldHaveSize
+import org.junit.jupiter.api.Test
+
+internal class ChannelMemberResponseParsingTest {
+    private val parser = ParserFactory.createMoshiChatParser()
+
+    @Test
+    fun `Parse the members of a MembersResponse`() {
+        val members = parser.fromJson(MEMBERS_JSON, MembersResponse::class.java).members
+
+        members shouldHaveSize 1
+        members.first().channelRole shouldBeEqualTo "channel_member"
+        members.first().userId shouldBeEqualTo "leandro"
+        members.first().custom["memberProbe"] shouldBeEqualTo "sentinel"
+    }
+
+    @Test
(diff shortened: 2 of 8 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ChannelMemberResponseParsingTest.kt
```diff
@@ -0,0 +1,77 @@
...
+        )
+    }
+
+    @Test
+    fun `Standard member fields do not leak into custom`() {
+        val member = parser.fromJson(MEMBERS_JSON, MembersResponse::class.java).members.first()
+
+        member.channelRole shouldBeEqualTo "channel_member"
+        member.userId shouldBeEqualTo "leandro"
+        member.user?.custom shouldBeEqualTo mapOf("birthland" to "Polis Massa")
```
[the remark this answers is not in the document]
gpunto: Added the other four to `MEMBERS_JSON` and switched the assertion to an exact map, so all five kept keys are covered and a key dropping out fails rather than passing a subset check. Each addition is mutation-checked: removing any one of `role`, `is_moderator`, `deleted_at`, `deleted_messages` from the keep set now fails this test.

Skipping the `MoshiChatApiTest` half: that test covers the request/response wiring rather than `extraData`, so deterministic members would not add coverage the parsing test lacks, and randomised fixtures are the convention in that file.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
```diff
@@ -728,6 +729,34 @@ internal object Mother {
...
+        id = id,
+        role = role,
+        language = language,
+        banned = randomBoolean(),
+        online = randomBoolean(),
+        createdAt = randomDate(),
+        updatedAt = randomDate(),
+    )
+
+    fun randomChannelMemberResponse(
```
andremion, line 732: Nothing asserts what `ChannelMemberResponse.toDomain()` produces. `testQueryMembers` runs it but only checks the `Result` type, and the generated `ChannelResponse` fixtures never carry a member (`ChannelDtoTestData.channelResponse` and `Mother.randomChannelResponse` both leave `members` at the default).

The part I'd most like covered is the `user?.toDomain() ?: User(id = userId.orEmpty())` fallback at `DomainMapping.kt:507`. It's new behaviour, `DownstreamMemberDto.user` was non-null, and the server sends `user` only when the member has one. This fixture always sets it, so that branch never runs. Could you add a `DomainMappingTest` case with and without `user`?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ChannelMemberResponseParsingTest.kt
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package io.getstream.chat.android.client.parser2
+
+import io.getstream.chat.android.network.models.MembersResponse
+import org.amshove.kluent.shouldBeEqualTo
+import org.amshove.kluent.shouldHaveSize
+import org.junit.jupiter.api.Test
+
+internal class ChannelMemberResponseParsingTest {
```
andremion, line 24: This overlaps `GeneratedExtraDataParityTest`, which already parses a `ChannelMemberResponse` carrying the same five kept keys plus a sentinel. What's new here is the exact-map assertion, the `MembersResponse` wrapper and the nested user's `custom`.

The exact-map assertion is stronger than the `shouldContainAll` over there, so maybe move it into the parity test and keep this file on the wrapper and the nested user? Two files pinning the same keep set will drift. Happy to be told otherwise.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
```diff
@@ -728,6 +729,34 @@ internal object Mother {
         last_reaction_at = lastReactionAt,
     )
 
+    fun randomUserResponse(
```
andremion: Small thing: `randomUserResponse` looks like a pure move, about 26 lines of diff for no change. Fine if grouping it next to `randomChannelMemberResponse` was on purpose.

andremion (commented):
Looks good, it really is just the wrapper. One question on test coverage and a nit inline.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
gpunto: Good catch, nothing was asserting that mapper. Added three cases to `DomainMappingTest`: the full `ChannelMemberResponse` to `Member` mapping, the no-user fallback, and `custom` reaching `extraData` with nulls dropped.

The fallback is the one that mattered: the hand-written DTO required a `user`, the generated model does not, so `User(id = userId)` is a path the old code never had.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ChannelMemberResponseParsingTest.kt
gpunto: Right, that overlapped `GeneratedExtraDataParityTest`. Moved the exact-map assertion there, replacing its `shouldContainAll` for the member so the keep set is pinned in one place.

This test now covers only what it uniquely can: the `MembersResponse` wrapper and the nested user's collected `custom`.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
gpunto: Not on purpose, it was a dedupe artifact from an earlier rebase where I kept this branch's copy and dropped develop's. Put back in place; `Mother.kt` is now additions only.

andremion (approved):

PR #6657, opened 2026-08-21
Migrate the OG scrape response to the generated GetOGResponse model

gpunto:
### Goal

Migrate the OG scrape response to the generated `GetOGResponse` model, so `enrichUrl` no longer parses
into the hand-written `AttachmentDto`.

### Implementation

- `OpenGraphApi.get()` returns the generated `GetOGResponse`, mapped to the domain `Attachment`.
- Add `GetOGResponseAdapter` so root-level custom fields are still collected into `extraData`, which is
  what `AttachmentDtoAdapter` did for this endpoint.
- Add the generated `GetOGResponse`, `Images`, `ImageData`, `Action` and `Field`. The last four come along
  because Go's `GetOGResponse` embeds the whole `Attachment` payload, so the schema inherits every field
  an attachment can carry, not only the ones a scrape produces.

`AttachmentDto` stays: message and draft attachments still use it.

### Notes

`GetOGResponse` embeds the whole attachment shape, so it declares more than a scrape produces. The mapper
covers all of it except `giphy`, which is not mapped here: it has a single producer, the giphy slash
command writing to a message attachment, so it cannot reach `/og`. The slice that adopts the shared
`Attachment` model owns it, and has to re-emit it into `extraData["giphy"]` or `Attachment.giphyInfo()`
stops finding gif urls; `closure.py` prints that reminder for any closure containing `Images`.

`file_size`, `image`, `mime_type` and `name` are absent from the payload struct, so when they are sent they
arrive inside `custom`. The mapper reads them back out and removes them, so they do not also sit in
`extraData` under their wire names, which is how the hand-written DTO behaved.

### E2E

`/og` is routed in the mock server, and it answered with a recorded message attachment carrying no
`duration`, the one field the generated model requires non-null, which turned `HyperLinksTests` red.
[mock-server #64](https://github.com/GetStream/stream-chat-test-mock-server/pull/64) sends `duration` on
that response and is merged, so the suite runs against the default mock branch with no override needed.

### Testing

- `GetOGResponseParsingTest` covers the collected custom data, an empty scrape falling back to domain
  defaults, the undeclared root fields being mapped and removed from `extraData`, and
  `author_name`/`author_link`, which are same-named on both sides so a swap would otherwise go unnoticed.
- Device-probed `enrichUrl` against three pages (an article, an image page and a video page), logging the
  raw response body next to every field of the resulting `Attachment`. The wire returned `type`, `title`,
  `title_link`, `text`, `image_url`, `thumb_url`, `asset_url`, `og_scrape_url` and `duration`, all mapping
  correctly (`duration` is response metadata with no domain field), and `extraData` stayed empty, so
  nothing arrived under an unexpected name. The remaining declared fields were not seen on those three
  pages, which is why they are mapped rather than dropped: three samples cannot show a field is never
  sent, and losing one would be silent.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Open Graph link previews by correctly processing metadata such as titles, images, authors, media details, and custom fields.
  - Ensured preview information is mapped consistently to attachment details.
  - Preserved additional Open Graph properties as extra attachment data instead of discarding them.
- **Tests**
  - Added coverage for Open Graph response parsing, empty responses, author information, and custom metadata handling.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
@@ -808,8 +808,16 @@ internal class DomainMapping(
         )
 
     internal fun GetOGResponse.toDomain(): Attachment {
-        // The spec does not declare file_size/image/mime_type/name, so when the wire sends them they
-        // arrive in `custom`. Read them back out and remove them, or they would also sit in extraData
+        // The model declares `author_icon`, `color`, `footer`, `footer_icon`, `pretext`, `actions`,
+        // `fields` and `giphy`, none of which the hand-written DTO declared, so they no longer reach
+        // `extraData`. A scrape cannot produce any of them: the response is built from a fresh attachment
+        // that only ever receives title, title_link, author_name, author_link, text, image_url, thumb_url,
+        // type, asset_url and og_scrape_url. `giphy` in particular has a single producer, the giphy slash
+        // command writing to a message attachment, so the slice adopting the shared Attachment model owns
+        // it and must keep `extraData["giphy"]` populated or `Attachment.giphyInfo()` stops finding urls.
+        //
+        // The spec does not declare file_size/image/mime_type/name either, so if the wire ever sends them
+        // they arrive in `custom`. Read them back out and remove them, or they would also sit in extraData
         // under their wire names, which the hand-written DTO never did.
         val extras = custom.toMutableMap()
         val fileSize = (extras.remove("file_size") as? Number)?.toInt() ?: 0
@@ -836,10 +844,6 @@ internal class DomainMapping(
             originalWidth = originalWidth,
             extraData = extras.mapNotNull { (key, value) -> value?.let { key to it } }.toMap(),
         )
-        // `giphy` is deliberately not mapped: unlike the fields above, it has a single producer -- the
-        // giphy slash command, which writes it to a message attachment -- so it cannot reach /og. The
-        // slice that adopts the shared Attachment model owns it, and must re-emit it into
-        // extraData["giphy"] or Attachment.giphyInfo() stops finding gif urls.
     }
 
     /**
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
@@ -139,6 +139,7 @@ import io.getstream.chat.android.network.models.ChatPreferencesResponse
 import io.getstream.chat.android.network.models.DeviceResponse
 import io.getstream.chat.android.network.models.FullUserResponse
 import io.getstream.chat.android.network.models.GetApplicationResponse
+import io.getstream.chat.android.network.models.GetOGResponse
 import io.getstream.chat.android.network.models.PollOptionResponseData
 import io.getstream.chat.android.network.models.PrivacySettingsResponse
 import io.getstream.chat.android.network.models.PushPreferencesResponse
@@ -805,6 +806,41 @@ internal class DomainMapping(
             originalWidth = original_width,
             extraData = extraData.toMutableMap(),
         )
+
+    internal fun GetOGResponse.toDomain(): Attachment {
+        // The spec does not declare file_size/image/mime_type/name, so when the wire sends them they
+        // arrive in `custom`. Read them back out and remove them, or they would also sit in extraData
+        // under their wire names, which the hand-written DTO never did.
+        val extras = custom.toMutableMap()
+        val fileSize = (extras.remove("file_size") as? Number)?.toInt() ?: 0
+        val image = extras.remove("image") as? String
+        val mimeType = extras.remove("mime_type") as? String
+        val name = extras.remove("name") as? String
+        return Attachment(
+            assetUrl = assetUrl,
+            authorName = authorName,
+            authorLink = authorLink,
+            fallback = fallback,
+            fileSize = fileSize,
+            image = image,
+            imageUrl = imageUrl,
+            mimeType = mimeType,
+            name = name,
+            ogUrl = ogScrapeUrl,
+            text = text,
+            thumbUrl = thumbUrl,
+            title = title,
+            titleLink = titleLink,
+            type = type,
+            originalHeight = originalHeight,
+            originalWidth = originalWidth,
+            extraData = extras.mapNotNull { (key, value) -> value?.let { key to it } }.toMap(),
+        )
+        // `giphy` is deliberately not mapped: unlike the fields above, it has a single producer -- the
+        // giphy slash command, which writes it to a message attachment -- so it cannot reach /og. The
+        // slice that adopts the shared Attachment model owns it, and must re-emit it into
+        // extraData["giphy"] or Attachment.giphyInfo() stops finding gif urls.
+    }
 
     /**
      * Transforms [BanResponse] to [BannedUser], or to null when the ban carries no user.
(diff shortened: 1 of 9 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -747,6 +748,41 @@ internal class DomainMapping(
...
+            text = text,
+            thumbUrl = thumbUrl,
+            title = title,
+            titleLink = titleLink,
+            type = type,
+            originalHeight = originalHeight,
+            originalWidth = originalWidth,
+            extraData = extras.mapNotNull { (key, value) -> value?.let { key to it } }.toMap(),
+        )
+        // `giphy` is deliberately not mapped: unlike the fields above, it has a single producer -- the
```
andremion: Small thing: the comment is after the return, so it reads as attached to unreachable code. Above the return, or in the function KDoc, would be easier to follow.

Same block: `color`, `pretext`, `author_icon`, `footer`, `footer_icon`, `actions` and `fields` are in the same position as `giphy`. They are declared now, so they stop reaching `extraData`, and the note only covers `giphy`. A scrape does not produce any of them, so nothing is lost, but the next reader has no way to know that. Maybe keep them with `alsoKeepInExtraData`, the way `GENERATED_CHANNEL_EXTRA_DATA_KEYS` does, or just add them to the note? Happy to be told otherwise.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/GetOGResponseAdapter.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.squareup.moshi.JsonAdapter
+import com.squareup.moshi.JsonReader
+import com.squareup.moshi.JsonWriter
+import com.squareup.moshi.ToJson
+import io.getstream.chat.android.network.models.GetOGResponse
+
+/**
+ * Collects the root-level custom fields of an OG scrape into `custom`, which is how the API sends them.
+ *
+ * The endpoint also ships `file_size`, `image`, `mime_type` and `name` at the root even though the spec
```
andremion: Not blocking: "The endpoint also ships" reads as a fact, but a scrape cannot carry those four. The wording in the description ("when they are sent") is the accurate one. Could you use that here too, so nobody later assumes the endpoint sends them?

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/GetOGResponse.kt
```diff
@@ -0,0 +1,101 @@
...
+    internal val type: String? = null,
+
+    @Json(name = "actions")
+    internal val actions: List<io.getstream.chat.android.network.models.Action>? = emptyList(),
+
+    @Json(name = "fields")
+    internal val fields: List<io.getstream.chat.android.network.models.Field>? = emptyList(),
+
+    @Json(name = "giphy")
+    internal val giphy: io.getstream.chat.android.network.models.Images? = null,
```
andremion, line 100: Just a note, not a change request: `Images` requires all seven sub-objects non-null, so a response with a partial `giphy` object now fails the whole parse, where before it landed in `extraData`. `/og` never sends it, so this is fine here. It is the first parse path that can reach `Images` though, and `giphyInfo()` reads that map with safe casts today because giphy data does arrive on message attachments. Worth a look at the required fields when that slice migrates.

andremion (approved):
Nice, this reads well. A few nits inline, none blocking.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
gpunto: Both fixed in `bd6ccd308f`. The comment is above the return now.

On the seven fields: I traced what a scrape can set. `HandleGetOG` builds the response through `types.NewAttachmentFromOG`, which starts from a fresh `&Attachment{}` and only ever assigns title, title_link, author_name, author_link, text, image_url, thumb_url, type, asset_url and og_scrape_url. `color`, `pretext`, `author_icon`, `footer`, `footer_icon`, `actions` and `fields` are never touched, so they cannot arrive here.

I went with documenting rather than `alsoKeepInExtraData`, for the same reason we dropped `latest_hidden_channels` from the keep set on #6664: keeping a key the endpoint cannot send is inert, and it makes the parity test pin a shape that never occurs. The note now names all eight and says what the scrape does populate, so the next reader does not have to go dig in the Go handler.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/GetOGResponseAdapter.kt
gpunto: Right, that overstated it. A scrape does not set `file_size`, `image`, `mime_type` or `name` either, so the recovery is a guard, not an observed shape. Reworded in `bd6ccd308f` to match the description.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/GetOGResponse.kt
gpunto: Useful note, and it is already handled in the follow-up: the attachment slice keeps the raw `giphy` object via `alsoKeepInExtraData`, so `extraData["giphy"]` stays the exact map `giphyInfo()` reads and the typed `Images` is never on the path that matters. That is deliberately different from the reference implementation, which rebuilt the map from the typed field with `Images.toLegacyMap()` and could only reproduce the seven variants and five inner keys the model declares.

On the strictness itself: a partial `giphy` cannot come off the wire. `giphy.Images` is seven value-type `ImageData` structs of five value-type strings with no `omitempty`, so Go cannot omit a variant or emit null; unfilled ones serialise as empty strings. Probed twice (`/giphy` send and shuffle): all seven variants and all five keys present each time, only `frames` empty off the original variant. The enclosing field is `giphy: Images? = null`, so an absent object is fine too. `AttachmentParsingTest` on that branch pins the seven-variant shape with that reasoning recorded.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -747,6 +748,41 @@ internal class DomainMapping(
...
+        // giphy slash command, which writes it to a message attachment -- so it cannot reach /og. The
+        // slice that adopts the shared Attachment model owns it, and must re-emit it into
+        // extraData["giphy"] or Attachment.giphyInfo() stops finding gif urls.
+    }
+
```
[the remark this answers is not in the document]
gpunto: Both fixed in `bd6ccd308f`. The comment is above the return now.

On the seven fields: I traced what a scrape can set. `HandleGetOG` builds the response through `types.NewAttachmentFromOG`, which starts from a fresh `&Attachment{}` and only ever assigns title, title_link, author_name, author_link, text, image_url, thumb_url, type, asset_url and og_scrape_url. `color`, `pretext`, `author_icon`, `footer`, `footer_icon`, `actions` and `fields` are never touched, so they cannot arrive here.

I went with documenting rather than `alsoKeepInExtraData`, for the same reason we dropped `latest_hidden_channels` from the keep set on #6664: keeping a key the endpoint cannot send is inert, and it makes the parity test pin a shape that never occurs. The note now names all eight and says what the scrape does populate, so the next reader does not have to go dig in the Go handler.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/GetOGResponse.kt
andremion: Makes sense, thanks. I checked the shape and it lines up, the variants and their inner keys are always serialised, so a partial object cannot happen. Nothing to change here.

PR #6658, opened 2026-08-21
Skip mark-read when the only messages are deleted

gpunto:
### Goal

v6 port of the deleted-message case from #6654, which came out of review there after #6646 had merged.

The server removes a deleted message from its read state (`HandleMessageDeleted` calls `ChannelMessagesDelete`), and nothing on our side keeps a soft-deleted message out of the message list, so a channel whose only message was deleted still reached `markRead` with nothing for the server to resolve, and the emitted `message.read` carried no `last_read_message_id`.

Part of [AND-1395](https://linear.app/stream/issue/AND-1395/mark-read-fires-on-server-rejected-message-echoes-emitting-messageread)

### Implementation

- Add `deletedAt == null` to the `Message.isInServerReadState` predicate the mark-read gate keys on. Deleted for the current user only does not count, since the message stays in the channel for everyone else and the server can still resolve the read state.
- Tests: a channel holding only a deleted message skips mark-read, and a message deleted for the current user only still marks read.
- `randomMessage` randomises `deletedAt` as well as `silent`, so gate-dependent tests build their messages through a `trackedMessage` helper that pins both, instead of pinning each field at every call site.

### Testing

Unit tests as above, with the `ui-common` suite rerun to confirm the gate no longer depends on randomised fields.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved read-status handling for deleted messages.
  * Messages deleted from the server are no longer marked as read.
  * Messages deleted only for the current viewer can still be marked as read.

### DIFF
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -1746,9 +1746,9 @@ public class MessageListController(
         }
     }
 
-    // The server keeps our own local-only messages out of its read state, along with silent,
-    // shadowed and deleted ones, so none of them can resolve a mark-read call. Deleted for the
-    // current user only does not count: the message is still there for everyone else.
+    // The server keeps our own local-only messages out of its channel read state, along with
+    // silent, shadowed and deleted ones, so none of them can resolve a mark-read call. Deleted for
+    // the current user only does not count: the message is still there for everyone else.
     private fun Message.isInServerReadState(currentUserId: String?): Boolean =
         !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed && deletedAt == null
 
--- a/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
+++ b/stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
@@ -1746,10 +1746,11 @@ public class MessageListController(
         }
     }
 
-    // The server keeps our own local-only messages out of its read state, and silent and shadowed
-    // ones from anyone, so none of them can resolve a mark-read call.
+    // The server keeps our own local-only messages out of its read state, along with silent,
+    // shadowed and deleted ones, so none of them can resolve a mark-read call. Deleted for the
+    // current user only does not count: the message is still there for everyone else.
     private fun Message.isInServerReadState(currentUserId: String?): Boolean =
-        !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed
+        !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed && deletedAt == null
 
     private fun markChannelAsRead() {
         val (channelType, channelId) = cid.cidToTypeAndId()
--- a/stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
+++ b/stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
@@ -351,9 +351,9 @@ internal class MessageListControllerTests {
     fun `When repetitive markLastMessageRead calls appear only single API call should be sent`() = runTest {
         val chatClient: ChatClient = mock()
         val messages = arrayListOf(
-            trackedMessage(id = "1"),
-            trackedMessage(id = "2"),
-            trackedMessage(id = "3"),
+            gateMessage(id = "1"),
+            gateMessage(id = "2"),
+            gateMessage(id = "3"),
         )
         val messagesState = MutableStateFlow(messages)
         val controller = Fixture(chatClient = chatClient)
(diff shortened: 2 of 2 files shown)


stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
```diff
@@ -1557,6 +1642,24 @@ internal class MessageListControllerTests {
 
         private fun nowDate() = Date(testCoroutines.dispatcher.scheduler.currentTime)
 
+        /**
+         * A message the server tracks in its read state. [randomMessage] randomises `silent` and
+         * `deletedAt`, which the mark-read gate keys on, so both are pinned here.
+         */
+        private fun trackedMessage(
```
andremion: Not blocking: the helper covers the positive tests, but the five negative gate tests still build with `randomMessage` and leave `deletedAt` random (lines 404, 428, 457, 517, 547). `randomDateOrNull` is 50/50, so each of them is now rejected by the new `deletedAt` clause on about half the runs instead of the clause it is named after. I reduced the gate to only `deletedAt == null` locally and they still passed on some runs each (5/8, 1/6, 3/6, 2/8, 6/8). Could the helper also take `type`, `silent` and `shadowed`, so those five go through it too? The one at 404 can already use it as is with `syncStatus = IN_PROGRESS`, like the peer test at 716.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
```diff
@@ -1746,10 +1746,11 @@ public class MessageListController(
...
     }
 
-    // The server keeps our own local-only messages out of its read state, and silent and shadowed
-    // ones from anyone, so none of them can resolve a mark-read call.
+    // The server keeps our own local-only messages out of its read state, along with silent,
+    // shadowed and deleted ones, so none of them can resolve a mark-read call. Deleted for the
+    // current user only does not count: the message is still there for everyone else.
     private fun Message.isInServerReadState(currentUserId: String?): Boolean =
-        !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed
+        !(isMine(currentUserId) && isLocalOnly()) && !silent && !shadowed && deletedAt == null
```
andremion: Small thing on the comment: this predicate also gates `markThreadRead`, and on the server side a deleted reply is not dropped from the thread read state the way it is from the channel one. The channel set also picks deleted messages back up when it is rebuilt after a cache miss. I do not think either is a problem, since the unread counts are already decremented when the message is deleted, so there is nothing left to clear. Maybe scope the comment to the channel delete path? Happy to be told otherwise.

andremion (approved):
Looks good. One test note and one comment-wording question inline, both optional.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
gpunto: You are right, and thanks for measuring it. Fixed in 79fb018 (and 485f7d7 on #6654).

The helper now exposes every field the gate reads, defaulting to a message the server tracks, so each test overrides only the clause it covers:

```kotlin
private fun gateMessage(
    id: String = randomString(),
    user: User = randomUser(),
    type: String = MessageType.REGULAR,
    syncStatus: SyncStatus = SyncStatus.COMPLETED,
    silent: Boolean = false,
    shadowed: Boolean = false,
    deletedAt: Date? = null,
    deletedForMe: Boolean = false,
)
```

All 19 call sites go through it now, negatives included, so the five you listed read as `gateMessage(id = "1", user = user1, type = MessageType.ERROR)` and friends.

I checked the property you were really after by removing each clause in turn and running the class:

| clause removed | tests that fail |
| --- | --- |
| own local-only | not-COMPLETED, ephemeral-only, error-echo-only |
| `!silent` | silent-only |
| `!shadowed` | shadowed-only |
| `deletedAt == null` | deleted-only |

Each clause is now pinned by exactly the test named after it, on both branches. My earlier reasoning was wrong in a way worth naming: when the `silent` clause made the positive tests flaky I pinned only those, on the grounds that the negatives "skip either way". Skipping either way is exactly the defect you found.

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListController.kt
gpunto: Good point, scoped to the channel read state in 79fb018:

```kotlin
// The server keeps our own local-only messages out of its channel read state, along with
// silent, shadowed and deleted ones, so none of them can resolve a mark-read call. Deleted for
// the current user only does not count: the message is still there for everyone else.
```

Both caveats check out on my side too: `FilterMessagesForUnreadCache` filters silent, shadowed, ephemeral and error but not deleted, so a rebuilt channel set does pick deleted messages back up, and the thread path decrements the counts rather than removing the message. So the deleted clause is conservative rather than wrong, and as you say there is nothing left to clear either way.

gpunto: Same comment change landed on #6654 (485f7d7), so the two branches stay identical.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/MessageListControllerTests.kt
andremion: Thanks, the new helper reads much better, and I get the same failure set per clause on my side.

PR #6659, opened 2026-08-24
Migrate the queryBannedUsers response to the generated BanResponse

gpunto:
### Goal

Migrate the `queryBannedUsers` response to the generated `BanResponse`.

### Implementation

- Replace the hand-written `BannedUserResponse` and `QueryBannedUsersResponse` with the generated
  `BanResponse` and `QueryBannedUsersResponse`, and point the endpoint at them.
- `shadow` is omitted from the payload when it is `false`, so the mapper reads an absent flag as not
  shadowed. The hand-written DTO relied on a Moshi default for the same thing.
- `user` is nullable on the generated model, so the mapper returns null for a ban without one and the
  endpoint turns that into a `Result.Failure` rather than dropping the ban or throwing.
- `created_at` becomes required. It is always sent, so the previous nullable field was wider than the
  wire.

The request keeps the hand-written body: the pager date fields carry a `query` tag, which leaves them out
of the schema, so migrating it would lose `created_at_after` and friends.

### Testing

- `DomainMappingTest` covers the full mapping plus the two branches the generated model adds: a ban with
  no user, and a ban with no `shadow` flag.
- Device-probed against a throwaway channel across every shape the endpoint returns: a ban with all
  optionals absent, a ban with a reason and a timeout, a shadow ban, and an empty result. The nested
  banned user and banning user both resolved with their names promoted out of `custom`.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -998,10 +998,16 @@ constructor(
...
+        ).flatMapDomain { response ->
+            val bans = response.bans.mapNotNull { it.toDomain() }
+            val result = if (bans.size == response.bans.size) {
+                Result.Success(bans)
+            } else {
+                Result.Failure(
+                    Error.GenericError("A ban in the query banned users response carried no user"),
+                )
             }
+            CoroutineCall(coroutineScope) { result }
```
andremion: Small thing: the other two `flatMapDomain` uses in this file build the failure with `ErrorCall(coroutineScope, error)` (lines 2039 and 2056), and it is already imported at the top. Any reason to go through `CoroutineCall` here instead? Behaviour is the same, so only worth it if you want the file consistent.

andremion (approved):
Looks good, one nit inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Good spot, no reason other than me not noticing the two below. Switched to `ErrorCall(coroutineScope, error)` so the file stays consistent.

Pushed as a separate commit so the change is visible on top of your review.

PR #6661, opened 2026-08-25
Deprecate the unflag APIs since the backend endpoint is no longer supported

andremion:
### Goal

The backend endpoint `POST /moderation/unflag` is deprecated and is a no-op. The SDK still exposes public APIs that call it, so an integrator can call something that silently does nothing.

This PR marks those APIs as deprecated. The removal happens in the next major version.

Resolves AND-1448

### Implementation

Added `@Deprecated` (WARNING level) to the public APIs that reach the endpoint:

- `ChatClient.unflagUser` and `ChatClient.unflagMessage`
- `MessageListController.unflagUser` and `MessageListController.ErrorEvent.UnflagUserError`
- Compose `MessageListViewModel.unflagUser`
- XML `MessageListViewModel.Event.UnflagUser`

Added `@Suppress("DEPRECATION")` where the SDK itself still routes through them, so the build stays free of new warnings: `MessageListController.unflagUser`, Compose `MessageListViewModel.unflagUser`, XML `MessageListViewModel.onEvent`, the `errorEventHandler` in `MessageListView`, and `ChatClientModerationApiTests`.

`ChatApi`, `MoshiChatApi` and `ModerationApi` are internal, so they are left as they are. They get removed together with the public APIs.

No behaviour changes. The `.api` dumps are unchanged, since the validator does not record annotations.

### Testing

- `./gradlew spotlessApply apiDump` (no `.api` changes)
- `./gradlew detekt`
- `./gradlew :stream-chat-android-client:testDebugUnitTest --tests '*ChatClientModerationApiTests*' --tests '*MoshiChatApiTest*'`

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified that unflagging users and messages is no longer supported by the backend.
  * Existing unflag actions remain available but have no effect.

* **Deprecations**
  * Added deprecation warnings across moderation and message-list APIs.
  * Unflagging functionality is planned for removal in the next major release.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/ChatClient.kt
@@ -4289,6 +4289,11 @@ internal constructor(
      *
      * @param userId The ID of the user to un-flag.
      */
+    @Deprecated(
+        message = "The backend no longer supports un-flagging. This call does nothing and will be removed " +
+            "in the next major version.",
+        level = DeprecationLevel.WARNING,
+    )
     @CheckResult
     public fun unflagUser(userId: String): Call<Flag> = api.unflagUser(userId)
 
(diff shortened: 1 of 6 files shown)


andremion:
@coderabbit review

gpunto (approved):

PR #6662, opened 2026-08-25
Fix the socket never reconnecting after a transient connectivity drop

aleksandar-apostolov:
### Goal
Fixes [AND-1450](https://linear.app/stream/issue/AND-1450/websocket-never-recovers-from-disconnectednetwork-leaving-the-client) — socket gets into `Disconnected.Network` and never comes back out, client stays offline until the process dies.

### Implementation
Provider now records direct `isConnected()` reads so the cached flag can't go stale. `Disconnected.Network` keeps a retry and handles `Event.Resume`. Lifecycle observer only swallows the replayed resume. Also dropped `NET_CAPABILITY_VALIDATED`. All internal, no API change.

### Testing
New unit tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lifecycle resume handling to avoid missed or duplicate resume events.
  * Improved network connectivity detection and transition reporting.
  * Preserved scheduled connection retries when network connectivity is lost.
  * Automatically restarts the chat connection when the app resumes after a network disconnection.

* **Tests**
  * Added coverage for lifecycle, connectivity, reconnection, and network-loss scenarios.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
@@ -47,16 +47,17 @@ internal class NetworkStateProvider(
 
         override fun onLost(network: Network) {
             availableNetworks.remove(network)
-            notifyListenersIfNetworkStateChanged()
             if (availableNetworks.isEmpty()) {
-                // No available networks, notify listeners about disconnection
-                listeners.onDisconnected()
+                // No available networks, the capability read may still lag behind
+                setConnected(false)
+            } else {
+                notifyListenersIfNetworkStateChanged()
             }
         }
     }
 
     @Volatile
-    private var isConnected: Boolean = isConnected()
+    private var lastKnownConnected: Boolean = queryConnectivity()
 
     @Volatile
     private var listeners: Set<NetworkStateListener> = setOf()
@@ -64,15 +65,20 @@ internal class NetworkStateProvider(
     private val isRegistered: AtomicBoolean = AtomicBoolean(false)
 
     private fun notifyListenersIfNetworkStateChanged() {
-        val isNowConnected = isConnected()
-        if (!isConnected && isNowConnected) {
-            logger.i { "Network connected." }
-            isConnected = true
-            listeners.onConnected()
-        } else if (isConnected && !isNowConnected) {
-            logger.i { "Network disconnected." }
-            isConnected = false
-            listeners.onDisconnected()
+        setConnected(queryConnectivity())
+    }
+
+    private fun setConnected(isNowConnected: Boolean) {
+        synchronized(lock) {
+            if (lastKnownConnected == isNowConnected) return
+            lastKnownConnected = isNowConnected
+            if (isNowConnected) {
+                logger.i { "Network connected." }
+                listeners.onConnected()
+            } else {
+                logger.i { "Network disconnected." }
+                listeners.onDisconnected()
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -88,14 +94,23 @@ internal class NetworkStateProvider(
...
-    fun isConnected(): Boolean {
+    /**
+     * Reports whether the network is currently usable.
+     *
+     * The answer is recorded, so that a caller which concludes "no network" from this cannot leave
+     * the last known state stale. If it did, the following reconnection would not be seen as a
+     * transition and listeners would never be notified that the network came back.
+     */
+    fun isConnected(): Boolean = queryConnectivity().also {
+        synchronized(lock) { lastKnownConnected = it }
```
VelikovPetar: I am thinking whether this should also notify the observers? Because calling `isConnected` now could also update the cache, but no observers would be informed about it.

For example:
1. We're offline, cache is false, socket is in `NetworkDisconnected`.                                                                                                             
2. Network comes back.                                                                                                                                                          
3. Something calls `isConnected()` before the system callback lands — it returns `true` and quietly sets the cache to `true`.                                                         
4. The callback arrives, `setConnected(true)` sees no change, stays quiet.                                                                                                        
5. Socket never hears the network is back.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
```diff
@@ -154,7 +154,9 @@ internal open class ChatSocket(
                             }
                             is State.Disconnected.NetworkDisconnected -> {
                                 streamWebSocket?.close()
-                                healthMonitor.stop()
+                                // Keep a backed-off retry armed. Recovery must not depend solely on
+                                // a network callback arriving, or a missed one strands the socket.
+                                healthMonitor.onDisconnected()
```
VelikovPetar: I am not entirely sure that this provides anything for the issue. `healthMonitor.onDisconnected`  postpones a reconnect -> which emits a `Event.WebSocketEventLost` after the defined delay.  But that event is not handled by the current state of the socket (`State.Disconnected.NetworkDisconnected`). Additionally, if the socket state changes in the meantime (i.e. to `State.Connecting`), receiving the `Event.WebSocketEventLost` will have bigger consequences - changing the socket state to `State.Disconnected.WebSocketEventLost`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/network/NetworkStateProvider.kt
```diff
@@ -88,14 +94,23 @@ internal class NetworkStateProvider(
...
+    private fun queryConnectivity(): Boolean {
         return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
             runCatching {
                 connectivityManager.run {
-                    getNetworkCapabilities(activeNetwork)?.run {
-                        hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) &&
-                            hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED)
-                    }
+                    getNetworkCapabilities(activeNetwork)
+                        ?.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
```
VelikovPetar: Is this change safe? From the official android docs: 
> The NET_CAPABILITY_INTERNET capability means that the network is configured to reach the general Internet. It may or may not actually provide connectivity ; the NET_CAPABILITY_VALIDATED bit indicates that the system found actual connectivity to the general Internet the last time it checked. Apps interested in actual connectivity should usually look at both these capabilities.

I am a bit afraid that removing this check would result in having false positives in some cases.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/socket/ChatSocket.kt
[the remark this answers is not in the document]
aleksandar-apostolov: recheck

VelikovPetar (approved):

PR #6664, opened 2026-08-26
Migrate the users endpoints to the generated FullUserResponse

gpunto:
### Goal

Migrate the `queryUsers`, `updateUsers` and `partialUpdateUser` responses to the generated
`FullUserResponse`.

### Implementation

- Replace the hand-written `UsersResponse` and `UpdateUsersResponse` with the generated
  `QueryUsersResponse` and `UpdateUsersResponse`, and point `UserApi` at them. All three endpoints decode
  the same `FullUserResponse`.
- Add mappers for the user's embedded `mutes`, `channel_mutes` and `privacy_settings`, which the previous
  DTO reached through its own types.
- `shadow_banned`, `ban_expires`, `deleted_at` and `revoke_tokens_issued_before` are declared on the
  generated model but were not on the hand-written DTO, so they would stop reaching `User.extraData`. `FullUserResponseAdapter` passes them to
  `alsoKeepInExtraData`, keeping them in the overflow map alongside their own fields. They go away with
  AND-1398. `latest_hidden_channels` is deliberately not kept: all three endpoints build their response
  through `MyUserFromUser`, which clears it before serialization, so it never reaches the payload.

### Notes

`push_preferences` is the one field the hand-written DTO mapped that the generated model does not, so
`User.pushPreference` is always null on these endpoints. Nothing changes in practice: the Go payload tags it
`json:"-"`, so the key was never sent here and the old mapping always produced null too.

`membership_deletion_task_id` is required on the generated `UpdateUsersResponse`. It is always sent, as an
empty string when unset, so the requirement holds.

The user's embedded `devices` are not exercised by these endpoints: the handlers build the response from a
user that carries none, so the list arrives empty here exactly as it did with the previous DTO. The
payload that does populate it is the connect response, which still uses the hand-written own-user DTO.

### Testing

- `GeneratedExtraDataParityTest` pins the five kept keys as an exact map, and fails if the keep set is
  emptied.
- `DomainMappingTest` covers the mapping, an absent name and image, and custom data reaching `extraData`
  with nulls dropped.
- Device-probed on the wire: a sentinel round-tripped through `partialUpdateUser` and survived a following
  `queryUsers`, `shadow_banned` stayed in `extraData` next to genuine custom data, and the user's mutes,
  channel mutes and privacy settings all parsed with their nested users and channel resolved.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -428,6 +431,61 @@ internal class DomainMapping(
...
+     * Transforms the generated channel mute to [ChannelMute].
+     */
+    internal fun io.getstream.chat.android.network.models.ChannelMute.toDomain(): ChannelMute =
+        ChannelMute(
+            user = user?.toDomain(),
+            channel = channel?.toDomain(),
+            createdAt = createdAt,
+            updatedAt = updatedAt,
+            expires = expires,
+        )
```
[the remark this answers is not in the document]
gpunto: Valid, and the gap was wider than the comment says: `randomFullUserResponse` leaves `mutes`, `channelMutes` and `privacySettings` at their defaults, so the `FullUserResponse` test never reached these three mappers at all. Nulling the mute's users and all three privacy settings left the whole client suite green.

Added six cases in `52635b72b8`, each mapper with its populated and its absent shape. The same mutation now fails two of them.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTest.kt
```diff
@@ -2352,7 +2352,7 @@ internal class MoshiChatApiTest {
 
     @ParameterizedTest
     @MethodSource("io.getstream.chat.android.client.api2.MoshiChatApiTestArguments#queryUsersInput")
-    fun testQueryUsers(call: RetrofitCall<UsersResponse>, expected: KClass<*>) = runTest {
+    fun testQueryUsers(call: RetrofitCall<QueryUsersResponse>, expected: KClass<*>) = runTest {
```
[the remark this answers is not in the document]
gpunto: Skipping this one. `MoshiChatApiTest` uses `testXxx` for 130 of its parameterized tests against 11 backtick names, so that is the file's convention for `@MethodSource` tests; renaming one of the 130 would make it less consistent. This PR also only changed the test's type parameter, not the test.

gpunto:
Skipping. These are vendored generated files, copied from the generator output unmodified: any edit here is silently reverted by the next regeneration, so the suppressions cannot be maintained at this end. The suggestion to change the generator template is the right one, but that lives in the backend repo rather than this PR.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
```diff
@@ -623,6 +629,137 @@ internal class DomainMappingTest {
...
+        val settings = with(sut) { PrivacySettingsResponse().toDomain() }
+
+        assertNull(settings.typingIndicators)
+        assertNull(settings.deliveryReceipts)
+        assertNull(settings.readReceipts)
+    }
+
+    @Test
+    fun `FullUserResponse is correctly mapped to User`() {
+        val userResponse = randomFullUserResponse()
```
andremion: `randomFullUserResponse` only sets the required scalars, so this covers 11 of the 24 fields the mapper sets. I hardcoded `privacySettings`, `devices`, `deactivatedAt`, `lastActive`, `mutes`, `teams`, `teamsRole`, `channelMutes`, `blockedUserIds` and `avgResponseTime` to null or empty and the whole client suite stayed green. `UserResponse is correctly mapped to User` above builds a full response and compares the whole `User`, would the same shape work here? Same fixture reason makes the `copy(name = null, image = null)` on 744 a no-op, both are already null.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/FullUserResponseAdapter.kt
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Wire keys the generated model declares and the hand-written user DTO did not, so they reached
+ * `User.extraData` and would otherwise stop doing so. Drop with AND-1398.
+ */
+internal val GENERATED_FULL_USER_EXTRA_DATA_KEYS = setOf(
+    "shadow_banned",
+    "ban_expires",
+    "deleted_at",
+    "revoke_tokens_issued_before",
+    "latest_hidden_channels",
```
andremion: Small thing: all three handlers clear `latest_hidden_channels` before building the response and the key is `omitempty`, so it never reaches the payload on these endpoints. It was not in `extraData` before either, so there is nothing to preserve. Fine to keep as a guard, but the parity test then pins a shape the endpoints do not send. Would a comment be clearer than an entry?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -428,6 +431,61 @@ internal class DomainMapping(
             extraData = extraData.toMutableMap(),
         ).let(userTransformer::transform)
 
+    /**
+     * Transforms [FullUserResponse] to [User].
+     */
+    internal fun FullUserResponse.toDomain(): User =
```
andremion, line 438: Not blocking: `push_preferences` is the one field the old DTO mapped that this one drops, so `User.pushPreference` is always null here now. I checked and the endpoints never send the key, so nothing actually changes. Could be nice to mention it in the description, since the five kept keys cover what the generated model adds and this is what it removes.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -428,6 +431,61 @@ internal class DomainMapping(
...
+            target = target?.toDomain(),
+            createdAt = createdAt,
+            updatedAt = updatedAt,
+            expires = expires,
+        )
+
+    /**
+     * Transforms the generated channel mute to [ChannelMute].
+     */
+    internal fun io.getstream.chat.android.network.models.ChannelMute.toDomain(): ChannelMute =
```
andremion: Nit: any reason not to alias the import? `import io.getstream.chat.android.network.models.ChannelMute as ChannelMuteResponse` would drop the fully qualified name here and the three in the test.

andremion (approved):
Looks good. One test coverage question and a few nits inline, none blocking.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/adapters/FullUserResponseAdapter.kt
gpunto: You're right, and I nearly argued the opposite. I first found the clearing inside `if redactUserInfo` (`clientSideAuth && currentUserID != u.ID`) and concluded it only applied to other users, so the key would still reach a self-query. That is the wrong code path: all three call sites wrap the user first.

```go
payload.NewFullUserResponse(types.MyUserFromUser(updatedUser), ...)   // update_users.go:309, update_users_partial.go:319
chatpayload.NewFullUserResponse(types.MyUserFromUser(u), ...)         // query_users.go:195
```

`MyUserFromUser` sets `LatestHiddenChannels = nil` unconditionally, and the field is `*[]string` with `omitempty`, so it is omitted before redaction is ever consulted. Removed from the keep set and from the parity fixture in `c37cf4b243`, with a comment recording why, so it does not get re-added. The other four entries stand: none of them is cleared by that helper.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
gpunto: Valid, and worse than 11 of 24: the fixture also left `name` and `image` null, which made the `copy(name = null, image = null)` on 744 a no-op asserting `"" == ""`.

Rewrote it to the `UserResponse` shape in `77c585347a`: a fully populated response compared against a whole expected `User`, including devices, mutes, channel mutes and privacy settings. `randomFullUserResponse` now sets `name` and `image`, so the empty-string test actually exercises `.orEmpty()`.

Mutation-checked both: dropping `.orEmpty()` fails 2 tests, emptying `mutes`/`channelMutes` fails 1. Before, either was silent.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
gpunto: Done in `20b5c36f39`, aliased as you suggested. It drops the fully qualified name in the mapper and the three in the test, and there is precedent for it in `DtoMapping` with `Attachment as AttachmentRequest`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
gpunto: Good point, added to the description. It is the one field the old DTO mapped that this model drops, and worth stating next to the keys it adds.

Matches what I found in the payload: the Go field is tagged `json:"-"`, so the key never reached these endpoints and the old mapping produced null here too. No behaviour change, just previously undocumented.

andremion (approved):

PR #6665, opened 2026-08-26
Migrate partialUpdateMember to the generated UpdateMemberPartialResponse model

gpunto:
### Goal

Migrate the `partialUpdateMember` response to the generated `UpdateMemberPartialResponse`.

### Implementation

- Replace the hand-written `UpdateMemberPartialResponse` with the generated model and point
  `ChannelApi.partialUpdateMember` at it. The member itself needs nothing: `ChannelMemberResponse`, its
  collecting adapter and the mapper to the domain `Member` all landed with the queryMembers slice.
- The member is optional in the generated response, so a payload without one becomes a `Result.Failure`
  instead of dropping the update or throwing inside the mapping.

### Testing

- `MoshiChatApiTest` covers the endpoint, including a 200 response carrying no member.
- Device-probed on the wire with a sentinel: a set returned `extraData={memberProbe=<sentinel>}`, an unset
  returned `extraData={}`, and `queryMembers` reported the same value for the member.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of partial member updates when the server response does not include member data.
  * Returns a clear failure instead of causing an application crash.
  * Preserves successful updates and server error handling.

* **Tests**
  * Added coverage for successful responses, server errors, and incomplete responses.


stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
```diff
@@ -2056,6 +2055,20 @@ constructor(
         }
     }
 
+    /**
+     * The member is optional in the response schema, so a missing one is surfaced as a failure rather
+     * than crashing the mapping.
+     */
+    private fun DomainMapping.toMemberCall(response: UpdateMemberPartialResponse): Call<Member> {
```
andremion, line 2063: Not blocking: this is the fourth place in the file branching on a nullable response field into `Success` or `ErrorCall` (queryBannedUsers, push preferences, `mapUserGroup`, and now this). `mapUserGroup` already does it generically. Worth a small shared helper as a follow-up, or do you prefer each one spelled out?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTestArguments.kt
```diff
@@ -376,13 +376,25 @@ internal object MoshiChatApiTestArguments {
...
         Arguments.of(
             RetroError<UpdateMemberPartialResponse>(statusCode = 500).toRetrofitCall(),
             Result.Failure::class,
         ),
+        // The member is optional in the response schema: a 200 without it must fail, not crash.
+        Arguments.of(
+            RetroSuccess(
+                UpdateMemberPartialResponse(duration = randomString(), channelMember = null),
+            ).toRetrofitCall(),
+            Result.Failure::class,
```
andremion, line 396: Small thing: `Call.runCatching` turns a throwable into a `Result.Failure` as well, so this case also passes if the mapping crashes. Asserting the error itself would pin the path the comment describes. The queryBannedUsers case has the same shape, so happy to leave it.

andremion (approved):
Looks good. One leftover though: `api2/model/requests/UpdateMemberPartialResponse.kt` is still in the tree and nothing references it anymore. The earlier slices deleted the hand-written model they replaced, so could you drop it here too? Two classes with that name in different packages is easy to import wrong. Two nits inline, both optional.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Agreed, and it is a follow-up rather than this PR: the same shape is in `queryBannedUsers` (already merged) and the push-preference path, so hoisting it here would make a single-endpoint slice touch three unrelated call sites. `mapUserGroup` is the right template to generalise from.

Filing it so it does not get lost.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/MoshiChatApiTestArguments.kt
gpunto: Good catch, and my comment on that row overclaims: `Call.runCatching` maps any throwable to `Result.Failure(Error.ThrowableError(...))`, so `Result.Failure::class` cannot tell a clean failure from a crash - which is exactly what the comment says it checks.

Added a focused test asserting the error itself. Verified it distinguishes the two: making the mapper throw instead of returning the failure now fails the test, where the parameterized row alone still passed.

`queryBannedUsers` has the same hole. It is merged, so I will fix it alongside the helper follow-up.

gpunto:
You're right, and thanks for spotting it. `ChannelApi` does point at the generated model, so `api2/model/requests/UpdateMemberPartialResponse.kt` had no references left. Removed it in its own commit; the module still compiles, which confirms it was dead.

The import hazard is real, not hypothetical: while doing the attachment slice I renamed an adapter to `AttachmentAdapter` and only later noticed `parser2.direct.AttachmentAdapter` already existed. Renamed that one to `NetworkAttachmentAdapter` for the same reason you give here.

Both nits addressed too: the error assertion is in, and the shared `Success`/`ErrorCall` helper is a follow-up so this stays one endpoint.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/MoshiChatApi.kt
gpunto: Follow-up filed: [AND-1451](https://linear.app/stream/issue/AND-1451/hoist-the-repeated-nullable-response-branching-in-moshichatapi-behind), covering both the shared helper and the `queryBannedUsers` error assertion.

PR #1565, opened 2021-03-16
[CAS-792] Feature - Enable disable replies

leandroBorgesFerreira:
[CAS-792](https://stream-io.atlassian.net/browse/CAS-792)

### Checklist

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [x] PR targets the `develop` branch
- [x] Changelog updated with client-facing changes
- ~[ ] New code is covered by unit tests~
- ~[ ] Comparison screenshots added for visual changes~
- [x] Reviewers added

PR #6666, opened 2026-08-26
Add inline playback for regular audio attachments

### LINKED ISSUE #6611: Add playback for regular audio attachments
Describe the feature you'd like to see in the SDK

We are using the same Stream Chat channel from a web client and an Android app. When someone sends a regular audio file from the web client, it appears as a playable audio attachment on the web. On Android, the same attachment is treated as a regular file and there is no play button.

The built-in Android player works for `audio_recording` attachments created by the voice recorder, but not for regular audio attachments with type `AUDIO`. In the Compose UI, regular audio is currently grouped with file attachments, while the audio player is only used for `AUDIO_RECORDING`.

Please add default playback support for regular audio attachments in the Android UI SDK. The attachment should use its audio URL and title or filename, and provide at least play/pause and progress controls. It should not require recording-only fields such as waveform data.

If a message contains multiple audio attachments, every attachment should remain visible and only one track should be active at a time. Mixed messages should keep rendering their image, video, file and link attachments normally.

This request concerns audio files received from another client. The existing voice-recording composer flow should remain unchanged.

Reproduced with Stream Chat Android 7.6.0.

## Related issues

UIKit: https://github.com/GetStream/stream-chat-swift/issues/4194
SwiftUI: https://github.com/GetStream/stream-chat-swiftui/issues/1554

gpunto:
### Goal

Attachments with `type: "audio"` rendered as a generic file card, so an audio file sent from another platform looked broken on Android while playing inline on web. Give them a first-party inline player in Compose, using the attachment's `assetUrl` and filename, without requiring waveform data or a duration.

Closes #6611
Closes [AND-1360](https://linear.app/stream/issue/AND-1360/regular-audio-attachments-have-no-inline-playback-ui-only-voice)

### Implementation

- An inline audio player: play/pause toggle, file name, elapsed time, seek bar and file type icon, matching the `Mobile / Message View Attachment / Audio File` design. Progress and duration come from the player, so nothing extra is needed on the attachment.
- Audio keeps sharing the file attachment path. The default `FileAttachmentItem` delegates audio to a new `ChatComponentFactory.AudioAttachmentItem` slot, so integrators overriding `FileAttachmentContent` or `FileAttachmentItem` keep rendering audio exactly as they do today. [AND-1439](https://linear.app/stream/issue/AND-1439/give-audio-attachments-their-own-render-path-in-the-next-major-compose) tracks giving audio its own path in the next major, matching iOS and React.
- `AudioPlayerController` accepts `AttachmentType.AUDIO` alongside `audio_recording`, and takes the track duration from the player when the attachment carries none. A seek made before the duration is known is held per track and applied on the first progress update that reports one, so scrubbing an unplayed audio file starts from the thumb rather than from zero. The player divides by the duration it reports, which can be zero for a track it cannot measure, so the progress the UI reads is now sanitised against `NaN`.
- Audio uploads are now typed `audio` instead of `file`, in `AttachmentMetaData`, `StorageHelper` and `AttachmentUploader`. This removes an Android-only inconsistency: stream-chat-js sends audio this way, and iOS's client has mapped audio MIME types to `.audio` since 2022, with its SwiftUI composer following in 5.9.0 ([#1565](https://github.com/GetStream/stream-chat-swiftui/pull/1565)). Voice recordings are unaffected: they set `audio_recording` explicitly and the uploader keeps an already-set type.
- Follow-ons from that type change: the XML quoted attachment factory, view and its sizing branch handle `audio` instead of falling through to the unsupported path and image dimensions, the channel-list preview counts audio in its files bucket, deleting a message stops a playing audio file and not just a voice recording, and the samples' shared-files screens include it.
- All three mappers match audio on the MIME type's top-level type, `startsWith("audio/")`, so a subtype that merely mentions the word does not count. The backend does not derive the type from the MIME type, it consumes whatever the client declares, so this is the client's call. iOS resolves it the same way, by splitting on the slash and comparing the first component. Their existing image and video branches are left exactly as they are. [AND-1452](https://linear.app/stream/issue/AND-1452/unify-the-three-mime-to-attachment-type-mappers-and-tighten-imagevideo) tracks unifying the three, which also means tightening those looser branches.
- Reuse: `PlaybackSlider` is lifted out of `VideoPlaybackControls` and shared, as is the per-attachment playback state both audio players derive from `AudioPlayerState`. Sharing that state also gives voice recordings the player-reported duration as a fallback.

### Notes

Compose only. The XML message list keeps rendering audio as a file card, unchanged, and its quoted attachments are fixed to keep doing the same after the type change, so [AND-1360](https://linear.app/stream/issue/AND-1360/regular-audio-attachments-have-no-inline-playback-ui-only-voice) stays open for the Views side. The `ui-common` player is generalised for both kits, so the XML player has nothing new to learn when it lands.

The typed-as-`audio` change is the one integrator-visible behaviour change. Code switching on `AttachmentType.FILE` to catch audio files needs to accept `AttachmentType.AUDIO` too. iOS shipped the same change as a Changed entry rather than a breaking one, which is why this is labelled `pr:new-feature`. Nothing is removed from the public API. Voice recordings keep their own UI and their playback is unchanged in normal use, but they now share the per-attachment playback state, so they pick up its edge-case handling: the player-reported duration as a fallback, a progress clamp, and a blank asset url counting as no source.

Two known rough edges, both inherited rather than introduced here, and both tracked in AND-1439:

- Tapping an audio row still opens the media preview, since that is currently the only way to open or save an audio attachment. The proper fix is the download action the design specifies, which no platform has shipped yet.
- Long pressing the seek bar does not open the message options, because the shared `dragPointerInput` consumes the gesture. The existing voice recording waveform behaves the same way.

The player shows elapsed time rather than total duration, since a regular audio attachment carries no duration and it is only known once the track loads. iOS loads it from the asset; worth revisiting if we want parity.

### UI Changes

| Before | After | Playing |
| ------ | ----- | ----- |
|  before      |    after | playing |

### Testing

- Snapshot coverage for the player idle, playing, uploading and multiple-attachment states, plus a mixed file and audio message rendered through the real `FileAttachmentContent` to `FileAttachmentItem` to `AudioAttachmentItem` path.
- `AudioPlayerControllerTest`: playback of a regular audio attachment, zero progress when no duration is known, the player-reported duration driving seeks, a seek held from both `play` and `seekTo` and applied once the duration arrives, and scrubbing leaving the track paused.
- `MessageListControllerTests`: deleting a message whose regular audio file is playing stops playback.
- The `isPlayableAudio` guard on every controller entry point, so a non-audio attachment is rejected by each.
- `AttachmentUploaderTests`: an audio mime type is typed `audio`, and an already-set type is preserved.
- `AudioPlaybackTest` pins the per-attachment playback state: no source for a missing or blank url, an attachment duration winning over the player's, a zero duration not counting as known, a stored seek surfacing on a track that is not loaded, and a progress the player could not compute being sanitised.
- `DefaultQuotedAttachmentMessageFactoryTest` covers the quoted attachment types the factory claims, audio included.
- Manually verified on an emulator: attaching an mp3 from the composer sends it as `audio`, and it renders, plays and reports its duration. The Before screenshot is that same message with develop's build installed over the top, so both columns show one message rather than two.


gpunto:
@coderabbitai review

stream-chat-android-ui-common/src/main/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/AudioPlayerController.kt
```diff
@@ -58,15 +65,16 @@ public class AudioPlayerController(
             logger.v { "[resetAudio] rejected (not playing): $audioHash" }
             return
         }
+        pendingSeek = null
         audioPlayer.resetAudio(audioHash)
     }
 
     /**
      * Plays or pauses the audio recording attachment.
      */
```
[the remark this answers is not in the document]
gpunto: Good catch, fixed in 80c376e5. Both KDocs now say "Plays a playable audio attachment: a voice recording or a regular audio file", matching the `isPlayableAudio()` gate they actually use.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/AudioAttachmentContent.kt
```diff
@@ -0,0 +1,319 @@
...
+ * @param modifier Modifier for styling.
+ * @param attachment The attachment to display.
+ * @param playerState The state of the audio player.
+ * @param isMine If the message is from the current user.
+ * @param onPlayToggleClick The callback for when the play button is clicked.
+ * @param onThumbDragStart The callback for when the thumb gets dragged.
+ * @param onThumbDragStop The callback for when the thumb gets released.
+ */
+@Composable
+public fun AudioAttachmentContentItem(
```
andremion, line 114: Most of the missing coverage is in this file and PlaybackSlider, and the snapshots only cover the static rendering. Since this overload takes state and callbacks directly, an interaction test in the style of VoiceMessageReplyContentTest could cover the rest: click on play fires the callback, uploading disables the button, and a swipe on the slider fires the three drag callbacks. The swipe one alone covers most of the slider branches. Worth adding?

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
```diff
@@ -2588,19 +2591,43 @@ public interface ChatComponentFactory {
...
+    /**
+     * Factory method for creating a single regular audio file attachment item, rendered with an inline player.
+     *
+     * Audio attachments are routed through the file attachment path, and [FileAttachmentItem] delegates here for
+     * them, so overriding [FileAttachmentContent] or [FileAttachmentItem] keeps control over audio rendering too.
+     *
+     * @param params Parameters for this component.
+     */
+    @Composable
+    public fun AudioAttachmentItem(params: AudioAttachmentItemParams) {
```
andremion, line 2611: The non-inspection branch here never runs in tests, since the snapshots run in inspection mode. Maybe render a message with an audio attachment through the real factory using the MockedChatClientTest base? That would also cover the defaultAudioPlayerViewModelFactory lines Sonar flags in AudioPlayerViewModel.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/AudioPlayerControllerTest.kt
```diff
@@ -314,6 +372,47 @@ internal class AudioPlayerControllerTest {
         assertEquals(0.25f, controller.state.value.seekTo.getOrDefault(attachment.audioHash, 0f))
     }
 
+    @Test
+    fun `seek made before the duration is known is applied once the player reports it`() {
```
andremion, line 376: Small thing: Sonar shows 5 conditions left in the controller. Three quick ones on the existing fixture would close them: an attachment duration winning over the player-reported one in `seekTo`, a progress update from a different track not applying the pending seek, and `changeSpeed` accepted for a regular audio file.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/content/AttachmentsContentTest.kt
```diff
@@ -106,6 +106,37 @@ internal class AttachmentsContentTest : PaparazziComposeTest {
         }
     }
 
+    @Test
+    fun `file attachment content with an audio file`() {
```
andremion, line 110: Could you also snapshot `FileAttachmentContentWithAudio(isMine = true)`? Covers the remaining isMine conditions in FileAttachmentContent.

andremion (approved):
Some ideas to get the coverage gate green, all inline. The first one alone probably does it.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/attachments/content/AudioAttachmentContent.kt
gpunto: Added as `AudioAttachmentContentInteractionTest`, in 81b30c33: play click reports the attachment, the button is disabled while uploading and when there is no source, and a drag reports the start plus the released progress. Thanks for pointing at `VoiceMessageReplyContentTest` as the template, that made it straightforward.

One thing the drag test turned up. `swipeRight()` reported a released progress of 0. That is real behaviour rather than a test artefact: `dragPointerInput` ends a drag with `onDragStop(null)`, so the slider falls back to its `progress` parameter, which only reflects the drag once it has recomposed. Splitting the gesture with `waitForIdle` between the move and the release makes it observable, and there is a comment in the test saying so.

stream-chat-android-compose/src/main/java/io/getstream/chat/android/compose/ui/theme/ChatComponentFactory.kt
gpunto: Done. The test renders `ChatTheme.componentFactory.AudioAttachmentItem` directly, which covers the non-inspection branch, the view model wiring in `AudioAttachmentItem`, and `defaultAudioPlayerViewModelFactory`. It works because `MockedChatClientTest` mocks with `RETURNS_MOCKS`, so `ChatClient.instance().audioPlayer` hands back a mock without any extra stubbing.

stream-chat-android-ui-common/src/test/kotlin/io/getstream/chat/android/ui/common/feature/messages/list/AudioPlayerControllerTest.kt
gpunto: All three added. The middle one is the most valuable of the set: a pending seek not applying to a different track pins the per-track keying of `pendingSeek`, which only exists because an earlier review pointed out the field was unkeyed. It would have silently regressed otherwise.

stream-chat-android-compose/src/test/kotlin/io/getstream/chat/android/compose/ui/attachments/content/AttachmentsContentTest.kt
gpunto: Added, the snapshot now renders both values so it covers incoming and outgoing in light and dark.

gpunto:
Thanks @andremion, all four are in, pushed as a separate commit (81b30c33) so the delta since your review is easy to read.

Sonar is green: **86.2% coverage on new code**, up from 75.8%. Your first suggestion did most of that on its own, as you guessed.

PR #6668, opened 2026-08-27
Expose upload context with message id to FileUploader

gpunto:
### Goal

Expose the id of the message an attachment belongs to to custom `FileUploader` implementations, so apps uploading to their own CDN can link the uploaded file to the message on their side before the message reaches the Stream API.

Closes [AND-1453](https://linear.app/stream/issue/AND-1453/expose-the-message-id-to-custom-fileuploader-implementations-at-upload)

### Implementation

- Add `FileUploadContext`, a read-only, SDK-constructed context (channel type/id, user id, message id). Its constructor is internal so new properties can be added later without breaking implementers.
- Add `FileUploader.sendFile`/`sendImage` overloads receiving the context. They delegate to the existing overloads by default and compile as JVM default methods (`-Xjvm-default=all`), so current `FileUploader` implementations, Kotlin or Java, keep working unchanged. The API diff is purely additive.
- Thread the message id from `UploadAttachmentsWorker` through `AttachmentUploader` and internal `ChatClient`/`ChatApi` overloads down to `MoshiChatApi`, which builds the context and calls the new overloads. Direct (non-message) uploads pass a null message id.

### Testing

- `MoshiChatApiTest` asserts the uploader receives the context with the right channel/user/message id, null message id for plain uploads, and that the legacy overloads are no longer called directly.
- `AttachmentUploaderTests` covers the message id being forwarded to the client call.
- Manually verified in the compose sample with a custom uploader: the id logged at upload time matches the message id stored server-side, for single and multiple attachments, including the offline → WorkManager retry path.

## Summary by CodeRabbit

* **New Features**
  * File and image uploads can now be associated with a specific message.
  * Custom uploaders receive consolidated upload context, including channel, user, and message details.
  * Existing uploader implementations remain supported through compatibility overloads.
* **Documentation**
  * Updated Java and Kotlin uploader examples to demonstrate message-linked uploads.
* **Tests**
  * Added coverage verifying message ID propagation and context-aware uploader behavior.

### DIFF
--- a/stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/client/helpers/MyFileUploader.java
+++ b/stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/client/helpers/MyFileUploader.java
@@ -19,8 +19,7 @@ public class MyFileUploader implements FileUploader {
     @Override
     public Result<UploadedFile> sendFile(@NotNull FileUploadContext uploadContext, @NotNull File file, @Nullable ProgressCallback callback) {
         try {
-            // uploadContext.getMessageId() is the id of the message this file belongs to. Use it to link
-            // the upload to the message on your own backend before the message is sent.
+            // uploadContext.getMessageId() is the id the message containing this file will be sent with.
             return new Result.Success<>(new UploadedFile("file url", "thumb url"));
         } catch (Exception e) {
             return new Result.Failure(new Error.ThrowableError("Could not send file.", e));
@@ -31,8 +30,7 @@ public class MyFileUploader implements FileUploader {
     @Override
     public Result<UploadedFile> sendImage(@NotNull FileUploadContext uploadContext, @NotNull File file, @Nullable ProgressCallback callback) {
         try {
-            // uploadContext.getMessageId() is the id of the message this image belongs to. Use it to link
-            // the upload to the message on your own backend before the message is sent.
+            // uploadContext.getMessageId() is the id the message containing this image will be sent with.
             return new Result.Success<>(new UploadedFile("url", null));
         } catch (Exception e) {
             return new Result.Failure(new Error.ThrowableError("Could not send image.", e));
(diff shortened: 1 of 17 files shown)


stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/client/helpers/MyFileUploader.kt
```diff
@@ -2,12 +2,27 @@ package io.getstream.chat.docs.kotlin.client.helpers
...
 import io.getstream.result.Error
 import io.getstream.chat.android.models.UploadedFile
+import io.getstream.chat.android.client.uploader.FileUploadContext
 import io.getstream.chat.android.client.uploader.FileUploader
 import io.getstream.chat.android.client.utils.ProgressCallback
 import io.getstream.result.Result
 import java.io.File
 
 class MyFileUploader : FileUploader {
+    override fun sendFile(
```
andremion, line 12: Both samples show only the `sendFile` context overload. Since images are the most common attachment type, should we also show `sendImage`, or add a one-line comment saying it works the same way?

andremion (approved):

stream-chat-android-docs/src/main/kotlin/io/getstream/chat/docs/kotlin/client/helpers/MyFileUploader.kt
gpunto: Good point, went with showing it: both samples now override the context-aware `sendImage` as well (a364a4edeb).

stream-chat-android-docs/src/main/java/io/getstream/chat/docs/java/client/helpers/MyFileUploader.java
```diff
@@ -8,12 +8,25 @@
...
 public class MyFileUploader implements FileUploader {
+    @Nullable
+    @Override
+    public Result<UploadedFile> sendFile(@NotNull FileUploadContext uploadContext, @NotNull File file, @Nullable ProgressCallback callback) {
+        try {
+            // uploadContext.getMessageId() is the id of the message this file belongs to. Use it to link
+            // the upload to the message on your own backend before the message is sent.
+            return new Result.Success<>(new UploadedFile("file url", "thumb url"));
+        } catch (Exception e) {
+            return new Result.Failure(new Error.ThrowableError("Could not send file.", e));
```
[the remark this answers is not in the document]
gpunto: Added the context-aware `sendImage` override to both the Kotlin and Java samples in a364a4edeb.

PR #6669, opened 2026-08-28
Expose upload context with message id to FileUploader (develop)

gpunto:
### Goal

Port of #6668 to develop: expose the id of the message an attachment belongs to to custom `FileUploader` implementations, so apps uploading to their own CDN can associate the uploaded file with the message before it reaches the Stream API.

Part of [AND-1453](https://linear.app/stream/issue/AND-1453/expose-the-message-id-to-custom-fileuploader-implementations-at-upload)

### Implementation

- Add `FileUploadContext`, a read-only, SDK-constructed context (channel type/id, user id, message id). Its constructor is internal so new properties can be added later without breaking implementers.
- Add `FileUploader.sendFile`/`sendImage` overloads receiving the context. They delegate to the existing overloads by default and compile as JVM default methods, so current `FileUploader` implementations, Kotlin or Java, keep working unchanged. The API diff is purely additive.
- Thread the message id from `UploadAttachmentsWorker` through `AttachmentUploader` and internal `ChatClient`/`ChatApi` overloads down to `MoshiChatApi`, which builds the context and calls the new overloads. Direct (non-message) uploads pass a null message id.

Differences from the v6 version are mechanical: the state-module tests live under `stream-chat-android-client` here, and the worker's parallel upload path is preserved.

### Testing

- Same coverage as #6668: `MoshiChatApiTest` (context contents, null message id for plain uploads), `AttachmentUploaderTests` (file and image forwarding), `ChatClientChannelFileUploaderTests` (overload forwarding), `FileUploaderTest` (default delegation contract), `FileUploadContextTest`, and the worker-to-uploader assertion in `UploadAttachmentsIntegrationTests`.

## Summary by CodeRabbit

* **New Features**
  * File and image uploads can now be associated with a specific message.
  * Added upload context support for channel, user, and message details.
  * Existing upload methods remain available for compatibility.
  * Attachment uploads now preserve message associations throughout the upload process.

* **Documentation**
  * Updated Java and Kotlin uploader examples for the new upload context.

* **Tests**
  * Added coverage verifying message ID forwarding and upload context behavior.


gpunto:
@coderabbitai Regarding the two findings:

- **Retain the previous `uploadAttachment` overload**: not applying. `uploadAttachment` is `@InternalStreamChatApi`, which this repo lists in the binary validator's `nonPublicMarkers` — it is deliberately outside the compatibility surface, and all of its call sites (the worker and the tests) were updated in the same change. JVM descriptor stability is not a guarantee we extend to internal-annotated API, and keeping a legacy overload would freeze an internal signature that was intentionally reordered for consistency across layers.
- **Document the suppression reason**: done in ea9f51f315 (and mirrored to the v6 PR).

andremion (approved):

PR #6672, opened 2026-08-28
Migrate the reaction read responses to the generated ReactionResponse models

gpunto:
### Goal

Migrate the reaction read responses (`getReactions`, `queryReactions`) from the hand-written DTOs to the generated network models.

### Implementation

- Add generated `ReactionResponse`, `GetReactionsResponse` and `QueryReactionsResponse`; remove the hand-written `ReactionsResponse` and `QueryReactionsResponse`. `sendReaction`/`deleteReaction` keep their hand-written wrapper, their generated responses embed `MessageResponse`.
- Add `ReactionResponseAdapter` so the flattened custom data the v1 endpoints send is collected into `custom`.
- Map `ReactionResponse` to the domain `Reaction` in `DomainMapping`. `emoji_code` is custom data on the backend rather than a declared field (the handler itself reads `reaction.Custom["emoji_code"]`), so it arrives inside `custom`; the mapper promotes it to `Reaction.emojiCode` and filters it out of `extraData`, matching what the hand-written DTO produced.

### Testing

- `ReactionResponseParsingTest` covers the response, the nested user and the collected custom fields.
- `DomainMappingTest` asserts `emoji_code` is promoted to `emojiCode` and does not appear in `extraData`.
- Device-probed both endpoints on the wire: sent a reaction carrying `emoji_code`, then `getReactions` and `queryReactions` both returned `emojiCode=<emoji>` with `extraData={}`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reaction retrieval and parsing from chat responses.
  * Preserved custom reaction data, including emoji codes and additional fields.
  * Added support for reaction metadata such as duration and nested user details.

* **Tests**
  * Expanded coverage for reaction retrieval, parsing, custom fields, and domain mapping.

### DIFF
--- a/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
+++ b/stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
@@ -145,6 +145,7 @@ import io.getstream.chat.android.network.models.PollVotesResponse
 import io.getstream.chat.android.network.models.PrivacySettingsResponse
 import io.getstream.chat.android.network.models.PushPreferencesResponse
 import io.getstream.chat.android.network.models.QueryPollsResponse
+import io.getstream.chat.android.network.models.ReactionResponse
 import io.getstream.chat.android.network.models.UnreadCountsChannel
 import io.getstream.chat.android.network.models.UnreadCountsChannelType
 import io.getstream.chat.android.network.models.UnreadCountsThread
@@ -489,6 +490,27 @@ internal class DomainMapping(
             updatedAt = updatedAt,
             expires = expires,
         )
+
+    /**
+     * Transforms [ReactionResponse] into [Reaction].
+     *
+     * `emoji_code` is custom data on the backend rather than a declared field, so it arrives inside
+     * [ReactionResponse.custom] and is promoted here, staying out of [Reaction.extraData].
+     */
+    internal fun ReactionResponse.toDomain(): Reaction = Reaction(
+        createdAt = createdAt,
+        messageId = messageId,
+        score = score,
+        type = type,
+        updatedAt = updatedAt,
+        user = user.toDomain(),
+        userId = userId,
+        emojiCode = custom[EMOJI_CODE_KEY] as? String,
+        extraData = custom
+            .filterKeys { it != EMOJI_CODE_KEY }
+            .mapNotNull { (key, value) -> value?.let { key to it } }
+            .toMap(),
+    )
 
     /**
      * Transforms [DownstreamReactionDto] to [Reaction].
@@ -1383,6 +1405,9 @@ internal class DomainMapping(
     )
 
     private companion object {
+        /** `emoji_code` is sent as custom data rather than a declared field. */
+        private const val EMOJI_CODE_KEY = "emoji_code"
+
         private const val FIELD_LAST_MESSAGE_AT = "last_message_at"
         private const val FIELD_LAST_UPDATED = "last_updated"
     }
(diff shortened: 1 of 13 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
```diff
@@ -1703,4 +1704,17 @@ internal class DomainMappingTest {
             )
         }
     }
+
+    @Test
+    fun `ReactionResponse maps emoji_code out of custom and keeps it out of extraData`() {
```
andremion, line 1745: Could be nice to add the full `ReactionResponse is correctly mapped to Reaction` test too, like the `DownstreamReactionDto` one above. Right now only `emojiCode` and `extraData` are asserted, so a swapped `createdAt`/`updatedAt`, or `userId` coming from the wrong field, would still pass. Not blocking.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
```diff
@@ -1386,6 +1387,21 @@ internal object Mother {
...
+        type: String = randomString(),
+        custom: Map<String, Any?> = emptyMap(),
+    ): ReactionResponse = ReactionResponse(
+        messageId = messageId,
+        type = type,
+        score = randomInt(),
+        createdAt = randomDate(),
+        updatedAt = randomDate(),
+        userId = randomString(),
+        user = randomUserResponse(),
```
andremion: Small thing: `user` gets its own random id here, so `userId` and `user.id` never match. `randomDownstreamReactionDto` just above ties them together with `randomDownstreamUserDto(id = userId)`. Maybe expose `userId` and pass `randomUserResponse(id = userId)`, so the fixture looks like what the API actually sends?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
```diff
@@ -1383,6 +1405,9 @@ internal class DomainMapping(
     )
 
     private companion object {
+        /** `emoji_code` is sent as custom data rather than a declared field. */
+        private const val EMOJI_CODE_KEY = "emoji_code"
```
andremion: Not blocking: the other direction still uses the plain string, in `Reaction.toDto()` in `DtoMapping` (`extraData + ("emoji_code" to emojiCode)`). Worth keeping the key in one place both directions can read, so they can't drift?

andremion (approved):
Looks good. Three optional nits inline.

stream-chat-android-client/src/main/java/io/getstream/chat/android/network/models/GetReactionsResponse.kt
```diff
@@ -14,12 +14,25 @@
  * limitations under the License.
  */
 
-package io.getstream.chat.android.client.api2.model.response
+@file:Suppress(
+    "ArrayInDataClass",
+    "EnumEntryName",
+    "RemoveRedundantQualifierName",
+    "UnusedImport",
+)
```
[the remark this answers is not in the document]
andremion: This block is what the model generator emits, and every file in `network/models` carries it. A hand edit here would be lost the next time the models are regenerated, so I'd leave it as is.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
gpunto: Added, asserting the whole object. One catch: with `userId` and `user.id` tied in the fixture, that test cannot tell which field the mapper reads, so I added a small separate test where the two ids differ to pin it. Rewriting the mapper to `userId = user.id` now fails.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
gpunto: Done, `randomReactionResponse` takes a `userId` and passes `randomUserResponse(id = userId)`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DomainMapping.kt
gpunto: `EMOJI_CODE_KEY` is now package-visible and `Reaction.toDto()` uses it, so both directions read the same constant. Left the hand-written reader in `parser2/direct/ReactionAdapter` on the literal since it goes away with AND-1398.

PR #6673, opened 2026-08-31
Map invisible to null when the wire omits it

gpunto:
### Goal

`invisible` is only ever sent for the connected user. The backend marks it `ignore_if_client_side` on the shared user payload and declares it unconditionally only on the own-user response, so for every other user the field never reaches a client-side SDK.

The hand-written DTOs defaulted it to `false` regardless, asserting something the server did not send. Endpoints already migrated to the generated `UserResponse` return `null`, because the generated model has no such field. The two parse paths disagreed on the same payload, and the disagreement is observable: the same message parsed over HTTP and over the websocket produced users differing on this field.

Closes [AND-1458](https://linear.app/stream/issue/AND-1458/nested-users-now-map-invisible-to-null-instead-of-false-decide-whether)

### Implementation

- Default `DownstreamUserDto.invisible` and the direct `UserAdapter` to `null` rather than `false`, so all parse paths agree with the wire.
- Move the expectations whose fixture JSON does not declare `invisible` to `null`. Fixtures that do send it are unchanged: `UserTestData.expectedAllFields` (`true`), `expectedWithExplicitNulls` (`null`), and the event fixtures, whose JSON carries the field.

### Consumer impact

- `User.invisible` returns `null` instead of `false` for any user other than the connected one. The declared type is unchanged (`Boolean?`) and the API diff is empty.
- `User.isInvisible` is `invisible == true`, so it still returns `false` and code using the accessor is unaffected.
- The connected user is unaffected, since the connect payload always carries the field.
- Offline storage flattens the value: `UserEntity.invisible` is non-null and written from `isInvisible`, so a user read back from the database reports `false`.
- Client-side sorting on `invisible` orders `null` before `false`, as `getComparableField` exposes the raw nullable.

### Testing

- Device probe against a live app: the connected user keeps a non-null `invisible` from the connect payload, while members, reads, message senders, poll voters and poll creators across three channels all report `null`.
- Reverting either default to `false` fails 22 tests, so the value is pinned on both the DTO and the direct parse path.
- `UserTestData` covers the three wire cases explicitly: field sent as `true`, sent as `null`, and absent.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected user visibility data handling when the `invisible` field is omitted.
  - Omitted values now remain `null` instead of being incorrectly interpreted as `false`.
  - Preserved explicit visibility values when provided by the server.

- **Tests**
  - Updated message, reaction, poll, event, and user parsing scenarios to reflect the corrected nullable behavior.


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/UserTestData.kt
```diff
@@ -171,7 +171,7 @@ internal object UserTestData {
         name = "",
         image = "",
         role = "admin",
-        invisible = false,
+        invisible = null,
```
andremion, line 174: Small thing: some fixtures outside this diff still say `invisible = false` where their JSON leaves the field out: `MemberTestData.kt:57`, `:75`, `ChannelTestData.kt:308`, `:327`, `:332`, `:336`, `:362`, `ChannelUserReadTestData.kt:56`, `:74`, `VoteTestData.kt:63`, `AnswerTestData.kt:59`. No test references those objects today, so nothing fails, but whoever wires them up later gets a failure that looks like a regression. Same for the `invisible = false` default in `randomDownstreamUserDto` (`Mother.kt:393`), which no longer matches the DTO. Worth flipping both while you are here?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/UserAdapter.kt
```diff
@@ -42,7 +42,7 @@ internal class UserAdapter(
         var name: String? = null
         var image: String? = null
         var role: String? = null
-        var invisible: Boolean? = false // DTO default
+        var invisible: Boolean? = null // the wire only sends invisible for the own user
```
andremion, line 45: Could be nice: the `@param invisible` KDoc on `User` (`User.kt:30`) still only says it can be changed while connecting. Since the value now stays null for everyone but the connected user, saying that there would point integrators at `isInvisible` instead of reading the field directly. Happy to leave it if you would rather keep this PR to the parse paths.

andremion (approved):
Looks good. Two nits inline, plus one scope question.

The DB entity and the outgoing user requests still flatten this to false. Any plan to follow up on those, or not worth it while nothing in the SDK reads the field?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/UserTestData.kt
gpunto: Good catch on `randomDownstreamUserDto`, flipped it to null so the helper matches the DTO.

The fixture list is the same set I ran into: `MemberTestData`, `ChannelTestData`, `ChannelUserReadTestData`, `VoteTestData` and `AnswerTestData` have no references anywhere in the module, and there is no adapter in `parser2/direct` for any of those types, so nothing can consume them. #6675 deletes them rather than updating expectations in files nothing runs. If you would rather keep them and fix the values, I can drop that PR and do it here instead.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/parser2/direct/UserAdapter.kt
gpunto: Done, the `@param invisible` KDoc now says it is only populated for the connected user and points at `isInvisible`.

gpunto:
On the scope question: neither is changed by this PR, but they are worth separate follow-ups for different reasons.

**Outgoing requests** are unaffected here, because both `UpstreamUserDto` and `UserRequest` are built from `isInvisible`, which is false for null and for false alike. There is a latent issue though: the generated `UserRequest.invisible` is nullable, so it could carry "unspecified" through as an omitted field, while today we always assert a concrete `false`. That means connecting with a `User` that never set the field sends `invisible: false` rather than leaving it alone. Changing it would change what we send, so it does not belong in a parse-path PR.

**The DB entity** is the one with a visible consequence. `UserEntity.invisible` is a non-null `Boolean` written from `isInvisible`, so null flattens to false on the way in and reads back as false. Before this PR network and cache agreed on false; after it, the same user reads null fresh from the network and false once cached. `isInvisible` still hides the difference, but anyone comparing `user.invisible == false` gets an answer that depends on where the object came from. Making the column nullable fixes it and keeps own-user fidelity, since `isInvisible` is lossless for a user whose value is always set. The cost is a Room schema bump, and the DB uses `fallbackToDestructiveMigration`, so it drops the offline cache. Worth batching with the next users-table change rather than spending a wipe on a field whose only in-SDK reader is `isInvisible`.

Happy to file both, or to pull the DB one into this PR if you would rather not leave the asymmetry in place at all.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/UserTestData.kt
andremion: Deleting them is better, and I get the same result: no references to any of the five. Let's keep #6675 as it is.

andremion:
Both as follow-ups sounds right. The destructive migration decides it for me: batching the nullable column with the next users-table change is better than spending a cache wipe on a field whose only reader in the SDK is `isInvisible`. No need to pull it into this PR.

Same for the request side, changing what we send does not belong in a parse-path PR.

PR #6674, opened 2026-08-31
Keep the loader up while a grouped channel list fetches its first page

gpunto:
### Goal

A grouped channel list with nothing cached for a group showed the empty state ("no channels") for the whole first `queryGroupedChannels` round trip instead of the loader. Standard channel lists do not behave this way: they hold the loader until the network answers.

Closes [AND-1460](https://linear.app/stream/issue/AND-1460/grouped-channel-list-shows-the-empty-state-instead-of-the-loader-while)

Targets `v6`.

### Implementation

- `loadOfflineGroupedChannels` cleared the first-page loading flag unconditionally, so a cache miss moved `channelsStateData` from `Loading` to `OfflineNoResults` while the request was still in flight, and nothing raised it again: `onQueryGroupedChannelsRequest` only captured the query config.
- Raise the flag in the request listener, the point that knows a fetch started, for each named first-page group. The guard is `!hasCompletedAQuery && getChannels().isNullOrEmpty()` and both halves are load-bearing. Without the marker a settled empty tab returns to a spinner on every reconnect, since `SyncManager` recovers grouped lists through a cursor-less request that reaches this listener. Without the emptiness check a request hides channels the offline read just loaded from cache, because `ChannelsStateData` reports `Loading` whenever the flag is set regardless of content.
- `hasCompletedAQuery` is needed because state alone cannot express the difference: a group that has never loaded and one that loaded and is genuinely empty both hold an empty channel map.
- A failure does not count as completion, so a retry raises the loader again, matching `queryOffline` for a standard list. An omitted group would count, since the server answered, but that branch is a guard rather than a live path: the server returns every requested group, with an empty channel list when the group has nothing. It stays because such a group would never reach `applyGroupedResult` and the loader it raised would never come down.
- Clear the flag only once there is something to show, otherwise the offline read immediately undoes the raise. End the load on failure, and defensively for a group the response left out, neither of which reaches `applyGroupedResult`; ending it initialises the channel map as well as clearing the flag, because `ChannelsStateData` reports `Loading` while channels are still null.

Scope is limited to grouped lists: the new methods are called only from `QueryGroupedChannelsListenerState`, and `loadOfflineGroupedChannels` returns early for non-grouped identifiers.

Two behaviours worth calling out. Recovery of a group whose first page *failed* raises the loader on grouped, where standard recovery never does, because `queryFirstPage` bypasses the listener entirely. That is deliberate: a spinner while the retry runs is more accurate than a stale empty state. And a request with `groups = null` keeps today's behaviour, since the keys are unknown until the response, so the mechanism is inert rather than half-engaged and cannot leave a list loading.

### Testing

- Unit tests over a real `QueryChannelsMutableState` assert the emitted `ChannelsStateData` rather than setter calls: an in-flight first page on a cache miss stays `Loading`, a request arriving after the offline read still raises, and a failed load ends on `OfflineNoResults`. Reinstating the unconditional clear, or the earlier `== null` guard, fails them.
- Mock tests cover the raise per first-page group, no raise while paginating, no raise over cached channels or a settled empty group, a failed page staying retryable, and the load being ended on failure and for an omitted group.
- Verified on a grouped-channels sample built against this branch, cold start with an empty database on a throttled network:
  - before: `OfflineNoResults -> Result(10) -> Result(20)`
  - after: `Loading -> Result(10) -> Result(20)`

  Also ran the sample with the ordering flipped so the offline read runs before the query, which is the case the completion marker adds: `Loading -> Result`, so the new raise clears. A warm start still renders straight from cache.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,3 +11,6 @@ android.suppressUnsupportedCompileSdk=34
 
 # Project version
 version=6.42.1
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,6 +11,3 @@ android.suppressUnsupportedCompileSdk=34
 
 # Project version
 version=6.42.1
-
-# Enabled parallel sync for Gradle 9.4+
-org.gradle.tooling.parallel=true
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/listener/internal/QueryGroupedChannelsListenerState.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/listener/internal/QueryGroupedChannelsListenerState.kt
@@ -43,16 +43,36 @@ internal class QueryGroupedChannelsListenerState(
         // parameters. When `groups == null` the caller is relying on the server's default group
         // set — we don't know the keys until the response, so we defer to the result-side capture
         // (which only fires on success, but in that case there is no failure to recover from).
+        // Only named groups get a loader: with `groups == null` the keys are unknown until the response.
         groups?.forEach { (key, groupQuery) ->
-            logic.queryChannels(QueryChannelsIdentifier.Grouped(key))
-                .setGroupedQueryConfig(
-                    GroupedQueryConfig(
-                        limit = limit,
-                        pageSize = groupQuery.limit,
-                        watch = watch,
-                        presence = presence,
-                    ),
-                )
+            val queryLogic = logic.queryChannels(QueryChannelsIdentifier.Grouped(key))
+            queryLogic.setGroupedQueryConfig(
+                GroupedQueryConfig(
+                    limit = limit,
+                    pageSize = groupQuery.limit,
+                    watch = watch,
+                    presence = presence,
+                ),
+            )
+            // The only point that knows a fetch is starting: grouped state is initialised separately
+            // from the query, so the offline load cannot tell an in-flight first page from no fetch.
+            if (groupQuery.isFirstPage()) {
+                queryLogic.startLoadingFirstPageIfNeverLoaded()
+            }
+        }
+    }
+
+    private fun GroupedChannelsGroupQuery.isFirstPage(): Boolean = next == null && prev == null
+
+    /** Ends the first-page load for every [keys] entry that was requested as a first page. */
+    private suspend fun finishFirstPageLoads(
+        keys: Set<String>,
+        groups: Map<String, GroupedChannelsGroupQuery>?,
+    ) {
+        keys.forEach { key ->
+            if (groups?.get(key)?.isFirstPage() == true) {
+                logic.queryChannels(QueryChannelsIdentifier.Grouped(key)).finishFirstPageLoad()
+            }
         }
     }
 
@@ -63,10 +83,14 @@ internal class QueryGroupedChannelsListenerState(
         watch: Boolean,
         presence: Boolean,
     ) {
-        if (result !is Result.Success) return
+        if (result !is Result.Success) {
+            // The success path ends the load while applying the result, so failures must do it here.
+            finishFirstPageLoads(groups.orEmpty().keys, groups)
+            return
+        }
 
         // Only the first page carries initial unread counts.
-        val isFirstPageRequest = groups.orEmpty().values.none { it.next != null || it.prev != null }
+        val isFirstPageRequest = groups.orEmpty().values.all { it.isFirstPage() }
         if (isFirstPageRequest) {
             val next = groupedUnreadChannelsUpdater.calculateUpdatedCounts(
                 current = globalState.groupedUnreadChannels.value,
@@ -78,11 +102,14 @@ internal class QueryGroupedChannelsListenerState(
         // Route each returned group's channels into the per-group state. The captured config lets
         // both ChannelListViewModel.loadMoreGroupedChannels and SyncManager.updateGroupedQueryChannels
         // reuse the caller's original parameters on paginated and recovery calls respectively.
+        // A requested group the response omits never reaches applyGroupedResult, so end it here.
+        finishFirstPageLoads(groups.orEmpty().keys - result.value.groups.keys, groups)
+
         result.value.groups.forEach { (key, group) ->
             // A request without `next`/`prev` cursors for this key (or no per-group query at all)
             // is a first-page request → replace channels. With a cursor → paginated → append.
             val perGroupQuery = groups?.get(key)
-            val isFirstPage = perGroupQuery?.let { it.next == null && it.prev == null } ?: true
+            val isFirstPage = perGroupQuery?.isFirstPage() ?: true
             val perGroupLimit = perGroupQuery?.limit
             val queryLogic = logic.queryChannels(QueryChannelsIdentifier.Grouped(key))
             queryLogic.setGroupedQueryConfig(
--- a/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/listener/internal/QueryGroupedChannelsListenerState.kt
+++ b/stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/listener/internal/QueryGroupedChannelsListenerState.kt
@@ -67,10 +67,11 @@ internal class QueryGroupedChannelsListenerState(
     private suspend fun finishFirstPageLoads(
         keys: Set<String>,
         groups: Map<String, GroupedChannelsGroupQuery>?,
+        completed: Boolean,
     ) {
         keys.forEach { key ->
             if (groups?.get(key)?.isFirstPage() == true) {
-                logic.queryChannels(QueryChannelsIdentifier.Grouped(key)).finishFirstPageLoad()
+                logic.queryChannels(QueryChannelsIdentifier.Grouped(key)).finishFirstPageLoad(completed)
             }
         }
     }
@@ -84,7 +85,7 @@ internal class QueryGroupedChannelsListenerState(
     ) {
         if (result !is Result.Success) {
             // The success path ends the load while applying the result, so failures must do it here.
-            finishFirstPageLoads(groups.orEmpty().keys, groups)
+            finishFirstPageLoads(groups.orEmpty().keys, groups, completed = false)
             return
         }
 
@@ -102,7 +103,7 @@ internal class QueryGroupedChannelsListenerState(
         // both ChannelListViewModel.loadMoreGroupedChannels and SyncManager.updateGroupedQueryChannels
         // reuse the caller's original parameters on paginated and recovery calls respectively.
         // A requested group the response omits never reaches applyGroupedResult, so end it here.
-        finishFirstPageLoads(groups.orEmpty().keys - result.value.groups.keys, groups)
+        finishFirstPageLoads(groups.orEmpty().keys - result.value.groups.keys, groups, completed = true)
 
         result.value.groups.forEach { (key, group) ->
             // A request without `next`/`prev` cursors for this key (or no per-group query at all)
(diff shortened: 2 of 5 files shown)


gpunto:
@coderabbitai review

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
```diff
@@ -147,6 +153,37 @@ internal class QueryChannelsLogic(
...
+    /**
+     * Marks a grouped first page as loading, but only for a group that has never loaded.
+     *
+     * The guard is `== null` rather than `isNullOrEmpty()`: a group that loaded and came back empty
+     * holds a non-null empty map, and re-raising for it would put every settled empty tab back into
+     * a spinner on each reconnect, since `SyncManager` recovers grouped lists through this listener.
+     *
+     * Shares [groupedResultMutex] so the check cannot read a half-applied update.
+     */
+    internal suspend fun startLoadingFirstPageIfNeverLoaded() {
```
andremion, line 180: `getChannels() == null` is doing double duty here: "never loaded" and "nothing has initialised the map yet". `initializeChannelsIfNeeded()` in the offline read flips it, so the loader only goes up when the request listener runs before that read. In the Compose path it does today, but that is timing rather than a rule, and an app that renders the list first and queries on a user action still gets the old behaviour.

Would a small in-flight counter per group work better? Increment in `onQueryGroupedChannelsRequest`, decrement in the result listener, and let `loadOfflineGroupedChannels` set the flag to `inFlight > 0` on a miss. Both orderings then behave the same, and it splits "never loaded" from "a fetch is running". Fine as a follow-up too, just raising it while the code is fresh.

stream-chat-android-state/src/test/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogicGroupedTest.kt
```diff
@@ -372,6 +458,48 @@ internal class QueryChannelsLogicGroupedTest {
...
+            ),
+        ) doReturn null
+
+        // When — the request starts, then the offline read misses
+        realLogic.startLoadingFirstPageIfNeverLoaded()
+        realLogic.loadOfflineGroupedChannels()
+
+        // Then — still Loading. Reinstating the unconditional clear in loadOfflineGroupedChannels
+        // turns this into OfflineNoResults, which is the flash this change removes.
+        assertEquals(ChannelsStateData.Loading, mutableState.channelsStateData.value)
```
andremion, line 514: Could be nice: this harness is real state, so it could cover the failure end too. Adding `realLogic.finishFirstPageLoad()` after this and asserting `OfflineNoResults` would give you raise, miss, failure in one test. That chain is only checked against mocks right now.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
```diff
@@ -147,6 +153,37 @@ internal class QueryChannelsLogic(
...
+     * Ends a grouped first-page load that produced no channels of its own: a failed request, or a
+     * requested group the response left out.
+     *
+     * Both steps are needed to leave `Loading`, which [ChannelsStateData] reports while the flag is
+     * set *or* while channels are still null.
+     */
+    internal suspend fun finishFirstPageLoad() {
+        groupedResultMutex.withLock {
+            queryChannelsStateLogic.initializeChannelsIfNeeded()
+            loadingPerPage(isLoading = false, hasOffset = false)
```
andremion: Small thing: with `hasOffset` fixed at false this is just `setLoadingFirstPage(false)`. Calling the setter directly reads a bit clearer?

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/listener/internal/QueryGroupedChannelsListenerState.kt
```diff
@@ -43,16 +43,36 @@ internal class QueryGroupedChannelsListenerState(
         // parameters. When `groups == null` the caller is relying on the server's default group
         // set — we don't know the keys until the response, so we defer to the result-side capture
         // (which only fires on success, but in that case there is no failure to recover from).
+        // Only named groups get a loader: with `groups == null` the keys are unknown until the response.
```
andremion: Nit: this repeats the paragraph right above it about `groups == null` and unknown keys. Drop it?

andremion (approved):
Looks good. One question on the ordering inline, plus a couple of nits.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
gpunto: Good catch, and my first answer to this was wrong, so thanks for pushing.

I prototyped the counter exactly as you described (increment on request, decrement on result, offline read setting the flag from `inFlight > 0` on a miss) and drove both orderings against real state:

```
order A (request first):     Loading
order B after offline read:  OfflineNoResults
order B after request:       OfflineNoResults   <- unchanged
```

Order B does not move, because `initializeChannelsIfNeeded()` has already made the map non-null by the time the request arrives, so the raise is skipped, and the only place reading the counter is the offline read, which has already run.

But the counter turned out not to be the missing piece. What was missing is a "has a query ever finished" marker, and once you have that the counter is redundant. Implemented in c68cf78d80: `hasCompletedAQuery`, set when a grouped result is applied and when a load is ended. The guard is now `!hasCompletedAQuery && getChannels().isNullOrEmpty()`.

Both halves are load-bearing. Without the marker, a settled empty group goes back to a spinner on every reconnect, since `SyncManager` recovers grouped lists with a cursor-less request through this listener. Without the emptiness check, a request hides channels the offline read just loaded from cache behind a spinner, because `ChannelsStateData` reports `Loading` whenever the flag is set regardless of content.

Order B now reaches `Loading` instead of showing the empty state for the whole round trip. There is a real-state test for it that fails against the previous guard, and I verified it on the sample with the ordering flipped: cold start goes `Loading -> Result`, so the new raise clears.

One judgement call in there: a failed query counts as completion, so a retry shows the empty state rather than returning to the loader. Happy to flip that if you disagree.

stream-chat-android-state/src/test/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogicGroupedTest.kt
gpunto: Done. It now asserts `OfflineNoResults` after `finishFirstPageLoad()`, so raise, miss and failure are covered against real state in one test rather than leaving the failure end to mocks.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
gpunto: Agreed, `hasOffset` was fixed at false so the branch was dead. Calling `setLoadingFirstPage(false)` directly now, matching the neighbouring writers.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/listener/internal/QueryGroupedChannelsListenerState.kt
gpunto: You are right, the paragraph above already covers `groups == null` and the unknown keys. Dropped.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
andremion: On the judgement call, I would lean toward flipping it. A standard list does raise the loader on a user retry after a failed empty first page: `queryOffline` sees `currentLoading` as false (empty channel map, loading false) and calls `loadingPerPage(true, ...)`. Grouped would now stay on the empty state, so the two differ on that path. Recovery is the case where standard also does not raise, so that one matches either way.

Only catch is that `finishFirstPageLoad` also runs for a group the response left out, and that one really is a completion, so it would need to tell the two apart. Not blocking if you would rather leave it as is.

gradle.properties
```diff
@@ -11,3 +11,6 @@ android.suppressUnsupportedCompileSdk=34
 
 # Project version
 version=6.42.1
+
+# Enabled parallel sync for Gradle 9.4+
+org.gradle.tooling.parallel=true
```
andremion: This one looks unrelated to the loader change, probably a local setting that came along by accident. Drop it from the PR?

andremion (commented):
One stray file in d656e3867b, inline.

gradle.properties
gpunto: Man this thing keeps popping back up 🤦

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
gpunto: Flipped, and you are right about the standard path. I checked: on failure `onQueryChannelsResult` calls `initializeChannelsIfNeeded()` and then `loadingPerPage(false, ...)`, so the map is empty and loading is false. On the retry `currentLoading` resolves to `loading`, `queryOffline` does not early-return, and it raises. Grouped would have diverged.

Good catch on the omitted-group case too. `finishFirstPageLoad` now takes `completed`: true for a group the response left out, since that group has been answered, and false for a failure, which leaves it never-loaded so a retry raises the loader again. Recovery matches either way, since standard does not raise there either.

Added a test that a failed first page stays retryable. In 11b643306f.

gradle.properties
gpunto: Reverted in 11b643306f. It was `org.gradle.tooling.parallel=true`, a local Studio setting that a `git add -A` of mine swept in. Sorry for the noise.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
andremion: Small correction to the rationale, not the code: with `completed = false` a failed group stays never-loaded, so SyncManager recovery does raise the loader for it, while standard recovery goes through `queryFirstPage()` and never sets the flag. They differ on that path.

I think grouped has the better behaviour there, a spinner while the retry runs is more accurate than a stale empty state, so nothing to change. Only worth fixing the "recovery matches either way" line in the commit message.

gradle.properties
andremion: No worries, thanks for the quick revert.

stream-chat-android-state/src/main/java/io/getstream/chat/android/state/plugin/logic/querychannels/internal/QueryChannelsLogic.kt
gpunto: You are right, and I had it backwards. With `completed = false` a failed group stays never-loaded, so grouped recovery does raise the loader for it, while standard recovery goes through `queryFirstPage` and never touches the flag. They differ on exactly that path.

Agreed the grouped behaviour is the better one, so leaving the code as is.

I have not amended the commit message, since the branch is already reviewed and I would rather not force-push over it. The PR description was stale anyway (it still described the earlier `== null` guard), so I have rewritten it to match the final state, including this recovery difference and why it is deliberate. The develop port carries the corrected wording from the start.

PR #6676, opened 2026-08-31
Migrate the channel config to the generated ChannelConfigWithInfo model

gpunto:
### Goal

Parse the channel config with the generated `ChannelConfigWithInfo` and delete the hand-written `ConfigDto`.

### Implementation

- Point `DownstreamChannelDto.config` at `ChannelConfigWithInfo` and drop `ConfigDto` with its mapper. The generated model, its four enum adapters and `ChannelConfigWithInfo.toDomain()` were already in place for the generated `ChannelResponse`, so this slice adds no model or mapper.
- Converge the two parallel config fixtures into one `configJson` and one `expectedConfig` serving both paths, with the real enum values (`disabled`, `flag`, `block`) in place of the invented `none` and `empty`.
- Rebuild `ChannelConfigWithInfo is correctly mapped to Config` to compare the whole `Config`, against a fixture where every value differs from the domain default it maps onto. It previously asserted five scalars against an all-true fixture. Differing from the default is what matters: a dropped field falls back to its `Config` default, so a fixture value that happens to equal that default hides the drop, and most of these default to `true`.
- Add the fields the wire always sends to the stale config fixtures in `EventChatJsonProvider` and `QueryGroupedChannelsResponseAdapterTest`.
- Make `DownstreamChannelDto.config` nullable, mapping an absent config to `Config()` as the generated `ChannelResponse` path already does. The backend leaves `config` nil for a channel whose type is no longer in the app config, and the field has no `omitempty`, so it arrives as `"config": null`. A non-null field there loses the entire response rather than degrading that one channel.

### Notes

The generated model marks 10 fields required that the hand-written DTO accepted as absent, including `created_at`, `updated_at`, `name`, and `quotes`, `reminders` and `count_messages`, which `ConfigDto` did not declare at all. Every one is a plain non-`omitempty` tag on the backend's channel config struct, so the current wire always sends them. That includes `delivery_events`, despite the DTO carrying a `= true` default for it.

The four enums resolve unknown values to `Unknown(value)` and the mapper passes `.value` to the domain, so an unrecognised mode maps to the same string the previous `String` field produced.

The generated model also parses eleven fields the domain `Config` has nowhere to hold, so they are read and dropped: `quotes`, `reminders`, `count_messages`, `blocklist`, `blocklists`, `automod_thresholds`, `allowed_flag_reasons`, `chat_preferences`, `grants`, `partition_size` and `partition_ttl`. `ConfigDto` did not declare any of them either, so nothing regressed, but the gap is now visible and tracked in [AND-1498](https://linear.app/stream/issue/AND-1498/surface-the-channel-config-fields-the-generated-model-parses-but).

### Testing

- Device probe over `queryChannels` (three channels) and the single-channel `watch`: every newly required field present, `automod=AI` carried through the enum, nested commands mapped, and the two endpoints produced an equal `Config`.
- Mutation sweep over the mapper, deleting each of the 27 mapped assignments in turn: every one is now caught by a test. Ten were unguarded before the fixture was rebuilt (`typingEventsEnabled`, `deliveryEventsEnabled`, `searchEnabled`, `isThreadEnabled`, `muteEnabled`, `urlEnrichmentEnabled`, `pushNotificationsEnabled`, `skipLastMsgUpdateForSystemMsgs`, `automod`, `sharedLocationsEnabled`).
- `ChannelConfigEnumParsingTest` pins the forward-compatibility claim above: unrecognised modes resolve to `Unknown(value)` and reach the domain as the raw wire string.
- `DownstreamChannelDtoAdapterTest` covers a channel arriving with `config: null`, which parses and maps to `Config()`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved channel configuration parsing and mapping to preserve additional settings, including message counting, quotes, reminders, shared locations, and user message reminders.
  * Updated channel configuration field handling to use consistent property names and values.
* **Tests**
  * Expanded coverage for channel configuration deserialization and domain mapping across all supported settings.


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
```diff
@@ -943,6 +943,11 @@ private fun createConfigJsonString() =
           "updated_at": "2020-06-29T06:14:28.000Z",
           "name": "team",
           "typing_events": true,
+          "count_messages": true,
```
aleksandar-apostolov, line 946: Did the uitests fixtures get missed? `http_channel.json` and `http_channels.json` are short a few of the now-required keys, so the mock responses won't parse. Snapshot tests are dispatch-only, so CI wouldn't have caught it.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/QueryGroupedChannelsResponseAdapterTest.kt
```diff
@@ -161,17 +175,17 @@ internal class QueryGroupedChannelsResponseAdapterTest {
...
         assertTrue(channelResponse.channel.config.reactions)
         assertTrue(channelResponse.channel.config.replies)
         assertTrue(channelResponse.channel.config.uploads)
-        assertTrue(channelResponse.channel.config.url_enrichment)
+        assertTrue(channelResponse.channel.config.urlEnrichment)
         assertTrue(channelResponse.channel.config.mutes)
-        assertEquals("infinite", channelResponse.channel.config.message_retention)
-        assertEquals(5000, channelResponse.channel.config.max_message_length)
+        assertEquals("infinite", channelResponse.channel.config.messageRetention)
+        assertEquals(5000, channelResponse.channel.config.maxMessageLength)
```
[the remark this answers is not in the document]
gpunto: Fair, the fixture gained those keys but nothing checked them. Added assertions for `deliveryEvents`, `quotes`, `countMessages`, `reminders`, `userMessageReminders`, `sharedLocations`, `skipLastMsgUpdateForSystemMsgs` and `name`, matching the fixture values. Flipping `delivery_events` in the fixture now fails the test, where before it passed.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/EventChatJsonProvider.kt
gpunto: Good catch, fixed it. Another reason to finish converting these ui tests to the e2e ones running in PRs

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/ChannelDtoTestData.kt
```diff
@@ -35,26 +34,29 @@ internal object ChannelDtoTestData {
...
+          "reminders": false,
+          "count_messages": true,
           "skip_last_msg_update_for_system_msgs": false,
           "polls": true,
           "message_retention": "retention",
           "max_message_length": 500,
-          "automod": "none",
-          "automod_behavior": "none",
-          "blocklist_behavior": "empty",
+          "automod": "disabled",
```
aleksandar-apostolov, line 57: Swapping the invented values for real ones is right, but it leaves the `Unknown(value)` path with no coverage anywhere in the repo. That is the guarantee you describe in the body: the old field was a raw `String` that took anything, the new one is a sealed enum, and `Unknown(s).value == s` is what stops a new server-side automod mode from changing behaviour. One fixture with an unrecognised value would lock it in.

gpunto: You are right, and it is worse than no coverage: the invented `none`/`empty` values were unrecognised, so the old fixture was exercising the `Unknown` path by accident, and swapping them for real values took that away.

Added `ChannelConfigEnumParsingTest`, which parses the shared fixture with `future_mode`, `future_behavior` and `future_blocklist` and asserts two things: the models resolve them to `Unknown(value)`, and the value still reaches the domain as the raw wire string. Deriving the JSON from `ChannelDtoTestData.configJson` keeps the required fields in one place, and it self-checks, since a replacement that stopped matching would leave a real value and fail the `Unknown` assertion.

That covers `automod`, `automod_behavior` and `blocklist_behavior`. `push_level` has the same sealed shape but is nullable and absent from the fixture, so I left it out rather than invent a payload for it.

gpunto:
Correcting something in the description above. It says the rebuilt mapping test means a flag read from the wrong source cannot pass, and that was only half true: it caught the swap I actually tested, where the two values differed, but not a field dropped entirely.

An exhaustive sweep over the mapper, deleting each mapped assignment in turn and checking whether any test notices, found **10 of the 27 fields unguarded**: `typingEventsEnabled`, `deliveryEventsEnabled`, `searchEnabled`, `isThreadEnabled`, `muteEnabled`, `urlEnrichmentEnabled`, `pushNotificationsEnabled`, `skipLastMsgUpdateForSystemMsgs`, `automod` and `sharedLocationsEnabled`.

The cause is that a dropped field falls back to its `Config` default, and ten of my fixture values happened to equal that default, most of them `true`. Alternating the booleans guarded against two flags being swapped with each other but not against one being dropped. The right property is that every value differs from the default it maps onto, which is strictly stronger.

Rebuilt the fixture on that rule in db05fa7: every boolean is the negation of its default, and `automod` is `AI` rather than `disabled`, which is what the device probe showed this app actually receives. All 27 fields are now caught.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/ChannelDtos.kt
```diff
@@ -41,7 +42,7 @@ internal data class DownstreamChannelDto(
     val members: List<DownstreamMemberDto> = emptyList(),
     val watchers: List<DownstreamUserDto> = emptyList(),
     val read: List<DownstreamChannelUserRead> = emptyList(),
-    val config: ConfigDto,
+    val config: ChannelConfigWithInfo,
```
andremion: `ChannelResponse` declares this same field as `ChannelConfigWithInfo? = null`, and the backend does send `config: null` when it cannot resolve the channel type (a custom type with no app config and no built-in default). With a non-null field here Moshi throws and the whole response is lost, not just that one channel. `ConfigDto` was non-null too, so nothing regressed. Still, now that this points at the generated model, would you make it nullable and map null to `Config()`? Happy to leave it if you'd rather keep the slice small.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ChannelConfigEnumParsingTest.kt
```diff
@@ -0,0 +1,62 @@
...
+            config.automodBehavior,
+        )
+        assertEquals(
+            ChannelConfigWithInfo.BlocklistBehavior.Unknown("future_blocklist"),
+            config.blocklistBehavior,
+        )
+    }
+
+    @Test
+    fun `Unrecognised config modes reach the domain as the raw wire value`() {
```
andremion, line 66: This one stays on the parsed model, and `Unknown(s).value == s` by construction, so the test above already covers it. What is not covered is the mapper passing `.value` through for an unrecognised value. Would you drop this test, or run the parsed config through `DomainMapping` and assert `config.automod == "future_mode"` instead?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
```diff
@@ -516,67 +516,39 @@ internal object Mother {
...
-        push_level = push_level,
+        userMessageReminders = randomBoolean(),
+        sharedLocations = randomBoolean(),
+        markMessagesPending = randomBoolean(),
     )
 
-    /**
-     * Provides a [QueryChannelsRequest] with random parameters (that can also be customized).
-     */
     fun randomQueryChannelRequest(
```
andremion, line 555: Small thing: the KDoc on `randomQueryChannelRequest` went away with the config mother, and this function is not related to the config. Put it back?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/ChannelDtoTestData.kt
```diff
@@ -100,12 +104,12 @@ internal object ChannelDtoTestData {
...
-        shared_locations = true,
-        mark_messages_pending = false,
+        userMessageReminders = false,
+        sharedLocations = true,
+        markMessagesPending = false,
+        messageRetention = "retention",
     )
 
-    @Language("JSON")
     val downstreamJson =
```
andremion, line 114: Not blocking: `downstreamJson` lost its `@Language("JSON")` while the three other JSON values in this file kept theirs. Any reason?

andremion (approved):
Looks good. One thing at the PR level: the model now parses `quotes`, `reminders` and `count_messages`, but `Config` has nowhere to put them, so they are read and then dropped. Is that tracked somewhere, or left for later on purpose? Rest is inline, all small.

gpunto:
On the PR-level point: it was not tracked, it is now, as [AND-1498](https://linear.app/stream/issue/AND-1498/surface-the-channel-config-fields-the-generated-model-parses-but). It is a bit wider than those three, the model parses eleven fields `Config` has nowhere to hold: `quotes`, `reminders`, `count_messages`, `blocklist`, `blocklists`, `automod_thresholds`, `allowed_flag_reasons`, `chat_preferences`, `grants`, `partition_size` and `partition_ttl`. `ConfigDto` did not declare any of them either, so none of them ever reached the domain and nothing regressed here. Deciding which belong on the public `Config` is its own change, with the DB entity and mapping to match, so I left it out of this slice and noted it in the Notes.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/ChannelDtos.kt
gpunto: Made it nullable in d4f5d1e6f1d, mapping an absent config to `Config()` so it matches the `ChannelResponse` path.

The mechanism is slightly different from what you describe, and it makes the case stronger. An unresolvable channel type on the create and `ApplyConfig` paths returns `ErrChannelTypeNotExist` rather than leaving `Config` nil, so the request fails outright. The null comes from `ResolvedConfig`, which returns nil when the channel type is no longer in the app config, and `channel_response.go` then leaves `config` nil. Its own comment says as much: "the response only lacks `config` when the channel type no longer exists". No `omitempty` on the field, so that serializes as `"config": null`.

So a channel whose type was removed from the app config is still returned, and a non-null field there costs the whole response. Covered by a new test in `DownstreamChannelDtoAdapterTest`; I did not re-probe it, since producing one would mean deleting a channel type on the shared app.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/ChannelConfigEnumParsingTest.kt
gpunto: You are right, it was a tautology. Rewritten in 82c3fd58410 to run the parsed config through `DomainMapping` and assert `config.automod == "future_mode"` and the other two.

Checked it now bites: replacing any of the three `.value` reads in the mapper with `""` fails the test, where the old version passed regardless.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/Mother.kt
gpunto: Restored. It was collateral from removing the config mother above it.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/ChannelDtoTestData.kt
gpunto: No reason, same collateral. Restored. The only other one that went is `channelResponseConfigJson`, which was deliberate since its fixture is gone.

PR #6677, opened 2026-09-01
Migrate the push preference chat preferences to the generated model

gpunto:
### Goal

Parse a push preference's chat preferences with the generated `ChatPreferencesResponse` and delete the hand-written `DownstreamChatPreferencesDto`.

### Implementation

- Point `DownstreamPushPreferenceDto.chat_preferences` at `ChatPreferencesResponse` and drop `DownstreamChatPreferencesDto`. Its mapper was identical field for field to the existing `ChatPreferencesResponse.toDomain()`, already used by `PushPreferencesResponse` and `ChannelPushPreferencesResponse`, so the duplicate goes with it and no new mapping code is needed.
- Rename `PushPreferenceDtos.kt` to `DownstreamPushPreferenceDto.kt`, since removing the class leaves a single declaration.
- Add a mapping test over the whole `PushPreference`, with the seven toggles alternating so a preference read from the wrong field cannot pass. `chat_preferences` had no coverage on this path before: neither `Mother.randomDownstreamPushPreferenceDto` nor `UserDtoTestData` populated it, and the closest existing test covered a different model while asserting two of the seven toggles.

The generated `ChatPreferences` model is a different, 8-field type belonging to channel config, not the push preference payload. `ChatPreferencesResponse` matches the removed DTO exactly, so no field is gained or lost.

### Testing

- Device probe: set channel chat preferences with the toggles alternating, then re-read through the channel query, which is the path that parses this field. All seven toggles round-tripped on both the write response and the re-read, with a populated `chatPreferences`.
- Mutation checks: dropping `chat_preferences` from the mapping, and reading one toggle from the wrong source field, each fail the new test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved push preference handling by consistently applying chat preference settings.
  - Preserved push levels, disabled-until dates, and individual notification toggles when converting preferences for app use.
- **Tests**
  - Added coverage to verify all chat preference options are correctly retained during push preference processing.

### DIFF
--- a/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
+++ b/stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
@@ -73,6 +73,7 @@ import io.getstream.chat.android.client.Mother.randomUserGroupMemberDto
 import io.getstream.chat.android.client.Mother.randomUserGroupResponse
 import io.getstream.chat.android.client.Mother.randomUserResponse
 import io.getstream.chat.android.client.api2.mapping.DomainMappingTest.Companion.toSortDomainArguments
+import io.getstream.chat.android.client.api2.model.dto.DownstreamPushPreferenceDto
 import io.getstream.chat.android.client.api2.model.dto.DownstreamUserGroupDto
 import io.getstream.chat.android.client.api2.model.dto.DownstreamUserGroupMemberDto
 import io.getstream.chat.android.client.api2.model.response.MessageResponse
@@ -89,6 +90,7 @@ import io.getstream.chat.android.models.ChannelMute
 import io.getstream.chat.android.models.ChannelTransformer
 import io.getstream.chat.android.models.ChannelUserRead
 import io.getstream.chat.android.models.ChatPreferenceToggle
+import io.getstream.chat.android.models.ChatPreferences
 import io.getstream.chat.android.models.Command
 import io.getstream.chat.android.models.Config
 import io.getstream.chat.android.models.Device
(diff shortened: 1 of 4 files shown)


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
```diff
@@ -1683,6 +1686,42 @@ internal class DomainMappingTest {
         )
     }
 
+    @Test
+    fun `DownstreamPushPreferenceDto keeps every chat preference toggle`() {
```
andremion, line 1690: Not blocking: the new test covers the mapping, but the change is really about which JSON keys Moshi reads, and nothing deserializes `chat_preferences` from JSON on this path. `DownstreamUserDtoAdapterTest` already parses `UserDtoTestData.downstreamJson`, and its `push_preferences` block has only `chat_level` and `disabled_until`. Adding the seven keys there plus the matching `ChatPreferencesResponse` in the expected DTO would cover it. Worth adding, or is the device probe enough for you?

andremion (approved):
Looks good. One optional test note inline.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DomainMappingTest.kt
gpunto: Good point, and worth doing rather than leaning on the probe: the mapping test would pass even if Moshi read none of those keys, which is the same shape as a draft bug I found yesterday where the mapper was correct and the JSON step silently dropped everything.

Added the seven keys to the `push_preferences` block in `UserDtoTestData.downstreamJson` with the matching `ChatPreferencesResponse` in the expected DTO, toggles alternating so a key read into the wrong property cannot look right. It turns out to cover more than just the user path: flipping one expected toggle fails 15 tests across seven adapter classes, since that fixture is embedded in the channel, thread, poll, reaction, member and message fixtures.

PR #6678, opened 2026-09-01
Build the outgoing message body with the generated MessageRequest model

gpunto:
### Goal

Build the outgoing message body with the generated `MessageRequest` and delete the hand-written `UpstreamMessageDto`.

### Implementation

- Point the `message` field of `SendMessageRequest`, `UpdateMessageRequest` and `TruncateChannelRequest` at `MessageRequest`, and move the four call sites in `MoshiChatApi` (send, update, draft create, truncate) onto `toMessageRequest()`. That mapper, the `MessageRequest` custom-flattening adapter and the `Type` adapter were already in place, serving the system-message paths.
- Add `DraftMessage.toMessageRequest()`, replacing `DraftMessage.toDto()`. The draft endpoint takes the same body: `CreateDraftRequest.Message` is the same `payload.MessageRequest` struct as sending a message.
- Delete `UpstreamMessageDto`, `Message.toDto()`, `UpstreamMessageDtoAdapter` and its registration. `MessageDtoAdapters.kt` becomes `DownstreamMessageDtoAdapter.kt`, since one declaration remains.
- Retarget the upstream fixtures and the adapter test to the new shape, keeping the `toJson` assertions that pin the serialised body.

The request no longer sends `args`, `cid`, `command`, `html`, `pinned_by`, `shadowed` or `thread_participants`. None are read on this path: `html` is `ignore_if_client_side`, `shadowed` is `readonly`, `cid` and `pinned_by` are `utils.JSONPlaceholder` in the block the struct comments as legacy fields kept only so the API does not break, and the other three are absent from the struct.

### Notes

Two fixture defaults let assertions pass that should not have. `randomMessage` defaults `extraData` to empty, so the existing `request.custom shouldBeEqualTo message.extraData` assertion held even with the mapper dropping `custom` entirely; `randomDraftMessage` defaults `silent` to `randomBoolean()`, so a mapper hardcoding it would have failed intermittently. Both are now pinned in the tests that depend on them.

### Testing

- Device probe over all four call sites against a throwaway channel: send, update, draft create and truncate were all accepted, with custom data flattened to the root and none of the dropped fields present. `poll_id` reached the server exactly once at the root, which is worth checking because it is now a declared field while the SDK still passes it through `extraData`.
- Mutation checks: dropping `custom` from either mapper, and hardcoding the draft's `silent`, each fail a test.
- The probe also surfaced [AND-1473](https://linear.app/stream/issue/AND-1473/draft-message-custom-data-is-dropped-when-reading-a-draft), custom data being dropped when reading a draft back. That is a pre-existing response-side bug, unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message request serialization for sending, updating, drafting, and truncating messages.
  * Ensured custom fields, mentions, threading, visibility, attachments, and related message properties are mapped consistently.
  * Updated message parsing and request handling to use the current request format.

* **Tests**
  * Expanded coverage for message request mapping, custom fields, and serialization behavior.


stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/MessageDtoTestData.kt
```diff
@@ -391,25 +391,26 @@ internal object MessageDtoTestData {
...
     @Language("JSON")
     val upstreamJson =
         """{
-          "attachments": [${AttachmentDtoTestData.json}],
-          "cid": "cid",
-          "html": "html",
           "id": "8584452-6d711169-0224-41c2-b9aa-1adbe624521b",
+          "text": "text",
           "type": "regular",
+          "attachments": [],
```
andremion: This fixture used to carry `AttachmentDtoTestData.json`, and it was the only place pinning how an attachment serialises inside the message body. With it empty, and `randomMessage` defaulting `attachments` to `listOf()`, replacing the attachment mapping with `emptyList()` in both mappers still passes the client suite. Since this is the PR that moves send and update off `AttachmentDto`, could you put `generatedAttachment` back here?

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DtoMappingTest.kt
```diff
@@ -269,7 +221,10 @@ internal class DtoMappingTest {
 
     @Test
     fun `Message is correctly mapped to the generated request model`() {
-        val message = randomMessage(type = MessageType.REGULAR)
+        val message = randomMessage(
+            type = MessageType.REGULAR,
+            extraData = mutableMapOf("messageKey" to "messageValue"),
```
andremion, line 234: `randomMessage` gives a non-null `sharedLocation`, so the structural comparison in the old test covered it. This one goes field by field and skips it, so a mapper without `sharedLocation` passes. Could you add an assertion for it?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/MessageDtos.kt
```diff
@@ -17,56 +17,9 @@
...
-
-/**
- * See [io.getstream.chat.android.client.parser2.adapters.DownstreamMessageDtoAdapter] for
- * special [extraData] handling.
- */
-@StreamHandsOff(
-    reason = "Field names can't be changed because [CustomObjectDtoAdapter] class uses reflections to add/remove " +
-        "content of [extraData] map",
-)
 @JsonClass(generateAdapter = true)
```
andremion, line 32: Any reason `DownstreamMessageDto` lost its `@StreamHandsOff` and the adapter comment? It still goes through `DownstreamMessageDtoAdapter`, and `CustomObjectDtoAdapter` reads the constructor parameter names by reflection, so the reason for it has not changed. The other DTOs on that adapter kept theirs.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
```diff
@@ -108,44 +107,6 @@ internal class DtoMapping(
...
-                    silent = silent,
-                    text = text,
-                    thread_participants = threadParticipants.map { it.toDto() },
-                    restricted_visibility = restrictedVisibility,
-                    shared_location = sharedLocation?.toDto(),
-                    extraData = extraData,
-                )
-            }
-
     internal fun Location.toDto(): UpstreamLocationDto = UpstreamLocationDto(
```
andremion, line 111: Small thing: the blank line before this function is gone, and the KDoc on `Mute.toDto()` further down was deleted too. Both look like leftovers from removing `Message.toDto()`. Could you restore them?

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
```diff
@@ -248,34 +209,23 @@ internal class DtoMapping(
...
-        shadowed = false,
-        show_in_channel = showInChannel,
-        silent = silent,
         text = text,
-        type = "regular",
-        thread_participants = emptyList(),
-        restricted_visibility = emptyList(),
-        shared_location = null,
-        extraData = extraData,
+        type = MessageRequest.Type.fromString("regular"),
```
andremion: Small thing: `MessageRequest.Type.Regular` instead of `fromString("regular")`? That's what the test uses.

andremion (approved):
Looks good. Two coverage notes and a couple of small things inline.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/parser2/testdata/MessageDtoTestData.kt
gpunto: Restored, and your reasoning held up in a way I had not expected. Putting the attachment back in the fixture fixed the serialisation pinning, but replacing the attachment mapping with `emptyList()` in either mapper still passed, because `randomMessage` and `randomDraftMessage` both default `attachments` to an empty list, so the mapping assertions could not tell the difference. Both mapping tests now start from a message that has an attachment, and each mutation fails a test.

I wrote the expected attachment out in `upstreamJson` rather than interpolating `AttachmentDtoTestData.json`. The request shape is not identical to the response one: the generated model serialises its empty `actions` and `fields`, and `file_size` comes back as an untyped `1.0`, so reusing the response fixture asserted a shape we never send.

stream-chat-android-client/src/test/java/io/getstream/chat/android/client/api2/mapping/DtoMappingTest.kt
gpunto: Added. Worth noting it was in the same state as the attachments: `sharedLocation` was asserted nowhere, so a mapper without it passed. Nulling it now fails a test.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/model/dto/MessageDtos.kt
gpunto: No reason, that was me deleting more than I meant to. Removing `UpstreamMessageDto` took the KDoc, the `@StreamHandsOff` and the `StreamHandsOff` import with it, and `DownstreamMessageDto` still needs all three for exactly the reason you give. Restored, pointing at `DownstreamMessageDtoAdapter`.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
gpunto: Same cause, both restored. The blank line and the `Mute.toDto()` KDoc were collateral from removing `Message.toDto()`. I compared function spacing against develop afterwards and there are no other losses.

stream-chat-android-client/src/main/java/io/getstream/chat/android/client/api2/mapping/DtoMapping.kt
gpunto: Done, `MessageRequest.Type.Regular`. It only ever went through `fromString` because it was a literal in the DTO it replaced.

